PR #8640, opened 2019-03-27
Add support for journaling path level configurations in meta master

cheng-chang:
1. adds new journal entries related to PathProperties
2. abstracts out a State class inside PathProperties to manage both the in-memory and persisted state
3. PathProperties is the first journaled state in MetaMaster

PR #8641, opened 2019-03-28
Add "fsadmin pathConf list" command

cheng-chang:
This command lists paths that have path level configurations.


shell/src/main/java/alluxio/cli/fsadmin/command/PathConfCommand.java
```diff
@@ -0,0 +1,122 @@
...
+      return mClass;
+    }
+  }
+
+  /** A map from sub-commands' names to the definition. */
+  private static final Map<String, SubCommand> SUB_COMMAND_MAP = new HashMap<>();
+
+  static {
+    for (SubCommand cmd : SubCommand.values()) {
+      SUB_COMMAND_MAP.put(cmd.getName(), cmd);
```
calvinjia: Do we need this map, could we use `valueOf` and some kind of case normalization?

shell/src/main/java/alluxio/cli/fsadmin/command/PathConfCommand.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Manages path level configuration.
+ */
+public final class PathConfCommand extends AbstractFsAdminCommand {
+  private enum SubCommand {
+    LIST("list", ListCommand.class),
+    ;
+
+    private final String mName;
+    private final Class<? extends Command> mClass;
```
aaudiber: consider storing a `Map<String, BiFunction<Context, AlluxioConfiguration, ? extends Command>>` instead. Then we don't need reflection at all

cheng-chang:
@calvinjia this command only lists the paths, there will be another command showing configurations for a path.

shell/src/main/java/alluxio/cli/fsadmin/command/PathConfCommand.java
cheng-chang: yes, possible, but then `LiSt` will also be recognized as the `list` command.

cheng-chang:
@calvinjia @aaudiber PTAL

shell/src/main/java/alluxio/cli/fsadmin/command/PathConfCommand.java
calvinjia: Got it, thanks.

calvinjia:
@cheng-chang If someone calls this command when they have thousands of paths configured, it will output all of them right? I am thinking of providing a prefix so they can see the ones under a subtree.

shell/src/main/java/alluxio/cli/fsadmin/command/PathConfCommand.java
```diff
@@ -0,0 +1,86 @@
...
+    StringBuilder usage = new StringBuilder(getCommandName());
+    for (String cmd : SUB_COMMANDS.keySet()) {
+      usage.append(" [").append(cmd).append("]");
+    }
+    return usage.toString();
+  }
+
+  @Override
+  public String getDescription() {
+    return "Manage path level configuration, see sub-commands' descriptions for more details.";
```
aaudiber: is there a way for users to discover the names of the subcommands?

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ListCommand.java
```diff
@@ -0,0 +1,67 @@
...
+import alluxio.wire.Configuration;
+
+import org.apache.commons.cli.CommandLine;
+
+import java.io.IOException;
+
+/**
+ * Lists paths that have path level configuration.
+ */
+public final class ListCommand extends AbstractFsAdminCommand {
```
aaudiber: does this need to extend `AbstractFsAdminCommand`? Usually things extending that can be called directly like `fsadmin list`, not via subcommand. See `ReportCommand` and its subcommands for example

shell/src/main/java/alluxio/cli/fsadmin/command/PathConfCommand.java
cheng-chang: `getUsage` lists the names of the subcommands.

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ListCommand.java
cheng-chang: by extending `AbstractFsAdminCommand`, the subcommand's interface and implementation can have the same structure as other commands. From a user's perspective, a subcommand is also a fsadmin command.

cheng-chang:
@calvinjia in the design, we expected only a small number of paths are configured with path defaults. So I think we're good at least for now.

aaudiber:
@cheng-chang can we add a test for the new command?

calvinjia:
@cheng-chang LGTM, please add testing

cheng-chang:
@aaudiber I'll add a test after https://github.com/Alluxio/alluxio/pull/8640 is merged, it depends on functionalities there.

cheng-chang:
@aaudiber @calvinjia PTAL

tests/src/test/java/alluxio/client/cli/fsadmin/pathconf/ListCommandIntegrationTest.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class ListCommandIntegrationTest extends PathConfTest {
+  @Test
+  public void list() {
+    int ret = mFsAdminShell.run("pathConf", "list");
+    Assert.assertEquals(0, ret);
+    String output = mOutput.toString();
+    String expected = DIR1 + "\n" + DIR2 + "\n";
+    Assert.assertEquals(expected, output);
+  }
```
calvinjia: The construction of the FS tree is in another class which makes this hard to follow. Also, should we test the case where there are some directories which do not have path level config?

cheng-chang: @calvinjia I made it more straightforward, PTAL.

calvinjia:

PR #8645, opened 2019-03-28
Improve the generate tarball script

gpang:
The Alluxio client jar does not depend on the hadoop distribution, so there is no need to build multiple tarballs for multiple distributions.


gpang:

PR #8655, opened 2019-03-30
Make RR policy work with replication

calvinjia:
Previously the round robin policy would ignore the input worker list. This causes problems with replication because the same block would be sent to the same worker multiple times.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
@@ -338,13 +338,13 @@ public final class AlluxioBlockStore {
 
     // Select N workers on different hosts where N is the value of initialReplicas for this block
     List<WorkerNetAddress> workerAddressList = new ArrayList<>();
+    List<BlockWorkerInfo> updatedInfos = Lists.newArrayList(workerOptions.getBlockWorkerInfos());
     for (int i = 0; i < initialReplicas; i++) {
       address = locationPolicy.getWorker(workerOptions);
       if (address == null) {
         break;
       }
       workerAddressList.add(address);
-      List<BlockWorkerInfo> updatedInfos = Lists.newArrayList(workerOptions.getBlockWorkerInfos());
       updatedInfos.removeAll(blockWorkersByHost.get(address.getHost()));
       workerOptions.setBlockWorkerInfos(updatedInfos);
     }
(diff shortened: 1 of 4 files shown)


core/client/fs/src/main/java/alluxio/client/block/policy/options/GetWorkerOptions.java
```diff
@@ -53,7 +53,7 @@ public BlockInfo getBlockInfo() {
   /**
    * @return the list of block worker infos
    */
-  public Iterable<BlockWorkerInfo> getBlockWorkerInfos() {
+  public List<BlockWorkerInfo> getBlockWorkerInfos() {
```
calvinjia: Questionable signature change, we previously modified the underlying object. Alternatively we could return something immutable and use the set API.

calvinjia: Updated, thanks.

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
```diff
@@ -345,7 +344,9 @@ public BlockOutStream getOutStream(long blockId, long blockSize, OutStreamOption
         break;
       }
       workerAddressList.add(address);
-      blockWorkers.removeAll(blockWorkersByHost.get(address.getHost()));
+      List<BlockWorkerInfo> updatedInfos = Lists.newArrayList(workerOptions.getBlockWorkerInfos());
```
ZacBlanco: Instead of making a copy of the current set of `BlockWorkerInfos` here could we utilize a single reference to a set?

ZacBlanco: Also `Lists.newArrayList` [shouldn't be used](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Lists.html#newArrayList--) any more

> Note for Java 7 and later: this method is now unnecessary and should be treated as deprecated. Instead, use the ArrayList constructor directly, taking advantage of the new "diamond" syntax.

calvinjia: `getBlockWorkerInfos` returns an iterable and not a collection so we cannot use the ArrayList constructor.

calvinjia: I updated it to use a single copied list. I still copied the list once since I think this is a special case and wanted to make it explicit.

ZacBlanco: Cool, thanks!

calvinjia:

PR #8663, opened 2019-04-02
Wait for processes to start by default in alluxio-start.sh

ZacBlanco:
This change will modify the default behavior of the
alluxio-start.sh script. This change moving forward will make it
easier and more straightforward to start an alluxio cluster and
know when it is ready to start serving requests.

The -m flag has been replaced in alluxio-start.sh with the -a flag.
-a is the logical opposite of what -m performed. Without
specifying any flags, the script will monitor each process to make
sure it has started properly. Otherwise, if the -a (async) flag is
specified, then the processes will start without monitoring and
the script will complete successfully as long as each process
was at least attempted to be started (This is the current
behavior)

The job master/worker functions were also added to the
alluxio-monitor script for the (all) and (local) cases.


LuQQiu:
@ZacBlanco for a small cluster, monitor by default is always a good idea.
When I do the local experiment, I have some concerns:
(1) processes start sequentially, do we need the sequential start behavior?
(2) when i run alluxio-start.sh all, the console output looks like
```
......
Starting to monitor all remote services.
Error: Invalid NODE_TYPE: masters
```
Not sure what error the monitor service threw.
(3) will the start script take a long time and looks like hanging when we start a large cluster?

Thanks!

bin/alluxio-monitor.sh
```diff
@@ -209,14 +209,18 @@ main() {
...
     all)
       prepare_monitor "Starting to monitor ${CYAN}all remote${NC} services."
-      run_monitors "master" "${HOSTS}" "${MODE}"
-      run_monitors "worker" "${HOSTS}" "${MODE}"
-      run_monitors "proxy" "${HOSTS}" "${MODE}"
+      run_monitors "masters" "${HOSTS}" "${MODE}"
+      run_monitors "job_masters" "${HOSTS}" "${MODE}"
+      run_monitors "workers" "${HOSTS}" "${MODE}"
+      run_monitors "job_workers" "${HOSTS}" "${MODE}"
+      run_monitors "proxies" "${HOSTS}" "${MODE}"
```
aaudiber: this code looks strange - isn't the set of hosts different for workers vs masters vs proxies? But they all use `${HOSTS}` here.

ZacBlanco: When passing arguments to `run_monitors`, an empty 2nd argument will result in pulling the the node list from the respective file  under the conf directory (See https://github.com/Alluxio/alluxio/blob/e52d802d352802fcf378017158d677c79b34b87b/bin/alluxio-monitor.sh#L124-L149). I will update these to pass `""` instead so that the node list will be pulled from the configuration files.

ZacBlanco: After doing some more digging - we actually don't need to change it from `${HOSTS}` to `""` because when `alluxio-start.sh` calls `alluxio-monitor.sh` it will only ever populate the `${HOSTS}` argument if `alluxio-start.sh` is restarting the workers, otherwise it is always an empty string.

ZacBlanco:
@LuQQiu thanks for the comments!

To address your concerns

1) We don't start monitoring processes until we've actually executed the commands which kick off the process. In theory - the monitors could run in any order, and if everything starts successfully it shouldn't matter. We'll catch the errors regardless.

However, I think there is a logical order that we should monitor processes by. I would first make sure the master, then job masters start. Then workers, job workers, and finally proxies. (This is the order we currently monitor by). It makes sense logically to check if the masters are up before the workers, and then check the workers once we know the masters have started successfully. The proxy is separate, and I think it is okay to check last as it's typically a non-critical part of the system.

2) You're correct there was an error - It's been fixed with a recent commit

3) This script will only take as long as it takes for each node to start and for the monitor to check it's been successfully started. For a large cluster regardless of monitoring or not, it will typically take a bit of time. The monitoring itself should not impose much of a bottleneck.

@aaudiber feel free to chime in

aaudiber:
@ZacBlanco Have you had a chance to test this out on a cluster with 3+ nodes, and different hostnames in `conf/masters` vs `conf/workers`? Unfortunately we don't currently have a good testing framework for shell components, so manual testing is required to verify that things work. Especially since this is being promoted from a little-known option to the default value, I'd like to do some additional testing before making the change

LuQQiu:
@ZacBlanco Thanks for fixing the error and explaining the monitor process.
I tried on my local computer again, all the other monitors work good except the job master
```
$./bin/alluxio-start.sh all
Executing the following command on all master nodes and logging to /Users/alluxio/alluxioFolder/alluxio/logs/task.log: /Users/alluxio/alluxioFolder/alluxio/bin/alluxio-start.sh job_master
Waiting for tasks to finish...
Task on 'localhost' fails, exit code: 1
```

The exception message in the task.log is 
```
The job_master @ Alluxios-MacBook-Pro-4.local is not serving requests.
```

When we start to use `embedde journal` by default, the alluxio master takes a relatively long time to start the raft journal system. Only when the raft journal system is ready, master and job_master can start serving their requests.

```
    HealthCheckClient client = new JobMasterHealthCheckClient(
        NetworkAddressUtils.getConnectAddress(NetworkAddressUtils.ServiceType.JOB_MASTER_RPC, conf),
        () -> new ExponentialBackoffRetry(50, 100, 2), conf);
```
The retry time may be shorter than the time job_master need to start serving.

ZacBlanco:
@ggezer Could you take a look at the health check clients and make sure I'm using the gRPC `ServiceType`s correctly?

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
```diff
@@ -166,39 +229,56 @@ public void run() {
...
-      Future<?> masterServingCheckFuture = mExecutorService.submit(
-              new MasterServingCheckRunnable());
+      ServiceType rpcService = ServiceType.FILE_SYSTEM_MASTER_CLIENT_SERVICE;
+      InetSocketAddress connectAddr = NetworkAddressUtils
+          .getConnectAddress(NetworkAddressUtils.ServiceType.MASTER_RPC, mConf);
+      if (MasterType.JOB_MASTER == mAlluxioMasterType) {
+        rpcService = ServiceType.JOB_MASTER_CLIENT_SERVICE;
+        connectAddr = NetworkAddressUtils
+            .getConnectAddress(NetworkAddressUtils.ServiceType.JOB_MASTER_RPC, mConf);
+      }
```
ggezer: Can you turn this into a switch statement with a default case in order to harden it against future master types

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
```diff
@@ -78,38 +115,64 @@ public Builder withConfiguration(AlluxioConfiguration alluxioConf) {
     }
 
     /**
-     * @param alluxioMasterName the Alluxio master process name
+     * @param masterType the Alluxio master type
      * @return an instance of the builder
      */
-    public Builder withAlluxioMasterName(String alluxioMasterName) {
-      mAlluxioMasterName = alluxioMasterName;
+    public Builder withAlluxioMasterName(MasterType masterType) {
```
ggezer: withAlluxioMasterType() ?

shell/src/main/java/alluxio/master/job/JobMasterRpcHealthCheckClient.java
```diff
@@ -25,8 +25,8 @@
 /**
  * JobMasterHealthCheckClient check if job_master is serving RPC.
  */
-public class JobMasterHealthCheckClient extends RpcPortHealthCheckClient {
-  private static final Logger LOG = LoggerFactory.getLogger(JobMasterHealthCheckClient.class);
+public class JobMasterRpcHealthCheckClient extends RpcPortHealthCheckClient {
```
ggezer: Actually, regular health-check clients are not doing RPC health check. Do we even need these separate classes?

ZacBlanco: Sorry, I don't quite understand. Are you saying that all the other health check clients aren't making any RPC's at all?

shell/src/main/java/alluxio/worker/AlluxioWorkerMonitor.java
```diff
@@ -43,7 +43,7 @@ public static void main(String[] args) {
 
     HealthCheckClient client = new WorkerHealthCheckClient(
         NetworkAddressUtils.getConnectAddress(NetworkAddressUtils.ServiceType.WORKER_RPC, conf),
-        () -> new ExponentialBackoffRetry(50, 100, 2), conf);
+        () -> new ExponentialBackoffRetry(50, 500, 10), conf);
```
aaudiber: Could we move this supplier to a constant with a descriptive name like `TWO_MINUTE_EXP_BACKOFF`? It difficult to tell from the numbers how long the retries go for

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
```diff
@@ -166,39 +229,65 @@ public void run() {
...
+      Future<?> masterServingFuture = mExecutorService.submit(masterRpcCheck);
       if (mProcessCheck) {
         Future<?> processCheckFuture = mExecutorService.submit(
-                new ProcessCheckRunnable(mAlluxioMasterName));
+                new ProcessCheckRunnable(mAlluxioMasterType.getClassName()));
         CommonUtils.sleepMs(Constants.SECOND_MS);
-        while (!masterServingCheckFuture.isDone()) {
-          if (processCheckFuture.isDone()) {
-            throw new IllegalStateException("One or more master processes are not running");
+        // If in HA mode, can't check the RPC service, because the server may not have started
```
aaudiber: server -> service

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
```diff
@@ -78,38 +115,64 @@ public Builder withConfiguration(AlluxioConfiguration alluxioConf) {
...
-      return new MasterHealthCheckClient(mAlluxioMasterName, mProcessCheck, mConf);
+      return new MasterHealthCheckClient(mAlluxioMasterType, mProcessCheck, mPolicySupplier, mConf);
     }
   }
 
   /**
    * Runnable for checking if the AlluxioMaster is serving RPCs.
    */
-  public final class MasterServingCheckRunnable implements Runnable {
+  public final class MasterServingHealthCheck extends RpcPortHealthCheckClient implements Runnable {
```
aaudiber: since this doesn't check an RPC port, it's strange for it to extend RpcPortHealthCheckClient

ZacBlanco: It actually does check an RPC port. it uses the `super.isServing` method in the `run` function

bin/alluxio-monitor.sh
```diff
@@ -108,38 +108,57 @@ run_monitor() {
...
       print_node_logs "${node_type}"
       return 1
     fi
   fi
-  echo -e "${BOLD}---${END_BOLD} ${GREEN}[ OK ] The ${node_type} service @ $(hostname -f) is in an healthy state.${NC}"
+  echo -e "${WHITE}---${NC} ${GREEN}[ OK ]${NC} The ${CYAN}${node_type}${NC} service @ ${PURPLE}$(hostname -f)${NC} is in a healthy state.${NC}"
   return 0
 }
 
+# $1 --> The path to the workers file
```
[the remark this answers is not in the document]
ZacBlanco: yes, will update

shell/src/main/java/alluxio/worker/AlluxioWorkerMonitor.java
ZacBlanco: For the exponential ones I just get a rough estimate for the max time by doing `maxWait * numRetries`. Technically it will always be less than that, but it gives a good enough ballpark if the min/max are close to one another.

As for adding a constant, I can find a place to put one. Especially because the masters also share a similar time.

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
ZacBlanco: fixed

shell/src/main/java/alluxio/worker/AlluxioWorkerMonitor.java
ZacBlanco: added constants in the Master/WorkerMonitor classes

bin/alluxio-monitor.sh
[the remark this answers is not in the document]
ZacBlanco: fixed

aaudiber:

PR #8669, opened 2019-04-02
Add a hint to check task.log for errors

cheng-chang:
When tasks fail, users might not know where to find errors.


cheng-chang:
@aaudiber PTAL

aaudiber:
@cheng-chang Please back-port this to branch-1.8 as well

PR #8679, opened 2019-04-03
Fix NPE in server address

bf8086:
Fixed an NPE in `GrpcServerAddress` when the address is passed in unresolved.

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcServerAddress.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcServerAddress.java
@@ -27,7 +27,7 @@ public class GrpcServerAddress {
    * @param socketAddress physical address
    */
   public GrpcServerAddress(InetSocketAddress socketAddress) {
-    mHostName = socketAddress.getAddress().getHostName();
+    mHostName = socketAddress.getHostName();
     mSocketAddress = socketAddress;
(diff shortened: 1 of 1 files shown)


bf8086:
@ggezer PTAL.

ggezer:
Thanks for fixing this.
Do we always create InetSocketAddress with a host name? 
There could be problems with Kerberos if there are places where we only specify the port and assume local address. (If Kerberos principals are configured to match reverse lookup name of hosts)

bf8086:
> Thanks for fixing this.
> Do we always create InetSocketAddress with a host name?
> There could be problems with Kerberos if there are places where we only specify the port and assume local address. (If Kerberos principals are configured to match reverse lookup name of hosts)

From Javadoc `InetSocketAddress.getHostName` will return the original host name if given, or in the case of an ip address, a reverse lookup will be used to acquire the hostname.

bf8086:
Jenkins, test this please.

ggezer:
So the earlier behavior is that we take the reverse lookup name even if hostname was provided. Now always using the given hostname may have effects on certificate resolution. We may want to run some tests with this before checking in.

ggezer:
LGTM.

gpang:

PR #8680, opened 2019-04-03
Fix timeouts in multi process checkpoint test

aaudiber:
2ms is too low for the session timeout. We were getting warnings saying
session timeout [2] is less than connection timeout [15000]

@LuQQiu @cheng-chang I think this could be the cause of the flaky test failures you've seen


aaudiber:

PR #8681, opened 2019-04-03
Refactor filesystem, fsContext into JobMaster/Worker contexts

ZacBlanco:
This change moves the creation of FileSystem/Context out of
the job definitions and into the respective master/worker
processes. The processes should aim to create clients as few
times as possible because of all the underlying threadpools
and resources that are needed to instantiate a client.

The JobMaster and JobWorkers now pass in a single instance
of a filesystem client+context when selecting executors and
running tasks. This greatly reduces the amount of resources
that the job master and workers consume when there are lots
of jobs being executed.

### DIFF
--- a/job/server/src/main/java/alluxio/job/JobMasterContext.java
+++ b/job/server/src/main/java/alluxio/job/JobMasterContext.java
@@ -11,6 +11,8 @@
 
 package alluxio.job;
 
+import alluxio.client.file.FileSystem;
+import alluxio.client.file.FileSystemContext;
 import alluxio.underfs.UfsManager;
 
 import javax.annotation.concurrent.ThreadSafe;
@@ -20,16 +22,39 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 @ThreadSafe
 public final class JobMasterContext {
+  private final FileSystem mFileSystem;
+  private final FileSystemContext mFsContext;
   private final long mJobId;
   private final UfsManager mUfsManager;
 
   /**
+   * Creates a new instance of {@link JobMasterContext}.
+   *
+   * @param filesystem the Alluxio client used to contact the master
+   * @param fsContext the filesystem client's underlying context
    * @param jobId the job id
    * @param ufsManager the UFS manager
    */
-  public JobMasterContext(long jobId, UfsManager ufsManager) {
+  public JobMasterContext(FileSystem filesystem, FileSystemContext fsContext, long jobId,
+      UfsManager ufsManager) {
+    mFsContext = fsContext;
+    mFileSystem = filesystem;
     mJobId = jobId;
     mUfsManager = ufsManager;
+  }
+
+  /**
+   * @return the {@link FileSystem} client
+   */
+  public FileSystem getFileSystem() {
+    return mFileSystem;
+  }
+
+  /**
+   * @return the FileSystemContext for the client
+   */
+  public FileSystemContext getFsContext() {
+    return mFsContext;
   }
 
   /**
--- a/job/server/src/main/java/alluxio/job/JobWorkerContext.java
+++ b/job/server/src/main/java/alluxio/job/JobWorkerContext.java
@@ -11,6 +11,8 @@
 
 package alluxio.job;
 
+import alluxio.client.file.FileSystem;
+import alluxio.client.file.FileSystemContext;
 import alluxio.underfs.UfsManager;
 
 import javax.annotation.concurrent.ThreadSafe;
@@ -20,6 +22,8 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 @ThreadSafe
 public final class JobWorkerContext {
+  private final FileSystem mFileSystem;
+  private final FileSystemContext mFsContext;
   private final long mJobId;
   private final int mTaskId;
   private final UfsManager mUfsManager;
@@ -27,14 +31,33 @@ public final class JobWorkerContext {
   /**
    * Creates a new instance of {@link JobWorkerContext}.
    *
+   * @param filesystem the filesystem client that is used when running jobs
+   * @param fsContext the filesystem client's underlying context
    * @param jobId the job id
    * @param taskId the task id
    * @param ufsManager the UFS manager
    */
-  public JobWorkerContext(long jobId, int taskId, UfsManager ufsManager) {
+  public JobWorkerContext(FileSystem filesystem, FileSystemContext fsContext, long jobId,
+      int taskId, UfsManager ufsManager) {
+    mFsContext = fsContext;
+    mFileSystem = filesystem;
     mJobId = jobId;
     mTaskId = taskId;
     mUfsManager = ufsManager;
+  }
+
+  /**
+   * @return the {@link FileSystem} client
+   */
+  public FileSystem getFileSystem() {
+    return mFileSystem;
+  }
+
+  /**
+   * @return the FileSystemContext for the client
+   */
+  public FileSystemContext getFsContext() {
+    return mFsContext;
   }
 
   /**
--- a/job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
+++ b/job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
@@ -15,11 +15,9 @@ import alluxio.AlluxioURI;
 import alluxio.client.WriteType;
 import alluxio.client.block.AlluxioBlockStore;
 import alluxio.client.block.BlockWorkerInfo;
-import alluxio.client.file.BaseFileSystem;
 import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileOutStream;
 import alluxio.client.file.FileSystem;
-import alluxio.client.file.FileSystemContext;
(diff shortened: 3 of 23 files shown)


ZacBlanco:
For context - Before this PR I tested this change locally by creating a directory tree with 5000 files with write type `MUST_CACHE`. I attempted to persist the directory with `./bin/alluxio fs persist /file-tree` and the master/job master basically came to halt and couldn't handle any more than a hundred or so persist jobs. Logs would fill with things errors like "too many open files", "Job failed", etc.

After this change I was able to run this without any problem, even adding more concurrent persist jobs (`./bin/alluxio fs persist -p 500`, 500 concurrent persists) executed without any hiccups

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -91,6 +93,17 @@ public WorkerNetAddress getFieldValue(MasterWorkerInfo o) {
...
+  /**
+   * The Filesystem context that the job master uses for its client.
+   */
+  private final FileSystemContext mFsContext =
+      FileSystemContext.create(ServerConfiguration.global());
+
+  /**
+   * The FileSystem client the job master uses to select executors for jobs.
+   */
+  private final FileSystem mFileSystem = FileSystem.Factory.create(mFsContext);
```
calvinjia: Any reason for now doing this in the constructor similar to `JobWorker`?

job/server/src/main/java/alluxio/job/JobMasterContext.java
```diff
@@ -20,18 +22,41 @@
...
   /**
+   * Creates a new instance of {@link JobMasterContext}.
+   *
+   * @param filesystem the Alluxio client used to contact the master
+   * @param fsContext the filesystem client's underlying context
    * @param jobId the job id
    * @param ufsManager the UFS manager
    */
-  public JobMasterContext(long jobId, UfsManager ufsManager) {
+  public JobMasterContext(FileSystem filesystem, FileSystemContext fsContext, long jobId,
```
calvinjia: Why does the `JobMasterContext` have a `jobId`? Is this more of a `JobContext`?

job/server/src/main/java/alluxio/job/JobWorkerContext.java
```diff
@@ -20,23 +22,44 @@
...
    *
+   * @param filesystem the filesystem client that is used when running jobs
+   * @param fsContext the filesystem client's underlying context
    * @param jobId the job id
    * @param taskId the task id
    * @param ufsManager the UFS manager
    */
-  public JobWorkerContext(long jobId, int taskId, UfsManager ufsManager) {
+  public JobWorkerContext(FileSystem filesystem, FileSystemContext fsContext, long jobId,
+      int taskId, UfsManager ufsManager) {
```
calvinjia: Why does the `JobWorkerContext` have a `jobId` and `taskId`? Is this more of a `JobTaskContext`?

job/server/src/main/java/alluxio/job/JobWorkerContext.java
```diff
@@ -20,23 +22,44 @@
...
+  private final FileSystemContext mFsContext;
   private final long mJobId;
   private final int mTaskId;
   private final UfsManager mUfsManager;
 
   /**
    * Creates a new instance of {@link JobWorkerContext}.
    *
+   * @param filesystem the filesystem client that is used when running jobs
+   * @param fsContext the filesystem client's underlying context
```
calvinjia: This doesn't guarantee that the context is the file system's underlying context. But I don't think we can just pass in a context and create a filesystem because the caching does not work on that code path.

job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
```diff
@@ -104,30 +102,15 @@
...
-   * @param fsContext the {@link FileSystemContext} used by the {@link FileSystem}
-   * @param fileSystem the {@link FileSystem} client
-   */
-  public MigrateDefinition(FileSystemContext fsContext, FileSystem fileSystem) {
-    mFsContext = fsContext;
-    mFileSystem = fileSystem;
-  }
-
-  private void checkMigrateValid(MigrateConfig config) throws Exception {
+  private void checkMigrateValid(MigrateConfig config, FileSystem fs) throws Exception {
```
calvinjia: Should we always pass in `Job*Context`?

job/server/src/main/java/alluxio/master/job/JobCoordinator.java
```diff
@@ -67,10 +77,13 @@
    */
   private UfsManager mUfsManager;
 
-  private JobCoordinator(CommandManager commandManager, UfsManager ufsManager,
+  private JobCoordinator(CommandManager commandManager, FileSystem filesystem,
```
calvinjia: Should this take a `JobMasterContext`?

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -227,8 +240,9 @@ public synchronized long run(JobConfig jobConfig)
       }
     }
     long jobId = mJobIdGenerator.getNewJobId();
-    JobCoordinator jobCoordinator = JobCoordinator.create(mCommandManager, mUfsManager,
-        getWorkerInfoList(), jobId, jobConfig, (jobInfo) -> {
+    JobCoordinator jobCoordinator = JobCoordinator.create(mCommandManager, mFileSystem, mFsContext,
```
calvinjia: We could create the `JobMasterContext` object here?

job/server/src/main/java/alluxio/job/JobMasterContext.java
ZacBlanco: The of executing a job is:

- get a scheduled Job ID
- Select the executors for that job (`selectExecutors`)
- send the jobconfig to worker
- Worker executes the job based on the config (`runTask`)

the `selectExecutors` portion takes an argument `JobMasterContext` which provides the necessary information for the Job Master to determine which job worker to run the job on. I assume the original author called it JobMasterContext because the JobMaster is the only part that executes `selectExecutors` and the `runTask` only runs on the Job worker so it takes a `JobWorkerContext`

job/server/src/main/java/alluxio/job/JobWorkerContext.java
ZacBlanco: see explanation from `JobMasterContext`, but essentially something like that. maybe `RunTaskContext` might be more appropriate

job/server/src/main/java/alluxio/job/JobWorkerContext.java
ZacBlanco: I guess we _could_ only pass in the `FileSystemContext` instead  of both, but then we still have to call `FileSystem.Factory.create(FileSystemContext)`.

That isn't as bad as calling `FileSystem.Factory.create(AlluxioConfiguration)` because we aren't creating any new thread pools but it would be unnecessarily going through and iterating over all of our configuration objects (see [here](https://github.com/Alluxio/alluxio/blob/master/core/client/fs/src/main/java/alluxio/client/file/FileSystem.java#L134-L148) for what I mean).

Passing both is better performance-wise, but we technically don't have the guarantee that the fsContext belongs to the client.

Personally I really wish we had some kind of way to access the context from outside the client, but that probably requires re-architecting, and probably some kind of wrapper object

job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
ZacBlanco: because this is a private method I only passed the necessary objects

job/server/src/main/java/alluxio/master/job/JobCoordinator.java
ZacBlanco: Not in the current architecture - but if we were to redesign the classes and objects in the Job service, then I would say it's a possibility

ZacBlanco: I see what you're getting at how the JobMasterContext should be created once, instead of making many of them. I agree. However, I usually like the keep bug fixes separate from refactors. I can roll both into this PR if you want though

calvinjia: Let's get the bug fix in first and then address style/code structure comments.

calvinjia:
@ZacBlanco Could you create a follow up PR which addresses the following two concerns:

- Use `Job*Context` as early as possible in the call stack
- Rename `Job*Context` to more suitable names, if any

job/server/src/main/java/alluxio/master/job/JobMaster.java
ZacBlanco: I just modified JobMaster so that the client/context are passed in the same way as the JobWorker

ZacBlanco:
Will do, thanks!

gpang:

Xenorith:

gpang:

gpang:

calvinjia:

PR #8682, opened 2019-04-03
Add an option to cp command to preserve file attributes

bf8086:
Sometimes user wants to preserve the original attributes of files and directories when using `cp` command. This change add a new `-p` option to the `cp` command to enable such functionality. Note that user will need to have the required permissions to perform the operation.

https://github.com/Alluxio/alluxio/issues/8685

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
@@ -112,7 +112,7 @@ public final class CpCommand extends AbstractFileSystemCommand {
   private int mCopyFromLocalBufferSize = COPY_FROM_LOCAL_BUFFER_SIZE_DEFAULT;
   private int mCopyToLocalBufferSize = COPY_TO_LOCAL_BUFFER_SIZE_DEFAULT;
   private int mThread = Runtime.getRuntime().availableProcessors() * 2;
-  private boolean mPreserveAttributes = false;
+  private boolean mPreservePermissions = false;
 
   /**
    * A thread pool executor for asynchronous copy.
@@ -329,7 +329,7 @@ public final class CpCommand extends AbstractFileSystemCommand {
       }
     }
     if (cl.hasOption(PRESERVE_OPTION.getLongOpt())) {
-      mPreserveAttributes = true;
+      mPreservePermissions = true;
     }
   }
 
@@ -556,7 +556,7 @@ public final class CpCommand extends AbstractFileSystemCommand {
    */
   private void preserveAttributes(AlluxioURI srcPath, AlluxioURI dstPath)
       throws IOException, AlluxioException {
-    if (mPreserveAttributes) {
+    if (mPreservePermissions) {
       URIStatus srcStatus = mFileSystem.getStatus(srcPath);
       mFileSystem.setAttribute(dstPath, SetAttributePOptions.newBuilder()
           .setOwner(srcStatus.getOwner())
--- a/shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
@@ -26,6 +26,9 @@ import alluxio.exception.FileAlreadyExistsException;
 import alluxio.exception.FileDoesNotExistException;
 import alluxio.exception.InvalidPathException;
 import alluxio.exception.status.InvalidArgumentException;
+import alluxio.grpc.SetAclAction;
+import alluxio.grpc.SetAttributePOptions;
+import alluxio.security.authorization.Mode;
 import alluxio.util.io.PathUtils;
 
 import com.google.common.base.Joiner;
@@ -98,10 +101,18 @@ public final class CpCommand extends AbstractFileSystemCommand {
               + "default is 8MB when copying from local, "
               + "and 64MB when copying to local")
           .build();
+  private static final Option PRESERVE_OPTION =
+      Option.builder("p")
+          .longOpt("preserve")
+          .required(false)
+          .desc("Preserve file attributes when copying files. "
+              + "All ownership, permissions and ACLs will be preserved")
+          .build();
 
   private int mCopyFromLocalBufferSize = COPY_FROM_LOCAL_BUFFER_SIZE_DEFAULT;
   private int mCopyToLocalBufferSize = COPY_TO_LOCAL_BUFFER_SIZE_DEFAULT;
   private int mThread = Runtime.getRuntime().availableProcessors() * 2;
+  private boolean mPreserveAttributes = false;
 
   /**
    * A thread pool executor for asynchronous copy.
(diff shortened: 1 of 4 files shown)


bf8086:
@aaudiber @gpang PTAL.

aaudiber:
@bf8086 can you create an issue to track this new feature?

bf8086:
> With this approach there will be a window where the copied file exists under the current user's permissions, between doing the copy and calling `preserveAttributes`. Is there a way to avoid this? How does HDFS handle this situation?

This is exactly how HDFS does it. We can potentially build the `setAttributeOptions` into the create file API. I am not convinced to add such complexity given that we can achieve the same result with combination of existing APIs, unless there is an important use case.

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
```diff
@@ -138,6 +161,86 @@ public void copyWildcard() throws Exception {
...
+    int ret = mFsShell.run("cp", "-R",  "-p", testDir, newDir);
+    AlluxioURI dstDir = new AlluxioURI(newDir);
+    Assert.assertEquals(0, ret);
+    Assert.assertTrue(mFileSystem.exists(dstDir));
+    verifyAttributes(srcDir, dstDir);
+    verifyAttributes(srcDir.join(subDir), dstDir.join(subDir));
+    verifyAttributes(srcDir.join(file), dstDir.join(file));
+  }
+
+  private void verifyAttributes(AlluxioURI src, AlluxioURI dst)
```
aaudiber: `verifyEqualAttributes`?

calvinjia:
One way around this is to first copy to a temporary location, set permissions, then rename to the final destination.

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
```diff
@@ -138,6 +161,86 @@ public void copyWildcard() throws Exception {
         equals(new AlluxioURI("/copy/foobar3"), new AlluxioURI(testDir + "/bar/foobar3")));
   }
 
+  /**
+   * Tests copying a file with attributes preserved.
+   */
+  @Test
+  public void copyFileWithPreservedAttributes() throws Exception {
```
aaudiber: can we also test that other attributes like TTL are *not* preserved

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
```diff
@@ -138,6 +161,86 @@ public void copyWildcard() throws Exception {
...
+    InstancedConfiguration conf = new InstancedConfiguration(ServerConfiguration.global());
+    conf.set(PropertyKey.USER_FILE_WRITE_TYPE_DEFAULT, "MUST_CACHE");
```
aaudiber: why do we need MUST_CACHE?

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
```diff
@@ -98,10 +101,18 @@
               + "default is 8MB when copying from local, "
               + "and 64MB when copying to local")
           .build();
+  private static final Option PRESERVE_OPTION =
+      Option.builder("p")
+          .longOpt("preserve")
+          .required(false)
+          .desc("Preserve file attributes when copying files. "
```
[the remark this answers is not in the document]
bf8086: We can either make this command also preserve TTL, pin state, and replication parameters, or we can just keep the current list and update the description/tests, thoughts?

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
bf8086: Because persisted file will fail to change owner unless the test is ran with root account.

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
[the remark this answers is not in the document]
aaudiber: I'd like to keep the functionality as-is since it mirrors what hdfs does. We just need to update the docs (and potentially command name) to make it clear that we're only preserving permissions, not other attributes. We can add additional options for preserving ttl/pin/replication in the future if there's a demand for it.

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
aaudiber: ahhh, makes sense. Can we add a comment for this?

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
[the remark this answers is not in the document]
bf8086: Updated description.

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
bf8086: Renamed.

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
bf8086: Updated.

tests/src/test/java/alluxio/client/cli/fs/command/CpCommandIntegrationTest.java
bf8086: Added a comment.

bf8086:
Thanks @aaudiber! I updated the PR to address your comment. PTAL.

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
```diff
@@ -98,10 +101,18 @@
...
+          .longOpt("preserve")
+          .required(false)
+          .desc("Preserve file permission attributes when copying files. "
+              + "All ownership, permissions and ACLs will be preserved")
+          .build();
 
   private int mCopyFromLocalBufferSize = COPY_FROM_LOCAL_BUFFER_SIZE_DEFAULT;
   private int mCopyToLocalBufferSize = COPY_TO_LOCAL_BUFFER_SIZE_DEFAULT;
   private int mThread = Runtime.getRuntime().availableProcessors() * 2;
+  private boolean mPreserveAttributes = false;
```
aaudiber: `mPreservePermissions`? Want to avoid using "attributes" to avoid confusing regarding ttl, etc

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
```diff
@@ -98,10 +101,18 @@
               + "default is 8MB when copying from local, "
               + "and 64MB when copying to local")
           .build();
+  private static final Option PRESERVE_OPTION =
```
aaudiber: what do you think of using the same option style as distcp? It extends well into preserving multiple types of attributes independently. https://hadoop.apache.org/docs/current/hadoop-distcp/DistCp.html#Command_Line_Options

aaudiber: talked, offline, the CLI parser we use doesn't handle that style of options well.

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
bf8086: renamed.

bf8086:
@aaudiber PTAL.

aaudiber:
LGTM, thanks for adding this

aaudiber:

aaudiber:

aaudiber:
@bf8086 can you manually cherry-pick to branch-1.8?

PR #8684, opened 2019-04-03
Add property to tolerate master state corruption

LuQQiu:


LuQQiu:

LuQQiu:
@aaudiber  I basically avoid `system.exit` and `throw t` when applying journal to metadata if tolerant corruption. PTAL, thanks

core/server/common/src/main/java/alluxio/ProcessUtils.java
```diff
@@ -67,6 +67,20 @@ public static void fatalError(Logger logger, String format, Object... args) {
...
+  /**
+   * Logs a fatal error and then exits the system if not tolerant master metadata corruption.
+   *
+   * @param tolerantCorruption whether or not to tolerant master metadata corruption
+   * @param logger the logger to log to
+   * @param t the throwable causing the fatal error
+   * @param format the error message format string
+   * @param args args for the format string
+   */
+  public static void fatalErrorWithCheck(boolean tolerantCorruption,
```
aaudiber: `ProcessUtils` contains util methods related to running processes, it shouldn't know anything about masters, metadata, or corruption

core/server/common/src/main/java/alluxio/ProcessUtils.java
```diff
@@ -75,7 +89,9 @@ public static void fatalError(Logger logger, Throwable t, String format, Object.
       throw new RuntimeException(message);
     }
     logger.error(message);
-    System.exit(-1);
+    if (!tolerantCorruption) {
```
aaudiber: It should be `tolerateCorruption`

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -284,8 +301,10 @@ public void applyAndJournal(Supplier<JournalContext> context, MutableInode<?> in
       applyCreateInode(inode);
       context.get().append(inode.toJournalEntry());
     } catch (Throwable t) {
-      ProcessUtils.fatalError(LOG, t, "Failed to apply %s", inode);
-      throw t; // fatalError will usually system.exit
+      ProcessUtils.fatalErrorWithCheck(mTolerantCorruption, LOG, t, "Failed to apply %s", inode);
```
aaudiber: can we move the error handling to a common method?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1376,6 +1376,16 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_JOURNAL_TOLERATE_CORRUPTION =
+      new Builder(Name.MASTER_JOURNAL_TOLERATE_CORRUPTION)
+          .setDefaultValue(false)
```
aaudiber: can we mark this as hidden so that it doesn't appear in the docs?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1376,6 +1376,16 @@ public String toString() {
...
+          .setDescription("Whether to tolerate master state corruption."
```
gpang: Does this control both for replaying journal entries, as well as applying entries during operation? Do we want a single flag to control both scenarios?

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Marked as hidden, thanks

core/server/common/src/main/java/alluxio/ProcessUtils.java
LuQQiu: I changed the parameter to `boolean shouldTolerate` means `whether we should tolerate this error and do not exit the system`. What's your opinion on that?

core/server/common/src/main/java/alluxio/ProcessUtils.java
LuQQiu: Changed, thanks!

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
LuQQiu: we cannot move the `throw t` to a method, thanks

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: I changed to only control replaying journal entries (basically `processJournalEntries`) since the reason of this propertykey is for advanced users to ignore some of the journal apply errors when starting master nodes. Thanks

LuQQiu:
@aaudiber @gpang Addressed the comments, PTAL thanks!

core/server/common/src/main/java/alluxio/ProcessUtils.java
aaudiber: I think adding this sort of boolean parameter makes the method API too complex/confusing. If we don't want to trigger a fatal error, we should not call this method in the first place

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
```diff
@@ -103,7 +104,16 @@ public static void restoreJournalEntryCheckpoint(CheckpointInputStream input, Jo
...
     while ((entry = reader.readEntry()) != null) {
-      journaled.processJournalEntry(entry);
+      try {
+        journaled.processJournalEntry(entry);
+      } catch (Throwable t) {
+        if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_TOLERATE_CORRUPTION)) {
+          ProcessUtils.fatalError(true,  LOG, t,
+              "Failed to process journal entry %s from a journal checkpoint", entry);
+        } else {
+          throw t;
```
aaudiber: I think we want to replay the rest of the journal still, not throw an exception

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
aaudiber: I think we can do something like
```java
handleJournalReplayFailure(masterName, journalEntry, t);
throw t;
```

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
LuQQiu: If not tolerate corruption, we directly throw any throwables.
If tolerate corruption, not exception will be thrown and we will still process the other journal entries.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -186,6 +186,13 @@ private void runInternal() {
         mJournalReader = new UfsJournalReader(mJournal, nextSequenceNumber, false);
         quietPeriodWaited = false;
         continue;
+      } catch (Throwable t) {
```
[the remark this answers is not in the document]
LuQQiu: Moved the catch to smallest scope, thanks

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
LuQQiu: Done!
change to something like
```
handleJournalReplayFailure(masterName, journalEntry, t);
if (!tolerrant) {
  throw t;
}
```

LuQQiu:
@aaudiber Addressed the comments, PTAL, thanks

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -91,7 +99,15 @@ public void initFromBackup(InputStream is) throws IOException {
       while ((entry = reader.readEntry()) != null) {
         String masterName = JournalEntryAssociation.getMasterForEntry(entry);
         Master master = mastersByName.get(masterName);
-        master.processJournalEntry(entry);
+        try {
+          master.processJournalEntry(entry);
+        } catch (Throwable t) {
+          JournalUtils.handleJournalReplayFailure(
+              LOG, t, "Failed to process journal entry %s when init from backup", entry);
+          if (!mTolerantCorruption) {
```
aaudiber: Can we move the throw into `handleJournalReplayFailure`? Wrapping the exception in a runtime exception is fine since we would only throw during tests.

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
```diff
@@ -157,5 +166,28 @@ public static void restoreFromCheckpoint(CheckpointInputStream input,
...
+   * Logs a fatal error and exits the system if required.
+   *
+   * @param logger the logger to log to
+   * @param t the throwable causing the fatal error
+   * @param format the error message format string
+   * @param args args for the format string
+   */
+  public static void handleJournalReplayFailure(
+      Logger logger, Throwable t, String format, Object... args) {
+    String message = String.format("Fatal error: " + format, args);
```
aaudiber: it's not a fatal error any more if we're tolerating corruption

LuQQiu:
@gpang PTAL, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -170,8 +172,8 @@ private synchronized void applyToMaster(JournalEntry entry) {
       LOG.trace("Applying entry to master {}: {} ", masterName, entry);
       master.processJournalEntry(entry);
     } catch (Throwable t) {
-      ProcessUtils.fatalError(LOG, t, "Failed to apply journal entry to master %s. Entry: %s",
-          masterName, entry);
+      JournalUtils.handleJournalReplayFailure(LOG, t,
+          "Failed to apply journal entry to master %s. Entry: %s", entry);
```
gpang: NIT: the string format has 2 placeholders, but only 1 argument.

LuQQiu: Added the masterName argument, Thanks for finding this error!

aaudiber:

PR #8691, opened 2019-04-04
Refactor JobMaster/WorkerContext into SelectExecutor/RunTaskContext

ZacBlanco:
This refactor introduces the SelectExecutor and
RunTaskContext. The fields of which will be the sole provider
of parameters for job definitions moving forward.

The reason for this changes is that is allows us to keep the
public API for job definitions consistent, but will allow us to
also add new parameters or values to the contexts as desired
in the future. It also allows us to push the creation of the
underlying JobServerContext closer to the beginning of the
callstack within each respective process, so that uneccesary
references aren't created and passed through multiple
methods. Overall it makes JobDefinitions more extensible
and the entire code path cleaner.

The current changes introduce an inheritance model from
JobServerContext which give both #selectExecutors and
#runTask access to a filesystem context, client, and a
UfsManager. By inheriting from JobServerContext the
SelectExecutorContext and RunTaskContext can have their
fields changed easily without worrying too much about the
other critical underlying fields that make up the
JobServerContext.


job/server/src/main/java/alluxio/job/load/LoadDefinition.java
```diff
@@ -58,14 +58,15 @@ public LoadDefinition() {
 
   @Override
   public Map<WorkerInfo, ArrayList<LoadTask>> selectExecutors(LoadConfig config,
-      List<WorkerInfo> jobWorkerInfoList, JobMasterContext jobMasterContext) throws Exception {
+      List<WorkerInfo> jobWorkerInfoList, SelectExecutorsContext selectExecutorsContext)
```
calvinjia: Consider naming the argument `context` since it is the only context used in this method.

ZacBlanco: done

job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
```diff
@@ -156,28 +156,30 @@ private void checkMigrateValid(MigrateConfig config, FileSystem fs) throws Excep
...
-      List<WorkerInfo> jobWorkerInfoList, JobMasterContext jobMasterContext) throws Exception {
+      List<WorkerInfo> jobWorkerInfoList, SelectExecutorsContext context)
+      throws Exception {
     AlluxioURI source = new AlluxioURI(config.getSource());
     AlluxioURI destination = new AlluxioURI(config.getDestination());
     if (source.equals(destination)) {
       return new HashMap<>();
     }
-    checkMigrateValid(config, jobMasterContext.getFileSystem());
+    checkMigrateValid(config, context.getFileSystem());
```
calvinjia: Can this be on one line now?

job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
```diff
@@ -156,28 +156,30 @@ private void checkMigrateValid(MigrateConfig config, FileSystem fs) throws Excep
    */
   @Override
   public Map<WorkerInfo, ArrayList<MigrateCommand>> selectExecutors(MigrateConfig config,
-      List<WorkerInfo> jobWorkerInfoList, JobMasterContext jobMasterContext) throws Exception {
+      List<WorkerInfo> jobWorkerInfoList, SelectExecutorsContext context)
+      throws Exception {
```
calvinjia: Could this be on the previous line now?

ZacBlanco:
I combed through the PR and fixed everything that could fit on a single line.

calvinjia:

Issue #8236, opened 2019-01-07
alluxio worker will throw "Failed to find any Kerberos tgt" with kerberos hadoop enviroment


apc999:
Note: this is an issue migrated from https://alluxio.atlassian.net/browse/ALLUXIO-3360 since we are deprecating JIRA by GIthub issues

**Alluxio Version:** <enter version here, e.g. 1.8.1>
1.8.1

**Describe the bug**

alluxio worker will throw "Failed to find any Kerberos tgt" with kerberos hadoop enviroment

when working with kerberos hdfs enviroment, the following exception will be thrown periodically
```
2018-11-13 11:01:47,993 WARN HdfsUnderFileSystem - 2 try to open hdfs://gdccluster/data/gamein/xyq_us/ods/ods_xyq_ach_show_day/dt=20171216/OdsXyqAchShow_20171216ACHSHOW-r-00000 : Failed on local exceptio
n: java.io.IOException: javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]; Host Details : lo
cal host is: "gdc-worker15-uspresto.i.nease.net/10.191.58.46"; destination host is: "gdc-nn02-formal.i.nease.net":9000;
2018-11-13 11:01:47,997 INFO RetryInvocationHandler - Exception while invoking getBlockLocations of class ClientNamenodeProtocolTranslatorPB over gdc-nn02-formal.i.nease.net/10.160.254.249:9000 after 8 fail over attempts. Trying to fail over immediately.
java.io.IOException: Failed on local exception: java.io.IOException: javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]; Host Details : local host is: "gdc-worker15-uspresto.i.nease.net/10.191.58.46"; destination host is: "gdc-nn02-formal.i.nease.net":9000;
at org.apache.hadoop.net.NetUtils.wrapException(NetUtils.java:776)
at org.apache.hadoop.ipc.Client.call(Client.java:1479)
at org.apache.hadoop.ipc.Client.call(Client.java:1412)
at org.apache.hadoop.ipc.ProtobufRpcEngine$Invoker.invoke(ProtobufRpcEngine.java:229)
at com.sun.proxy.$Proxy46.getBlockLocations(Unknown Source)
[...]

PR #8693, opened 2019-04-05
Fix kerberos login for root mount

bf8086:
#8236 

Alluxio master currently initiates login for UFS by calling `connectFromMaster` when a UFS is mounted. However, this is not called on the root mount, therefore user has to rely on `kinit` to enable Alluxio to talk to Kerberized HDFS.

This change ensures root mount also has a chance to log in the user with the credentials from Alluxio properties, which frees up user from having to use kinit and activates HDFS client relogin mechanism, so user does not have to periodically kinit to refresh the Kerberos ticket.

Note: this chance is targeted `branch-1.8`. Alluxio 2.0 has changed how root mount is added to the mount table thus might need a different fix.


bf8086:
@gpang PTAL.

gpang:
Jenkins, test this please

bf8086:
Closing this PR in favor of a more thorough fix https://github.com/Alluxio/alluxio/pull/8696

PR #8695, opened 2019-04-05
Use InodeView in DefaultPermissionChecker

madanadit:


core/server/master/src/main/java/alluxio/master/file/DefaultPermissionChecker.java
```diff
@@ -74,7 +75,7 @@ public void checkParentPermission(Mode.Bits bits, LockedInodePath inodePath)
 
     // collects existing inodes info on the path. Note that, not all the components of the path have
     // corresponding inodes.
-    List<Inode> inodeList = inodePath.getInodeList();
+    List<InodeView> inodeList = (List<InodeView>) (List<?>) inodePath.getInodeList();
```
aaudiber: can we avoid these casts?

madanadit: don't think so. can we?

aaudiber: what if we give `inodeList` the type `? extends InodeView`?

madanadit:

PR #8696, opened 2019-04-06
Fix authentication with kerberized HDFS

bf8086:
#8236

Alluxio master currently initiates login for UFS by calling connectFromMaster when a UFS is mounted through commandline. However, this is not called when root ufs is mounted or when nested ufs is remounted on restart, therefore user has to rely on kinit to enable Alluxio to talk to Kerberized HDFS.

This change ensures before connecting to a UFS, it always logs in the user with the credentials from Alluxio properties, which frees up user from having to use kinit and activates HDFS client relogin mechanism, so user does not have to periodically kinit to refresh the Kerberos ticket.

### DIFF
--- a/core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
+++ b/core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
@@ -127,9 +127,16 @@ public abstract class AbstractUfsManager implements UfsManager {
       UnderFileSystem fs = UnderFileSystem.Factory.create(ufsUri.toString(), ufsConf);
       mUnderFileSystemMap.putIfAbsent(key, fs);
       mCloser.register(fs);
+      try {
+        connectUfs(fs);
+      } catch (IOException e) {
+        LOG.warn("Failed to perform initial connect to UFS {}: {}", ufsUri, e.getMessage());
+      }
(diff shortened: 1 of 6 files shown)


bf8086:
@gpang This is an improved fix for the Kerberized HDFS login issue. PTAL.

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
```diff
@@ -127,10 +127,17 @@ private UnderFileSystem getOrAdd(AlluxioURI ufsUri, UnderFileSystemConfiguration
...
+      try {
+        connectUfs(fs);
+      } catch (IOException e) {
+        LOG.warn("Failed to perform initial connect to UFS {}: {}", ufsUri, e.getMessage());
+      }
       return fs;
     }
   }
 
+  protected abstract void connectUfs(UnderFileSystem fs) throws IOException;
```
gpang: can you add a javadoc on what this does?

bf8086:
@gpang The ticket cache will be ignored in favor of the keytab login. Once the keytab is given, the server can talk to Kerberized HDFS even if the TGT from ticket cache expires.

gpang:

PR #8702, opened 2019-04-07

GiantSeaweed:
Use the @Nullable annotation for all methods which may return null in /alluxio/core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java


madanadit:

PR #8722, opened 2019-04-08
Fix sync point status not found error

yuzhu:
name matching of enums is tricky and i think this way is safer.


yuzhu:
@ggezer PTAL, this was a problem because the capitalization of the status was different between grpc enum and our internal enum.

ggezer:
We are in track to get rid of all wire types that are simple wrappers. This one looks like one of such.
How about deleting the whole wire class?

calvinjia:

PR #8723, opened 2019-04-08
Remove waiting for the full sync to complete, so that meta master can…

yuzhu:
… start up

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
+++ b/core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
@@ -192,7 +192,7 @@ public class ActiveSyncManager implements Journaled {
                       LOG.warn("IOException encountered during active sync while starting {}", e);
                     }
                   }
-          )).get();
+          ));
         }
       } catch (Exception e) {
         LOG.warn("exception encountered during initial sync {}", e);


yuzhu:
@madanadit this should address the issue where master take forever to start up because it is blocked by a full sync.  I removed the get on the future, which allows it to execute asynchronously.

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
```diff
@@ -192,7 +192,7 @@ public void start() throws IOException {
                       LOG.warn("IOException encountered during active sync while starting {}", e);
                     }
                   }
-          )).get();
+          ));
```
madanadit: @yuzhu just to make sure. operations on the sub-tree being synced are still blocked because of locking correct? by not waiting we are allowing operations on other parts of the tree which are in sync to proceed?

madanadit:
Thanks @yuzhu. Can we add a test case?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
yuzhu: Yes, that is correct.

madanadit:

PR #8726, opened 2019-04-08
Add "pathConf show" command

cheng-chang:
Adds "pathConf show" to show path level configurations.

It has two modes:
1. only show configurations for the specific path;
2. show all configurations applicable to the path.

For example, suppose there are two path level configurations:
/a: property1=value1
/a/b: property2=value2
Then in mode 1, only property2=value2 will be shown for /a/b,
but in mode 2, since /a is the prefix of /a/b, both properties will be shown for /a/b.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/AlluxioProperties.java
+++ b/core/common/src/main/java/alluxio/conf/AlluxioProperties.java
@@ -18,6 +18,7 @@ import com.google.common.collect.Maps;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
@@ -199,14 +200,14 @@ public class AlluxioProperties {
   public Set<PropertyKey> keySet() {
     Set<PropertyKey> keySet = new HashSet<>(PropertyKey.defaultKeys());
     keySet.addAll(mUserProps.keySet());
-    return keySet;
+    return Collections.unmodifiableSet(keySet);
   }
 
   /**
    * @return the key set of user set properties
    */
   public Set<PropertyKey> userKeySet() {
-    return new HashSet<>(mUserProps.keySet());
+    return Collections.unmodifiableSet(mUserProps.keySet());
   }
 
   /**
--- a/core/common/src/main/java/alluxio/conf/AlluxioProperties.java
+++ b/core/common/src/main/java/alluxio/conf/AlluxioProperties.java
@@ -203,6 +203,13 @@ public class AlluxioProperties {
   }
 
   /**
+   * @return the key set of user set properties
+   */
+  public Set<PropertyKey> userKeySet() {
+    return new HashSet<>(mUserProps.keySet());
+  }
+
+  /**
    * Iterates over all the key value pairs and performs the given action.
    *
    * @param action the operation to perform on each key value pair
--- a/core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
@@ -40,7 +40,7 @@ public interface PathConfiguration {
    * @param path the Alluxio path
    * @return all property keys in the path level configuration that is applicable to path
    */
-  Set<PropertyKey> getAllPropertyKeys(AlluxioURI path);
+  Set<PropertyKey> getPropertyKeys(AlluxioURI path);
 
   /**
    * Factory method to create an implementation of {@link PathConfiguration}.
--- a/core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
@@ -15,6 +15,7 @@ import alluxio.AlluxioURI;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.PropertyKey;
 
+import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -33,4 +34,14 @@ public interface PathConfiguration {
    * @return the chosen configuration matching the path and containing the key
    */
   Optional<AlluxioConfiguration> getConfiguration(AlluxioURI path, PropertyKey key);
+
+  /**
+   * Factory method to create an implementation of {@link PathConfiguration}.
+   *
+   * @param pathConf the map from paths to path level configurations
+   * @return the implementation of {@link PathConfiguration}
+   */
+  static PathConfiguration create(Map<String, AlluxioConfiguration> pathConf) {
+    return new PrefixPathConfiguration(pathConf);
+  }
 }
--- a/core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
@@ -17,6 +17,7 @@ import alluxio.conf.PropertyKey;
 
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 /**
  * Path level configuration.
@@ -34,6 +35,12 @@ public interface PathConfiguration {
    * @return the chosen configuration matching the path and containing the key
    */
   Optional<AlluxioConfiguration> getConfiguration(AlluxioURI path, PropertyKey key);
+
+  /**
+   * @param path the Alluxio path
+   * @return all property keys in the path level configuration that is applicable to path
+   */
+  Set<PropertyKey> getAllPropertyKeys(AlluxioURI path);
 
   /**
    * Factory method to create an implementation of {@link PathConfiguration}.
--- a/shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
+++ b/shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
@@ -52,14 +52,15 @@ import java.util.Set;
  * but in mode 2, since /a is the prefix of /a/b, both properties will be shown for /a/b.
  */
 public final class ShowCommand extends AbstractFsAdminCommand {
-  public static final String RESOLVE_OPTION_NAME = "resolve";
+  public static final String ALL_OPTION_NAME = "all";
 
-  private static final Option RESOLVE_OPTION =
+  private static final Option ALL_OPTION =
       Option.builder()
-          .longOpt(RESOLVE_OPTION_NAME)
+          .longOpt(ALL_OPTION_NAME)
           .required(false)
           .hasArg(false)
-          .desc("Whether to show all path level configurations applicable to the path")
+          .desc("Whether to show all path level configurations applicable to the path, including"
+              + " those set on ancestor paths")
           .build();
 
   /**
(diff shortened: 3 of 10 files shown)


cheng-chang:
@calvinjia @aaudiber PTAL

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+ * 2. show all configurations applicable to the path.
+ * <p>
+ * For example, suppose there are two path level configurations:
+ * /a: property1=value1
+ * /a/b: property2=value2
+ * Then in mode 1, only property2=value2 will be shown for /a/b,
+ * but in mode 2, since /a is the prefix of /a/b, both properties will be shown for /a/b.
+ */
+public final class ShowCommand extends AbstractFsAdminCommand {
+  public static final String RESOLVE_OPTION_NAME = "resolve";
```
aaudiber: what do you think of naming this "--all" instead? It's shorter and might be more intuitive. If I was a user, I'd need to read the documentation to understand what either "--all" or "--resolve" are supposed to mean

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+    Configuration configuration = mMetaConfigClient.getConfiguration();
+
+    if (cl.hasOption(RESOLVE_OPTION_NAME)) {
+      Set<PropertyKey> pathPropertyKeys = new HashSet<>();
+      Map<String, AlluxioConfiguration> pathConfMap = new HashMap<>();
+
+      configuration.getPathConf().forEach((path, conf) -> {
+        AlluxioProperties properties = new AlluxioProperties();
+        conf.forEach(property -> {
+          pathPropertyKeys.add(PropertyKey.fromString(property.getName()));
```
aaudiber: consider saving `PropertyKey.fromString(property.getName())` to a local variable since we use it twice here

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+  public int run(CommandLine cl) throws IOException {
+    String targetPath = cl.getArgs()[1];
+    Configuration configuration = mMetaConfigClient.getConfiguration();
+
+    if (cl.hasOption(RESOLVE_OPTION_NAME)) {
+      Set<PropertyKey> pathPropertyKeys = new HashSet<>();
+      Map<String, AlluxioConfiguration> pathConfMap = new HashMap<>();
+
+      configuration.getPathConf().forEach((path, conf) -> {
+        AlluxioProperties properties = new AlluxioProperties();
```
aaudiber: Consider adding a static builder in `AlluxioProperties` which turns a `Map<String, String>` into `AlluxioProperties`

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+          properties.set(PropertyKey.fromString(property.getName()),
+              property.getValue());
+        });
+        pathConfMap.put(path, new InstancedConfiguration(properties));
+      });
+      PathConfiguration pathConf = PathConfiguration.create(pathConfMap);
+
+      AlluxioURI targetUri = new AlluxioURI(targetPath);
+      List<PropertyKey> propertyKeys = new ArrayList<>(pathPropertyKeys);
+      propertyKeys.sort(Comparator.comparing(PropertyKey::getName));
```
aaudiber: could we build a list of just the relevant properties, then sort those by name? Then we don't need `propertyKeys` or `pathPropertyKeys` at all

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+      propertyKeys.forEach(key -> {
+        pathConf.getConfiguration(targetUri, key).ifPresent(conf -> {
+          mPrintStream.println(format(key.getName(), conf.get(key)));
+        });
+      });
+    } else {
+      if (configuration.getPathConf().containsKey(targetPath)) {
+        List<Property> properties = configuration.getPathConf().get(targetPath);
+        properties.sort(Comparator.comparing(Property::getName));
+        properties.forEach(property -> {
```
aaudiber: could move the sorting and printing outside the if/else

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+        });
+      }
+    }
+    mPrintStream.close();
+    return 0;
+  }
+
+  @Override
+  public String getUsage() {
+    return "show [--" + RESOLVE_OPTION_NAME + "] <path>";
```
aaudiber: consider using `String.format` to make this a little more readable

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,143 @@
...
+ */
+public final class ShowCommand extends AbstractFsAdminCommand {
+  public static final String RESOLVE_OPTION_NAME = "resolve";
+
+  private static final Option RESOLVE_OPTION =
+      Option.builder()
+          .longOpt(RESOLVE_OPTION_NAME)
+          .required(false)
+          .hasArg(false)
+          .desc("Whether to show all path level configurations applicable to the path")
```
aaudiber: can we mention that this includes path level configuration set on ancestors?

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
cheng-chang: it's constructed from a `List<Property>`, it's only used here for now, if the pattern reoccurs again in the future, let's add a new constructor in `AlluxioProperties` then.

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
cheng-chang: The problem is that all properties might be "relevant", for example, if there is no path level configuration set for path `/a/b/c/d`, but there are configurations for path `/a`, `/a/b`, `/a/b/c`, then all properties in these configurations are "relevant".

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
cheng-chang: consolidated into one `else if` block, I think it's more intuitive to put them together in one block

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
aaudiber: What do you think of adding a `List<PropertyKey> getAllProperties(uri)` method in `PathConfiguration`? I could see this being useful for debugging or other code in the future.

cheng-chang:
@aaudiber PTAL

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:
@calvinjia PTAL

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Show path level configurations.
+ *
+ * <p>
+ * It has two modes:
```
calvinjia: Specify `--all`?

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,141 @@
...
+      Option.builder()
+          .longOpt(ALL_OPTION_NAME)
+          .required(false)
+          .hasArg(false)
+          .desc("Whether to show all path level configurations applicable to the path, including"
+              + " those set on ancestor paths")
+          .build();
+
+  /**
+   * @param context     fsadmin command context
```
calvinjia: We don't align the Javadoc

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,141 @@
...
+   * @param alluxioConf Alluxio configuration
+   */
+  public ShowCommand(Context context, AlluxioConfiguration alluxioConf) {
```
calvinjia: We ignore the configuration?

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
```diff
@@ -0,0 +1,141 @@
...
+
+  @Override
+  public String getCommandName() {
+    return "show";
+  }
+
+  @Override
+  public Options getOptions() {
+    return new Options()
+        .addOption(ALL_OPTION);
```
calvinjia: One line?

core/common/src/main/java/alluxio/conf/path/PathConfiguration.java
```diff
@@ -33,4 +35,20 @@
    * @return the chosen configuration matching the path and containing the key
    */
   Optional<AlluxioConfiguration> getConfiguration(AlluxioURI path, PropertyKey key);
+
+  /**
+   * @param path the Alluxio path
+   * @return all property keys in the path level configuration that is applicable to path
+   */
+  Set<PropertyKey> getAllPropertyKeys(AlluxioURI path);
```
calvinjia: Why do we call this `getAllPropertyKeys` instead of `getPropertyKeys`?

core/common/src/main/java/alluxio/conf/AlluxioProperties.java
```diff
@@ -202,6 +202,13 @@ public boolean isSetByUser(PropertyKey key) {
     return keySet;
   }
 
+  /**
+   * @return the key set of user set properties
+   */
+  public Set<PropertyKey> userKeySet() {
+    return new HashSet<>(mUserProps.keySet());
```
calvinjia: Should this use ImmutableSet?

shell/src/main/java/alluxio/cli/fsadmin/pathconf/ShowCommand.java
cheng-chang: not used for now

cheng-chang:
@calvinjia PTAL

calvinjia:

PR #8729, opened 2019-04-09
[SMALLFIX] Replace anonymous type with lambda

cockroach20168:
replace anonymous type with lambda in alluxio/job/client/src/main/java/alluxio/client/job/RetryHandlingJobMasterClient.java
Replace

    retryRPC(new RpcCallable<Void>() {
      public Void call() {
        mClient.cancel(CancelPRequest.newBuilder().setJobId(jobId).build());
        return null;
      }
    });
with

    retryRPC((RpcCallable<Void>) () -> {
      mClient.cancel(CancelPRequest.newBuilder().setJobId(jobId).build());
      return null;
    });


aaudiber:

PR #8730, opened 2019-04-09

Acc-plus:
Change

        import org.junit.Assert;
to

        import static org.junit.Assert.fail;
AND

Update all

        Assert.{{ method }}*
to

        {{ method }}*


aaudiber:
@Acc-plus The PR shows no changes, please open a new pull request with the changes you want to make

PR #8731, opened 2019-04-09

ZZH98:
Use static imports for standard test utilities in alluxio/core/server/worker/src/test/java/alluxio/worker/block/evictor/EvictorContractTest.java


aaudiber:

PR #8755, opened 2019-04-11
Support changing journal type in multi process cluster

LuQQiu:
Change journal type to embedded journal by default in multi process cluster.
By passing in `-Dtest.journal.type=UFS`, those tests support changing journal can change to UFS journal.


minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -162,15 +162,23 @@ public synchronized void start() throws Exception {
...
     switch (mDeployMode) {
-      case NON_HA:
-        MasterNetAddress masterAddress = mMasterAddresses.get(0);
+      case UFS_NON_HA:
         mProperties.put(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS.toString());
         mProperties.put(PropertyKey.MASTER_HOSTNAME, masterAddress.getHostname());
         mProperties.put(PropertyKey.MASTER_RPC_PORT, Integer.toString(masterAddress.getRpcPort()));
         mProperties.put(PropertyKey.MASTER_WEB_PORT, Integer.toString(masterAddress.getWebPort()));
         break;
+      case EMBEDDED_NON_HA:
```
aaudiber: Why do we need a separate `EMBEDDED_NON_HA` from `EMBEDDED_HA`? It looks like they result in the same configuration.

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
```diff
@@ -125,7 +126,10 @@ public static InstancedConfiguration defaults() {
     conf.put(PropertyKey.WORKER_NETWORK_SHUTDOWN_TIMEOUT, "0ms");
 
     conf.put(PropertyKey.Template.WORKER_TIERED_STORE_LEVEL_ALIAS.format(0), "MEM");
-    conf.put(PropertyKey.USER_RPC_RETRY_MAX_DURATION, "1s");
+    if (journalType.equals("UFS")) {
+      // Raft journal system need longer to start, job worker connect should wait for longer time
+      conf.put(PropertyKey.USER_RPC_RETRY_MAX_DURATION, "1s");
```
aaudiber: can we make the raft journal start faster by tuning `MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT` and `MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL`? We want the test to run as fast as possible without introducing flakiness.

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
LuQQiu: The reason why I use `EMBEDDED_NON_HA` and `EMBEDDED_HA` before is because I found out the `EMBEDDED_HA` configuration with embedded journal addresses and master rpc addresses do not work. This is because our workers still use `master.rpc.port` when building the master inquire client and cannot connect to the leading master. This bug has been fixed in the PR.

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
LuQQiu: Added these properties to make the raft journal start faster. 

```
      conf.put(PropertyKey.USER_RPC_RETRY_MAX_DURATION, "30s");
      conf.put(PropertyKey.MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT, "1s");
      conf.put(PropertyKey.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL, "100ms");
```

LuQQiu:
@aaudiber Thanks for the review. I fixed the master inquire client bug which makes workers unable to connect with the leading master when using embedded journal in single master situation. PTAL, thanks

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
aaudiber: it still takes 30 seconds for the raft journal to start?

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -162,16 +162,16 @@ public synchronized void start() throws Exception {
     mState = State.STARTED;
 
     mMasterAddresses = generateMasterAddresses(mNumMasters);
+    MasterNetAddress masterAddress = mMasterAddresses.get(0);
     LOG.info("Master addresses: {}", mMasterAddresses);
     switch (mDeployMode) {
-      case NON_HA:
-        MasterNetAddress masterAddress = mMasterAddresses.get(0);
+      case UFS_NON_HA:
```
aaudiber: why do we need to change the deploy types?

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -162,16 +162,16 @@ public synchronized void start() throws Exception {
     mState = State.STARTED;
 
     mMasterAddresses = generateMasterAddresses(mNumMasters);
+    MasterNetAddress masterAddress = mMasterAddresses.get(0);
```
aaudiber: why do we move this outside? I think we only need it for `UFS_NON_HA`

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
```diff
@@ -75,7 +74,7 @@
       new ConfigurationRule(new ImmutableMap.Builder<PropertyKey, String>()
           .put(PropertyKey.MASTER_JOURNAL_TAILER_SHUTDOWN_QUIET_WAIT_TIME_MS, "100")
           .put(PropertyKey.MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES, "2")
-          .put(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX, "32")
+          .put(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX, "65")
```
aaudiber: why is this change needed?

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
```diff
@@ -137,9 +137,9 @@ public void multiMasterJournalStopIntegration() throws Exception {
     MultiProcessCluster cluster =
         MultiProcessCluster.newBuilder(PortCoordination.JOURNAL_STOP_MULTI_MASTER)
             .setClusterName("multiMasterJournalStopIntegration")
+            .setDeployMode(MultiProcessCluster.getDeployMode(TEST_NUM_MASTERS))
```
aaudiber: could the cluster figure out the deploy mode on its own based on how many masters are configured and whether embedded journal is configured?

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
LuQQiu: because `EMBEDDED` include the `EMBEDDED_NON_HA` and `EMBEDDED_HA` situation. `NON_HA` deploy mode should include the `EMBEDDED_NON_HA` and `UFS_NON_HA` cases

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
LuQQiu: Raft journal system requires that the journal loge size should be bigger than 64KB.

aaudiber: can we make this 128 then? also, do you mean 64 bytes?

LuQQiu: Sorry, I mean 64 bytes, will change it to 128 bytes

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
LuQQiu: Changed to `3s` which is sufficient in my local test. Will change it to larger values if jenkins tests says to.

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
```diff
@@ -45,10 +45,11 @@ public static InstancedConfiguration defaults() {
    *
    * @param hostname the master hostname
    * @param workDirectory the work directory in which to configure the journal and tiered storage
+   * @param journalType the journal type to configure the related configuration
    * @return the configuration
    */
   public static Map<PropertyKey, String> testConfigurationDefaults(AlluxioConfiguration alluxioConf,
-      String hostname, String workDirectory) {
+      String hostname, String workDirectory, String journalType) {
```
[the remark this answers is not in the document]
LuQQiu: Changed to get journal type from alluxioConf

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
LuQQiu: Moved back to `UFS_NON_HA`

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
LuQQiu: Now if users do not give their desired deploy mode, we get deploy mode according to the num of masters and master journal type by default.

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
aaudiber: Can you create an issue to track this bug?

LuQQiu: This bug is already fixed in this PR, but I will open an issue to describe this bug and point the this PR.

LuQQiu:
@aaudiber Created the issue and added the fix for job master inquire client, PTAL, thanks

LuQQiu:
Thanks for the review @aaudiber 

@calvinjia This PR support changing journal type in multi process cluster and fix the issue describe in #8815. PTAL, thanks

calvinjia:

PR #8757, opened 2019-04-11
Add versioning target for GrpcDataServer worker service

ggezer:


ggezer:
@bf8086 PTAL.

ggezer:
@calvinjia PTAL

core/transport/src/grpc/version.proto
```diff
@@ -8,20 +8,21 @@ package alluxio.grpc.version;
...
+  FILE_SYSTEM_MASTER_JOB_SERVICE = 3;
+  BLOCK_MASTER_CLIENT_SERVICE = 4;
+  BLOCK_MASTER_WORKER_SERVICE = 5;
+  META_MASTER_CONFIG_SERVICE = 6;
+  META_MASTER_CLIENT_SERVICE = 7;
+  META_MASTER_MASTER_SERVICE = 8;
+  METRICS_MASTER_CLIENT_SERVICE = 9;
+  JOB_MASTER_CLIENT_SERVICE = 10;
+  JOB_MASTER_WORKER_SERVICE = 11;
+  KEY_VALUE_MASTER_CLIENT_SERVICE = 12;
```
ZacBlanco: Since we've removed KV from alluxio, can the `KEY_VALUE_*` services also be removed?

ggezer: Nice catch.

core/transport/src/grpc/version.proto
```diff
@@ -8,20 +8,19 @@ package alluxio.grpc.version;
...
+  FILE_SYSTEM_MASTER_JOB_SERVICE = 3;
+  BLOCK_MASTER_CLIENT_SERVICE = 4;
+  BLOCK_MASTER_WORKER_SERVICE = 5;
+  META_MASTER_CONFIG_SERVICE = 6;
+  META_MASTER_CLIENT_SERVICE = 7;
+  META_MASTER_MASTER_SERVICE = 8;
+  METRICS_MASTER_CLIENT_SERVICE = 9;
+  JOB_MASTER_CLIENT_SERVICE = 10;
+  JOB_MASTER_WORKER_SERVICE = 11;
+  FILE_SYSTEM_WORKER_WORKER_SERVICE = 12;
```
calvinjia: How does this work with versioning? Should we instead keep the old enum values but change them to be called `*_UNUSED`?

ggezer: There is no version constants that are used to match them at the service implementation.
I think deleting them is safe as versioning in 2.0 is completely changed and not compatible with the old clients.

calvinjia:

PR #8762, opened 2019-04-12
Add pathConf add and remove services and commands

cheng-chang:
This PR adds `alluxio fsadmin pathConf add` and `alluxio fsadmin pathConf remove` commands.


core/transport/src/grpc/meta_master.proto
```diff
@@ -132,9 +139,14 @@ service MetaMasterConfigurationService {
   rpc GetConfiguration(GetConfigurationPOptions) returns (GetConfigurationPResponse);
 
   /**
-   * Sets a property for a path pattern.
+   * Sets a property for a path.
    */
   rpc SetPathConfiguration(SetPathConfigurationPRequest) returns (SetPathConfigurationPResponse);
+
+  /**
+   * Removes properties for a path, if the keys are empty, it means removing all properties.
```
aaudiber: nit: removing -> remove

core/transport/src/grpc/meta_master.proto
```diff
@@ -115,12 +115,19 @@ message GetMetricsPResponse {
...
-  required string key = 2;
-  required string value = 3;
-  optional SetPathConfigurationPOptions options = 4;
+  map<string, string> properties = 2;
+  optional SetPathConfigurationPOptions options = 3;
 }
 message SetPathConfigurationPResponse {}
 
+message RemovePathConfigurationPOptions {}
+message RemovePathConfigurationPRequest {
```
aaudiber: the meaning of this type isn't very clear - does the nth path correspond to the nth key? Or does each key get applied to each path? I'd like to make it as clear as possible from this type signature what the request means

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -30,8 +33,35 @@
    * Sets a property for a path.
    *
    * @param path the path
-   * @param key the property key
+   * @param key the proeprty key
```
calvinjia: `property`

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -30,8 +33,35 @@
...
    * @param value the property value
    */
-  void setPathConfiguration(String path, PropertyKey key, String value) throws IOException;
+  default void setPathConfiguration(String path, PropertyKey key, String value) throws IOException {
```
calvinjia: `AlluxioURI` instead of `String`?

core/server/master/src/main/java/alluxio/master/meta/MetaMasterConfigurationServiceHandler.java
```diff
@@ -54,13 +62,35 @@ public void getConfiguration(GetConfigurationPOptions options,
...
+      mMetaMaster.setPathConfiguration(path, props);
       return SetPathConfigurationPResponse.getDefaultInstance();
     }, "setPathConfiguration", "request=%s", responseObserver, request);
   }
+
+  @Override
+  public void removePathConfiguration(RemovePathConfigurationPRequest request,
+      StreamObserver<RemovePathConfigurationPResponse> responseObserver) {
+    String path = request.getPath();
+    List<String> keys = request.getKeysList();
```
calvinjia: Should we check `hasKeysList`? Default is empty but I think this is more explicit

core/server/master/src/main/java/alluxio/master/meta/PathProperties.java
```diff
@@ -146,10 +146,12 @@ public Journaled getDelegate() {
      */
     public void addProperties(Supplier<JournalContext> ctx, String path,
         Map<PropertyKey, String> properties) {
-      Map<String, String> newProperties = mProperties.getOrDefault(path, new HashMap<>());
-      properties.forEach((key, value) -> newProperties.put(key.getName(), value));
-      applyAndJournal(ctx, PathPropertiesEntry.newBuilder().setPath(path)
-          .putAllProperties(newProperties).build());
+      if (!properties.isEmpty()) {
```
aaudiber: can we move `addProperties` and `removeProperties` outside of `State`? The correctness of `State` is critical for the journal to be able to replay correctly, so I'd like to keep it as simple as possible. For example, we create journal entries in `InodeTree` and then apply them to `InodeTreePersistentState` instead of creating the entries in `InodeTreePersistentState`.

In general, all methods in `State` should take a journal entries as arguments

core/server/master/src/main/java/alluxio/master/meta/MetaMasterConfigurationServiceHandler.java
cheng-chang: there is no such method

core/transport/src/grpc/meta_master.proto
cheng-chang: there is just one path and a list of keys for that path, what's the confusing part?

aaudiber: oh sorry my eyes are bad, thought `path` was `repeated` not `required` >.<

on that note, we have a best practice of never using `required`. There is little benefit, and it can make it harder to evolve the API in the future. Better to use `optional` instead

cheng-chang:
jenkins, test this please

cheng-chang:
@aaudiber @calvinjia PTAL

core/server/master/src/main/java/alluxio/master/meta/MetaMasterConfigurationServiceHandler.java
calvinjia: Oh that's because the field is not optional, please use optional and not any required fields.

cheng-chang: it's a `repeated` field, cannot add `optional`

cheng-chang:

cheng-chang:

PR #8763, opened 2019-04-12
[SMALLFIX] Add @Nullable annotation

KSDeng:
Add @nullable annotation for methods which may return null in /alluxio/core/server/common/src/main/java/alluxio/RpcUtils.java


aaudiber:
@KSDeng Thanks for the contribution! It looks like there's an additional npm shrinkwrap file added by accident. Can you remove it? The `@Nullable` change looks good to me

PR #8767, opened 2019-04-12
Refactor getConf in fs client to use path level config

cheng-chang:
This PR marks phase 1 of path level configuration as done.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
@@ -155,7 +155,7 @@ public class BaseFileSystem implements FileSystem {
   public void createDirectory(AlluxioURI path, CreateDirectoryPOptions options)
       throws FileAlreadyExistsException, InvalidPathException, IOException, AlluxioException {
     checkUri(path);
-    options = FileSystemOptions.createDirectoryDefaults(mFsContext.getConf(path))
+    options = FileSystemOptions.createDirectoryDefaults(mFsContext.getPathConf(path))
         .toBuilder().mergeFrom(options).build();
     FileSystemMasterClient masterClient = mFsContext.acquireMasterClient();
     try {
@@ -184,7 +184,7 @@ public class BaseFileSystem implements FileSystem {
   public FileOutStream createFile(AlluxioURI path, CreateFilePOptions options)
       throws FileAlreadyExistsException, InvalidPathException, IOException, AlluxioException {
     checkUri(path);
-    options = FileSystemOptions.createFileDefaults(mFsContext.getConf(path))
+    options = FileSystemOptions.createFileDefaults(mFsContext.getPathConf(path))
         .toBuilder().mergeFrom(options).build();
     FileSystemMasterClient masterClient = mFsContext.acquireMasterClient();
     URIStatus status;
@@ -204,7 +204,7 @@ public class BaseFileSystem implements FileSystem {
     }
 
     OutStreamOptions outStreamOptions =
-        new OutStreamOptions(options, mFsContext.getConf(path));
+        new OutStreamOptions(options, mFsContext.getPathConf(path));
     outStreamOptions.setUfsPath(status.getUfsPath());
     outStreamOptions.setMountId(status.getMountId());
     outStreamOptions.setAcl(status.getAcl());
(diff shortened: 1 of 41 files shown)


core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -683,7 +683,7 @@ private void checkUri(AlluxioURI uri) {
        * user passes. If not, throw an exception letting the user know they don't match.
        */
       Authority configured =
-          MasterInquireClient.Factory.create(mFsContext.getConf())
+          MasterInquireClient.Factory.create(mFsContext.getConf(uri))
```
aaudiber: I don't think we need `uri` here since it's looking up connection details, and it doesn't make sense for those to be path-based

core/client/hdfs/src/test/java/alluxio/hadoop/AbstractFileSystemTest.java
```diff
@@ -645,17 +647,20 @@ void verifyBlockLocations(List<WorkerNetAddress> blockWorkers, List<String> ufsL
...
     FileSystemContext fsContext = mock(FileSystemContext.class);
     when(fsContext.getClientContext()).thenReturn(ClientContext.create(mConfiguration));
     when(fsContext.getConf()).thenReturn(mConfiguration);
+    when(fsContext.getConf(uri)).thenReturn(new SpecificPathConfiguration(mConfiguration,
+        new PrefixPathConfiguration(), uri));
     alluxio.client.file.FileSystem fs = alluxio.client.file.FileSystem.Factory.create(fsContext);
     alluxio.client.file.FileSystem spyFs = spy(fs);
     doReturn(new URIStatus(fileInfo))
-        .when(spyFs).getStatus(new AlluxioURI(HadoopUtils.getPathWithoutScheme(path)));
+        .when(spyFs).getStatus(uri);
```
calvinjia: This can be on one line

cheng-chang:
@calvinjia sometimes, `getConf()` is still needed

calvinjia:
@cheng-chang how about allowing `getConf(null)`? Do you share my concern about the API leading to mistakes?

aaudiber:
I share @calvinjia's concern that it would be easy to miss a path default in the RPC layer. We could use the type system to prevent the issue by having our default options creators take `SpecificPathConfiguration` instead of `AlluxioConfiguration`.

cheng-chang:
@aaudiber I agree it's a good idea to use type system to prevent missing path defaults, but the path configs are not only used in default options, for example https://github.com/Alluxio/alluxio/pull/8767/files#diff-bd724ad92fc2948fda03d6b1d869daf6L327. I think it's the programmer's responsibility to decide where path configs are needed instead of cluster configs.

aaudiber:
@cheng-chang hm good point, another idea is to split `BaseFileSystem` into two layers, one which resolves path configuration, and another which implements the operations. e.g.

**BaseFileSystem.java**
```java
public List<BlockLocationInfo> getBlockLocations(AlluxioURI path) {
  SpecificPathConfiguration conf = resolveConf(path);
  return mFs.getBlockLocations(path, conf);
}
```

**FileSystemImpl.java**
```java
public List<BlockLocationInfo> getBlockLocations(AlluxioURI path, SpecificPathConfiguration conf) {
  // getBlockLocations implementation, using conf
}
```

what do you think?

cheng-chang:
@aaudiber I think it's unnecessary, the whole point of the design behind `SpecificPathConfiguration implements AlluxioConfiguration` and `getConf(path)` is to keep the usage of path configs and cluster configs consistent. Whatever abstractions or refactoring we do, ultimately, there must be some cases where the programmer has to determine whether to use cluster or path configs.

cheng-chang:
@calvinjia @aaudiber I'm fine with calvin's idea of `getConf(null)` to force API users to think about whether path or cluster configs should be applied. WDYT?

aaudiber:
I also prefer `getClusterConf()/getPathConf()` over `getConf(null)`. I don't see the interchangeability of path-resolved and path-unresolved configuration as a good thing, since it opens us up to bugs. But if you think it will create too much overhead to enforce the usage of path-resolved config, I'm ok with it.

cheng-chang:
@aaudiber @calvinjia I refactored `getConf` according to my proposal above. PTAL.

cheng-chang:
After this PR is merged, I'm going to open a PR to merge the feature branch `pathconf` into `master`.

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:

PR #8812, opened 2019-04-15
Deflake JournalToolTest

aaudiber:
Saw this test fail a new contributor's build.

Previously, the tests could fail if a checkpoint is written
too early, and the journalTool reads that checkpoint. This
PR increases the number of entries required to trigger a
checkpoint, so that we don't take a checkpoint until we
want to.


calvinjia:

PR #8816, opened 2019-04-16
[SMALLFIX]Modify due to #478

161220012:
[SMALLFIX] Java 8 improvement: replace collections.sort() with list.sort in /alluxio/core/server/worker/src/main/java/alluxio/worker/block/evictor/LRFUEvictor.java#getSortedCRF #478
Change
    Collections.sort(sortedCRF, Comparator.comparingDouble(Entry::getValue));
to
    sortedCRF.sort(Comparator.comparingDouble(Entry::getValue));


161220012:
I have already modified the file asked. Why did I still fail？

PR #8818, opened 2019-04-16
[SMALLFIX]Modify due to #478

161220012:


161220012:
I have already modified the file asked. Why did I still fail?

PR #8829, opened 2019-04-17
[SMALLFIX] change import org.junit.Assert; import static org.junit.Assert.assert…

DaleTian8820:
…Equals; and Update all

Assert.{{ method }}*
to

{{method}}*

### DIFF
--- a/core/common/src/test/java/alluxio/underfs/UfsDirectoryStatusTest.java
+++ b/core/common/src/test/java/alluxio/underfs/UfsDirectoryStatusTest.java
@@ -11,7 +11,9 @@
 
 package alluxio.underfs;
 
-import org.junit.Assert;
+//import org.junit.Assert;
+import static org.junit.Assert.assertEquals;
+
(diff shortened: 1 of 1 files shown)


core/common/src/test/java/alluxio/underfs/UfsDirectoryStatusTest.java
```diff
@@ -11,7 +11,9 @@
 
 package alluxio.underfs;
 
-import org.junit.Assert;
+//import org.junit.Assert;
```
aaudiber: please remove this entirely

aaudiber:
@DaleTian8820 It looks like a lot of new changes are getting added now, it might be easier to create a new commit/PR for your change

apc999:
@DaleTian8820 It looks like a lot of new changes are getting added now, it might be easier to create a new commit/PR for your change

DaleTian8820:
I changed

PR #8843, opened 2019-04-17
Implement RocksStore checkpoints using Rocks Checkpoints

aaudiber:
Previously we used Rocks BackupEngine to create backups,
and build our checkpoints from those. That was slow because
backups require copying the full metastore. Checkpoints are
faster because they use hard links instead of copying. With
this change, the time needed to take checkpoints and restore
checkpoints is reduced by 40%.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/checkpoint/CheckpointType.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/checkpoint/CheckpointType.java
@@ -36,10 +36,6 @@ public enum CheckpointType {
    */
   ROCKS(3, new TarballCheckpointFormat()),
   /**
-   * A RocksDB checkpoint in .tar.gz format.
-   */
-  ROCKS_CHECKPOINT(6, new TarballCheckpointFormat()),
-  /**
    * This format sequentially writes delimited InodeMeta.Inode protocol buffers one after another
    * using the protocol buffer writeDelimitedTo method.
    */
--- a/core/server/common/src/main/java/alluxio/master/journal/checkpoint/CheckpointType.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/checkpoint/CheckpointType.java
@@ -36,6 +36,10 @@ public enum CheckpointType {
    */
   ROCKS(3, new TarballCheckpointFormat()),
   /**
+   * A RocksDB checkpoint in .tar.gz format.
+   */
+  ROCKS_CHECKPOINT(6, new TarballCheckpointFormat()),
+  /**
    * This format sequentially writes delimited InodeMeta.Inode protocol buffers one after another
    * using the protocol buffer writeDelimitedTo method.
    */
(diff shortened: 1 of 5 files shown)


core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
```diff
@@ -159,19 +171,22 @@ private void createDb() throws RocksDBException {
...
+    LOG.info("Creating rocksdb checkpoint at {}", mDbCheckpointPath);
     long startNano = System.nanoTime();
+
     CheckpointOutputStream out = new CheckpointOutputStream(output, CheckpointType.ROCKS);
-    try (BackupEngine backupEngine = BackupEngine.open(Env.getDefault(),
-        new BackupableDBOptions(mDbBackupPath).setMaxBackgroundOperations(4))) {
-      backupEngine.createNewBackup(mDb, true);
-      backupEngine.purgeOldBackups(1);
+    try {
+      // createCheckpoint requires that the directory not already exist.
```
gpang: what sets it to `ROCKS_CHECKPOINT`? What happens to backup engine?

core/server/common/src/main/java/alluxio/master/journal/checkpoint/CheckpointType.java
```diff
@@ -35,6 +35,10 @@
    * A RocksDB backup in .tar.gz format.
    */
   ROCKS(3, new TarballCheckpointFormat()),
+  /**
+   * A RocksDB checkpoint in .tar.gz format.
```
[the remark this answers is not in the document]
aaudiber: oops, this wasn't supposed to slip into this PR. We never released back-based checkpoints, so we don't need to differentiate between them and checkpoint-based backups

backups and checkpoints are RocksDB concepts. Basically a backup is information set aside by the backup engine so that it can restore a database from the backup. Backups are full copies of all state. Checkpoints use hard links instead of copying to essentially fork a database. Unlike a backup, a checkpoint can be opened and used like a regular database

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
aaudiber: backup engine is no longer being used in any way

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
```diff
@@ -114,24 +112,37 @@ private void resetDb() throws RocksDBException {
...
+          synchronized (this) {
+            mDb.close();
+          }
+        });
+        closer.register(() -> {
+          synchronized (this) {
+            mCheckpoint.close();
+          }
+        });
+        closer.close();
```
calvinjia: Does using a closer for all of them make sense, given that `mDb.close` will fail if any of the ColumnFamilyHandles fail?

aaudiber: good point, removed the closer

calvinjia:

PR #8844, opened 2019-04-17
Enable embedded journal in single master local alluxio cluster

LuQQiu:
For all the tests that
(1) use single master local alluxio cluster
(2) can support both embedded journal and ufs journal 

This PR:
use embedded journal as the default journal type
support easy change embedded journal to ufs journal


LuQQiu:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -206,8 +208,8 @@ private synchronized void initServer() {
     mStateMachine = new JournalStateMachine(mJournals);
     mServer = CopycatServer.builder(getLocalAddress(mConf))
         .withStorage(storage)
-        .withElectionTimeout(Duration.ofMillis(mConf.getElectionTimeoutMs()))
         .withHeartbeatInterval(Duration.ofMillis(mConf.getHeartbeatIntervalMs()))
+        .withElectionTimeout(Duration.ofMillis(mConf.getElectionTimeoutMs()))
```
aaudiber: I think this could fail if someone wants to set (heartbeat timeout, election timeout) to (10s, 30s). Could we add a test for this?

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterSync.java
```diff
@@ -124,7 +124,17 @@ public void heartbeat() {
     Command cmdFromMaster = null;
     List<alluxio.grpc.Metric> metrics = new ArrayList<>();
     for (Metric metric : MetricsSystem.allWorkerMetrics()) {
-      metrics.add(metric.toProto());
+      try {
+        metrics.add(metric.toProto());
```
aaudiber: why do we hit NPE?

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
```diff
@@ -128,6 +130,21 @@ public void initConfiguration() throws IOException {
...
+    if (mJournalType == JournalType.UFS) {
+      ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS);
+    } else {
+      ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.EMBEDDED);
+      ServerConfiguration.set(PropertyKey.MASTER_EMBEDDED_JOURNAL_ADDRESSES,
+          String.format("%s:%s", mHostname,
+              ServerConfiguration.get(PropertyKey.MASTER_EMBEDDED_JOURNAL_PORT)));
+      ServerConfiguration.set(PropertyKey.MASTER_RPC_ADDRESSES, String.format("%s:%s", mHostname,
+          ServerConfiguration.get(PropertyKey.MASTER_RPC_PORT)));
+      if (ServerConfiguration.getBytes(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX) < 65) {
```
aaudiber: it's not good to silently change the configuration when other code explicitly tries to set the configuration to something illegal. This could be confusing to the person trying to set the log size to 64. We should allow it to fail so that we can fix their code

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
```diff
@@ -128,6 +130,21 @@ public void initConfiguration() throws IOException {
...
+        ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX,
+            PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX.getDefaultValue());
+      }
+      ServerConfiguration.set(PropertyKey.MASTER_EMBEDDED_JOURNAL_SHUTDOWN_TIMEOUT, "100ms");
```
aaudiber: can we make this a test configuration default?

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
```diff
@@ -128,6 +130,21 @@ public void initConfiguration() throws IOException {
...
       ServerConfiguration.set(entry.getKey(), entry.getValue());
     }
+    if (mJournalType == JournalType.UFS) {
+      ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS);
+    } else {
+      ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.EMBEDDED);
+      ServerConfiguration.set(PropertyKey.MASTER_EMBEDDED_JOURNAL_ADDRESSES,
+          String.format("%s:%s", mHostname,
+              ServerConfiguration.get(PropertyKey.MASTER_EMBEDDED_JOURNAL_PORT)));
+      ServerConfiguration.set(PropertyKey.MASTER_RPC_ADDRESSES, String.format("%s:%s", mHostname,
```
aaudiber: this shouldn't be needed since we've already set the embedded journal addresses

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
```diff
@@ -128,6 +130,21 @@ public void initConfiguration() throws IOException {
...
+          ServerConfiguration.get(PropertyKey.MASTER_RPC_PORT)));
+      if (ServerConfiguration.getBytes(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX) < 65) {
+        ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX,
+            PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX.getDefaultValue());
+      }
+      ServerConfiguration.set(PropertyKey.MASTER_EMBEDDED_JOURNAL_SHUTDOWN_TIMEOUT, "100ms");
+    }
     ServerConfiguration.set(PropertyKey.TEST_MODE, true);
     ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS);
```
aaudiber: looks like we're overriding the journal type here anyway

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
```diff
@@ -139,6 +156,7 @@ public void initConfiguration() throws IOException {
   public void startMasters() throws Exception {
     mMaster = LocalAlluxioMaster.create(mWorkDirectory, true);
     mMaster.start();
+    CommonUtils.PROCESS_TYPE.set(CommonUtils.ProcessType.CLIENT);
```
aaudiber: why is this needed?

LuQQiu:
@aaudiber Yeah, the `PROCESS_TYPE` is a little annoying in the test, 
We mainly use it when we need to build the raft journal system or format it.
The main place we use is in the `JournalSystem.build()`, are we able to get rid of it?

aaudiber:
@LuQQiu Can we pass the `ServiceType` to the builder?

aaudiber:
@LuQQiu could we pass `ServiceType` instead of `ProcessType`?

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
```diff
@@ -78,7 +83,9 @@ public void start() throws Exception {
       mLeaderSelector.waitForState(State.PRIMARY);
       if (gainPrimacy()) {
         mLeaderSelector.waitForState(State.SECONDARY);
-        losePrimacy();
+        if (mRunning) {
```
aaudiber: when mRunning is false, how do we exit the while loop?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: The copycat server is really strange.  `electionTimeout=100ms, heartbeatTimeout=20ms` failed no master we set heartbeat interval or election timeout first.
but `electionTimeout=120ms, heartbeatTimeout=20ms` succeed when setting the heartbeat interval first

LuQQiu: I changed back the sequence (`withElection` first and `withHeartbeat` later).
As long as election timeout is bigger than the default heartbeat interval `250ms`, there will be no issue with copycat.

```
      conf.put(PropertyKey.MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT, "260ms");
      conf.put(PropertyKey.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL, "50ms");
```

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterSync.java
LuQQiu: This is because previously we change PROCESS_TYPE=MASTER in the tests which cause the metrics to be messy. Some metrics are master metrics and  `mHostname=null` which make the `Metric.toProto()` to throw NPE.

When `PROCESS_TYPE=CLIENT`, all metrics have `mHostname` and will not cause NPE now.

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
LuQQiu: I changed to set the `MASTER_JOURNAL_LOG_SIZE` to its default value in `ConfigurationTestUtils.testConfigurationDefaults`. Users can set explicit values to this property in their test and if the value is smaller than 65, the person will notice that the test fail and find out the reason.

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
LuQQiu: Moved to test configuration default

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
LuQQiu: Removed, thanks

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
LuQQiu: Removed, thanks

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
LuQQiu: Add the `if(mRunning) {losePrimacy()} else {break the while loop}", thanks!

minicluster/src/main/java/alluxio/master/LocalAlluxioCluster.java
LuQQiu: Emmm, looks like both `EMBEDDED_JOURNAL_ADDRESSES` and `RPC_ADDRESSES` are not needed, since we set the `MASTER_HOSTNAME` and all the ports.

LuQQiu:
@aaudiber Addressed the comments, PTAL, thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
aaudiber: What would it take to make it work even when the election timeout is smaller than `250ms`?

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
```diff
@@ -44,11 +44,12 @@ public static InstancedConfiguration defaults() {
    * them to test specific functionality.
    *
    * @param hostname the master hostname
+   * @param journalType the master journal type
    * @param workDirectory the work directory in which to configure the journal and tiered storage
    * @return the configuration
    */
   public static Map<PropertyKey, String> testConfigurationDefaults(AlluxioConfiguration alluxioConf,
-      String hostname, String workDirectory) {
+      String hostname, String journalType, String workDirectory) {
```
aaudiber: can we make the defaults independent of the journal type? We don't know at this point what the actual journal type will be, since the journal type might be overridden in the cluster configuration

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
```diff
@@ -189,7 +189,7 @@ public Builder setQuietTimeMs(long quietTimeMs) {
     /**
      * @return a journal system
      */
-    public JournalSystem build() {
+    public JournalSystem build(CommonUtils.ProcessType processType) {
```
aaudiber: since we only use processType to determine serviceType, why not directly pass the serviceType?

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
```diff
@@ -78,7 +83,11 @@ public void start() throws Exception {
       mLeaderSelector.waitForState(State.PRIMARY);
       if (gainPrimacy()) {
         mLeaderSelector.waitForState(State.SECONDARY);
-        losePrimacy();
+        if (mRunning) { // To identify whether stop is called
```
aaudiber: we can short circuit this:
```java
if (mRunning) {
  break;
}
losePrimacy();
```

Should we also check after `mLeaderSelector.waitForState(State.PRIMARY)`?

minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java
```diff
@@ -62,16 +63,19 @@
 
   protected String mWorkDirectory;
   protected String mHostname;
+  protected JournalType mJournalType;
```
aaudiber: I don't think we need this, we already have `ServerConfiguration` giving us the journal type

LuQQiu: This is because some tests do not support embedded journal. For those test, we allow them to set journal type. 
but the `journal type in ServerConfiguration` may also not be correct, it could be set by previous test and we do not reset!

I may think about adding a `default test journal type` property or make sure `journal type in ServerConfiguration` is reset after every test.

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
LuQQiu: Because we are not necessarily building Embedded journal system (there are also noop journal system and ufs journal system), always passing in `ServiceType.MASTER_RAFT` and `ServiceType.JOB_MASTER_RAFT` feels like we are building embedded journal system.

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
LuQQiu: Now the `testConfigurationDefaults` is isolated from the actual journal type. The journal type is not a must-have in this method, thanks

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
LuQQiu: short circuit it and also check for primary state.

LuQQiu:
@aaudiber Since the local Alluxio cluster change the global configuration and added the configuration rule to all the tests using local alluxio cluster is complicated, I added back the `surefire` property `alluxio.test.journal.type`. PTAL, thanks!

minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java
aaudiber: if a test doesn't support embedded journal, should it explicitly configure itself to use ufs journal?

I think we already have this line to reset configuration after each test: https://github.com/Alluxio/alluxio/blob/master/minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java#L199

aaudiber:
@LuQQiu I'm still trying to understand why we need a separate journal property outside `ServerConfiguration`. Can you give an example of where we need to call `setJournalType` instead of setting the journal type configuraiton property?

LuQQiu:
@aaudiber 
Suppose we only have one Property key: `MASTER_JOURNAL_TYPE` to control the global test journal type. All the tests that support both embedded journal and UFS journal should use the journal type defined by `MASTER_JOURNAL_TYPE`.

(1) It works well in `MultiProcessCluster` since the value of `MASTER_JOURNAL_TYPE` will never be changed.

For example,  Even if `MASTER_JOURNAL_TYPE=EMBEDDED`, `JournalBackupIntegrationTest.backupRestoreZk()` test still use a `MultiProcessCluster` with Zookeeper. Instead of changing the global configuration `MASTER_JOURNAL_TYPE` to `UFS`, we store process level properties in a map and passing them to each jvm process by `-Dkey=value`.

(2) However, in multi-thread cluster -> the `LocalAlluxioCluster`, we do not have a separate environment allow us to set thread level configuration. 

If all the tests can support both embedded journal and ufs journal, it will be easy, we can directly set once the `MASTER_JOURNAL_TYPE` before tests and never modify this value. However, some tests target embedded journal or ufs journal. For example, the `JournalToolTest.java` and `SecondaryMasterTest` can only use `UFS` journal and they need to set the global configuration `MASTER_JOURNAL_TYPE=UFS` when initializing the `LocalAlluxioClusster` configuration. The `MASTER_JOURNAL_TYPE` will be changed in this case.

That's why I want another property key that is write-once and never be modified.

aaudiber:
@LuQQiu Even in `LocalAlluxioCluster`, tests are run one at a time, not concurrently, so each test sets its own configuration independent from other tests. The environment is controlled by setting the global configuration at the start, then launching the cluster. As far as I can tell, we never need to change the journal type in the middle of a test.

What if we change from calling

```java
  new LocalAlluxioClusterResource.Builder()
  // ...
  .setJournalType(JournalType.UFS)
```

to

```java
  new LocalAlluxioClusterResource.Builder()
  // ...
  .setProperty(PropertyKey.MASTER_JOURNAL_TYPE, "UFS")
```

LuQQiu:
@aaudiber Thanks for the advice! I agreed that we do not need the `setJournalType` in `LocalAlluxioClusterResource.Builder`. Will remove this part later.

In `LocalAlluxioClusterResource.start` we still need to override the `MASTER_JOURNAL_TYPE` IN `ServerConfiguration`. So I still think it will be safer to have another global test journal type. 
```
// Init configuration for integration test
    mLocalAlluxioCluster.initConfiguration();
    // Overwrite the test configuration with test specific parameters
    for (Entry<PropertyKey, String> entry : mConfiguration.entrySet()) {
      ServerConfiguration.set(entry.getKey(), entry.getValue());
    }
```

aaudiber:
@LuQQiu why do we need to override `MASTER_JOURNAL_TYPE` in `ServerConfiguration`? If we want a test to use the default journal type (may be overridden by passing system property `-Dalluxio.master.journal.type`), we don't need to do anything because the configuration will automatically pick up the system property. For tests that should always use a specific journal type, they can set that type with
```java
 new LocalAlluxioClusterResource.Builder()
  // ...
  .setProperty(PropertyKey.MASTER_JOURNAL_TYPE, "UFS")
```
which will override the default journal type in `LocalAlluxioClusterResource.start`.

My concern with adding a second way to set the journal type is that it could be confusing which journal configuration should be respected, and it adds unnecessary complexity.

LuQQiu:
@aaudiber 
Still confusing.
(1) Situation one:
 our default `alluxio.master.journal.type=EMBEDDED`.
The first several tests using `LocalAlluxioCluster` work as expected.

Then a test 
```
 new LocalAlluxioClusterResource.Builder()
  // ...
  .setProperty(PropertyKey.MASTER_JOURNAL_TYPE, "UFS")
```
The `alluxio.master.journal.type` in `ServerConfiguration` will be changed to `UFS` in this test.

Then following tests in the same thread will running with the `alluxio.master.journal.type=UFS` or `alluxio.master.journal.type=EMBEDDED`?

(2) Situation two:
`-Dalluxio.master.journal.type=UFS` is passed in.
The first several tests using `LocalAlluxioCluster` work as expected.

Then a test `  .setProperty(PropertyKey.MASTER_JOURNAL_TYPE, "EMBEDDED")`. Can the following tests in the same thread still pick up the system property value `UFS`?

aaudiber:
@LuQQiu I suggest trying it out to double check, but I believe in situation (1) the following tests will go back to using EMBEDDED because the configuration is reset at the end of each test. If it isn't it's a bug and it should be.

In situation (2), later tests should still be able to pick up the value of the system property. We never modify system properties, so their values will stay the same. When we reset the configuration, we will re-discover the default journal type from the system property.

minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java
LuQQiu: Yeah! In both situations, the configuration will be reset or reload the system property. The master journal type will be consistent through tests. Thanks!

LuQQiu:
@aaudiber Thanks!! I finally got it! The `LocalAlluxioCluster.stop` will reset the server configuration which will reload the default property value or system property. Both situation 1 & 2 will maintain the master.journal.type across different tests. Changed back to use one property to control the journal type and remove the `setJournalType`.

aaudiber:

PR #8845, opened 2019-04-17
Improve user logging file structure

ns1123:
Modified the user logging directory to prevent errors from occurring when the user running a command is different from the alluxio installation user.

@ZacBlanco or @aaudiber PTAL


dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -305,10 +305,11 @@ func generateTarball(hadoopClients []string) error {
 	fmt.Printf("Creating %s:\n", tarball)
 
 	for _, dir := range []string{
-		"assembly", "client", "logs", "integration/fuse", "integration/checker",
+		"assembly", "client", "logs", "integration/fuse", "integration/checker", "logs/user",
 	} {
 		mkdir(filepath.Join(dstPath, dir))
 	}
+    os.Chmod(filepath.Join(dstPath, "logs/user"), 0777)
```
ZacBlanco: Don't forget to handle the error if we can't chmod the logs/user folder

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -330,6 +330,13 @@ public String toString() {
           .setIgnoredSiteProperty(true)
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
+  public static final PropertyKey USER_LOGS_DIR =
+      new Builder(Name.USER_LOGS_DIR)
+          .setDefaultValue(String.format("${%s}/user", Name.LOGS_DIR))
+          .setDescription("The path to store user log files.")
```
ZacBlanco: nit: I think someone reading this without background might not know what "user log files" are. Could we say something similar to the tune of "logs from alluxio client operations"? Even I'm not a huge fan of my suggestion, but I think I just want to make it more clear it's logs from client operations.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3628,6 +3635,7 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String INTEGRATION_YARN_WORKERS_PER_HOST_MAX =
         "alluxio.integration.yarn.workers.per.host.max";
     public static final String LOGGER_TYPE = "alluxio.logger.type";
+    public static final String USER_LOGS_DIR = "alluxio.user.logs.dir";
```
ZacBlanco: nit: I think we try to keep these in alphabetical order

core/common/src/main/java/alluxio/conf/PropertyKey.java
aaudiber: +1 for clarifying what a user log contains. How about "logs from alluxio CLI operations"?

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -334,7 +335,7 @@ func generateTarball(hadoopClients []string) error {
 	addAdditionalFiles(srcPath, dstPath, hadoopVersion, version)
 
 	chdir(cwd)
-	run("creating the distribution tarball", "tar", "-czvf", tarball, dstDir)
+	run("creating the distribution tarball", "tar", "-czpvf", tarball, dstDir)
```
aaudiber: Does this `-p` do anything? My understanding is that permissions are preserved by default on the create path, and `-p` only affects the extract path

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -305,11 +305,16 @@ func generateTarball(hadoopClients []string) error {
 	fmt.Printf("Creating %s:\n", tarball)
 
 	for _, dir := range []string{
-		"assembly", "client", "logs", "integration/fuse", "integration/checker",
+		"assembly", "client", "logs", "integration/fuse", "integration/checker", "logs/user",
 	} {
 		mkdir(filepath.Join(dstPath, dir))
 	}
 
+    err := os.Chmod(filepath.Join(dstPath, "logs/user"), 0777)
```
aaudiber: the Gopher idiom for this common pattern is
```go
if err := os.Chmod(filepath.Join(dstPath, "logs/user"), 0777); err != nil {
    return err
}

core/common/src/main/java/alluxio/conf/PropertyKey.java
ns1123: Fixed.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ns1123: done.

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
ns1123: left in from me figuring out the same thing :). Fixed.

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
ns1123: Noted and fixed.

ns1123:
@aaudiber fixes done. Please help to merge.

aaudiber:
@ns1123 just waiting for build to pass, then I'll merge

aaudiber:

aaudiber:
alluxio-bot, cherry-pick to branch-1.8 please

Issue #8404, opened 2019-02-14
Propagate options from OutStreamOptions when scheduling persist


ZacBlanco:
**Alluxio Version:** 2.0.0-SNAPSHOT

**Describe the bug**
When calling  `scheduleAsyncPersist` inside of `FileOutStream` the `CommonPOptions` are not propagated in the `persist` request. This could lead to odd behavior when clients make calls with different request options.

Bug found when discussing #8369 

**To Reproduce**
N/A

**Expected behavior**
The applicable fields in `OutStreamOptions` should be used in the `scheduleAsyncPersist` call

**Additional context**
N/A

ZacBlanco:
@calvinjia can you close this?

PR #8855, opened 2019-04-18
Propagate common options for async persist request

ggezer:
Fix for https://github.com/Alluxio/alluxio/issues/8404

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
@@ -367,9 +367,9 @@ public final class OutStreamOptions {
       return false;
     }
     OutStreamOptions that = (OutStreamOptions) o;
-    return mCommonOptions.equals(that.mCommonOptions)
-        && Objects.equal(mAcl, that.mAcl)
+    return Objects.equal(mAcl, that.mAcl)
         && Objects.equal(mBlockSizeBytes, that.mBlockSizeBytes)
+        && Objects.equal(mCommonOptions, that.mCommonOptions)
         && Objects.equal(mGroup, that.mGroup)
         && Objects.equal(mLocationPolicy, that.mLocationPolicy)
         && Objects.equal(mMode, that.mMode)
@@ -386,9 +386,9 @@ public final class OutStreamOptions {
   @Override
   public int hashCode() {
     return Objects.hashCode(
-        mCommonOptions,
         mAcl,
         mBlockSizeBytes,
+        mCommonOptions,
         mGroup,
         mLocationPolicy,
         mMode,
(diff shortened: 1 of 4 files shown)


ggezer:
@ZacBlanco @calvinjia PTAL.

core/client/fs/src/test/java/alluxio/client/file/options/OutStreamOptionsTest.java
```diff
@@ -119,14 +119,14 @@ public void fields() throws Exception {
     int writeTier = random.nextInt();
     WriteType writeType = WriteType.NONE;
 
+    mConf.set(PropertyKey.USER_FILE_CREATE_TTL_ACTION, "FREE");
```
ZacBlanco: I noticed below you're only checking the ttlAction instead of both. Is that intentional? Can we set the ttl here as well?

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -437,15 +406,14 @@ public int hashCode() {
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
+        .add("common-options", mCommonOptions)
```
ZacBlanco: could we do `commonOptions` to follow the pattern for the rest of the toString items? And also put this in alphabetical order?

what does the `toString` of `mCommonOptions` look like?

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -416,15 +386,14 @@ public boolean equals(Object o) {
   @Override
   public int hashCode() {
     return Objects.hashCode(
+        mCommonOptions,
```
ZacBlanco: Can we use alphabetical order? :)

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -396,15 +367,14 @@ public boolean equals(Object o) {
       return false;
     }
     OutStreamOptions that = (OutStreamOptions) o;
-    return Objects.equal(mAcl, that.mAcl)
+    return mCommonOptions.equals(that.mCommonOptions)
```
ZacBlanco: Can we use alphabetical order? :)

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
ggezer: It's pretty much harmonious with MoreObjects.toStringHelper.
You can check any RPC log to see what it looks like.

ggezer:
After checking the code I see that common options for async-persist request is not even used on master side. So no  meaningful testing is possible at this point. @calvinjia WDYT?

ggezer:
Ping!

ZacBlanco:
I think we could still add a unit test that checks that the RPC which is being made at least contains the expected common options. Even if not used on the master, it is good verification that options are propagated properly

ggezer:
Opened a separate issue for tracking the issue of not making use of common options on master side.
https://github.com/Alluxio/alluxio/issues/8891

madanadit:

PR #8859, opened 2019-04-18
Refactor common server code

gpang:
Move server common code into the server/common module.


ggezer:
Managed channel pool is a client code, so GrpcManagedChannelPoolTest.java should stay in core-common.

PR #8871, opened 2019-04-19
Reduce number of startup consistency check threads

aaudiber:
It isn't good to pin all processors to consistency checking.


aaudiber:
Moved to https://github.com/Alluxio/alluxio/pull/8889

PR #8872, opened 2019-04-19
Add a framework for configuration propagation

cheng-chang:
This PR adds grpc services to return the computed versions (hashes) of both cluster and path level configurations, and adds client heartbeats to continuously monitor the versions, if versions change, then the FileSystemContext will be re-initialized. Note that this PR only adds the grpc communication and client-server heartbeat framework, details like how to compute the versions are left for future work.


cheng-chang:
@aaudiber @calvinjia PTAL

PR #8873, opened 2019-04-20
Add a framework for configuration propagation

cheng-chang:
This PR adds grpc services to return the computed versions (hashes) of both cluster and path level configurations, and adds client heartbeats to continuously monitor the versions, if versions change, then the FileSystemContext will be re-initialized.

Note that this PR only adds the grpc communication and client-server heartbeat framework, details like how to compute the versions are left for future work.


cheng-chang:
@aaudiber @calvinjia PTAL

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -300,6 +331,20 @@ public synchronized void close() throws IOException {
     }
   }
 
+  /**
+   * Closes the context, update configuration from meta master, then re-initializes the context.
+   *
+   * @throws IOException when failed to close the context or update configuration
+   */
+  public synchronized void reinit() throws IOException {
```
aaudiber: how does this affect outstanding RPCs?

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -97,15 +102,26 @@
   private volatile BlockMasterClientPool mBlockMasterClientPool;
 
   // Closed flag for debugging information.
```
calvinjia: I don't think this comment is accurate anymore?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -352,6 +352,18 @@ public ConfigCheckReport getConfigCheckReport() {
     return pathConfig;
   }
 
+  @Override
+  public String getConfigurationVersion() {
+    // TODO(cc)
```
calvinjia: Will this be implemented in another PR, is there significant logic here?

cheng-chang: yes, it's in another PR, core logic is in https://github.com/Alluxio/alluxio/pull/8896.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -97,15 +102,26 @@
...
-  private final AtomicBoolean mClosed;
+  private AtomicBoolean mClosed = new AtomicBoolean(true);
```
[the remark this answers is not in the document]
cheng-chang: added back final, I thought it is recreated in reinit.

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
calvinjia: Got it thanks

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
cheng-chang: When closing, all grpc channels and underlying connections will be closed, and each grpc channel has its ID, a response from the rpc sent by old client will not be received by the new client.

aaudiber: Will config propagation lead to user-facing errors, or will we perform retries, similar to what we do when the primary master fails over to a standby

cheng-chang: It will lead to user-facing errors

aaudiber: I don't think this acceptable, what would it take to prevent updating conf from auto-failing outstanding jobs?

cheng-chang: Retrying will cause inconsistencies, e.g. when writing a file, if previously it's CACHE but the conf is updated to THROUGH, then part of the file could be only cached while the remaining is persisted.

aaudiber: @calvinjia is it acceptable to fail all outstanding rpcs when conf is updated? I could see this being ok for compute frameworks if they clean up and retry on failure, but not all clients will do this. We put a lot of effort into making master failover transparent to compute, and this seems similar.

@cheng-chang retrying doesn't necessarily cause inconsistencies - we could allow the file write to complete using the configuration it was started with.

calvinjia: For the stateful interactions (file in/out streams), could we keep the configuration constant (ie. once a stream is created, it always has that configuration). For everything else I think we can retry.

cheng-chang:

Issue #8870, opened 2019-04-19
Compensate ForkJoinPool during UFS ops


ggezer:
**Is your feature request related to a problem? Please describe.**
Currently, master RPCs are handled by ForkJoinPool and that needs external hints (via managed blocker API) for compensating the pool whenever a blocking operation is taking place. AsyncJournalWriter is giving out such hints during journal flushes. We would need to do the same for FileSystemMaster calls while they're doing UFS operations (create/delete). That way slow UFSs will not block progress on RPCs.

**Describe the solution you'd like**
Migrate ObjectUnderFileSystem class for managed blocking, as base class of the potentially slowest UFSs.

**Describe alternatives you've considered**
FileSystemMaster level analysis on potentially slow paths could yield better overall performance bump.

**Urgency**
Better before 2.0 due to unknowns.

PR #8884, opened 2019-04-21
Call UFS operations under ForkJoinPool managed blocking

ggezer:
After switching to ForkJoinPool, master RPC throughput with some UFS types have been degradation. This is due to UFS operations being slow on them, and ForkJoinPool not knowing that its worker is stuck on a blocking operation. This fix introduces a wrapper over UnderFileSystem that notifies the ForkJoinPool during certain UFS operations.

https://github.com/Alluxio/alluxio/issues/8870

### DIFF
--- a/core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
+++ b/core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
@@ -1,0 +1,654 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.concurrent;
+
+import alluxio.AlluxioURI;
+import alluxio.SyncInfo;
+import alluxio.collections.Pair;
+import alluxio.concurrent.jsr.ForkJoinPool;
+import alluxio.security.authorization.AccessControlList;
+import alluxio.security.authorization.AclEntry;
+import alluxio.security.authorization.DefaultAccessControlList;
+import alluxio.underfs.UfsDirectoryStatus;
+import alluxio.underfs.UfsFileStatus;
+import alluxio.underfs.UfsMode;
+import alluxio.underfs.UfsStatus;
+import alluxio.underfs.UnderFileSystem;
+import alluxio.underfs.options.CreateOptions;
+import alluxio.underfs.options.DeleteOptions;
+import alluxio.underfs.options.FileLocationOptions;
+import alluxio.underfs.options.ListOptions;
+import alluxio.underfs.options.MkdirsOptions;
+import alluxio.underfs.options.OpenOptions;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Forwarder for {@link UnderFileSystem} objects that works through with ForkJoinPool's
+ * managed blocking.
+ *
+ * If UFS calls are being done on a {@link alluxio.concurrent.jsr.ForkJoinWorkerThread}, then
+ * this forwarder will make sure UFS operations are treated as blocking operations
+ * for compensating the ForkJoinPool.
+ *
+ */
+public class ManagedBlockingUfsForwarder implements UnderFileSystem {
+
+  /** Underlying {@link UnderFileSystem}. **/
+  private UnderFileSystem mUfs;
+
+  /**
+   * Creates {@link ManagedBlockingUfsForwarder} instance.
+   *
+   * @param ufs the underlying ufs
+   */
+  public ManagedBlockingUfsForwarder(UnderFileSystem ufs) {
+    mUfs = ufs;
+  }
+
+  @Override
+  public void cleanup() throws IOException {
+    mUfs.cleanup();
+  }
+
+  @Override
+  public void connectFromMaster(String hostname) throws IOException {
+    mUfs.connectFromMaster(hostname);
+  }
+
+  @Override
+  public void connectFromWorker(String hostname) throws IOException {
+    mUfs.connectFromWorker(hostname);
+  }
+
+  @Override
+  public OutputStream create(String path) throws IOException {
+    return new ManagedBlockingUfsMethod<OutputStream>() {
+      @Override
+      public OutputStream execute() throws IOException {
+        return mUfs.create(path);
+      }
+    }.get();
+  }
+
+  @Override
+  public OutputStream create(String path, CreateOptions options) throws IOException {
+    return new ManagedBlockingUfsMethod<OutputStream>() {
+      @Override
+      public OutputStream execute() throws IOException {
+        return mUfs.create(path, options);
+      }
+    }.get();
+  }
+
+  @Override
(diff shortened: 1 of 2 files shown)


ggezer:
@calvinjia PTAL.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
```diff
@@ -0,0 +1,654 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    mUfs.close();
+  }
+
+  /**
+   * Utility class used to isolate calls into underlying UFS from concurrency compensation logic.
+   * Note: This class used to make calls with a return value.
```
calvinjia: Could `Void` be used for the calls without a return value?

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
```diff
@@ -0,0 +1,654 @@
...
+  }
+
+  @Override
+  public void cleanup() throws IOException {
+    mUfs.cleanup();
+  }
+
+  @Override
+  public void connectFromMaster(String hostname) throws IOException {
+    mUfs.connectFromMaster(hostname);
```
calvinjia: Should this be wrapped in the managed blocker?

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
```diff
@@ -0,0 +1,654 @@
...
+      @Override
+      public UfsFileStatus execute() throws IOException {
+        return mUfs.getExistingFileStatus(path);
+      }
+    }.get();
+  }
+
+  @Override
+  public String getFingerprint(String path) {
+    return mUfs.getFingerprint(path);
```
calvinjia: Should this be wrapped?

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
```diff
@@ -0,0 +1,654 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Forwarder for {@link UnderFileSystem} objects that works through with ForkJoinPool's
+ * managed blocking.
+ *
+ * If UFS calls are being done on a {@link alluxio.concurrent.jsr.ForkJoinWorkerThread}, then
```
calvinjia: The idea is if there possibly is a UFS interaction, it should be wrapped in a managed blocker?

ggezer: True. Preferably an interaction that is slow.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: Does getting a fingerprint entail an UFS interaction? If it does, is it slow?

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: What does this do? If it's slow and entails an UFS interaction, then yes.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: We would need to return "null" (or anything) from void methods, like below:

```
@Override
  public void setAclEntries(String path, List<AclEntry> aclEntries) throws IOException {
    new ManagedBlockingUfsMethod<Void>() {
      @Override
      public Void execute() throws IOException {
        mUfs.setAclEntries(path, aclEntries);
        return null;
      }
    }.get();
}
```

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: Based on this reasoning, please advise on all methods that you think should be or should not have been wrapped.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
calvinjia: Take a look at the implementations of `connectFromMaster`. I think some may require UFS interaction but most do not.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
calvinjia: Right, I prefer this over having two separate classes, WDYT?

ggezer: Agreed.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
calvinjia: It varies based on the implementation, some call `getStatus` which will be slow (but that is already wrapped in ManagedBlocker).

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
calvinjia: Is there a downside to wrapping a call with `ManagedBlocker`? Also if a method calls another method, is it fine to have chained `ManagedBlockers`?

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: I see the slower UFSs don't even implement this method. So it should be fine to exclude them from managed blocking. Because managing blocking has its own overhead, which could make fast operations slower when used liberally.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: Then it's better to exclude it and postpone managed blocking only when `getStatus` is called.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
ggezer: Managed blocking has its overhead, so yes wrapping a fast code with `ManagedBlocker` will reduce executor's efficiency. With the current state of this code, S3 UFS has been recovered completely. So I believe there could only be methods to exclude from managed blocking.

For chain blocking, I have made an experiment by implementing `managedBlockEnter` and `managedBlockExit` methods which keep call state in TLS and don't compensate when called in a chain. I didn't see any difference between that and this usage, so continued with this instead to not increase complexity of `ForkJoinPool` by adding a custom API to it.

core/common/src/main/java/alluxio/concurrent/ManagedBlockingUfsForwarder.java
```diff
@@ -483,30 +483,36 @@ public AlluxioURI resolveUri(AlluxioURI ufsBaseUri, String alluxioPath) {
 
   @Override
   public void setAclEntries(String path, List<AclEntry> aclEntries) throws IOException {
-    new ManagedBlockingUfsFunction() {
+    new ManagedBlockingUfsMethod<Void>() {
       @Override
-      public void execute() throws IOException {
+      public Void execute() throws IOException {
         mUfs.setAclEntries(path, aclEntries);
+        // For obeying the contract. Returned object not utilized.
```
calvinjia: These comments are not necessary, the signature should be sufficient explanation.

calvinjia:

Issue #8825, opened 2019-04-16
fs mount command show all mount point as ufs type local after master restart


LuQQiu:
**Alluxio Version:**
2.0.0-preview

**Describe the bug**
```
// Start alluxio
$./bin/alluxio-start.sh all SudoMount
// Mount s3 to alluxio
$./bin/alluxio fs mount --option aws.accessKeyId=**** --option aws.secretKey=*** /s3-mnt s3a://alluxio-test/
// fs mount show the right ufs type s3
$./bin/alluxio fs mount
s3a://alluxio-test-fuse/                             on  /s3-mnt  (s3, capacity=-1B, used=-1B, not read-only, not shared, properties={aws.secretKey=*****, aws.accessKeyId=*****})
/Users/alluxio/alluxioFolder/alluxio/underFSStorage  on  /        (local, capacity=464.77GB, used=-1B(0%), not read-only, not shared, properties={})

// restart alluxio
$./bin/alluxio-start.sh all SudoMount

// All mount point ufs type is local now
$./bin/alluxio fs mount
s3a://alluxio-test-fuse/                             on  /s3-mnt  (local, capacity=-1B, used=-1B, not read-only, not shared, properties={aws.secretKey=*****, aws.accessKeyId=*****})
/Users/alluxio/alluxioFolder/alluxio/underFSStorage  on  /        (local, capacity=464.77GB, used=-1B(0%), not read-only, not shared, properties={})

```

**Expected behavior**
Ufs type should always show the right value across master restart

aaudiber:
Does this only affect how the CLI displays info, or could there be other impact on the mount point?

LuQQiu:
It affects one of the future PR which is about hiding the aws (or other object stores) secret keys.
I will take a look into this issue when I am working on that PR.

apc999:
@LuQQiu has this issue been fully addressed by PR https://github.com/Alluxio/alluxio/pull/8885? If yes, you can close this Issue

apc999:
also is this issue also in 1.8? if yes, can you cherry-pick your fix to 1.8 branch?

LuQQiu:
I checked that branch 1.8 do not have this issue. It's a newly occurred issue. Fixed in #8885

PR #8885, opened 2019-04-22
Fix the wrong argument in UfsManager.addMount()

LuQQiu:
#8825

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -560,7 +560,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
         UnderFileSystemConfiguration ufsConf =
             UnderFileSystemConfiguration.defaults(ServerConfiguration.global())
             .createMountSpecificConf(mountInfo.getOptions().getPropertiesMap());
-        mUfsManager.addMount(mountInfo.getMountId(), new AlluxioURI(key), ufsConf);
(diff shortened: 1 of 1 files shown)


LuQQiu:
@madanadit A small fix for the mount information. The key here is the mounted address in alluxio namespace, but the parameter required in `addMount` is ufs uri. This PR fixed the issue that after `fs mount` get wrong ufs type after restart. PTAL, thanks

LuQQiu:
@madanadit The bug cannot be tested in unit test and I found it hard to test in integration test as well.
In the designed integration test, my steps are:
(1) mount a ufs other than local ufs
(2) getMountPointInfo and make sure the ufs type is correct (is not local)
(3) restart the local cluster
(4) getMountPointInfo again and make sure the ufs type is correct (is not local)

but step one is hard to fulfill in our integration test: we do not support mounting systems other than local ufs.
and the 
```
    String localPath = com.google.common.io.Files.createTempDir().getAbsolutePath();
    UnderFileSystem ufs =
        new DelegatingUnderFileSystem(UnderFileSystem.Factory.create(localPath,
            ServerConfiguration.global())) {
          @Override
          public String getUnderFSType() {
            return "FakeType";
          }
        };
    UnderFileSystemFactoryRegistry.register(new DelegatingUnderFileSystemFactory(ufs));
    mFsMaster.mount(alluxioUri, ufsUri, MountContext.defaults());
    MountPointInfo info = mFsMaster.getMountPointInfo(alluxioUri);
```
the `ufsType` in `info.getUfsType` is still `local`.

I think this bug is a rare happen case and hard to add tests for it. Could we leave it without test? 
Thanks

madanadit:

PR #8889, opened 2019-04-23
Remove startup consistency check

aaudiber:
The startup consistency check doesn't scale well,
especially when metadata is stored off-heap. Users
can still call checkConsistency to get the same effect.
There are also active and passive sync mechanisms which
have been introduced since the time we implemented
startup consistency check.


calvinjia:

Issue #8857, opened 2019-04-18
gRPC authentication should not close an already closed call


ggezer:
**Alluxio Version:**
2.0-preview

**Describe the bug**
Below stack has been observed in master logs;

```
Apr 19, 2019 3:25:46 AM io.grpc.internal.SerializingExecutor run
SEVERE: Exception while executing runnable io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed@700831d5
java.lang.IllegalStateException: call already closed
        at com.google.common.base.Preconditions.checkState(Preconditions.java:507)
        at io.grpc.internal.ServerCallImpl.close(ServerCallImpl.java:172)
        at alluxio.security.authentication.AuthenticatedUserInjector.authenticateCall(AuthenticatedUserInjector.java:90)
        at alluxio.security.authentication.AuthenticatedUserInjector.access$000(AuthenticatedUserInjector.java:34)
        at alluxio.security.authentication.AuthenticatedUserInjector$1.onHalfClose(AuthenticatedUserInjector.java:66)
        at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:283)
        at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed.runInContext(ServerImpl.java:710)
        at io.grpc.interna
```
**To Reproduce**
Unknown

**Expected behavior**
gRPC authentication should remember/detect that call is closed.

**Urgency**
It's a benign error message.

**Additional context**

PR #8890, opened 2019-04-24
Stop intercepting closed/cancelled server calls

ggezer:
There are sporadic occurrences of "IllegalStateException: call already closed" in server logs. This PR fixes https://github.com/Alluxio/alluxio/issues/8857 by making sure we don't close an already closed or cancelled server call.

### DIFF
--- a/core/common/src/main/java/alluxio/security/authentication/AuthenticatedUserInjector.java
+++ b/core/common/src/main/java/alluxio/security/authentication/AuthenticatedUserInjector.java
@@ -77,9 +77,9 @@ public final class AuthenticatedUserInjector implements ServerInterceptor {
    * that is kept on auth-server.
    */
   private <ReqT, RespT> boolean authenticateCall(ServerCall<ReqT, RespT> call, Metadata headers) {
-    // Fail validation for closed/cancelled server calls.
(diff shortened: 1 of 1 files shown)


ggezer:
@bf8086 PTAL.

core/common/src/main/java/alluxio/security/authentication/AuthenticatedUserInjector.java
```diff
@@ -77,6 +77,13 @@ public void onHalfClose() {
    * that is kept on auth-server.
    */
   private <ReqT, RespT> boolean authenticateCall(ServerCall<ReqT, RespT> call, Metadata headers) {
+    // Fail validation for closed/cancelled server calls.
+    if (!call.isReady() || call.isCancelled()) {
```
bf8086: Is checking `isReady()` appropriate here? That is just an indicator whether the server can send message immediately or not. It can be false when there are too many messages in the buffer.

ggezer: Nice catch. Just checking for cancelled now.

madanadit:

PR #8894, opened 2019-04-24
Remove source  when removing key from properties

cheng-chang:
When a key is removed from AlluxioProperties, the source should also be removed. Otherwise, when getting the key, the default value will be returned, but getSource might not return Source.DEFAULT.


cheng-chang:
jenkins, test this please

cheng-chang:

PR #8895, opened 2019-04-24
Support versioning AlluxioProperties

cheng-chang:
This PR computes a hex encoded md5 has of list of key:value:source strings as the version for properties. The version is lazily computed, cached, and updated.


cheng-chang:
Closing because it relies on another PR to be merged.

PR #8905, opened 2019-04-25
Add non-generated gRPC classes under checkstyle/findbugs coverage

ggezer:
Earlier maven exclusions for generated classes were covering gRPC helper/manager classes as well since they use the same package. This PR narrows exclusions down to core-transport module and fixes hidden checkstyle/findbugs issues.


ggezer:
@gpang PTAL.

build/findbugs/findbugs-exclude.xml
```diff
@@ -2,13 +2,7 @@
 
   <Match>
     <Or>
-      <Package name="org.apache.thrift"/>
```
gpang: we don't actually have this package now, right?

core/common/src/main/java/alluxio/grpc/ServiceVersionClientServiceHandler.java
```diff
@@ -31,11 +32,16 @@
   /** Set of services that are going to be recognized by this versioning service. */
   private Set<ServiceType> mServices;
 
+  /**
+   * Creates service version handler that allows given services.
+   * @param services services to allow
```
gpang: should there be a blank line before this?

build/findbugs/findbugs-exclude.xml
ggezer: We don't

core/common/src/main/java/alluxio/grpc/ServiceVersionClientServiceHandler.java
ggezer: You mean a space between general doc and the @param doc? I can but it's pretty common through out the code.

ggezer:
@gpang No code change. PTAL.

gpang:

PR #8910, opened 2019-04-25
Implement persist on rename

calvinjia:
Adds a new flag to rename, `persist` which determines whether or not the file should be persisted after the rename operation. This can greatly improve the performance of compute frameworks using Hadoop commit protocols.


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3106,23 +3116,27 @@ private void setAttributeInternal(RpcContext rpcContext, LockedInodePath inodePa
...
-      mPersistRequests.put(inodePath.getInode().getId(), new alluxio.time.ExponentialTimer(
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_INTERVAL_MS),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_INTERVAL_MS),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_WAIT_TIME_MS),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_TOTAL_WAIT_TIME_MS)));
+      scheduleAsyncPersistenceInternal(rpcContext, inodePath.getInodeFile());
+    }
+  }
+
+  private void scheduleAsyncPersistenceInternal(RpcContext rpcContext, InodeFile inode)
```
calvinjia: @gpang Due to how rename works, we don't have a LockedInodePath that can be used, I would prefer to make the internal method take LockedInodePath, do you have any suggestions? Another alternative is not to have the internal method and have slightly modified implementation at both call sites.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2127,6 +2127,16 @@ private void renameInternal(RpcContext rpcContext, LockedInodePath srcInodePath,
 
     // Now we remove srcInode from its parent and insert it into dstPath's parent
     renameInternal(rpcContext, srcInodePath, dstInodePath, false, context);
+
+    LOG.info("Checking if persist after rename is enabled");
```
ZacBlanco: INFO seems high. Did you mean DEBUG? Same for the rest of the statements below

calvinjia: Whoops the commit that removed these seemed to have disappeared

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: Is the issue that we don't have a locked inode path from the call site? It looks like we have the full src path, but only the dst dir, so no path exists for the full dst path?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1993,6 +1993,12 @@ private void renameInternal(RpcContext rpcContext, LockedInodePath srcInodePath,
 
     // Now we remove srcInode from its parent and insert it into dstPath's parent
     renameInternal(rpcContext, srcInodePath, dstInodePath, false, context);
+
+    // Check options and determine if we should schedule async persist. This is helpful for compute
+    // frameworks that use rename as a commit operation.
+    if (context.getPersist() && srcInode.isFile() && !srcInode.isPersisted()) {
```
gpang: Does this work if the file was already written with async persist?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: I updated it to repeat some of the logic but avoid breaking any invariants around the method signatures.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: Yeah this works exactly the same as calling `scheduleAsyncPersist` on the file.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2811,6 +2811,13 @@ public String toString() {
               + "not occur).")
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_FILE_RENAME_PERSIST =
+      new Builder(Name.USER_FILE_RENAME_PERSIST)
+          .setDefaultValue("false")
+          .setDescription("Whether or not to persist any files which have been renamed. This is "
```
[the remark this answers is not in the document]
calvinjia: Async persist

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2972,18 +2986,18 @@ private void setAttributeInternal(RpcContext rpcContext, LockedInodePath inodePa
...
       throws AlluxioException, UnavailableException {
-    // We retry an async persist request until ufs permits the operation
     try (RpcContext rpcContext = createRpcContext();
         LockedInodePath inodePath = mInodeTree.lockFullInodePath(path, LockPattern.WRITE_INODE)) {
-      if (!inodePath.getInodeFile().isCompleted()) {
+      InodeFile inode = inodePath.getInodeFile();
+      if (!inode.isCompleted()) {
         throw new InvalidPathException(
-            "Cannot persist an incomplete Alluxio file: " + inodePath.getUri());
+            "Cannot persist an incomplete Alluxio file: " + inode.getName());
```
gpang: I think we want the full URI?

calvinjia: Updated, thanks.

calvinjia:

PR #8912, opened 2019-04-26

niruntao:
Change

import org.junit.Assert;
to

AND

Update all

Assert.{{ method }}*
to

{{ method }}*


niruntao:
again

apc999:
Can you remove the change to `alluxio-ui/npm-shrinkwrap.json` which is unrelated to this PR?

Issue #8916, opened 2019-04-26
Switch for disabling managed blocking for UFS


ggezer:
**Is your feature request related to a problem? Please describe.**
There could be cases where managed blocking would decrease the performance in an environment where UFS is fast. An ability to disable UFS could easily recoup such performance degradation.

**Describe the solution you'd like**
Configuration to skip managed blocking on UFS.

**Describe alternatives you've considered**
Enable managed blocking only on ObjectStoreUnderFileSystem implementations.

calvinjia:
@ggezer I think this could be a good mount level option.

ggezer:
@calvinjia Do you think UnderFileSystemConfiguration would suffice? Or you are thinking some place else?

ZacBlanco:
@ggezer has the PR for fixing this been merged?

ggezer:
> @ggezer has the PR for fixing this been merged?
Yes, this issue can be closed.

PR #8918, opened 2019-04-26
Enable managed-blocking for object stores only

ggezer:
https://github.com/Alluxio/alluxio/issues/8916

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -884,6 +884,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_RUN_WITH_MANAGEDBLOCKING =
+      new Builder(Name.UNDERFS_RUN_WITH_MANAGEDBLOCKING)
+          .setDescription("Whether to run UFS operations with managed blocking. "
+              + " This will provide RPC layer a hint that UFS is possible slow."
+              + "The default is true for object stores and false for the rest. "
+              + "unless set explicitly.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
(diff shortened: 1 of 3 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3753,6 +3762,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String UNDERFS_KODO_CONNECT_TIMEOUT =
         "alluxio.underfs.kodo.connect.timeout";
     public static final String UNDERFS_KODO_REQUESTS_MAX = "alluxio.underfs.kodo.requests.max";
+    public static final String UNDERFS_RUN_WITH_MANAGEDBLOCKING =
+        "alluxio.underfs.run.managed.block";
```
calvinjia: We generally try to make the text the same as the variable name

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
```diff
@@ -125,6 +126,17 @@ private UnderFileSystem getOrAdd(AlluxioURI ufsUri, UnderFileSystemConfiguration
         return cachedFs;
       }
       UnderFileSystem fs = UnderFileSystem.Factory.create(ufsUri.toString(), ufsConf);
+
+      // Detect whether to use managed blocking on UFS operations.
+      boolean useManagedBlocking = fs.isObjectStorage();
```
ggezer: There could be a worse design with a better Ceph performance like below:
boolean useManagedBlocking = (fs.getUnderFSType() != "s3");

calvinjia: If the Ceph performance is not substantially worse, I think we can keep it as is

calvinjia:

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
ggezer: Ceph just lost the extra gain from ForkJoinPool for add/delete. List is still better than before. So overall it's either on par or better.

PR #8922, opened 2019-04-27
[SMALLFIX] Use static imports

mana-wrym:


core/base/src/test/java/alluxio/collections/PairTest.java
```diff
@@ -11,7 +11,11 @@
 
 package alluxio.collections;
 
-import org.junit.Assert;
+import static
+org.junit.Assert.assertEquals;
```
aaudiber: this should go on the same line as `import static` - see other uses of import static in this repository for reference

apc999:
can you undo your change to `alluxio-ui/npm-shrinkwrap.json`? this file is unrelated to your change

apc999:
@mana-wrym  currently this PR still changes 40k files. which is not correct. I assume you only want to touch one single file as the new contributor task. You may close this PR and create a new one with one file changed

PR #8938, opened 2019-04-30
Optimize client metrics heartbeat and decrease client logging

ZacBlanco:
This PR aims to address two concerns.

1. The client logs a lot of information
2. The client metrics heartbeat is inefficient and
noisy in the logs.

The solutions are described below:

--- Client Logging ---
Our client liberally logs messages upon failures. Due to retries
when connecting to the master the logs tend to fill up with
useless duplicates of messages.

This commit combats this by utilizing Alluxio's Sampling
logger in a few key cases where we see excessive logging.
Namely in the case of the metrics client heartbeat.
We also decrease the logging to DEBUG within the 
AbstractClient and when loading cluster default configuration.

- A message about attempting to load the cluster defaults has
been moved to debug
- The AbstractClient now has most of its logging at DEBUG
- The client metrics now utilize the sampling logger with a
30 second cooldown period.

--- Metrics Heartbeat ---

Currently, for every Alluxio client in a JVM, the client object will
instantiate a new executor with a single thread responsible for
heartbeating metrics to the master. This is quite resource
intensive and non-ideal for this type of task.

This change aims to decrease the resource utilization of the
metrics heartbeat threads within the client. Instead of utilizing
Alluxio's heartbeat threads and a fixed executor service we
instead, utilize an instance of Java'sScheduledExecutorService
with a fixed threadpool size of 1. We laily initialize the executor
service when a new context is created, and shut it down when
the last one is destroyed. This is all controlled the public
methods within MetricsHeartbeatContext. The
FileSystemContext is responsible for adding a heartbeat and
removing it once the FSContext is closed. We also consolidate
all RPCs in the heartbeat into a single RPC that contains all the
client metrics information destined for the master.

The possible downsides to this solution:

- If there are many clients with only a single thread in the
executor, then it is possible that it will be overwhelmed and
metrics heartbeats may arrive at delayed intervals.
- Currently the implementation only supports per-JVM
granularity when reporting metrics.


ZacBlanco:
Another consideration we may want to make is bumping the default value for the client metrics heartbeat (`alluxio.user.metrics.heartbeat.interval`) from `3s` to something higher. I'm not quite as aware of standard use cases so maybe `3s` is still a sane default, but something we have the opportunity to revisit for this PR. @aaudiber @calvinjia thoughts?

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -286,8 +285,7 @@ public synchronized void close() throws IOException {
       mBlockWorkerClientPool.clear();
 
       if (mMetricsMasterClient != null) {
-        ThreadUtils.shutdownAndAwaitTermination(mExecutorService,
-            mClientContext.getConf().getMs(PropertyKey.METRICS_CONTEXT_SHUTDOWN_TIMEOUT));
+        mMetricsHeartbeat.cancel(false);
```
aaudiber: what is the purpose of this line?

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -238,12 +238,11 @@ private synchronized void init(MasterInquireClient masterInquireClient) {
...
           .build());
       mClientMasterSync = new ClientMasterSync(mMetricsMasterClient, mAppId);
-      mExecutorService = Executors.newFixedThreadPool(1,
-          ThreadFactoryUtils.build("metrics-master-heartbeat-%d", true));
-      mExecutorService
-          .submit(new HeartbeatThread(HeartbeatContext.MASTER_METRICS_SYNC, mClientMasterSync,
-              (int) mClientContext.getConf().getMs(PropertyKey.USER_METRICS_HEARTBEAT_INTERVAL_MS),
-              mClientContext.getConf()));
+
+      mMetricsHeartbeat = EXECUTOR_SERVICE.scheduleWithFixedDelay(mClientMasterSync::heartbeat, 0,
```
aaudiber: Do we need a separate heartbeat per-context, or could they be aggregated so that there's only ever 1 heartbeat?

core/common/src/main/java/alluxio/AbstractClient.java
```diff
@@ -51,7 +52,8 @@
  */
 @ThreadSafe
 public abstract class AbstractClient implements Client {
-  private static final Logger LOG = LoggerFactory.getLogger(AbstractClient.class);
+  private static final Logger LOG =
```
aaudiber: This will impact all clients, not just metrics

ZacBlanco: I am aware

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: By calling this we're removing it from the scheduled context.

Imagine there are 10 clients, there will be 10 tasks scheduled in `EXECUTOR_SERVICE`. They will run based on the schedule in the configuration. When `close()` is called on one of a client object we don't want to heartbeat more. `mMetricsHeartbeat.cancel` will remove the scheduled task for that client object, leaving the `EXECUTOR_SERVICE` responsible for only the 9 remaining client objects.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: It's possible we could aggregate them, but will likely require some more changes to the RPC. We would also not be able to ever `stop` this thread from running after the JVM starts

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: See the example in the javadoc here for a possibly better explanation of how `ScheduledFuture.cancel` works: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html

core/common/src/main/java/alluxio/AbstractClient.java
aaudiber: In particular, if metrics master client fails to connect to master and logs some warnings, other clients, e.g. file system master client, job master client, etc, will suppress their logs for 10 seconds. So users may only see metrics master client log messages, even though other clients are failing to connect as well.

In general, I don't think we want to use sampling loggers at a static scope because it makes it easy to suppress log messages that we actually want.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
aaudiber: If we have 50 of these heartbeat threads, all 50 are sending the same exact set of metrics every 3 seconds. It doesn't really matter whether it's being done from 1 thread or 50 threads, either way it's wasteful and putting unnecessary load on the master

aaudiber: If we count the number of open contexts, we can stop the heartbeat thread when the last context closes

core/common/src/main/java/alluxio/AbstractClient.java
ZacBlanco: I disagree. I feel that the amount of logging that the AbstractClient does is very high. I understand that using a SamplingLogger will cause contention of log messages by different clients but I feel that the amount of logging that is produced from this class is too high. Take for example the following outputs before and after attempting to open a file over a 10-second period.

```log
2019-04-30 12:08:02,933 WARN  alluxio.AbstractClient (SamplingLogger.java:warn) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:08:13,112 WARN  alluxio.AbstractClient (SamplingLogger.java:warn) - Failed to connect (13) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
```

Without the `SamplingLogger` I get 

```log
2019-04-30 12:05:22,454 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:22,520 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (2) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:22,637 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (3) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:22,848 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (4) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:23,280 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (5) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:24,114 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (6) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:25,519 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:25,867 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (7) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:28,525 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:28,910 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (8) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:31,534 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:32,132 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (9) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:34,542 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:35,362 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (10) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:37,549 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:38,604 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (11) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:40,557 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (1) with MetricsMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
2019-04-30 12:05:41,792 WARN  alluxio.AbstractClient (AbstractClient.java:connect) - Failed to connect (12) with FileSystemMasterClient @ zac-alluxio/10.0.0.8:19998: Failed to handshake with master zac-alluxio/10.0.0.8:19998 to load cluster default configuration values: UNAVAILABLE: io exception
```

Most of this is just the same message for multiple different clients for the same reason. Do you feel that this kind of output is still useful?

aaudiber: I'm totally in agreement that the logging is too much.

For the example you give, I see only two clients, metrics heartbeat and filesystem master. In this situation, I'd want to see two warnings, one for metrics and one for filesystem master. Do you think one warnings per client is too verbose?

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: Well it's not that they are sending the same set of metrics. Even if the object is static, they technically will report different app ID's with each metric heartbeat. If we want to continue to support application client-level granularity I'd need to modify the RPC.

On top of this: another issue I've found is that `MetricsSystem` is a static class and is only instantiated on a per-JVM basis. This leads to an issue where multiple clients (`FileSystemContexts`) in the same JVM  will end up reporting metrics for other clients in the same JVM. Until we remove the singleton `MetricsSystem` the functionality for client-object level granularity will be broken.

core/common/src/main/java/alluxio/AbstractClient.java
ZacBlanco: I think for the number of separate clients that talk to the master within the JVM, that together and with retries they perform too much logging. Preferably each type of client would have more fine-grained control over when they log messages. Unfortunately I don't think we'll easily be able get that kind of granularity because we don't propagate any errors from AbstractClient back to FileSystemMasterClient, MetricsMasterClient, etc..

What I've seen is that if a client fails to connect, usually all the other clients fail with the exact same reason. I can't recall ever seeing a case where clients failed for different reasons at the same time. I was aware that adding the SamplingLogger would cause us to lose logs, but due to the liberal nature that we log within the AbstractClient I felt it warranted the use. Maybe you've seen other cases where errors are different- but I haven't

Otherwise we could possibly move the `WARN` levels to `DEBUG`, and then `ERROR` or `WARN` when the retryPolicy reaches certain points (multiples of 15 or 20 maybe?)

aaudiber: How about we put all `AbstractClient` retry logging at debug level? If it continues to fail, we will eventually throw an exception, which should be handled or logged by whatever called the RPC.

If we **really** want to put non-debug error logging in `AbstractClient`, we could put it after the final retry fails, right before throwing an exception. But I think this is unnecessary since the place that catches the exception will log it anyway.

ZacBlanco: I have moved everything to `DEBUG` logging

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: I've completely overhauled the implementation for adding heartbeats. Please see 0edab3c

core/common/src/main/java/alluxio/AbstractClient.java
calvinjia: Will using the sampling logger lead to confusing output since we log things like the attempt count?

ZacBlanco: No. My bad, I missed removing the sampling logger. Will be fixed soon

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +38,75 @@
...
-      mMasterClient.disconnect();
-    }
+  /**
+   * Adds a new application Id to be included in the metrics heartbeat.
+   *
+   * @param appId the application identifier to be added
+   */
+  public void addAppId(String appId) {
+    Preconditions.checkNotNull(appId);
+    Preconditions.checkArgument(!appId.equals(""));
```
calvinjia: `appId.isEmpty`?

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +38,75 @@
...
-public final class ClientMasterSync implements HeartbeatExecutor {
-  private static final Logger LOG = LoggerFactory.getLogger(ClientMasterSync.class);
+public final class ClientMasterSync {
+  private static final Logger LOG =
+      new SamplingLogger(LoggerFactory.getLogger(ClientMasterSync.class), 30 * Constants.SECOND_MS);
 
   /** Client for communicating to metrics master. */
   private final MetricsMasterClient mMasterClient;
-  private final String mAppId;
+  private final ConcurrentHashSet<String> mApplicationId;
```
calvinjia: `mApplicationIds`?

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +38,75 @@
...
+            .setHostname(hostname)
+            .setClientId(appId)
+            .addAllMetrics(metrics)
+            .build());
+      }
+      try {
+        mMasterClient.heartbeat(clientMetrics);
+      } catch (IOException e) {
+        // WARN instead of ERROR as metrics are not critical to the application function
+        LOG.warn("Failed to send metrics to master: ", e);
```
calvinjia: We prefer to avoid entire stack traces for `warn` level logging, would `master: {}", e` suffice?

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +38,75 @@
...
+        mMasterClient.disconnect();
```
calvinjia: Move the disconnect logic to the call to `heartbeat`?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -0,0 +1,241 @@
...
+      MasterInquireClient inquireClient, String appId) {
+    Preconditions.checkNotNull(ctx);
+    Preconditions.checkNotNull(inquireClient);
+    Preconditions.checkNotNull(appId);
+    if (sExecutorService == null) {
+      sExecutorService = Executors.newSingleThreadScheduledExecutor(
+          ThreadFactoryUtils.build("metrics-master-heartbeat-%d", true));
+
+      // register the shutdown hook if it hasn't been set up already
+      if (!sAddedShudownHook) {
```
calvinjia: Should this be in its own condition instead of in the `sExecutorService == null` branch?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -0,0 +1,241 @@
...
+    mConf = ctx.getConf();
+    // setup metrics master client sync
+    mMetricsMasterClient = new MetricsMasterClient(MasterClientContext
+        .newBuilder(ctx)
+        .setMasterInquireClient(inquireClient)
+        .build());
+    mClientMasterSync = new ClientMasterSync(mMetricsMasterClient, mConf);
+  }
+
+  private synchronized void addContext(String appId) {
```
calvinjia: Why is this called `addContext`?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -0,0 +1,241 @@
...
+  private static boolean sAddedShudownHook = false;
+
+  /** The service which executes metrics heartbeats. */
+  private static ScheduledExecutorService sExecutorService;
+
+  private final String mMasterAddr;
+  private final MetricsMasterClient mMetricsMasterClient;
+  private final ClientMasterSync mClientMasterSync;
+  private ScheduledFuture<?> mMetricsMasterHeartbeatTask;
+  private final AtomicInteger mCtxCount;
```
calvinjia: Does this need to be an `AtomicInteger`? All accesses are synchronized (aside from the constructor)?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+      });
+      mLastHeartbeatThread.setDaemon(true);
+    }
+
+    @Override
+    public void run() {
+      mLastHeartbeatThread.start();
+      try {
+        // Shutdown hooks should run quickly, so we limit the wait time to 500ms. It isn't the end
```
calvinjia: This comment is no longer consistent with the code?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
ZacBlanco: It was called addContext because it's called when a FileSystemContext adds a heartbeat

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
ZacBlanco: It should only ever execute for one time in the life of the JVM, being within the `sExecutorService == null` branches just saves a condition check every time we call `addHeartbeat`. We're guaranteed on the first call to `addHeartbeat` that `sExecutorService` is `null`, so that check will always happen at least once.

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
ZacBlanco: In a previous implementation this wasn't the case, but I could change this to a normal integer.

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
ZacBlanco: You may have been looking at an older version of the PR. It is consistent now

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
ZacBlanco: Will update

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
ZacBlanco: This comment no longer applies

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
ZacBlanco: This comment no longer applies

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
ZacBlanco: Sorry I don't quite understand what you mean

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
ZacBlanco: Actually, I changed the log level to error in a later commit. Should I keep it at warn?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
calvinjia: It is a bit confusing because we take an `appId` as the argument, how about something like `registerApplication` or `addApplication`?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
calvinjia: Right, I don't think there's anything wrong with the logic. From a code legibility perspective it is not obvious why these two conditions are coupled, and if its just a condition check we are saving, it seems more worthwhile to make the logic decoupled.

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
calvinjia: The `MetricsMasterClient` is responsible for connecting on each `heartbeat` call, but it does not clean up (disconnect) on any network errors and instead relies on the caller to explicitly disconnect (like what you are doing here). The `heartbeat` call should make the decision on when to disconnect before throwing an exception (since some exceptions could be unrelated to the connection).
I looked at some of the other clients and we don't have a strict convention (most others use `retryRPC` so this problem is avoided).

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
ZacBlanco: Technically this is outdated now as it doesn't even take an appId anymore. @aaudiber and I decided it would be best to only allow 1 appId per-JVM because MetricsSystem is still `static`. So now it is no-arg.

As of now, its only function is to track `FileSystemContext`s through `addContext` or `removeContext`. I would like to keep it the way it is for now, and we may change it in the future.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -284,19 +241,15 @@ public synchronized void close() throws IOException {
...
       }
       mBlockWorkerClientPool.clear();
-
-      if (mMetricsMasterClient != null) {
-        ThreadUtils.shutdownAndAwaitTermination(mExecutorService,
-            mClientContext.getConf().getMs(PropertyKey.METRICS_CONTEXT_SHUTDOWN_TIMEOUT));
-        mMetricsMasterClient.close();
-        mMetricsMasterClient = null;
-        mClientMasterSync = null;
+      if (mClientContext.getConf().getBoolean(PropertyKey.USER_METRICS_COLLECTION_ENABLED)) {
```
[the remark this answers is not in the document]
ZacBlanco: Essentially, yes - technically, no. Upon creation, it is likely that the cluster defaults have not been loaded. Pulling in the cluster defaults could change this property. Other than that change, it is immutable.

I think I would prefer to save it to a variable, and read that value. In this case, cluster metrics can only be enabled by client-side configuration, not cluster defaults. Is that behavior we would like to have?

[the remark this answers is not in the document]
aaudiber: I think it would be quite valuable to be able to enable/disable cluster defaults from the master. Could we make sure cluster defaults have been downloaded before checking the property?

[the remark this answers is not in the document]
ZacBlanco: It wouldn't really fit with the design of the current implementation because we load cluster defaults lazily upon first connection. Typically, that means when the user application opens or reads a file. @cheng-chang was also having issues with cluster defaults being loaded lazily as it was affecting path configuration. We may be changing it in the near future.

Personally, I don't think the effort is worth changing right now. If we decide to pro-actively load cluster defaults, then supporting enabling metrics via cluster defaults should become trivial.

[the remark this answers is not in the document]
aaudiber: ok, we can let the dust settle on Cheng's changes before putting work in here

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +37,54 @@
  * it before retrying.
  */
 @ThreadSafe
-public final class ClientMasterSync implements HeartbeatExecutor {
-  private static final Logger LOG = LoggerFactory.getLogger(ClientMasterSync.class);
+public final class ClientMasterSync {
+  private static final Logger LOG =
```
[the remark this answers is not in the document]
ZacBlanco: fixed

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +37,54 @@
...
 
-  @Override
-  public synchronized void heartbeat() throws InterruptedException {
+  /**
+   * Sends metrics to the master keyed with appId and client hostname.
+   */
+  public synchronized void heartbeat() {
+    // Only perform the heartbeat if there are applications to track
+    List<alluxio.grpc.ClientMetrics> clientMetrics = new ArrayList<>();
+    List<Metric> allClientMetrics = MetricsSystem.allClientMetrics();
```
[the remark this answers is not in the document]
ZacBlanco: updated

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -33,41 +37,55 @@
...
-public final class ClientMasterSync implements HeartbeatExecutor {
-  private static final Logger LOG = LoggerFactory.getLogger(ClientMasterSync.class);
+public final class ClientMasterSync {
+  private static final Logger SAMPLING_LOG =
+      new SamplingLogger(LoggerFactory.getLogger(ClientMasterSync.class), 30 * Constants.SECOND_MS);
 
-  /** Client for communicating to metrics master. */
+  /**
+   * Client for communicating to metrics master.
+   */
```
calvinjia: Could you change this to `/** Comment */` format?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -0,0 +1,261 @@
...
+  private final MetricsMasterClient mMetricsMasterClient;
+  private final ClientMasterSync mClientMasterSync;
+  private final AlluxioConfiguration mConf;
+
+  // This can only be a primitive if all accesses are synchronized
+  private int mCtxCount;
+  private ScheduledFuture<?> mMetricsMasterHeartbeatTask;
+
+  private MetricsHeartbeatContext(ClientContext ctx,
+      MasterInquireClient inquireClient) {
```
calvinjia: One line?

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
ZacBlanco: done

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
ZacBlanco: done

calvinjia:

Issue #8571, opened 2019-03-15
Multi-threaded backup


aaudiber:
**Is your feature request related to a problem? Please describe.**
Alluxio can back up around 3 million files per minute. When the master contains hundreds of millions of files, backups can take hours. This is largely because a single thread does all the backup work.

**Describe the solution you'd like**
We could use multiple threads (e.g. equal to number of processors on the machine) to speed up the backup process. One thread writes serialized inodes to the backup while other threads walk the inode tree and feed the writer thread serialized inodes to write.

**Describe alternatives you've considered**
Right now we write directly to a single backup file. To improve performance further, we could write to multiple files within a directory, and then archive them at the end. Then we won't be bottlenecked on a single writer thread. This is a bigger change since it requires modifying the backup file format.

gpang:
Can we consider taking a journal backup from a standby? We could briefly "stop" the leader master to write out a journal entry to all the masters, and then the assigned standby could pause all the journals at that point in time, to run the full backup.

aaudiber:
I like the idea of enlisting the help of standby masters to avoid leader downtime. With a secondary taking the backup, there shouldn't be any need at all to stop the leader. We could use either RPC or special journal entry to communicate a standby that it needs to take a backup. RPC might be better so that we can respond to the CLI synchronously, and so that we can control which standby tries to take the backup. 

Some other thoughts:
- Standbys only see completed logs, so we need to complete the current log as a part of running `backup`, and also tell the standby what sequence number it should catch itself up to.
- To extend this approach to single-master-node deployments, we could share the backup functionality across both standbys and dedicated `SecondaryMaster` processes.

gpang:
The potential issue I see is if the standby cannot replay the journal entries for some reason.

But, in general, offloading the backup from the primary to the standby avoids the pause that is directly proportional to the metadata size.

aaudiber:
Created https://github.com/Alluxio/alluxio/issues/8935 for the idea of using standbys to take snapshots

PR #8942, opened 2019-04-30
Do multi-threaded backups

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/8571

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/BackupManager.java
+++ b/core/server/common/src/main/java/alluxio/master/BackupManager.java
@@ -54,7 +54,7 @@ import java.util.regex.Pattern;
 public class BackupManager {
   private static final Logger LOG = LoggerFactory.getLogger(BackupManager.class);
   // Used to mark termination of reading/writing journal entries.
-  private final long mTerminationSequence = -1;
+  private static final long TERMINATION_SEQ = -1;
 
   /**
    * The name format of the backup files (E.g., alluxio-backup-2018-12-14-1544749081561.gz).
@@ -114,7 +114,7 @@ public class BackupManager {
         }
         // Put termination entry for signaling the writer.
         journalEntryQueue
-            .put(JournalEntry.newBuilder().setSequenceNumber(mTerminationSequence).build());
+            .put(JournalEntry.newBuilder().setSequenceNumber(TERMINATION_SEQ).build());
         return true;
       } catch (InterruptedException ie) {
         Thread.currentThread().interrupt();
@@ -138,7 +138,7 @@ public class BackupManager {
           // Write entries to back-up stream.
           for (JournalEntry journalEntry : pendingEntries) {
             // Check for termination entry.
-            if (journalEntry.getSequenceNumber() == mTerminationSequence) {
+            if (journalEntry.getSequenceNumber() == TERMINATION_SEQ) {
               // Reading finished.
               return true;
             }
@@ -202,7 +202,7 @@ public class BackupManager {
           }
           // Put termination entry for signaling the applier.
           journalEntryQueue
-              .put(JournalEntry.newBuilder().setSequenceNumber(mTerminationSequence).build());
+              .put(JournalEntry.newBuilder().setSequenceNumber(TERMINATION_SEQ).build());
           return true;
         } catch (InterruptedException ie) {
           // Continue interrupt chain.
@@ -243,7 +243,7 @@ public class BackupManager {
               // Apply entries.
               for (JournalEntry entry : drainedEntries) {
                 // Check for termination entry.
-                if (entry.getSequenceNumber() == mTerminationSequence) {
+                if (entry.getSequenceNumber() == TERMINATION_SEQ) {
                   // Reading finished.
                   return true;
                 }
--- a/core/server/common/src/main/java/alluxio/master/BackupManager.java
+++ b/core/server/common/src/main/java/alluxio/master/BackupManager.java
@@ -11,13 +11,16 @@
 
 package alluxio.master;
 
+import alluxio.conf.PropertyKey;
+import alluxio.conf.ServerConfiguration;
 import alluxio.master.journal.JournalContext;
 import alluxio.master.journal.JournalEntryAssociation;
 import alluxio.master.journal.JournalEntryStreamReader;
-import alluxio.master.journal.JournalUtils;
 import alluxio.proto.journal.Journal.JournalEntry;
+import alluxio.util.ThreadFactoryUtils;
 
 import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.SettableFuture;
 import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
 import org.apache.commons.compress.compressors.gzip.GzipCompressorOutputStream;
 import org.slf4j.Logger;
@@ -26,9 +29,17 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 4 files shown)


ggezer:
@aaudiber , @LuQQiu PTAL.

aaudiber:
@ggezer What speedup do you get with these changes?

ggezer:
@aaudiber,
With 5M files and an UFS journal.
Time for applying a backup, went down from 5mins to 50secs.
Time for taking a backup didn't reduce much(55 secs to 45 secs). I'm still investigating that.

aaudiber:
@ggezer I think the problem is that the backup time is dominated by the time to back up the filesystem master, so parallelizing across masters doesn't do too much. We need to parallelize the filesystem master backup

ggezer:
@aaudiber That's very likely. I get same readings for taking backup regardless of where I persist the backup.

Do you think master level concurrency while reading/applying will be useful? If FSM will always dominate, then I may consider reducing this change to single thread for both readers/writers in take/apply backup paths.

Do you think parallelizing journaled components of FSM would suffice? Or FSM is also dominated by its`InodeTree` component?

aaudiber:
I think 90% of the gains are in parallelization of the `InodeTree` component. Other FSM components are much much smaller. The next biggest is block master, but that still less than 10% the size of the inode tree

ggezer:
@aaudiber Doing a simple read-ahead buffering in `InodeTree` during enumeration brought backup time for 5M files from 48secs to 12secs. I feel like this is good enough gain for now. WDYT?

Also, what's your take on eliminating master-level concurrency on `BackupManager`?

aaudiber:
@ggezer I'm hoping we can take advantage of all cores since usually production masters are quite powerful, with 24+ cores. Can the read-ahead buffering scale to that?

Another thing to consider is that the metadata reading speed will drop significantly in off-heap scenarios, e.g. 1B files with 95% of metadata on disk. I'm hoping that with enough parallelism and a powerful machine, we can take a 1B files backup in under an hour.

I'm in favor of eliminating master-level concurrency since it isn't buying us much

ggezer:
@aaudiber Parallel enumeration of tree will need some algorithmic design. The fundemental challenge for a crawler thread is to determine when to keep branching and when to offload a child to another thread. We can simplify the design with constraints inherent to file systems but I believe we should tackle that as a separate work.

ggezer:
@aaudiber Updated the PR based on our discussions. 
* Simplified buffering in `BackupManager`
* Multi-threaded iteration of `InodeTree`

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -83,27 +215,130 @@ public void backup(OutputStream os) throws IOException {
...
+      backupReaderExecutor.submit(() -> {
         try {
-          master.processJournalEntry(entry);
-        } catch (Throwable t) {
-          JournalUtils.handleJournalReplayFailure(
-              LOG, t, "Failed to process journal entry %s when init from backup", entry);
+          JournalEntry entry;
+          while (!applierTaskFuture.isDone() && (entry = reader.readEntry()) != null) {
+            // Try to push the entry as long as applier is alive.
+            while (!applierTaskFuture.isDone()) {
```
aaudiber: is this needed when it's also checked in the outer while loop?

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -83,27 +215,130 @@ public void backup(OutputStream os) throws IOException {
...
+        Throwable cause = ee.getCause();
+        if (cause instanceof IOException) {
+          throw (IOException) cause;
+        } else {
+          throw new IOException(cause);
+        }
+      } finally {
+        // Close journal contexts to ensure applied entries are flushed.
+        for (JournalContext journalContext : masterJCMap.values()) {
+          journalContext.close();
```
aaudiber: will this work if the metadata size is large? We can't afford to buffer the full backup in memory. I think we might need to write it in chunks instead

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -677,35 +685,195 @@ public void restoreFromCheckpoint(CheckpointInputStream input) throws IOExceptio
...
+    final class BufferedIterator implements Iterator<Journal.JournalEntry> {
+
+      /** Buffered entry queue. */
+      BlockingQueue<JournalEntry> mEntryBuffer;
+      /** Iterator thread completion queue. */
+      BlockingQueue<Boolean> mCompletionQueue;
+      /** How many threads to use for enumeration. */
+      private int mIteratorThreadCount;
+      /** How many entry to pre-fetch during enumeration. */
+      private int mEntryBufferSize;
```
aaudiber: can this and `mIteratorThreadCount` be local variables instead of fields?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -677,35 +685,195 @@ public void restoreFromCheckpoint(CheckpointInputStream input) throws IOExceptio
...
+            // Crawling completed.
+            mCrawlerExecutor.shutdown();
+          }
+        });
+        // Will be shutdown after current coordinator task is complete.
+        mCoordinatorExecutor.shutdown();
+      }
+
       @Override
       public boolean hasNext() {
```
aaudiber: Can we make this blocking so that `hasNext()` is true iff there really is a next element? It will make the iterator much more pleasant to work with

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -63,18 +74,139 @@ public BackupManager(MasterRegistry registry) {
...
+    SettableFuture<Void> writerTaskFuture = SettableFuture.create();
+
+    // Submit master reader task.
+    backupMasterReaderExecutor.submit(() -> {
+      try {
+        for (Master master : mRegistry.getServers()) {
+          Iterator<JournalEntry> it = master.getJournalEntryIterator();
+          if (it != null) {
+            while (!writerTaskFuture.isDone() && it.hasNext()) {
+              JournalEntry je = it.next();
```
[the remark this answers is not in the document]
ggezer: Any attempt to guarantee a deterministic answer for hasNext() will need a synchronization with buffering process. Since hasNext() is called as the same amount as next(), any synchronization will cause a contention. I don't see a way without badly hurting the concurrency of buffering.

core/server/common/src/main/java/alluxio/master/BackupManager.java
ggezer: offer() obeys the limits. if an applier is failed, that offer can stuck there forever. That's there is a timeout and checking if if offering a new entry still makes sense. In short, first one is for iterating the entries from backup, the second is for pushing an individual one to buffer.

core/server/common/src/main/java/alluxio/master/BackupManager.java
ggezer: Journal writer has a dedicated thread that continuously writes. So it'll never be all of the entries of backup. We can have intermittent flushes here at each drain time, if you think it's still necessary with dedicated thread on journal flushing side.

core/server/common/src/main/java/alluxio/master/BackupManager.java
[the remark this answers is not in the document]
aaudiber: I'm thinking of pushing the "it might be null" handling code into the iterator instead of here. The synchronization shouldn't need to change

core/server/common/src/main/java/alluxio/master/BackupManager.java
aaudiber: got it, makes sense

core/server/common/src/main/java/alluxio/master/BackupManager.java
aaudiber: I see, it will only be an issue if the tree crawling threads produce inodes faster than the writer thread can write them to the backup.

Perhaps `JournalContext` is the wrong tool to use here. It is designed to support many threads writing to the journal concurrently, but in this situation we have only a single thread. Perhaps we could replace it with a simple bounded queue (to prevent unlimited memory usage) plus a regular `JournalWriter` instead of `AsyncJournalWriter`?

core/server/common/src/main/java/alluxio/master/BackupManager.java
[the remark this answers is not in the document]
ggezer: I think it could done by an additional buffering of next entry and moving blocking part from next() to hasNext().

core/server/common/src/main/java/alluxio/master/BackupManager.java
ggezer: I think doing an intermittent flushes per each drained entry list is acceptable enough. For 5M files I didn't see any perf change. For bigger loads, it shouldn't be any worse. The fix is to open and close contexts per each round of draining from queue. This will make sure the amount of memory we use is proportional to the configured buffer size.

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -63,18 +80,87 @@ public BackupManager(MasterRegistry registry) {
    * @param os the stream to write to
    */
   public void backup(OutputStream os) throws IOException {
-    int count = 0;
+    // Create gZIP compressed stream as back-up stream.
```
aaudiber: This method is getting pretty long. I think it could be cleaner to move this logic to a new stateful `BackupWriter`/`BackupReader` classes which handle a single calls to backup/restore. The tasks could be modeled with inner classes instead of anonymous functions. WDYT?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -677,35 +692,250 @@ public void restoreFromCheckpoint(CheckpointInputStream input) throws IOExceptio
...
-        return !inodes.isEmpty();
+        /**
```
aaudiber: nit: since it's inside a function, this should be regular comment `/*`

core/server/common/src/main/java/alluxio/master/BackupManager.java
ggezer: Since both taking a backup and applying it revolve around a blocking queue and both use draining, I think there could be a shared base class that provides reader/writer infra. So then, splitted classes could be differing only on the actual operations they're doing. I would like to put a TODO and work on it later if that's fine.

aaudiber: SG

aaudiber:

PR #8944, opened 2019-04-30
Fix the eventual consistency issue in S3 contract test

LuQQiu:
the `deleteExistingDirectory` is able to solve the create-then-delete eventual consistency issue, but the following `isFile` and `isDirectory` check does not support eventual consistency since they are the S3 current status check.


madanadit:

PR #8946, opened 2019-05-01
Add command to trigger a snapshot in primary master journal system

LuQQiu:
Add a SnapshotCommand to trigger a snapshot/checkpoint in primary master journal system.
This command is added for debugging or avoid primary master log from going unbounded.


LuQQiu:
jenkins, test this please

LuQQiu:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +328,63 @@ public synchronized void losePrimacy() {
...
+      try {
+        future.get(5, TimeUnit.SECONDS);
+      } catch (TimeoutException | ExecutionException e) {
+        LOG.info("Exception submitting entry: {}", e.toString());
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        return builder.setTriggered(false)
+            .setMessage("Interrupted while submitting entry to trigger snapshot").build();
+      }
+      if (!mStateMachine.isSnapshotting()) {
```
aaudiber: could the snapshot have already finished?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +328,63 @@ public synchronized void losePrimacy() {
...
+   * @param stateMachine the journal state machine
+   * @param builder the snapshot response builder
+   */
+  private void waitForSnapshotting(JournalStateMachine stateMachine,
+      SnapshotPResponse.Builder builder) {
+    try {
+      CommonUtils.waitFor("snapshotting to finish", () -> !stateMachine.isSnapshotting(),
+          WaitForOptions.defaults().setTimeoutMs(20 * Constants.MINUTE_MS));
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
aaudiber: when would we hit this code?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +328,63 @@ public synchronized void losePrimacy() {
...
+            .setMessage("Do not fulfill Raft journal system snapshot requirement").build();
+      }
+
+      waitForSnapshotting(mStateMachine, builder);
+      if (builder.hasSucceed()) {
+        return builder.build();
+      }
+      return builder.setTriggered(true).setSucceed(true).build();
+    } finally {
+      if (mPrimarySelector.getState() == PrimarySelector.State.PRIMARY) {
```
aaudiber: this method is synchronized with `gainPrimacy` and `losePrimacy`, so I think it's safe to always set `mSnapshotAllowed` to `false` at the end. If the primary selector switched to secondary while the snapshot was happening, we will call `losePrimacy` afterward, which will set `mSnapshotAllowed` to `true`

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -311,6 +312,19 @@ public void format() throws IOException {
         .toString());
   }
 
+  /**
+   * Triggers a checkpoint in the primary master ufs journal.
+   *
+   * @return the snapshot response
+   */
+  public synchronized SnapshotPResponse checkpoint() {
+    if (mTailerThread == null) {
```
aaudiber: isn't this normally `null` on the primary?

core/transport/src/grpc/meta_master.proto
```diff
@@ -112,6 +112,20 @@ message GetMetricsPResponse {
...
+/**
+ * The snapshot response has three cases:
+ * 1. failed to trigger (triggered=false and message explains why snapshot is not triggered)
+ * 2. triggered but failed (triggered=false, succeed=false and message explains why snapshot is failed)
+ * 3. succeed (triggered=true, succeed=true)
+ */
+message SnapshotPResponse {
+ optional bool triggered = 1;
+ optional bool succeed = 2;
+ optional string message = 3;
```
aaudiber: usually we use exceptions for this sort of thing. If the rpc fails to trigger a snapshot, we throw an exception saying why it failed to trigger. If the rpc starts a snapshot but the snapshot fails, we throw an exception saying why the snapshot failed. If we don't throw anything, that means it succeeded.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: I tested on the alluxio cluster with fsmeta operations, the snapshot always fall into this if block. In this way, we are not exactly sure if `statemachine` snapshots. I will come up with other ways to make sure snapshot happens. You are right, the snapshot could have already finished.

core/transport/src/grpc/meta_master.proto
LuQQiu: Yeah, will use exceptions instead of message here to indication the rpc fails.
But Snapshot has one more circumstances other than snapshotted or failed: the snapshot is not triggered. 

Is it OK that we only responsible for giving the possibility of taking a snapshot, but not guarantee snapshotted in the Snapshot CLI? Giving console message like: successfully send the snapshot signal.

aaudiber: I think we can throw an exception with a relevant message if we fail to checkpoint because the journal doesn't want to be checkpointed, e.g. Copycat decides not to snapshot, or in the UFS journal no entries have been written since the last checkpoint. I don't see a need for the client side to differentiate between "Failed because the journal refused" and "Failed for some other reason". Either way I think we'll just forward the error message to the client

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterClient.java
```diff
@@ -56,4 +57,11 @@
    * @return a map of metrics information
    */
   Map<String, MetricValue> getMetrics() throws AlluxioStatusException;
+
+  /**
+   * Triggers a snapshot of the primary master state in master journal system.
+   *
+   * @return the snapshot response
+   */
+  SnapshotPResponse snapshot() throws IOException;
```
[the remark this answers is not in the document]
LuQQiu: Change the name from snapshot to checkpoint. Thanks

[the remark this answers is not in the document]
LuQQiu: Now i only return the hostname of the master that trigger the checkpoint.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +328,63 @@ public synchronized void losePrimacy() {
     LOG.info("Raft server successfully restarted");
   }
 
+  @Override
+  public synchronized SnapshotPResponse snapshot() {
+    SnapshotPResponse.Builder builder = SnapshotPResponse.newBuilder();
+    mSnapshotAllowed.set(true);
+    try {
+      CopycatClient client = createClient();
+      long sequence = ThreadLocalRandom.current().nextLong(Long.MIN_VALUE, 0);
```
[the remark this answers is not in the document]
LuQQiu: Changed to use `JournalEntry.getDefaultInstance()`, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +328,63 @@ public synchronized void losePrimacy() {
...
+      CompletableFuture<Void> future = client.submit(new JournalEntryCommand(
+          JournalEntry.newBuilder().setSequenceNumber(sequence).build()));
+      try {
+        future.get(5, TimeUnit.SECONDS);
+      } catch (TimeoutException | ExecutionException e) {
+        LOG.info("Exception submitting entry: {}", e.toString());
```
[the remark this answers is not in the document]
LuQQiu: Also throw exception here.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Add a `mLastSnapshotStartTime` to trigger when the last snapshot start. If it is bigger than the `RaftJournalSystem.checkpoint()` start time, we successfully trigger the snapshot.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: The `CommonUtils.waitFor()` method requires handling of the `InterruptedException`, we may not be able to hit this code.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Make sense, remove the Primary selector state check

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
LuQQiu: Oh yeah, now change to directly `mMaster.writeToCheckpoint(journalWriter);` now.

core/transport/src/grpc/meta_master.proto
LuQQiu: Make sense. Now I will also throw exceptions when not be able to trigger snapshot. Only return results when snapshot is triggered and succeed.

LuQQiu:
@aaudiber Addressed comments and tested in Alluxio cluster with embedded journal and ufs journal. Now is working on adding tests. PTAL, thanks

core/server/common/src/main/java/alluxio/master/journal/noop/NoopJournalSystem.java
```diff
@@ -57,4 +57,9 @@ public void start() {}
 
   @Override
   public void stop() {}
+
+  @Override
+  public void checkpoint() {
+    throw new UnsupportedOperationException("Checkpoint is not supported in NoopJournalSystem");
```
aaudiber: more consistent to make this a no-op like other methods

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,58 @@ public synchronized void losePrimacy() {
...
+      long start = System.currentTimeMillis();
+      LOG.info("Submitting empty journal entry to trigger snapshot");
+      // If snapshot requirements are fulfilled, a snapshot will be triggered
+      // after sending an empty journal entry to Copycat
+      CompletableFuture<Void> future = client.submit(new JournalEntryCommand(
+          JournalEntry.getDefaultInstance()));
+      try {
+        future.get(1, TimeUnit.MINUTES);
+      } catch (TimeoutException | ExecutionException e) {
+        LOG.info("Exception submitting entry to trigger snapshot: {}", e.toString());
```
aaudiber: can this be LOG.warn?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,58 @@ public synchronized void losePrimacy() {
...
+        throw new IOException("Exception submitting entry to trigger snapshot", e);
+      } catch (InterruptedException e) {
+        LOG.info("Interrupted when submitting entry to trigger snapshot: {}", e.toString());
```
aaudiber: ditto

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,58 @@ public synchronized void losePrimacy() {
...
+        throw new CancelledException("Interrupted when submitting entry to trigger snapshot", e);
+      }
+      if (mStateMachine.getLastSnapshotStartTime() < start) {
+        throw new IOException("Do not fulfill Copycat snapshot requirements. "
+            + "No snapshot is triggered");
+      }
+      if (mStateMachine.isSnapshotting()) {
+        waitForSnapshotting(mStateMachine);
+      }
+      LOG.info("Snapshotted in raft journal system");
```
aaudiber: I think the snapshot implementation will already log about this, so we don't need a second log here

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,58 @@ public synchronized void losePrimacy() {
     LOG.info("Raft server successfully restarted");
   }
 
+  @Override
+  public synchronized void checkpoint() throws IOException {
+    mSnapshotAllowed.set(true);
+    CopycatClient client = createAndConnectClient();
```
aaudiber: this should be inside the `try` - otherwise if we throw an exception, we won't set `mSnapshotAllowed` back to `false`

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,58 @@ public synchronized void losePrimacy() {
...
+        future.get(1, TimeUnit.MINUTES);
+      } catch (TimeoutException | ExecutionException e) {
+        LOG.info("Exception submitting entry to trigger snapshot: {}", e.toString());
+        throw new IOException("Exception submitting entry to trigger snapshot", e);
+      } catch (InterruptedException e) {
+        LOG.info("Interrupted when submitting entry to trigger snapshot: {}", e.toString());
+        Thread.currentThread().interrupt();
+        throw new CancelledException("Interrupted when submitting entry to trigger snapshot", e);
+      }
+      if (mStateMachine.getLastSnapshotStartTime() < start) {
```
aaudiber: what if the snapshot is about to start, but just hasn't started yet? Perhaps we could retry for some period before giving up

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,58 @@ public synchronized void losePrimacy() {
...
+
+  /**
+   * Waits for snapshotting to finish.
+   *
+   * @param stateMachine the journal state machine
+   */
+  private void waitForSnapshotting(JournalStateMachine stateMachine) throws IOException {
+    try {
+      CommonUtils.waitFor("snapshotting to finish", () -> !stateMachine.isSnapshotting(),
+          WaitForOptions.defaults().setTimeoutMs(20 * Constants.MINUTE_MS));
```
aaudiber: unrelated to this PR, but can we make this configurable and increase the default to 2 hours? With 1 billion files this might time out due to the large scale

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -311,6 +312,31 @@ public void format() throws IOException {
...
   }
 
+  /**
+   * Triggers a checkpoint in the primary master ufs journal.
+   */
+  public synchronized void checkpoint() throws IOException {
+    long nextSequenceNumber = getNextSequenceNumberToWrite();
+    if (nextSequenceNumber - getNextSequenceNumberToCheckpoint() < ServerConfiguration.getLong(
+        PropertyKey.MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES)) {
+      LOG.info("{}: No enough journal entries to checkpoint.",
```
aaudiber: can we force a checkpoint anyway in this situation? The user is indicating that they want a checkpoint, so we should do our best to create one

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -154,4 +154,11 @@ public void format() throws IOException {
       journal.format();
     }
   }
+
+  @Override
+  public void checkpoint() throws IOException {
+    for (UfsJournal  journal : mJournals.values()) {
```
aaudiber: nit: extra space

shell/src/main/java/alluxio/cli/fsadmin/command/CheckpointCommand.java
```diff
@@ -0,0 +1,75 @@
...
+    return "checkpoint";
+  }
+
+  @Override
+  public int run(CommandLine cl) throws IOException {
+    Timer timer = new Timer();
+    timer.schedule(new TimerTask() {
+        @Override
+        public void run() {
+          mPrintStream.println("Checkpointing may take a while.");
```
aaudiber: this is pretty verbose - in other places we've logged a single "." every few seconds without adding newlines, see https://github.com/Alluxio/alluxio/blob/master/job/client/src/main/java/alluxio/client/job/JobGrpcClientUtils.java#L97

shell/src/main/java/alluxio/cli/fsadmin/command/CheckpointCommand.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }, Constants.MINUTE_MS);
+    String masterHostname = mMetaClient.checkpoint();
+    timer.cancel();
+    mPrintStream.println("Successfully checkpointed in " + masterHostname);
```
aaudiber: `Successfully took a checkpoint on master %s`

core/server/common/src/main/java/alluxio/master/journal/noop/NoopJournalSystem.java
LuQQiu: Changed, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Changed, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Changed, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Yeah, we don't need a second log here, removed, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Move inside the try finally. Thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Add the retry check start time, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Add a property key to control how long we wait for the snapshotting to finish.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
LuQQiu: Now only when the `nextSequenceNumber == getNextSequenceNumberToCheckpoint()` we don't trigger snapshot.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
LuQQiu: Fixed, thanks

shell/src/main/java/alluxio/cli/fsadmin/command/CheckpointCommand.java
LuQQiu: Used the `...`  interpreting waiting

shell/src/main/java/alluxio/cli/fsadmin/command/CheckpointCommand.java
LuQQiu: Changed, thanks

LuQQiu:
@aaudiber Look into the snapshot requirements, the main requirement that always fails to achieve is Copycat need to have new segments (I think it is a file contains many logs). The segment size is controlled by `alluxio.master.journal.log.size.bytes.max` which by default is `10MB`. When I set it to `1kb`, the snapshot will be triggered and succeed definitely.
Addressed the comments, add the triggered embedded journal snapshot integration test and wait for snapshot to start logic. PTAL, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,79 @@ public synchronized void losePrimacy() {
...
+    try {
+      CommonUtils.waitFor("snapshotting to start", () ->
+              stateMachine.getLastSnapshotStartTime() > start,
+          WaitForOptions.defaults().setTimeoutMs(10 * Constants.SECOND_MS));
+      return;
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new CancelledException("Interrupted when waiting for snapshotting to start", e);
+    } catch (TimeoutException e) {
+      throw new IOException("Do not fulfill Copycat snapshot requirements. "
```
aaudiber: this message will likely be propagated to users, who might not know about snapshots or copycat. Can we say something more general like "A recent checkpoint already exists, not creating a new one"?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -327,6 +334,79 @@ public synchronized void losePrimacy() {
...
+      } catch (TimeoutException | ExecutionException e) {
+        LOG.warn("Exception submitting entry to trigger snapshot: {}", e.toString());
+        throw new IOException("Exception submitting entry to trigger snapshot", e);
+      } catch (InterruptedException e) {
+        LOG.warn("Interrupted when submitting entry to trigger snapshot: {}", e.toString());
+        Thread.currentThread().interrupt();
+        throw new CancelledException("Interrupted when submitting entry to trigger snapshot", e);
+      }
+      waitForSnapshotStart(mStateMachine, start);
+      if (mStateMachine.isSnapshotting()) {
```
aaudiber: unnecessary check

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -311,6 +312,29 @@ public void format() throws IOException {
         .toString());
   }
 
+  /**
+   * Triggers a checkpoint in this ufs journal.
```
aaudiber: "Triggers" suggests asynchronous, I think "Creates" is a better word here

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -311,6 +312,29 @@ public void format() throws IOException {
...
+   */
+  public synchronized void checkpoint() throws IOException {
+    long nextSequenceNumber = getNextSequenceNumberToWrite();
+    if (nextSequenceNumber == getNextSequenceNumberToCheckpoint()) {
+      LOG.info("{}: No enough journal entries to create checkpoint.",
```
aaudiber: We can be more specific, saying that no entries have been written since the last checkpoint

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Yeah, changed. Now is more user-friendly.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
LuQQiu: Removed the check

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
LuQQiu: Changed the `triggers` to `creates` to sound more synchronous

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
LuQQiu: Changed, more clear. thanks

LuQQiu:
@aaudiber Thanks for correcting the documentation, Addressed the comments, PTAL, thanks

aaudiber:

PR #8948, opened 2019-05-01
Make "alluxio bootstrapConf" to create alluxio-site.properties

### LINKED ISSUE #8615: Make alluxio bootstrap-conf create and modify alluxio-site.properties
**Summary**
Make `alluxio bootstrap-conf` command to create and modify `conf/alluxio-site.properties` rather than `conf/alluxio-env.sh`. This is because `conf/alluxio-env.sh` modifies the environment variable, which we should limit the use and prefer use property files.

**Urgency**
Low

apc999:
Fix #8615 

- We shall always use `alluxio-site.properties` over `alluxio-env.sh` when possible
- MEM size logic is now inside the default properties, no need to be in shell script


apc999:

PR #8952, opened 2019-05-01
Remove the unneeded S3A consistency timeout property

LuQQiu:
UNDERFS_S3A_CONSISTENCY_TIMEOUT_MS is not used anywhere


LuQQiu:
@apc999 PTAL, remove one outdated property key

PR #8954, opened 2019-05-01
Move shaded module from a top module to a submodule inside underfs

apc999:
and rename the module to be consistent with other modules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,7 +173,6 @@
 
   <modules>
     <module>alluxio-ui</module>
-    <module>alluxio-shaded-hadoop</module>
     <module>assembly</module>
     <module>core</module>
     <module>examples</module>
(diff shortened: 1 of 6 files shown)


apc999:
@calvinjia Agree. But having it as a top level module seems an overkill from codebase organization perspective. How about having a `dep` or `shaded` top level module and move this module inside as  `hadoop`? I would imagine in the future, we may need to add more shaded module?

calvinjia:
@apc999 Yeah I think having another top level module for `shaded` would be good. The modules inside should just be shaded libraries and have no Alluxio dependencies.

apc999:
@calvinjia @yuzhu  PTAL

apc999:
Jenkins, test this please

calvinjia:

PR #8964, opened 2019-05-03
Hide credentials in mount command

bf8086:
This fix changed the API so we hides values of mount options that are credentials.

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
+++ b/core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
@@ -16,13 +16,11 @@ import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.AlluxioProperties;
 import alluxio.conf.ConfigurationValueOptions;
 import alluxio.conf.InstancedConfiguration;
-import alluxio.conf.PropertyKey;
 import alluxio.conf.Source;
 import alluxio.util.ConfigurationUtils;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 
 import javax.annotation.concurrent.NotThreadSafe;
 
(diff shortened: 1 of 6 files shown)


bf8086:
@gpang PTAL. Thanks!

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1310,28 +1310,28 @@ public long getNewBlockIdForFile(AlluxioURI path) throws FileDoesNotExistExcepti
...
     for (Map.Entry<String, MountInfo> mountPoint : mMountTable.getMountTable().entrySet()) {
-      mountPoints.put(mountPoint.getKey(), getMountPointInfo(mountPoint.getValue()));
+      mountPoints.put(mountPoint.getKey(), getDisplayMountPointInfo(mountPoint.getValue()));
     }
     return mountPoints;
   }
 
   @Override
-  public MountPointInfo getMountPointInfo(AlluxioURI path) throws InvalidPathException {
+  public MountPointInfo getDisplayMountPointInfo(AlluxioURI path) throws InvalidPathException {
```
gpang: Which is the API which the workers use to get UFS mount point properties?

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -125,4 +128,18 @@ public UnderFileSystemConfiguration createMountSpecificConf(Map<String, String>
...
+
+  /**
+   * @param userPropertiesOnly  whether to use only the user properties
+   * @param options options for formatting the configuration values
+   * @return a map from all configuration property names to their values; values may potentially be
+   *         null
+   */
+  public Map<String, String> toMap(boolean userPropertiesOnly, ConfigurationValueOptions options) {
+    Map<String, String> map = new HashMap<>();
+    Set<PropertyKey> keys = userPropertiesOnly ? userKeySet() : keySet();
```
gpang: What are user properties? Are UFS properties user properties? If so, this parameter `userPropertiesOnly` should always be `true`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: `alluxio.master.file.DefaultFileSystemMaster#getUfsInfo`

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
bf8086: I refactored the method to make it clear.

bf8086:
Thanks @gpang! I replied to the comments and added a test. 

I am not aware of other places we leak credentials. Do you have any idea where to look at?

gpang:
@bf8086 I am not aware of other leakage, but it is unclear what else uses mount options. Is there a way to make the credentials hidden by default, and have a separate mechanism for getting access to the real values? I'd imagine only a few places really need the credentials (UFS creation, worker)?

I'm wondering what we can do to prevent leaking the values in the future.

bf8086:
@gpang At API level there are 3 APIs expose UFS options: `getMountTable`, `get(Display)MountPointInfo` and `getUfsInfo`. Only `getMountTable` is exposed in client RPC and is currently used by mount command. `get(Display)MountPointInfo` is used by web UI. Both APIs are hiding credentials. `getUfsInfo` is the only API gives the actual credentials but it is only exposed in master-worker RPC interface.

For a more secure solution, we will need to refactor how we store credentials internally and never store it unencrypted. That will help preventing the future leaks. I would still like to get this change in first so we don't have to block 2.0 release on this security concern. I can create a tracking issue for the refactor work.

gpang:

PR #8971, opened 2019-05-05
[SMALLFIX] modify something 

1072158516:
Use the @Nullable annotation for all methods which may return null in /alluxio/core/server/common/src/main/java/alluxio/cli/validation/HadoopConfigurationFileParser.java#parseXmlConfiguration #492


1072158516:
modify the title

1072158516:
modify it again

1072158516:
lll

PR #8973, opened 2019-05-05

hsc0902:
[SMALLFIX] Use the @Nullable annotation for all methods which may return null in /alluxio/core/server/worker/src/main/java/alluxio/worker/block/allocator/RoundRobinAllocator.java#allocateBlock #499


apc999:

PR #8990, opened 2019-05-06

courseralxy:
Change
`import org.junit.Assert;`
to
`import static org.junit.Assert.assertEquals;`
**AND**
Update all
`Assert.{{ method }}*`
to
`{{ method }}*`


madanadit:

PR #8992, opened 2019-05-06
Add runUfsTest CLI to trigger UnderFileSystem Contract Test

LuQQiu:
This PR moves all the `UnderFileSystemContractTest` to `alluxio-examples` package. Users can directly trigger the tests through CLI `./bin/alluxio runUfsTest --path <ufs_path> -D(aws credentials)`.

If the given ufs path is S3A path, the tests can also be treated as S3A compatibility tests. Users can pass in their S3 credentials to see if the target storage system can fulfill the minimum S3 compatibility requirements in order to run with Alluxio via Alluxio's integration with S3.


LuQQiu:
@calvinjia Added the S3 streaming upload contract test, PTAL.
I am investigating how to trigger this contract test together with the s3 under filesystem contract test in client CLI. The trigger part will be included in future PR. Thanks

LuQQiu:
Jenkins, test this please

LuQQiu:
@calvinjia @aaudiber The PR needs testing and docs improvements, but I want to get some opinions regarding the high-level implementation of this PR.

The `./bin/alluxio runS3ATests` will trigger the `S3ACompatibilityTest` inside the `alluxio-examples/src`
which will use the `JUnitCore` to run all the tests inside the `S3AStreamingUploadContractTest.class` and `S3AUnderFileSystemContractTest.class`.

`S3ACompatibilityContractTest` is put under `alluxio-underfs-s3a/test` which also use `JUnitCore` to run `S3AStreamingUploadContractTest.class` and `S3AUnderFileSystemContractTest.class`.

The `S3AStreamingUploadContractTest.class` and `S3AUnderFileSystemContractTest.class` are all placed in the `alluxio-underfs-s3a/src/` instead of `alluxio-underfs-s3a/test`.

`S3AStreamingUploadContractTest.class` extends `AbstractUnderFileSystemContractTest`. `AbstractUnderFileSystemContractTest` has some test dependencies: `ConfigurationRule` and `ConfigurationTestUtils`. All the three test classes `AbstractUnderFileSystemContractTest`, `ConfigurationRule`, and `ConfigurationTestUtils` are moved to `alluxio-core-common/src`.

the `alluxio-examples` package now depend on the `alluxio-underfs-s3a` and `alluxio-core-common`.

The way of doing it is based on my discussion with @aaudiber that we don't want to include test jar with dependencies in our assembly jar. If we maintain `S3AStreamingUploadContractTest.class` and `S3AUnderFileSystemContractTest.class` in `alluxio-underfs-s3a/test`, the trigger class `S3ACompatibilityContractTest` should also be in `alluxio-underfs-s3a/test` to be able to trigger other tests. We need to use `maven assembly jar` to create a jar including test code with their dependencies..

Please raise any questions or suggestions. Thanks!

LuQQiu:
> Test FAILed.
> Refer to this link for build results (access rights to CI server needed):
> https://amplab.cs.berkeley.edu/jenkins//job/Alluxio-Pull-Request-Builder/3487/
> Test FAILed.

LuQQiu:
@adit @calvinjia PTAL, thanks!

bin/alluxio
```diff
@@ -31,6 +31,8 @@ function printUsage {
   echo -e "                \t NOTE: This command will stop the existing server and creates a new one!"
   echo -e "  runMesosTest  \t Test the Alluxio integration with Mesos. Try 'alluxio mesosTest -help' for more help."
   echo -e "                \t NOTE: This command requires Mesos to be running and will stop any Alluxio servers that are currently running."
+  echo -e "  runS3ATests    \t Test the S3A compatibility of the target storage system. Try 'alluxio runS3Test -help' for more help."
```
calvinjia: Is the spacing off?

LuQQiu:
Based on the discussion with @calvinjia, will work to turn `AbstractUnderFileSystemContractTest` and other extended contract tests to normal java classes (instead of maven junit test classes) that creates an actual under filesystem with the credentials and connect details and runs operations on this created under filesystem. 

The reason for this change is that including test classes in the src code is somewhat strange and we don't run the contract tests normally. 

All the contract tests will be triggered through CLI via
`./bin/alluxio runContractTest <ufs_name>`.

examples/src/main/java/alluxio/cli/S3ACompatibilityTest.java
```diff
@@ -0,0 +1,220 @@
...
+    runJUnitClass(S3AStreamingUploadContractTest.class);
+    runJUnitClass(S3AUnderFileSystemContractTest.class);
+  }
+
+  /**
+   * Prints the help message.
+   */
+  private static void printHelp() {
+    System.out.println("bin/alluxio runS3Tests");
+    System.out.println("Test if the target storage system which declares it's S3 compatible "
```
madanadit: nit: it's -> it is

examples/src/main/java/alluxio/cli/S3ACompatibilityTest.java
```diff
@@ -0,0 +1,220 @@
...
+        + "can fulfill the minimum S3A compatibility requirements in order to work with Alluxio "
+        + "through Alluxio's integration with S3. A test bucket `testS3ABucket` is required to "
```
madanadit: nit: remove "through Alluxio's integration with S3"
nit: "A test bucket `testS3ABucket`" -> "A test bucket specified by the option `testS3ABucket`"

examples/src/main/java/alluxio/cli/S3ACompatibilityTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Prints the failure information.
+   *
+   * @param failure the failure to print
+   */
+  private static void printFailureInfo(Failure failure) {
+    String testHeader = failure.getTestHeader();
+    System.out.println("Failed test: " + testHeader);
+    int endIndex = testHeader.indexOf("(");
```
madanadit: how does this check work?

examples/src/main/java/alluxio/cli/S3ACompatibilityTest.java
```diff
@@ -0,0 +1,220 @@
...
+import org.junit.runner.Result;
+import org.junit.runner.notification.Failure;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Class to perform S3A compatibility tests.
+ */
+public final class S3ACompatibilityTest {
```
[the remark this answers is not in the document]
LuQQiu: The `./bin/alluxio` trigger tests in Alluxio client jar or alluxio server jar. Most of the `./bin/alluxio` tests are located in `alluxio-examples`. put the `runContractTest` also in the `alluxio-examples` jar is to match our current usage.

but one issue that I put `runContractTest` (which is in progress) in `alluxio-examples` is that `alluxio-examples` will depend on all the `alluxio-underfs` jars.

underfs/swift/src/test/java/alluxio/underfs/swift/SwiftUnderFileSystemContractTest.java
```diff
@@ -36,7 +35,7 @@ public static void beforeClass() throws Exception {
   @Override
   public UnderFileSystem createUfs(String path, UnderFileSystemConfiguration conf)
       throws Exception {
-    return new SwiftUnderFileSystemFactory().create(path, conf, ConfigurationTestUtils.defaults());
+    return new SwiftUnderFileSystemFactory().create(path, conf, mConfiguration);
```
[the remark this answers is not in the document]
LuQQiu: I moved all the `UnderFileSystem contract tests` to `alluxio-examples`. Will solve the conflicts after your PR merged.

LuQQiu:
@madanadit @calvinjia Push all the contract tests in the alluxio-examples and change them from junit tests to normal classes. PTAL, thanks

examples/pom.xml
```diff
@@ -67,5 +67,46 @@
...
+      <artifactId>alluxio-underfs-swift</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-underfs-wasb</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
calvinjia: Any reason for explicitly stating `compile` scope? Would it be possible to make these `provided` since the user should be linking their ufs jars from the `lib` folder? Otherwise the examples jar will become quite large after including all the ufses

examples/src/main/java/alluxio/cli/CliUtils.java
```diff
@@ -56,4 +63,23 @@ public static boolean runExample(final Callable<Boolean> example) {
...
   }
+
+  /**
+   * Cleans all the files or sub directories inside the given directory
+   * in the under filesystem.
+   *
+   * @param ufs the under filesystem
+   * @param directory the directory to clean
+   */
+  public static void cleanup(UnderFileSystem ufs, String directory) throws IOException {
```
calvinjia: Maybe this should be `cleanupUfs`?

examples/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  private void cleanup(UnderFileSystem ufs, String testDir) throws IOException {
+    ufs.deleteDirectory(testDir, DeleteOptions.defaults().setRecursive(true));
+    ufs.close();
+  }
+
+  private void createUnderFileSystemFactory() throws IOException {
+    switch (mUfsType) {
+      case "HDFS":
```
madanadit: instead if enumerating all ufs's, can't we service load from the libs directory?

madanadit: with a dependency of `alluxio-underfs` as `provided` (instead of individual `compiled` ufs) and using `UnderFileSystemFactoryRegistry.find(path, UnderFileSystemConfiguration.defaults())` to load the factory, we are saying that contract tests can only be run from the servers (which have the `lib` and `extensions` directory) instead of bundling up all ufs modules into the client examples jar

examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
```diff
@@ -0,0 +1,1345 @@
...
+    mUfsPath = ufsPath;
+    mTopLevelTestDirectory = topLevelTestDirectory;
+    mUfs = ufs;
+    mConfiguration = configuration;
+  }
+
+  /**
+   * Runs all the tests.
+   */
+  public void runTests() throws Exception {
```
madanadit: use reflection instead?

examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
```diff
@@ -0,0 +1,1345 @@
...
+
+  /**
+   * Gets the S3A operations related to the failed test. This method
+   * should only be called when the ufs is S3A.
+   *
+   * @return the related S3A operations
+   */
+  public List<String> getRelatedS3AOperations() {
+    List<String> operations = new ArrayList<>();
+    switch (mFailedTestName) {
```
madanadit: same here

madanadit:
@LuQQiu I made another pass. The 2 things I don't like (both related to the long enumerations):
1. test method names. can we get rid of this by reflection? something like [this](https://stackoverflow.com/questions/19025761/how-can-i-run-all-methods-of-an-external-class)
2. listing of the ufs modules. the situation here is that the client (example) is loading the ufs modules (which is a deviation from only alluxio servers loading the ufses so far). i think that part is fine. but our service loader logic already discovers ufses from libs and extensions directory. why don't we re-use that and only allow running examples from a node which has the jars pre-deployed (like the server but not necessarily). @calvinjia thoughts?

calvinjia:
@LuQQiu I agree with @madanadit that we should put the jars in provided scope and discover them during runtime.

examples/src/main/java/alluxio/examples/S3ASpecificOperations.java
```diff
@@ -0,0 +1,172 @@
...
+    if (getMultipartUploadList().size() != 0) {
+      throw new IOException("Failed to clean the in progress multipart upload");
+    }
+  }
+
+  /**
+   * @return a list of intermediate multipart uploads
+   */
+  private List<MultipartUpload> getMultipartUploadList() {
+    return mUfs.getS3Client().listMultipartUploads(
```
LuQQiu: The common operations run well, but there are still `ClassNotFoundException: alluxio.underfs.s3a.S3AUnderFileSystem` because I need `S3AUnderFileSystem` to list the multipart upload and `((S3ALowLevelOutputStream) outputStream).waitForAllPartsUpload();`. Now working to address this issue.

LuQQiu:
@madanadit @calvinjia Addressed the comments, PTAL, thanks!

examples/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -0,0 +1,325 @@
...
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Integration tests for Alluxio under filesystems. It describes the contract of Alluxio
+ * with the UFS through the UFS interface.
+ */
+public final class UnderFileSystemContractTest {
```
madanadit: nit: add a comment upfront specifying that all methods ending with the suffix `Test` will be run

examples/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -0,0 +1,325 @@
...
+      System.exit(1);
+    }
+
+    mConf.set(PropertyKey.UNDERFS_LISTING_LENGTH, "50");
+    mConf.set(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, "512B");
+    // Increase the buffer time of journal writes to speed up tests
+    mConf.set(PropertyKey.MASTER_JOURNAL_FLUSH_BATCH_TIME_MS, "1sec");
+
+    createUnderFileSystem();
+    mIsS3Ufs = mUfs.getUnderFSType().equals(S3_IDENTIFIER);
```
madanadit: nit: i prefer not having this redundant field. is it expensive to query `getUnderFSType`?

examples/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  /**
+   * Gets the S3A operations related to the failed test. This method
+   * should only be called when the ufs is S3A.
+   *
+   * @param testName the name of the failed test
+   * @return the related S3A operations
+   */
+  private List<String> getRelatedS3AOperations(String testName) {
```
madanadit: will using annotations on methods in `S3ASpecificOperations` and using reflection to query annotations make this cleaner? not sure if there's a way around this ugly `switch`

examples/src/main/java/alluxio/examples/S3ASpecificOperations.java
```diff
@@ -0,0 +1,159 @@
...
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.examples;
+
+import alluxio.Constants;
+import alluxio.cli.UnderFileSystemContractTest.RelatedS3Operations;
```
madanadit: nit: should we move the class `RelatedS3Operations` under `alluxio.examples`?

LuQQiu:
@madanadit Thanks a lot!! The code looks much clean and easy to read than before!!

examples/src/main/java/alluxio/examples/S3ASpecificOperations.java
LuQQiu: Make sense. The import does not look good.
Moved to `alluxio.examples`

LuQQiu:
@calvinjia Could you take a look at the new changes. Now it uses reflection to trigger tests and get the method annotation. Thanks

examples/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -0,0 +1,228 @@
...
+    mConf.set(PropertyKey.UNDERFS_LISTING_LENGTH, "50");
+    mConf.set(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, "512B");
+    // Increase the buffer time of journal writes to speed up tests
+    mConf.set(PropertyKey.MASTER_JOURNAL_FLUSH_BATCH_TIME_MS, "1sec");
+
+    createUnderFileSystem();
+
+    runCommonOperations();
+
+    if (mUfs.getUnderFSType().equals(S3_IDENTIFIER)) {
```
calvinjia: Wouldn't we want to run this for all object stores?

LuQQiu: The test `S3SpecificOperation` is only related to S3. Only S3 uses streaming upload. The logs inside the test are also S3 specific. The other object store can also run this test but it is not that useful and the logs may be misleading.

calvinjia: Ah okay, I still would prefer a configurable way to enable this test for other object stores in case we provide these features in future.

LuQQiu: The other tests in the `S3ASpecificOperations` are also valid for other ufs, excepted the `createAndAbortTest` which I add a line for checking the ufs is s3. Added the `--force` that users can use to force run all the tests.

LuQQiu:
@calvinjia Addressed the comment, PTAL, thanks

calvinjia:
@LuQQiu Instead of adding a force flag, could we allow the user to modify what the s3 identifier is? That way if they want to test something other than s3 with the additional s3 tests, they can just pass in the identifier of the storage they are testing.

calvinjia:

PR #8995, opened 2019-05-07
Implement sink for journal entries

gpang:
The journal sink is a new interface for the journal system to write journal entries to. The journal sink is useful for examining journal entries applied to the master. Only LOG journal entries are appended to the sink, so no checkpoint entries are written to the sink.

The journal system writes entries to the journal sink, as the leader and also as the standby. If the journal is the leader, the journal system appends to the sink when writing/flushing to the journal. If the journal is a standby, the journal system appends to the sink when replaying the journal to catch up.

Journal sink implementations must not block any of the methods, since they are called within the journal system, and any delay will negatively impact operational clients on the leader master.


core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
```diff
@@ -42,6 +56,60 @@ public synchronized void stop() throws InterruptedException, IOException {
     stopInternal();
   }
 
+  @Override
+  public void addJournalSink(Master master, JournalSink journalSink) {
+    mSinkLock.writeLock().lock();
```
aaudiber: would the `LockResource` pattern work here?

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
```diff
@@ -42,6 +56,60 @@ public synchronized void stop() throws InterruptedException, IOException {
...
+    try {
+      mJournalSinks.compute(master.getName(), (key, entry) -> {
```
aaudiber: Since we're locking externally, this can be done in one line with
```java
mJournalSinks.computeIfAbsent(master.getName(), (k, v) -> new HashSet<>()).add(journalSink);
```

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
```diff
@@ -42,6 +56,60 @@ public synchronized void stop() throws InterruptedException, IOException {
...
+    mSinkLock.writeLock().lock();
+    try {
+      mJournalSinks.computeIfPresent(master.getName(), (key, entry) -> {
+        if (entry != null) {
+          entry.remove(journalSink);
+          return entry;
+        }
+        return null;
+      });
+      // Compute the full set of sinks on removal.
```
aaudiber: why not remove just `journalSink`?

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
```diff
@@ -42,6 +56,60 @@ public synchronized void stop() throws InterruptedException, IOException {
...
+      for (Set<JournalSink> sinks : mJournalSinks.values()) {
+        mAllJournalSinks.addAll(sinks);
+      }
+    } finally {
+      mSinkLock.writeLock().unlock();
+    }
+  }
+
+  @Override
+  public Set<JournalSink> getJournalSinks(Master master) {
```
aaudiber: annotate master with `@Nullable`

core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
```diff
@@ -143,12 +146,21 @@ public boolean isReleasable() {
    */
   private volatile boolean mStopFlushing = false;
 
+  /**
+   * Collection of sinks for this journal.
+   */
+//  private final Set<JournalSink> mJournalSinks = new ConcurrentHashSet<>();
```
aaudiber: should we remove this?

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
```diff
@@ -145,6 +148,26 @@
...
+   * @param master the master from which to remove the journal sink
+   * @param journalSink the journal sink to remove
+   */
+  default void removeJournalSink(Master master, JournalSink journalSink) {}
+
+  /**
+   * @param master the master from which to remove the journal sink
+   * @return a set of {@link JournalSink} for the given master, or all sinks if master is null
+   */
+  default Set<JournalSink> getJournalSinks(Master master) {
```
aaudiber: are these default implementations necessary? I think it's safer to leave them non-default so that `JournalSystems` are required to specify how they manage journal sinks

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
gpang: Yep, updated.

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
gpang: Updated. 
I forgot to change this after I added the external locking.

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
gpang: It is possible that a sink instance is associated with multiple masters, so removing it may not be correct.

core/server/common/src/main/java/alluxio/master/journal/AbstractJournalSystem.java
gpang: Done

core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
gpang: Ooops, removed

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
gpang: Done

gpang:

PR #9014, opened 2019-05-07
Track metrics over worker and client restarts

ZacBlanco:
This PR improves the Metrics collection on the master and
workers by allowing cluster-aggregated metrics on the master
to reflect accurate values even when workers (or clients)
restart.

In order to support properly tracking metrics over restarts we
need to know which values will differ over restarts. Alluxio's
MetricsSystem uses 4 types of metrics: GAUGE, COUNTER,
METER, and TIMER. COUNTER metrics are the only ones
that change over restarts. The rest vary with time, and can
safely assume the same behavior they had previously.

This change adds MetricType, and updates the client and
master behavior such that when the client or worker sends
metrics over the wire, that the metrics reported for COUNTER
will only include the incremental values that the given metrics
should change, rather than the absolute value since the
process started. This allows the master to track metrics for
the life of its process and its workers, even if the workers
restart.


core/base/src/main/java/alluxio/collections/IndexedSet.java
```diff
@@ -130,12 +130,8 @@ public IndexedSet(IndexDefinition<T, ?> primaryIndexDefinition,
...
 
     for (IndexDefinition<T, ?> indexDefinition : indexDefinitions) {
-      FieldIndex<T, ?> index;
-      if (indexDefinition.isUnique()) {
-        index = new UniqueFieldIndex<>(indexDefinition);
-      } else {
-        index = new NonUniqueFieldIndex<>(indexDefinition);
-      }
+      FieldIndex<T, ?> index = indexDefinition.isUnique()
+          ? new UniqueFieldIndex<>(indexDefinition) : new NonUniqueFieldIndex<>(indexDefinition);
```
calvinjia: Consider
```
if (indexDefinition.isUnique) {
   indices.put(indexDefinition, new Unique ...)
...
```

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -412,6 +415,53 @@ public static void resetAllCounters() {
...
+      // contain the difference of the current value, and the last value sent. If it doesn't
+      // yet exist, just send the current value
+      if (m.getMetricType() == MetricType.COUNTER) {
+        Metric prev = LAST_REPORTED_METRICS.replace(m.getFullMetricName(), m);
+        // On restarts counters will be reset to 0, so whatever the value is the first time
+        // this method is called represents the value which should be added to the master's
+        // counter.
+        double diff = prev != null ? m.getValue() - prev.getValue() : m.getValue();
+        rpcMetrics.add(new Metric(m.getInstanceType(), m.getHostname(), m.getInstanceId(),
+            m.getMetricType(), m.getName(), diff).toProto());
```
calvinjia: Could do `toProto`, `toBuilder` and modify the value?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -412,6 +415,53 @@ public static void resetAllCounters() {
...
+   */
+  public static List<alluxio.grpc.Metric> reportWorkerMetrics() {
+    return reportMetrics(InstanceType.WORKER);
+  }
+
+  /**
+   * @return the client metrics to send via RPC
+   */
+  public static List<alluxio.grpc.Metric> reportClientMetrics() {
+    return reportMetrics(InstanceType.CLIENT);
```
calvinjia: How are the metrics currently tied to instance type? I think workers could potentially have client metrics due to using an Alluxio client for some operations under the hood?

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -111,21 +107,55 @@ public void putWorkerMetrics(String hostname, List<Metric> metrics) {
...
-        mClientMetrics.add(metric);
+      putReportedMetrics(mClientMetrics, getFullInstanceId(hostname, clientId), metrics);
+    }
+  }
+
+  /**
+   * Update the reported metrics with the given instanceId and set of metrics received from a
+   * worker or client.
+   *
+   * Any metrics from the given instanceId which are not reported in the new set of metrics are
```
calvinjia: Why do we remove the old metrics if they are not contained in the new set?

ZacBlanco: This was the previous behavior of the metrics heartbeat. The behavior itself may be questionable, but I didn't want to modify it for the scope of this PR.

One issue, I think, is that the metric hash code depends upon the value of the metric itself. If we're getting different values over heartbeats, its possible to store more than one metric of the same type if the previous isn't removed. I, however, don't know if there is ever a case where the lists of worker metrics differs between heartbeats (whether they become a subset/superset).

core/base/src/main/java/alluxio/collections/IndexedSet.java
ZacBlanco: that seems more verbose, no?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
ZacBlanco: Yes, it is possible that workers also have "clients" but those clients should act like any other application and report their metrics via the client metrics heartbeat. However, that all hinges on the `allMetrics` method actually returning _only_ worker or client metrics.

Because it's static and will return everything for the JVM, we may need to do some more filtering....investigation required...

ZacBlanco: Similarly, it also hinges on the fact that metrics registered within the JVM also are registered properly under Client, Master, Worker, etc.. 

When I'm working in my IDE this isn't implemented properly such that different Alluxio processes can live alongside one another in the metric system.

I think my vote is to actually just get rid of the static system entirely in this PR. It bothers me enough that I really just want to do away with it, especially if it is something on the radar anyways. Maybe an extra day of work to finish.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
ZacBlanco: Fixed

calvinjia:

PR #9015, opened 2019-05-07
Add address in exception message when binding port for web server fails

apc999:
No port is given in the exception message if the port to bind is taken

image


apc999:

PR #9016, opened 2019-05-07
Update metrics servlet to respond with valid Content-Type

ZacBlanco:
`text/json` is not a valid content type. See the following resource:
https://www.iana.org/assignments/media-types/media-types.xhtml

We should be using the standard `application/json` which also
defaults its charset to utf-8, so we won't need to specify it explicitly
either.


ZacBlanco:
@calvinjia PTAL

calvinjia:

PR #9019, opened 2019-05-07
Change pin behavior to pin to specific tier / media type

yuzhu:
There are several things in this PR. 
1. Changes to Master inode metadata to have a BitSet of possible storage media. 
2. Pass a desired pinning storage medium type to `loadBlock` method of JobUtils.java. 
3. This storage medium type eventually gets passed to the `createBlock` method on the worker, and worker finds a storage medium which matches the desired storage medium type. 

TODO: 
BlockHeartBeatReport related changes


core/transport/src/grpc/block_worker.proto
```diff
@@ -123,14 +124,15 @@ message OpenLocalBlockResponse {
...
-// next available id: 6
+// next available id: 8
 message CreateLocalBlockRequest {
   optional int64 block_id = 1;
   optional int32 tier = 3;
   optional int64 space_to_reserve = 4;
   // If set, only reserve space for the block.
   optional bool only_reserve_space = 5;
   optional bool cleanup_on_failure = 6;
+  optional string medium_type = 7;
```
calvinjia: I would prefer to have a consistent name for this type of field, currently we have `writeMedium`, `mediumType`, and `pinnedMedia`.

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -239,6 +250,18 @@ public WriteType getWriteType() {
...
+  /**
+   * Set the write medium type of the file
+   *
+   * @param mediumType write medium type
+   * @return the updated options object
+   */
+  public OutStreamOptions setWriteMedium(String mediumType) {
+    mWriteMedium = mediumType;
+    return this;
+  }
```
calvinjia: Update `toString`, `Equals`, `Hashcode`

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1626,6 +1626,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_TIERED_STORE_GLOBAL_MEDIA =
+      new Builder(Name.MASTER_TIERED_STORE_GLOBAL_MEDIA)
+          .setDefaultValue("MEM, SSD, HDD")
+          .setDescription("The list of medias we support in the system.")
```
calvinjia: Media is already plural

core/transport/src/grpc/block_worker.proto
yuzhu: I changed `writeMedium` to  `mediumType`, I think `pinnedMedia` refers to a list of pinned medium specifically, so that one is kind of different?

core/common/src/main/java/alluxio/wire/FileInfo.java
```diff
@@ -298,6 +299,13 @@ public DefaultAccessControlList getDefaultAcl() {
     return (mDefaultAcl == null) ? new ArrayList<>() : mDefaultAcl.toStringEntries();
   }
 
+  /**
+   * @return a list of pinned locations
+   */
+  public List<String> getPinnedLocation() {
```
calvinjia: Does `Location` differ from `Medium`? If not, I would prefer to use the same name

core/common/src/main/java/alluxio/client/file/URIStatus.java
```diff
@@ -236,6 +236,13 @@ public boolean isPinned() {
     return mInfo.isPinned();
   }
 
+  /**
+   * @return the pinned location list
+   */
+  public List<String> getPinnedLocation() {
```
calvinjia: Should this be plural? Also, should we use a `Set` instead of a list, or do we expect use cases for duplicated locations?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -865,25 +865,40 @@ public void deleteInode(RpcContext rpcContext, LockedInodePath inodePath, long o
     }
   }
 
+  private boolean checkPinningValidity(List<String> pinnedLocations) {
+    List<String> mediaList = ServerConfiguration.getList(PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIA, ",");
+    for (String medium : pinnedLocations) {
+      if (!mediaList.remove(medium)) {
```
calvinjia: I would prefer a contains check and having `pinnedLocations` be a set, unless you have a use case for duplicate medium types?

core/server/master/src/main/java/alluxio/master/file/meta/InodeView.java
```diff
@@ -107,6 +108,16 @@
...
 
+  /**
+   * @return the pinned location
+   */
+  BitSet getPinnedLocation();
+
+  /**
+   * @return the pinned location list
+   */
+  List<String> getPinnedLocationList();
```
calvinjia: Do we need both of these methods?

core/server/master/src/main/java/alluxio/master/file/meta/MutableInode.java
```diff
@@ -162,6 +170,20 @@ public AccessControlList getACL() {
     return mAcl;
   }
 
+  @Override
+  public BitSet getPinnedLocation() { return mPinnedLocation; }
+
+  @Override
+  public List<String> getPinnedLocationList() {
+    List<String> mediaList = ServerConfiguration.getList(PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIA, ",");
```
calvinjia: Because we are implicitly associating the bit set with media, if the user changes the property (for example removes a tier), all the previously stored state will become incorrect?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -141,11 +141,11 @@ private void check(Set<Long> inodes, ReplicationHandler handler, Mode mode) {
...
     for (long inodeId : inodes) {
-      // TODO(binfan): calling lockFullInodePath locks the entire path from root to the target
-      // file and may increase lock contention in this tree. Investigate if we could avoid
-      // locking the entire path but just the inode file since this access is read-only.
-      try (LockedInodePath inodePath =
-          mInodeTree.lockFullInodePath(inodeId, LockPattern.READ)) {
+        // TODO(binfan): calling lockFullInodePath locks the entire path from root to the target
+        // file and may increase lock contention in this tree. Investigate if we could avoid
+        // locking the entire path but just the inode file since this access is read-only.
+        try (LockedInodePath inodePath =
```
calvinjia: Unnecessary indent?

core/transport/src/grpc/block_worker.proto
calvinjia: Yeah you are right, there are two concepts, singular medium type to write to and a list of media to pin to

yuzhu:
@calvinjia PTAL, thanks.

yuzhu:
This was addressed in another PR #9084

PR #9039, opened 2019-05-08
Fix docker doc

aaudiber:
Some embedded journal docs slipped in when cherry-picking from master


apc999:

Issue #8947, opened 2019-05-01
Docker does not work with certain property keys


apc999:
**Alluxio Version:**
2.0.0-SNAPSHOT

**Describe the bug**

```bash
$ bin/alluxio runClass alluxio.cli.GetConfKey ALLUXIO_MASTER_MOUNT_TABLE_ROOT_OPTION_AWS_ACCESSKEYID
alluxio.master.mount.table.root.option.aws.accesskeyid
```
and `alluxio.master.mount.table.root.option.aws.accesskeyid` will not work but only `alluxio.master.mount.table.root.option.aws.accessKeyId` will work

**To Reproduce**
passing `ALLUXIO_MASTER_MOUNT_TABLE_ROOT_OPTION_AWS_ACCESSKEYID` as the property to docker

**Expected behavior**
We should have one way to pass the property

**Urgency**
High

**Additional context**
One solution might be making Alluxio properties case independent by fixing `alluxio.conf.PropertyKey#isValid` and `alluxio.conf.PropertyKey#fromString`

ZacBlanco:
As an alternative solution, instead of relying on camel case I think we should always `.` between words as is convention for (probably) > 90% of the property keys. I think we should follow the same convention for all keys.

While not affected by the casing of the characters I'm specifically thinking about tiered storage keys that use `level0`, `level1`, etc.. instead of `level.0`, `level.1`...

apc999:
@ZacBlanco , I discussed this issue with @aaudiber , another approach might be just pass all properties to `docker run`  through env variable like `$ALLUXIO_JAVA_OPTS`, `$ALLUXIO_MASTER_JAVA_OPTS` or `$ALLUXIO_WORKER_JAVA_OPTS`. Then we don't need to modify the code but only the recommended way to run docker

mingfang:
Looks like there are some keys that must be in mixed cases https://github.com/Alluxio/alluxio/blob/12d159a0d74369e256b3ded8cdf129d4b3e0a8c2/shell/src/main/java/alluxio/cli/GetConfKey.java#L31

I think we should just use JAVA_OPTS for everything.

aaudiber:
Created https://github.com/Alluxio/alluxio/pull/8993 to support spaces in environment variables, and also recommend using `-e ALLUXIO_JAVA_OPTS="-Dkey1=value1 -Dkey2=value2 ..."` for everything

apc999:
Both 1.8 branch and master branch have this fix. Close this ticket now

PR #9040, opened 2019-05-08
Support multiple ALLUXIO_JAVA_OPTS and update doc

aaudiber:
Addresses https://github.com/Alluxio/alluxio/issues/8947

ALLUXIO_JAVA_OPTS can support any type of property key without
needing translation from environment variable name to property
key name. This PR fixes two bugs in supplying space-separated
options.

1. To iterate over environment variables, we need to set
IFS='\n'. Otherwise, environment variables with values
containing spaces will be parsed as multiple variables.
2. Since environment variables may contain spaces, we need to
wrap their values in strings when writing to alluxio-env.sh
e.g. ALLUXIO_JAVA_OPTS="-Dfoo=bar -Dbaz=bat" instead of
ALLUXIO_JAVA_OPTS=-Dfoo=bar -Dbaz=bat.

Now that ALLUXIO_JAVA_OPTS can support multiple options, it is
the preferred way to configure Alluxio in Docker. This PR
updates the Docker doc to use ALLUXIO_JAVA_OPTS. We also
switch hostname examples to use `-` instead of `_` since `_` isn't
a valid character in a hostname.

With this PR, we no longer support passing and translating
individual configuration properties


apc999:

PR #9042, opened 2019-05-09
Remove implicit inode locking

aaudiber:
Previously, we would skip taking "unnecessary" locks on inodes and edges. A lock was considered unnecessary if one of its ancestors was already write-locked, e.g. if we've write-locked /a/b, we don't need to acquire any more locks to lock /a/b/c/d. This is what we called implicit locking.

Implicit locking works well when all locks are taken starting from the root. However, we sometimes want to lock individual inodes without needing to lock their ancestors. In particular, CachingInodeStore needs to evict inodes from memory to RocksDB, which requires locking the inodes directly (locking their ancestors would be too expensive).

To support random locks, we remove the concept of implicit locking. With this PR, we replace implicit locks with explicit write locks. For the above example, we will now take write locks on all inodes and edges leading from /a/b to /a/b/c/d.

One new invariant we follow is that all lock lists contain some number of read-locked entries followed by some number of write-locked entries. In particular, we never take a read lock following a write lock. So if you've write-locked /a/b and then request to extend the path with a read-lock on /a/b/c, you will actually take a write lock.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/CompositeInodeLockList.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/CompositeInodeLockList.java
@@ -11,7 +11,10 @@
 
 package alluxio.master.file.meta;
 
-import java.util.ArrayList;
+import alluxio.concurrent.LockMode;
+
+import com.google.common.base.Preconditions;
+
 import java.util.List;
 
 import javax.annotation.concurrent.NotThreadSafe;
@@ -26,9 +29,11 @@ import javax.annotation.concurrent.NotThreadSafe;
  * Modification is only supported for the non-base part of the lock list.
  */
 @NotThreadSafe
-public class CompositeInodeLockList extends InodeLockList {
+public class CompositeInodeLockList implements InodeLockList {
   /** The base lock list for this composite list. */
   private final InodeLockList mBaseLockList;
+  private final InodeLockList mSubLockList;
+  private final int mBaseListSize;
 
   /**
    * Constructs a new lock list, using an existing lock list as the base list.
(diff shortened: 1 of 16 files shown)


core/server/master/src/main/java/alluxio/master/file/meta/CompositeInodeLockList.java
```diff
@@ -26,56 +29,144 @@
  * Modification is only supported for the non-base part of the lock list.
  */
 @NotThreadSafe
-public class CompositeInodeLockList extends InodeLockList {
+public class CompositeInodeLockList implements InodeLockList {
   /** The base lock list for this composite list. */
   private final InodeLockList mBaseLockList;
+  private final InodeLockList mSubLockList;
```
gpang: Can you add comments for these fields?

core/server/master/src/main/java/alluxio/master/file/meta/InodeLockList.java
```diff
@@ -12,366 +12,204 @@
...
+   * @param childName the child name of the edge
    * @param mode the mode to lock in
    */
-  public void lockRootEdge(LockMode mode) {
-    Preconditions.checkState(mEntries.isEmpty());
-
-    mEntries.add(new EdgeEntry(mInodeLockManager.lockEdge(ROOT_EDGE, mode), ROOT_EDGE));
-    mLockMode = mode;
-  }
+  void lockEdge(Inode inode, String childName, LockMode mode);
```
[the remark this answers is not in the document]
aaudiber: If a lock list starts with a non-root edge, there isn't a previous inode to work off of. If the lock list does have a previous inode, we check that its id matches the specified inode

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,381 @@
...
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * A simple inode lock list.
+ */
+@NotThreadSafe
+public class SimpleInodeLockList implements InodeLockList {
+  private static final Logger LOG = LoggerFactory.getLogger(InodeLockList.class);
```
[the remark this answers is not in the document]
aaudiber: good catch, fixed

aaudiber:
@gpang comments addressed, PTAL

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+    addEntry(createInodeEntry(inode, mode));
+  }
+
+  @Override
+  public void lockEdge(Inode lastInode, String childName, LockMode mode) {
+    mode = nextLockMode(mode);
+    long edgeParentId = lastInode.getId();
+    if (!mEntries.isEmpty()) {
+      Preconditions.checkState(endsInInode());
```
gpang: can we add messages for these preconditions?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,381 @@
...
+      Preconditions.checkState(lastInode().getId() == edgeParentId);
+    }
+    addEntry(createEdgeEntry(new Edge(edgeParentId, childName), mode));
+  }
+
+  @Override
+  public void lockRootEdge(LockMode mode) {
+    Preconditions.checkState(mEntries.isEmpty());
+    addEntry(createEdgeEntry(ROOT_EDGE, mode));
+    if (mode == LockMode.WRITE) {
```
gpang: does `addEntry` not take care of this?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,381 @@
...
+    addEntry(newEntry);
+  }
+
+  @Override
+  public void downgradeLastEdge() {
+    Preconditions.checkState(!endsInInode());
+    Preconditions.checkState(!mEntries.isEmpty());
+    Preconditions.checkState(endsInWriteLock());
+
+    if (!endsInMultipleWriteLocks()) {
```
gpang: NIT: can we mirror the logic in `downgradeLastInode`?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+    if (mEntries.get(0) instanceof InodeEntry) {
+      return (mEntries.size() + 1) / 2;
+    }
+    return mEntries.size() / 2;
+  }
+
+  @Override
+  public boolean isEmpty() {
+    return mEntries.isEmpty();
```
gpang: can `mEntries` have just a single edge (and no inode), and if so, is that considered empty?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
aaudiber: good idea, added

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
aaudiber: good call, removed

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
aaudiber: done

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
aaudiber: that's a valid state. I would not consider it empty since the list contains a locked edge.

core/server/master/src/main/java/alluxio/master/file/meta/InodeLockList.java
```diff
@@ -12,366 +12,204 @@
...
- * Represents a locked path within the inode tree. All lock lists must start with the root. Both
- * inodes and edges are locked along the path.
+ * Represents a locked path within the inode tree. Both inodes and edges are locked along the path.
+ * The path may start from either an edge or an inode.
+ *
+ * A lock list always contains some number of read locks (possibly zero) followed by some number of
+ * write locks (possibly zero).
+ *
+ * This class uses list notation to describe lock lists. Nodes are single letters, edges are
+ * parentName->childName, and write-locks are indicated with '*'. For example, a lock list for
```
LuQQiu: Not `parentId to childName` ??

core/server/master/src/main/java/alluxio/master/file/meta/InodeLockList.java
```diff
@@ -12,366 +12,204 @@
...
   /**
    * Downgrades the last inode from a write lock to a read lock. The read lock is acquired before
    * releasing the write lock.
+   *
+   * Example
+   * Starting from [a, a->b, b*]
+   *
+   * downgradeLastInode() results in [a, a->b, b]
+   *
+   * If the last inode is not the only write-locked inode, no downgrade occurs.
```
LuQQiu: If the last inode is **not**(should it be removed?) the only write-locked inode, no downgrade occurs

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+  }
+
+  @Override
+  public void unlockLastInode() {
+    Preconditions.checkState(endsInInode(),
+        "Cannot unlock last inode when the lock list %s ends in an edge", this);
+    Preconditions.checkState(!mEntries.isEmpty(),
+        "Cannot unlock last inode when the lock list is empty");
+    removeLastEntry();
```
LuQQiu: Will it be better to swap these two `checkState`? 
since `endsInInode==false` when `mEntries.isEmpty` or `endsInEdge`.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,404 @@
...
+        "Cannot unlock last edge when the lock list is empty");
+    removeLastEntry();
+  }
+
+  @Override
+  public void downgradeLastInode() {
+    Preconditions.checkState(endsInInode(),
+        "Cannot downgrade last inode when lock list %s ends in an edge", this);
+    Preconditions.checkState(!mEntries.isEmpty(),
+        "Cannot downgrade last inode when the lock list is empty");
```
LuQQiu: Same here

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,404 @@
...
+   *
+   * @return the edge entry
+   */
+  private EdgeEntry createEdgeEntry(Edge edge, LockMode mode) {
+    LockResource lock = mInodeLockManager.lockEdge(edge, mode);
+    return new EdgeEntry(lock, edge, mode);
+  }
+
+  /**
+   * Locks an inode and creates and entry without adding the entry to the lock list yet.
```
LuQQiu: `creates and entry` to `creates an entry`

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -0,0 +1,404 @@
...
+    if (!endsInWriteLock() && entry.getMode() == LockMode.READ) {
+      mFirstWriteLockIndex++;
+    }
+    mEntries.add(entry);
+  }
+
+  /**
+   * Removes and unlocks the last entry from mEntries.
+   */
+  private void removeLastEntry() {
```
LuQQiu: Are all the calls to operations inside `SimpleInodeLockList` sequential?
Who maintains the synchronization of `SimpleInodeLockList`?

gpang:

PR #9044, opened 2019-05-09
Update example insert statement for EMR documentation

apc999:
Example insert statement for presto fails as it expects a INT and the example uses a string.


calvinjia:

PR #9047, opened 2019-05-10
Consolidate journal tools

ggezer:
Currently Alluxio supports two journal systems, (UFS, RAFT) and have separate tools for both. (JournalTool, RaftJournalTool). This PR aims to enhance JournalTool to work with RAFT journals as well.


ggezer:
@LuQQiu PTAL.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -76,16 +117,17 @@
           "The start log sequence number (inclusive). Set to 0 by default.")
       .addOption("end", true,
           "The end log sequence number (exclusive). Set to +inf by default.")
+      .addOption("inputDir", true,
+          "The input directory to read journal content from.")
```
LuQQiu: If it's easy to do so, could we add something to tell users more information about `inputDir`, like it is not required and the input dir should be disk directory

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +212,467 @@ private static void usage() {
...
+    private NetworkAddressUtils.ServiceType getRaftServiceType() {
+      if (mMaster.contains("job")) {
+        return NetworkAddressUtils.ServiceType.JOB_MASTER_RAFT;
+      } else {
+        return NetworkAddressUtils.ServiceType.MASTER_RAFT;
+      }
+    }
+
+    /**
+     * Writes given entry after goin through range and validity checks.
```
LuQQiu: `goin` -> `go`

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +212,467 @@ private static void usage() {
...
+    }
+
+    /**
+     * Implementation of {@link StateMachine} that echoes its state for dumping.
+     *
+     * EchoJournalStateMachine is attached to the cluster as PASSIVE, which makes it receive all
+     * snapshots happening in the FOLLOWER node that it attaches to. For practical purposes, we
+     * will dump the first snapshot only and stop dumping snapshots afterwards.
+     */
+    public static class EchoJournalStateMachine extends StateMachine implements Snapshottable {
```
LuQQiu: The expected behavior of copycat passive state is that the state machine provides one snapshot and the journal entries after the snapshot. 

But in our case, we will receive multiple snapshots, then the journal entries that the copycat provides is starting from the first snapshot or from the last snapshot?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +212,467 @@ private static void usage() {
...
+        }
+        // Don't read snapshots coming after this one.
+        mFirstSnapshotRead = true;
+        LOG.debug("Installing snapshot: {}", snapshotReader.readLong());
+        // Reading the snapshot.
+        try (CheckpointInputStream checkpointStream =
+            new CheckpointInputStream(new SnapshotReaderStream(snapshotReader))) {
+          mCheckpointConsumer.accept(checkpointStream);
+        } catch (Exception exc) {
+          LOG.error(exc.toString());
```
LuQQiu: Why not throw the exception?

LuQQiu:
@ggezer Could you add a test for dumping embedded journal checkpoint from copycat cluster?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +212,467 @@ private static void usage() {
...
+      super(master, start, end, outputDir, inputDir);
+    }
+
+    @Override
+    public void dumpJournal() throws Throwable {
+      UfsJournal journal = new UfsJournalSystem(getJournalLocation(mInputDir), 0)
+          .createJournal(new NoopMaster(sMaster));
+      try (
+          PrintStream out =
+              new PrintStream(new BufferedOutputStream(new FileOutputStream(mJournalEntryFile)));
```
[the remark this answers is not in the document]
ggezer: It's already within try, along with "JournalReader reader". What do you mean combine?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +212,467 @@ private static void usage() {
...
+     * @param outputDir output dir for journal dump
+     * @param inputDir [optional] input dir for journal files
+     */
+    public RaftJournalDumper(String master, long start, long end, String outputDir,
+        @Nullable String inputDir) throws IOException {
+      super(master, start, end, outputDir, inputDir);
+      /*
+       * We reset dump time whenever, - Writing to a snapshot is complete. - New JournalEntry
+       * command is written. So, starting the last dump time from snapshot wait time will ensure,
+       * the tool wait at most COPYCAT_SNAPSHOT_READ_TIMEOUT_SECONDS for snapshot, if there is any
```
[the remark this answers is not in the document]
ggezer: Rephrased the comment. Let me know if it's still vague.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
ggezer: We'll still get latest snapshot, then entries (if snapshot is not complete). 
However, we'll keep getting snapshots after then. Basically, the passive node follows follower node's internal state, so as the follower takes a snapshot, it's sent to passive node.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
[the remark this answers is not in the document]
LuQQiu: put them in one line instead of two

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
LuQQiu: Got it, thanks. Then by getting the first snapshot + entries, we are able to get the full state.

ggezer:
@bf8086 PTAL as secondary.

ggezer:
Jenkins, test this please

tests/src/test/java/alluxio/client/cli/JournalToolTest.java
```diff
@@ -112,23 +196,114 @@ public void dumpHeapCheckpoint() throws Exception {
...
+      mFs.createFile(new AlluxioURI("/" + i)).close();
+    }
+
+    // Wait until current snapshot index changes.
+    CommonUtils.waitFor("copycat checkpoint to be written", () -> {
+      try {
+        long latestSnapshotIdx = getCurrentCopyCatSnapshotIndex(followerJournalFolder);
+        return latestSnapshotIdx != curentSnapshotIdx;
+      } catch (Throwable err) {
+        return false;
```
bf8086: Are the errors expected? If not, shall we throw the error to bail out early?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +214,469 @@ private static void usage() {
...
+      }
+    }
+
+    /**
+     * Writes given entry after going through range and validity checks.
+     *
+     * @param out out stream to write the entry to
+     * @param entry the entry to write to
+     */
+    private void writeSelected(PrintStream out, JournalEntry entry) {
```
bf8086: Probably out of scope of this PR - we should consider consolidate methods writing journal entries for both raft journal and ufs journal.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +214,469 @@ private static void usage() {
...
+          return JournalEntryAssociation.getMasterForEntry(entry).equalsIgnoreCase(mMaster);
+        } catch (IllegalStateException e) {
+          return false;
+        }
+      }
+      return false;
+    }
+
+    /**
+     * Implementation of {@link StateMachine} that echoes its state for dumping.
```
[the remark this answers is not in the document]
ggezer: Yes, I'll remove that as the last step. It's defunct for the most cases anyway.

tests/src/test/java/alluxio/client/cli/JournalToolTest.java
ggezer: There could be errors.
Throwing it now.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
ggezer: For UFS it's pretty straightforward because the journal could be selected for a master.
For RAFT, we need to do filtering, because we have the whole journal system coming in the same stream.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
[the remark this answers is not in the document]
ggezer: Removed.

ggezer:
@LuQQiu @bf8086 
There seems to be no stable way for reading embedded journal from a live cluster. CopyCat only allows PASSIVE state for reading state without participating in quorum and that mode gives no timing or consistency guarantees in receiving the cluster state. The snapshots I receive with that mode seems to be a non-committed internal snapshot states of participants. There is an updated copycat library that we are considering for switch, and that supports PROMOTABLE mode that's documented to receive state changes as any other master without participating in quorum. So it's promising but that's a bigger undertaking.

For this PR, I'll leave out the functionality of reading from a live cluster. Currently I ignored the test but I'll go ahead and remove the code from the tool itself and only leave support for reading from disk.

Please let me know WDYT...

ggezer:
@calvinjia PTAL.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -76,16 +94,19 @@
           "The start log sequence number (inclusive). Set to 0 by default.")
       .addOption("end", true,
           "The end log sequence number (exclusive). Set to +inf by default.")
+      .addOption("inputDir", true,
+          "The input directory on-disk to read journal content from. "
+              + "(Default: Read from system configuration.)")
       .addOption("outputDir", true,
-          "The output directory to write journal content to. Default: journal_dump-${timestamp}");
+          "The output directory to write journal content to. "
+          + "(Default: journal_dump-${timestamp})");
```
calvinjia: Is this relative to the directory the command is run from?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -104,101 +125,34 @@ public static void main(String[] args) throws IOException {
...
-          default:
-            throw new RuntimeException("Unknown state: " + state);
-        }
-      }
-    } catch (Exception e) {
-      LOG.error("Failed to read from journal.", e);
+    try {
+      dumpJournal();
+    } catch (Exception exc) {
+      System.err.println(exc);
```
calvinjia: Will this be intuitive to use, if run from the command line usually stderr is not tracked?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +194,299 @@ private static void usage() {
         "Read an Alluxio journal and write it to a directory in a human-readable format.", OPTIONS,
         "", true);
   }
+
+  /**
+   * An abstract class for journal dumpers.
+   */
+  abstract static class AbstractJournalDumper {
```
calvinjia: The JournalTool class is quite long, should we separate these out to their own classes?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
ggezer: Dumping to std stream now.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
ggezer: Correct.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +194,299 @@ private static void usage() {
...
+        Path checkpointPath = path.resolve(entry.getName().toString());
+        LOG.debug("Reading checkpoint for %s to %s%n", entry.getName(), checkpointPath);
+        readCheckpoint(entry.getStream(), checkpointPath);
+      }
+    }
+
+    private void readRocksCheckpoint(CheckpointInputStream checkpoint, Path path)
+        throws IOException {
+      TarballCheckpointReader reader = new TarballCheckpointReader(checkpoint);
+      reader.unpackToDirectory(path);
```
calvinjia: Does this output human readable data?

ggezer: No. We might need another tool for dumping RocksDB content.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
```diff
@@ -239,4 +194,299 @@ private static void usage() {
...
+      }
+    }
+
+    private void readCopycatLogFromDir() {
+      try (PrintStream out =
+          new PrintStream(new BufferedOutputStream(new FileOutputStream(mJournalEntryFile)))) {
+        Log log = Storage.builder().withDirectory(mInputDir).build().openLog("copycat");
+        for (long i = log.firstIndex(); i < log.lastIndex(); i++) {
+          io.atomix.copycat.server.storage.entry.Entry entry = log.get(i);
+          if (entry instanceof CommandEntry) {
```
calvinjia: Are command entries the only ones we care about and are all command entries our journal entries that we want to output?

ggezer: `writeSelected()` does the filtering for edit logs.
For snapshots though we can't do filtering. (Or at least it's not easy)

ggezer:
@calvinjia Updated the PR. PTAL. Thanks.

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
calvinjia: okay, this won't be addressed in this PR?

calvinjia:
@ggezer Thanks, could you see if the test failures are related (seems at least one of them is)?

core/server/common/src/main/java/alluxio/master/journal/JournalTool.java
ggezer: No. That's out of scope for this PR.

ggezer:
Jenkins, test this please

calvinjia:

calvinjia:

PR #9048, opened 2019-05-10
Add inode paths to journal

bf8086:
Add inode path to file create, delete and rename journal entries. This enables path related operations based on only journal entries.


bf8086:
@gpang PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -703,7 +706,8 @@ public InodeDirectory getRoot() {
         newDir.setInternalAcl(pair.getFirst());
         newDir.setDefaultACL(pair.getSecond());
       }
-      mState.applyAndJournal(rpcContext, newDir);
+      mState.applyAndJournal(rpcContext, newDir,
+          inodePath.getUri().getLeadingPath(k));
```
gpang: is there a way to avoid `getLeadingPath`? That method re-parses the path, which could be expensive.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -293,11 +293,13 @@ public void applyAndJournal(Supplier<JournalContext> context, UpdateInodeFileEnt
    *
    * @param context journal context supplier
    * @param inode an inode to add and create a journal entry for
+   * @param path path of the new inode
```
gpang: can `path` ever be `null` (optional)?

core/server/master/src/test/java/alluxio/master/journal/sink/JournalSinkTest.java
```diff
@@ -240,6 +314,29 @@ private long findRename(String newName) throws Exception {
...
+        return entry.getRename().getId();
+      }
+    }
+    return INVALID_ID;
+  }
+
+  private long findCompleteFile(long id, String path) throws Exception {
+    while (!mEntries.isEmpty()) {
+      JournalEntry entry = mEntries.poll();
+      if (entry.hasUpdateInodeFile() && entry.getUpdateInodeFile().getId() == id
```
gpang: shouldn't this check for the `isCompleted` flag to be set and `true`?

core/server/master/src/test/java/alluxio/master/journal/sink/JournalSinkTest.java
```diff
@@ -220,6 +272,17 @@ private long findFile(String name) throws Exception {
     return INVALID_ID;
   }
 
+  private long findFile(String name, String path) throws Exception {
```
gpang: these methods with the path is strictly more specific than the existing ones, right? Should we just replace the existing tests to use these more specific methods?

core/server/master/src/test/java/alluxio/master/journal/sink/JournalSinkTest.java
```diff
@@ -210,6 +210,58 @@ public void replayEvents() throws Exception {
         standbyEntries);
   }
 
+  @Test
+  public void writeInodePaths() throws Exception {
+    String path = "/file";
+    createFile(path);
+    final long createId1 = findFile("file", path);
+    assertNotEquals(-1, createId1);
```
gpang: can we use `INVALID_ID` throughout?

core/server/master/src/test/java/alluxio/master/journal/sink/JournalSinkTest.java
bf8086: I would like to differentiate tests for sink interface and tests for path population so it is easy to pinpoint the issue when some test fails.

bf8086: Using the new method in the existing sink tests means that both sink tests and the new path tests will fail if the path population logic breaks, which makes it hard to pinpoint the root cause.

core/server/master/src/test/java/alluxio/master/journal/sink/JournalSinkTest.java
bf8086: Done.

core/server/master/src/test/java/alluxio/master/journal/sink/JournalSinkTest.java
bf8086: Done.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
bf8086: It should not be `null` - All new inode entry (not from checkpoint) will contain a path.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
bf8086: Refactored.

bf8086:
Thanks @gpang. I updated the PR to address your comments. PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -706,8 +707,10 @@ public InodeDirectory getRoot() {
         newDir.setInternalAcl(pair.getFirst());
         newDir.setDefaultACL(pair.getSecond());
       }
+      String newDirPath = k == 0 ? ROOT_PATH
```
gpang: The loop starts `k` from `pathIndex`. That doesn't have to start from `0`, right? If it starts from a non-zero index, does this still work?

bf8086: Good catch! Fixed.

bf8086:
Thanks @gpang! I updated the PR to address your comment. PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
gpang: Should the creation of `pathBuilder` take care of the `k==0` case, and here we can just call `toString` on the builder?

bf8086: To simplify the logic I am keeping the `pathBuilder` free from the trailing slash. Therefore I have to special case the `k=0` situation where it has to output `/` but the `pathBuilder` cannot contains trailing slash.

bf8086:
@gpang PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
gpang: SG

bf8086:
> @bf8086 Do the tests exercise the path building part in `InodeTree`? I feel like that is the trickiest part.
@gpang Yeah they do I added some tests for k starting from both zero and non-zero cases.

gpang:

PR #9049, opened 2019-05-10
Fix ping service client

bf8086:
This PR fixes two issues:

- Alluxio client can use ping service to query the health of Alluxio services. On client the authentication was incorrectly enabled by default causing it to fail to start when there is no client credentials given.

- Another problem is that the client fails to retry because the gRPC `UnavailableException` is not converted to the `AlluxioStatusException`, effectively render the retry logic invalid.


bf8086:
@ggezer PTAL. Thanks!

ggezer:
LGTM. Thanks for fixing it.

bf8086:
@gpang PTAL. Thanks!

gpang:

PR #9052, opened 2019-05-10
Add to be persisted files to pinned file list

calvinjia:
Changes the semantic of `getPinIdList` to return both the explicitly pinned inodes and inodes which are in `TO_BE_PERSISTED` state. Inodes in that state should not be evicted or users can experience data loss.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -2227,6 +2227,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
 
   @Override
   public Set<Long> getPinIdList() {
+    // return both the explicitly pinned inodes and not persisted inodes which should not be evicted
     return Sets.union(mInodeTree.getPinIdSet(), mInodeTree.getToBePersistedIds());
   }
 
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -2227,7 +2227,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
 
   @Override
   public Set<Long> getPinIdList() {
-    return mInodeTree.getPinIdSet();
+    return Sets.union(mInodeTree.getPinIdSet(), mInodeTree.getToBePersistedIds());
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2227,7 +2227,7 @@ public AlluxioURI getPath(long fileId) throws FileDoesNotExistException {
 
   @Override
   public Set<Long> getPinIdList() {
-    return mInodeTree.getPinIdSet();
+    return Sets.union(mInodeTree.getPinIdSet(), mInodeTree.getToBePersistedIds());
```
apc999: shall we unpinn these files after they are persisted?

madanadit: nit: 
1. can we rename the method or update the comments in `BlockMetadataManagerView:isBlockEvictable` to make it clear that we the evictable check checks for both pinned and to_be_persisted state?
2. ideally the check for free and evict should be the same? in free we can only free PERSISTED files. in evict we allow evicting NOT_PERSISTED files.

✔ ~/os-fork/alluxio [cheng-chang-config-prop-test L|✚ 2]
10:35 $ afs -Dalluxio.user.file.writetype.default=MUST_CACHE copyFromLocal LICENSE /
Copied file:///Users/adit.madan/os-fork/alluxio/LICENSE to /
✔ ~/os-fork/alluxio [cheng-chang-config-prop-test L|✚ 2]
-rw-r--r--  adit.madan     staff                    27040   NOT_PERSISTED 05-10-2019 10:35:44:578 100% /LICENSE
✔ ~/os-fork/alluxio [cheng-chang-config-prop-test L|✚ 2]
10:35 $ afs free /LICENSE
Cannot free file /LICENSE which is not persisted

madanadit: we never pin, so unpinning is not required

calvinjia: `InodeTree` handles adding and removing inodes from the set of `ToBePersisted` inodes. This call is used for a snapshot of inodes which should be considered as unevictable.

calvinjia: Updated, thanks.

apc999: but workers don't know the difference. can worker keep them indefinitely?

calvinjia:
Jenkins, test this please.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: The worker periodically polls for the current set of pinned inodes, it does not store the state across calls to `getPinIdList`

apc999: please document this here as the correctness of returning the union of pinned file ids and tobepersisted ids replying on some assumptions externally. Also it looks "confusing" too here to return the union.

calvinjia:

PR #9056, opened 2019-05-10
Update generated proto to include MetricType enum comments

ZacBlanco:
When generating the proto files from commit 
042f956bb055db9c0fadeb886d643eac9f95725b
They were not all generated with the latest updates.
This commit regenerates the proto files. It only adds
some comments to the enum MetricType which was
created during from that commit.


calvinjia:

PR #9057, opened 2019-05-10
Add extended attributes to inodes

ZacBlanco:
This change will introduce a new field on our inode termed
`xAttr` which is an abbreviation for "Extended Attributes". This
is a `map<string, string>` structure can be utilized in various
ways to add more information to inodes past the standard set
of fields. It gives much more flexibility when adding new features
which may not be used on all inodes (i.e. pinning, TTL). This can
increase the space efficiency of inodes, and possibly decrease
heap usage if utilized only when necessary.

This change will not expose the `xattr` to clients yet, however it
could be in the future. For now the `xattr` kv-pairs are solely for
use between the FileSystemMaster and the Ufs Modules


ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/meta/ExtendedAttribute.java
```diff
@@ -0,0 +1,73 @@
...
+    return String.join(SEPARATOR, components);
+  }
+
+  @Override
+  public String toString() {
+    return mName;
+  }
+
+  private enum NamespacePrefix {
+    SYSTEM("s"),
```
gpang: What does system mean? Shouldn't there be a "ufs" prefix? System seems to vague, since the system could be master, worker, BlockMaster, etc.

core/server/master/src/main/java/alluxio/master/file/meta/MutableInode.java
```diff
@@ -70,6 +73,7 @@ protected MutableInode(long id, boolean isDirectory) {
     mPinned = false;
     mAcl = new AccessControlList();
     mUfsFingerprint = Constants.INVALID_UFS_FINGERPRINT;
+    mXAttr = new HashMap<>(1);
```
gpang: we may want to consider not creating the object if it is unused, since java map objects take up some mem.

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeDirectory.java
```diff
@@ -111,6 +111,8 @@ public MutableInodeDirectory setDefaultACL(DefaultAccessControlList acl) {
   /**
    * Generates client file info for a folder.
    *
+   * xAttr is intentionally left out as the information is not yet exposed to clients
```
gpang: Would it make sense to support the read path of the xattrs, meaning, we expose the map in the file info. We can always add the "write" portion of the API at a later time. This way, if a ufs wants to set values in xattrs, clients/cli can still view the values.

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeDirectory.java
```diff
@@ -200,6 +202,9 @@ public static MutableInodeDirectory fromJournalEntry(InodeDirectoryEntry entry)
     } else {
       ret.mDefaultAcl = new DefaultAccessControlList();
     }
+    if (entry.getXAttrCount() > 0) {
+      ret.setXAttr(entry.getXAttrMap());
```
gpang: Do we not need to update `toJournalEntry` and `toProto`, like the file counterpart?

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeFile.java
```diff
@@ -394,6 +395,8 @@ public static MutableInodeFile fromJournalEntry(InodeFileEntry entry) {
   /**
    * Creates an {@link MutableInodeFile}.
    *
+   * Setting xAttributes is not yet supported on clients. It is intentionally left out.
```
gpang: Similar question, from the directory counterpart. (implement read-only portion)

core/common/src/main/java/alluxio/master/file/meta/PersistenceState.java
```diff
@@ -22,4 +25,59 @@
...
   PERSISTED, // the file is persisted in the under FS
   LOST // the file is lost but not persisted in the under FS
+  ;
+
+  /**
+   * Encode a state into a single byte.
+   *
+   * @return the state encoded as a byte
+   */
+  public byte encode() {
```
gpang: Is all this new code for "replacing" the current persistence state? I don't think we are ready to replace that yet?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2316,6 +2316,7 @@ private void loadMetadataInternal(RpcContext rpcContext, LockedInodePath inodePa
           } else {
             listOptions.setRecursive(false);
           }
+          listOptions.setXAttr(inodePath.getInode().getXAttr());
```
gpang: Why aren't the other UFS calls updated to set the xattrs?

core/base/src/main/java/alluxio/underfs/UfsStatus.java
```diff
@@ -42,15 +46,17 @@
...
   protected UfsStatus(String name, boolean isDirectory, String owner, String group, short mode,
-      Long lastModifiedTimeMs) {
+      Long lastModifiedTimeMs, @Nullable Map<String, String> xAttrs) {
     mIsDirectory = isDirectory;
     mName = name;
     mOwner = owner;
     mGroup = group;
     mMode = mode;
     mLastModifiedTimeMs = lastModifiedTimeMs;
+    mXAttr = xAttrs == null ? new HashMap<>(0) : xAttrs;
```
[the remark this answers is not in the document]
ZacBlanco: So I'm seeing two options

1. Make the internal nullable if none are set, upon get, check for null, return an empty map if it is null.
2. Make the internal nullable, return null when `getXattr()`.

1. can avoid potential NPEs, but may still result in memory usage
2. Will avoid extra memory usage, but introduce possibility of NPEs if we are not careful

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeDirectory.java
ZacBlanco: I may have missed that...

core/common/src/main/java/alluxio/master/file/meta/PersistenceState.java
ZacBlanco: So it does not need to be included in this PR, but I gave it as a sample of how we could encode the state in a small portion of memory

core/server/master/src/main/java/alluxio/master/file/meta/ExtendedAttribute.java
ZacBlanco: I wanted to keep prefixes short because longer names will take more bytes. Either way, This `ExtendedAttribute` enum is more or less provided as an example of how we can define properties that we use within extended attributes. I can remove the examples before the PR gets merged and add more javadoc instead.

core/common/src/main/java/alluxio/underfs/options/CreateOptions.java
```diff
@@ -60,6 +64,7 @@ private CreateOptions(String authUmask) {
     mOwner = null;
     mGroup = null;
     mMode = ModeUtils.applyFileUMask(Mode.defaults(), authUmask);
+    mXAttr = new HashMap<>();
```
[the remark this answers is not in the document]
ZacBlanco: Ah I missed the constructor here. I was trying to set the default size of the map to `1` so that a minimal amount of memory is used. I can make this `Nullable` similar to the others though. I'm just wary of NPEs

core/base/src/main/java/alluxio/underfs/UfsStatus.java
```diff
@@ -152,9 +159,17 @@ public String getOwner() {
     return mOwner;
   }
 
+  /**
+   * Returns the extended attributes from the Ufs, if any.
+   * @return a map of the extended attributes. If none, the map will be empty
+   */
+  public Map<String, String> getXAttr() {
+    return mXAttr;
```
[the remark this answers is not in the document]
ZacBlanco: Should we make this map immutable then?

core/server/master/src/main/java/alluxio/master/file/meta/MutableInode.java
ZacBlanco: I actually think that is a good idea, however I'm concerned about users getting back a null value when calling `getXAttr()`. If it is null, could we return an empty map there? or do you think we should just return `null`?

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeDirectory.java
ZacBlanco: For the time being, if we don't plan to expose the extended attributes to users to write, I don't think they should be able to read them. I think it keeps the API cleaner and less confusing. If clients can't set them, why do they need to be read? If you have a strong use case for clients reading UFS attributes now I could be convinced to add it.

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeDirectory.java
ZacBlanco: good catch!

core/server/master/src/main/java/alluxio/master/file/meta/ExtendedAttribute.java
```diff
@@ -0,0 +1,73 @@
...
+public enum ExtendedAttribute {
+
+  PERSISTENCE_STATE("ps"), // system (ufs) property
+  PIN_STATE(NamespacePrefix.USER, "pin"), // property set by user
+  ;
+
+  private static final String SEPARATOR = ".";
+
+  private final String mName;
+
```
bf8086: Can we include encoding / decoding information so we can automate the value conversion?

core/base/src/main/java/alluxio/underfs/UfsStatus.java
```diff
@@ -152,9 +159,17 @@ public String getOwner() {
     return mOwner;
   }
 
+  /**
+   * Returns the extended attributes from the Ufs, if any.
+   * @return a map of the extended attributes. If none, the map will be empty
+   */
+  public Map<String, String> getXAttr() {
```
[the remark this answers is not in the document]
ZacBlanco: I am for using `byte[]` instead of string. I did some digging through the man pages (http://man7.org/linux/man-pages/man7/xattr.7.html and http://man7.org/linux/man-pages/man2/getxattr.2.html) and I didn't find anything specifying it had to be a string. I will convert to using `byte[]`

core/server/master/src/main/java/alluxio/master/file/meta/ExtendedAttribute.java
ZacBlanco: Actually, that's not a bad idea. Instead of an enum, maybe I can create an `ExtendedAttribute` interface where each implementation defines the identifier, namespace, and encode/decode functions

core/base/src/main/java/alluxio/underfs/UfsStatus.java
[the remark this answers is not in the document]
ZacBlanco: Actually, I think we should want to be careful using `byte[]`. Proto's default representation of `bytes` in Java is as a `ByteString`. Unless we want to actively convert proto values, this will force inodes to depend on this proto-specific class, unless we perform conversion to `byte[]` actively whenever converting to/from proto. That could result in an increase in memory usage copying underlying byte arrays back and forth.

core/base/src/main/java/alluxio/underfs/UfsStatus.java
```diff
@@ -33,6 +36,7 @@
   protected final String mGroup;
   protected final short mMode;
 
+  protected final Map<String, String> mXAttr;
```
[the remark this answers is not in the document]
ZacBlanco: done

core/base/src/main/java/alluxio/underfs/UfsStatus.java
```diff
@@ -65,6 +71,7 @@ protected UfsStatus(UfsStatus status) {
     mGroup = status.mGroup;
     mMode = status.mMode;
     mLastModifiedTimeMs = status.mLastModifiedTimeMs;
+    mXAttr = status.mXAttr;
```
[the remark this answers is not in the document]
ZacBlanco: fixed

core/common/src/main/java/alluxio/underfs/UfsDirectoryStatus.java
```diff
@@ -11,23 +11,29 @@
...
    * @param owner of the directory
    * @param group of the directory
    * @param mode of the directory
+   * @param lastModifiedTimeMs of the directory
+   * @param xAttr extended attributes, if any
    */
-  public UfsDirectoryStatus(String name, String owner, String group, short mode) {
-    super(name, true, owner, group, mode, null);
+  public UfsDirectoryStatus(String name, String owner, String group, short mode,
+      long lastModifiedTimeMs, Map<String, String> xAttr) {
```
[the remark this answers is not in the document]
ZacBlanco: done

core/common/src/main/java/alluxio/underfs/UfsFileStatus.java
```diff
@@ -33,14 +35,31 @@
    * @param owner of the file
    * @param group of the file
    * @param mode of the file
+   * @param xAttr extended attributes, if any
    */
   public UfsFileStatus(String name, String contentHash, long contentLength, long lastModifiedTimeMs,
-      String owner, String group, short mode) {
-    super(name, false, owner, group, mode, lastModifiedTimeMs);
+      String owner, String group, short mode, Map<String, String> xAttr) {
```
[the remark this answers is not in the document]
ZacBlanco: done

core/base/src/main/java/alluxio/underfs/UfsStatus.java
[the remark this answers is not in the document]
bf8086: We can store `byte[]` in inode and convert to/from `ByteString` when needed. 
The conversion between `ByteString` and `byte[]` won't be more expensive than the conversion between `String` and `byte[]` since creating a string involves copying byte arrays as well.

core/server/master/src/main/java/alluxio/master/file/meta/ExtendedAttribute.java
bf8086: Yeah that sounds good to me.

core/server/master/src/main/java/alluxio/master/file/meta/xattr/ExtendedAttribute.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  T decode(ByteString bytes) throws IOException;
+
+  /**
+   * Encode a list of objects into a byte string.
+   *
+   * @param objects the objects to encode
+   * @return a bytestring with a size equal to {@link #getEncodedSize()} * {@code objects.size()}
+   */
+  ByteString multiEncode(List<T> objects);
```
bf8086: Do we need to differentiate between single encode/decode and multi encode/decode? As a user I would assume just getting the correct value from extended attributes by calling ExtendedAttribute.MY_ATTRIBUTE.decode(some_bytes), regardless of the attribute value being a list or not.

bf8086: Actually can we require attribute values to be a protobuf type? That will make it consistent with everywhere else and save the work to write the decode/encode function.

ZacBlanco: In my opinion, it is useful to have an explicit implementation for encode/decode because we can get tight control over exactly how many bytes are stored when an object is serialized. I don't think these attributes are intended for complex values, but I think having tighter control over the individual bytes is beneficial.

Additionally, there's nothing stopping someone from adding a proto message and simply encoding an object as the proto message and then calling `toBytes()` for encode and then using `Message.parseFrom(byte[])` in the decode function.

I tested using a simple message 

```proto
enum PersistencePState {
  NOT_PERSISTED = 1;
  TO_BE_PERSISTED = 2;
  PERSISTED = 3;
  LOST = 4;
}

message PersistenceStateAttribute {
  required PersistencePState state = 1;
}
```

In this example, when on the wire, the field `state` requires two bytes, where in fact, it should only need 1. If the field is `repeated` instead of required, the usage is still two bytes/value

For more advanced cases, we can utilize proto, but if we can encode in a smaller number of bytes than proto can, is there a reason we can't implement our own?

Lastly, I implemented multiEncode because it's possible you want to store an array/list of some kind of value as opposed to a single one. The caveat is that it requires knowledge of the size of the message before (that's why the encoding size is a parameter in the constructor).

I can see the tradeoffs of ease of use vs better memory utilization. I think we just need to decide between them. My vote is to allow our own encoding, but still support proto (the way it is now). Is there a downside to supporting both? Also, I gave persistence state as more of an example implementation than something that actually needs to be committed. So this doesn't have to necessarily be the final implementation.

core/base/src/main/java/alluxio/underfs/UfsStatus.java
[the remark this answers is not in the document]
bf8086: Do we still want to update this class? If we do, should this now be `Map<String, byte[]>`?

core/server/master/src/main/java/alluxio/master/file/meta/xattr/AbstractExtendedAttribute.java
```diff
@@ -0,0 +1,75 @@
...
+   * Builds an attribute by joining namespace components together.
+   *
+   * @param components the components to join
+   * @return the attribute joined by {@link #SEPARATOR}
+   */
+  private static String buildAttr(String... components) {
+    return String.join(SEPARATOR, components);
+  }
+
+  enum NamespacePrefix {
```
bf8086: Is this the same as the `NamespacePrefix` in `ExtendedAttribute.java`?

core/server/master/src/main/java/alluxio/master/file/meta/ExtendedAttribute.java
```diff
@@ -0,0 +1,77 @@
...
+ * For example, the persistence state uses the default {@link NamespacePrefix#SYSTEM} prefix and
+ * appends {@code ps} (an acronym for persistence state) so that the attribute when stored is
+ * "{@code s.ps}" which is only a 4-character long string. The reason the prefix and is so short
+ * and attribute identifier are so small is to try to reduce the memory footprint of the
+ * attribute when stored in the inode. Compare this to using the attribute name
+ * "system.persistance_state". The second one is 6x longer and will use at least 6x more memory.
+ * Over a couple hundred million inodes this difference can contribute to a significant
+ * increase in memory usage.
+ */
+public enum ExtendedAttribute {
```
bf8086: Do we still need this now there is a `ExtendedAttribute` interface?

core/transport/src/proto/journal/file.proto
```diff
@@ -95,7 +95,8 @@ message UpdateInodeEntry {
   optional PTtlAction ttlAction = 13 [default = DELETE];
   optional alluxio.proto.shared.AccessControlList acl = 14;
   optional string ufs_fingerprint = 15;
-  repeated string medium_type = 16;
+  repeated string medium_type = 17;
```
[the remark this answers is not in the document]
ZacBlanco: whoops, mistake on my part

core/server/master/src/main/java/alluxio/master/file/meta/xattr/AbstractExtendedAttribute.java
bf8086: Should we make this enum public and put it in `ExtendedAttribute`? Do you think it would be useful for other classes to use it?

ZacBlanco: I don't think other classes will need to use it. But I think it is still useful to be in `ExtendedAttribute` rather than `AbstractExtendedAttribute`

bf8086: I see if we don't need to use it outside of ExtendedAttributes and derived classes, then we can leave it here.

core/common/src/main/java/alluxio/master/file/meta/xattr/ExtendedAttribute.java
```diff
@@ -0,0 +1,81 @@
...
+   * @throws IOException if the size of the byte string isn't equal to the encoding length
+   */
+  T decode(byte[] bytes) throws IOException;
+
+  /**
+   * The namespace for extended attributes.
+   *
+   * The values provided in this enum are the standard namespaces for all {@code xattr}
+   */
+  enum NamespacePrefix {
```
gpang: no "ufs" prefix?

ZacBlanco: Namespaces are standardized: http://man7.org/linux/man-pages/man7/xattr.7.html

calvinjia:

PR #9060, opened 2019-05-11

ScorpionFray:
Use static imports for standard test utilities in /alluxio/core/common/src/test/java/alluxio/util/network/GetMasterWorkerAddressTest.java

Change
import org.junit.Assert;
to
Put the static imports before non-static imports.

AND

Update all
Assert.assertEquals*
to
assertEquals*


apc999:

PR #9061, opened 2019-05-11

JimZong:
change import org.junit.Assert; to import org.junit.Assert.fail;
modify all assert.fail(* to fail(*
Put the static imports before non-static imports.


JimZong:

apc999:

PR #9062, opened 2019-05-11
Add option for skipping caching and eviction for scanning inode tree

cheng-chang:
Scanning all a large portion of the inode tree can thrash metastore when using the CachingInodeStore. Frequent eviction would negatively impact the performance of the cache.

This PR introduces an option to disable caching and eviction when getting inodes from CachingInodeStore.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/metastore/ReadOnlyInodeStore.java
+++ b/core/server/master/src/main/java/alluxio/master/metastore/ReadOnlyInodeStore.java
@@ -28,29 +28,114 @@ import java.util.Set;
  * Read-only access to the inode store.
  */
 public interface ReadOnlyInodeStore extends Closeable {
+  /**
+   * Options of reading from the inode store.
+   */
+  class ReadOption {
+    private static final ReadOption DEFAULT = new ReadOption(false);
+
+    private final boolean mSkipCachingAndEviction;
+
+    private ReadOption(boolean skipCachingAndEviction) {
+      mSkipCachingAndEviction = skipCachingAndEviction;
+    }
+
+    /**
+     * @return whether to skip the caching and eviction when reading from the inode store
+     */
+    public boolean shouldSkipCachingAndEviction() {
+      return mSkipCachingAndEviction;
+    }
+
+    /**
+     * @return a new builder
+     */
+    public static Builder newBuilder() {
+      return new Builder();
+    }
+
+    /**
+     * @return the singleton instance of the default option
+     */
+    public static ReadOption defaults() {
+      return DEFAULT;
+    }
+
+    /**
+     * Builder for {@link ReadOption}.
+     */
+    public static class Builder {
+      private boolean mSkipCachingAndEviction = false;
+
+      /**
+       * Sets whether to skip caching and eviction.
+       *
+       * @param skip skip or not
+       * @return the builder
+       */
+      public Builder setSkipCachingAndEviction(boolean skip) {
+        mSkipCachingAndEviction = skip;
+        return this;
+      }
+
+      /**
+       * @return the built option
+       */
+      public ReadOption build() {
+        return new ReadOption(mSkipCachingAndEviction);
+      }
+    }
+  }
 
   /**
    * @param id an inode id
+   * @param option the options
    * @return the inode with the given id, if it exists
    */
-  Optional<Inode> get(long id);
+  Optional<Inode> get(long id, ReadOption option);
+
+  /**
+   * @param id an inode id
+   * @return the result of {@link #get(long, ReadOption)} with default option
+   */
+  default Optional<Inode> get(long id) {
+    return get(id, ReadOption.defaults());
+  }
 
   /**
    * Returns an iterable for the ids of the children of the given directory.
    *
    * @param inodeId an inode id to list child ids for
+   * @param option the options
    * @return the child ids iterable
    */
-  Iterable<Long> getChildIds(Long inodeId);
+  Iterable<Long> getChildIds(Long inodeId, ReadOption option);
+
+  /**
+   * @param inodeId an inode id to list child ids for
+   * @return the result of {@link #getChildIds(Long, ReadOption)} with default option
+   */
+  default Iterable<Long> getChildIds(Long inodeId) {
+    return getChildIds(inodeId, ReadOption.defaults());
+  }
 
   /**
    * Returns an iterable for the ids of the children of the given directory.
    *
    * @param inode the inode to list child ids for
+   * @param option the options
    * @return the child ids iterable
    */
+  default Iterable<Long> getChildIds(InodeDirectoryView inode, ReadOption option) {
+    return getChildIds(inode.getId(), option);
+  }
+
+  /**
+   * @param inode the inode to list child ids for
+   * @return the result of {@link #getChildIds(InodeDirectoryView, ReadOption)} with default option
+   */
   default Iterable<Long> getChildIds(InodeDirectoryView inode) {
-    return getChildIds(inode.getId());
+    return getChildIds(inode, ReadOption.defaults());
   }
 
   /**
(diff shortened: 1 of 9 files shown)


cheng-chang:
@gpang @ggezer PTAL

core/server/master/src/main/java/alluxio/master/metastore/ReadOnlyInodeStore.java
```diff
@@ -28,29 +28,114 @@
  * Read-only access to the inode store.
  */
 public interface ReadOnlyInodeStore extends Closeable {
+  /**
+   * Options of reading from the inode store.
+   */
+  class ReadOption {
```
gpang: Should we move this out to its own class, since multiple classes use this?

core/server/master/src/main/java/alluxio/master/metastore/ReadOnlyInodeStore.java
```diff
@@ -28,29 +28,114 @@
...
+    private static final ReadOption DEFAULT = new ReadOption(false);
+
+    private final boolean mSkipCachingAndEviction;
```
gpang: Is there any way to skip the cache but NOT skip eviction? Maybe just `skipCache` is a good enough name?

core/server/master/src/main/java/alluxio/master/metastore/caching/Cache.java
```diff
@@ -126,6 +128,29 @@ public Cache(CacheConfiguration conf, String name) {
...
+  }
+
+  /**
+   * Retrieves a value from the cache if already cached, otherwise, loads from the backing store
+   * without caching the value. Eviction is not triggered.
+   *
+   * @param key the key to get the value for
+   * @return the value, or empty if the key doesn't exist in the cache or in the backing store
+   */
+  private Optional<V> getWithoutCachingAndEviction(K key) {
```
gpang: NIT: maybe `getSkipCache()`?

cheng-chang:
@gpang PTAL

cheng-chang:
jenkins, test this please

LuQQiu:
@cheng-chang How could the end user provide the read option? It looks like we always use `ReadOption.defaults()` and there is no way (property key or add it to `GetStatusContext, ListStatusContext`.

I agree that we can use `skipCache()` name but also skip eviction in the listing process.

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
```diff
@@ -691,10 +697,13 @@ public void clear() {
...
-      evictIfNecessary();
+    private Map<String, Long> loadChildren(Long inodeId, ListingCacheEntry entry,
+        ReadOption option) {
+      if (!option.shouldSkipCache()) {
+        evictIfNecessary();
+      }
       entry.mModified = false;
-      Map<String, Long> listing = mEdgeCache.getChildIds(inodeId);
+      Map<String, Long> listing = mEdgeCache.getChildIds(inodeId, option);
       mMap.computeIfPresent(inodeId, (key, value) -> {
```
LuQQiu: Do we also want to skip caching in `getChildIds`?  inside the `mMap.computeIfPresent` is the place where we cache the listing results in the`mMap of ListingCache`.

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
```diff
@@ -680,9 +686,9 @@ public void removeEdge(Edge edge) {
       }
       if (entry == null || !createdNewEntry.get()) {
         // Skip caching if the cache is full or someone else is already caching.
-        return mEdgeCache.getChildIds(inodeId).values();
+        return mEdgeCache.getChildIds(inodeId, option).values();
```
LuQQiu: If we want to skip caching in `getChildIds`, we should add the 
```
if (entry == null || !createdNewEntry.get() || option.shouldSkipCache()) {
return mEdgeCache.getChildIds(inodeId, option).values();
}
```
so that we don't walk to the `loadChildren` caching place.

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
```diff
@@ -659,10 +662,13 @@ public void removeEdge(Edge edge) {
      * Gets all children of an inode, falling back on the edge cache if the listing isn't cached.
      *
      * @param inodeId the inode directory id
+     * @param option the read options
      * @return the ids of all children of the directory
      */
-    public Collection<Long> getChildIds(Long inodeId) {
-      evictIfNecessary();
+    public Collection<Long> getChildIds(Long inodeId, ReadOption option) {
```
ggezer: They are most likely entries that are loaded with shouldSkipCache set to false. A read operation suppressing eviction for other entries is not safe as in, we could end up with full cache. I think we shouldn't interfere with the eviction of entries that are already in the cache, and just avoiding cache should be good enough. WDYT?

core/server/master/src/main/java/alluxio/master/metastore/caching/Cache.java
```diff
@@ -94,16 +95,16 @@ public Cache(CacheConfiguration conf, String name) {
    * If the value needs to be loaded, concurrent calls to get(key) will block while waiting for the
    * first call to finish loading the value.
    *
+   * If skipCachingAndEviction is true, then the value loaded from the backing store will not be
```
ggezer: If skipCachingAndEviction is true -> if options.shouldSkipCache() is true

cheng-chang:
@LuQQiu The convention is for each read method, there are two overrided methods, one with the explicit `ReadOption`, another with the implicit default ReadOption. E.g. https://github.com/Alluxio/alluxio/pull/9062/files#diff-1dc1130da7795efb313cea4bdd361d03L39

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
cheng-chang: yes

cheng-chang:
@ggezer this is just the infra

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
cheng-chang: The idea of `skipCache` is to not interfere with the caching and eviction behavior between cache and backing store. For operations with `skipCache=false`, they will evict the cache if it's full. Think of it this way:  the caching inode store works normally for operations other than scan, for scan, it will neither cache nor evict, so scan a caching inode store will have no effect on the caching status of the store.

cheng-chang:
@LuQQiu @ggezer thanks for comments, PTAL.

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
ggezer: During scanning the nodes that are evicted are brought by it, hence the trashing. I'm in for keeping the logic simple and only skip caching. That will already prevent trashing since cache won't be filled by scanning and eviction won't be necessary due to scanning.

cheng-chang: SG, since the eviction thread keeps running in background, the "skip eviction" does not make much sense.

cheng-chang:
@ggezer @LuQQiu PTAL, thanks.

LuQQiu:
@cheng-chang Yeah, there are two overridden methods, but looks like the `DefaultFileSystemMaster` only call the one with default ReadOption. Any places call the read method with `ReadOption and skipCacahe = true` ?

core/server/master/src/main/java/alluxio/master/metastore/caching/CachingInodeStore.java
ggezer: Eviction here is in-memory. So no effect on trashing as well.

cheng-chang:
@LuQQiu yes, that's for now, when the `scan` API is added, it will use ReadOption with skipCache=true.

cheng-chang:

PR #9068, opened 2019-05-12
[ALLUXIO-9068]update file

niruntao:


naive4E4A55:
change the import rules

PR #9082, opened 2019-05-12
Change the initial persist wait time to user property

LuQQiu:
Change the persist wait time to user property so that users can pass in the value when creating the file. Persist wait time is suggested to set to a big enough value when using WriteType=ASYNC_THOUGH


tests/src/test/java/alluxio/client/cli/JournalToolTest.java
```diff
@@ -65,7 +65,6 @@
           .setProperty(PropertyKey.MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES,
               Integer.toString(CHECKPOINT_SIZE))
           .setProperty(PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX, "100")
-          .setProperty(PropertyKey.MASTER_PERSISTENCE_INITIAL_WAIT_TIME_MS, "1min")
```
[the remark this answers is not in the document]
LuQQiu: I took a look into the test, set the persistence wait time to 0 will not affect the pass rate of this test but reduce the wait time when running `fs persist` command.

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeFile.java
```diff
@@ -423,7 +441,8 @@ public static MutableInodeFile create(long blockContainerId, long parentId, Stri
         .setMode(context.getMode().toShort())
         .setAcl(context.getAcl())
         .setPersistenceState(context.isPersisted() ? PersistenceState.PERSISTED
-            : PersistenceState.NOT_PERSISTED);
+            : PersistenceState.NOT_PERSISTED)
```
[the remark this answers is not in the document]
LuQQiu: This cannot be fit in one line, but it makes sense to remain `.setPersistenceState` as the last one to set. I move the `setPersistenceWaitTime` above.

ZacBlanco:
@LuQQiu lgtm, aside from the fact we may not want to put this information into the inode. Still want to hear thoughts from @calvinjia

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -577,7 +577,7 @@ public void start(Boolean isPrimary) throws IOException {
           mPersistRequests.put(inodeFile.getId(), new alluxio.time.ExponentialTimer(
               ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_INTERVAL_MS),
               ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_INTERVAL_MS),
-              ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_WAIT_TIME_MS),
+              inodeFile.getPersistenceWaitTime(),
```
calvinjia: This calculation needs to be a bit different on recovery since the file should not be treated as created on restart (for example if wait time is 1 day and restart happened 23hrs after file creation, we should not wait 1 more day)

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2009,7 +2010,7 @@ private void renameInternal(RpcContext rpcContext, LockedInodePath srcInodePath,
       mPersistRequests.put(srcInode.getId(), new alluxio.time.ExponentialTimer(
           ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_INTERVAL_MS),
           ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_INTERVAL_MS),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_WAIT_TIME_MS),
+          srcInode.asFile().getPersistenceWaitTime(),
```
calvinjia: This should not obey the file's persist wait time.

For the persist on rename feature, we should expect users to set a persist wait time of `-1` (or some other special value) to indicate that the file will be persisted but should not be automatically persisted by Alluxio (rather through a rename or CLI persist).

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2817,6 +2809,14 @@ public String toString() {
               + "to commit results.")
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_FILE_PERSISTENCE_INITIAL_WAIT_TIME =
```
calvinjia: Should we add an option to our persist cli which let's them specify the wait time?

LuQQiu: I am considering either we can add a persist option like `forcePersist` which will use 0 wait time. Or we directly use 0 wait time. I think when the user call persist, they want to persist as soon as possible.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
LuQQiu: Maybe instead of the persist wait time, I use a `should persist time(create file time + persist wait time)`. In here I can pass in the`(should persist time - current time)`. If it's smaller than 0, than I set 0 wait time.

calvinjia: That sounds reasonable.

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: By default the wait time should be zero, I just meant to provide an option in the CLI so the user does not need to pass something like `-Dalluxio.user.file...`

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
LuQQiu: In the rename context, we have a `context.getPersist`. Only if `Context.getPersist (a persist operation should follow the rename)`, will we trigger the rename. Then here it should also obey the `srcInode.getShouldPersistTime`. 

If users don't want the file be automatically persisted by Alluxio, then they didn't need to pass in the `Context.getPersist` and directly trigger `persist CLI` when they need. Looks like we don't need the special persist wait time special value.

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: I agree, the persist CLI should have a persist wait time option. if it's not provided, we directly use the default initial wait time value.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2002,16 +2007,21 @@ private void renameInternal(RpcContext rpcContext, LockedInodePath srcInodePath,
...
-          .setPersistenceState(PersistenceState.TO_BE_PERSISTED.name())
-          .build());
-      mPersistRequests.put(srcInode.getId(), new alluxio.time.ExponentialTimer(
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_INTERVAL_MS),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_INTERVAL_MS),
-          srcInode.asFile().getPersistenceWaitTime(),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_TOTAL_WAIT_TIME_MS)));
+    if (srcInode.isFile() && !srcInode.isPersisted()) {
+      long shouldPersistTime = srcInode.asFile().getShouldPersistTime();
+      if (context.getPersist() || shouldPersistTime == Constants.NO_AUTO_PERSIST_VALUE) {
```
calvinjia: I think this condition should just be `context.getPersist`, its possible the user does not want to persist on rename.

LuQQiu: In Alluxio's integration with Hive and maybe other compute frameworks, will they be able to provide detail rename options like `context.getPersist` here?

core/base/src/main/java/alluxio/Constants.java
```diff
@@ -194,6 +194,8 @@
   public static final int REPLICATION_MAX_INFINITY = -1;
 
   // Persistence
+  // The file should only be persisted after rename operation or persist CLI
+  public static final int NO_AUTO_PERSIST_VALUE = -1;
```
ZacBlanco: nit: I think `_VALUE` is redundant in naming? Can we just call it `NO_AUTO_PERSIST`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: Yeah we can use the property key alluxio.user.file.persist.on.rename=true to get the desired behavior.

core/client/fs/src/main/java/alluxio/client/file/FileSystemUtils.java
```diff
@@ -143,14 +145,16 @@ public static void persistAndWait(final FileSystem fs, final AlluxioURI uri)
...
    * @param timeoutMs max amount of time to wait for persist in milliseconds. -1 to wait
    *                  indefinitely
    * @throws TimeoutException if the persist takes longer than the timeout
    */
-  public static void persistAndWait(final FileSystem fs, final AlluxioURI uri, int timeoutMs)
-      throws FileDoesNotExistException, IOException, AlluxioException, TimeoutException,
-      InterruptedException {
-    fs.persist(uri);
+  public static void persistAndWait(final FileSystem fs, final AlluxioURI uri,
+      long persistenceWaitTime, int timeoutMs) throws FileDoesNotExistException, IOException,
```
ZacBlanco: To help prevent user error, can we check that `persistenceWaitTime < timeoutMs`? When that condition is true, the persist will always timeout and fail here. It would be useful to fail-fast by checking this condition.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
LuQQiu: Then I am good with only `context.getPersist`, thanks!

core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
```diff
@@ -42,6 +42,17 @@
...
 public class FileSystemOptions {
+  /**
+   * @param conf Alluxio configuration
+   * @return options based on the configuration
+   */
+  public static ScheduleAsyncPersistencePOptions scheduleAsyncPersistenceDefaults(
+      AlluxioConfiguration conf) {
+    return ScheduleAsyncPersistencePOptions.newBuilder()
+        .setCommonOptions(commonDefaults(conf))
+        .setPersistenceWaitTime(0)
```
ZacBlanco: Do we always want this to be `0`? Shouldn't it be set based on the configuration?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -572,12 +573,15 @@ public void start(Boolean isPrimary) throws IOException {
         if (inode.getPersistenceState() != PersistenceState.TO_BE_PERSISTED) {
           continue;
         }
+        if (inode.asFile().getShouldPersistTime() == Constants.NO_AUTO_PERSIST_VALUE) {
```
ZacBlanco: nit: instead of a separate conditional, could we just add an `||` condition to the last statement?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3705,21 +3716,32 @@ public void stopSync(AlluxioURI syncPoint)
...
-  private void addPersistJob(long fileId, long jobId, AlluxioURI uri, String tempUfsPath) {
+  private void addPersistJob(long fileId, long jobId, long persistenceWaitTime, AlluxioURI uri,
+      String tempUfsPath) {
     alluxio.time.ExponentialTimer timer = mPersistRequests.remove(fileId);
     if (timer == null) {
       timer = new alluxio.time.ExponentialTimer(
           ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_INTERVAL_MS),
           ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_MAX_INTERVAL_MS),
-          ServerConfiguration.getMs(PropertyKey.MASTER_PERSISTENCE_INITIAL_WAIT_TIME_MS),
+          persistenceWaitTime,
```
ZacBlanco: I am under the assumption that `MASTER_PERSISTENCE_MAX_TOTAL_WAIT_TIME_MS` represents the total amount of time since adding to the job that we will allow before removing/erroring. If that is the case, then the following comment applies:

Similar to another comment I made, is there a case here where a user could specify `persistenceWaitTime` larger than `MASTER_PERSISTENCE_MAX_TOTAL_WAIT_TIME_MS`? It might be useful to fail-fast in those cases.

shell/src/main/java/alluxio/cli/fs/command/PersistCommand.java
```diff
@@ -68,6 +68,15 @@
               + DEFAULT_TIMEOUT)
           .required(false)
           .build();
+  private static final int DEFAULT_WAIT_TIME = 0;
+  private static final Option WAIT_OPTION =
+      Option.builder("w")
+          .longOpt("wait")
+          .argName("the initial persistence wait time")
+          .numberOfArgs(1)
+          .desc("The time in milliseconds to wait before persisting. default: " + DEFAULT_WAIT_TIME)
```
ZacBlanco: Can we utilize the InstancedConfiguration's `getMs` function (I think it references `CommonUtils.Xyz` something) to parse values like `1min`, `1hour` etc instead of only supporting milliseconds?

core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
LuQQiu: If a user directly call persist, they usually want the file/directory to persist immediately. If not, they can just set the wait time through CLI.

calvinjia:
@LuQQiu Thanks for implementing this. Could you add unit tests for the cases where persist wait time is > 0 and == -1?

LuQQiu:
I tested locally and the behavior is not as expected, 
```
./bin/alluxio fs -Dalluxio.user.file.persistence.initial.wait.time=5s -Dalluxio.user.file.writetype.default=ASYNC_THROUGH copyFromLocal LICENSE /LICENSE
The files are persisted immediately
```
need to look into ASYNC_THROUGH persist to which places cause the issue.

shell/src/main/java/alluxio/cli/fs/command/PersistCommand.java
LuQQiu: Change to allow given time formats. Thanks

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
LuQQiu: I add the comment to `USER_FILE_PERSISTENTCE_INITIAL_WAIT_TIME` to say it's not bigger than `TOTAL_WAIT_TIME`. The default total wait time is 1 day, usually it will not be exceeded.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
LuQQiu: Changed, thanks

core/client/fs/src/main/java/alluxio/client/file/FileSystemUtils.java
LuQQiu: Add the preconditions check here. Thanks

core/base/src/main/java/alluxio/Constants.java
LuQQiu: Removed `_value`

LuQQiu:
@ZacBlanco @calvinjia Addressed the comments and added the tests. Once the master build fail solved, I will resolve the conflicts. PTAL, thanks

LuQQiu:
@calvinjia @ZacBlanco Added the tests and fix the bug. Please take a look, thanks

LuQQiu:
@ZacBlanco Please take a look at the new commits that I addressed your comments and added the tests. Thanks

calvinjia:

PR #9083, opened 2019-05-12
Use JAVA_OPTS for k8s config

madanadit:
Fix spec to use JAVA_OPTS instead of capitalized env variables following https://github.com/Alluxio/alluxio/pull/9040


madanadit:

PR #9084, opened 2019-05-13
Add Pinning to specific tier

yuzhu:
There are several things in this PR.

Change Inode metadata to include a set of strings that are the desired pinning medium for the file. 
Pass a desired pinning storage medium type to loadBlock method of JobUtils.java.
This storage medium type eventually gets passed to the createBlock method on the worker, and worker finds a storage medium which matches the desired storage medium type.
Change the blockInfo in fileInfo to contain medium type information. 

TODO (defer to phase 2):
BlockHeartBeatReport related changes


yuzhu:
jenkins, test this please

yuzhu:
jenkins, test this please

yuzhu:
jenkins, test this please

yuzhu:
@calvinjia PTAL at this one instead of #9019 , thanks.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3453,7 +3453,9 @@ private void setAttributeSingleFile(RpcContext rpcContext, LockedInodePath inode
     Inode inode = inodePath.getInode();
     SetAttributePOptions.Builder protoOptions = context.getOptions();
     if (protoOptions.hasPinned()) {
-      mInodeTree.setPinned(rpcContext, inodePath, context.getOptions().getPinned(), opTimeMs);
+      mInodeTree.setPinned(rpcContext, inodePath, context.getOptions().getPinned(),
+          context.getOptions().getPinnedMediaList(),
+          opTimeMs);
```
calvinjia: Can be on previous line.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -865,25 +865,41 @@ public void deleteInode(RpcContext rpcContext, LockedInodePath inodePath, long o
...
+  private boolean checkPinningValidity(Set<String> pinnedLocations) {
+    List<String> mediaList = ServerConfiguration.getList(
+        PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIA, ",");
+    for (String medium : pinnedLocations) {
+      if (!mediaList.remove(medium)) {
+        // mediaList does not contains medium
+        return false;
+      }
+    }
+    return true;
```
calvinjia: Since `pinnedLocations` is a set, you can check it without using the `remove` call.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -865,25 +865,41 @@ public void deleteInode(RpcContext rpcContext, LockedInodePath inodePath, long o
...
    * @param rpcContext the rpc context
    * @param inodePath the {@link LockedInodePath} to set the pinned state for
    * @param pinned the pinned state to set for the inode (and possible descendants)
+   * @param mediumTypes the list of pinned media that that the file can reside in
    * @param opTimeMs the operation time
    * @throws FileDoesNotExistException if inode does not exist
    */
   public void setPinned(RpcContext rpcContext, LockedInodePath inodePath, boolean pinned,
-      long opTimeMs) throws FileDoesNotExistException, InvalidPathException {
+      List<String> mediumTypes, long opTimeMs)
```
calvinjia: Could `mediumTypes` be of type `Set<String>` instead?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -865,25 +865,41 @@ public void deleteInode(RpcContext rpcContext, LockedInodePath inodePath, long o
...
+      throws FileDoesNotExistException, InvalidPathException {
     Preconditions.checkState(inodePath.getLockPattern().isWrite());
-
+    Set<String> mediumSet = new HashSet<>(mediumTypes);
+    if (!checkPinningValidity(mediumSet)) {
```
calvinjia: Could this be a precondition?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -400,6 +400,19 @@ private void applyUpdateInode(UpdateInodeEntry entry) {
...
       if (entry.getPinned()) {
         MutableInodeFile file = inode.asFile();
+        List<String> mediaList = ServerConfiguration.getList(
+            PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIA, ",");
+        if (entry.getMediumTypeList().isEmpty()) {
+          // if user does not specify a pinned media list, any location is OK
+          file.setMediumTypes(new HashSet<>());
+        } else {
+          for (String medium : entry.getMediumTypeList()) {
+            int index = mediaList.indexOf(medium);
```
calvinjia: Use `contains` here?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -145,7 +145,7 @@ private void check(Set<Long> inodes, ReplicationHandler handler, Mode mode) {
       // file and may increase lock contention in this tree. Investigate if we could avoid
       // locking the entire path but just the inode file since this access is read-only.
       try (LockedInodePath inodePath =
-          mInodeTree.lockFullInodePath(inodeId, LockPattern.READ)) {
+               mInodeTree.lockFullInodePath(inodeId, LockPattern.READ)) {
```
calvinjia: Formatting

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -920,6 +920,25 @@ public void removeDir(StorageDir dir) {
     }
   }
 
+  /**
+   * Get the block store location based on a storage medium.
+   *
+   * @param mediumName storage medium
+   * @return the storage location
+   */
+  public BlockStoreLocation findLocationByMedium(String mediumName) {
```
calvinjia: Can't multiple block store locations provide the same medium type, we should return a set here?

core/server/worker/src/main/java/alluxio/worker/block/BlockMetadataManager.java
```diff
@@ -468,4 +484,19 @@ public void resizeTempBlockMeta(TempBlockMeta tempBlockMeta, long newSize)
...
+
+  /**
+   * Get the medium type of the block store location.
+   *
+   * @param location the block store location
+   * @return empty string if block store location is a range, otherwise the medium type
+   */
+  public String getMediumType(BlockStoreLocation location) {
+    if (location.tierAlias().equals(BlockStoreLocation.ANY_TIER)
+        || location.dir() == BlockStoreLocation.ANY_DIR) {
```
calvinjia: Is this necessarily true, there could only be one medium type provided in all the tiers or all the dirs in a tier? Do we only need to resolve medium type for specific block store locations?

core/server/worker/src/main/java/alluxio/worker/block/BlockMetadataManager.java
```diff
@@ -242,6 +242,22 @@ public StorageDir getDir(BlockStoreLocation location) {
...
+   * Get the block store location based on the medium type.
+   *
+   * @param mediumType medium type
+   * @return the blockstorelocation object
+   */
+  public BlockStoreLocation getMedium(String mediumType) {
+    for (StorageTier tier : getTiers()) {
+      for (StorageDir dir : tier.getStorageDirs()) {
+        if (dir.getDirMedium().equals(mediumType)) {
+          return new BlockStoreLocation(tier.getTierAlias(), dir.getDirIndex());
```
calvinjia: Won't this always return the first storage dir that matches the medium type (but there could be multiple)?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
yuzhu: The user inputs a list, somewhere we have to do the list to set conversion since we are using set internally, where do you think this conversion should happen?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
yuzhu: actually adding it to the `BlockStoreLocation` removes the need for these two functions

yuzhu:
@calvinjia addressed comments, PTAL. thanks.

yuzhu:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
calvinjia: The conversion from user input to set should happen immediately since we only use set semantics

yuzhu:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -403,6 +403,18 @@ private void applyUpdateInode(UpdateInodeEntry entry) {
...
+        List<String> mediaList = ServerConfiguration.getList(
+            PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIUMTYPES, ",");
+        if (entry.getMediumTypeList().isEmpty()) {
+          // if user does not specify a pinned media list, any location is OK
+          file.setMediumTypes(new HashSet<>());
+        } else {
+          for (String medium : entry.getMediumTypeList()) {
+            if (mediaList.contains(medium)) {
+              file.getMediumTypes().add(medium);
+            }
```
calvinjia: Should we throw an exception if the medium type is invalid?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -137,15 +136,27 @@ public void close() {
...
+    Set<String> pinnedMediumTypes = file.getMediumTypes();
+    if (blockInfo == null && !pinnedMediumTypes.isEmpty()) {
+      return false;
+    }
+    Set<String> storedMediumTypes = new HashSet<>();
+    for (BlockLocation loc : blockInfo.getLocations()) {
+      storedMediumTypes.add(loc.getMediumType());
+    }
+    // if all pinned medium type is found in block locations, then it is correctly pinned.
+    return storedMediumTypes.containsAll(pinnedMediumTypes);
```
calvinjia: Couldn't we simplify the logic to just fail fast if we find a medium type which is not contained?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -167,12 +178,18 @@ private void check(Set<Long> inodes, ReplicationHandler handler, Mode mode) {
...
               }
               break;
             case REPLICATE:
               int minReplicas = file.getReplicationMin();
+              // check if it is correctly replicated
+              if (currentReplicas > 0 && (!checkCorrectlyReplicated(file, blockInfo))) {
+                // some copy is in memory and it is not correctly replicated
+                requests.add(new ReplicationRequest(Mode.EVICT, inodePath.getUri(), blockId, 1));
+                currentReplicas--;
+              }
```
calvinjia: We plan to add the logic for moving the incorrectly place copy in a later PR?

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -344,9 +345,15 @@ public void commitBlockInUfs(long blockId, long length) throws IOException {
...
+  public String createBlock(long sessionId, long blockId, String tierAlias,
+      String medium, long initialBytes)
       throws BlockAlreadyExistsException, WorkerOutOfSpaceException, IOException {
-    BlockStoreLocation loc = BlockStoreLocation.anyDirInTier(tierAlias);
+    BlockStoreLocation loc;
+    if (medium.isEmpty()) {
+      loc = BlockStoreLocation.anyDirInTier(tierAlias);
+    } else {
+      loc = BlockStoreLocation.anyDirInTierWithMedium(medium);
+    }
```
calvinjia: Do we need to handle the case where `tierAlias` and `medium` are populated?

core/server/worker/src/main/java/alluxio/worker/block/allocator/GreedyAllocator.java
```diff
@@ -73,6 +73,20 @@ private StorageDirView allocateBlock(long sessionId, long blockSize,
...
+      for (StorageTierView tierView : mManagerView.getTierViews()) {
+        for (StorageDirView dirView : tierView.getDirViews()) {
+          if (dirView.getMediumType().equals(mediumType)
+              && dirView.getAvailableBytes() >= blockSize) {
+            return dirView;
+          }
+        }
+      }
+      return null;
+    }
```
calvinjia: Do we need to check the medium type in the first loop?

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
```diff
@@ -62,14 +62,22 @@ private StorageDirView allocateBlock(long sessionId, long blockSize,
 
     if (location.equals(BlockStoreLocation.anyTier())) {
       for (StorageTierView tierView : mManagerView.getTierViews()) {
-        candidateDirView = getCandidateDirInTier(tierView, blockSize);
+        candidateDirView = getCandidateDirInTier(tierView, blockSize,
+            BlockStoreLocation.ANY_MEDIUM);
```
calvinjia: Should we be using the medium provided by `location`?

calvinjia:
@yuzhu had a few more comments

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
yuzhu: in this case location has a mediumType of ANY_MEDIUM, that is why it is equal to `anyTier()`.

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
yuzhu: i don't think there is a case where both would be populated right now. maybe add a comment if we do decide to support that case later?

core/server/worker/src/main/java/alluxio/worker/block/allocator/GreedyAllocator.java
yuzhu: first loop? outer loop?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: i had this change in another branch.. really not sure how this change got into this PR.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: we still need to build the set though right? i am not sure if the logic is simpler?

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
calvinjia: Ah I see, that was not immediately obvious since tier and medium are orthogonal. The BlockStoreLocation is not straightforward anymore, could you add some documentation which includes some of the assumptions we make (for example that tier and medium are not concurrently set (I think this might be a valid use case?))

core/server/worker/src/main/java/alluxio/worker/block/allocator/GreedyAllocator.java
calvinjia: Nevermind, I was mistaken about the definition of `anyTier`

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
calvinjia: Oh hmm shouldn't this method return more than a boolean then? Because we need to know which blocks are not correctly replicated?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -167,12 +178,18 @@ private void check(Set<Long> inodes, ReplicationHandler handler, Mode mode) {
...
+                requests.add(new ReplicationRequest(Mode.EVICT, inodePath.getUri(), blockId,
                     currentReplicas - maxReplicas));
               }
               break;
             case REPLICATE:
               int minReplicas = file.getReplicationMin();
+              // check if it is correctly replicated
+              if (currentReplicas > 0 && (!checkCorrectlyReplicated(file, blockInfo))) {
+                // some copy is in memory and it is not correctly replicated
+                requests.add(new ReplicationRequest(Mode.EVICT, inodePath.getUri(), blockId, 1));
```
calvinjia: I'm not sure about this, since couldn't some replicas be correctly replicated?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
yuzhu: unfortunately we can't safely throw exceptions in this context, while we are in `applyUpdateInode`.   Maybe we can move the check to outside of this section.

yuzhu:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
calvinjia: I'm not too concerned, we will just quietly do a best effort, it might be worth doing a preemptive check on the client side and warn if the global types don't match (but not fail).

yuzhu:
jenkins, test this please

yuzhu:
jenkins, test this please

yuzhu:
@calvinjia comments addressed, PTAL. thanks

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -412,6 +412,8 @@ private void applyUpdateInode(UpdateInodeEntry entry) {
           for (String medium : entry.getMediumTypeList()) {
             if (mediaList.contains(medium)) {
               file.getMediumTypes().add(medium);
+            } else {
+              LOG.warn("pinned medium types contain invalid medium");
```
calvinjia: Logging this on the master is not that useful, since it can't be mapped to the client request. I would prefer to do a client side check

shell/src/main/java/alluxio/cli/fs/command/PinCommand.java
```diff
@@ -49,8 +50,13 @@ protected void runPlainPath(AlluxioURI path, CommandLine cl)
       throws AlluxioException, IOException {
     String[] args = cl.getArgs();
     // args[0] is the path, args[1] to args[end] is the list of possible media to pin
-    List<String> mediumTypes = Arrays.asList(Arrays.copyOfRange(args, 1, args.length));
-    FileSystemCommandUtils.setPinned(mFileSystem, path, true, mediumTypes);
+    List<String> pinnedMediumTypes = Arrays.asList(Arrays.copyOfRange(args, 1, args.length));
+    List<String> availableMediumList = mFsContext.getPathConf(path).getList(
+        PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIUMTYPES, ",");
+    if (!availableMediumList.containsAll(pinnedMediumTypes)) {
+      throw new IllegalArgumentException("Invalid medium to pin the file");
```
calvinjia: Could you print which one is invalid / give a list of valid ones?

calvinjia:
Jenkins, test this please

shell/src/main/java/alluxio/cli/fs/command/PinCommand.java
yuzhu: done.

yuzhu:
jenkins, test this please

shell/src/main/java/alluxio/cli/fs/command/PinCommand.java
```diff
@@ -53,8 +54,20 @@ protected void runPlainPath(AlluxioURI path, CommandLine cl)
...
+        validMediumType.add(mediumType);
+      } else {
+        invalidMediumType.add(mediumType);
+      }
+    }
+    if (!invalidMediumType.isEmpty()) {
+      throw new IllegalArgumentException("Invalid medium to pin the file. "
+          + String.join(",", invalidMediumType) + " are invalid. "
+          + (validMediumType.isEmpty() ? ""
+          : String.join(",", validMediumType) + " are valid"));
```
calvinjia: I think its more helpful to print out the `availableMediumList`

yuzhu:
@apc999 PTAL

yuzhu:
jenkins, test this please

calvinjia:

PR #9086, opened 2019-05-13
Implement UserState

gpang:
The main contribution is to remove the LoginUser singleton, and introduce a new concept called `UserState`. The `UserState` manages the login information and state of any client (master, worker, client). There are specific implementations for UserState depending on the security type, so every implementation has its own class.


gpang:
@bf8086 @ZacBlanco

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -102,16 +116,21 @@ public OutStreamOptions(CreateFilePOptions options, AlluxioConfiguration alluxio
...
         alluxioConf);
     mWriteTier = alluxioConf.getInt(PropertyKey.USER_FILE_WRITE_TIER_DEFAULT);
     mWriteType = alluxioConf.getEnum(PropertyKey.USER_FILE_WRITE_TYPE_DEFAULT, WriteType.class);
-    mOwner = SecurityUtils.getOwnerFromLoginModule(alluxioConf);
-    mGroup = SecurityUtils.getGroupFromLoginModule(alluxioConf);
+    try {
+      mOwner = context.getUserState().getUser().getName();
+      mGroup = CommonUtils.getPrimaryGroupName(mOwner, context.getConf());
+    } catch (IOException e) {
+      mOwner = "";
```
bf8086: (nit) Log an error message for the exception

core/common/src/main/java/alluxio/security/user/TestUserState.java
```diff
@@ -0,0 +1,51 @@
...
+import alluxio.conf.AlluxioConfiguration;
+import alluxio.exception.status.UnauthenticatedException;
+import alluxio.security.User;
+
+import javax.security.auth.Subject;
+
+/**
+ * UserState for testing only.
+ */
+public class TestUserState extends BaseUserState {
```
bf8086: Can we move this under the test folder?

tests/src/test/java/alluxio/client/cli/fs/command/SetFaclCommandIntegrationTest.java
```diff
@@ -12,12 +12,16 @@
...
 
 import alluxio.AlluxioURI;
+import alluxio.client.file.FileSystem;
+import alluxio.client.file.FileSystemContext;
 import alluxio.conf.PropertyKey;
 import alluxio.client.cli.fs.AbstractFileSystemShellTest;
 import alluxio.client.file.FileSystemTestUtils;
 import alluxio.client.file.URIStatus;
+import alluxio.conf.ServerConfiguration;
 import alluxio.exception.AlluxioException;
```
bf8086: (nit) This is not introduced by your change, can you reorder the import?

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
ZacBlanco: We could set the level to `DEBUG` if we ever need to trace security exceptions?

core/common/src/main/java/alluxio/security/CurrentUser.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(mName, mServiceName);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    return Objects.equal(mName, mServiceName);
```
ZacBlanco: Why is equals testing whether `mName == mServiceName`? Shouldn't it be something like:
```java
    public boolean equals(Object o) {
      if (!(o instanceof CurrentUser)) {
        return false;
      }
      CurrentUser other = (CurrentUser) o;
      return Objects.equals(mName, other.mName)
          && Objects.equals(mServiceName, other.mServiceName);
    }
```

core/common/src/main/java/alluxio/security/user/BaseUserState.java
```diff
@@ -0,0 +1,110 @@
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (this.getClass() != o.getClass()) {
+      return false;
+    }
+    BaseUserState that = (BaseUserState) o;
+    return Objects.equal(mSubject, that.mSubject);
```
ZacBlanco: should we also compare `mUser`? Same for hashCode

core/common/src/main/java/alluxio/security/CurrentUser.java
gpang: good catch! fixed

core/common/src/main/java/alluxio/security/user/BaseUserState.java
gpang: Subject equality will compare the principals as well. `mUser` is basically a cached reference of the User in the Subject.

gpang:

PR #9087, opened 2019-05-13
Propagate configuration updates automatically to clients

cheng-chang:
This PR merges the feature branch into master.


cheng-chang:
@madanadit @ggezer PTAL

madanadit:
@cheng-chang is there anything in particular you need a review on? all commits to the feature branch have been reviewed before?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2678,6 +2696,7 @@ public String toString() {
           .setDescription("The maximum number of workers to retry before the client gives up on "
               + " reading a block")
           .build();
+  // TODO(cc): remove this since configuration propagation is always enabled?
```
madanadit: did we decide to use a single property for refresh mode: never, once or using heartbeat?

cheng-chang: for path conf, if the mode is never or once, the path conf feature is of no use to clients?

madanadit: makes sense

madanadit:
alluxio-bot, feature-merge this please

PR #9091, opened 2019-05-14
Change default hadoop version from 2.2.0 to 2.7.1

maobaolong:


maobaolong:
@apc999 PTAL.

gpang:
@maobaolong @apc999 Would it make sense to target this to master, then cherry-pick to 1.8?

apc999:
@gpang agree. @maobaolong  can you create the same change into master branch? For branch 1.8, I am hesitated to cherry-pick and bump the hadoop version as we typically keep the dependency stable across patch releases (e.g., from 1.8.1 to 1.8.2)

Independently, I agree to bump the default HDFS version to 2.7 from 2.2 in Alluxio 2.0. Latest Spark depends on HDFS 2.7, which will make the src code dependency easy to handle. @calvinjia @gpang , are you good?

Separately, @gpang @calvinjia , I think we should bump

apc999:
@calvinjia , do we want to include this change to 2.0.0? On the binary side, it should make no difference. Only affecting src code level dependency. Right?

dev/scripts/src/alluxio.org/build-distribution/cmd/common.go
```diff
@@ -32,11 +32,11 @@ var hadoopDistributions = map[string]version{
 	"hadoop-2.4": parseVersion("2.4.1"),
 	"hadoop-2.5": parseVersion("2.5.2"),
 	"hadoop-2.6": parseVersion("2.6.5"),
-	"hadoop-2.7": parseVersion("2.7.3"),
+	"hadoop-2.7": parseVersion("2.7.1"),
```
calvinjia: Is there a reason we want to use `.1` instead of `.3` or latest (`.7`)?

apc999:
Link issue https://github.com/Alluxio/alluxio/issues/9753 to this PR. We are aiming to solve it in Alluxio v2.1

ns1123:
@maobaolong you will also need to update the default value of the UNDERFS_VERSION constant in PropertyKey.java. It is currently set to "2.2".

apc999:
This PR https://github.com/Alluxio/alluxio/pull/9897/files takes care of the default HDFS version bumping

PR #9092, opened 2019-05-14
Reduce memory footprint of InodeLockList

cheng-chang:
Although the Entry abstraction in SimpleInodeLockList makes code more readable, but it unnecessarily stores extra representations of Edges, each consumes a long and a String. Since Each FileSystemMaster operation will create at least one InodeLockList, the lower memory consumption the better.

This PR removes the Entry abstraction without sacrificing readability too much.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
@@ -51,9 +51,10 @@ public class SimpleInodeLockList implements InodeLockList {
    */
   private List<LockResource> mLocks;
   /**
-   * The index of the first write lock entry. All locks before this index are read locks, and all
-   * locks after and including this index are write locks. If all locks are read locks,
-   * mFirstWriteLockIndex == NO_WRITE_LOCK.
+   * The index of the first write lock entry.
+   * If all locks are read locks, mFirstWriteLockIndex == NO_WRITE_LOCK.
+   * Otherwise, all locks before this index are read locks, and all
+   * locks after and including this index are write locks.
    */
   private int mFirstWriteLockIndex;
 
--- a/core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
@@ -31,7 +31,10 @@ import javax.annotation.concurrent.NotThreadSafe;
 public class SimpleInodeLockList implements InodeLockList {
   private static final Logger LOG = LoggerFactory.getLogger(SimpleInodeLockList.class);
 
-  private static final int NO_WRITE_LOCK = -1;
+  /**
+   * Default value for {@link #mFirstWriteLockIndex} when there is no write lock.
+   */
+  private static final int NO_WRITE_LOCK_INDEX = -1;
   private static final int INITIAL_CAPACITY = 8;
   private static final Edge ROOT_EDGE = new Edge(-1, "");
 
@@ -47,11 +50,13 @@ public class SimpleInodeLockList implements InodeLockList {
    */
   private Edge mLastEdge;
   /**
-   * Lock list. The locks always alternate between Inode lock and Edge lock.
+   * Lock list.
+   * The locks always alternate between Inode lock and Edge lock.
+   * The first lock can be either Inode or Edge lock.
    */
   private List<LockResource> mLocks;
   /**
-   * The index of the first write lock entry.
+   * The index of the first write lock entry in {@link #mLocks}.
    * If all locks are read locks, mFirstWriteLockIndex == NO_WRITE_LOCK.
    * Otherwise, all locks before this index are read locks, and all
    * locks after and including this index are write locks.
@@ -67,7 +72,7 @@ public class SimpleInodeLockList implements InodeLockList {
     mInodeLockManager = inodeLockManager;
     mInodes = new ArrayList<>(INITIAL_CAPACITY);
     mLocks = new ArrayList<>(INITIAL_CAPACITY);
-    mFirstWriteLockIndex = NO_WRITE_LOCK;
+    mFirstWriteLockIndex = NO_WRITE_LOCK_INDEX;
   }
 
   @Override
@@ -250,7 +255,7 @@ public class SimpleInodeLockList implements InodeLockList {
   private void removeLastLock() {
     mLocks.remove(mLocks.size() - 1).close();
     if (mFirstWriteLockIndex >= mLocks.size()) {
-      mFirstWriteLockIndex = NO_WRITE_LOCK;
+      mFirstWriteLockIndex = NO_WRITE_LOCK_INDEX;
     }
     if (mLastEdge != null) {
       mLastEdge = null;
@@ -323,11 +328,11 @@ public class SimpleInodeLockList implements InodeLockList {
    * @return whether this lock list ends in a write lock
    */
   private boolean endsInWriteLock() {
-    return mFirstWriteLockIndex != NO_WRITE_LOCK;
+    return mFirstWriteLockIndex != NO_WRITE_LOCK_INDEX;
   }
 
   private boolean endsInMultipleWriteLocks() {
-    return mFirstWriteLockIndex != NO_WRITE_LOCK && mFirstWriteLockIndex < mLocks.size() - 1;
+    return mFirstWriteLockIndex != NO_WRITE_LOCK_INDEX && mFirstWriteLockIndex < mLocks.size() - 1;
   }
 
   @Override
--- a/core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
@@ -22,7 +22,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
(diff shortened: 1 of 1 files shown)


cheng-chang:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -186,67 +197,72 @@ public void downgradeEdgeToInode(Inode inode, LockMode mode) {
...
-   * If the lock list is write locked, returns LockMode.WRITE. Otherwise, returns the given mode.
+   * If mode is read but the lock list is write locked, precondition check will fail.
    *
    * This helps us preserve the invariant that there is never a READ lock following a WRITE lock.
    *
    * @param mode a lock mode
-   * @return the mode
    */
-  private LockMode nextLockMode(LockMode mode) {
+  private void checkMode(LockMode mode) {
```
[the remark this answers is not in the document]
cheng-chang: I think the implicit lock upgrade makes sense, I changed back.

cheng-chang:
@ggezer thanks for comments, PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -32,20 +31,30 @@
...
   /**
-   * The index of the first write lock entry. All locks before this index are read locks, and all
-   * locks after and including this index are write locks. If all locks are read locks,
-   * mFirstWriteLockIndex == mEntries.size().
+   * If last lock in mLocks is an edge lock, then this is the edge.
+   * Otherwise, null.
+   */
+  private Edge mLastEdge;
+  /**
+   * Lock list. The locks always alternate between Inode lock and Edge lock.
```
gpang: Is inode lock always first?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -32,20 +31,30 @@
...
+   */
+  private List<LockResource> mLocks;
+  /**
+   * The index of the first write lock entry.
```
gpang: index into `mLocks`?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -56,43 +65,43 @@
    */
   public SimpleInodeLockList(InodeLockManager inodeLockManager) {
     mInodeLockManager = inodeLockManager;
-    mEntries = new ArrayList<>(INITIAL_CAPACITY);
-    mFirstWriteLockIndex = 0;
+    mInodes = new ArrayList<>(INITIAL_CAPACITY);
+    mLocks = new ArrayList<>(INITIAL_CAPACITY);
```
gpang: Is the initial capacity an issue for memory consumption? This change effectively made the initial size twice as large

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: Added comment `The first lock can be either Inode or Edge lock.`

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -32,20 +31,30 @@
 public class SimpleInodeLockList implements InodeLockList {
   private static final Logger LOG = LoggerFactory.getLogger(SimpleInodeLockList.class);
 
+  private static final int NO_WRITE_LOCK = -1;
```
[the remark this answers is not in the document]
cheng-chang: Added comment `Default value for {@link #mFirstWriteLockIndex} when there is no write lock.`.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: Updated comment: ` The index of the first write lock entry in {@link #mLocks}`.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: I changed to reserve less capacity.

cheng-chang:
@gpang PTAL

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
ggezer: 2 is too few as it only covers locking a hierarchy of 1 level deep.
Since this struct grows progressively with every lock and we remove only from tail, I believe LinkedList could be of use here instead of ArrayList. WDYT?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -56,43 +79,43 @@
...
     mode = nextLockMode(mode);
-    if (!mEntries.isEmpty()) {
+    if (!mLocks.isEmpty()) {
       Preconditions.checkState(!endsInInode(),
           "Cannot lock inode %s for lock list %s because the lock list already ends in an inode",
           inode.getId(), this);
-      String lastEdgeName = ((EdgeEntry) lastEntry()).getEdge().getName();
-      Preconditions.checkState(inode.getName().equals(lastEdgeName),
-          "Expected to lock inode %s but locked inode %s", lastEdgeName, inode.getName());
+      Preconditions.checkState(inode.getName().equals(mLastEdge.getName()),
```
gpang: `mLastEdge` could be `null`, right?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -186,67 +212,71 @@ public void downgradeEdgeToInode(Inode inode, LockMode mode) {
...
-      mFirstWriteLockIndex = mEntries.size();
+  private void removeLastLock() {
+    mLocks.remove(mLocks.size() - 1).close();
+    if (mFirstWriteLockIndex >= mLocks.size()) {
+      mFirstWriteLockIndex = NO_WRITE_LOCK_INDEX;
+    }
+    if (mLastEdge != null) {
+      mLastEdge = null;
+    } else {
+      Inode last = mInodes.remove(mInodes.size() - 1);
```
gpang: Why is this required? Don't we already unlock and remove earlier?

Could this fail because there is nothing to remove?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -186,67 +212,71 @@ public void downgradeEdgeToInode(Inode inode, LockMode mode) {
...
+      if (!mLocks.isEmpty()) {
+        if (mInodes.isEmpty()) {
+          mLastEdge = ROOT_EDGE;
+        } else {
+          mLastEdge = new Edge(mInodes.get(mInodes.size() - 1).getId(), last.getName());
```
gpang: Is `mInodes.get(mInodes.size() - 1)` going to be the one before `last`? Do all of these indexes exist?

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: no, the previous precondition has checked that it ends in an edge, so mLastEdge must not be null.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: in line https://github.com/Alluxio/alluxio/pull/9092/files#diff-74f13917d59a65b6dcbc4f09e382edd3R265, we only remove the lock, here, we remove the corresponding inode.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: when this `else` block is executed, neither `mLocks` nor `mInodes`is empty.

cheng-chang:
@gpang PTAL

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
cheng-chang: My only concern of using LinkedList is that the implementation of `get` is O(n). But I checked that the only usage of `get` is to get the last or the second last element, which is O(1) in LinkedList. So I changed to LinkedList.

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
```diff
@@ -257,32 +278,22 @@ public LockMode getLockMode() {
...
-        .map(e -> ((InodeEntry) e).getInode())
-        .collect(Collectors.toList());
+    return new ArrayList<>(mInodes);
   }
 
   @Override
   public Inode get(int index) {
-    int entryIndex = mEntries.get(0) instanceof InodeEntry ? index * 2 : index * 2 + 1;
-    return ((InodeEntry) mEntries.get(entryIndex)).getInode();
+    return mInodes.get(index);
```
gpang: This will be slow with the linked list now?

I'd imagine most of the usage only wants the last index? Is that efficient for the implementation of the linked list?

[later] Maybe we should consider adding a `getLastInode` api.

cheng-chang: LinkedList is a double linked list, so getting the last index is O(1)

cheng-chang:
@gpang PTAL

core/server/master/src/main/java/alluxio/master/file/meta/SimpleInodeLockList.java
gpang: looks like LinkedList.get() starts from the end that the index is closer to.

cheng-chang:

PR #9094, opened 2019-05-14
Fix a bug where last edge is not locked in WRITE_INODE pattern

cheng-chang:
Suppose WRITE_INODE on path /a/b/c, and only /a exists, then both inode a and edge a->b should be read locked, otherwise, a concurrent rename from /a/c -> a/b will succeed.

See the enhanced unit test in LockedInodePathTest.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
@@ -347,9 +347,9 @@ public class InodeTree implements DelegatingJournaled {
         new LockedInodePath(uri, mInodeStore, mInodeLockManager, getRoot(), lockPattern);
     try {
       inodePath.traverse();
-    } catch (Throwable t) {
+    } catch (InvalidPathException e) {
       inodePath.close();
-      throw t;
+      throw e;
     }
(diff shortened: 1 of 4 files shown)


core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -159,6 +159,26 @@
...
+     *
+     * path to lock: /a/b/c
+     * existing inodes: /a/b/c
+     * last edge: b->c
+     *
+     * path to lock: /a/b/c
+     * existing inodes: /a
+     * last edge: a->b
+     *
+     * @return whether the last edge should be locked
```
gpang: write locked?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -159,6 +159,26 @@
...
+     */
+    public boolean shouldLockLastEdge() {
```
gpang: should this be named "writeLockLastEdge"?

cheng-chang:
@gpang PTAL

cheng-chang:

PR #9095, opened 2019-05-15
Use Swagger to generate rest docs

calvinjia:
Resulting documentation example is attached. We can change the styling to other bootswatch templates. I've chosen one that is similar to Alluxio colors (blue).

Docs are generated in `generated/<service>/index.html`. So far the job, master, worker, and proxy are supported. Deprecated APIs are not documented, and internal APIs (ie. for the web ui) are not documented because we do not want others to rely on them.
Screen Shot 2019-05-16 at 5 47 20 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,12 +40,14 @@
 /docs/api/
 /docs/serve/
 /extensions/
+/generated/
 /journal/
 /lib/
 /logs/*.log*
 /logs/*.out*
 /logs/user/*
 /metastore
+/templates/
 /underFSStorage/
 /underfs/hdfs/src/main/java/alluxio/UfsConstants.java
 /web/
--- a/.gitignore
+++ b/.gitignore
@@ -47,7 +47,6 @@
 /logs/*.out*
 /logs/user/*
 /metastore
-/templates/
 /underFSStorage/
 /underfs/hdfs/src/main/java/alluxio/UfsConstants.java
 /web/
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterClientRestServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterClientRestServiceHandler.java
@@ -18,7 +18,6 @@ import alluxio.master.AlluxioMasterProcess;
 import alluxio.web.MasterWebServer;
 
 import com.google.common.base.Preconditions;
-import com.qmino.miredot.annotations.ReturnType;
 
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.servlet.ServletContext;
@@ -65,7 +64,7 @@ public final class BlockMasterClientRestServiceHandler {
    */
   @GET
   @Path(SERVICE_NAME)
-  @ReturnType("java.lang.String")
+  //@ReturnType("java.lang.String")
   public Response getServiceName() {
     return RestUtils.call(() -> Constants.BLOCK_MASTER_CLIENT_SERVICE_NAME,
         ServerConfiguration.global());
@@ -77,7 +76,7 @@ public final class BlockMasterClientRestServiceHandler {
    */
   @GET
   @Path(SERVICE_VERSION)
-  @ReturnType("java.lang.Long")
+  //@ReturnType("java.lang.Long")
   public Response getServiceVersion() {
     return RestUtils.call(() -> Constants.BLOCK_MASTER_CLIENT_SERVICE_VERSION,
         ServerConfiguration.global());
@@ -90,7 +89,7 @@ public final class BlockMasterClientRestServiceHandler {
    */
   @GET
   @Path(GET_BLOCK_INFO)
-  @ReturnType("alluxio.wire.BlockInfo")
+  //@ReturnType("alluxio.wire.BlockInfo")
   public Response getBlockInfo(@QueryParam("blockId") final Long blockId) {
     return RestUtils.call(() -> {
       Preconditions.checkNotNull(blockId, "required 'blockId' parameter is missing");
--- a/job/server/pom.xml
+++ b/job/server/pom.xml
@@ -133,14 +133,17 @@
           <apiSource>
             <springmvc>false</springmvc>
             <schemes>http</schemes>
-            <host>[Alluxio Job Master Hostname]</host>
+            <host>[Alluxio Job Master or Job Worker Hostname]</host>
             <basePath>/api/v1</basePath>
             <info>
               <version>v1</version>
-              <title>Alluxio Job Master REST API Documentation</title>
-              <description>The Alluxio Job Master is a component of the Job Service that coordinates Alluxio Job Workers to execute distributed tasks scheduled by the Alluxio system.</description>
+              <title>Alluxio Job Service REST API Documentation</title>
+              <description>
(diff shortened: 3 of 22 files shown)


core/server/master/src/main/java/alluxio/master/block/BlockMasterClientRestServiceHandler.java
```diff
@@ -65,7 +64,7 @@ public BlockMasterClientRestServiceHandler(@Context ServletContext context) {
    */
   @GET
   @Path(SERVICE_NAME)
-  @ReturnType("java.lang.String")
+  //@ReturnType("java.lang.String")
```
calvinjia: Currently commenting these out because the return type will be reused when we reannotate with Swagger's annotations.

.gitignore
```diff
@@ -40,12 +40,14 @@
 /docs/api/
```
Xenorith: are there folders to unignore after removing miredot?

job/server/pom.xml
```diff
@@ -124,6 +123,22 @@
...
+
+    <!-- REST API documentation -->
+    <plugin>
+      <groupId>com.github.kongchen</groupId>
+      <artifactId>swagger-maven-plugin</artifactId>
+      <configuration>
+        <apiSources>
+          <apiSource>
+            <locations>alluxio.master</locations>
+            <templatePath>${project.parent.parent.basedir}/templates/strapdown.html.hbs</templatePath>
```
Xenorith: seems like a provided file rather than a generated file; should templates/ be ignored? is this the entrypoint to styling the output HTML to whatever we like?

`project.parent.parent`... o.O

calvinjia: Good catch on the templates, and yeah its because this is nested in two modules.

.gitignore
calvinjia: Possibly, would it be cleaner if we outputted the documentation in the same location?

Xenorith: i'm not a fan of how miredot spread all the generated files in each of the packages so i would prefer if we could output all the docs into the same directory if possible

pom.xml
```diff
@@ -1034,6 +974,47 @@
...
+                <basePath>/api/v1</basePath>
+                <info>
+                  <title>Alluxio REST API Documentation</title>
+                  <version>v1</version>
+                  <termsOfService>
+                    http://www.github.com/alluxio/alluxio
+                  </termsOfService>
+                  <description>Documentation for Alluxio RESTful APIs</description>
+                  <contact>
+                    <email>haoyuan@alluxio.com</email>
```
Xenorith: this email is used for the "Contact the developer" link. i'm guessing we shouldn't use HY's email here but i'm not sure which email alias would be best given that this is going into open source...

job/server/pom.xml
```diff
@@ -124,6 +123,22 @@
...
+            <outputPath>${project.parent.parent.basedir}/generated/job/document.html</outputPath>
```
Xenorith: document.html -> index.html please for backcompat and silly web naming conventions

templates/strapdown.html.hbs
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<title>API Document</title>
```
Xenorith: i'd remove this block; the internal page links (ex. document.html#/definitions/WorkerInfo) navigate the page in a way where the title (ex. WorkerInfo in the above example) is covered up by this title

ZacBlanco: This template doesn't conform to the standard that most all HTML documents follow on the web. The format should be the following:

```html
<!DOCTYPE html>
<html>
  <head>
    <title></title>
  </head>
  <body>
    {{ content }}
    {{ scripts to load asynchronously }}
  </body>
</html>
```

pom.xml
calvinjia: I used the same email/name associated with the Alluxio repository, but another option is just to remove this section.

Xenorith:
related to this PR, we currently hardcode the following rest doc endpoints in the sidebar HTML:
```
                        <a class="dropdown-item" href="{{site.restdocBaseurl}}/master/index.html">Master REST API</a>
                        <a class="dropdown-item" href="{{site.restdocBaseurl}}/worker/index.html">Worker REST API</a>
                        <a class="dropdown-item" href="{{site.restdocBaseurl}}/proxy/index.html">Proxy REST API</a>
```
but i noticed here you generated the rest endpoints for `job`. @ZacBlanco we may have to make the set of rest doc links configurable per version :/

ZacBlanco:
All I have to say to that is _UGH_. We'll definitely need to add a bit of logic for the docs to make sure that `job/` endpoints are only displayed on the proper versions

templates/strapdown.html.hbs
calvinjia: Unfortunately the js code gets buggy if I try it that way. How important is the formatting?

calvinjia: Removed

job/server/pom.xml
calvinjia: Updated, thanks.

pom.xml
calvinjia: I've gotten rid of this.

templates/strapdown.html.hbs
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<body>
+<xmp theme="cerulean" style="display:none;">
```
ZacBlanco: actually, [because `<xmp>` is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/xmp), and this is simply a tag, can we change this to something more modern i.e. `<div>` or `<section>`? and keep the same theme and style attributes? It shouldn't affect the output unless the CSS relies specifically on the `xmp` element.

templates/strapdown.html.hbs
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<body>
```
ZacBlanco: Also, we should probably still have a `<head>` above the body with `<title>` inside so that when users open the page, there is something shown on the tab in their browser. try `{{title}}`?

templates/strapdown.html.hbs
calvinjia: The script is hard coded to look for the xmp tag.

ZacBlanco: This makes me sad :/

templates/strapdown.html.hbs
calvinjia: @Xenorith suggested removing the title because it making jumping by anchor tags difficult due to the title bar covering the sub section title.

ZacBlanco: the header is completely different from `<head><title></title></head>`. This is page metadata only and does not affect rendering of the page. It will just put the page title in your tab of the browser

calvinjia: The CSS will automatically add a header bar if there is a title present.

ZacBlanco: This is the worst piece of javascript I've ever encountered

ZacBlanco:
I'm okay leaving the strapdown dependency for now. If we ever want to update the docs in the future, it shouldn't be terribly hard to replace.

calvinjia:

PR #9099, opened 2019-05-16
Pin to specific tier, phase 2

yuzhu:
In this phase, we address the problem that some blocks may not be at the pinned desired medium. 
This PR roughly includes: 
- Updates to `blockMaster` to include medium information and dirIndex so that we can reconstruct a `BlockStoreLocation` as the source location of the moveBlock command
- Logic to determine which candidate to move to satisfy the pinning constraint

TODO:
Actually call the moveBlock function


job/server/src/main/java/alluxio/job/replicate/MoveDefinition.java
```diff
@@ -0,0 +1,129 @@
...
+    BlockStoreLocationProto srcProto = config.getSource().toProto();
+    BlockStoreLocationProto dstProto = config.getDestination().toProto();
+
+    MoveBlockRequest request = MoveBlockRequest.newBuilder().setBlockId(blockId).setSrc(srcProto)
+        .setDst(dstProto).build();
+    BlockWorkerClient blockWorker = null;
+    try {
+      blockWorker = context.getFsContext().acquireBlockWorkerClient(localNetAddress);
+      blockWorker.moveBlock(request);
+    } finally {
```
calvinjia: Can change this to use `try-with-resources`?

job/server/src/main/java/alluxio/job/replicate/MoveDefinition.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  @Override
+  public Map<WorkerInfo, SerializableVoid> selectExecutors(MoveConfig config,
+      List<WorkerInfo> jobWorkerInfoList, SelectExecutorsContext context) {
+    Preconditions.checkArgument(!jobWorkerInfoList.isEmpty(), "No worker is available");
+
+    String workerHost = config.getWorkerHost();
+
+    Map<WorkerInfo, SerializableVoid> result = Maps.newHashMap();
```
calvinjia: This constructor is deprecated in favor of `new HashMap`?

job/server/src/main/java/alluxio/job/replicate/MoveDefinition.java
```diff
@@ -0,0 +1,129 @@
...
+
+    Collections.shuffle(jobWorkerInfoList);
+    for (WorkerInfo workerInfo : jobWorkerInfoList) {
+      // Select job workers that have this block locally to move
+      if (workerHost.equals(workerInfo.getAddress().getHost())) {
+        result.put(workerInfo, null);
+        return result;
```
calvinjia: Do we only need to select one worker?

yuzhu: if we are sending n job worker requests for n movements, then we are selecting one worker in each of these requests

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -124,20 +134,108 @@ public void heartbeat() throws InterruptedException {
     TimeUnit.SECONDS.sleep(mQuietPeriodSeconds);
     Set<Long> inodes;
 
+
```
calvinjia: Remove extra new line

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -124,20 +134,108 @@ public void heartbeat() throws InterruptedException {
...
     // Nothing to clean up
   }
 
+  /**
+   * Find a set of from and to locations for our moveBlock operation.
+   * After the move, the file will be in the desired pinned medium at least minReplication times.
+   *
+   * @param file
+   * @param blockInfo
+   * @return
```
calvinjia: Please complete the javadoc

yuzhu:
jenkins, test this please

yuzhu:
jenkins, test this please

yuzhu:
@calvinjia Addressed comments, PTAL, thanks.

core/client/fs/src/main/java/alluxio/client/block/stream/DefaultBlockWorkerClient.java
```diff
@@ -167,6 +169,12 @@ public RemoveBlockResponse removeBlock(final RemoveBlockRequest request) {
         .removeBlock(request);
   }
 
+  @Override
+  public MoveBlockResponse moveBlock(MoveBlockRequest request) {
+    return mRpcBlockingStub.withDeadlineAfter(mDataTimeoutMs, TimeUnit.MILLISECONDS)
+        .moveBlock(request);
```
calvinjia: I don't think this is the right timeout to use. The MoveBlock request could take a significantly longer period of time due to data movement.

job/server/src/main/java/alluxio/job/replicate/MoveDefinition.java
calvinjia: @yuzhu ping

core/server/worker/src/main/java/alluxio/worker/block/BlockHeartbeatReporter.java
```diff
@@ -139,13 +142,13 @@ private void removeBlockInternal(long blockId) {
...
    */
-  private void addBlockToAddedBlocks(long blockId, String tierAlias) {
-    if (mAddedBlocks.containsKey(tierAlias)) {
-      mAddedBlocks.get(tierAlias).add(blockId);
+  private void addBlockToAddedBlocks(long blockId, BlockStoreLocation location) {
+    if (mAddedBlocks.containsKey(location)) {
+      mAddedBlocks.get(location).add(blockId);
     } else {
-      mAddedBlocks.put(tierAlias, Lists.newArrayList(blockId));
+      mAddedBlocks.put(location, Lists.newArrayList(blockId));
```
calvinjia: Does this work because the hash will be different for different dirs but we need to aggregate across dirs when we report to the master?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -131,13 +137,101 @@ public void heartbeat() throws InterruptedException {
...
+  /**
+   * Find a set of from and to locations for our moveBlock operation.
+   * After the move, the file will be in the desired pinned medium at least minReplication times.
+   *
+   * @param file the file to scan for misplaced blocks
+   * @param blockInfo blockInfo containing information about the block locations
+   * @return a map that maps from workerHost to desired medium
+   */
+  private Map<String, String> findMisplacedBlock(
+      InodeFile file, BlockInfo blockInfo) {
```
calvinjia: Does it make sense to disassociate the `file` concept and simply pass in the pinned medium types and min replication? Also should this deal with max replication or is that handled in another check?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -131,13 +137,101 @@ public void heartbeat() throws InterruptedException {
...
+      // nothing needs to be moved
+      return Collections.emptyMap();
+    }
+    // at least pinned to one medium type
+    String firstPinnedMedium = pinnedMediumTypes.iterator().next();
+    int minReplication = file.getReplicationMin();
+    int correctReplication = 0;
+    List<String> candidates = new ArrayList<>();
+    for (BlockLocation loc: blockInfo.getLocations()) {
+      if (loc.getMediumType().equals(firstPinnedMedium)) {
```
calvinjia: Why do we check this condition instead of `pinnedMediumTypes.contains(loc.getMediumType())`?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -131,13 +137,101 @@ public void heartbeat() throws InterruptedException {
...
+        candidates.add(loc.getWorkerAddress().getHost());
+      }
+    }
+    if (correctReplication >= minReplication) {
+      // there are more than minReplication in the right location
+      return Collections.emptyMap();
+    } else {
+      int toMove = minReplication - correctReplication;
+      for (String candidate : candidates) {
+        movement.put(candidate, firstPinnedMedium);
```
calvinjia: If multiple medium types are pinned, should this be more flexible? Can leave a comment if we do not plan to address this use case now.

calvinjia:
@yuzhu Thanks for the update, I left some more comments.

core/server/worker/src/main/java/alluxio/worker/block/BlockHeartbeatReporter.java
yuzhu: Good catch, this was from before when we passed information keyed by tieralias/dirIndex/storageMedium to the master.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: this does not handle min replication, there are essentially three checks. One check for minReplication, one check for MaxReplication one check for number of replications that match but medium doesn't match.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: because we are not address multiple medium types being pinned right now. but yeah I agree, your solution covers both cases.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: will add comment

core/client/fs/src/main/java/alluxio/client/block/stream/DefaultBlockWorkerClient.java
yuzhu: do you have a recommended timeout value in this case?

calvinjia: Not any existing value, take a look at the worker side move block call and see what kind of upper bound it has in terms of the retries.

yuzhu: the default value for `mDataTimeoutMs` is 30 secs and the default value for worker side move block call timeout is 10 secs.   Maybe we can use this value if the default is that high?

job/server/src/main/java/alluxio/job/replicate/MoveDefinition.java
yuzhu: `blockWorker.close()` here does not seem to do what i want though, at least not the same.  
`context.getFsContext().releaseBlockWorkerClient` seems more appropriate?

calvinjia:

PR #9103, opened 2019-05-16
Change Zookeeper path defaults

LuQQiu:
Currently the default leader path is "/leader" and the default election path is "/election". These defaults make sense when Zookeeper is being used for Alluxio only, but in real-world deployments, there will often be multiple services making use of Zookeeper. Change the defaults to something like "/alluxio/leader" and "/alluxio/election" so that there's a better chance that users will not have to change them. It is especially annoying to use non-default paths because clients need to be configured with the leader path, and this configuration can't be specified through a centralized cluster config, or by adding something to the URI.


calvinjia:

PR #9106, opened 2019-05-17
Replace lock with atomic counter in FileSystemContextReinitializer

cheng-chang:
Previously, lock is used. But in FileInStream and FileOutStream, the lock may be held in one thread during construction and unlocked in another thread during close(), this will cause IllegalMonitorStateException.

This is observed when testing Alluxio with Presto.


cheng-chang:
jenkins, test this please

calvinjia:
@cheng-chang should we be doing this logic in the streams given that they are not intended to be threadsafe?

cheng-chang:
@calvinjia what is not threadsafe?

calvinjia:
`FileInStream` and `FileOutStream` aren't intended to be used by multiple threads

cheng-chang:
@calvinjia this PR takes this into account, even if different threads are using the streams, the reference counting is thread safe.

calvinjia:
Right, my question was should we be making the streams closer to thread safe or just move the critical logic out of the streams?

cheng-chang:
@calvinjia currently, this is the cleanest way I can come up with.

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
```diff
@@ -89,51 +87,48 @@ public void onSuccess() {
...
    *
    * If there is an existing reinitialization exception, immediately throw an exception
    * without trying to block further reinitialization.
    *
    * @throws IOException when reinitialization fails before or during this method
-   * @return a resource holding the locked read lock
+   * @return the resource
    */
-  public LockResource acquireReadLockResource() throws IOException {
+  public Optional<CountResource> acquireBlockResource() throws IOException {
```
ggezer: NIT: acquireResourceForBlock

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
```diff
@@ -89,51 +87,48 @@ public void onSuccess() {
...
-   * @throws TimeoutException if timed out
-   * @throws InterruptedException if the current thread is interrupted while being blocked
+   * @return the resource
    */
-  public LockResource acquireWriteLockResource() throws TimeoutException, InterruptedException {
-    long timeout = mContext.getClusterConf().getMs(PropertyKey.USER_CONF_SYNC_TIMEOUT);
-    if (!mLock.writeLock().tryLock(timeout, TimeUnit.MILLISECONDS)) {
-      throw new TimeoutException("Failed to begin reinitialization after being blocked for "
-          + DurationFormatUtils.formatDurationWords(timeout, true, true));
+  public Optional<CountResource> acquireAllowResource() {
```
ggezer: NIT: acquireResourceForAllow

core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
```diff
@@ -92,8 +92,9 @@ public synchronized void heartbeat() throws InterruptedException {
         mContext.getClientContext().getPathConfHash());
     if (isClusterConfUpdated || isPathConfUpdated) {
       try {
-        mContext.reinit(isClusterConfUpdated, isPathConfUpdated);
-        mException = null;
+        if (mContext.reinit(isClusterConfUpdated, isPathConfUpdated)) {
+          mException = null;
+        }
```
ggezer: Some warning would be useful if not synced. Basically the same message as in the deleted TimeoutException section.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
```diff
@@ -89,51 +87,48 @@ public void onSuccess() {
...
-  public LockResource acquireReadLockResource() throws IOException {
+  public Optional<CountResource> acquireBlockResource() throws IOException {
     Optional<IOException> exception = mExecutor.getException();
     if (exception.isPresent()) {
       throw exception.get();
     }
-    mLock.readLock().lock();
+    if (mCount.updateAndGet(c -> c >= 0 ? c + 1 : c) <= 0) {
+      return Optional.empty();
+    }
```
ggezer: Would it make sense to embed this logic and the allow logic to CountResource itself. CountResource by itself already is pretty much restricted to this scenario. So I feel it makes sense to make it expose its own allow/block functions and avoid exposing in its constructor a parameter for whether to increase/decrease the value.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -306,33 +334,43 @@ public LockResource acquireBlockReinitLockResource() {
...
-      throws TimeoutException, InterruptedException, UnavailableException, IOException {
-    try (LockResource r = mReinitializer.acquireWriteLockResource()) {
-      InetSocketAddress masterAddr;
-      try {
-        masterAddr = getMasterAddress();
-      } catch (IOException e) {
-        throw new UnavailableException("Failed to get master address during reinitialization", e);
+  public boolean reinit(boolean updateClusterConf, boolean updatePathConf)
+      throws UnavailableException, IOException {
+    if (!mReinitLock.writeLock().tryLock()) {
```
ggezer: With these changes the only chance for HashSync to happen is that there are no active RPCs at the time of heartbeat. That is pretty unfair to hash-sync. How about try-locking with a timeout value? You could use the same timeout that you used before. That way has-sync will have a better chance to run.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -394,18 +394,18 @@ public FileInStream openFile(AlluxioURI path)
...
-    }
-    // getStatus should have loaded path level configs.
-    OpenFilePOptions mergedOptions = FileSystemOptions.openFileDefaults(
-        mFsContext.getPathConf(path)).toBuilder().mergeFrom(options).build();
-    InStreamOptions inStreamOptions = new InStreamOptions(status, mergedOptions,
-        mFsContext.getPathConf(path));
-    // FileSystemContext reinitialization is blocked during construction and close of FileInStream.
-    return new FileInStream(status, inStreamOptions, mFsContext);
+    return rpc(client -> {
+      URIStatus status = client.getStatus(path, GetStatusPOptions.getDefaultInstance());
```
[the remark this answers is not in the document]
cheng-chang: Nop, I changed to use path config here. My intention is to use the `rpc` convention to ensure the mFsContext used in `openFile` is consistent, otherwise, there is a possibility that `getStatus` uses an old version of `mFsContext`, then `mFsContext` is reinitialized, then the `FileInStream` uses the new version.

cheng-chang:
jenkins, test this please

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
cheng-chang: changed to acquireResourceToBlockReinit

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
cheng-chang: changed to acquireResourceToAllowReinit

cheng-chang:
jenkins, test this please

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
cheng-chang: I imagine the `CountResource` can be used in other scenarios that need reference counting.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
cheng-chang: but even if reinit acquires this lock, it still needs to acquire the counter resource to allow reinit. If there are streaming operations happening, reinit will not happen even if this write lock is acquired.

I changed to use a fair lock, and just lock instead of tryLock here, so that reinit has a better chance to run.

cheng-chang:
@ggezer your concern is correct, reinitialization only happens only when there are no ongoing RPCs. When reinitialization is needed while there are ongoing RPCs, we can block further RPCs and wait for the ongoing RPCs to finish for a certain amount of time, but by design, we would not block RPCs unless we are actually reinitializing.

calvinjia:

PR #9107, opened 2019-05-17
Add a command to enable updating mount options

bf8086:
This change adds a CLI command to update mount options without losing existing metadata.

Usage example:
`bin/alluxio fs updateMount /mnt --readonly --option abc.def=xyz`


bf8086:
@calvinjia @ns1123 PTAL. Thanks!

bf8086:
Jenkins, test this please.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -354,6 +354,19 @@ public void mount(AlluxioURI alluxioPath, AlluxioURI ufsPath, final MountPOption
...
 
+  @Override
+  public void updateMount(AlluxioURI alluxioPath, final MountPOptions options)
+      throws IOException, AlluxioException {
+    checkUri(alluxioPath);
+    rpc(client -> {
+      MountPOptions mergedOptions = FileSystemOptions.mountDefaults(
+          mFsContext.getPathConf(alluxioPath)).toBuilder().mergeFrom(options).build();
+      client.updateMount(alluxioPath, mergedOptions);
+      LOG.info("UpdateMount on " + alluxioPath.getPath());
```
calvinjia: debug level and use `{}` syntax

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2580,6 +2581,61 @@ private void loadMetadataIfNotExist(RpcContext rpcContext, LockedInodePath inode
...
+      String ufsUri = mountInfo.getUfsUri().toString();
+      try (UnderFileSystem ufs = UnderFileSystem.Factory.create(ufsUri, ufsConf)) {
+        // check the root path is valid
+        ufs.getFileStatus(ufsUri);
+      } catch (Throwable t) {
+        throw new InvalidArgumentException("Failed to mount under file system."
+            + " Please ensure the options are valid", t);
+      }
+      unmountInternal(rpcContext, inodePath, false);
+      long mountId = IdUtils.createMountId();
```
calvinjia: We modify the mount id without journaling?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2580,6 +2581,61 @@ private void loadMetadataIfNotExist(RpcContext rpcContext, LockedInodePath inode
...
+      try {
+        mountInternal(rpcContext, inodePath, mountInfo.getUfsUri(), mountId, context, true);
+      } catch (Throwable t) {
+        // With ignoreError = true we should not normally get an exception. At this point we failed
+        // to add the mount point back to the table so we need to remove all child inodes and
+        // nested mount.
+        try {
+          deleteInternal(rpcContext, inodePath, DeleteContext
+              .mergeFrom(DeletePOptions.newBuilder().setRecursive(true).setAlluxioOnly(true)));
```
calvinjia: It is a bit counter intuitive if failing the update causes the mount the be unmounted

shell/src/main/java/alluxio/cli/fs/command/UpdateMountCommand.java
```diff
@@ -0,0 +1,120 @@
...
+  @Override
+  public String getUsage() {
+    return "updateMount [--readonly] [--shared] [--option <key=val>] <alluxioPath>";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Updates options for a mount point.\n"
+        + "WARNING: updating mount point with incorrect options may cause mount point to"
+        + " become unusable or unmounted.";
```
calvinjia: This seems like a very steep cost for using the command, given that there is no option to dry-run it

tests/src/test/java/alluxio/client/cli/fs/command/UpdateMountCommandTest.java
```diff
@@ -0,0 +1,184 @@
...
+    AlluxioURI mountPoint = new AlluxioURI("/mnt");
+    String ufsPath = mFolder.getRoot().getAbsolutePath();
+    Assert.assertEquals(0,
+        mFsShell.run("mount", "--option", "key=k=v", mountPoint.toString(), ufsPath));
+    Assert.assertEquals(0,
+        mFsShell.run("updateMount", "--option", "key=k=v2", mountPoint.toString()));
+  }
+
+  @Test
+  public void updateRootMountFailed() throws Exception {
```
calvinjia: Do you want to validate that it deletes the mount point cleanly?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: The only down side I can think of is the interruption of current UFS operations. Since the updateMount request is equivalent to user doing an unmount and a mount immediate after, it will immediately invalidate existing mount id, causing any client operation using the old mount id to fail. I talked to @ns1123 and he is okay with the behavior.

The reason I want to use a different mount Id is for worker to invalidate the cached UFS and get a new updated UFS client. If we use the same mount id then there has to be a new RPC call to inform worker of the update.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: yeah, that's why I added all the extra code to:

1. creating a UFS client to validate the new options.
2. disable most of the checks that can cause a mount to fail.

Then, if the mount operation still fails(this will be highly unlikely), we have to do some clean up to make sure that we don't have the file metadata existing for a non-exist mount point. So this is a second level fail-safe mechanism. 

Most of the time if user specifies some invalid options for the mount point they will get an error message and the UFS should still work as before and its options are not updated.

shell/src/main/java/alluxio/cli/fs/command/UpdateMountCommand.java
bf8086: I already added a dry run within this command (on the server side), so 99% of the time the old mount point will continue to work even if a user put something invalid in the command. But given the possibilities this command enables (we don't have control over what options they updates and that can be any valid HDFS configuration property), a user can still potentially make the UFS client malfunction in a way that even the dry run cannot detect. This is just to make sure they understand the potential impact of this command before running it.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
bf8086: We already log similar information for mount and unmount and I feel it can be useful for understanding the state of the file system. If you think it can be too chatty, I can demote all similar log entries.

tests/src/test/java/alluxio/client/cli/fs/command/UpdateMountCommandTest.java
bf8086: I actually don't know how to get it to fail that bad.
Normally if an option does not work it will be captured in the dry run logic at  https://github.com/Alluxio/alluxio/pull/9107/files#diff-7f82122f923de058e3c4a20b43cd7e4eR2613
The only thing I can think of that will get the mount point deleted is if in the future we accidentally add some check that throws even when `ignoreError` is set to true.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
bf8086: Actually it seems `mount` is the only other command that logs with info. I will update the level for both.

bf8086:
@gpang PTAL. Thanks!

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: I see, which services currently cache the mount id? I think the worker does?

bf8086: yep.

bf8086: The work ufs manager maintains a UFS client cache which will reuse the clients of the same id.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2674,16 +2730,24 @@ private void mountInternal(Supplier<JournalContext> journalContext, LockedInodeP
...
-      // Check that the alluxioPath we're creating doesn't shadow a path in the parent UFS
-      MountTable.Resolution resolution = mMountTable.resolve(alluxioPath);
-      try (CloseableResource<UnderFileSystem> ufsResource =
-               resolution.acquireUfsResource()) {
-        String ufsResolvedPath = resolution.getUri().getPath();
-        if (ufsResource.get().exists(ufsResolvedPath)) {
-          throw new IOException(
-              ExceptionMessage.MOUNT_PATH_SHADOWS_PARENT_UFS.getMessage(alluxioPath,
-                  ufsResolvedPath));
+      if (!ignoreError) {
```
calvinjia: Why do we allow this in the `ignoreError` case?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2727,31 +2791,37 @@ public void unmount(AlluxioURI alluxioPath) throws FileDoesNotExistException,
...
     if (!inodePath.fullPathExists()) {
       throw new FileDoesNotExistException(
           "Failed to unmount: Path " + inodePath.getUri() + " does not exist");
     }
+    MountInfo mountInfo = mMountTable.getMountTable().get(inodePath.getUri().getPath());
+    if (mountInfo == null) {
+      throw new InvalidPathException("Failed to unmount " + inodePath.getUri() + ". Please ensure"
+          + " the path is an existing mount point.");
+    }
+    mSyncManager.stopSyncForMount(mountInfo.getMountId());
```
calvinjia: Does this mean if the user updates mount for a previously active synced mount point, they will need to set up the active sync again?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: Are you talking about the code from `2719` to `2739`? That is because the logic updates mount options based on a API call result from a UFS client, so skipping it will result in incorrect mount options.

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
```diff
@@ -165,15 +167,17 @@ public boolean delete(Supplier<JournalContext> journalContext, AlluxioURI uri) {
     try (LockResource r = new LockResource(mWriteLock)) {
       if (mState.getMountTable().containsKey(path)) {
         // check if the path contains another nested mount point
-        for (String mountPath : mState.getMountTable().keySet()) {
-          try {
-            if (PathUtils.hasPrefix(mountPath, path) && (!path.equals(mountPath))) {
-              LOG.warn("The path to unmount {} contains another nested mountpoint {}",
-                  path, mountPath);
-              return false;
+        if (checkNestedMount) {
```
calvinjia: Why do we need to introduce this flag?

bf8086: So basically if user execute this `updateMount`command on a mount point that contains a nested mount. The below check would fail because for a normal unmount we do not want to unmount if a nested mount point exist. In our case for updateMount command we do not want to limit the user to only be able to update mount point that does not have a nested mount, so this logic is disable in the `updatedMount` case. It is safe in this case because the `updateMount` command will keep the inode tree under the mount point so the nested mount point can still exist after the command.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: yeah looks like this command will remove all the sync points under the mount point. I can try adding them back after the remount is completed.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: Will the worker gracefully handle needing the refresh the cache after it fails once with the old mount id?

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
calvinjia: This seems quite tricky in cases of partial success, if there is a partial mount, will we even be able to delete the data (alluxio only) in the nested mount?

bf8086: yeah at the end of the `updateMount` function it will capture any exception and delete the subtree which also removes any nested mount along the way.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: no. The current expectation is that the updateMount is a interruptive command similar to unmount which will cause all existing data access request to fail due to the mount id invalidation.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2632,6 +2642,22 @@ public void updateMount(AlluxioURI alluxioPath, MountContext context)
           throw t;
         }
       }
+      if (syncPoints != null && !syncPoints.isEmpty()) {
+        // add back all sync points
+        Exception syncEx = null;
```
calvinjia: Do we do anything with `syncEx`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2632,6 +2642,22 @@ public void updateMount(AlluxioURI alluxioPath, MountContext context)
...
+        for (AlluxioURI syncPoint : syncPoints) {
+          try {
+            startSyncAndJournal(rpcContext, syncPoint);
```
calvinjia: Is this equivalent to keeping the sync point, or will we incur a large cost of restarting the sync (same as if you needed to add a mount as a new sync point)

bf8086: There will be some cost equivalent to adding the sync points again, but that is necessary because we have to reinitiate the sync hook using the new UFS client.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: Good catch! Fixed.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2644,19 +2644,22 @@ public void updateMount(AlluxioURI alluxioPath, MountContext context)
...
-              syncEx = new UnexpectedAlluxioException("Some sync points in this mount point"
+              syncEx = new InvalidPathException("Some sync points in this mount point"
                   + " failed to initialize. Please add them manually.");
             }
             syncEx.addSuppressed(new InvalidPathException(
                 String.format("Failed to add sync point %s", syncPoint), e));
           }
         }
+        if (syncEx != null) {
+          throw syncEx;
```
calvinjia: I'm wary of relying on the exception to hint that there was a partial success, especially if another `InvalidPathException` could mean that the entire operation failed.

bf8086: Do we have a convention on how non-fatal warnings are returned?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2580,6 +2581,90 @@ private void loadMetadataIfNotExist(RpcContext rpcContext, LockedInodePath inode
...
+          }
+        } catch (Throwable t) {
+          throw new InvalidArgumentException("Failed to mount under file system."
+              + " Please ensure the options are valid", t);
+        }
+        unmountInternal(rpcContext, inodePath, false);
+      }
+      long mountId = IdUtils.createMountId();
+      try {
+        mountInternal(rpcContext, inodePath, mountInfo.getUfsUri(), mountId, context, true);
```
gpang: should there be a `updateMountInternal` and an associated `updateMount` in the mount table? It looks like we are reusing delete and add, but they are not performed atomically, which could cause strange behavior?

shell/src/main/java/alluxio/cli/fs/command/UpdateMountCommand.java
```diff
@@ -0,0 +1,120 @@
...
+import java.io.IOException;
+import java.util.Properties;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Update options for an Alluxio mount point.
+ */
+@ThreadSafe
+public final class UpdateMountCommand extends AbstractFileSystemCommand {
```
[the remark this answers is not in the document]
bf8086: I thought about it and there are a couple reasons for using a separate command:
- In mount command the user has to specify a UFS URI; That is not necessary for updateMount.
- We already have `mount` and `unmount ` instead of a single command with `-u` option. In the spirit of "Do One Thing And Do It Well", I made it a separate command.
- It is harder to describe the command and usage if we have the mount command and updateMount command combined.
- It's easier to maintain if we later wants to add options to only one of the command.
What do you think?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: I want to reuse the `unmountInternal` and `mountInternal` to minimize the risk of having a new code path and get out of sync in the future when we add new logic to only one of the function. 

Regarding atomicity, the current implementation use a single write lock on the mount point and also does a dry run to load the UFS client with new configuration options before making change to the mount table, so it should be safe to use 99% of the time. If we think it is crucial to achieve atomicity for the rest 1% of the cases, I can refactor the mount table to support it but I feel that will also make mount table less maintainable.

shell/src/main/java/alluxio/cli/fs/command/UpdateMountCommand.java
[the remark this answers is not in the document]
gpang: Ok, sounds good.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: My concern would not be that other threads have write access to the inode, but rather, the update mount operation itself. It is represented by 2 journal entries in sequence: unmount and mount. If something happens in between, then the resulting state would be very inconsistent?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2727,18 +2786,22 @@ public void unmount(AlluxioURI alluxioPath) throws FileDoesNotExistException,
    *
    * @param rpcContext the rpc context
    * @param inodePath the Alluxio path to unmount, must be a mount point
+   * @param removeInodes whether to remove inodes after mount point is deleted
```
gpang: no longer needed

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
```diff
@@ -186,6 +190,40 @@ public boolean delete(Supplier<JournalContext> journalContext, AlluxioURI uri) {
...
+        throw new InvalidPathException(String.format("Failed to update mount point at %s."
+            + " Please ensure the path is an existing mount point and not root.",
+            alluxioUri.getPath()));
+      }
+      try {
+        add(journalContext, alluxioUri, mountInfo.getUfsUri(), newMountId, newOptions);
+      } catch (FileAlreadyExistsException | InvalidPathException e) {
+        // This should never happen since the path is guaranteed to exist and the mount point is
+        // just removed from the same path.
+        LOG.error("Failed to add back the mount point at {}", alluxioUri, e);
```
gpang: This message is misleading. This failed at adding the updated mount point, not the old one, right? Maybe "Failed to add the updated mount point at "

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: Removed.

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
bf8086: Updated.

bf8086:
@gpang @calvinjia I updated the PR to address the comments. PTAL.

calvinjia:

calvinjia:

PR #9110, opened 2019-05-18
Use default options in HdfsFileInputStream

cheng-chang:
The original code will only use cluster configurations, which suppresses path configurations in openFile. This behavior will cause long running clients like Presto to never be able to use path level configurations even after restarting.


calvinjia:

PR #9111, opened 2019-05-18
Allow properties with scope broader than client for path config

cheng-chang:


calvinjia:

PR #9112, opened 2019-05-19
Run alluxio-start.sh local without creating alluxio-site.properties

apc999:
Previously, it requires redudant steps to start a local & single node cluster:
- Set `alluxio.master.hostname=localhost` in `conf/alluxio-site.properties`
- Pass `local` as an arg to `alluxio-start.sh`

After this PR, running `alluxio-start.sh local` implies setting `alluxio.master.hostname=localhost`. Therefore reduce the steps to test-drive alluxio in a local mode

### DIFF
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -516,6 +516,11 @@ main() {
       start_proxies
       ;;
     local)
+      ALLUXIO_MASTER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_WORKER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_PROXY_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_JOB_MASTER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_JOB_WORKER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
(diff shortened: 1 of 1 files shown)


bin/alluxio-start.sh
```diff
@@ -516,6 +516,11 @@ main() {
       start_proxies
       ;;
     local)
+      ALLUXIO_MASTER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_WORKER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_PROXY_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_JOB_MASTER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
+      ALLUXIO_JOB_WORKER_JAVA_OPTS+=" -Dalluxio.master.hostname=localhost"
```
calvinjia: Is there no common Java opt we can use? Also what will the behavior be if the user does specify a master address?

apc999: - setting `ALLUXIO_JAVA_OPTS` will not work as `ALLUXIO_MASTER_JAVA_OPTS` will be set in `libexec/config.sh` before reaching here. 
- `alluxio-start.sh local` will overite the master hostname setting in `alluxio-site.properties` which I think should be the correct behavior (when arg `local` is conflicting with parameters in `alluxio-site.properties` )

calvinjia:

calvinjia:

PR #9115, opened 2019-05-20
Fix generate-tarballs - exclude generated restdoc from license check

ZacBlanco:
If we don't exclude `./generated` the `generate-tarballs` script
will fail when building UFS modules because of the license check
after compiling the codebase


calvinjia:

PR #9122, opened 2019-05-22
Replace atomic counter with CountLatch in FileSystemContextReinitializer

cheng-chang:
This PR adds a new synchronization primitive: CountLatch.
After replacing the atomic counter with CountLatch, the new behavior of reinitializing FileSystemContext is:
If reinit is called while there are ongoing RPCs, reinit blocks until a time point where there is no active RPC. When reinit is happening, new RPCs are blocked until reinit succeeds or fails.


gpang:
@cheng-chang This should target master branch, and then be cherry-picked to other branches.

cheng-chang:
@gpang reopened against master in https://github.com/Alluxio/alluxio/pull/9123

PR #9134, opened 2019-05-24
Add a white list to the monitor script

calvinjia:
Otherwise running `bin/alluxio-start.sh logserver` or `safe` will fail due to monitoring not being supported.

### DIFF
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -594,9 +594,12 @@ main() {
     wait
   fi
 
-  if [[ ! "${async}" ]]; then
-    start_monitor "${ACTION}" "${MONITOR_NODES}"
-  fi
+  case "${ACTION}" in
+      all | local | master | masters | secondary_master | job_master | job_masters | proxy | proxies | worker | workers | job_worker | job_workers)
+    if [[ ! "${async}" ]]; then
+      start_monitor "${ACTION}" "${MONITOR_NODES}"
+    fi
+  esac
 }
 
 main "$@"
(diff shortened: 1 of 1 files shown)


bin/alluxio-start.sh
```diff
@@ -594,9 +594,12 @@ main() {
     wait
   fi
 
-  if [[ ! "${async}" ]]; then
-    start_monitor "${ACTION}" "${MONITOR_NODES}"
-  fi
+  case "${ACTION}" in
+      all | local | master | masters | secondary_master | job_master | job_masters | proxy | proxies | worker | workers | job_worker | job_workers)
```
ZacBlanco: Making a comment to point out the fact that if we ever add/remove a monitor type in `alluxio-monitor.sh`, this list will need to be updated. I don't really see clear way to do this without calling the monitor script in the first place though

calvinjia: Another alternative is to just no-op in the monitor call if it is invalid instead of throwing an error.

ZacBlanco: The proper behavior of the script should be to `exit 1` in the case that an option passed isn't valid.

One option could be to add a function to the script `alluxio-monitor.sh checkValid` that will print `true/false` and we can check that output within the start script instead.

calvinjia: In the start script, currently we have

```
  elif [[ "${action}" == "logserver" || "${action}" == "safe" ]]; then
    echo -e "Error: Invalid Monitor ACTION: ${action}" >&2
    exit 1
  fi
```
We could just no-op instead of exit 1, I agree that `alluxio-monitor.sh` should exit 1 if invalid args are passed.

calvinjia:

calvinjia:

PR #9143, opened 2019-05-25
Fix locality in containerized environments

madanadit:
In a containerized env such as k8s, the worker and client hostname may not match. To determine if a worker is local, file system inspection is used instead when property `alluxio.worker.data.server.domain.socket.as.uuid=true` is set.

In this PR:
- update the method to choose a local worker for BlockInStream/OutStream
- if client container does not mount domain socket dir but has the same hostname as worker, do not attempt to open local file short circuit stream if worker supports domain socket (as /dev/shm is not accessible)
- unmount as part of fuse startup in case k8s shutdown hook is flaky

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
@@ -48,7 +48,7 @@ public final class BlockLocationUtils {
     if (conf.getBoolean(PropertyKey.WORKER_DATA_SERVER_DOMAIN_SOCKET_AS_UUID)) {
       // Determine by inspecting the file system if worker is local
       for (WorkerNetAddress addr : addresses) {
-        if (NettyUtils.isDomainSocketSupported(addr, conf)) {
+        if (NettyUtils.isDomainSocketAccessible(addr, conf)) {
           LOG.debug("Found local worker by file system inspection of path {}",
               addr.getDomainSocketPath());
           // Returns the first local worker and does not shuffle
--- a/core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
@@ -11,12 +11,6 @@
 
 package alluxio.client.block.util;
 
-import java.util.List;
-import java.util.Optional;
-import java.util.stream.Collectors;
-
-import javax.annotation.concurrent.ThreadSafe;
-
 import alluxio.Constants;
 import alluxio.collections.Pair;
 import alluxio.conf.AlluxioConfiguration;
@@ -25,6 +19,12 @@ import alluxio.util.TieredIdentityUtils;
 import alluxio.util.network.NettyUtils;
 import alluxio.wire.TieredIdentity;
 import alluxio.wire.WorkerNetAddress;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.annotation.concurrent.ThreadSafe;
 
 /**
  * Utility functions for working with block locations.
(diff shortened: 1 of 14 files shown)


core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
```diff
@@ -0,0 +1,77 @@
...
+@ThreadSafe
+public final class BlockLocationUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(BlockLocationUtils.class);
+
+  /**
+   * @param tieredIdentity the tiered identity
+   * @param addresses the candidate worker addresses
+   * @param conf Alluxio configuration
+   * @return the address closest to this one. If none of the identities match, the first address is
+   *         returned
```
calvinjia: Also include that the boolean hints as to whether or not the location is local.

core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
```diff
@@ -0,0 +1,77 @@
...
+  public static Optional<Pair<WorkerNetAddress, Boolean>> nearest(TieredIdentity tieredIdentity,
+      List<WorkerNetAddress> addresses, AlluxioConfiguration conf) {
+    if (conf.getBoolean(PropertyKey.WORKER_DATA_SERVER_DOMAIN_SOCKET_AS_UUID)) {
+      // Determine by inspecting the file system if worker is local
+      for (WorkerNetAddress addr : addresses) {
+        if (NettyUtils.isDomainSocketAccessible(addr, conf)) {
+          LOG.debug("Found local worker by file system inspection of path {}",
+              addr.getDomainSocketPath());
+          // Returns the first local worker and does not shuffle
+          return Optional.of(new Pair<>(addr, true));
```
calvinjia: If the domain socket is not accessible, won't we still execute the code below which likely will return `true` based on the locality match?

madanadit: if domain socket is not accessible, we choose nearest worker based on locality match (isLocal is potentially true if client and worker have the same hostname -> as intended)

madanadit:
jenkins, test this please.

madanadit:
@calvinjia do you have further comments?

core/client/fs/src/main/java/alluxio/client/block/util/BlockLocationUtils.java
calvinjia: Ah I see, we have another check for domain socket availability later.

core/client/fs/src/test/java/alluxio/client/block/policy/LocalFirstPolicyTest.java
```diff
@@ -120,6 +131,33 @@ public void chooseClosestTier() throws Exception {
     assertEquals("node4", chosen.getTieredIdentity().getTier(0).getValue());
   }
 
+  @Test
+  public void chooseLocalAccessibleDomainSocket() throws Exception {
```
calvinjia: Does this test belong in `LocalFirstPolicyTest`?

madanadit: it doesn't matter much as `LocalFirstPolicy::nearest` is a thin wrapper over `BlockLocationUtils::nearest`. i'll move it to a to a new class for `BlockLocationUtilsTest` to be more granular

calvinjia: Thanks, I was more concerned about putting the test in a reasonable place and not the test correctness. It seemed weird to test for this in a policy since this should work regardless of which policy is used.

madanadit: done

calvinjia:
@madanadit LGTM, could you fix the checkstyle failures?

madanadit:
jenkins, test this please.

madanadit:

madanadit:

Issue #9089, opened 2019-05-13
fuse client  the alluxio worker doesn't go through short-circuit in the same node in Kubernetes


cheyang:
**Alluxio Version:**
2.0.0-snapshot

**Describe the bug**
Deploy master, workers and fuse in Kubernetes.

[allxuio.zip](https://github.com/Alluxio/alluxio/files/3175058/allxuio.zip)

And go to the machine A(192.168.0.109), copy whole the directory from alluxio to local directory

```
time cp -r /alluxio-fuse/training-data/images /var/lib/docker/test
```

But I noticed the file is not located in the same node, it's in 192.168.0.110:

```
/opt/alluxio/bin/alluxio fs stat /training-data/images/validation-00108-of-00128
/training-data/images/validation-00108-of-00128 is a file path.
FileInfo{fileId=35802578943, name=validation-00108-of-00128, path=/training-data/images/validation-00108-of-00128, ufsPath=oss://imagenet-huabei5/images/validation-00108-of-00128, length=50475323, blockSizeBytes=536870912, creationTimeMs=1557791864620, completed=true, folder=false, pinned=false, cacheable=true, persisted=true, blockIds=[35785801728], inMemoryPercentage=0, lastModificationTimesMs=1557109076000, ttl=-1, ttlAction=DELETE, owner=, group=, mode=511, persistenceState=PERSISTED, mountPoint=false, replicationMax=-1, replicationMin=0, fileBlockInfos=[FileBlockInfo{blockInfo=BlockInfo{id=35785801728, length=50475323, locations=[BlockLocation{workerId=4272994094758843366, address=WorkerNetAddress{host=192.168.0.110, rpcPort=29998, dataPort=29998, webPort=29996, domainSocketPath=/opt/domain/2956ad2e-f596-4cbb-a31e-9a92d934b6d4, tieredIdentity=TieredIdentity(node=192.168.0.110, rack=null)}, tierAlias=SSD}]}, offset=0, ufsLocations=[]}], mountId=344496208015086572, inAlluxioPercentage=100, ufsFingerprint=TYPE|FILE UFS|oss OWNER|_ GROUP|_ MODE|511 CONTENT_HASH|373069F926A32F9FAF9DA7B4E36A0B34 , acl=user::rwx,group::rwx,other::rwx, defaultAcl=}
Containing the following blocks:
[...]

PR #9149, opened 2019-05-28
Implement authentication long polling

ggezer:
`AuthenticationServer` has a setting for revoking authentication after a period of inactivity. To handle that on the client side, metadata clients, after a period of inactivity, will retry after getting `Unauthenticated` code. However, due to nature of streaming, data clients can not retry after getting the error because they might have pipelined more data before seeing the error. And since this revocation will not change the connection state, they used to continue getting `Unauthenticated`. See https://github.com/Alluxio/alluxio/issues/9089 for an instance of this problem.

This PR introduces long polling to authentication handshake. Client and server will not complete streams used for authentication and instead will use it for notifying end of an authentication session. With this change, revocation on server will be propagated to client channel via health status, causing a client recreation for later use of the same channel. Also client closing the channel will notify server and it'll clean its state of the recently closed channel. 

Periodic cleanup has not been disabled in order to not prolong a duration for a channel to remain authenticated.

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -34,6 +34,7 @@ public final class GrpcChannel extends Channel {
   private final GrpcManagedChannelPool.ChannelKey mChannelKey;
   private final Supplier<Boolean> mChannelHealthState;
   private Channel mChannel;
+  private Runnable mAuthCloseCallback;
   private boolean mChannelReleased;
   private boolean mChannelHealthy = true;
   private final long mShutdownTimeoutMs;
@@ -51,6 +52,11 @@ public final class GrpcChannel extends Channel {
     mChannelHealthState = channel instanceof AuthenticatedChannel
         ? () -> (((AuthenticatedChannel) channel).isAuthenticated() && mChannelHealthy)
             : () -> mChannelHealthy;
+    if (channel instanceof AuthenticatedChannel) {
+      // Store {@link AuthenticatedChannel::#close) for signaling end of
+      // authenticated session during shutdown.
+      mAuthCloseCallback = ((AuthenticatedChannel) channel)::close;
+    }
     mChannel = ClientInterceptors.intercept(channel, new ChannelResponseTracker((this)));
     mChannelReleased = false;
     mShutdownTimeoutMs = shutdownTimeoutMs;
@@ -79,6 +85,10 @@ public final class GrpcChannel extends Channel {
    * Shuts down the channel.
    */
   public void shutdown() {
+    if (mAuthCloseCallback != null) {
+      // Stop authenticated session with server.
+      mAuthCloseCallback.run();
+    }
     if (!mChannelReleased) {
       GrpcManagedChannelPool.INSTANCE().releaseManagedChannel(mChannelKey, mShutdownTimeoutMs);
     }
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -32,7 +32,7 @@ import java.util.function.Supplier;
  */
 public final class GrpcChannel extends Channel {
   private final GrpcManagedChannelPool.ChannelKey mChannelKey;
-  private final Supplier<Boolean> mChannelHealthState;
+  private Supplier<Boolean> mChannelHealthState;
   private Channel mChannel;
   private Runnable mAuthCloseCallback;
   private boolean mChannelReleased;
@@ -49,17 +49,29 @@ public final class GrpcChannel extends Channel {
   public GrpcChannel(GrpcManagedChannelPool.ChannelKey channelKey, Channel channel,
       long shutdownTimeoutMs) {
     mChannelKey = channelKey;
-    mChannelHealthState = channel instanceof AuthenticatedChannel
-        ? () -> (((AuthenticatedChannel) channel).isAuthenticated() && mChannelHealthy)
-            : () -> mChannelHealthy;
-    if (channel instanceof AuthenticatedChannel) {
-      // Store {@link AuthenticatedChannel::#close) for signaling end of
-      // authenticated session during shutdown.
-      mAuthCloseCallback = ((AuthenticatedChannel) channel)::close;
-    }
+    mChannelHealthState = () -> mChannelHealthy;
     mChannel = ClientInterceptors.intercept(channel, new ChannelResponseTracker((this)));
     mChannelReleased = false;
     mShutdownTimeoutMs = shutdownTimeoutMs;
+  }
+
+  /**
+   * Create a new instance of {@link GrpcChannel} with an authenticated channel.
+   *
+   * @param channelKey the channel key
+   * @param channel the authenticated grpc channel
+   * @param shutdownTimeoutMs shutdown timeout in milliseconds
+   */
+  public GrpcChannel(GrpcManagedChannelPool.ChannelKey channelKey, AuthenticatedChannel channel,
+      long shutdownTimeoutMs) {
+    this(channelKey, (Channel) channel, shutdownTimeoutMs);
+    // Update the channel health supplier for authenticated channel.
+    mChannelHealthState =
+        () -> (((AuthenticatedChannel) channel).isAuthenticated() && mChannelHealthy);
+
+    // Store {@link AuthenticatedChannel::#close) for signaling end of
+    // authenticated session during shutdown.
+    mAuthCloseCallback = ((AuthenticatedChannel) channel)::close;
   }
 
   @Override
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -66,8 +66,7 @@ public final class GrpcChannel extends Channel {
       long shutdownTimeoutMs) {
     this(channelKey, (Channel) channel, shutdownTimeoutMs);
     // Update the channel health supplier for authenticated channel.
-    mChannelHealthState =
-        () -> (((AuthenticatedChannel) channel).isAuthenticated() && mChannelHealthy);
+    mChannelHealthState = () -> (channel.isAuthenticated() && mChannelHealthy);
 
     // Store {@link AuthenticatedChannel::#close) for signaling end of
     // authenticated session during shutdown.
--- a/core/common/src/main/java/alluxio/grpc/GrpcServer.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcServer.java
@@ -13,7 +13,9 @@ package alluxio.grpc;
 
 import alluxio.retry.ExponentialBackoffRetry;
 import alluxio.retry.RetryUtils;
+import alluxio.security.authentication.AuthenticationServer;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.grpc.Server;
 
 import java.io.IOException;
@@ -25,6 +27,7 @@ import java.util.concurrent.TimeUnit;
  */
 public final class GrpcServer {
   private Server mServer;
+  private AuthenticationServer mAuthServer;
 
   /** Set to TRUE when the server has been successfully started. **/
   private boolean mStarted = false;
@@ -34,11 +37,21 @@ public final class GrpcServer {
    * Create a new instance of {@link GrpcServer}.
    *
    * @param server the wrapped server
+   * @param authServer the authentication server
    * @param serverShutdownTimeoutMs server shutdown timeout in milliseconds
    */
-  public GrpcServer(Server server, long serverShutdownTimeoutMs) {
+  public GrpcServer(Server server, AuthenticationServer authServer, long serverShutdownTimeoutMs) {
     mServer = server;
+    mAuthServer = authServer;
     mServerShutdownTimeoutMs = serverShutdownTimeoutMs;
+  }
+
+  /**
+   * @return the authentication server associated with this server
+   */
+  @VisibleForTesting
+  public AuthenticationServer getAuthenticationServer() {
+    return mAuthServer;
   }
 
   /**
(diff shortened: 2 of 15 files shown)


core/common/src/main/java/alluxio/grpc/GrpcChannel.java
```diff
@@ -51,6 +52,11 @@ public GrpcChannel(GrpcManagedChannelPool.ChannelKey channelKey, Channel channel
     mChannelHealthState = channel instanceof AuthenticatedChannel
         ? () -> (((AuthenticatedChannel) channel).isAuthenticated() && mChannelHealthy)
             : () -> mChannelHealthy;
+    if (channel instanceof AuthenticatedChannel) {
```
gpang: we do various things for the instance of `AuthenticatedChannel`. Can we just have a bool for that?

Also, what other types of channels (not AuthenticatedChannel) are possible? All this checking looks like it can be abstracted away.

core/common/src/main/java/alluxio/grpc/GrpcServer.java
```diff
@@ -34,13 +37,23 @@
...
     mServer = server;
+    mAuthServer = authServer;
     mServerShutdownTimeoutMs = serverShutdownTimeoutMs;
   }
 
+  /**
+   * @return the authentication server associated with this server
+   */
+  @VisibleForTesting
+  public AuthenticationServer getAuthenticationServer() {
```
gpang: Can't the test just pass in the authentication server?

tests/src/test/java/alluxio/client/fuse/FuseFileSystemIntegrationTest.java
```diff
@@ -256,6 +256,31 @@ public void tail() throws Exception {
     Assert.assertEquals("Test File Content\n", result);
   }
 
+  @Test
+  @LocalAlluxioClusterResource.Config(
+      confParams = {PropertyKey.Name.AUTHENTICATION_INACTIVE_CHANNEL_REAUTHENTICATE_PERIOD, "3sec"})
```
gpang: can we shorten this (and the sleep)?

core/common/src/main/java/alluxio/security/authentication/ChannelAuthenticator.java
```diff
@@ -127,12 +128,14 @@ public Channel authenticate(GrpcServerAddress serverAddress, ManagedChannel mana
     private final GrpcServerAddress mServerAddress;
     private final ManagedChannel mManagedChannel;
     private Channel mChannel;
-    private boolean mAuthenticated;
+    private AtomicBoolean mAuthenticated;
```
gpang: can you add some comments for these two variables?

core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
```diff
@@ -125,6 +125,13 @@ public void unregisterChannel(UUID channelId) {
         LOG.warn("Failed to dispose sasl client for channel-Id: {}. Error: {}", channelId,
             e.getMessage());
       }
+
+      try {
+        channelInfo.getSaslServerDriver().onCompleted();
+      } catch (Exception e) {
+        LOG.warn("Failed to authentication session for channel-Id: {}. Error: {}", channelId,
```
gpang: NIT: what does this message mean?

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
```diff
@@ -41,20 +42,24 @@
...
+  /** Whether the authentication is active. */
+  private AtomicBoolean mAuthenticationActive;
 
   private final long mGrpcAuthTimeoutMs;
 
   /**
    * Creates client driver with given handshake handler.
    *
    * @param handshakeClient client handshake handler
+   * @param authenticationActive boolean reference to receive authentication state changes
```
gpang: Is this class doing the "receiving" of the state change? It looks like the state is set here?

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
```diff
@@ -72,10 +77,11 @@ public void onNext(SaslMessage saslMessage) {
       LOG.debug("SaslClientDriver received message: {}",
           saslMessage != null ? saslMessage.getMessageType().toString() : "<NULL>");
       SaslMessage response = mSaslHandshakeClientHandler.handleSaslMessage(saslMessage);
-      if (response == null) {
-        mRequestObserver.onCompleted();
```
gpang: `onCompleted` is not required anymore?

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
```diff
@@ -41,20 +42,24 @@
   private SaslHandshakeClientHandler mSaslHandshakeClientHandler;
   /** Used to wait until authentication is completed. */
   private SettableFuture<Boolean> mAuthenticated;
+  /** Whether the authentication is active. */
```
gpang: can you add comments on how this is different from `mAuthenticated` and how they interact? Are both required? Having both is very confusing.

core/common/src/main/java/alluxio/grpc/GrpcServer.java
ggezer: `GrpcServer` is built by `ServerBuilder` and it dynamically creates the auth server. Removing this would mean adding a new method to builder with a slightly more complex code.

core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
ggezer: Oh. It's gotta be "Failed to **complete** the authentication ......"

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
ggezer: This class knows whether the client channel is authenticated or not. So from client's perspective this class manages the state, end purpose of this parameter is to receive that current state from outside.

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
ggezer: onCompleted() calls for both client and server are reserved now for long polling. Calling them would indicate that the end of authenticated session. 
When called from client, it means "I don't want to be authenticated anymore". 
When called from server, it means, for the client, that "I'm not authenticated anymore."

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
ggezer: `Channel` is not our class so can't replace `AuthenticatedChannel` with a boolean. I did some changes for better abstracting the authenticated channel handling.

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
ggezer: SettableFuture is unfortunately not resettable, so it can't be used to reflect state changes following the initial state. I changed names and comments to try to make it less confusing.

ggezer:
@gpang Updated PR after your feedback. PTAL.

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
gpang: I see, I thought it meant this class (SaslStreamClientDriver) is receiving the state, but what it means is the caller is receiving the state via this bool. Thanks for the clarification.

However, can we rename to `authenticated` to match the internal variable name?

core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
gpang: I see. Would it be useful to have this information somewhere in the comments? I'm just not sure where is the best place to put this info.

gpang:
jenkins, test this please

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

core/common/src/test/java/alluxio/security/authentication/GrpcSecurityTest.java
```diff
@@ -131,6 +140,64 @@ public void testAuthMismatch() throws Exception {
...
+    // Grab internal authenticated channel reference.
+    Channel interceptorChannel = Whitebox.getInternalState(channel, "mChannel");
+    AuthenticatedChannel authenticatedChannel =
+        Whitebox.getInternalState(interceptorChannel, "channel");
+    // Assert that authentication server has a login info for the channel.
+    Assert.assertNotNull(server.getAuthenticationServer()
+        .getUserInfoForChannel(authenticatedChannel.getChannelId()));
+    // Close the underlying authenticated channel.
+    authenticatedChannel.close();
+    // Assert that authentication server don't contain login info for the channel anymore.
```
bf8086: (nit) `don't` -> `doesn't`

gpang:
@ggezer I think this looks good, but the tests keep failing

core/common/src/test/java/alluxio/security/authentication/GrpcSecurityTest.java
```diff
@@ -131,6 +140,64 @@ public void testAuthMismatch() throws Exception {
...
+     * Sleeping will ensure that authentication sessions for the channel will
+     * expire on the server. This should have propagated back to the client and
+     * its health status should reflect that.
+     *
+     * Sleep more than authentication revocation timeout.
+     */
+    Thread.sleep(500);
+    Assert.assertFalse(channel.isHealthy());
+
+    server.shutdown();
```
bf8086: (nit) should this be in a try-final block?

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
```diff
@@ -48,14 +49,31 @@
...
+   * @param channelKey the channel key
+   * @param channel the authenticated grpc channel
+   * @param shutdownTimeoutMs shutdown timeout in milliseconds
+   */
+  public GrpcChannel(GrpcManagedChannelPool.ChannelKey channelKey, AuthenticatedChannel channel,
+      long shutdownTimeoutMs) {
+    this(channelKey, (Channel) channel, shutdownTimeoutMs);
+    // Update the channel health supplier for authenticated channel.
+    mChannelHealthState =
+        () -> (((AuthenticatedChannel) channel).isAuthenticated() && mChannelHealthy);
```
bf8086: (nit) is the explicit cast necessary?

core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
```diff
@@ -73,11 +73,11 @@ public DefaultAuthenticationServer(String hostName, AlluxioConfiguration conf) {
     mHostName = hostName;
     mConfiguration = conf;
     mCleanupIntervalMs =
-            conf.getMs(PropertyKey.AUTHENTICATION_INACTIVE_CHANNEL_REAUTHENTICATE_PERIOD);
+        conf.getMs(PropertyKey.AUTHENTICATION_INACTIVE_CHANNEL_REAUTHENTICATE_PERIOD);
```
bf8086: Shall we bump up the default reauthentication period now that the closed channels are cleaned up automatically?

ggezer: That would prolong duration at which clients could remain authenticated. Are we fine with that? I'm mostly in favor to revoke auth after a period of inactivity.

bf8086: yeah I would think the period should be somewhat larger than the time to complete a normal user task. That should be no more than a couple days. Any concern for making it 3 days?

core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
```diff
@@ -122,8 +123,15 @@ public void unregisterChannel(UUID channelId) {
...
-        LOG.warn("Failed to dispose sasl client for channel-Id: {}. Error: {}", channelId,
-            e.getMessage());
+        LogUtils.warnWithException(LOG,
+            String.format("Failed to dispose sasl client for channel-Id: {}", channelId), e);
+      }
+
+      try {
+        channelInfo.getSaslServerDriver().onCompleted();
+      } catch (Exception e) {
+        LogUtils.warnWithException(LOG, String.format(
```
bf8086: Remove `String.format`: `LogUtils.warnWithException(LOG, "Failed to complete the authentication session for channel-Id: {}", channelId, e);`

core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
```diff
@@ -122,8 +123,15 @@ public void unregisterChannel(UUID channelId) {
       try {
         channelInfo.getSaslServer().dispose();
       } catch (SaslException e) {
-        LOG.warn("Failed to dispose sasl client for channel-Id: {}. Error: {}", channelId,
-            e.getMessage());
+        LogUtils.warnWithException(LOG,
+            String.format("Failed to dispose sasl client for channel-Id: {}", channelId), e);
```
bf8086: Remove `String.format`: `LogUtils.warnWithException(LOG, "Failed to dispose sasl client for channel-Id: {}", channelId, e);`

core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
ggezer: That should be OK.

gpang:

bf8086:
@gpang Can you help cherry-pick this PR to 2.0 branch?

calvinjia:

PR #9153, opened 2019-05-28
Update LockCache's eviction behavior to avoid deadlock

cheng-chang:
Previously, when the soft limit is reached, and no lock can be evicted, calls to get and tryGet will deadlock because the eviction thread will be blocked in evictIfOverLimit because no lock can be evicted, while other threads will block waiting for the eviction thread to finish.

This is a temporary fix for 2.0 release. The root cause is that in master, some operations will try to hold locks for all inodes and edges under a path recursively, and these locks will only be released when the operation finishes. This causes a deadlock because if the number of locks to be acquired is larger than the LockCache size, then no lock can be acquired,  so the operation is blocked, consequently, the existing locks can not be released.

A better solution is to add an iterator interface to InodeTree to lock and unlock paths while traversing.

### DIFF
--- a/core/common/src/main/java/alluxio/collections/LockCache.java
+++ b/core/common/src/main/java/alluxio/collections/LockCache.java
@@ -48,18 +48,11 @@ import java.util.function.Function;
  */
 public class LockCache<K> {
   private static final Logger LOG = LoggerFactory.getLogger(LockCache.class);
-  private static final long OVER_HARD_LIMIT_LOG_INTERVAL = 60000; // milliseconds
   private static final float DEFAULT_LOAD_FACTOR = 0.75f;
   private static final float SOFT_LIMIT_RATIO = 0.9f;
-  /**
-   * Eviction happens whenever the evictor thread is signaled,
-   * or is blocked for this period of time.
-   */
-  private static final int EVICTION_MAX_AWAIT_TIME = 30000; // milliseconds
   private static final String EVICTOR_THREAD_NAME = "LockCache Evictor";
 
   private final Map<K, ValNode> mCache;
-  private Iterator<Map.Entry<K, ValNode>> mIterator;
   /**
    * SoftLimit = hardLimit * softLimitRatio
    * once the size reaches softlimit, eviction starts to happen,
@@ -71,7 +64,6 @@ public class LockCache<K> {
 
   private final Lock mEvictLock = new ReentrantLock();
   private final Condition mOverSoftLimit = mEvictLock.newCondition();
-  private volatile long mLastOverHardLimitTime = 0;
   private final ExecutorService mEvictor;
 
   /**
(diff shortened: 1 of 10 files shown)


tests/src/test/java/alluxio/client/fs/FreeAndDeleteIntegrationTest.java
```diff
@@ -127,4 +130,20 @@ public void freeAndDeleteIntegration() throws Exception {
     // Verify the blocks are not in mLostBlocks.
     assertTrue(bm.getLostBlocks().isEmpty());
   }
+
+  /**
+   * Tests that deleting a directory with number of files larger than maximum lock cache size will
+   * not be blocked.
+   */
+  @Test(timeout = 3000)
+  public void deleteDir() throws Exception {
```
cheng-chang: without this PR, this test will timeout.

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -70,45 +72,28 @@ public LockCache(Function<? super K, ? extends ReentrantReadWriteLock> defaultLo
...
-          if (candidate.mIsAccessed) {
-            candidate.mIsAccessed = false;
-          } else {
-            if (candidate.mRefCount.compareAndSet(0, Integer.MIN_VALUE)) {
-              // the value object can be evicted, at the same time we make refCount minValue
-              mIterator.remove();
-              numToEvict--;
-            }
-          }
+    try (LockResource l = new LockResource(mEvictLock)) {
```
gpang: should this be a try-lock, since blocking will stop all progress for other threads?

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -267,14 +219,12 @@ public int getSoftLimit() {
   /**
    * Node containing value and other information to be stored in the cache.
    */
-  public class ValNode {
-    private ReentrantReadWriteLock mValue;
-    private boolean mIsAccessed;
```
gpang: there is no more clock approximation?

tests/src/test/java/alluxio/client/fs/FreeAndDeleteIntegrationTest.java
cheng-chang: it's enough on my machine

core/common/src/main/java/alluxio/collections/LockCache.java
cheng-chang: I intentionally make it block other threads since I don't want to lose chances for eviction.
But after second thought, try-lock works too.

core/common/src/main/java/alluxio/collections/LockCache.java
cheng-chang: I think the reference counter is enough, what's the point of this boolean?

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -174,51 +202,24 @@ public ReentrantReadWriteLock getRawReadWriteLock(K key) {
...
     Preconditions.checkNotNull(key, "key can not be null");
-    ValNode cacheEntry;
-    while (true) {
-      // repeat until we get a cacheEntry that is not in the process of being removed.
-      cacheEntry = mCache.compute(key, (k, v) -> {
-        if (v != null) {
-          // we want to maintain the invariant that only one thread is removing from the map.
-          // Returning anything other than v would break that invariant.
-          v.mRefCount.incrementAndGet();
-          v.mIsAccessed = true;
```
cheng-chang: if the above two lines are executed between `mRefCount.compareAndSet(0, INT_MIN)` and `mIterator.erase` in `evictIfOverLimit`, then the returned lock is no longer in the cache, so a further `get` for the same key will get a different lock, which causes locking issues.

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -174,51 +203,24 @@ public ReentrantReadWriteLock getRawReadWriteLock(K key) {
...
+        v.mIsAccessed = true;
+        return v;
       }
-      evictIfOverLimit();
-      if (cacheEntry.mRefCount.get() > 0) {
-        return cacheEntry;
-      } else {
+      return new ValNode(mDefaultLoader.apply(k));
+    });
+    if (mCache.size() > mSoftLimit) {
```
gpang: we probably want to redefine what soft and hard limits mean. I think this should signal the evictor if the hard limit is exceeded, to avoid needing to evict constantly around soft limit.

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -71,44 +89,56 @@ public LockCache(Function<? super K, ? extends ReentrantReadWriteLock> defaultLo
...
-          if (candidate.mIsAccessed) {
-            candidate.mIsAccessed = false;
-          } else {
-            if (candidate.mRefCount.compareAndSet(0, Integer.MIN_VALUE)) {
-              // the value object can be evicted, at the same time we make refCount minValue
-              mIterator.remove();
-              numToEvict--;
-            }
+  private void evictIfOverLimit() throws InterruptedException {
+    try (LockResource l = new LockResource(mEvictLock)) {
```
gpang: Do we need to hold the lock the whole time while evicting? We only have 1 evictor thread, and we only need the lock to receive the signal from other threads. Can't we just use the lock to detect if evictor should run, and then release it?

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -267,14 +269,14 @@ public int getSoftLimit() {
...
-    private boolean mIsAccessed;
+  private static final class ValNode {
+    private final ReentrantReadWriteLock mValue;
+    private volatile boolean mIsAccessed;
     private AtomicInteger mRefCount;
 
     private ValNode(ReentrantReadWriteLock value) {
       mValue = value;
-      mIsAccessed = false;
+      mIsAccessed = true;
```
gpang: why does this start as `true` now?

cheng-chang: after second thought, changed to false

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -25,36 +26,53 @@
...
   private final int mHardLimit;
   private final int mSoftLimit;
   private final Function<? super K, ? extends ReentrantReadWriteLock> mDefaultLoader;
-  private final Lock mEvictLock;
 
-  private Iterator<Map.Entry<K, ValNode>> mIterator;
-  private long mLastSizeWarningTime;
+  private final Lock mEvictLock = new ReentrantLock();
+  private final Condition mOverSoftLimit = mEvictLock.newCondition();
+  private volatile long mLastOverHardLimitTime = 0;
```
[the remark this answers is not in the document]
cheng-chang: added

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -71,44 +89,56 @@ public LockCache(Function<? super K, ? extends ReentrantReadWriteLock> defaultLo
...
-            candidate.mIsAccessed = false;
-          } else {
-            if (candidate.mRefCount.compareAndSet(0, Integer.MIN_VALUE)) {
-              // the value object can be evicted, at the same time we make refCount minValue
-              mIterator.remove();
-              numToEvict--;
-            }
+  private void evictIfOverLimit() throws InterruptedException {
+    try (LockResource l = new LockResource(mEvictLock)) {
+      while (mCache.size() <= mSoftLimit) {
```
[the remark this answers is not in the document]
cheng-chang: refactored into `Evictor` class, and renamed to `awaitAndEvict`

core/common/src/main/java/alluxio/collections/LockCache.java
cheng-chang: softlimit is 0.9 * hardlimit, reaching softlimit is like a warning signal that hardlimit will be reached soon, so doing eviction around softlimit makes sense.

core/common/src/main/java/alluxio/collections/LockCache.java
cheng-chang: It makes no difference. We can release the lock once the signal is received, then the `tryLock` in other `get` can lock and signal the condition, but if the eviction thread is in the while loop, the signal will be a no-op anyway. So if eviction is happening, it can keep holding the lock, because even if other threads can acquire the lock, they cannot trigger the eviction anyway.

cheng-chang:
@gpang PTAL

cheng-chang:
jenkins, test this please

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -70,45 +80,86 @@ public LockCache(Function<? super K, ? extends ReentrantReadWriteLock> defaultLo
...
               mIterator.remove();
               numToEvict--;
             }
           }
         }
-      } finally {
-        mEvictLock.unlock();
+        if (mCache.size() >= mHardLimit) {
+          if (System.currentTimeMillis() - mLastOverHardLimitTime > OVER_HARD_LIMIT_LOG_INTERVAL) {
+            LOG.warn("LockCache at hard limit, cache size = " + mCache.size()
```
gpang: can we use the `{}` notation?

core/common/src/main/java/alluxio/collections/LockCache.java
gpang: I think hard and soft limit should behave like high and low watermarks. When greater than high watermark, evict down to the lower watermark.

core/common/src/main/java/alluxio/collections/LockCache.java
```diff
@@ -174,51 +180,23 @@ public ReentrantReadWriteLock getRawReadWriteLock(K key) {
...
+        return v;
       }
-      evictIfOverLimit();
-      if (cacheEntry.mRefCount.get() > 0) {
-        return cacheEntry;
-      } else {
+      return new ValNode(mDefaultLoader.apply(k));
+    });
+    if (mCache.size() > mSoftLimit) {
+      if (mEvictLock.tryLock()) {
```
yuzhu: It seems to me that `mEvictLock.trylock` would never succeed while the evicting thread is awaiting? because the eviction thread would be holding the lock?

yuzhu: nvm, should have read the javadoc more closely

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1371,22 +1371,28 @@ public String toString() {
...
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_LOCKCACHE_MAXSIZE =
-      new Builder(Name.MASTER_LOCKCACHE_MAXSIZE)
-          .setDefaultValue(100000)
-          .setDescription("Maximum inodelock cache size")
+  public static final PropertyKey MASTER_LOCK_POOL_LOW_WATERMARK =
+      new Builder(Name.MASTER_LOCK_POOL_LOW_WATERMARK)
+          .setDefaultValue(500000)
+          .setDescription("Low watermark of lock pool size")
```
gpang: Please explain how the low and high watermark work/interact.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1371,22 +1371,28 @@ public String toString() {
...
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_LOCK_POOL_HIGH_WATERMARK =
+      new Builder(Name.MASTER_LOCK_POOL_HIGH_WATERMARK)
+          .setDefaultValue(1000000)
+          .setDescription("High watermark of lock pool size")
```
gpang: Please explain how the low and high watermark work/interact.

core/server/master/src/main/java/alluxio/master/file/meta/InodeLockManager.java
```diff
@@ -121,7 +123,7 @@ public void assertAllLocksReleased() {
     assertAllLocksReleased(mInodeLocks);
   }
 
-  private <T> void assertAllLocksReleased(LockCache<T> cache) {
+  private <T> void assertAllLocksReleased(LockPool<T> cache) {
```
gpang: NIT: rename to pool

core/common/src/main/java/alluxio/collections/LockPool.java
```diff
@@ -0,0 +1,293 @@
...
+          }
+        }
+        if (mPool.size() >= mHighWatermark) {
+          if (System.currentTimeMillis() - mLastOverHighWatermarkTime
+              > OVER_HIGH_WATERMARK_LOG_INTERVAL) {
+            LOG.warn("LockPool size grows over high watermark: "
+                + "pool size = {}, low watermark = {}, high watermark = {}",
+                mPool.size(), mLowWatermark, mHighWatermark);
+          }
+          mLastOverHighWatermarkTime = System.currentTimeMillis();
```
gpang: this should be within the previous if-block, right?

cheng-chang:
jenkins, test this please

cheng-chang:
@gpang @yuzhu PTAL

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

cheng-chang:
jenkins, test this please

yuzhu:
jenkins, test this please

gpang:
Jenkins, test this please

gpang:
@cheng-chang overall, this looks good, but the tests keep failing.

cheng-chang:
jenkins, test this please

cheng-chang:
@gpang JournalToolTest is known to be flaky.

cheng-chang:
@gpang test passed

gpang:

bf8086:
@gpang Can you help cherry-pick this PR to 2.0 branch?

calvinjia:

calvinjia:

PR #9174, opened 2019-05-29
Prevent async eviction of ASYNC_THROUGH because of race condition

madanadit:
When a new block is written to Alluxio it becomes a candidate for eviction. The worker fetches the pinned list (including to be persisted file ids) from the master in a heartbeat to prevent eviction of files which are not yet persisted. If a worker is running near storage capacity, a to be persisted block may be evicted before the worker heartbeat fetches the pinned list from the master.

To overcome this issue, let’s look at different scenarios when a block is created on the worker TieredBlockStore. For each of these scenarios if the block create request contains the pinned flag at create time, we can prevent eviction.
 
- Client short-circuit writes to local worker: alluxio.grpc.CreateLocalBlockRequest will need an additional field
- Client remote writes to worker: alluxio.grpc.WriteRequest will need an additional field
- Worker reads from remote worker: We do not inherit the pinned flag from a remote worker.
- Worker reads from ufs: No changes are required for this case as block was already persisted

Proposed Solution:
- When a block is created in TieredBlockStore, the BlockMetadataManagerView is updated with an additional pinned id if the flag was set
- On the next heartbeat, the pinned list is overwritten with the list from the master => the create request flag only impacts current heartbeat interval

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataWriter.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataWriter.java
@@ -156,9 +156,7 @@ public final class GrpcDataWriter implements DataWriter {
       builder.setCreateUfsBlockOptions(ufsBlockOptions);
     }
     // check if we need to pin block on create
-    if (options.getWriteType() == WriteType.ASYNC_THROUGH) {
-      builder.setPinned(true);
-    }
+    builder.setPinOnCreate(options.getWriteType() == WriteType.ASYNC_THROUGH);
     mPartialRequest = builder.buildPartial();
     mChunkSize = chunkSize;
     mClient = client;
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataWriter.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataWriter.java
@@ -11,6 +11,7 @@
 
 package alluxio.client.block.stream;
 
+import alluxio.client.WriteType;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.options.OutStreamOptions;
 import alluxio.conf.AlluxioConfiguration;
@@ -153,6 +154,10 @@ public final class GrpcDataWriter implements DataWriter {
           .setMountId(options.getMountId())
           .setFallback(alreadyFallback).build();
       builder.setCreateUfsBlockOptions(ufsBlockOptions);
+    }
+    // check if we need to pin block on create
+    if (options.getWriteType() == WriteType.ASYNC_THROUGH) {
+      builder.setPinned(true);
     }
     mPartialRequest = builder.buildPartial();
     mChunkSize = chunkSize;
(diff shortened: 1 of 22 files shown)


core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -549,6 +552,12 @@ private BlockStoreLocation commitBlockInternal(long sessionId, long blockId)
           | WorkerOutOfSpaceException e) {
         throw Throwables.propagate(e); // we shall never reach here
       }
+
+      // Check if block is pinned on commit
+      if (isPinned) {
+        updatePinnedInodes(Collections.singleton(BlockId.getFileId(blockId)));
```
calvinjia: I think we need a different API, this is currently an override operation, not a union

madanadit:
@calvinjia @bf8086 this PR is ready for review for the most part except new test cases

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataWriter.java
```diff
@@ -154,6 +155,10 @@ private GrpcDataWriter(FileSystemContext context, final WorkerNetAddress address
           .setFallback(alreadyFallback).build();
       builder.setCreateUfsBlockOptions(ufsBlockOptions);
     }
+    // check if we need to pin block on create
+    if (options.getWriteType() == WriteType.ASYNC_THROUGH) {
+      builder.setPinned(true);
+    }
```
calvinjia: Should we use the `builder.setPinned(options.getWriteType() == ...)` pattern here as well?

core/server/worker/src/main/java/alluxio/worker/grpc/WriteRequest.java
```diff
@@ -25,11 +25,15 @@
   /** This ID can either be block ID or temp UFS file ID. */
   private final long mId;
 
+  /** Is block pinned to create. */
```
calvinjia: `on create`?

madanadit: +1

madanadit:
Thanks @calvinjia. Adding the test cases. All comments make sense

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -883,6 +892,18 @@ public void updatePinnedInodes(Set<Long> inodes) {
     }
   }
 
+  /**
+   * Add a single inode to set of pinned ids.
+   *
+   * @param inode an inode that is pinned
+   */
+  private void addToPinnedInodes(Long inode) {
```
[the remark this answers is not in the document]
madanadit: nope

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
```diff
@@ -81,9 +81,9 @@ public static LocalFileDataWriter create(final FileSystemContext context,
       long dataTimeout = conf.getMs(PropertyKey.USER_NETWORK_DATA_TIMEOUT_MS);
 
       CreateLocalBlockRequest.Builder builder =
-          CreateLocalBlockRequest.newBuilder().setBlockId(blockId)
-              .setTier(options.getWriteTier()).setSpaceToReserve(fileBufferByes)
-              .setMediumType(options.getMediumType());
+          CreateLocalBlockRequest.newBuilder().setBlockId(blockId).setTier(options.getWriteTier())
+              .setSpaceToReserve(fileBufferByes).setMediumType(options.getMediumType())
```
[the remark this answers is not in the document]
madanadit: done

madanadit:
Thanks @bf8086, @calvinjia. Addressed the comments

madanadit:

madanadit:

PR #9186, opened 2019-06-01
Do harden exception handling for inode crawler

ggezer:
In current code there are cases where failure during inode buffering could go unnoticed. This could  raise false positives while taking a backup, and cause backup file to not contain the complete master state. This fix is to ensure buffering failures are not overlooked during enumeration.


calvinjia:

calvinjia:

PR #9192, opened 2019-06-02
Support "s3://" in addition to "s3a://"

apc999:
# Motivation
- Users should not need to understand the difference between all s3 related Hadoop FSes (s3, s3a, s3n) when using Alluxio. All they need to know is to specify the S3 bucket using URL like `s3://bucketname/dir`
- Some S3 related properties are `s3a` and some others are `s3`, very confusing to hard to remember.

# PR Summary
"s3a" is a Hadoop client-side concept to denote a HDFS compatible implementation, whereas when mounting external s3 bucket into Alluxio, it has nothing to do with Hadoop client.  We should support the canonical S3 bucket URL by "s3://" 

# Changes include
- Support "s3://" in addition to "s3a://" in UFS URL
- Improve property names to be "alluxio.s3.foo" rather than "alluxio.s3a.foo" to be consistent and make "alluxio.s3a.foo" the alias.
- Update corresponding docs


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3643,30 +3655,30 @@ private static String javadocLink(String fullyQualifiedClassname) {
...
-    public static final String UNDERFS_S3A_STREAMING_UPLOAD_ENABLED =
-        "alluxio.underfs.s3a.streaming.upload.enabled";
-    public static final String UNDERFS_S3A_STREAMING_UPLOAD_PARTITION_SIZE =
-        "alluxio.underfs.s3a.streaming.upload.partition.size";
+    public static final String UNDERFS_S3_BULK_DELETE_ENABLED =
+        "alluxio.underfs.s3.bulk.delete.enabled";
+    public static final String UNDERFS_S3_DEFAULT_MODE = "alluxio.underfs.s3.default.mode";
+    public static final String UNDERFS_S3_DIRECTORY_SUFFIX =
+        "alluxio.underfs.s3.directory.suffix";
+    public static final String UNDERFS_S3_INHERIT_ACL = "alluxio.underfs.s3.inherit.acl";
```
apc999: @calvinjia I changed this property to `alluxio.underfs.s3.inherit.acl` to be consistent with other property names. Let me know if you are good with it

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -588,7 +598,7 @@ private ObjectPermissions getPermissionsInternal() {
 
   @Override
   protected String getRootKey() {
-    return Constants.HEADER_S3A + mBucketName;
```
apc999: this string concatenation is very __inefficient__ as `getRootKey` gets called on almost every S3 UFS ops, multiple times even. @calvinjia , I would even move this up to `ObjectStoreUFS` to share the optimization across all object stores.

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -91,6 +91,11 @@
   /** AWS-SDK S3 client. */
   private final AmazonS3Client mClient;
 
+  // TODO(binfan): make this a member of ObjectUnderFileSystem rather than a calculated string
+  // each time on getRootKey() for efficiency
```
calvinjia: Isn't this already done?

calvinjia: Oh I see, this is only for S3A at the moment. It would be nice to put this in the abstract class if this performance improvement is nontrivial.

apc999: Yes, will do that in a follow up PR to minimize the change of this one

apc999:
@calvinjia  addressed your comments. PTAL

calvinjia:
@apc999 LGTM, please fix the checkstyle and merge conflicts.

apc999:

apc999:

PR #9197, opened 2019-06-03
Avoid unnecessary logs from user injector

ggezer:


ggezer:
@bf8086 PTAL. Thanks.

core/common/src/main/java/alluxio/security/authentication/AuthenticatedUserInjector.java
```diff
@@ -102,13 +104,31 @@ public void onHalfClose() {
...
+   * This is mostly to avoid dumping "already closed" exceptions to logs.
+   *
+   * @param call call to close
+   * @param status status to close the call with
+   * @param headers headers to close the call with
+   */
+  private <ReqT, RespT> void safeClose(ServerCall<ReqT, RespT> call, Status status,
+      Metadata headers) {
+    try {
+      call.close(status, headers);
```
bf8086: Add a debug logging message for the authentication error on the server side?

ggezer: Could you clarify your ask? (This is server side)

ggezer: If that's what you means; "Status: {}" in the below debug print will contain with what the call was being closed for.

bf8086: Right now the authentication error message is only reported to the client side through the Status object. We should consider adding debug logging on the server side so even if we don't have the client side message we still have a way to know what went wrong with authentication. This should be logged regardless whether the close of the call result in an exception or not.

ggezer: Got it. Added trace before closing the call.

ggezer:
Jenkins, test this please

core/common/src/main/java/alluxio/security/authentication/AuthenticatedUserInjector.java
```diff
@@ -102,13 +104,33 @@ public void onHalfClose() {
...
+
+  /**
+   * Closes the call while blanketing runtime exceptions.
+   * This is mostly to avoid dumping "already closed" exceptions to logs.
+   *
+   * @param call call to close
+   * @param status status to close the call with
+   * @param headers headers to close the call with
+   */
+  private <ReqT, RespT> void safeClose(ServerCall<ReqT, RespT> call, Status status,
```
calvinjia: Usually this is called `closeQuietly`?

ggezer: Sounds better.

calvinjia:

calvinjia:

PR #9198, opened 2019-06-03
Throw away sasl client/server handlers after authentication

ggezer:


ggezer:
@bf8086 PTAL. Thanks.

core/common/src/main/java/alluxio/security/authentication/SaslStreamServerDriver.java
```diff
@@ -81,9 +82,15 @@ public void onNext(SaslMessage saslMessage) {
...
       // Because client will assume authenticated after receiving the success message.
       if (response.getMessageType() == SaslMessageType.SUCCESS) {
+        // Register authorized user with the authentication server.
         mAuthenticationServer.registerChannel(mChannelId,
-            mSaslServerHandler.getAuthenticatedUserInfo(), this,
-            mSaslServerHandler.getSaslServer());
+            mSaslServerHandler.getAuthenticatedUserInfo(), this);
+        // Finished with the handler.
+        try {
+          mSaslServerHandler.close();
```
bf8086: Should we also close the handle when channel is not authenticated successfully?

ggezer: Definitely. Fixed.

core/common/src/main/java/alluxio/security/authentication/SaslStreamServerDriver.java
```diff
@@ -111,6 +120,13 @@ public void onError(Throwable throwable) {
       LOG.warn("Closing authenticated channel: {} due to error: {}", mChannelId, throwable);
       mAuthenticationServer.unregisterChannel(mChannelId);
     }
+    if (mSaslServerHandler != null) {
+      try {
+        mSaslServerHandler.close();
+      } catch (IOException exc) {
```
bf8086: Do we also need to close the handler when `SaslStreamServerDriver.close()` is called?

ggezer:
Jenkins, test this please.

core/common/src/main/java/alluxio/security/authentication/SaslStreamServerDriver.java
ggezer: No, SaslServer will be closed before reaching there for both success and failure paths.

ggezer: I couldn't find any cleanar way for serverdriver to close its handler. Now closing it during close call as well.

core/common/src/main/java/alluxio/security/authentication/plain/SaslClientHandlerPlain.java
```diff
@@ -92,4 +96,15 @@ public ChannelAuthenticationScheme getClientScheme() {
...
     return mSaslClient;
   }
+
+  @Override
+  public void close() {
+    if (mSaslClient != null) {
+      try {
+        mSaslClient.dispose();
+      } catch (SaslException exc) {
+        LOG.debug("Failed to close SaslClient.", exc);
```
calvinjia: Is it a problem if we fail to close the client or server?

ggezer: Not really. But I don't expect that to happen often, that's why the debug trace.

calvinjia:

calvinjia:

PR #9201, opened 2019-06-03
Fix remove function for ControllableQueue

bf8086:
There is a bug in `ControllableQueue.remove` where it is trying to remove the item from priority queue which holds a list of wrappers. This change fixed it by comparing the value of the wrapper to the item instead.

### DIFF
--- a/core/common/src/main/java/alluxio/util/executor/ControllableQueue.java
+++ b/core/common/src/main/java/alluxio/util/executor/ControllableQueue.java
@@ -89,7 +89,7 @@ public class ControllableQueue<T> {
    * @return true if the element removed successfully, false otherwise
    */
   public boolean remove(T element) {
-    return mQueue.removeIf(x -> x.mValue == element);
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/util/executor/ControllableQueue.java
```diff
@@ -89,7 +89,7 @@ public T pop() {
    * @return true if the element removed successfully, false otherwise
    */
   public boolean remove(T element) {
-    return mQueue.remove(element);
+    return mQueue.removeIf(x -> x.mValue == element);
```
gpang: NIT: use `getValue()` instead of accessing private member?

bf8086: Done.

bf8086:
Jenkins, test this please.

gpang:

PR #9206, opened 2019-06-04
[SMALLFIX] Change INFO to DEBUG logging of removed block information

LuQQiu:
The master log is easy to be full of the removed block information if we use INFO logging.


LuQQiu:
@gpang I do many create/delete file operations on the cloud, the master.log is full of the information of removed block. It will be better to put it in debug logging. PTAL, thanks

PR #9208, opened 2019-06-04
[SMALLFIX] Fix the inputStream.read() in runUfsTests

LuQQiu:
Some `inputStream.read()` throws IOE exception when unable to read instead of returning -1.
This test consider the object store eventual consistency and allow retry `inputStream.read()` when fail to read.

### DIFF
--- a/examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
+++ b/examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
@@ -26,6 +26,7 @@ import alluxio.util.CommonUtils;
 import alluxio.util.UnderFileSystemUtils;
 import alluxio.util.WaitForOptions;
 import alluxio.util.io.PathUtils;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
+++ b/examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
@@ -218,7 +218,7 @@ public final class UnderFileSystemCommonOperations {
       try {
         bytesRead = inputStream.read(buf, offset, buf.length - offset);
       } catch (Exception e) {
-        LOG.info("Failed to read from file {}", testFile, e);
+        LOG.info("Failed to read from file {}", testFile, e.toString());
         bytesRead = -1;
       }
       if (bytesRead != -1) {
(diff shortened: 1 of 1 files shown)


examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
```diff
@@ -210,7 +214,13 @@ public void createOpenLargeTest() throws IOException {
     int offset = 0;
     int noReadCount = 0;
     while (offset < buf.length && noReadCount < 3) {
-      int bytesRead = inputStream.read(buf, offset, buf.length - offset);
+      int bytesRead;
+      try {
+        bytesRead = inputStream.read(buf, offset, buf.length - offset);
+      } catch (Exception e) {
+        LOG.info("Failed to read from file {}", testFile, e);
```
calvinjia: Does logging make sense or should we just print to console?

LuQQiu: In console, if this will cause test to fail, we should print out. But here, read throws exception because of eventual consistency is kind of expected. 
All logging is write to user.log to make console look cleaner, otherwise object store will has tons of logging

calvinjia: Makes sense, for this log message, let's use `e.toString`, otherwise this will print the entire stack trace which is generally reserved for `log.error`

examples/src/main/java/alluxio/examples/UnderFileSystemCommonOperations.java
```diff
@@ -218,7 +218,7 @@ public void createOpenLargeTest() throws IOException {
       try {
         bytesRead = inputStream.read(buf, offset, buf.length - offset);
       } catch (Exception e) {
-        LOG.info("Failed to read from file {}", testFile, e);
+        LOG.info("Failed to read from file {}", testFile, e.toString());
```
calvinjia: Need an additional `{}` in the log message.

LuQQiu:
Tests in error: 
alluxio.client.cli.fs.command.DistributedMvCommandTest.crossMountMove(alluxio.client.cli.fs.command.DistributedMvCommandTest)
  Run 1: DistributedMvCommandTest>AbstractFileSystemShellTest.before:123 » Runtime Time...
  Run 2: DistributedMvCommandTest>AbstractFileSystemShellTest.after:132 » NullPointer
  Run 3: DistributedMvCommandTest.crossMountMove:41->run:50 Runtime Failed command <dis...
  Run 4: DistributedMvCommandTest.crossMountMove:41->run:50 Runtime Failed command <dis...

LuQQiu:
Jenkins, test this please

LuQQiu:
@calvinjia Addressed the comments, PTAL, thanks!

calvinjia:

calvinjia:

PR #9226, opened 2019-06-06
Implement close() for BlockStore interface

ggezer:
Leaving `RocksBlockStore` unclosed, will block creating further stores at the same path. This prevents running fail-over scenarios in single process tests with ROCKS meta-store. This PR makes necessary plumbing to make sure `RocksBlockStore` is closed during master shutdown.

### DIFF
--- a/core/server/common/src/main/java/alluxio/Registry.java
+++ b/core/server/common/src/main/java/alluxio/Registry.java
@@ -140,12 +140,22 @@ public class Registry<T extends Server<U>, U> {
   }
 
   /**
-   * Stops all {@link Server}s in reverse dependency order. If A depends on B, B is stopped
-   * before A.
+   * Stops all {@link Server}s in reverse dependency order. If A depends on B, A is stopped
+   * before B.
    */
   public void stop() throws IOException {
     for (T server : Lists.reverse(getServers())) {
       server.stop();
+    }
+  }
+
+  /**
+   * Closes all {@link Server}s in reverse dependency order. If A depends on B, A is closed
+   * before B.
+   */
+  public void close() throws IOException {
+    for (T server : Lists.reverse(getServers())) {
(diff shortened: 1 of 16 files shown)


core/server/common/src/main/java/alluxio/Registry.java
```diff
@@ -149,6 +149,16 @@ public void stop() throws IOException {
     }
   }
 
+  /**
+   * Closes all {@link Server}s in reverse dependency order. If A depends on B, B is closed
+   * before A.
```
calvinjia: Is this accurate? Shouldn't we close A before B?

ggezer: Code does the right thing, but the doc was wrong.

ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

ggezer:
@LuQQiu @calvinjia 
PTAL. Thanks!

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
```diff
@@ -208,10 +208,16 @@ protected void startMasters(boolean isLeader) {
 
   /**
    * Stops all masters, including block master, fileSystem master and additional masters.
+   *
+   * @param isClose {@code true} if process is exiting
    */
-  protected void stopMasters() {
+  protected void stopMasters(boolean isClose) {
```
LuQQiu: Will it be better to use `closeMasters()` instead of a boolean flag?

ggezer: Makes sense.

calvinjia:

calvinjia:

PR #9230, opened 2019-06-07
Propagate xattr from UfsStatus to Inode

ZacBlanco:
This change aims to allow the calls to the UFS from the
FileSystemMaster to propagate into the inode tree. This will
allow developers in the future to leverage any xattr that are
returned from the UFS implementations in order to make
optimizations or other assumptions about the state of a node
past the standard set of fields.

Most of these changes are mechanical, it's just a matter of
finding exactly where to set the xattr and where not to.

For now, the only usecase for setting xattr is during sync or
loadMetadata calls. The implementation adds the xattr to the
create*Context so that the inode tree can utilize set the xattr
on inodes from within the inode tree class.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/contexts/CreatePathContext.java
+++ b/core/server/master/src/main/java/alluxio/master/file/contexts/CreatePathContext.java
@@ -27,6 +27,7 @@ import com.google.protobuf.GeneratedMessageV3;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Wrapper for {@link CreateFilePOptions} or {@link CreateDirectoryPOptions} with additional context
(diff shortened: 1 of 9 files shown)


core/server/master/src/main/java/alluxio/master/file/contexts/CreatePathContext.java
```diff
@@ -258,6 +261,22 @@ public K setMetadataLoad(boolean metadataLoad) {
     return getThis();
   }
 
+  /**
+   * @return extended attributes on this context
+   */
+  public Map<String, byte[]> getXAttr() {
```
gpang: does this need the `@Nullable` annotation?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -668,6 +670,7 @@ public InodeDirectory getRoot() {
           mState.applyAndJournal(rpcContext, UpdateInodeEntry.newBuilder()
               .setId(currentId)
               .setLastModificationTimeMs(context.getOperationTimeMs())
+              .putAllXAttr(CommonUtils.convertToByteString(context.getXAttr()))
```
gpang: Can `context.getXAttr()` be `null`?

gpang:
Jenkins, test this please

gpang:

calvinjia:

PR #9239, opened 2019-06-08
Use lazy calculation on root key for object storage

apc999:


underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -575,7 +565,11 @@ private ObjectPermissions getPermissionsInternal() {
 
   @Override
   protected String getRootKey() {
-    return mRootKey;
+    if (mUri.getScheme() != null && mUri.getScheme().equals("s3a")) {
```
calvinjia: `"s3a".equals(mUri.getScheme())`?
Should we also check for `s3`, is there any risk of the uri scheme not being `s3a` or `s3`?

apc999: - updated
- not really possible to be any other scheme rather than `s3a` or `s3`

calvinjia:

calvinjia:

PR #9261, opened 2019-06-11
Add documentation for extracting k8s artifacts from docker image

madanadit:
In this PR:
- instructions for extract k8s specs and alluxio client jar from docker image

### DIFF
--- a/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
+++ b/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
@@ -281,9 +281,10 @@ func generateTarball(hadoopClients []string) error {
 		replace(filepath.Join("integration/checker", file), "target/alluxio-checker-${VERSION}-jar-with-dependencies.jar", "alluxio-checker-${VERSION}.jar")
 	}
 	// Update the kubernetes image tags
-	for _, file := range []string{"alluxio-master.yaml.template", "alluxio-worker.yaml.template", "alluxio-fuse.yaml.template"} {
+	for _, file := range []string{"alluxio-master.yaml.template", "alluxio-worker.yaml.template"} {
 		replace(filepath.Join("integration/kubernetes", file), "alluxio/alluxio:latest", fmt.Sprintf("alluxio/alluxio:%v", version))
 	}
+	replace("integration/kubernetes/alluxio-fuse.yaml.template", "alluxio/alluxio-fuse:latest", fmt.Sprintf("alluxio/alluxio-fuse:%v", version))
 
 	mvnArgs := getCommonMvnArgs(hadoopVersion)
 	run("compiling repo", "mvn", mvnArgs...)
--- a/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
+++ b/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
@@ -280,6 +280,10 @@ func generateTarball(hadoopClients []string) error {
 	for _, file := range []string{"bin/hive-checker.sh", "bin/mapreduce-checker.sh", "bin/spark-checker.sh"} {
 		replace(filepath.Join("integration/checker", file), "target/alluxio-checker-${VERSION}-jar-with-dependencies.jar", "alluxio-checker-${VERSION}.jar")
 	}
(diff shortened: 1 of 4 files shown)


docs/_config.yml
```diff
@@ -12,6 +12,8 @@ ALLUXIO_MASTER_VERSION_SHORT: master
 # For release candidates and release branches, this should be the major Alluxio version, e.g. both 1.5.0 and 1.5.0-RC1 should use "1.5"
 # For master branch, this should be "master"
 ALLUXIO_MAJOR_VERSION: master
+# The version string must match the version string part of the client jar path
+ALLUXIO_VERSION_STRING: 2.0.0-SNAPSHOT
```
calvinjia: Could we avoid adding this new variable?

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -280,6 +280,11 @@ func generateTarball(hadoopClients []string) error {
 	for _, file := range []string{"bin/hive-checker.sh", "bin/mapreduce-checker.sh", "bin/spark-checker.sh"} {
 		replace(filepath.Join("integration/checker", file), "target/alluxio-checker-${VERSION}-jar-with-dependencies.jar", "alluxio-checker-${VERSION}.jar")
 	}
+	// Update the kubernetes image tags
+	for _, file := range []string{"alluxio-master.yaml.template", "alluxio-worker.yaml.template"} {
+		replace(filepath.Join("integration/kubernetes", file), "alluxio/alluxio:latest", fmt.Sprintf("alluxio/alluxio:%v", version))
+	}
+	replace("integration/kubernetes/alluxio-fuse.yaml.template", "alluxio/alluxio-fuse:latest", fmt.Sprintf("alluxio/alluxio-fuse:%v", version))
```
calvinjia: Should this be done in the release branch similar to how we update for the DockerFile?

madanadit: not sure i follow. do you mean manually updating the ARG_TARBALL in the release branch? 
and, why do you prefer the alternate approach? to me it seems like all generated tarballs would benefit from having the correct tag

calvinjia: Yeah updating the version in the release branch. The difference would be if they check out the tag/release branch directly, they still get the correct version.

docs/_config.yml
calvinjia: I don't have a good solution. The problem here is `master` is not a valid string for the places you want to use it. It is quite painful to maintain a bunch of variables in `_config.yml` though.

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
madanadit: i see
- our docs will point to using the specs from the docker image. i imagine the common flow to not use the github specs, so i was trying to save the manual effort of updating the release branch
- the best of both would be an automated script to generate the tag/release branch but i'm not aware of any?

let me know what you think @calvinjia

docs/_config.yml
madanadit: yep, that's the trouble I had. but since users often copy paste, i would like to reduce the manual steps of replacing strings from the doc

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
calvinjia: There already is kind of a checklist for updating strings when making a new tag, so I think we could just add this to it.

madanadit: sgtm

madanadit: reverted

docs/_config.yml
madanadit: @calvinjia reverting this here and opening a separate PR to ease cherry-pick

madanadit:
Tests in error: 

madanadit:
jenkins, test this please.

madanadit:

madanadit:

PR #9263, opened 2019-06-11
[DOCFIX] Update metrics docs

wyykwan:


wyykwan:
@ZacBlanco PTAL, thanks!

docs/en/operation/Metrics-System.md
```diff
@@ -71,6 +76,9 @@ If Alluxio is deployed in a cluster, this file needs to be distributed to all th
 Then, start Alluxio, CSV files containing metrics will be found in the `sink.csv.directory`. The
 file name will correspond with the metric name.
 
+To configure another sink, simply update `sink.csv.class` to sink of choice. Sink specific parameters can be found
```
ZacBlanco: I actually think that the property name changes depending on which sink is used. i.e. the JMX sink. https://github.com/Alluxio/alluxio/blob/master/conf/metrics.properties.template#L70

I think there are some instructions earlier which say a user can use the examples in the `metrics.properties.template`. I think that's probably enough. Maybe just another reminder sentence here to tell people where to look for more configurations.

wyykwan: got it, updated!

docs/en/operation/Metrics-System.md
```diff
@@ -23,13 +23,13 @@ following instances are currently supported:
 * Master: The Alluxio master process.
 * Worker: The Alluxio worker process.
 
-Each instance can report to zero or more sinks.
+Each instance can report to zero or more sinks, found in `$ALLUXIO_HOME/core/common/src/main/java/alluxio/metrics/sink/`.
```
ZacBlanco: Because this is user documentation, we don't expect users to have the anything under `./**/src**/` because they will be working with a distribution tarball. I would either point them to a direct github link like: https://github.com/Alluxio/alluxio/tree/master/core/common/src/main/java/alluxio/metrics/sink

docs/en/operation/Metrics-System.md
```diff
@@ -71,6 +76,9 @@ If Alluxio is deployed in a cluster, this file needs to be distributed to all th
 Then, start Alluxio, CSV files containing metrics will be found in the `sink.csv.directory`. The
 file name will correspond with the metric name.
 
+All possible sink specific configurations can be found within their respective class files: 
+`$ALLUXIO_HOME/core/common/src/main/java/alluxio/metrics/sink/`.
```
ZacBlanco: Same comment as above about users having access to this directory.

Also, it would probably be pretty hard to users to figure out the configuration just by looking at the source file (it's unfortunately not very straightforward). I would probably just tell them to look at the `metrics.properties.template`

wyykwan: ah i see, thanks for clarifying! updated again

wyykwan:
jenkins, test this please

wyykwan:
jenkins, test this please

wyykwan:
jenkins, test this please

madanadit:

madanadit:

PR #9264, opened 2019-06-12
Add docs for instanced conf

calvinjia:
Adds docs about modifying configuration programmatically.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
@@ -53,8 +53,6 @@ public class InstancedConfiguration implements AlluxioConfiguration {
   private final boolean mClusterDefaultsLoaded;
 
   /**
-   * WARNING: This API is unstable and may be changed in a future minor release.
-   *
    * Users should use this API to obtain a configuration for modification before passing to a
    * FileSystem constructor. The default configuration contains all default configuration params
    * and configuration properties modified in the alluxio-site.properties file.
@@ -65,6 +63,8 @@ public class InstancedConfiguration implements AlluxioConfiguration {
    * conf.set(...);
    * FileSystem fs = FileSystem.Factory.create(conf);
    *
+   * WARNING: This API is unstable and may be changed in a future minor release.
+   *
    * @return an instanced configuration preset with defaults
    */
   public static InstancedConfiguration defaults() {
@@ -72,6 +72,8 @@ public class InstancedConfiguration implements AlluxioConfiguration {
   }
 
   /**
+   * Creates a new instance of {@link InstancedConfiguration}.
+   *
    * WARNING: This API is not intended to be used outside of internal Alluxio code and may be
    * changed or removed in a future minor release.
    *
@@ -85,6 +87,8 @@ public class InstancedConfiguration implements AlluxioConfiguration {
   }
 
   /**
+   * Creates a new instance of {@link InstancedConfiguration}.
+   *
    * WARNING: This API is not intended to be used outside of internal Alluxio code and may be
    * changed or removed in a future minor release.
    *
@@ -99,6 +103,8 @@ public class InstancedConfiguration implements AlluxioConfiguration {
   }
 
   /**
+   * Creates a new instance of {@link InstancedConfiguration}.
+   *
    * WARNING: This API is not intended to be used outside of internal Alluxio code and may be
    * changed or removed in a future minor release.
    *
--- a/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
@@ -14,6 +14,7 @@ package alluxio.conf;
 import alluxio.conf.PropertyKey.Template;
 import alluxio.exception.ExceptionMessage;
 import alluxio.exception.PreconditionMessage;
+import alluxio.util.ConfigurationUtils;
 import alluxio.util.FormatUtils;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


docs/en/api/FS-API.md
```diff
@@ -78,6 +78,36 @@ CreateFilePOptions options = CreateFilePOptions.newBuilder().setBlockSizeBytes(6
...
+FileSystem normalFs = FileSystem.Factory.get();
+AlluxioURI normalPath = new AlluxioURI("/normalFile");
+// Create a file with default properties
+FileOutStream normalOut = normalFs.createFile(normalPath);
+...
+out.close();
+
+// Create a file system with custom configuration
+InstancedConfiguration conf = new InstancedConfiguration(ConfigurationUtils.defaults());
+conf.set(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, "256MB");
```
ZacBlanco: This might not be the best property to use because it can be set on a per-RPC basis anyways. Maybe setting a security parameter to authenticate as a different user instead? Or maybe a write/read block policy?

docs/en/api/FS-API.md
```diff
@@ -78,6 +78,36 @@ CreateFilePOptions options = CreateFilePOptions.newBuilder().setBlockSizeBytes(6
 FileOutStream out = fs.createFile(path, options);
 ```
 
+#### Programmatically Modifying Configuration
+
+Alluxio configuration can be set through alluxio-site.properties, but these properties apply to all
```
ZacBlanco: maybe use backticks around `alluxio-site.properties` to denote it's a file?

docs/en/api/FS-API.md
```diff
@@ -78,6 +78,36 @@ CreateFilePOptions options = CreateFilePOptions.newBuilder().setBlockSizeBytes(6
...
+InstancedConfiguration conf = new InstancedConfiguration(ConfigurationUtils.defaults());
+conf.set(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, "256MB");
+FileSystem customizedFs = FileSystem.Factory.create(conf);
+AlluxioURI normalPath = new AlluxioURI("/customizedFile");
+// The newly created file will have the custom block size 256MB
+FileOutStream customizedOut = customizedFs.createFile(customizedPath);
+...
+customizedOut.close();
+
+// NormalFs can still be used to create files with the default block size.
```
ZacBlanco: lowercase `normalFs`

docs/en/api/FS-API.md
```diff
@@ -78,6 +78,36 @@ CreateFilePOptions options = CreateFilePOptions.newBuilder().setBlockSizeBytes(6
...
+```java
+FileSystem normalFs = FileSystem.Factory.get();
+AlluxioURI normalPath = new AlluxioURI("/normalFile");
+// Create a file with default properties
+FileOutStream normalOut = normalFs.createFile(normalPath);
+...
+out.close();
+
+// Create a file system with custom configuration
+InstancedConfiguration conf = new InstancedConfiguration(ConfigurationUtils.defaults());
```
apc999: - It is really uncommon to see `XXXXUtils.defaults`. Why not to provide a factory method `InstancedConfiguration.defaults()` directly? --- I feel really strong about it from a developer's perspective.
- Since original `Configuration` interface is no more used, can we rename `InstancedConfiguration` to `Configuration` to make it easy to write?

ZacBlanco: These require code changes which would take a fair amount of effort to overhaul. For being very close to the release this is probably not the best time to go about changing the API design.

That being said. I agree that `InstancedConfiguration.defaults()` would be a good to have, but given the number of people that directly use the java API being so low I don't feel strongly either way. One additional note is that in order to synchronously persist a file using the Java API, a user has to user `FileSystemUtils` too. So by that argument shouldn't we also incorporate those utils into the client somehow?

For both points it would require some amount of change to the way we instantiate, populate, and expose the configuration parameters that get read from `alluxio-site.properties`. Based on some initial IntelliJ data there are at least a few hundred call sites which would need to change or be updated.

I think any change at this point to public-facing APIs should require some more time and careful consideration around naming, and which public-facing methods are available rather than being pushed quickly right before release. I think both concerns here are valid but I feel the amount of work to refactor this is larger compared to the somewhat minor usability benefits.

apc999: Once the API is out, it will be sticky. Meanwhile I could think of some solution in the middle that we could still provide a better public API without really changing the implementation substantially. Let's talk offline

docs/en/api/FS-API.md
calvinjia: Updated, thanks.

docs/en/api/FS-API.md
calvinjia: Updated, thanks.

docs/en/api/FS-API.md
calvinjia: Updated, thanks.

docs/en/api/FS-API.md
calvinjia: Changed to use a slightly nicer API and added warnings to give us some flexibility in the next releases.

core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
```diff
@@ -52,6 +53,30 @@
...
+   * FileSystem fs = FileSystem.Factory.create(conf);
+   *
+   * @return an instanced configuration preset with defaults
+   */
+  public static InstancedConfiguration defaults() {
+    return new InstancedConfiguration(ConfigurationUtils.defaults());
+  }
+
+  /**
+   * WARNING: This API is not intended to be used outside of internal Alluxio code and may be
```
ZacBlanco: How would you feel about adding to the `alluxio.annotation` package? We use `@Deprecated` and `@PublicApi` quite a bit. Maybe this could warrant a new `@Unstable`? Or I think guava also has some similar annotations too.

I don't feel too strongly, but I think using annotations for things like this is pretty standard.

ZacBlanco: In the javadoc typically the first sentence is taken as the preview/short description of what a method does or constructor does. In this case it will always be shown as `WARNING: This API is not intended to be used outside of internal Alluxio code and may be changed or removed in a future minor release.` I would prefer the first sentence for these to say something like `Creates a new instance of {@link InstancedConfiguration}`

calvinjia: I think its a good idea but wouldn't really want to introduce it now since we would need to re-evaluate the codebase and apply it everywhere

calvinjia: Isn't that pretty good?

ZacBlanco: If you feel it's too much effort at this point then I am okay without it. However annotating the API may give developers' IDEs information about using unstable or deprecated APIs. I know that in the guava library when things are marked as experimental or unstable, I get a warning in my IDE about using it.

apc999:
@ZacBlanco  once you LGTMs this PR, I will merge it and cherry-pick it to 2.0

core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
calvinjia: If we mark some apis as unstable, I think it will make all the other ones implicitly stable.

ZacBlanco: Isn't that standard though?

calvinjia:

calvinjia:

PR #9269, opened 2019-06-12
Update docker image tags in k8s spec

madanadit:


calvinjia:
I think master should always just use latest?

madanadit:
@calvinjia since we publish SNAPSHOT docker images, it makes sense to update the tag as well. other someone following along the master docs will end up installing 1.8.1 (or released version) with master docs

calvinjia:
Oh it looks like latest is not really suggested use in Dockerhub.

madanadit:

PR #9283, opened 2019-06-13
Change journal backup default

ns1123:
Change the default backup directory to be '/opt/alluxio/alluxio_backups' on the current Alluxio Primary Master
If a path is provided via CLI, then fall back to using UFS (same behaviour)
If a path is provided via CLI with the --local option, then backup to the specified directory on the primary master (same behaviour)

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
@@ -268,7 +268,7 @@ public final class DefaultMetaMaster extends CoreMaster implements MetaMaster {
     String dir = options.hasTargetDirectory() ? options.getTargetDirectory()
         : ServerConfiguration.get(PropertyKey.MASTER_BACKUP_DIRECTORY);
     UnderFileSystem ufs = mUfs;
-    if (options.getLocalFileSystem() && !ufs.getUnderFSType().equals("local")) {
+    if ((options.getLocalFileSystem() && !ufs.getUnderFSType().equals("local")) || !options.hasTargetDirectory()) {
       ufs = UnderFileSystem.Factory.create("/",
           UnderFileSystemConfiguration.defaults(ServerConfiguration.global()));
       LOG.info("Backing up to local filesystem in directory {}", dir);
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -268,7 +268,7 @@ public BackupResponse backup(BackupPOptions options) throws IOException {
     String dir = options.hasTargetDirectory() ? options.getTargetDirectory()
         : ServerConfiguration.get(PropertyKey.MASTER_BACKUP_DIRECTORY);
     UnderFileSystem ufs = mUfs;
-    if (options.getLocalFileSystem() && !ufs.getUnderFSType().equals("local")) {
+    if ((options.getLocalFileSystem() && !ufs.getUnderFSType().equals("local")) || !options.hasTargetDirectory()) {
```
calvinjia: Can the user specify a non-local destination for periodic backups with this condition?

ns1123: they can't specify a new UFS type. If the user provides a path, it will backup to that path on the root UFS (consistent with current behavior). Allowing a destination aside from local and root UFS would require more complexity (ex: providing an S3 path when Alluxio doesn't have the Access/Secret Key)

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -268,7 +268,8 @@ public BackupResponse backup(BackupPOptions options) throws IOException {
     String dir = options.hasTargetDirectory() ? options.getTargetDirectory()
         : ServerConfiguration.get(PropertyKey.MASTER_BACKUP_DIRECTORY);
     UnderFileSystem ufs = mUfs;
-    if (options.getLocalFileSystem() && !ufs.getUnderFSType().equals("local")) {
+    if ((options.getLocalFileSystem() && !ufs.getUnderFSType().equals("local"))
+            || !options.hasTargetDirectory()) {
```
calvinjia: I think this condition should be
`((options.getLocalFileSystem() || !options.hasTargetDirectory) && !ufs.getUnderFSType().equals("local"))`

ns1123: done.

ns1123:
@calvinjia please summon the bot

calvinjia:

calvinjia:

PR #9296, opened 2019-06-14
Review configuration properties

apc999:
- Review all properties in `PropertyKey.java` and remove a few keys no longer used in the code base
- Modified "alluxio.foo.bar.ms" to "alluxio.foo.bar" in raw keys --- the Property Key names are not changed.
- Renamed a bunch keys to be consistent with Alluxio convention and like to see reaction from the other maintainers.


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3710,8 +3686,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String UNDERFS_KODO_CONNECT_TIMEOUT =
         "alluxio.underfs.kodo.connect.timeout";
     public static final String UNDERFS_KODO_REQUESTS_MAX = "alluxio.underfs.kodo.requests.max";
-    public static final String UNDERFS_RUN_WITH_MANAGEDBLOCKING =
-        "alluxio.underfs.run.with.managedblocking";
+    public static final String UNDERFS_MANAGED_BLOCKING_ENABLED =
+        "alluxio.underfs.managed.blocking.enabled";
```
calvinjia: Is this really an underfs property?

calvinjia: Also we use `underfs` and `ufs` in the property keys, should we consolidate?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4095,10 +4064,6 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.user.file.metadata.sync.interval";
     public static final String USER_FILE_PASSIVE_CACHE_ENABLED =
         "alluxio.user.file.passive.cache.enabled";
-    public static final String USER_FILE_LOAD_TTL =
-        "alluxio.user.file.load.ttl";
-    public static final String USER_FILE_LOAD_TTL_ACTION =
-        "alluxio.user.file.load.ttl.action";
```
calvinjia: Are these no longer used?

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -34,63 +34,68 @@
...
-      put(Name.MASTER_ACTIVE_UFS_SYNC_BATCH_INTERVAL, V2_REMOVED);
+      put(Name.UNDERFS_S3A_CONSISTENCY_TIMEOUT_MS, V2_0_0_REMOVED);
+      put(Name.SWIFT_USE_PUBLIC_URI_KEY, V2_0_0_REMOVED);
+      put(Name.MASTER_CLIENT_SOCKET_CLEANUP_INTERVAL, V2_0_0_REMOVED);
+      put(Name.MASTER_CONNECTION_TIMEOUT, V2_0_0_REMOVED);
+      put(Name.MASTER_HEARTBEAT_INTERVAL, V2_0_0_REMOVED + " Use "
+          + PropertyKey.Name.MASTER_WORKER_HEARTBEAT_INTERVAL + " instead.");
       put(Name.MASTER_JOURNAL_FORMATTER_CLASS, "v2.0 removed the ability to specify the master "
           + "journal formatter");
-      put(Name.LOGSERVER_LOGS_DIR, V2_REMOVED);
```
apc999: these two LOGSERVER related are actually still used. @calvinjia

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -34,63 +34,68 @@
...
+      put(Name.WORKER_TIERED_STORE_RETRY, V2_0_0_REMOVED);
 
       put(Name.TEST_REMOVED_KEY, "This key is used only for testing. It is always removed");
     }
   };
 
   static final class Name {
-    public static final String LOGSERVER_HOSTNAME = "alluxio.logserver.hostname";
-    public static final String LOGSERVER_LOGS_DIR = "alluxio.logserver.logs.dir";
-    public static final String MASTER_ACTIVE_UFS_SYNC_BATCH_INTERVAL =
```
apc999: this and the following one were not in 1.8 but only introduced in 2.0-preview, IMO we don't need to annotate these but only the ones deleted from 1.8

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1123,7 +1118,6 @@ public String toString() {
       new Builder(Name.MASTER_AUDIT_LOGGING_ENABLED)
           .setDefaultValue(false)
           .setDescription("Set to true to enable file system master audit.")
-          .setIgnoredSiteProperty(true)
```
apc999: @calvinjia do you know why this was marked as `IgonredSiteProperty`? This property is not really used any where else (like log4j or sh script) but only in Java land. I tried remove this line and audit log still works

core/common/src/main/java/alluxio/conf/RemovedKey.java
apc999: plus, the logserver configuration is not fully working. do you want to fix it in 2.0?

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: no longer

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1895,50 +1911,28 @@ public String toString() {
...
       new Builder(Name.WORKER_BLOCK_HEARTBEAT_TIMEOUT_MS)
-          .setAlias(new String[]{"alluxio.worker.block.heartbeat.timeout.ms"})
+          .setAlias("alluxio.worker.block.heartbeat.timeout.ms")
           .setDefaultValue(String.format("${%s}", Name.WORKER_MASTER_CONNECT_RETRY_TIMEOUT))
           .setDescription("The timeout value of block workers' heartbeats. If the worker can't "
               + "connect to master before this interval expires, the worker will exit.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
-  public static final PropertyKey WORKER_BLOCK_THREADS_MAX =
```
apc999: these two properties are only used in tests after 2.0. removed

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3878,22 +3884,24 @@ private static String javadocLink(String fullyQualifiedClassname) {
...
+    public static final String MASTER_UFS_ACTIVE_SYNC_EVENT_RATE_INTERVAL =
+        "alluxio.master.ufs.active.sync.event.rate.interval";
+    public static final String MASTER_UFS_ACTIVE_SYNC_MAX_AGE =
+        "alluxio.master.ufs.active.sync.max.age";
+    public static final String MASTER_UFS_ACTIVE_SYNC_INITIAL_SYNC_ENABLED =
+        "alluxio.master.ufs.active.sync.initial.sync.enabled";
     public static final String MASTER_UFS_BLOCK_LOCATION_CACHE_CAPACITY =
         "alluxio.master.ufs.block.location.cache.capacity";
+    public static final String MASTER_UFS_MANAGED_BLOCKING_ENABLED =
+        "alluxio.master.ufs.managed.blocking.enabled";
```
apc999: @calvinjia : discussed with @ggezer . this is the new property name we agreed

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1812,45 +1818,55 @@ public String toString() {
...
+          .setAlias("alluxio.master.worker.timeout.ms")
           .setDefaultValue("5min")
           .setDescription("Timeout between master and worker indicating a lost worker.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_EXECUTOR_PARALLELISM =
-      new Builder("alluxio.master.executor.parallelism")
+  public static final PropertyKey MASTER_RPC_EXECUTOR_PARALLELISM =
+      new Builder(Name.MASTER_RPC_EXECUTOR_PARALLELISM)
```
apc999: @ggezer let me know if you are good with this change to the following property names. e.g. "alluxio.master.executor.parallelism" -> "alluxio.master.rpc.executor.parallelism" to avoid confusion (there are multiple executors inside master). Also I updated the property description. PTAL if you are good

docs/_data/table/en/common-configuration.yml
```diff
@@ -95,7 +99,7 @@ alluxio.proxy.s3.deletetype:
 alluxio.proxy.s3.multipart.temporary.dir.suffix:
   'Suffix for the directory which holds parts during a multipart upload.'
 alluxio.proxy.s3.writetype:
-  'Write type when creating buckets and objects through S3 API. Valid options are `MUST_CACHE` (write will only go to Alluxio and must be stored in Alluxio), `CACHE_THROUGH` (try to cache, write to UnderFS synchronously), `THROUGH` (no cache, write to UnderFS synchronously).'
+  'Write type when creating buckets and objects through S3 API. Valid options are `MUST_CACHE` (write will only go to Alluxio and must be stored in Alluxio), `CACHE_THROUGH` (try to cache, write to UnderFS synchronously), `ASYNC_THROUGH` (try to cache, write to UnderFS asynchronously), `THROUGH` (no cache, write to UnderFS synchronously).'
```
apc999: @cheng-chang , can you check here to confirm my update is correct?

apc999:
Jenkins, test this please

apc999:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -356,9 +359,41 @@ public String toString() {
           .setIsHidden(true)
           .setScope(Scope.ALL)
           .build();
+  public static final PropertyKey NETWORK_CHANNEL_AUTH_TIMEOUT =
```
apc999: @ggezer Please take a look for the next 4 properties. Actually, do you think channel means connection here? If yes, I would even rename Channel to connection. Also. I would avoid mentioning or at least give more general description in property descriptions as people may not know what GRPC is but they simply want to bump up network connection time out. what do you think?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2514,6 +2502,7 @@ public String toString() {
               + "Valid options are "
               + "`MUST_CACHE` (write will only go to Alluxio and must be stored in Alluxio), "
               + "`CACHE_THROUGH` (try to cache, write to UnderFS synchronously), "
+              + "`ASYNC_THROUGH` (try to cache, write to UnderFS asynchronously), "
```
apc999: @cheng-chang please help to confirm if this is a correct fix

core/server/master/src/main/java/alluxio/master/file/async/AsyncPersistHandler.java
```diff
@@ -53,10 +50,7 @@ private Factory() {} // prevent instantiation
      * @return the generated {@link AsyncPersistHandler}
      */
     public static AsyncPersistHandler create(FileSystemMasterView view) {
-      return CommonUtils.createNewClassInstance(
```
apc999: we shall remove this class in a separate PR

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3754,17 +3757,13 @@ private static String javadocLink(String fullyQualifiedClassname) {
...
         "alluxio.master.cluster.metrics.update.interval";
-    public static final String MASTER_CONNECTION_TIMEOUT_MS =
-        "alluxio.master.connection.timeout";
     public static final String MASTER_DAILY_BACKUP_ENABLED =
         "alluxio.master.daily.backup.enabled";
     public static final String MASTER_DAILY_BACKUP_FILES_RETAINED =
         "alluxio.master.daily.backup.files.retained";
     public static final String MASTER_DAILY_BACKUP_TIME =
         "alluxio.master.daily.backup.time";
-    public static final String MASTER_FILE_ASYNC_PERSIST_HANDLER =
```
apc999: no longer needed as job persist becomes the default and hard-coded too

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: Talked offline. This is supported now

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Does audit logging work if you remove this line and set it through the site properties?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -356,9 +359,43 @@ public String toString() {
           .setIsHidden(true)
           .setScope(Scope.ALL)
           .build();
+  public static final PropertyKey NETWORK_CONNECTION_AUTH_TIMEOUT =
```
apc999: @ggezer  are you good with the following?

ggezer: These LGTM. Thanks!

apc999:

apc999:

Issue #9022, opened 2019-05-08
alluxio sometimes load the local / dir


liuzx32:
**Alluxio Version:**
master

**Describe the bug**
The cmd "alluxio ls -f" will sometimes load the local / dir, just like the picture.

<img src="https://user-images.githubusercontent.com/19620810/57345297-e39b4d00-717c-11e9-86b7-a412f418f86b.png" width="100" align=center>

liuzx32:
/bin /boot /dev /etc /mnt /sys ... are the local machine dirs, not the cloud bucket dir.

ZacBlanco:
Thanks for reporting this! We'll need some more information to be able to debug this. Do you have a configuration that reliably reproduces this problem? and could you describe your environment?

If we can reliably reproduce the issue, then we should be able to fix it.

liuzx32:
@ZacBlanco Thanks for the answer.
#
The case is not necessarily recurring, and It only happens on alluxio-2.0.0-SNAPSHOT.
After alluxio runs for a while, produces files on s3 or oss, then alluxio ls -f to update the alluxio meta.

liuzx32:
Refer: https://github.com/Alluxio/alluxio/pull/9308

rastogiasr:
If this issue is still valid, please re-open it.

PR #9308, opened 2019-06-15
AbstractFileSystem#listStatus returns a list of LocatedFileStatus instances

witgo:
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/InMemoryFileIndex.scala#L315

### DIFF
--- a/core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
+++ b/core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
@@ -12,9 +12,11 @@
 package alluxio.hadoop;
 
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
 
 import alluxio.AlluxioURI;
 import alluxio.ClientContext;
+import alluxio.client.block.BlockWorkerInfo;
 import alluxio.client.file.FileOutStream;
 import alluxio.client.file.FileSystem;
 import alluxio.client.file.URIStatus;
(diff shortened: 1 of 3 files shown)


liuzx32:
@witgo Hello, Is the same question with the  issue https://github.com/Alluxio/alluxio/issues/9022 ?

witgo:
@liuzx32  They're not the same bug.

apc999:
@witgo  thanks for the patch. Can you elaborate here what's the issue we are fixing here? it is not clear to me by reading the link in this issue description

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -52,6 +54,7 @@
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.LocatedFileStatus;
```
apc999: which hadoop version introduces `org.apache.hadoop.fs.LocatedFileStatus`?
If it is a recent version, by having this line of src code, we might not be able to compile Alluxio client with earlier hadoop version. In this case, we may need to introduce multiple implementations of `AbstractFileSystem.java` for different Hadoop versions.

witgo: `hadoop 2.0 +` version introduces `org.apache.hadoop.fs.LocatedFileStatus`?

apc999: It looks like this is introduced in 2.x. So without your change @witgo , does this prevent you using any data source for Spark? If that is the case, please let us know and we will see the importance and address it accordingly.

PR #9315, opened 2019-06-18
Fix job-master gRPC context cascading

ggezer:
gRPC requires `io.grpc.Context` instance to be forked if an RPC implementation calls another RPC. Because termination of first call will invalidate context of the cascaded calls. Even if calls are contained within each other, with authentication long polling, authentication stream opened for the second call will be canceled after first call returns. That means every call to first RPC will cause tearing up of channel used for the second RPC. 

This PR forks gRPC context inside `JobMaster:run RPC call` since some job definitions require RPC calls during initialization.

### DIFF
--- a/job/server/src/main/java/alluxio/master/job/JobMaster.java
+++ b/job/server/src/main/java/alluxio/master/job/JobMaster.java
@@ -49,6 +49,7 @@ import alluxio.wire.WorkerNetAddress;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import io.grpc.Context;
 import net.jcip.annotations.GuardedBy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

ggezer:
@bf8086 PTAL. Thanks!

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -205,46 +206,54 @@ public String getName() {
...
-        if (completedBeforeMs < RETENTION_MS) {
-          // mFinishedJobs is sorted. Can't iterate to a job within retention policy
-          break;
+    // This RPC service implementation triggers another RPC.
+    // Run the implementation under forked context to avoid interference.
+    // Then restore the current context at the end.
+    Context forkedCtx = Context.current().fork();
+    Context prevCtx = forkedCtx.attach();
+    try {
+      if (mIdToJobCoordinator.size() == mCapacity) {
```
ggezer: Change looks big but I just wrapped existing method with try-finally block and do context forking/attaching/detaching.

bf8086: thanks, turning off space diff helps.

ggezer:
@calvinjia PTAL.

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -205,46 +206,54 @@ public String getName() {
...
+        if (mFinishedJobs.isEmpty()) {
+          // The job master is at full capacity and no job has finished.
+          throw new ResourceExhaustedException(
+              ExceptionMessage.JOB_MASTER_FULL_CAPACITY.getMessage(mCapacity));
+        }
+        // Discard old jobs that have completion time beyond retention policy
+        Iterator<JobInfo> jobIterator = mFinishedJobs.iterator();
+        // Used to denote whether space could be reserved for the new job
+        // It's 'true' if job master is at full capacity
+        boolean isfull = true;
```
calvinjia: `isFull`

calvinjia:

calvinjia:

PR #9325, opened 2019-06-18
[DOCFIX] Update vagrant docs, vb and vagrant settings

wyykwan:


deploy/vagrant/conf/alluxio.yml
```diff
@@ -2,15 +2,15 @@
...
 
 # github repo and version(can be branch, commit hash)
 Github:
   Repo: https://github.com/alluxio/alluxio
-  Version: branch-1.6
+  Version: branch-2.0
 
 Release:
-  Version: 1.6.1 # name of release version
+  Version: 2.0.0-RC3 # name of release version
```
cheng-chang: @wyykwan use 2.0.0?

deploy/vagrant/conf/alluxio.yml
```diff
@@ -2,15 +2,15 @@
...
 
 # github repo and version(can be branch, commit hash)
 Github:
   Repo: https://github.com/alluxio/alluxio
-  Version: branch-1.6
+  Version: branch-2.0
 
 Release:
-  Version: 1.6.1 # name of release version
+  Version: 2.0.0 # name of release version
```
calvinjia: What is this version string used for? Does this need to be equivalent to the version of Alluxio (ie. 2.0.0 and not 2.0.0-RCx)? If so we should use a github tag instead of a branch.

cheng-chang: it's a the version matching the downloads, for example, 1.8.1 matches tarballs under https://downloads.alluxio.io/downloads/files/1.8.1/

PR #9335, opened 2019-06-19
Reduce the information allocator need in createBlockMeta

LuQQiu:
In each `createBlockMetaInternal`, Space allocator will get a updated `BlockMetadataManagerView` which will do a full copy of pinList. Copy the pinList is slow when the list is huge (especially in async through) and copy is down in a `synchronized lock` which reduce the concurrency.  And actually, the allocator in the `createBlockMetaInternal` do not need the pin list info. It only needs the storage tier and directory information. 

This PR is to reduce the information given to allocator and thus reduce the contention and improve performance.


core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -586,15 +585,15 @@ private TempBlockMeta createBlockMetaInternal(long sessionId, long blockId,
       if (newBlock) {
         checkTempBlockIdAvailable(blockId);
       }
-      StorageDirView dirView =
-          mAllocator.allocateBlockWithView(sessionId, initialBlockSize, location, getUpdatedView());
-      if (dirView == null) {
+      StorageDir dir = mAllocator.allocateBlockWithTierInfo(sessionId, initialBlockSize, location,
+          mMetaManager.getTiers(), mMetaManager.getAliasToTiersMap());
```
LuQQiu: Do we also want to create another `BlockMetadataManagerView` to provide read only access to the tiers and directories?

madanadit: +1. i like that better than exposing the tiers and alias to tiers maps

LuQQiu:
@madanadit I added the `StorageMetadataView`, `StorageTierView`, and `StorageDirView` to only contains the information needed for allocator.

I renamed the original `StorageTierView` and `StorageDirView` to `StorageTierEvictableView` and `StorageDirEvictableView` to distinguish from the new ones.

In Allocator, there are many similar codes. I think of modifying the `StorageTierEvictableView` to extend `StorageTierView`, `StorageDirEvictableView` to extend `StorageDirView` to reduce the duplicate code, but then there will be many cast in our evictor. The evictor logics are much more complicated then allocator. I want to limit the impact of my changes as small as possible. 

PTAL, thanks!

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -586,8 +587,8 @@ private TempBlockMeta createBlockMetaInternal(long sessionId, long blockId,
       if (newBlock) {
         checkTempBlockIdAvailable(blockId);
       }
-      StorageDirView dirView =
-          mAllocator.allocateBlockWithView(sessionId, initialBlockSize, location, getUpdatedView());
+      StorageDirView dirView = mAllocator.allocateBlockWithView(sessionId,
+          initialBlockSize, location, new StorageMetadataView(mMetaManager));
```
madanadit: nit; did you consider calling `StorageMetadatView` `BlockMetadataManagerView` instead and renaming the existing `BlockMetadataManagerView` to `BlockMetadataManagerEvictableView` for consistency?

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
```diff
@@ -58,9 +76,9 @@ public static Allocator create(BlockMetadataManagerView view) {
...
    * @param location the location in block store
    * @param view of the metadata manager
-   * @return a {@link StorageDirView} in which to create the temp block meta if success, null
-   *         otherwise
+   * @return a {@link StorageDirEvictableView} in which to create the temp block meta if success,
+   *         null otherwise
    */
-  StorageDirView allocateBlockWithView(long sessionId, long blockSize, BlockStoreLocation location,
-      BlockMetadataManagerView view);
+  StorageDirEvictableView allocateBlockWithEvictableView(long sessionId, long blockSize,
```
madanadit: hmm, does allocator need evict-able view?

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTierView.java
```diff
@@ -25,29 +23,36 @@
...
+    this(tier, false);
+  }
+
+  /**
+   * Creates a {@link StorageTierView} using the actual {@link StorageTier}.
+   *
+   * @param tier which the tierView is constructed from
+   * @param evictable whether this class contains evict-related methods
+   */
+  public StorageTierView(StorageTier tier, boolean evictable) {
```
LuQQiu: @madanadit Are there better ways than using a boolean?
Here i use `evictable` is because the `StorageDirView` and `StorageDirEvictableView` have different parameters.

madanadit:
Thanks a lot @LuQQiu! Looks good after the comment / variable name fixes. @ggezer PTAL

core/server/worker/src/main/java/alluxio/worker/block/evictor/LRFUEvictor.java
```diff
@@ -78,9 +79,9 @@ public LRFUEvictor(BlockMetadataManagerView view, Allocator allocator) {
         "Attenuation factor should be no less than 2.0");
 
     // Preloading blocks
-    for (StorageTierView tier : mManagerView.getTierViews()) {
+    for (StorageTierView tier : mMetadataView.getTierViews()) {
       for (StorageDirView dir : tier.getDirViews()) {
-        for (BlockMeta block : dir.getEvictableBlocks()) {
+        for (BlockMeta block : ((StorageDirEvictorView) dir).getEvictableBlocks()) {
```
ggezer: `StorageDirView` is casted to `StorageDirEvictorView` at many places. Can we avoid that somehow? Maybe `StorageTierEvictorView` could implement a one that returns `StorageDirEvictorView` instead.

LuQQiu: I tried to use generic view of `StorageTierView` and `StorageDirView` and implement the abstract methods in allocator and evictor classes. The detail code listed in https://github.com/LuQQiu/alluxio/pull/1. This way reduce many unneeded cast.

However, in the allocators, we directly operate on abstract class `BlockMetadataView`.
And intellij will complain:
```
for (StorageTierView tierView : mMetadataView.getTierViews()) {
        for (StorageDirView dirView : tierView.getDirViews()) { <-- complain about incompatible types
          if (dirView.getMediumType().equals(mediumType)
```
Incompatible types: required object, but found StorageDirView

LuQQiu:
Mistakenly push one more commit, force push back to the commit that has been reviewed and approved

calvinjia:

calvinjia:

PR #9343, opened 2019-06-20
Add root cause to our logging

yuzhu:


yuzhu:
@calvinjia This change gives us more information to work with in case of a deeply nested exception. PTAL

core/server/common/src/main/java/alluxio/ProcessUtils.java
```diff
@@ -37,11 +37,13 @@ public static void run(Process process) {
       System.exit(0);
     } catch (Throwable t) {
       LOG.error("Uncaught exception while running {}, stopping it and exiting.", process, t);
+      LOG.error("Root cause of the exception: ", ExceptionUtils.getRootCause(t));
```
calvinjia: Can we include this in the previous log message?

yuzhu: done

yuzhu:
jenkins, test this please

calvinjia:

calvinjia:

apc999:
@calvinjia @yuzhu this pr introduced a side-effect that the stack trace is lost from the log. Note that the original logging will output the entire exception including stack trace of `t`:
```
LOG.error("Uncaught exception while running {}, stopping it and exiting.", process, t)
```
After this PR, one will see sth like, to me it actually becomes harder to debug.
```
‘2019-07-03 14:58:01,065 ERROR ProcessUtils - Uncaught exception while stopping Alluxio worker, simply exiting. Exception java.lang.NullPointerException, Root Cause null‘ 
```
I would put back `t` in the end as the last arg

PR #9353, opened 2019-06-23
Update Removed Keys

apc999:
- Updated the delete keys by running diff between getConf in 1.8 and 2.0-RC4
- Update the reference to these deleted keys in doc and src code
- Simplify the `RemovedKey` implementation a bit to make my life a bit easier


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1615,6 +1615,7 @@ public String toString() {
       .build();
   public static final PropertyKey MASTER_RPC_PORT =
       new Builder(Name.MASTER_RPC_PORT)
+          .setAlias("alluxio.master.rpc.port")
```
apc999: I found an inconsistency that we have key `alluxio.job.master.rpc.port` but for master it is `alluxio.master.port`. 
@calvinjia let me know what you think about this change (I'm not strong on this)
maybe alternatively, we can use ``alluxio.master.port` and `alluxio.job.master.port`

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2256,6 +2257,7 @@ public String toString() {
       .build();
   public static final PropertyKey WORKER_RPC_PORT =
       new Builder(Name.WORKER_RPC_PORT)
+          .setAlias("alluxio.worker.rpc.port")
```
apc999: same here

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2588,6 +2590,7 @@ public String toString() {
           .build();
   public static final PropertyKey LOCALITY_TIER_NODE =
       new Builder(Template.LOCALITY_TIER, Constants.LOCALITY_NODE)
+          .setAlias("alluxio.user.hostname")
```
apc999: In some previous PR, I found `alluxio.user.hostname` was totally removed and replaced by `alluxio.locality.node`. I actually don't think it is correct. IMO,  `alluxio.locality.node` is not specific to `alluxio.user.hostname` but also applies to `alluxio.{master,worker}.hostname`, right? I think the current implementation is weird by treating `alluxio.{master,worker}.hostname` in parallel with `alluxio.locality.node`

@calvinjia  please let me know what you think. I think we ought to fix this

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2928,7 +2932,7 @@ public String toString() {
           .build();
   public static final PropertyKey USER_METRICS_HEARTBEAT_INTERVAL_MS =
       new Builder(Name.USER_METRICS_HEARTBEAT_INTERVAL_MS)
-          .setAlias("alluxio.user.heartbeat.interval.ms")
+          .setAlias("alluxio.user.metrics.heartbeat.interval.ms")
```
apc999: fix a bug I introduced in the previous PR which sets the wrong alias

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2864,6 +2867,7 @@ public String toString() {
           .setDefaultValue(false).build();
   public static final PropertyKey USER_BLOCK_WRITE_LOCATION_POLICY =
       new Builder(Name.USER_BLOCK_WRITE_LOCATION_POLICY)
+          .setAlias("alluxio.user.block.write.location.policy.class")
```
apc999: I changed this property to be `alluxio.user.block.write.location.policy` without suffixed `class`. if you guys don't like the alias, I could remove it.

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -24,110 +24,144 @@
...
+   * @return a warning message indicating a property has been removed since a given version
+   */
+  private static String replacedSince(String version, String newProperty) {
+    return String.format("this property has been removed since %s, use %s instead.", version,
+        newProperty);
+  }
 
   private static final Map<String, String> REMOVED_KEYS = new HashMap<String, String>(20) {
     {
-      put(Name.WEB_TEMP_PATH, "The alluxio web UI overhaul in v2.0 removed this parameter.");
```
apc999: having another nested `Name` class is not necessary IMO, but adding more work. Without the `Name`s, the code of this class is much shorter.

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -24,110 +24,144 @@
...
+      put("alluxio.worker.network.netty.file.transfer", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.file.writer.threads.max", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.reader.buffer.size.packets", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.rpc.threads.max", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.shutdown.timeout", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.writer.buffer.size.packets", removedSince(V2_0_0));
+      put("alluxio.worker.tieredstore.reserver.enabled", removedSince(V2_0_0));
+      put("alluxio.worker.tieredstore.retry", removedSince(V2_0_0));
+      put("fs.swift.apikey", replacedSince(V2_0_0, PropertyKey.Name.SWIFT_PASSWORD_KEY));
+      put("fs.swift.use.public.url", removedSince(V2_0_0));
```
apc999: @madanadit  I found this property `fs.swift.use.public.url` is referenced quite a bit in doc but never used in code (even in 1.8), can you double check if it is safe to delete this property?

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -191,4 +225,6 @@ public static String getMessage(String key) {
     }
     return null;
   }
+
+  private RemovedKey() {} // prevent instantiation of this class
```
apc999: @ZacBlanco we don't expect anyone to instantiate this class, do we?

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -24,110 +24,144 @@
...
+      put("alluxio.worker.network.netty.block.writer.threads.max", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.buffer.receive", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.buffer.send", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.file.transfer", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.file.writer.threads.max", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.reader.buffer.size.packets", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.rpc.threads.max", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.shutdown.timeout", removedSince(V2_0_0));
+      put("alluxio.worker.network.netty.writer.buffer.size.packets", removedSince(V2_0_0));
+      put("alluxio.worker.tieredstore.reserver.enabled", removedSince(V2_0_0));
```
apc999: @calvinjia , do you know if this (deleting this flag) is correct? does that mean we will always turn on async cache (as opposed to turning it on by default)? Also can you check the doc? do we still have two modes of eviction mixed?

docs/en/compute/Presto.md
```diff
@@ -212,12 +212,6 @@ For Alluxio 1.6 or earlier,
 it is recommended to set this size no less than Alluxio's block size to avoid the read contention within the same block. For later Alluxio versions, this is no more an issue due to
 async cache on Alluxio workers.
 
-### Avoid Presto timeout reading large files
-
-It is recommended to increase `alluxio.user.network.netty.timeout` to a bigger value (e.g.
```
apc999: @bf8086  do you have any recommendation here after  `alluxio.user.network.netty.timeout` is removed in 2.0?

docs/en/operation/Troubleshooting.md
```diff
@@ -191,19 +178,6 @@ the property `alluxio.master.journal.folder` setting. There can be multiple reas
 some HDFS datanodes serving the journal files are under heavy load or running out of disk space. Please ensure the
 HDFS deployment is connected and healthy for Alluxio to store journals when the journal directory is set to be in HDFS.
 
-### Q: I'm seeing that client connection was rejected by master
-
-A: When you see errors from applications like `"alluxio.exception.status.UnavailableException:
-Failed to connect to BlockMasterClient @ hostname:19998 after 13 attempts"`, a possibility is the
-Alluxio master server has run out threads in its thread pool to serve new incoming client requests.
```
apc999: @bf8086 do we possibly still run into this issue? Any suggestion here after `alluxio.master.worker.threads.max` was removed?

docs/en/advanced/Performance-Tuning.md
```diff
@@ -33,11 +33,11 @@ The following is a checklist to run through to address common problems when tuni
    the compute application is likely not interfacing with a local Alluxio worker.
    The Alluxio client uses hostname matching to discover a local Alluxio worker;
    check that the client and worker use the same hostname string.
-   Configuring `alluxio.locality.node` and `alluxio.worker.hostname` sets the client and worker
```
apc999: > Configuring `alluxio.locality.node` and `alluxio.worker.hostname` sets the client and worker
This is very counterintuitive compared to 
> Configuring `alluxio.user.hostname` and `alluxio.worker.hostname` sets the client and worker

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: `alluxio.user.hostname` was removed because it was marked as deprecated.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: Personally, I like the class at the end because it points to a specific implementation. It also gives more of a hint as to what the value of the property should be. The `.class` suffix is used commonly in `hdfs-site`, `mapred-site`, and `yarn-site` as well.

I don't think we should change the name of this property.

docs/_data/table/cn/master-configuration.yml
```diff
@@ -1,12 +1,10 @@
 alluxio.master.bind.host:
   Alluxio master绑定的主机名。参考<a href="#configure-multihomed-networks">多宿主网络</a>
-alluxio.master.heartbeat.interval:
-  Alluxio master心跳间隔。
+alluxio.master.master.heartbeat.interval:
```
ZacBlanco: `master.master`?

docs/en/advanced/Performance-Tuning.md
ZacBlanco: Yes, but then why was `user.hostname` deprecated? I think it's fine because if users are configuring this they're probably more experienced/knowledgable in this realm anyways.

A possible stop-gap here is to make `alluxio.user.hostname` and alias of `alluxio.locality.node`

docs/en/advanced/Performance-Tuning.md
```diff
@@ -52,7 +52,7 @@ The following is a checklist to run through to address common problems when tuni
    Alluxio clients maintain a connection to the master to avoid using a new connection each time.
    Each client will occupy a server thread while an RPC request is pending.
    This may deplete the master's thread pool; its size can be increased by setting
-   `alluxio.master.executor.fork.pool.size.max`, which has a default value of 500.
+   `alluxio.master.rpc.executor.max.pool.siz`, which has a default value of 500.
```
ZacBlanco: `size`

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: My $.02: break backcompat here and add the aliases. I think it is much better to user `*.rpc.port` and `*.web.port`. `*.port` isn't quite descriptive enough

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4091,14 +4095,14 @@ private static String javadocLink(String fullyQualifiedClassname) {
...
+        "alluxio.user.block.write.location.policy";
     public static final String USER_CONF_CLUSTER_DEFAULT_ENABLED =
         "alluxio.user.conf.cluster.default.enabled";
     public static final String USER_CONF_SYNC_INTERVAL = "alluxio.user.conf.sync.interval";
     public static final String USER_DATE_FORMAT_PATTERN = "alluxio.user.date.format.pattern";
     public static final String USER_FILE_BUFFER_BYTES = "alluxio.user.file.buffer.bytes";
-    public static final String USER_FILE_COPY_FROM_LOCAL_BLOCK_LOCATION_POLICY =
-        "alluxio.user.file.copyfromlocal.block.location.policy.class";
+    public static final String USER_FILE_COPYFROMLOCAL_BLOCK_LOCATION_POLICY =
+        "alluxio.user.file.copyfromlocal.block.location.policy";
```
ZacBlanco: This is new in 2.0 right? Can we change this to be `*.copy.from.local.*`? We should follow paradigm of the rest of the property keys and separate words by `.`

ZacBlanco: Also would like to add that I like using `.class` as well as keeping words in the var name separated by `_`

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: right. I guess what puzzled me is why `alluxio.user.hostname` can be deprecated by `alluxio.locality.node` as  `alluxio.locality.node` seems to apply to client, master and worker

docs/en/compute/Presto.md
bf8086: The equivalent property in Alluxio 2.0 should be `alluxio.user.network.data.timeout`.

docs/en/operation/Troubleshooting.md
bf8086: It should be less likely for user to run into this issue in 2.0 because of the new shared thread model, but users can tweak `alluxio.master.rpc.executor.parallelism` if they see delays or timeout on RPC request. @ggezer WDYT?

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: I'm in favor of option 3 as well.

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: I think we tried to consolidate the concepts to only use `alluxio.locality.node` (this is a superset of the functionality provided by `alluxio.user.hostname`). On the client, the functionality is equivalent, and `alluxio.user.hostname` should not be set on the servers. However the property key itself is not very descriptive of what it does so I think the alias is fine.

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Since the value needs to be a class name I think having the suffix is fine.

core/common/src/main/java/alluxio/conf/RemovedKey.java
calvinjia: There is only one mode of eviction now. It looks like there are still references to synchronous eviction in the docs.

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: -> option 3

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: keep `.class`, make a ticket to support passing a short policy name

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: mark as deprecated , create an issue to fix local first.

core/common/src/main/java/alluxio/conf/RemovedKey.java
apc999: only async eviction supported now

docs/_data/table/cn/master-configuration.yml
apc999: `standby.master`

core/common/src/main/java/alluxio/conf/RemovedKey.java
apc999: confirmed, we are not using this property key any more

docs/en/compute/Presto.md
```diff
@@ -212,6 +212,12 @@ For Alluxio 1.6 or earlier,
 it is recommended to set this size no less than Alluxio's block size to avoid the read contention within the same block. For later Alluxio versions, this is no more an issue due to
 async cache on Alluxio workers.
 
+### Avoid Presto timeout reading large files
+
+It is recommended to increase `alluxio.user.network.data.timeout` to a bigger value (e.g
+`10min`) to avoid the timeout
```
ZacBlanco: I would typically write "...avoid **a** timeout failure when reading..."

apc999: done

apc999:
@calvinjia @ZacBlanco  PTAL

docs/_data/table/en/worker-configuration.yml
```diff
@@ -86,10 +86,10 @@ alluxio.worker.network.writer.buffer.size.messages:
   'When a client writes to a remote worker, the maximum number of data messages to buffer by the server for each request.'
 alluxio.worker.network.zerocopy.enabled:
   'Whether zero copy is enabled on worker when processing data streams.'
-alluxio.worker.port:
-  'The port Alluxio''s worker node runs on.'
 alluxio.worker.principal:
   'Kerberos principal for Alluxio worker.'
+alluxio.worker.rpc.port:
+  'The port Alluxio''s worker node runs on.'
```
ZacBlanco: is this supposed to have two `''`?

ZacBlanco: Ah nevermind, `''` is used to escape `'` within single quotes in YAML.

http://docs.octoprint.org/en/master/configuration/yaml.html#string

apc999: fixed

docs/en/operation/Troubleshooting.md
apc999: @bf8086 @ggezer  if you think it is possible, can you contribute a paragraph on this in a separate doc fix pr?

calvinjia:

calvinjia:

LuQQiu:
@apc999 @ZacBlanco 
In the removed keys, I found those properties are not removed and are still be using in code
```
      put("alluxio.user.block.master.client.pool.gc.interval", removedSince(V2_0_0));
      put("alluxio.user.block.master.client.pool.gc.threshold", removedSince(V2_0_0));
      put("alluxio.user.block.master.client.pool.size.max", removedSince(V2_0_0));
      put("alluxio.user.block.master.client.pool.size.min", removedSince(V2_0_0));
      put("alluxio.user.file.master.client.pool.gc.interval", removedSince(V2_0_0));
      put("alluxio.user.file.master.client.pool.gc.threshold", removedSince(V2_0_0));
      put("alluxio.user.file.master.client.pool.size.max", removedSince(V2_0_0));
      put("alluxio.user.file.master.client.pool.size.min", removedSince(V2_0_0));
```
Not sure if other added removedKeys also have the same circumstances

ZacBlanco:
@LuQQiu the changes that added those keys were only introduced a month ago...how did they get into this PR?

PR #9357, opened 2019-06-25
Support fixes for ROCKS meta-store

ggezer:


ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -80,6 +82,14 @@ public RocksBlockStore(String baseDir) {
         .setCreateMissingColumnFamilies(true);
     String dbPath = PathUtils.concatPath(baseDir, BLOCKS_DB_NAME);
     String backupPath = PathUtils.concatPath(baseDir, BLOCKS_DB_NAME + "-backups");
+    // Create block store db path if it does not exist.
+    if (!FileUtils.exists(dbPath)) {
+      try {
+        FileUtils.createDir(dbPath);
+      } catch (IOException ioe) {
```
apc999: nit: unless `e` is taken, just name exception to be `e` in a `catch` clause

ggezer: Hmm. I didn't know about this convention. I might have broken it many times..

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -86,8 +86,8 @@ public RocksBlockStore(String baseDir) {
     if (!FileUtils.exists(dbPath)) {
       try {
         FileUtils.createDir(dbPath);
-      } catch (IOException ioe) {
-        LOG.warn("Failed to create nonexistent db path at: {}. Error:{}", dbPath, ioe);
+      } catch (IOException e) {
+        LOG.warn("Failed to create nonexistent db path at: {}. Error:{}", dbPath, e);
```
calvinjia: do we need `e.toString` here?

ggezer: It's implicit I think, because it's interpreted as `Object`

calvinjia: I was worried about it treating it differently when the last arg is a throwable, but looking at the implementation I think you are right.

calvinjia:

calvinjia:

PR #9359, opened 2019-06-25
Close shell resources properly after command execution

ggezer:
Failure to close resources that has gRPC channels will cause abrupt channel termination, thus annoying `CANCELLED` exception in the master logs.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/FileSystemShell.java
+++ b/shell/src/main/java/alluxio/cli/fs/FileSystemShell.java
@@ -47,11 +47,6 @@ public final class FileSystemShell extends AbstractShell {
       .build();
 
   /**
-   * Shared context for loaded commands.
-   */
-  private FileSystemContext mContext;
-
-  /**
    * Main method, starts a new FileSystemShell.
    *
    * @param argv array of arguments given by the user's input from the terminal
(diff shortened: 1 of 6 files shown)


ggezer:
@LuQQiu PTAL. Thanks!

shell/src/main/java/alluxio/cli/fsadmin/FileSystemAdminShell.java
```diff
@@ -79,15 +83,24 @@ protected String getShellName() {
...
     return CommandUtils.loadCommands(FileSystemAdminShell.class.getPackage().getName(),
-        new Class[] {Context.class, AlluxioConfiguration.class}, new Object[] {context,
+        new Class[] {Context.class, AlluxioConfiguration.class}, new Object[] {mContext,
             mConfiguration});
   }
+
+  @Override
+  public void close() throws IOException {
+    super.close();
+    mContext.getBlockClient().close();
```
apc999: why not make `Context` closable and using `Closer` inside `Context.close()` to make sure we try the best to close all resources?

shell/src/main/java/alluxio/cli/fs/FileSystemShell.java
```diff
@@ -84,6 +89,13 @@ protected String getShellName() {
...
   @Override
   protected Map<String, Command> loadCommands() {
-    return FileSystemShellUtils.loadCommands(FileSystemContext.create(mConfiguration));
+    mContext = FileSystemContext.create(mConfiguration);
+    return FileSystemShellUtils.loadCommands(mContext);
+  }
+
+  @Override
+  public void close() throws IOException {
+    super.close();
```
apc999: an simpler way can be introduce a new member `Closer` in `AbstractShell`, and add resource (e.g., the `mContext` here to the closer from `AbstractShell`. Then you can directly close this `Closer` in `AbstractShell.close()` rather than having the same code on close repeatedly across different impl of `AbstractShell`

ggezer: Thanks Bin. Closer made it much simpler and cleaner.

calvinjia:

calvinjia:

PR #9366, opened 2019-06-27
Reset the counter in MetricsSystemTest

calvinjia:
The counter needs to be reset to 0 if we don't want tests stepping on each other.


apc999:
is this the only way to reset to zero? it is  implicit

calvinjia:
@apc999 There is no `set` API for counters if that's what you are asking. Another way of doing this is to make the tests not rely on absolute values (ie. instead of checking for 1, it checks for counter value + 1), but that is more burden on each test which I would rather put in a single place (`Before` in this case).
Yeah we can cherry-pick this to 2.0 if you are okay with the PR.

apc999:
@calvinjia SGTM as the fix. 
.

apc999:

apc999:

PR #9371, opened 2019-06-27
Extract SaslClient/SaslServer logic from handlers as abstract classes

ggezer:


ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

ggezer:
@bf8086 PTAL. Thanks!

ggezer:
@bf8086 PTAL.

core/common/src/main/java/alluxio/security/authentication/AbstractSaslServerHandler.java
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.security.sasl.SaslException;
+import javax.security.sasl.SaslServer;
+
+/**
+ * Abstract {@link SaslServerHandler} implementation that maintains {@link SaslServer} instance.
+ */
+public abstract class AbstractSaslServerHandler implements SaslServerHandler {
```
bf8086: These classes looks identical in logic. Is there a way to further unify the code path?

ggezer: You mean dropping `SaslServerHandler` maybe?

bf8086: not sure about that, I was just thinking of having something like a `AbstractSaslHandler<SaslServer>` so we don't have to repeat the logic in two classes.

ggezer: `SaslClient` and `SaslServer` are not sharing an ancestor even though their logic looks identical. That makes it hard to unify their abstract handlers without lots of boilerplate...

ggezer:
@gpang PTAL. Thanks!

gpang:

Issue #9355, opened 2019-06-24
Concurrent monitoring of Alluxio processes during startup


ggezer:
**Is your feature request related to a problem? Please describe.**
For a big cluster, it takes quite a lot of time for alluxio-start.sh to be done with monitoring worker processes.

**Describe the solution you'd like**
For the same process types, monitors could be spawned concurrently

PR #9374, opened 2019-06-27
Do parallelize node monitoring during startup

ggezer:
Fix https://github.com/Alluxio/alluxio/issues/9355

### DIFF
--- a/bin/alluxio-monitor.sh
+++ b/bin/alluxio-monitor.sh
@@ -176,7 +176,7 @@ run_monitors() {
     if [[ $? -ne 0 ]]; then
       # if there is an error, print the log tail for the remaining master nodes.
       for node in $(echo ${nodes}); do
-        run_on_node ${node} "${BIN}/alluxio-monitor.sh" -L "${node_type}"
+        run_on_node ${node} "${BIN}/alluxio-monitor.sh" -L "${node_type}" &
       done
     else
       HA_ENABLED=$(${BIN}/alluxio getConf ${ALLUXIO_MASTER_JAVA_OPTS} alluxio.zookeeper.enabled)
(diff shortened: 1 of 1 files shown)


ggezer:
@ZacBlanco PTAL. Thanks!

ggezer:
@ZacBlanco PTAL.

bin/alluxio-monitor.sh
```diff
@@ -175,27 +175,44 @@ run_monitors() {
...
+
+# Used to run a command on multiple hosts concurrently.
+# By default it limits concurrent tasks to 100.
+batch_run_on_nodes() {
+  # String of nodes, seperated by a new line
+  local nodes=$1
+  # Command to run on each node
+  local command=$2
+  # Parameter for command
+  local params=$3
```
ZacBlanco: I'm thinking now it might be better to utilize `shift` (http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_09_07.html) and then for params you should be able to use `$@` instead to get the remaining arguments. This way you don't need to quote the final set of args to the function.

ZacBlanco: Alternatively without using `shift` you could do `${@:3}` to append the 3rd and every subsequent arg into a single var here

ggezer: I liked the second one better. Thanks!

bin/alluxio-monitor.sh
```diff
@@ -175,27 +175,44 @@ run_monitors() {
...
-          run_on_node ${node} "${BIN}/alluxio-monitor.sh" ${mode} "${node_type}"
-        done
+        batch_run_on_nodes "$(echo ${nodes})" "${BIN}/alluxio-monitor.sh" ${mode} ${node_type}
       fi
     fi
   else
-    for node in $(echo "${nodes}"); do
-      run_on_node ${node} "${BIN}/alluxio-monitor.sh" ${mode} "${node_type}"
-    done
+    batch_run_on_nodes "$(echo ${nodes})" "${BIN}/alluxio-monitor.sh" ${mode} ${node_type}
```
ZacBlanco: should we still quote the `${node_type}` here? same for other locations

ggezer: Test will tell.

ggezer: It works this way, I guess because our node types are always single word.

ZacBlanco: It might not be a bad idea to quote these anyway? I would prefer that anyways, otherwise LGTM

ggezer: how about `${mode}` ?

ggezer:
@calvinjia PTAL.

ggezer:
Jenkins, test this please

bin/alluxio-monitor.sh
ZacBlanco: I think generally it is good practice to quote put variables expansions within quotations anyways. I would not be opposed to quoting both

ggezer: Makes sense. Done for both.

calvinjia:

calvinjia:

PR #9384, opened 2019-06-28
Load EMPTY_URI lazy in class AlluxioURI

liuzx32:
**Description**
AlluxioURI and URI are depend on each other, becuse AlluxioURI.EMPTY_URI is a static final variable.

**Improvement**
Adjust EMPTY_URI to lazy load with inner static class.
When other class need AlluxioURI.EMPTY_URI, AlluxioURI will really load EMPTY_URI instance.

### DIFF
--- a/core/base/src/main/java/alluxio/AlluxioURI.java
+++ b/core/base/src/main/java/alluxio/AlluxioURI.java
@@ -61,7 +61,19 @@ public final class AlluxioURI implements Comparable<AlluxioURI>, Serializable {
   public static final String CUR_DIR = ".";
   public static final String WILDCARD = "*";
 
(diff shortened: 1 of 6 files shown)


apc999:
@liuzx32 regarding "AlluxioURI and URI are depend on each other, becuse AlluxioURI.EMPTY_URI is a static final variable", can you describe more here? Does this introduce issues in compilation? Typically we prevent cross-module inter-dependency.

Introducing `EMPTY_URI` as a method like `AlluxioURI.EMPTY_URI()` is an unusual practice. There are different ways as a workaround like leveraging `Suppliers.memoize `

liuzx32:
@apc999 Thanks for your answer. It's not a big problem, I will close this PR.
Perhaps, we should add a class AlluxioHelper for AlluxioURI which like your consider.

PR #9397, opened 2019-07-01
Add import to fix build

ns1123:


core/common/src/test/java/alluxio/util/executor/ControllableSchedulerTest.java
```diff
@@ -15,6 +15,7 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import org.junit.Assert;
```
calvinjia: Instead of adding this, could you use the static imports in the `cancel` methods?

ns1123: done

apc999:

PR #9413, opened 2019-07-03
Fix read-only mount flag after restart

madanadit:


madanadit:
I'll add a test case as well

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -561,7 +561,9 @@ public void start(Boolean isPrimary) throws IOException {
         MountInfo mountInfo = mMountTable.getMountTable().get(key);
         UnderFileSystemConfiguration ufsConf =
             UnderFileSystemConfiguration.defaults(ServerConfiguration.global())
-            .createMountSpecificConf(mountInfo.getOptions().getPropertiesMap());
+                .createMountSpecificConf(mountInfo.getOptions().getPropertiesMap())
```
calvinjia: The API here is pretty confusing, it is not obvious we need to set the read only or shared fields separately.
For the fix to released branches I'm fine with this, to master we should should consider something else. How about a `fromMountInfo` method?

madanadit: makes sense. not targeting this PR though as its a big refactor

madanadit:
jenkins, test this please.

shaneknapp:
failure caused by testing out a new build node

shaneknapp:
test this please

madanadit:
jenkins, test this please.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: Can we file a tech-debt issue for this?

madanadit: done

madanadit:
@ZacBlanco do you have more comments?

ZacBlanco:
Can you link the issue to this thread/PR? I can't seem to find it.

Other than that, LGTM

madanadit:

madanadit:
alluxio-bot, cherry-pick this to master please.

PR #9422, opened 2019-07-06
Add integration test for read-only mount after restart

madanadit:
@ns1123 @ZacBlanco @calvinjia FYI this was not an issue in 1.8 so only adding the integration test here.


madanadit:

Issue #9231, opened 2019-06-07
Transparent URI


yuzhu:
**Is your feature request related to a problem? Please describe.**
Alluxio users often switch between accessing UFS directly and using Alluxio to access the data. In this process, they need to switch URI from something like s3a:// ... to alluxio://... . Sometimes making this change is easy, but in some cases, making this change is difficult because it could be long-lived, as is the case when using Hive metastore.

**Describe the solution you'd like**
Compute frameworks define the mapping from FileSystem scheme to FileSystem implementation. Instead of switching all URI usages from hdfs -> alluxio, switch the FileSystem scheme mapping so that URIs with hdfs:// schemes are routed to Alluxio's FileSystem implementation. Then users can enable Alluxio with nothing but a configuration update.

**Urgency**
Alluxio 2.1

liuzx32:
fs.s3.impl or fs.hdfs.impl can't directly depend on Alluxio's FileSystem, because Alluxio's FileSystem will check the scheme and the method getCheme() only return one sheme.
The scheme checking will throw exceptions below:  (s3 or hdfs)
-ls: URI scheme s3 does not match the expected scheme alluxio 
-ls: URI scheme hdfs does not match the expected scheme alluxio 
#

yuzhu:
@liuzx32 I think the intention of this issue is to change the logic in Alluxio's File System so that we can use URI schemes such as s3 and hdfs directly. Currently they won't work, because Alluxio File System expect alluxio scheme.

liuzx32:
@yuzhu Yes, we have the same mind.

ggezer:
Some low-level details on this issue:

There will be a new file system implementation called `ShimFileSystem` that could be passed in `fs.***.impl`.

`ShimFileSystem` will simply forward every request to master without doing any URI check. On master, a reverse-lookup will be done against active Alluxio mount points and given URI will be converted on the fly to Alluxio path. 

For example:
- Application, with `ShimFileSystem` configured for `fs.s3.impl`, accesses `s3://some-bucket/foo/bar`.
- `ShimFileSystem` relays the request to Alluxio with original URI. 
- Alluxio detects `some-bucket` s3 bucket is already mounted at `/s3-buckets/some-bucket`.
- Given URI is transparently translated to alluxio://s3-buckets/some-bucket/foo/bar` and request is served.

[...]

PR #9434, opened 2019-07-09
Adjust load command options readType from user config

liuzx32:
**Description**
The command load options for readtype is CACHE_PROMOTE which is hard coded.
When user set alluxio.user.file.read.type.default=NO_CACHE, the command load can also load data to CACHE, so the options of load command should be consistent with the user config alluxio.user.file.read.type.default.

**Improvement**
Adjust load command options readType from user config: alluxio.user.file.read.type.default.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
@@ -17,6 +17,8 @@ import alluxio.cli.CommandUtils;
 import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.URIStatus;
+import alluxio.conf.AlluxioConfiguration;
+import alluxio.conf.PropertyKey;
 import alluxio.exception.AlluxioException;
 import alluxio.exception.status.InvalidArgumentException;
 import alluxio.grpc.OpenFilePOptions;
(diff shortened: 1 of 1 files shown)


apc999:
thanks for the contribution @liuzx32 . do you have an example when users want to load data but with `alluxio.user.file.read.type.default=NO_CACHE`? It is not clear to me why users want to achieve this

liuzx32:
@apc999 In my opinion, Alluxio is not only for Cache, it contains cache, proxy and namespace.
Sometimes, we don't have enough memory and SSD, but we can unify the access api through alluxio. So we set alluxio.user.file.read.type.default=NO_CACHE, and then Alluxio can provide functions about proxy and namespace. When we have enough memory and ssd，we will open the Cache by setting alluxio.user.file.read.type.default=[CACHE|CACHE_PROMOTE].

liuzx32:
@apc999 I think that the transparent uri from https://github.com/Alluxio/alluxio/issues/9231 is an important function at the same time. Transparent uri can unify the storage for many cloud api without Cache. It should be a single important function.

ZacBlanco:
@liuzx32 if you're interested in loading metadata without loading data into the cache, then you can use `alluxio ls -Rf /path/to/file` to force loading the metadata into the alluxio namespace from the UFS without moving data around at all.

Is there something else that calling `alluxio fs load /path/to/file` with `NO_CACHE` provides that `alluxio ls -Rf` doesn't?

apc999:
> @apc999 I think that the transparent uri from #9231 is an important function at the same time. Transparent uri can unify the storage for many cloud api without Cache. It should be a single important function.

Definitely. Do you want to take on #9231 ?

liuzx32:
@apc999 It's not easy for me to take on #9231. And I haven't got time to do that yet.

liuzx32:
@ZacBlanco Yeah, you are right! thanks for the solution. But when I set alluxio.user.file.read.type.default=NO_CACHE, the load command will always load data to our little MEM CACHE, and it's a slowly process. We just want to keep consistent with user config.

maobaolong:
@liuzx32 @apc999 @ZacBlanco It this PR still active?

PR #9437, opened 2019-07-09
Fix leaking zk servers during various multi-master tests

ggezer:
Multi master test clusters makes use of embedded zookeeper server. This PR is an attempt to stabilize cluster initialization/termination code on some tests for making sure we don't leak embedded zookeeper servers, which could adversely effect following tests.

Since multi-master tests are usually run on a same JVM, one flake was causing whole bunch of tests to fail.

### DIFF
--- a/tests/src/test/java/alluxio/server/ft/MasterFaultToleranceIntegrationTest.java
+++ b/tests/src/test/java/alluxio/server/ft/MasterFaultToleranceIntegrationTest.java
@@ -278,11 +278,12 @@ public class MasterFaultToleranceIntegrationTest extends BaseIntegrationTest {
     // Stop the default cluster.
     after();
 
-    // Create a new cluster, with no workers initially
-    final MultiMasterLocalAlluxioCluster cluster = new MultiMasterLocalAlluxioCluster(2, 0);
-    cluster.initConfiguration();
-    cluster.start();
+    MultiMasterLocalAlluxioCluster cluster = null;
     try {
+      // Create a new cluster, with no workers initially
+      cluster = new MultiMasterLocalAlluxioCluster(2, 0);
+      cluster.initConfiguration();
+      cluster.start();
       // Get the first block master
       BlockMaster blockMaster1 =
           cluster.getLocalAlluxioMaster().getMasterProcess().getMaster(BlockMaster.class);
@@ -343,7 +344,9 @@ public class MasterFaultToleranceIntegrationTest extends BaseIntegrationTest {
           Collections.EMPTY_MAP, Collections.EMPTY_MAP, Collections.EMPTY_MAP,
           RegisterWorkerPOptions.getDefaultInstance());
     } finally {
-      cluster.stop();
+      if (cluster != null) {
+        cluster.stop();
+      }
     }
 
     // Start the default cluster.
--- a/tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
+++ b/tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
@@ -183,7 +183,8 @@ public class JournalShutdownIntegrationTest extends BaseIntegrationTest {
     try {
       cluster = new MultiMasterLocalAlluxioCluster(TEST_NUM_MASTERS);
       cluster.initConfiguration();
-      cluster.stop();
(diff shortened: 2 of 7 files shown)


ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

tests/src/test/java/alluxio/server/ft/MasterFaultToleranceIntegrationTest.java
```diff
@@ -278,11 +278,12 @@ public void failoverWorkerRegister() throws Exception {
...
     after();
 
-    // Create a new cluster, with no workers initially
-    final MultiMasterLocalAlluxioCluster cluster = new MultiMasterLocalAlluxioCluster(2, 0);
-    cluster.initConfiguration();
-    cluster.start();
+    MultiMasterLocalAlluxioCluster cluster = null;
     try {
+      // Create a new cluster, with no workers initially
+      cluster = new MultiMasterLocalAlluxioCluster(2, 0);
```
ZacBlanco: maybe we in the future can make MultiMasterAlluxioCluster implement the `AutoCloseable` interface so that we can use the `try-with-resources` syntax

ggezer:
@calvinjia PTAL.

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
```diff
@@ -178,15 +178,24 @@ public void singleMasterMountUnmountJournal() throws Exception {
...
-    // Kill the leader one by one.
-    for (int kills = 0; kills < TEST_NUM_MASTERS; kills++) {
-      cluster.waitForNewMaster(120 * Constants.SECOND_MS);
-      assertTrue(cluster.stopLeader());
+    MultiMasterLocalAlluxioCluster cluster = null;
+    UnderFileSystemFactory factory = null;
+    try {
+      cluster = new MultiMasterLocalAlluxioCluster(TEST_NUM_MASTERS);
+      cluster.initConfiguration();
+      cluster.stop();
```
calvinjia: Why can we stop the cluster here?

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
```diff
@@ -178,15 +178,24 @@ public void singleMasterMountUnmountJournal() throws Exception {
...
-    MultiMasterLocalAlluxioCluster cluster = setupMultiMasterCluster();
-    UnderFileSystemFactory factory = mountUnmount(cluster.getClient());
-    // Kill the leader one by one.
-    for (int kills = 0; kills < TEST_NUM_MASTERS; kills++) {
-      cluster.waitForNewMaster(120 * Constants.SECOND_MS);
-      assertTrue(cluster.stopLeader());
+    MultiMasterLocalAlluxioCluster cluster = null;
+    UnderFileSystemFactory factory = null;
+    try {
+      cluster = new MultiMasterLocalAlluxioCluster(TEST_NUM_MASTERS);
```
calvinjia: Should construction of a new cluster always be in the try block? We include it here but leave it outside the try for `MultiProcessClusters`.

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
ggezer: Nice catch. It should have been stopLeader.

tests/src/test/java/alluxio/server/ft/journal/JournalShutdownIntegrationTest.java
ggezer: Most tests are using `before` and `after` methods for setting up and tearing clusters, which makes using try statements a bigger change.

tests/src/test/java/alluxio/server/ft/MasterFaultToleranceIntegrationTest.java
ggezer: I realized most tests are using `before` and `after` methods for setting up and tearing clusters, which makes using try statements a bigger change.

calvinjia:

PR #9438, opened 2019-07-09
Isolate zk client config for suppressing authentication

ggezer:
For when more than one zk servers are accessed within a single jvm, global authentication settings are shared via system properties. This PR bumps zk client version for getting a support for client config isolation and makes use of it. Curator has also been updated for compatibility.


ggezer:
@bf8086 @madanadit PTAL. Thanks!

core/common/pom.xml
```diff
@@ -57,7 +57,8 @@
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
-      <artifactId>netty-all</artifactId>
+      <artifactId>netty</artifactId>
+      <version>3.7.0.Final</version>
```
madanadit: why change the version only for this specific module?

core/common/src/main/java/alluxio/master/ZkMasterInquireClient.java
```diff
@@ -78,16 +97,20 @@ public static synchronized ZkMasterInquireClient getClient(String zookeeperAddre
...
     mElectionPath = electionPath;
 
     LOG.info("Creating new zookeeper client for {}", connectDetails);
-    // Start the client lazily.
-    mClient = CuratorFrameworkFactory.newClient(connectDetails.getZkAddress(),
-        new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
+    CuratorFrameworkFactory.Builder curatorBuilder = CuratorFrameworkFactory.builder();
+    curatorBuilder.connectString(connectDetails.getZkAddress());
+    curatorBuilder.retryPolicy(new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
+    curatorBuilder.zookeeperFactory(new NoSaslZookeeperFactory());
```
madanadit: i don't get it. where is authType used now?

core/common/pom.xml
ggezer: This module uses `ConcurrentIdentityMap` implementation from netty which is dropped after this version. And that's the only use of netty from this module.

Actually we don't need an explicit netty dependency anymore. I can have another PR to remove netty dependencies throughout the code. (Locally I tested that removing netty dependencies, except this one, has no effect on build.)

core/common/src/main/java/alluxio/master/ZkMasterInquireClient.java
ggezer: It's for future use.

core/common/src/main/java/alluxio/master/MasterInquireClient.java
```diff
@@ -97,7 +99,8 @@ public static MasterInquireClient createForJobMaster(AlluxioConfiguration conf)
         return ZkMasterInquireClient.getClient(conf.get(PropertyKey.ZOOKEEPER_ADDRESS),
             conf.get(PropertyKey.ZOOKEEPER_JOB_ELECTION_PATH),
             conf.get(PropertyKey.ZOOKEEPER_JOB_LEADER_PATH),
-            conf.getInt(PropertyKey.ZOOKEEPER_LEADER_INQUIRY_RETRY_COUNT));
+            conf.getInt(PropertyKey.ZOOKEEPER_LEADER_INQUIRY_RETRY_COUNT),
+            conf.getEnum(PropertyKey.SECURITY_AUTHENTICATION_TYPE, AuthType.class));
```
madanadit: does this mean the auth type is now a required client side property? i.e. we cannot depend on cluster defaults handshake to determine the auth type.

ggezer: I don't think so. All creators of `MasterInquireClient` use cluster conf directly or indirectly.

ggezer: Doesn't sound OK though, given it's the inquire client. I'll need to test this.

ggezer: Actually, I'm kind of sure there is a path that calls this before fetching configs.

One solution would be to disable zk client authentication only when instructed by a config property. So authentication could disabled only when it's explicitly required by specific deployments. Sent an iteration for implementing it this way.

core/common/src/main/java/alluxio/master/ZkMasterInquireClient.java
```diff
@@ -53,22 +56,43 @@
   private final CuratorFramework mClient;
   private final int mInquireRetryCount;
 
+  /**
+   * Zookeeper factory for curator that explicitly disables authentication.
```
madanadit: nit: update comment

core/common/src/main/java/alluxio/master/MasterInquireClient.java
madanadit: sounds good

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -525,7 +525,11 @@ public String toString() {
           .setDescription("Session timeout to use when connecting to Zookeeper")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
-
+  public static final PropertyKey ZOOKEEPER_AUTH_ENABLED =
+      new Builder(Name.ZOOKEEPER_AUTH_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("If true, enable client-side Zookeeper authentication.")
+          .build();
```
bf8086: (nit) set the scope and consistency level for this property.

ggezer:
@madanadit PTAL.

core/common/pom.xml
madanadit: did you create a tech dept issue for this?

calvinjia: If we can remove the direct dependency on Netty that would be helpful.

ggezer: https://github.com/Alluxio/alluxio/issues/9449

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -525,7 +525,13 @@ public String toString() {
           .setDescription("Session timeout to use when connecting to Zookeeper")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
-
+  public static final PropertyKey ZOOKEEPER_AUTH_ENABLED =
+      new Builder(Name.ZOOKEEPER_AUTH_ENABLED)
```
ggezer: @ns1123 FYI.

madanadit:

madanadit:

madanadit:
@ggezer please cherry-pick to branch-1.8

PR #6497, opened 2017-11-21
[SMALLFIX] Update Netty minor version

calvinjia:
Requires the changes to remove usages of chmv8. Version >= 4.0.50 is required to support shading without needing to specify system properties.

PR #9455, opened 2019-07-12
Clean io.netty dependencies

ggezer:
Cleaning up io.netty configuration from modules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,6 +142,7 @@
     <glusterfs-hadoop.version>2.3.13</glusterfs-hadoop.version>
     <gmetric4j.version>1.0.7</gmetric4j.version>
     <grpc.version>1.17.1</grpc.version>
+    <netty.version>4.1.30.FINAL</netty.version>
     <hadoop.version>2.2.0</hadoop.version>
     <hadoop-openstack.version>2.6.0</hadoop-openstack.version>
     <java.version>1.8</java.version>
@@ -334,6 +335,11 @@
         <groupId>io.grpc</groupId>
         <artifactId>grpc-stub</artifactId>
         <version>${grpc.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-all</artifactId>
+        <version>${netty.version}</version>
       </dependency>
       <dependency>
         <groupId>io.prometheus</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -335,12 +335,6 @@
         <artifactId>grpc-stub</artifactId>
         <version>${grpc.version}</version>
       </dependency>
-      <!-- Netty 4.0.50+ required for shading in client runtime jar to work -->
-      <dependency>
-        <groupId>io.netty</groupId>
-        <artifactId>netty-all</artifactId>
-        <version>4.1.30.Final</version>
-      </dependency>
       <dependency>
         <groupId>io.prometheus</groupId>
         <artifactId>simpleclient</artifactId>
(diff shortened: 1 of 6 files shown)


ggezer:
@apc999 PTAL. Thanks!

ggezer:
@calvinjia PTAL. Thanks!

pom.xml
```diff
@@ -400,37 +394,16 @@
         <groupId>org.apache.curator</groupId>
         <artifactId>curator-client</artifactId>
         <version>${apache.curator.version}</version>
-        <exclusions>
-          <!-- Excluded to avoid implementation conflict with io.netty:netty-all -->
-          <exclusion>
-            <groupId>org.jboss.netty</groupId>
-            <artifactId>netty</artifactId>
-          </exclusion>
-        </exclusions>
```
calvinjia: Although we have removed our explicit dependency on netty, don't several of our dependencies (grpc, zk) transitively depend on netty? Is this safe to remove?

pom.xml
```diff
@@ -335,12 +335,6 @@
         <artifactId>grpc-stub</artifactId>
         <version>${grpc.version}</version>
       </dependency>
-      <!-- Netty 4.0.50+ required for shading in client runtime jar to work -->
-      <dependency>
-        <groupId>io.netty</groupId>
-        <artifactId>netty-all</artifactId>
-        <version>4.1.30.Final</version>
```
calvinjia: Should we remove this since our dependencies pull netty and we may want to fix the version?

pom.xml
ggezer: They'll bring their own versions. We won't be interfering with their dependencies anymore.

pom.xml
ggezer: Why do we need this for shading to work? @apc999 ?

apc999: - first, this is in `<dependencyManagement>` and I don't think version set here affect transitive dependencies
- regarding why we need this for shading to work, this is actually a fix by calvin https://github.com/Alluxio/alluxio/pull/6497 . w.r.t. that PR, should we also remove this line https://github.com/Alluxio/alluxio/blob/master/core/client/runtime/pom.xml#L90 , @calvinjia

apc999:
hooray! can not wait for this moment!

pom.xml
ggezer: Some dependencies might be bringing these as well. I am for not touching the shading directives.

pom.xml
calvinjia: don't our dependencies rely on different versions?

pom.xml
apc999: @calvinjia  Ping

calvinjia: @apc999 Waiting on response for question above:
Isn't dependency management also used to manage the version brought in by the transitive dependencies? For example if A and B depend on Netty 1.0 and Netty 1.4, we can use dependency management to enforce the version used instead of implicitly (based on order of declaration/dependency depth).

ggezer:
@calvinjia @apc999 I'll park this PR for a while. I'm currently working on getting another transitive netty dependency out of our way, then will continue this one.

pom.xml
ggezer: @calvinjia Right. Maven will enforce transitive dependency version.
At current state, without this direction we're pulling 2 versions: 4.1.30 for grpc, and 4.1.29 for zookeeper
With this directive we can unify the versions.
I'm reverting this directive.

pom.xml
ggezer: Current curator version depends on `netty-all`, indirectly via zookeeper. I don't think we need this directive anymore.

calvinjia:
Jenkins, test this please.

ggezer:
Jenkins, test this please

calvinjia:

PR #9456, opened 2019-07-12
Enable owner to chmod, even without write privilege

bf8086:
Currently owner of file/directory cannot change permission when they don't have write privilege. This is inconsistent with Unix and HDFS. This fix updated the permission requirement so the user should be able to perform chmod if and only if the user is the owner or super user.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3016,9 +3016,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
     // for chgrp, chmod
     boolean ownerRequired = (options.hasGroup()) || (options.hasMode());
     // for other attributes
-    boolean writeRequired = options.hasPersisted() || options.hasPinned()
-        || options.hasReplicationMax() || options.hasReplicationMin()
-        || options.getPinnedMediaCount() > 0;
+    boolean writeRequired = !rootRequired && !ownerRequired;
     if (options.hasOwner() && options.hasGroup()) {
       try {
         checkUserBelongsToGroup(options.getOwner(), options.getGroup());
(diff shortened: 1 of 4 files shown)


bf8086:
@gpang PTAL.

bf8086:
@madanadit @ns1123 FYI.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3015,6 +3015,10 @@ public void setAttribute(AlluxioURI path, SetAttributeContext context)
     boolean rootRequired = options.hasOwner();
     // for chgrp, chmod
     boolean ownerRequired = (options.hasGroup()) || (options.hasMode());
+    // for other attributes
+    boolean writeRequired = options.hasPersisted() || options.hasPinned()
```
gpang: when does this list of operations need to be updated? What is not included in this list of operations?

bf8086: This is is everything beside permission related updates. It needs to be updated when a new field is added to the option / inode.

gpang: will this be hard to maintain? Should it be checked the other way around, meaning, if it is a permission change, then only owner is required, otherwise, write is required?

It will be very easy to forget to update this list of non-permission related updates.

bf8086: It's is hard to maintain either way. This original issue happens because we added set owner and set mode and we did not exclude them from checking write permission. I agree that if we forget to update it should at least check write permission, but we can still have bugs if we add new attributes which requires different permissions to set.

bf8086:
Jenkins, test this please.

bf8086:
@gpang Thanks! I updated the PR. PTAL.

gpang:

PR #9462, opened 2019-07-15
insert a space between two word.

maobaolong:
A minor bugfix, just insert a space between two word.

### DIFF
--- a/docs/en/Architecture-DataFlow.md
+++ b/docs/en/Architecture-DataFlow.md
@@ -108,7 +108,7 @@ recovery. Different from standby masters, a secondary master can never upgrade t
 ### Job Masters
 
 The Alluxio Job Master is a standalone process which is responsible for scheduling a number of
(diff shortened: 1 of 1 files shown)


docs/en/Architecture-DataFlow.md
```diff
@@ -108,7 +108,7 @@ recovery. Different from standby masters, a secondary master can never upgrade t
 ### Job Masters
 
 The Alluxio Job Master is a standalone process which is responsible for scheduling a number of
-more heavyweight file system operations asynchronously within Alluxio.
```
ZacBlanco: Thanks for looking at this @maobaolong! The usage of "heavyweight" was intentional when the section was authored.

A more accurate wording might be "resource intensive"? What do you think?

apc999: ping @maobaolong

ZacBlanco:
There's been no progress on this PR for a while. Please re-open if you would like to continue the work.

Issue #9399, opened 2019-07-01
Add noEvictor class to one layer of tieredstore


cheyang:
**Is your feature request related to a problem? Please describe.**
In Deep learning scenario, we are using `MEM` layer to speed up training speed. But the eviction action in `MEM` layer produces additional overhead. 

**Describe the solution you'd like**
Can we add `noEvictor` to `alluxio.worker.evictor.class`, and make it for MEM layer? it just doesn't add more data to `MEM` layer. And the deep learning application can be benefit with the memory layer without swap in and out.

**Describe alternatives you've considered**

**Urgency**
Explain why the feature is important

**Additional context**

apc999:
@cheyang  thanks for proposing the idea. Wondering if command `pin`([doc](https://docs.alluxio.io/os/user/2.0/en/basic/Command-Line-Interface.html#pin)) work for you for this particular purpose? Basically, "pinned" data in Alluxio space will not be evicted out.

cheyang:
Yes, I think this one should be good for my scenario. I will try it later.

cheyang:
@apc999 , I see the following error. As you know the data for training is always read-only. Any suggestions?

```
/opt/alluxio/bin/alluxio fs pin  /training-data/images/
A write operation on /training-data/images under a readonly mount point /training-data/images is not allowed
```

apc999:
Pin should not be a write operation.  @cheyang  let me fix this issue

PR #9460, opened 2019-07-13
Fix failed Pin on readonly mount

apc999:
Fix https://github.com/Alluxio/alluxio/issues/9399 where `pin` failed due to readonly mount

PR #9466, opened 2019-07-15
Fix failed Pin on readonly mount

apc999:
Fix https://github.com/Alluxio/alluxio/issues/9399 where `pin` failed
due to readonly mount
Also fix a bug on wrong command name in audit log

pr-link: Alluxio/alluxio#9460
change-id: cid-2e62e472a4aa270b838fc419c2b5d6fbc71252c0

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3146,12 +3146,16 @@ public final class DefaultFileSystemMaster extends AbstractMaster implements Fil
       }
     }
     String commandName;
+    boolean checkWritableMountPoint = false;
     if (options.getOwner() != null) {
       commandName = "chown";
+      checkWritableMountPoint = true;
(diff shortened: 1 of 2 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3146,12 +3146,16 @@ public void setAttribute(AlluxioURI path, SetAttributeOptions options)
       }
     }
     String commandName;
+    boolean checkWritableMountPoint = false;
     if (options.getOwner() != null) {
       commandName = "chown";
+      checkWritableMountPoint = true;
     } else if (options.getGroup() != null) {
       commandName = "chgrp";
-    } else if (options.getMode() != null) {
```
apc999: this is a bug that `options.getMode()` will never be `null` as the default value is already `Constants.INVALID_MODE`.
So it will show an wrong command name as `chmod` if it is really Pin or Setttl

apc999:
Jenkins, test this please

calvinjia:

PR #9471, opened 2019-07-16
Add metadataSyncInterval to client RPCs by default

### LINKED ISSUE #9470: SetAttribute operations will always loadmetadata
**Alluxio Version:**
v2.0.0

**Describe the bug**
For FS operations implemented by the master-side API setAttribute, including pin, setTtl, chmod, chown, chgrp, setReplication, `loadMetadataInterval` on server side will always be `0` due to the missing `commonOptions` in `SetAttributeOptions`. This results always unnecessary, possibly too much UFS load metadata operations even the file metadata has been cached already.

**To Reproduce**

mount a local fs containing file `foo` then `fs pin` this file which will succeed.
versus in 1.8 one needs `fs ls` first before being able to `fs pin`

**Expected behavior**
Two possible options:
1. convert back to behavior of 1.8 that makes default `synInterval` -1 instead of 0 (due to protobuf)
2. introduce a new behavior that it will sync on-demand for the first time, and follows sync interval in the future

**Urgency**
High

**Additional context**

ZacBlanco:
The SetAttribute RPC in the client works via omission of
properties. If a property is included within the RPC options
object, then it is assumed the user set this property and
should then be set on the inode. This works for all fields
within the SetAttribute call except for the sync interval.

The metadata sync interval must be included on client RPCs.
Otherwise, if it is not included, then the
DefaultFileSystemMaster will assume that the user has
implicitly set a sync interval of 0 - which results in syncing
metadata for every RPC. Users can still override this value
from their configuration. However, by default it should be
sending a value of -1 (the default for the property key).

This change makes sure that the interval is set and adds a
small unit test.

Fixes #9470

### DIFF
--- a/core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
+++ b/core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
@@ -261,6 +261,9 @@ public class FileSystemOptions {
    * @return options based on the configuration
    */
   public static SetAttributePOptions setAttributeClientDefaults(AlluxioConfiguration conf) {
+    // Specifically set and override *only* the metadata sync interval
+    // Setting other attributes by default will make the server think the user is intentionally
+    // setting the values. Most fields withinSetAttributePOptions are set by inclusion
     return SetAttributePOptions.newBuilder()
         .setCommonOptions(FileSystemMasterCommonPOptions.newBuilder()
             .setSyncIntervalMs(conf.getMs(PropertyKey.USER_FILE_METADATA_SYNC_INTERVAL))
--- a/core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
+++ b/core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
@@ -256,7 +256,8 @@ public class FileSystemOptions {
   }
 
   /**
-   * Sets the defaults for the SetAttribute RPC which should be used on the client side.
+   * Defaults for the SetAttribute RPC which should only be used on the client side.
+   *
    * @param conf Alluxio configuration
    * @return options based on the configuration
    */
--- a/core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
+++ b/core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
@@ -256,6 +256,19 @@ public class FileSystemOptions {
   }
 
   /**
+   * Sets the defaults for the SetAttribute RPC which should be used on the client side.
+   * @param conf Alluxio configuration
+   * @return options based on the configuration
+   */
+  public static SetAttributePOptions setAttributeClientDefaults(AlluxioConfiguration conf) {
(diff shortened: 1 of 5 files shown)


ZacBlanco:
@apc999 FYI

@ggezer @gpang PTAL

ZacBlanco:
The issue is that if I run

```console
$ ./bin/alluxio fs -Dalluxio.user.file.metadata.sync.interval=1s pin /path/to/file
```

As of right now, the file's metadata won't be sending the `1s` sync interval to the master.

ZacBlanco:
jenkins, test this please

core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
```diff
@@ -255,6 +255,19 @@ public static SetAttributePOptions setAttributeDefaults(AlluxioConfiguration con
         .build();
   }
 
+  /**
+   * Sets the defaults for the SetAttribute RPC which should be used on the client side.
+   * @param conf Alluxio configuration
+   * @return options based on the configuration
+   */
+  public static SetAttributePOptions setAttributeClientDefaults(AlluxioConfiguration conf) {
+    return SetAttributePOptions.newBuilder()
```
gpang: can we add a comment or javadoc being explicit that we are only setting sync interval?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4540,8 +4540,11 @@ public RpcContext createRpcContext() throws UnavailableException {
 
   private LockingScheme createLockingScheme(AlluxioURI path, FileSystemMasterCommonPOptions options,
       LockPattern desiredLockMode) {
-    boolean shouldSync =
-        mUfsSyncPathCache.shouldSyncPath(path.getPath(), options.getSyncIntervalMs());
+    // If client options didn't specify the interval, fallback to whatever the server has
+    // configured to prevent unnecessary syncing due to the default value being 0
+    long syncInterval = options.hasSyncIntervalMs() ? options.getSyncIntervalMs() :
```
gpang: so the default proto will return 0 for `get` but `hasSyncIntervalMs` will return `false`?

ZacBlanco: Yes, that is correct. If the client specifically set `0` for the interval, then `hasSyncIntervalMs` will return `true`

core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
ZacBlanco: moved an old comment

ZacBlanco:
jenkins, test this please

gpang:

ZacBlanco:
@gpang cherry pick branch 2.0?

apc999:

PR #9473, opened 2019-07-16
Fix unencoded paths for the web ui

### LINKED ISSUE #9465: Path components are not uri encoded in the webui
**Alluxio Version:**
2.0.0

**Describe the bug**
Path components are not uri encoded causing pages for files with reserved characters (ie. `=`) to fail to render.

**To Reproduce**
Browse to a file with and `=` in it.

calvinjia:
Fixes #9465 

Invariants -
Paths presented to the user are not encoded.
Paths in URLs/links are encoded, except for specific references to the root path (a constant `/`).

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
@@ -96,6 +96,7 @@ import java.io.FileInputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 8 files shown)


calvinjia:
Jenkins, test this please.

calvinjia:
@ggezer Have we deflaked this test? `alluxio.client.cli.JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal`

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -367,8 +368,8 @@ public Response getWebUIBrowse(@DefaultValue("/") @QueryParam("path") String req
           ServerConfiguration.getBoolean(PropertyKey.SECURITY_AUTHORIZATION_PERMISSION_ENABLED))
           .setMasterNodeAddress(mMasterProcess.getRpcAddress().toString()).setInvalidPathError("");
       List<FileInfo> filesInfo;
-      String path = requestPath;
-      if (path == null || path.isEmpty()) {
+      String path = URLDecoder.decode(requestPath, "UTF-8");
+      if (path.isEmpty()) {
         path = AlluxioURI.SEPARATOR;
```
apc999: don't we need to encode this too?

ggezer:
@calvinjia it still flakes, but less frequently. I'll need to have a deeper look into checkpoint rpc in raft journal as it mostly fails during that.

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
calvinjia: We only need to decode what is sent to us by the client in `requestPath`.

calvinjia:

calvinjia:

PR #9486, opened 2019-07-18
Support writing to zk 3.4.x servers

ggezer:
After bumping zk client version to 3.5.5, leader election was broken against 3.4.x servers.
This fix restores the support by making sure unsupported features are not used in write path.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
+++ b/core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
@@ -174,20 +174,26 @@ public final class PrimarySelectorClient extends AbstractPrimarySelector
    * @return a new {@link CuratorFramework} client to use for leader selection
    */
   private CuratorFramework getNewCuratorClient() {
-    CuratorFramework client = CuratorFrameworkFactory.newClient(mZookeeperAddress,
-        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_SESSION_TIMEOUT),
-        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_CONNECTION_TIMEOUT),
-        new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
-    client.start();
+    LOG.info("Creating new zookeeper client for primary selector {}", mZookeeperAddress);
+    CuratorFrameworkFactory.Builder curatorBuilder = CuratorFrameworkFactory.builder();
+    curatorBuilder.connectString(mZookeeperAddress);
+    curatorBuilder.retryPolicy(new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
+    curatorBuilder
+        .sessionTimeoutMs((int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_SESSION_TIMEOUT));
+    curatorBuilder.connectionTimeoutMs(
+        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_CONNECTION_TIMEOUT));
+    // Force compatibility mode to support writing to 3.4.x servers.
+    curatorBuilder.zk34CompatibilityMode(true);
+    // Prevent using container parents as it breaks compatibility with 3.4.x servers.
+    // This is only required if the client is used to write data to zookeeper.
+    curatorBuilder.dontUseContainerParents();
 
     // Sometimes, if the master crashes and restarts too quickly (faster than the zookeeper
     // timeout), zookeeper thinks the new client is still an old one. In order to ensure a clean
     // state, explicitly close the "old" client and recreate a new one.
-    client.close();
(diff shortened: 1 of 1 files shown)


ggezer:
@calvinjia PTAL. Thanks!

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -174,20 +174,26 @@ public void takeLeadership(CuratorFramework client) throws Exception {
    * @return a new {@link CuratorFramework} client to use for leader selection
    */
   private CuratorFramework getNewCuratorClient() {
-    CuratorFramework client = CuratorFrameworkFactory.newClient(mZookeeperAddress,
-        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_SESSION_TIMEOUT),
-        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_CONNECTION_TIMEOUT),
-        new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
-    client.start();
+    LOG.info("Creating new zookeeper client for primary selector {}", mZookeeperAddress);
```
calvinjia: Will this be noisy?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -174,20 +174,26 @@ public void takeLeadership(CuratorFramework client) throws Exception {
...
+    CuratorFrameworkFactory.Builder curatorBuilder = CuratorFrameworkFactory.builder();
+    curatorBuilder.connectString(mZookeeperAddress);
+    curatorBuilder.retryPolicy(new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
+    curatorBuilder
+        .sessionTimeoutMs((int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_SESSION_TIMEOUT));
+    curatorBuilder.connectionTimeoutMs(
+        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_CONNECTION_TIMEOUT));
+    // Force compatibility mode to support writing to 3.4.x servers.
+    curatorBuilder.zk34CompatibilityMode(true);
```
calvinjia: Should this be configurable?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -174,20 +174,26 @@ public void takeLeadership(CuratorFramework client) throws Exception {
...
 
     // Sometimes, if the master crashes and restarts too quickly (faster than the zookeeper
     // timeout), zookeeper thinks the new client is still an old one. In order to ensure a clean
     // state, explicitly close the "old" client and recreate a new one.
-    client.close();
-    client = CuratorFrameworkFactory.newClient(mZookeeperAddress,
-        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_SESSION_TIMEOUT),
-        (int) ServerConfiguration.getMs(PropertyKey.ZOOKEEPER_CONNECTION_TIMEOUT),
-        new ExponentialBackoffRetry(Constants.SECOND_MS, 3));
+    curatorBuilder.build().close();
```
calvinjia: Unrelated to this PR - how expensive is this close/recreate?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: This is created once during master process init.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: I don't know really. All I can say is it just steps down immediately during debugging, so not much really. I think it's OK even if it's costly, considering it's only executed once during process creation.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: Curator recipe we're using `LeaderElection` is working fine with these limitations. I don't think we'll ever need to switch this off for zk 3.5.x servers since limited functionality is not that essential. 

For our use of zookeeper via curator:

- `zk34CompatibilityMode` causes some previously internal functionality to be invoked via reflection. Rather than going through exposed API in 3.5.x.
- `dontUseContainerParents` just causes to use specific type of `create` calls rather than calling new overload that has an option to create parents recursively.

ggezer:
@calvinjia Please see comments for details on cost of compatibility switches. Additionally, since this compatibility is limited to our curator instance, we don't have any side effect to the other users of zookeeper in the classpath.

calvinjia:

PR #9487, opened 2019-07-18
Improve the alluxio-fuse script

gpang:
Cherry-pick of Alluxio/alluxio#8365
Cherry-pick commit is Alluxio/alluxio@4af760027df217f5833091465deaf33b77853869

### DIFF
--- a/integration/fuse/bin/alluxio-fuse
+++ b/integration/fuse/bin/alluxio-fuse
@@ -116,26 +116,6 @@ fi
 
 case $1 in
   mount)
-<<<<<<< HEAD
-    if [[ $# -eq 2 ]]; then
-      mount_fuse $2 /
-||||||| parent of 4af760027d...  Fix various minor issues in alluxio-fuse script (#8365)
-    shift
-    while getopts "o:" option; do
-      case "${option}" in
-        o)
-          mount_option=${OPTARG}
-          ;;
-        *)
-          echo -e "${USAGE}" >&2
-          exit 1
-          ;;
-      esac
-    done
-    shift $((${OPTIND} - 1))
-    if [[ $# -eq 1 ]]; then
-      mount_fuse $1 /
-=======
     shift
     while getopts "o:" option; do
       case "${option}" in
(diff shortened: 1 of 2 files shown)


gpang:
@apc999 Please take a look, since I had to change some of the script.

integration/fuse/bin/alluxio-fuse
```diff
@@ -88,17 +90,17 @@ umount_fuse() {
...
     return 1
   else
     echo -e "pid\tmount_point\talluxio_path"
     echo -e "$(ps aux | grep [A]lluxioFuse | awk -F' ' '{print $2 "\t" $(NF-2) "\t" $NF}')"
-    return 0    
+    return 0
   fi
 }
 
-USAGE_MSG="Usage:\n\t$0 [mount|umount|stat]"
```
liuhongtong: $0 represents current name of this script, even though alluxio-fuse may be renamed. I think there is no need to modify $0

integration/fuse/bin/alluxio-fuse
```diff
@@ -114,30 +116,47 @@ fi
...
-      mount_fuse $2 $3
+    if [[ $# -eq 1 ]]; then
+      mount_fuse $1 /
+      exit $?
+    fi
+    if [[ $# -eq 2 ]]; then
+      mount_fuse $1 $2
       exit $?
     fi
-    echo -e "Usage\n\t$0 mount mount_point [alluxio_path]\n\t alluxio_path is default to root" >&2
```
liuhongtong: same here

integration/fuse/bin/alluxio-fuse
```diff
@@ -114,30 +116,47 @@ fi
...
+    echo -e "Usage\n\talluxio-fuse mount mount_point [alluxio_path]\n\t alluxio_path is default to root" >&2
     exit 1
     ;;
-  umount)    
+  umount|unmount)
     if [[ $# -eq 2 ]]; then
       umount_fuse $2
       exit $?
     fi
-    echo -e "Usage\n\t$0 umount mount_point\n\tuse mount stat to show mount points" >&2
```
liuhongtong: same here

integration/fuse/bin/alluxio-fuse
apc999: $0 will generate the full path, which is not common in the usage of a command

gpang: @apc999 what do you think?

liuhongtong: basename $0 may be more appropriate

apc999: @gpang I think  `basename $0` can be better

gpang:
@apc999 
I would prefer to have this cherry-pick PR correctly match the behavior of the original commit, and in a separate PR, change the behavior. Therefore, this PR will not implement the `basename $0` changes.

apc999:
@gpang  SG. let's stick to the original cherry-pick . I will submit separate PR to change `$0` to `basename $0` and make it available in different branches

apc999:

PR #9498, opened 2019-07-18
Fix 'Display from byte offset' input

### LINKED ISSUE #9483: Fix display from byte offset field in the web ui
**Alluxio Version:**
2.0.0

**Describe the bug**
Display from byte offset field will not allow empty values. This is confusing to the user because they generally will try to backspace the `0` before entering the number they want.

**To Reproduce**
Go to browse -> click a file. Try to modify the display from byte offset field.

**Expected behavior**
The zero is deleted and the user is free to type the number of their choice.

tieujason330:
- Fixes #9483
- making sure we initialize the field with `0` offset even if parsed query has `undefined` offset


tieujason330:
Future improvement suggestion - `Browse.tsx` and `Logs.tsx` are both similar - they either display a directory or a file. We can create a parent component that houses shared logic. Then each can derive from the parent, adding their own functionality

tieujason330:
@calvinjia PTAL, thanks

calvinjia:

PR #9505, opened 2019-07-19
Refactor Alluxio EMR script

### LINKED ISSUE #9477: Unified way to  append arbitrary Alluxio property keys in EMR bootstrap and docker CLI

https://github.com/Alluxio/alluxio/blob/master/integration/emr/alluxio-emr.sh#L82

> # Arg 3. Extra Alluxio Options. These will be appended to alluxio-site.properties. Multiple options can be specified using ';' as a delimiter

However, ';' as a delimiter might introduce issues as it may be a part of a property value, see https://docs.alluxio.io/os/user/v2.0/en/deploy/Running-Alluxio-On-a-HA-Cluster.html

A more general solution to pass in multiple property keys

**Urgency**
Medium

**Additional context**

### LINKED ISSUE #9478: Support specifying worker memory size in EMR bootstrap script

Currently "20GB" per worker was hard-coded in https://github.com/Alluxio/alluxio/blob/master/integration/emr/alluxio-emr.sh#L76

The script should be able to accept different worker memory size. Maybe let users to specify the value in `alluxio.worker.memory.size` in `args`.

**Urgency**
Medium

**Additional context**

### LINKED ISSUE #9489: Support ability to pass in a file as an arg to EMR bootstrap to append to alluxio-site.properties
Some time Alluxio properties are quite complex. If a bootstrap doesn't allow a file to be passed in, updating a large number of Alluxio properties every time an EMR cluster comes up, can lead to human errors. 

pass a file in with properties that get appended to alluxio-site.properties

ZacBlanco:
This refactor includes a number of improvements to the Alluxio EMR
bootstrap script.

- Users can optionally specify a remote `alluxio-site.properties` file
to add arbitrary properties without adding them when launching
through a local `emr` command or in the web UI (Fixes #9489)
- Dynamically sets memory to 1/3 of available system RAM on
workers. Previously was hard-coded to 20GB (Closes #9478)
- Supports an optional argument for the user to specify the
delimiter used when passing arbitrary properties on the command
line. Previously this was hard-coded to `;` by default. Users can
choose to continue using `;` by specifying an empty string `""` or
`";"`. Additionally, in place of `;`, the third argument of the script
now allows the user to specify delimiter of their choice.
(Closes #9477)

Other various refactoring and bash programming style has been
utilized to reduce the amount of copy-and-paste between
commands throughout the script. Structure has been updated to
utilize functions more effectively along with a `main` function.

### DIFF
--- a/docs/en/compute/AWS-EMR.md
+++ b/docs/en/compute/AWS-EMR.md
@@ -52,13 +52,15 @@ upload the script to your S3 bucket.
     - (Optional) A path path to a remote `alluxio-site.properties` that is used during the bootstrap process.
       Properties in this file will be set on all master+worker nodes.
       Can be set to empty string (`""`) to ignore.
-    - (Optional) A delimiter value for additional properties.
-      Empty string (`""`) defaults to `;`.
     - (Optional) Extra alluxio options.
       These are specified as a delimited list of key-values in the
       format `<key>=<value>`.
-      The value from the previous argument is used as the delimiter.
+      The value for the delimiter is `;` by default.
+      It can be changed by setting the 5th argument.
       For example, `alluxio.user.file.writetype.default=CACHE_THROUGH`
+    - (Optional) A delimiter value for additional properties.
+      Empty string (`""`) defaults to `;`.
+      This property can be left out.
 ```console
 $ aws emr create-cluster \
 --release-label emr-5.23.0 \
@@ -68,7 +70,7 @@ $ aws emr create-cluster \
 --name '<cluster-name>' \
 --bootstrap-actions \
 Path=s3://bucket/path/to/alluxio-emr.sh,\
-Args=[<download-url>,<root-ufs-uri>,<path<property-delimiter>,<additional-properties>] \
+Args=[<download-url>,<root-ufs-uri>,<path>,<path-to-site-properties>,<additional-delimited-properties>,<property-delimiter>] \
 --configurations file://${ALLUXIO_HOME}/integration/emr/alluxio-emr.json \
 --ec2-attributes KeyName=<ec2-keypair-name>
 ```
@@ -104,7 +106,6 @@ Path=s3://alluxio-test/emr/bootstrap-actions/alluxio-emr.sh,\
 Args=[http://downloads.alluxio.io/downloads/files/{{site.ALLUXIO_RELEASED_VERSION}}/alluxio-{{site.ALLUXIO_RELEASED_VERSION}}-bin.tar.gz,\
 s3://alluxio-test/emr/mount/,\
 s3://alluxio-test/alluxio/conf/alluxio-site.properties,\
-"|",\
 alluxio.underfs.s3.owner.id.to.username.mapping=f1234123412341234123412341234123412341234123412341234123412341234=hadoop] \
 --configurations file://${ALLUXIO_HOME}/integration/emr/alluxio-emr.json \
 --ec2-attributes KeyName=admin-key
--- a/docs/en/compute/AWS-EMR.md
+++ b/docs/en/compute/AWS-EMR.md
@@ -70,7 +70,7 @@ $ aws emr create-cluster \
 --name '<cluster-name>' \
 --bootstrap-actions \
 Path=s3://bucket/path/to/alluxio-emr.sh,\
-Args=[<download-url>,<root-ufs-uri>,<path>,<path-to-site-properties>,<additional-delimited-properties>,<property-delimiter>] \
+Args=[<download-url>,<root-ufs-uri>,<optional-path-to-site-properties>,<optional-additional-delimited-properties>,<optional-property-delimiter>] \
 --configurations file://${ALLUXIO_HOME}/integration/emr/alluxio-emr.json \
 --ec2-attributes KeyName=<ec2-keypair-name>
 ```
--- a/docs/en/compute/AWS-EMR.md
+++ b/docs/en/compute/AWS-EMR.md
@@ -17,7 +17,7 @@ AWS EMR provides great options for running clusters on-demand to handle compute 
 deployment of various Hadoop Services and allows for hooks into these services for customizations. Alluxio
 can run on EMR to provide functionality above what EMRFS currently provides. Aside from the added performance
 benefits of caching, Alluxio also enables users to run compute workloads against on-premise storage or even a
-different cloud provider's storage i.e. GCS, Azure Blob Store. 
+different cloud provider's storage i.e. GCS, Azure Blob Store.
 
 ## Prerequisites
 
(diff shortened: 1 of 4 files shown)


integration/emr/alluxio-emr.sh
```diff
@@ -10,99 +10,187 @@
...
+  sudo tar -xvf /opt/${release} -C /opt/
+  sudo rm -R /opt/${release}
+  sudo mv /opt/${release_unzip} ${ALLUXIO_HOME}
+  sudo chown -R alluxio:alluxio ${ALLUXIO_HOME}
+  rm ${release}
+  doas alluxio "cp ${ALLUXIO_SITE_PROPERTIES}.template ${ALLUXIO_SITE_PROPERTIES}"
+
+  #Get hostnames and load into masters/workers file
+  local emr_cluster=`jq '.jobFlowId' /mnt/var/lib/info/job-flow.json | sed -e 's/^"//' -e 's/"$//'`
+  local hostlist=`aws emr list-instances --cluster-id ${emr_cluster} --region us-east-1 | jq '.Instances[].PrivateDnsName' | sed -e 's/^"//' -e 's/"$//'`
```
ZacBlanco: Will eventually need to figure out a way to programmatically determine region. Current it is hardcoded to `us-east-1`

ZacBlanco: Fixed in later commit

ZacBlanco:
@ns1123 @apc999 FYI

ZacBlanco:
Also, for testing I created a number of clusters with this PR and everything worked fine with these modifications

docs/en/compute/AWS-EMR.md
```diff
@@ -28,59 +28,72 @@ different cloud provider's storage i.e. GCS, Azure Blob Store.
...
+see the cluster going through the different stages of setup.
+Once the cluster is in the 'Waiting' stage, click on the cluster details to get the
+'Master public DNS'.
+SSH into this instance using the keypair provided in the previous command.
+If a security group isn't specified via CLI, the default EMR security group will not allow inbound
+SSH.
+To SSH into the machine, a new rule will need to be added.
 5. Test that Alluxio is running as expected
-```
-sudo runuser -l alluxio -c "/opt/alluxio/bin/alluxio runTests
```
ns1123: Seems like the close quote was lost somewhere?

integration/emr/alluxio-emr.sh
```diff
@@ -10,99 +10,217 @@
...
+#   $1: property name
+#   $2: property value
+append_alluxio_property() {
+  local property=$1
+  local value=$2
+
+  # OK to fail in this section
+  set +o errexit
+  # /opt/alluxio/conf must exist for this to work
+  grep -q ${property} ${ALLUXIO_SITE_PROPERTIES} 2> /dev/null
```
ns1123: What happens if a property is a substring of an existing property? Can we do an exact string match? With some of the templates, this can get a little dicey.

Ex:
alluxio.master.mount.table.root.option.alluxio.underfs.version=2.7
alluxio.underfs.version=3.1

integration/emr/alluxio-emr.sh
```diff
@@ -10,99 +10,217 @@
...
+  local property_delimeter
+  local delimited_properties
+  alluxio_tarball=${1}
+  root_ufs_uri=${2}
+  site_properties_uri=${3:-""}
+  property_delimeter=${4:-";"}
+  delimited_properties=${5:-""}
+
+  # Create user
+  sudo groupadd alluxio -g 600 || true
```
ns1123: Do we want this to have the || true appended? If the user isn't successfully created, most following commands will fail

integration/emr/alluxio-emr.sh
ZacBlanco: Ooh, good catch. I think the quick fix for this is for grep to use `${property}=`

ZacBlanco: I can also use regex to force matching from beginning of the line

    grep -q -e "^${property}="

integration/emr/alluxio-emr.sh
ZacBlanco: This was some debugging code. I'll remove. good catch.

integration/emr/alluxio-emr.sh
ns1123: I think our current file allows whitespace at the beginning of the line. So if we expect users to maybe try out a local mode alluxio-site.properties in EMR, then we should have consistent behavior.

docs/en/compute/AWS-EMR.md
```diff
@@ -28,59 +28,72 @@ different cloud provider's storage i.e. GCS, Azure Blob Store.
...
+```console
+$ sudo runuser -l alluxio -c "/opt/alluxio/bin/alluxio runTests
```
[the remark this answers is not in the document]
ZacBlanco: good catch!

docs/en/compute/AWS-EMR.md
```diff
@@ -90,30 +103,33 @@ aws emr create-cluster \
...
+"|",\
 alluxio.underfs.s3.owner.id.to.username.mapping=f1234123412341234123412341234123412341234123412341234123412341234=hadoop] \
 --configurations file://${ALLUXIO_HOME}/integration/emr/alluxio-emr.json \
 --ec2-attributes KeyName=admin-key
 ```
 
-Notes: The default Alluxio Worker memory is set to 20GB. If the instance type has less than 20GB of memory, change
-the value in the `alluxio-emr.sh` script.
+> Note: The default Alluxio Worker memory is set to 1/3 of the physical memory on the instance.
+If specific value is desired, set `alluxio.worker.memory.size` in the provided
```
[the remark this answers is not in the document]
ZacBlanco: :+1:

integration/emr/alluxio-emr.sh
ZacBlanco: Regex for allowing the beginning whitespace is `^\s*${property}=`

ns1123:
LGTM. Please run a sanity test with the modifications.

ZacBlanco:
@ns1123 FYI ran with some clusters with the new changes and everything is looking good

apc999:
@ZacBlanco  is the change to `docs/css/main.css` in this PR intended?

ZacBlanco:
It was initially, but I'll remove it as it probably shouldn't be in this PR

docs/en/compute/AWS-EMR.md
```diff
@@ -28,59 +28,72 @@ different cloud provider's storage i.e. GCS, Azure Blob Store.
...
-       This is a mandatory property.
-    - Extra alluxio options. These are specified as a comma-separated list of key-values in the format `<key>=<value>`.
-       For example, `alluxio.user.file.writetype.default=CACHE_THROUGH`
-```
-aws emr create-cluster \
+3. There are 5 input arguments for the bootstrap script. They are described below:
+    - (Mandatory) The URI from where to download the Alluxio Release `.tar.gz` file.
+    This can be an `https://` URI, `s3://` URI or an `http://` URI.
+    - (Mandatory) The root-ufs-uri.
+      This should be an `s3://` or `hdfs://` URI designating the root mount of the Alluxio file
```
apc999: why is this required to be s3 or hdfs? @ns1123

ZacBlanco:
jenkins, test this please

docs/en/compute/AWS-EMR.md
```diff
@@ -28,59 +28,74 @@ different cloud provider's storage i.e. GCS, Azure Blob Store.
...
+$ aws emr create-cluster \
 --release-label emr-5.23.0 \
 --instance-count <num-instances> \
 --instance-type <instance-type> \
 --applications Name=Presto Name=Hive Name=Spark \
 --name '<cluster-name>' \
 --bootstrap-actions \
 Path=s3://bucket/path/to/alluxio-emr.sh,\
-Args=[<download-url>,<root-ufs-uri>,<additional-properties>] \
+Args=[<download-url>,<root-ufs-uri>,<path>,<path-to-site-properties>,<additional-delimited-properties>,<property-delimiter>] \
```
apc999: - remove `<path>`? You are taking 6 args where 5 are expected
- How about `<optional-path-to-site-properties>`, `<optional-additional-properties>`,`<optional-property-delimiter>`

integration/emr/alluxio-emr.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
apc999: is `#!/usr/bin/env bash` more portable?

integration/emr/alluxio-emr.sh
```diff
@@ -10,99 +10,216 @@
...
+
+  # Create user
+  sudo groupadd alluxio -g 600
+  sudo useradd alluxio -u 600 -g 600
+
+  # Download the release
+  if [[ -z ${alluxio_tarball} ]]
   then
-    aws s3 cp ${1} ./
+    echo "No Download URL Provided. Please go to http://downloads.alluxio.io to see available release downloads."
```
apc999: exit here?

docs/en/compute/AWS-EMR.md
apc999: Ping @ns1123  this is the Alluxio UFS URI and can be any UFS alluxio supports, right?

integration/emr/alluxio-emr.sh
ZacBlanco: yes

integration/emr/alluxio-emr.sh
ZacBlanco: Don't need to. The script specifies at the top 

```bash
set -o errexit  # exit when a command fails - append "|| true" to allow a
                # command to fail
```

apc999:
@ZacBlanco @ns1123 do we need to cherry-pick this to 2.0 branch?

apc999:

PR #9507, opened 2019-07-19
Enable owner to chmod, even without write privilege

bf8086:
Cherry-pick https://github.com/Alluxio/alluxio/pull/9456 to `branch-1.8`

Currently owner of file/directory cannot change permission when they
don't have write privilege. This is inconsistent with Unix and HDFS.
This fix updated the permission requirement so the user should be able
to perform chmod if and only if the user is the owner or super user.

pr-link: Alluxio/alluxio#9456
change-id: cid-1de7dc5511daf2b91761db8014560f07b0bfd746


bf8086:
@gpang PTAL.

gpang:

PR #9510, opened 2019-07-19
Fix UI issue with unused var

tieujason330:
- `lastFetched` was used to visually show which link was previously clicked on. Didn't seem to work since when we switch from file view to directory view, we always send out a new fetch request, so `lastFetched` keeps getting cleared
- it's best to just remove it since it has no real usage

### DIFF
--- a/webui/common/src/utilities/misc/renderFileNameLink.tsx
+++ b/webui/common/src/utilities/misc/renderFileNameLink.tsx
@@ -1,13 +1,7 @@
 import {Link} from 'react-router-dom';
 import React from 'react';
 
-export const renderFileNameLink = function (this: any, path: string, urlPrefix: string) {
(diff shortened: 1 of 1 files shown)


calvinjia:
@tieujason330 Since we changed the signature do we need to update the call sites?

tieujason330:
@calvinjia not in this case since we're calling the method via [`.call()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call). We don't need to explicitly state `this` as an argument in the signature. It'll be called in the scope of the caller (so using the caller's `this`). Plus we no longer care about the scope since we'll return an object based on the passed in arguments. One of javascript's many caveats I guess

calvinjia:

PR #9500, opened 2019-07-19
Fix Unmount failure by Alluxio FUSE

liuhongtong:
Fix #9461

If umount an Alluxio FUSE mount point,
fs.mount() returns without any exception.
but fs.umount() will conflict with the running umount command,
that results in "target is busy" error and umount command failing.

Signed-off-by: liuhongtong <hongtongliu@126.com>

PR #9516, opened 2019-07-19
Remove close because there is no close API in 1.8

calvinjia:
Changed introduced from #9500


apc999:

PR #9529, opened 2019-07-22
Change journal tool test to use must cache instead of async

calvinjia:
When using async through there can be timing related issues with the persist step


ggezer:
Does this help with flakiness? If so, how?

calvinjia:
Jenkins, test this please.

calvinjia:
@ggezer I think it should help with the JournalToolTest. When that test fails it is usually do to a timing issue caused by the large number of async persists going on in the system.

calvinjia:

PR #9531, opened 2019-07-22
Add to be persisted files to pinned file list

yuzhu:
Changes the semantic of `getPinIdList` to return both the explicitly
pinned inodes and inodes which are in `TO_BE_PERSISTED` state. Inodes in
that state should not be evicted or users can experience data loss.

pr-link: Alluxio/alluxio#9052
change-id: cid-3d03eca419de07c8efc4f238d095b8c58c2b2c41


yuzhu:
@calvinjia PTAL

calvinjia:
@yuzhu you will need to make some logical conflict resolutions (ie. libraries not imported)

PR #9533, opened 2019-07-22
Implement quorum management and shell commands

ggezer:


ggezer:
@LuQQiu PTAL. Thanks!

LuQQiu:
Thanks for adding the quorum management! Two high level comments are left:

(1) Could we put `getQuorumInfo` and `removeQuorumServer` in  `MetaMaster` instead of `JournalMaster`? 
The names of embedded journal and UFS journal is more about the Journal (which determines the ways of storing journal logs). Embedded journal comes with Copycat quorum which helps us to do the internal leader election. For the leader election part, we all use `internal leader election` to refer to it in our official docs instead of embedded journal. Besides that, UFS journal has nothing to do with leader election or quorum. `MetaMaster` on the other hand contains methods about masters. @gpang What's your opinion about where we put the quorum methods in?

(2) do we want to use the word `quorum` externally?
It makes sense if we only want experience users to use the `fsadmin quorum` to get internal leader election information or remove a master node from the internal leader election. For normal users, `quorum` may make them confuse. I will prefer `fsadmin leaderElection` or something easier for users to understand. @calvinjia Any ideas of what term should we expose to users?

ggezer:
@LuQQiu 

This PR hasn't introduced any distinction between embedded journal and UFS. Distinction is between MASTER and JOB_MASTER. So I'm not sure why you mentioned that.

`MetaMaster` only lives in alluxio master process. Because of that it can't be used to manage job_master journal. This is the reason I have a dedicated `JournalMaster` service that is served by both alluxio and job_master processes.

Existing `checkpoint`command lives in `MetaMaster` and can't work with job_master journal because of this. Please see https://github.com/Alluxio/alluxio/issues/9538. Once this PR is in, we can move `checkpoint` command to new `JournalMaster` for overcoming this limitation.

I agree `quorum` command might be confusing when it's detached from the context of journals. Once we're done with this and the above issue, we can unify both commands under high-level `journal` command, under which both `checkpoint` and `quorum` would make better sense. Please see https://github.com/Alluxio/alluxio/issues/9542.

LuQQiu:
Got it, if the `JournalMaster` can work with both master and job_master, include more other journal related methods (like checkpoint), then it makes sense to have this separate master.

For CLI, I agree to uniform the different journal related commands and improve user experience.
Thanks!

ggezer:
@calvinjia PTAL. Thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -509,6 +516,50 @@ public synchronized void stopInternal() throws InterruptedException, IOException
     LOG.info("Journal shutdown complete");
   }
 
+  /**
+   * Used to export information of internal RAFT quorum.
```
LuQQiu: `Used to export` -> `Gets`

job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
```diff
@@ -203,6 +206,10 @@ protected void startServingRPCServer() {
           ServerUserState.global());
       registerServices(serverBuilder, mJobMaster.getServices());
 
+      // Add journal master service.
+      serverBuilder.addService(alluxio.grpc.ServiceType.JOURNAL_MASTER_CLIENT_SERVICE,
```
LuQQiu: Will it better to put in `JobMaster.getServices`?

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -362,6 +364,18 @@ public synchronized MetaMasterClient getMetaMasterClient() {
         .build());
   }
 
+  /**
+   * @return a meta master client
```
LuQQiu: `meta` -> `journal`

shell/src/main/java/alluxio/cli/fsadmin/command/QuorumCommand.java
```diff
@@ -0,0 +1,88 @@
...
+      mSubCommands.put(name, constructor.apply(context, alluxioConf));
+    });
+  }
+
+  /**
+   * @return command's description
+   */
+  @VisibleForTesting
+  public static String description() {
+    return "Manage embedded journal quorum configuration. "
```
LuQQiu: Change to `Manage internal leader election quorum`?

shell/src/main/java/alluxio/cli/fsadmin/quorum/QuorumInfoCommand.java
```diff
@@ -0,0 +1,124 @@
...
+  public QuorumInfoCommand(Context context, AlluxioConfiguration alluxioConf) {
+    super(context);
+  }
+
+  /**
+   * @return command's description
+   */
+  @VisibleForTesting
+  public static String description() {
+    return "Shows quorum information for embedded journal.";
```
LuQQiu: Change to `Shows internal leader election quorum information`?

shell/src/main/java/alluxio/cli/fsadmin/quorum/QuorumInfoCommand.java
```diff
@@ -0,0 +1,124 @@
...
+    try {
+      JournalDomain.valueOf(domainVal);
+    } catch (IllegalArgumentException e) {
+      throw new InvalidArgumentException(ExceptionMessage.INVALID_OPTION_VALUE
+          .getMessage(DOMAIN_OPTION_NAME, Arrays.toString(JournalDomain.values())));
+    }
+
+    JournalMasterClient jmClient = mMasterJournalMasterClient;
+
+    if (cl.hasOption(DOMAIN_OPTION_NAME)) {
```
LuQQiu: Is it necessary? Can we directly use the previous try clause to get the `domain`.

shell/src/main/java/alluxio/cli/fsadmin/quorum/QuorumInfoCommand.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public String getDescription() {
+    return description();
+  }
+
+  @Override
+  public void validateArgs(CommandLine cl) throws InvalidArgumentException {
+    if (cl.getOptions().length != 1) {
```
LuQQiu: Could we make `master` domain the default? Usually users are not that interest in job_master domain compared to master.

job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
ggezer: Theoretically, It could have been put to any master. However it's very confusing as non of them owns the journalsystem. Only master processes owns them so that's why I'm doing the insertion at that level for both alluxio and job_master processes.

core/server/common/src/main/java/alluxio/master/journal/DefaultJournalMaster.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Default implementation for {@link JournalMaster}.
+ * This is used by both master and job_master processes.
+ */
+public class DefaultJournalMaster extends
+    JournalMasterClientServiceGrpc.JournalMasterClientServiceImplBase implements JournalMaster {
```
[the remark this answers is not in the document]
ggezer: Agreed. It was confusing to me as well.

shell/src/main/java/alluxio/cli/fsadmin/command/QuorumCommand.java
ggezer: I think that would be too explicit.

shell/src/main/java/alluxio/cli/fsadmin/quorum/QuorumInfoCommand.java
ggezer: I think that would be too explicit.

shell/src/main/java/alluxio/cli/fsadmin/quorum/QuorumInfoCommand.java
ggezer: I thought about that but couldn't find a nice way to plug optional arguments into our command subsystem and didn't want to complicate the code. I feel like it's lay out arguments explicitly for this particular command as it's for advanced use. I don't have a strong preference though.

core/base/src/main/java/alluxio/Constants.java
```diff
@@ -91,6 +91,7 @@
   public static final long META_MASTER_CLIENT_SERVICE_VERSION = 2;
   public static final long META_MASTER_MASTER_SERVICE_VERSION = 1;
   public static final long METRICS_MASTER_CLIENT_SERVICE_VERSION = 2;
+  public static final long JOURNAL_MASTER_CLIENT_SERVICE_VERSION = 2;
```
calvinjia: Why start this at 2?

core/base/src/main/java/alluxio/exception/ExceptionMessage.java
```diff
@@ -190,6 +190,10 @@
   INVALID_ARGS_NUM_TOO_MANY("Command {0} requires at most {1} arguments ({2} provided)"),
   INVALID_ARGS_SORT_FIELD("Invalid sort option `{0}` for --sort"),
   INVALID_ARG_TYPE("Arg {0} is not type {1}"),
+  INVALID_OPTION_COUNT("Invalid option count. Expected: {0}, Found: {1}"),
+  INVALID_OPTION("Invalid option provided. Supported options: {0}"),
```
calvinjia: Reorder

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -509,6 +516,50 @@ public synchronized void stopInternal() throws InterruptedException, IOException
...
+
+      quorumMemberStateList.add(QuorumServerInfo.newBuilder().setServerAddress(memberAddress)
+          .setServerState(QuorumServerState.valueOf(member.status().name())).build());
+    }
+    return quorumMemberStateList;
+  }
+
+  /**
+   * Removes from RAFT quorum, a server with given address.
+   * For server to be removed, it should be in unavailable state in quorum.
```
calvinjia: Is this explicitly checked?

ggezer: That is validated by Copycat.

ggezer:
@calvinjia `domain` argument is used by both commands to select the correct `JournalMasterClient`.

calvinjia:
@ggezer Right, but the implementation of both `remove` calls is unrelated to the domain?

ggezer:
@calvinjia, Correct. But we still need an argument at command level for forwarding the call to appropriate implementation.

calvinjia:
An alternative could be to show quorum info for both through the status call?

job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
LuQQiu: Make sense

calvinjia:

PR #9536, opened 2019-07-22
Fix UI issue with unused var

tieujason330:
- `lastFetched` was used to visually show which link was previously
clicked on. Didn't seem to work since when we switch from file view to
directory view, we always send out a new fetch request, so `lastFetched`
keeps getting cleared
- it's best to just remove it since it has no real usage

pr-link: Alluxio/alluxio#9510
change-id: cid-f2f749ede019f89908da3310ea9598932c835708


tieujason330:
@calvinjia PTAL, thanks

calvinjia:

PR #9537, opened 2019-07-22
Reduce the information allocator need in createBlockMeta

yuzhu:


yuzhu:
@calvinjia PTAL

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
```diff
@@ -43,16 +43,8 @@ private Factory() {} // prevent instantiation
...
       return CommonUtils.createNewClassInstance(
-<<<<<<< HEAD
-          Configuration.<Allocator>getClass(PropertyKey.WORKER_ALLOCATOR_CLASS),
-          new Class[] {BlockMetadataManagerView.class}, new Object[] {managerView});
-||||||| parent of ec9f9ceb90... Reduce the information allocator need in createBlockMeta
-          ServerConfiguration.<Allocator>getClass(PropertyKey.WORKER_ALLOCATOR_CLASS),
-          new Class[] {BlockMetadataManagerView.class}, new Object[] {managerView});
-=======
-          ServerConfiguration.<Allocator>getClass(PropertyKey.WORKER_ALLOCATOR_CLASS),
+          Configuration.getClass(PropertyKey.WORKER_ALLOCATOR_CLASS),
```
calvinjia: Removing the `<Allocator>` type?

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDirView.java
```diff
@@ -90,96 +90,6 @@ public StorageTierView getParentTierView() {
   }
 
   /**
-<<<<<<< HEAD
-   * Returns an indication whether the given block is marked to be moved out.
-   *
-   * @param blockId the block ID
-   * @return whether the block is marked to be moved out
-   */
-  public boolean isMarkedToMoveOut(long blockId) {
```
calvinjia: Don't we need these methods still?

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
yuzhu: IDE prompted me to remove it, since it can be inferred. Do we have a convention to write it out?

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDirView.java
yuzhu: they are moved to another file

PR #9540, opened 2019-07-22
Fix 'Display from byte offset' input

### LINKED ISSUE #9483: Fix display from byte offset field in the web ui

tieujason330:
- Fixes #9483
- making sure we initialize the field with `0` offset even if parsed
query has `undefined` offset

pr-link: Alluxio/alluxio#9498
change-id: cid-fa8efbc9fe26f3efcdee75eab725c05c4f2cdf6b


calvinjia:

PR #9547, opened 2019-07-23
Add logic to track last access time

bf8086:
Access time is useful for understanding the access pattern for files and directories. This PR implemented the core logic for updating and persisting the last access time. Last access is updated on open, create, complete for files, and list, child delete, child rename, child create for directories. When updating, Alluxio master uses the following strategy:
- When the action updates inode metadata, the last access time updates and journal synchronously.
- When the action does not update inode metadata, the current operation time is compared to existing last access time. if the difference is bigger than `MASTER_FILE_ACCESS_TIME_UPDATE_PRECISION`, then the last access time of the inode is set to the operation time, and the operation will be journaled at some point of time. Otherwise the last access time is not updated.
- For journaling the access time update, `MASTER_FILE_ACCESS_TIME_JOURNAL_FLUSH_INTERVAL` determines the timing of writing the update to journal. If the interval is non-positive, then the journal is updated synchronously. Otherwise Alluxio master will schedule the journal update to happen no later than the interval. The first update will schedule a one time task to write to journal after the interval, and any update request before the scheduled task is executed will be performed in batch when the task is executed.

Most relevant code are in:

- AccessTimeUpdater.java: contains configurable update logic which allows updating with certain precision and delay.
- DefaultFileSystemMaster.java: contains entry points for updating last access time on various read and write operations.


bf8086:
@gpang PTAL.

gpang:
Jenkins, test this please

gpang:
@bf8086 in the description, can you add some information on how the last accessed time is updated/journaled, and some of the implications of it? Also, does last access also include updates, or it just for reads?

bf8086:
@gpang Updated.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -354,7 +354,7 @@ public FileStatus getFileStatus(Path path) throws IOException {
     return new FileStatus(fileStatus.getLength(), fileStatus.isFolder(),
         getReplica(fileStatus), fileStatus.getBlockSizeBytes(),
         fileStatus.getLastModificationTimeMs(),
-        fileStatus.getCreationTimeMs(), new FsPermission((short) fileStatus.getMode()),
+        fileStatus.getLastAccessTimeMs(), new FsPermission((short) fileStatus.getMode()),
```
gpang: why is creation time changed to last access time?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -1117,6 +1129,7 @@ public void syncPersistNewDirectory(MutableInodeDirectory dir)
       Long lastModificationTime = status.getLastModifiedTime();
       if (lastModificationTime != null) {
         dir.setLastModificationTimeMs(lastModificationTime, true);
+        dir.setLastAccessTimeMs(lastModificationTime, true);
```
gpang: why does this depend on `lastModificationTime != null`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1243,6 +1259,7 @@ private void completeFileInternal(RpcContext rpcContext, LockedInodePath inodePa
         .setId(inode.getId())
         .setUfsFingerprint(ufsFingerprint)
         .setLastModificationTimeMs(opTimeMs)
+        .setLastAccessTimeMs(opTimeMs)
```
gpang: this is for create, but where do other accesses like mv get updated? Also, when you delete an inode, does the parent get updated to? I don't see that one either.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,176 @@
...
+          journalAccessTime(context, entry);
+        }
+      }
+    }
+  }
+
+  private void scheduleJournalUpdate(UpdateInodeEntry entry) {
+    mAccessTimeUpdates.add(entry.getId());
+    if (mUpdateScheduled.compareAndSet(false, true)) {
+      mExecutorService.schedule(this::flushScheduledUpdates, mFlushInterval, TimeUnit.MILLISECONDS);
```
gpang: can you use this instead? https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit)

gpang: can you use this instead? https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit)

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Update inode last access time.
+   *
+   * @param context the journal context
+   * @param inode the target inode
+   * @param opTimeMs the last access time to set on the inode
+   */
+  public void updateAccessTime(JournalContext context, Inode inode, long opTimeMs) {
```
gpang: is this method run while already holding the lock on the inode, in FSM?

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  public void start() {
+    start(mFlushInterval > 0 ? Executors.newSingleThreadScheduledExecutor(
+        ThreadFactoryUtils.build("AccessTimeUpdater-%d", true)) : null);
+  }
+
+  @VisibleForTesting
+  public void start(ScheduledExecutorService executorService) {
+    mExecutorService = executorService;
```
gpang: what if `mExecutorService` was already  not `null` from a previous `start`?

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  private void flushUpdates() {
+    long filesRemoved = 0;
+    try (JournalContext context = mFileSystemMaster.createJournalContext()) {
+      for (Iterator<Long> iterator = mAccessTimeUpdates.iterator();
+           iterator.hasNext();) {
+        long inodeId = iterator.next();
+        iterator.remove();
+        try (LockedInodePath path = mInodeTree.lockFullInodePath(inodeId, LockPattern.READ);
```
gpang: So, this will re-lock all the inodes? Will this be too expensive? What are the tradeoffs of just journaling without locking?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -762,6 +768,12 @@ public FileInfo getFileInfo(AlluxioURI path, GetStatusContext context)
         ensureFullPathAndUpdateCache(inodePath);
       }
       FileInfo fileInfo = getFileInfoInternal(inodePath);
+      Mode.Bits accessMode = Mode.Bits.fromProto(context.getOptions().getAccessMode());
+      if (context.getOptions().hasAccessMode()
+           && (accessMode.imply(Mode.Bits.READ) || accessMode.imply(Mode.Bits.WRITE))) {
```
gpang: why can't we just update the access time for any access (remove this check)?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: where do other updates, like mv, or delete, accesses get updated?

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: A previous `mExecutorService` should have been already shutdown by a previous `stop()`.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
bf8086: From HDFS interface this argument is the access time. We just used creation time as a fallback.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -189,7 +189,11 @@ public synchronized JournalContext createJournalContext() throws UnavailableExce
       // We throw UnavailableException here so that clients will retry with the next primary master.
       throw new UnavailableException("Failed to write to journal: journal is in state " + mState);
     }
-    return new MasterJournalContext(mAsyncWriter);
+    AsyncJournalWriter writer = mAsyncWriter;
+    if (writer == null) {
```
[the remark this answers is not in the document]
bf8086: This happens if any thread tries to write to a journal after it has been shutdown(e.g. due to loss of primacy). It should throw an `UnavailableException` per interface instead of a NPE.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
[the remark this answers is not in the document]
bf8086: We want to limit access time update to only apply to file content access. 

From user perspective, it is more useful for knowing when user actually opened the file and read data, rather than just someone listed or changed attributes, especially because the policy engine is focused on moving the actual data than the metadata.

From technical perspective, updating inode tree and journal on every `getStatus` call is just too expensive, even with all the optimization in the update logic. It effectively makes every metadata read a write operation.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: Yes. I will add a javadoc comment to note that.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: `scheduleWithFixedDelay` schedules a task to run periodically. Here the task should be run lazily if:
- at least one inode is pending update
- an update task is not already scheduled

For example, if the update interval is 1 minute and two inodes are read 1 hour apart, `scheduleWithFixedDelay` will make the update task run 60 times while this function will only schedule the update to run twice.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: The main risk is that journal entry might be written after an inode is removed, causing the journal replay to fail later on.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: rename and delete will only update the timestamps of the immediate parent directory. It is handled via `InodeTreePersistentState::updateTimestampsAndChildCount()`

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
bf8086: Currently UFS last access time is not exposed or sync'ed.  This is just to give last access time a default value.

core/server/common/src/main/java/alluxio/master/journal/sink/JournalSink.java
```diff
@@ -23,10 +23,16 @@
...
   /**
    * Signals the sink that the journal is flushed.
    */
-  void flush();
+  default void flush() {}
+
+  /**
+   * Signals the sink that the journal is about to shutdown.
+   */
+  default void beforeShutdown() {
```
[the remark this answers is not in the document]
bf8086: Updated.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.Iterator;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * This class represents the executor for periodic inode ttl check.
```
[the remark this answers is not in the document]
bf8086: Oops, updated.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -896,14 +908,15 @@ public PersistenceState getPersistenceState(long fileId) throws FileDoesNotExist
    * the descendantType. The result is returned via a list specified by statusList, in postorder
    * traversal order.
    *
+   * @param rpcContext  the context for the RPC call
```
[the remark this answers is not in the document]
bf8086: Removed.

bf8086:
Thanks @gpang! I updated the PR to address your comments. PTAL.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
gpang: Is this because we are re-using the `UpdateInode` journal entry? Would it make sense to have a separate `UpdateAccessTime` entry?

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * Update inode last access time. Requires at least read lock acquired for the inode.
+   *
+   * @param context the journal context
+   * @param inode the target inode
+   * @param opTimeMs the last access time to set on the inode
+   */
+  public void updateAccessTime(JournalContext context, Inode inode, long opTimeMs) {
+    if (opTimeMs - inode.getLastAccessTimeMs() > mUpdatePrecision) {
+      try (LockResource lr = mInodeTree.getInodeLockManager().lockUpdate(inode.getId())) {
```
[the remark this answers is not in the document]
bf8086: This is to acquire update lock which is only for updating a timestamp. This needs to be done in addition to the expected read lock.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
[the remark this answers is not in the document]
bf8086: Yes it is consistent with HDFS. HDFS update access time on acquiring block locations, which is only called when a client reads a file. getFileInfo is called by client for many metadata only operations such as checking existence or running stat command. The access mode is an optional parameter for the `getFileInfo()` RPC. It is currently only set by the client when used inside the client file `open` API.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: Rename on file system master eventually calls `InodeTreePersistentState::applyRename()` which calls `InodeTreePersistentState::updateTimestampsAndChildCount()` to update parent timestamps. It is considered a "content update" of parent directory thus should update both last modification time and last access time. This behavior is consistent with UNIX file system, while HDFS does not implement access time for directories.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -292,6 +292,15 @@ public void updateInode(Supplier<JournalContext> context, UpdateInodeEntry entry
     mState.applyAndJournal(context, entry);
   }
 
+  /**
+   * @param inodeId id of inode to be updated
+   * @param accessTime the new access time
+   * @return the applied inode entry
+   */
+  public UpdateInodeEntry updateInodeAccessTime(long inodeId, long accessTime) {
+    return mState.applyInodeAccessTime(inodeId, accessTime);
```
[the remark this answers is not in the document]
bf8086: Current implementation can result in inconsistency in access time if master is interrupted before it has a chance to flush access time changes. The only way to guarantee journal updates is to turn of async update for access time. 

Regarding API design, I made this call specific to access time so user of the class cannot use it to introduce inconsistent journal for updating other more important attributes. To avoid ppl forget to add journal entry, the only way I can think of is to make the access time updater part of the inode tree class so the journal update will be managed exclusively within the API call. However I am concerned about adding attribute specific logic to InodeTree to make it overly complicated. What do you think?

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: It looks like we recently deprecated a similar journal entry `UpdateModificationTimeEntry` to unify journaling. I can just ignore missing inode error when applying entries that only update access time.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: `AccessTimeUpdater.stop()` is called on  `FileSystemMaster.stop()`. If that is not called before another start, the master will certainly be in a bad state. But just to be defensive, I will update it to try stop before starting a new executor, if needed.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
bf8086: Added.

bf8086:
Thanks @gpang! I updated the PR to address your comments. PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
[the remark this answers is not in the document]
gpang: If this is going to be the non-journaling update, then should call this something like  `updateInodeAccessTimeNoJournal`? This a special case since this method will not journal like all the other methods.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -419,6 +438,10 @@ private void applyUpdateInode(UpdateInodeEntry entry) {
     updateToBePersistedIds(inode);
   }
 
+  private boolean isJournalUpdateAsync(UpdateInodeEntry entry) {
```
gpang: NIT: comment on what this does/returns.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,166 @@
...
+        UpdateInodeEntry entry = mInodeTree.updateInodeAccessTime(inode.getId(), opTimeMs);
```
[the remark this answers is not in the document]
bf8086: updated.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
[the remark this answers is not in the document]
bf8086: renamed.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
bf8086: Added a javadoc for the method.

bf8086:
Jenkins, test this please.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
gpang: do we have to update any documentation now, saying this is now last accessed time?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1278,6 +1278,30 @@ public String toString() {
               + "used when Zookeeper is enabled, since Zookeeper already stores the master "
               + "addresses.")
           .build();
+  public static final PropertyKey MASTER_FILE_ACCESS_TIME_JOURNAL_FLUSH_INTERVAL =
+      new Builder(Name.MASTER_FILE_ACCESS_TIME_JOURNAL_FLUSH_INTERVAL)
+          .setDefaultValue("1h")
+          .setDescription("The minimum interval between files access time update journal entries "
+              + "get flushed. Setting it to a non-positive value will make the the journal update "
+              + "synchronous.")
```
gpang: In the description, can you add the implications of synchronous flushes (always consistent but slows down reads), as well as an async flush interval (does not affect reads, but some accesses may be lost on crashes)?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1278,6 +1278,30 @@ public String toString() {
...
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_FILE_ACCESS_TIME_UPDATE_PRECISION =
+      new Builder(Name.MASTER_FILE_ACCESS_TIME_UPDATE_PRECISION)
+          .setDefaultValue("1d")
+          .setDescription("The file last access time is precise up to this value. Setting it to"
+              + "a non-positive value will update last access time on every file access operation.")
```
gpang: can you add some implications of short vs long precisions?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -566,32 +593,36 @@ private void applyRename(RenameEntry entry) {
...
-      updateLastModifiedAndChildCount(oldParent, entry.getOpTimeMs(), -1);
-      updateLastModifiedAndChildCount(newParent, entry.getOpTimeMs(), 1);
+      updateTimestampsAndChildCount(oldParent, entry.getOpTimeMs(), -1);
+      updateTimestampsAndChildCount(newParent, entry.getOpTimeMs(), 1);
     }
   }
 
   /**
-   * Updates the last modified time (LMT) for the indicated inode directory, and updates its child
+   * Updates the timestamps for the indicated inode directory, and updates its child
```
gpang: can you add what the "timestamps" are? (last modified, and last accessed)

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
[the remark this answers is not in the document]
gpang: How is this possible? We are holding a read lock, and then we are trying to write lock it? Lock upgrades is not possible with java locks.

Also, is it required to lock again? If we only have the read lock, we could have out-of-order access time entries, but I thought that was fine, since we handle that on the replay side?

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
```diff
@@ -0,0 +1,171 @@
...
+
+  private void scheduleJournalUpdate(UpdateInodeEntry entry) {
+    mAccessTimeUpdates.put(entry.getId(), entry.getLastAccessTimeMs());
+    if (mUpdateScheduled.compareAndSet(false, true)) {
+      mExecutorService.schedule(this::flushScheduledUpdates, mFlushInterval, TimeUnit.MILLISECONDS);
+    }
+  }
+
+  private void journalAccessTime(JournalContext context, UpdateInodeEntry entry) {
+    context.append(Journal.JournalEntry.newBuilder().setUpdateInode(entry)
```
gpang: I think we can just inline this at the call site.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
bf8086: Not really. We never mention this is creation time in the first place. Hadoop interface has always refer to this parameter as last access time so we are good. Basically our Hadoop client now implement access time correctly.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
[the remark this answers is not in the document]
bf8086: Looking at the implementation of `lockUpdate()`, it basically locks a different type of lock instead of the inode lock. Here is the explanation from the javadoc:

> Locks for guarding changes to last modified time and size on read-locked parent inodes.
> 
> When renaming, creating, or deleting, we update the last modified time and size of the parent inode while holding only a read lock. In the presence of concurrent operations, this could cause the last modified time to decrease, or lead to incorrect directory sizes. To avoid this, we guard the parent inode read-modify-write with this lock. To avoid deadlock, a thread should never acquire more than one of these locks at the same time, and no other locks should be taken while holding one of these locks.

core/common/src/main/java/alluxio/conf/PropertyKey.java
bf8086: Done.

core/common/src/main/java/alluxio/conf/PropertyKey.java
bf8086: Updated.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
bf8086: Added.

core/server/master/src/main/java/alluxio/master/file/AccessTimeUpdater.java
bf8086: Inlined.

bf8086:
@gpang Thanks! I updated the PR. PTAL.

bf8086:
Jenkins, test this please.

gpang:

PR #9550, opened 2019-07-24
Remove path encoding for UI Browse tab

### LINKED ISSUE #9506: Remove URLEncoding in browse field
**Alluxio Version:**
2.1.0-SNAPSHOT

**Describe the bug**
In the browse GO field, paths will be URL encoded.

**To Reproduce**
Go to the browse tab and navigate to a directory, the new browse textbox will show `%2F` instead of `/`

**Expected behavior**
The path in the input box is not URL encoded.

tieujason330:
- Fixes #9506 
- make sure parsed path comes in decoded or else it'll get into an unreliable state

### DIFF
--- a/webui/master/src/containers/pages/Browse/Browse.tsx
+++ b/webui/master/src/containers/pages/Browse/Browse.tsx
@@ -73,9 +73,9 @@ export class Browse extends React.Component<AllProps, IBrowseState> {
     super(props);
 
     let {path, offset, limit, end} = parseQuerystring(this.props.location.search);
-    path = decodeURIComponent(path || '');
+    path = decodeURIComponent(path || '/');
     offset = offset || '0';
-    this.state = {end, limit, offset, path: path || '/'};
+    this.state = {end, limit, offset, path};
   }
 
   public componentDidUpdate(prevProps: AllProps) {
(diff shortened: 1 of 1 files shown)


gpang:
@tieujason330 can you be a bit more descriptive with the title (avoid using "fix")? The issue title is "Remove URLEncoding in browse field", so could this title mention removing the url encoding, if that is in fact what this is doing? Thanks!

webui/master/src/containers/pages/Browse/Browse.tsx
```diff
@@ -73,6 +73,7 @@ export class Browse extends React.Component<AllProps, IBrowseState> {
     super(props);
 
     let {path, offset, limit, end} = parseQuerystring(this.props.location.search);
+    path = decodeURIComponent(path || '');
```
calvinjia: Should the default value be `/`? 2 lines down we try to set path with `|| '/'`

webui/master/src/containers/pages/Browse/Browse.tsx
```diff
@@ -215,7 +215,7 @@ export class Browse extends React.Component<AllProps, IBrowseState> {
                    `/browse?path=${path}${queryStringSuffix}`)}/>
           </FormGroup>
           <FormGroup className="mb-2 mr-sm-2">
-            <Button tag={Link} to={`/browse?path=${encodeURIComponent(path || "/")}${queryStringSuffix}`}
+            <Button tag={Link} to={`/browse?path=${encodeURIComponent(path || '/')}${queryStringSuffix}`}
```
calvinjia: Unrelated to this PR, is it possible in typescript to infer that path cannot be undefined here?

calvinjia:

webui/master/src/containers/pages/Browse/Browse.tsx
tieujason330: it depends how `path` is defined. In this case, it's defined as an optional string (`path?: string`) so typescript complains about that. I can look into simplifying this if possible

PR #9551, opened 2019-07-24
Remove query parameters from UI browse button

### LINKED ISSUE #9484: Query parameters are propagated incorrectly in the web ui
**Alluxio Version:**
2.0.0

**Describe the bug**
When pressing the `Go` button on the browse page, it keeps all query parameters including offset and limit. This causes problems when browsing a large directory and going to a smaller directory.

**To Reproduce**
Create a folder with > 20 files (to activate pagination).
Go to the second page.
Press Go to return to the root (< 20 files)

**Expected behavior**
User is brought to the appropriate first paginated page of the destination.

tieujason330:
- Fixes #9484 
- Removed `queryStringSuffix` for directory rendering since we should only base it off the input path field


tieujason330:
@calvinjia hmm i'm not sure what you mean. the fileview page is still needs browse's other state/props fields

calvinjia:
@tieujason330 I think there is some overlap but I'm not convinced that the overlap vs difference is large enough to use the same views. Perhaps we could further modularize the overlapping portion? In the backend code, these two go through two separate if blocks.

calvinjia:
For this PR, could you resolve the conflicts?

tieujason330:
@calvinjia conflicts are resolved. 

Hm, yeah I agree with modularizing. I'm looking into commonalities between several components right now to see what i can extract (`Logs`, `Browse`, `BlockInfo`)

calvinjia:

calvinjia:
@tieujason330 Thanks for the fix! Could you cherry-pick this to branch-2.0?

PR #9555, opened 2019-07-24
Parallelize checkConsistency repairs

### LINKED ISSUE #9527: Parallelize checkConsistency repair
The checkConsistency -r command is useful for resyncing Alluxio metadata with the UFS in the case of out-of-band writes. Currently the repair part is done in one thread.

Since the list of inconsistent files is already determined at this point, there should be an option to specify the number of threads when running the command.

Current work around is using multiple CLI's to accomplish this. It is inefficient and a pain to manage.

**Urgency**
Explain why the feature is important
Better support for syncing out-of-band operations

**Additional context**

ZacBlanco:
Instead of serially repairing each URI from the list of
inconsistent URIs this PR utilizes an ExecutorService on the
client to run the repairs in parallel. The default number of
threads to use is `2*<number of cpu cores>` but users can
optionally specify their own number with the `-t` or
`--threads` option

Fixes #9527

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
@@ -28,7 +28,6 @@ import alluxio.util.FileSystemOptions;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
-import org.apache.commons.lang.WordUtils;
 
 import java.io.IOException;
 import java.util.Collection;
--- a/shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
@@ -16,6 +16,8 @@ import alluxio.cli.CommandUtils;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.FileSystemMasterClient;
 import alluxio.client.file.URIStatus;
+import alluxio.collections.ConcurrentHashSet;
+import alluxio.exception.AggregateException;
 import alluxio.exception.AlluxioException;
 import alluxio.exception.status.InvalidArgumentException;
 import alluxio.grpc.CheckConsistencyPOptions;
@@ -26,11 +28,16 @@ import alluxio.util.FileSystemOptions;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
+import org.apache.commons.lang.WordUtils;
 
 import java.io.IOException;
-import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * Command for checking the consistency of a file or folder between Alluxio and the under storage.
@@ -44,6 +51,18 @@ public class CheckConsistencyCommand extends AbstractFileSystemCommand {
           .desc("repair inconsistent files")
           .build();
 
+  private static final Option THREADS_OPTION =
+      Option.builder("t")
+          .longOpt("threads")
+          .required(false)
+          .hasArg(true)
+          .desc("Number of threads used when repairing consistency. Defaults to <number of cores>"
+              + " * 2. This option has no effect if -r is not specified")
+          .build();
+
+  private static final String PARSE_THREADS_FAILURE_MSG = "The threads option must be a positive "
+      + "integer";
+
   /**
    * @param fsContext the filesystem of Alluxio
    */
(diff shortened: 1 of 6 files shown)


ZacBlanco:
Some numbers from local testing:

4C/8T machine running Alluxio locally with a LocalUFS

- For a directory with 1k files, I was able to reduce the repair time by > 50%. 1.8s to ~750s
- For a directory with 10k files, I was able to reduce the repair time by > 50%. 9.8s to ~4.2s

gpang:
@ZacBlanco is the description correct? It says `The default number of
threads to use is 2* but users can `. I feel like it should say "number of cores" in there somewhere

ZacBlanco:
You're correct. It was actually in the markdown but it looks like github doesn't properly escape `<` and `>` characters unless I put them in `<pre>` blocks

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
```diff
@@ -54,7 +72,18 @@ public CheckConsistencyCommand(FileSystemContext fsContext) {
...
   protected void runPlainPath(AlluxioURI plainPath, CommandLine cl)
       throws AlluxioException, IOException {
-    runConsistencyCheck(plainPath, cl.hasOption("r"));
+    int threads;
+    try {
+      threads = cl.hasOption(THREADS_OPTION.getOpt())
+          ? Integer.parseInt(cl.getOptionValue(THREADS_OPTION.getOpt())) :
+          Runtime.getRuntime().availableProcessors() * 2;
+      if (threads < 1) {
+        throw new IOException(PARSE_THREADS_FAILURE_MSG);
```
gpang: in the message, can we include what `threads` was parsed as?

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
```diff
@@ -122,43 +151,85 @@ private void runConsistencyCheck(AlluxioURI path, boolean repairConsistency) thr
...
+          } catch (AlluxioException | IOException e) {
+            exceptions.add(e);
+          }
+        }, true);
       }
+
+      waitForTasks(completionService, totalUris, exceptions);
+
+      int totalDirs = inconsistentDirs.size();
       for (AlluxioURI uri : inconsistentDirs) {
```
gpang: why are dirs repaired separately from files?

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
```diff
@@ -122,43 +151,85 @@ private void runConsistencyCheck(AlluxioURI path, boolean repairConsistency) thr
...
+      waitForTasks(completionService, totalDirs, exceptions);
+      svc.shutdown();
+    }
+  }
+
+  private void waitForTasks(CompletionService svc, int nTasks, Collection<Exception> exceptions)
+      throws IOException {
+    for (int i = 0; i < nTasks; i++) {
+      try {
+        svc.take();
```
gpang: this could wait forever?

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
```diff
@@ -122,43 +151,85 @@ private void runConsistencyCheck(AlluxioURI path, boolean repairConsistency) thr
...
+            mFileSystem.exists(inconsistentUri);
+            System.out.println(inconsistentUri + " repaired");
+            System.out.println();
+          } catch (AlluxioException | IOException e) {
+            exceptions.add(e);
+          }
+        }, true);
       }
+
+      waitForTasks(completionService, totalUris, exceptions);
```
gpang: should we print out that we are waiting for everything to finish?

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
ZacBlanco: This was the behavior from previously. I just kept it the same to parallelize everything

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
ZacBlanco: :+1:

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
ZacBlanco: It's possible that it hangs forever, but it will only hang until there is a completed task available in the queue (regardless of whether the task finished successfully or with an error). That is, it would only hang if a Runnable from the service hung forever - if that's the case the repair portion of the command would have hung forever before regardless if this was parallelized or not

shell/src/main/java/alluxio/cli/fs/command/CheckConsistencyCommand.java
ZacBlanco: I don't think there's a need to. This is more or less just the code making sure we do files before directories. While this is "waiting" there will be plenty of tasks running on background threads which are printing to the CLI. To them they won't be able to tell that anything is "waiting". They'll probably end up seeing some out-of-order log messages

gpang:

PR #9556, opened 2019-07-24
Fix Job master monitor in HA mode to use correct class name

### LINKED ISSUE #9526: Job_master health check failed
**Alluxio Version:**
v2.0.0

**Describe the bug**
JOB_MASTER‘s main class is alluxio.master.AlluxioJobMaster  , but MasterHealthCheckClient check command were as follows:
> ssh -o StrictHostKeyChecking=no -o ConnectTimeout=5 hostname ps -ef | grep "alluxio.master.job.JobMaster$" | grep "java" | awk '{ print $2; }'

Class alluxio.master.job.JobMaster not on the process list so that will never sucessed. 

**To Reproduce**

**Expected behavior**

**Urgency**

**Additional context**

ZacBlanco:
The MasterHealthCheckClient uses a hard-coded fully-qualified
class name in order to do a ps -ef to check for the process on
another node.

The class name of the job master was previously updated, but the
enum with  the value here was not. This change fixes the class
name so that the health check can pass when running in HA
mode.

Fixes #9526

Initially reported by github user @lodocj

### DIFF
--- a/shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
+++ b/shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
@@ -50,7 +50,7 @@ public class MasterHealthCheckClient implements HealthCheckClient {
    */
   public enum MasterType {
     MASTER("alluxio.master.AlluxioMaster"),
-    JOB_MASTER("alluxio.master.job.JobMaster")
+    JOB_MASTER("alluxio.master.AlluxioJobMaster")
     ;
 
     private String mClassName;


calvinjia:
@ZacBlanco Thanks for the fix, is there a unit test we can add to prevent future regressions?

ZacBlanco:
I think this would be difficult because it relies on programmatically determining the main class of the Master/Job master. With my limited knowledge, I don't think there is a way to do such a thing.

Plus, the only reason we need to know the main class if because the monitor does a check with `ps -ef | grep <class name>` which is hacky to begin with.

There's been a requested feature floating around for a while now which would allow us to do a check via an RPC port instead of this - but it's not implemented yet because secondary masters don't open their RPC ports.

calvinjia:

calvinjia:
@ZacBlanco do you mind opening a cherry-pick to branch-2.0?

Issue #8260, opened 2019-01-12
Redacting aws keys for mount command


jaiseban:
**Alluxio Version:** <enter version here, e.g. 1.8.1>

**Describe the bug**
Nested mount information shows aws access and secret keys.
**To Reproduce**

1. Execute a nested mount to S3 using access key and secret key using Alluxio client
2. After this, if you just execute mount command - it displays all the mount along with access and secret keys.

**Expected behavior**
Sensitive information should be redacted

**Additional context**

apc999:
@LuQQiu  this issue has been addressed in 2.0.0 and 1.8.2-SNAPSHOT, right? please confirm

LuQQiu:
@bf8086 Is it addressed in both 2.0.0 and 1.8.2?

apc999:
I checked that this issue is solved in branch 2.0, but not in branch 1.8.
 can we cherrypick this to 1.8 branch @bf8086 ?

apc999:
ping @LuQQiu  @bf8086

LuQQiu:
Agree to cherry pick to 1.8. @bf8086 Could you help cherry pick to 1.8?

PR #9565, opened 2019-07-25
Hide credentials in mount command

bf8086:
Port to 1.8 from master branch.

This fix changed the API so we hides values of mount options that are
credentials.

pr-link: https://github.com/Alluxio/alluxio/pull/8964
change-id: cid-aff754dd27d54939eb325e2e34c26cf8508f6abc


bf8086:
@gpang PTAL.
@ns1123 FYI.

related issue: https://github.com/Alluxio/alluxio/issues/8260

core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
```diff
@@ -225,12 +225,12 @@ long getNewBlockIdForFile(AlluxioURI path)
   Map<String, MountPointInfo>  getMountTable();
 
   /**
-   * Gets the mount point information of an Alluxio path.
+   * Gets the mount point information of an Alluxio path for display purpose.
    *
    * @param path an Alluxio path which must be a mount point
    * @return the mount point information
    */
-  MountPointInfo getMountPointInfo(AlluxioURI path) throws InvalidPathException;
```
apc999: instead of rename the internal API in a patch release, can we keep its original name "`getMountPointInfo`"?

bf8086: I explicitly renamed it so that others won't accidentally use it and expect to extract raw information such as credentials. If you are worrying about compatibility, I can fork a copy of the original method just in case someone depends on it in their application.

apc999: yeah, it will be good to keep the original API

bf8086: Done.

apc999:

Issue #9469, opened 2019-07-16
Rewire Copycat communications through gRPC


ggezer:
**Is your feature request related to a problem? Please describe.**
Copycat provides a way to implement transport for communication between participating nodes. Currently we use stock NettyTransport which is not secure.

**Describe the solution you'd like**
New transport implementation that leverages our unified gRPC security.

PR #9568, opened 2019-07-25
Implement copycat transport with gRPC

ggezer:
fix https://github.com/Alluxio/alluxio/issues/9469


ggezer:
@bf8086 PTAL. Thanks!

ggezer:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+   * @param context copycat thread context
+   * @param responseObject response object  to send
+   */
+  private void sendResponse(long requestId, ThreadContext context, Object responseObject) {
+    LOG.debug("Sending response of type: {} for request: {}", responseObject.getClass().getName(),
+        requestId);
+    mTargetObserver.onNext(CopycatMessage.newBuilder()
+        .setResponseHeader(CopycatResponseHeader.newBuilder().setRequestId(requestId)
+            .setFailed(responseObject instanceof Throwable))
+        .setMessage(ByteString.copyFrom(context.serializer().writeObject(responseObject).array()))
```
bf8086: use `unsafeByteOperations.unsafeWrap()`?

ggezer:
@calvinjia PTAL. Thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatMessageServiceClientHandler.java
```diff
@@ -0,0 +1,68 @@
...
+   * RPC for establishing bi-di stream between client/server.
+   *
+   * @param responseObserver client's stream observer
+   * @return server's stream observer
+   */
+  public StreamObserver<CopycatMessage> connect(StreamObserver<CopycatMessage> responseObserver) {
+    // Create server connection that is bound to given client stream.
+    CopycatGrpcConnection clientConnection = new CopycatGrpcConnection(
+        CopycatGrpcConnection.ConnectionOwner.SERVER, mContext, mRequestTimeoutMs);
+    clientConnection.setTargetObserver(responseObserver);
```
bf8086: Can we just pass the observer in the constructor?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+        .build());
+  }
+
+  protected void handleResponseMessage(CopycatMessage response) {
+    CopycatGrpcConnection.ContextualFuture future =
+        mResponseFutures.remove(response.getResponseHeader().getRequestId());
+
+    if (future == null) {
+      return;
```
bf8086: Log a warning on missing future?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+    } else if (message.hasResponseHeader()) {
+      handleResponseMessage(message);
+    } else {
+      throw new RuntimeException("Message should contain a request/response header.");
+    }
+  }
+
+  @Override
+  public void onError(Throwable t) {
+    LOG.debug("Connection faulted. Owner: {}. Error: {}", mConnectionOwner, t);
```
bf8086: (nit) `faulted` -> `failed`?

also you don't need to explicitly print the error as an argument. log4j will print out the full exception that is specified as an extra argument.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+    if (mConnectionOwner == ConnectionOwner.SERVER) {
+      // Complete the close on client side.
+      mTargetObserver.onCompleted();
+    }
+  }
+
+  /**
+   * Times out pending requests.
+   */
+  void timeout() {
```
bf8086: Make this function private?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+      mClosed = true;
+      // Stop timeout timer.
+      mTimeoutScheduler.cancel();
+
+      // Close outstanding calls.
+      for (CopycatGrpcConnection.ContextualFuture<?> responseFuture : mResponseFutures.values()) {
+        responseFuture.getContext().executor().execute(
+            () -> responseFuture.completeExceptionally(new ConnectException("Connection closed.")));
+      }
+      mResponseFutures.clear();
```
bf8086: What happens if a new response future is added by another thread right before this line is executed?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+    mRequestCounter = new AtomicLong(0);
+    // Initialize timeout scheduler.
+    mTimeoutScheduler = context.schedule(Duration.ofMillis(mRequestTimeoutMs),
+        Duration.ofMillis(mRequestTimeoutMs / 2), this::timeout);
+    // Initialize listeners.
+    mExceptionListeners = new Listeners<>();
+    mCloseListeners = new Listeners<>();
+    // Initialize handler and response maps.
+    mHandlers = new ConcurrentHashMap<>();
+    mResponseFutures = new ConcurrentSkipListMap<>();
```
bf8086: Why do we need a `ConcurrentSkipListMap` for futures?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcServer.java
```diff
@@ -0,0 +1,94 @@
...
+            .build();
+
+    try {
+      mGrpcServer.start();
+      mActiveAddress = address;
+      resultFuture.complete(null);
+      LOG.info("Successfully started gRPC server for copycat transport at: {}", address);
+    } catch (IOException e) {
+      mGrpcServer = null;
+      LOG.debug("Failed to create gRPC server for copycat transport at: {}. Error: {}", address, e);
```
bf8086: (nit) remove  "Error: {}" so log4j will output the full exception.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcTransport.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Override
+  public synchronized void close() {
+    if (!mClosed) {
+      mClosed = true;
+
+      // Close created clients.
+      for (CopycatGrpcClient client : mClients) {
+        client.close();
```
bf8086: Can we make sure all client get closed even if one of the close call throws?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcTransport.java
```diff
@@ -0,0 +1,100 @@
...
+      }
+      mClients.clear();
+
+      // Close created servers.
+      for (CopycatGrpcServer server : mServers) {
+        server.close();
```
bf8086: Make sure all server closes regardless of errors.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
ggezer: To favor precision for timing out pending requests by iterating keys in order.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
ggezer: Nice catch. Could cause hangs as copycat trusts transport layer to timeout requests properly.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatMessageServiceClientHandler.java
ggezer: We can do here for the server, but we can't for the client side.

This is because retrieving server stream requires an existing client stream. This is cost of implementing `StreamObserver` interface in the connection implementation itself. So we can't use the connection until target's obsorver is sent explicitly.

We have this very same limitation in `SaslStreamClientDriver` and `SaslStreamServerDriver` in channel authentication logic.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -215,7 +216,11 @@ private synchronized void initServer() {
         .withHeartbeatInterval(Duration.ofMillis(mConf.getHeartbeatIntervalMs()))
         .withSnapshotAllowed(mSnapshotAllowed)
         .withSerializer(createSerializer())
-        .withTransport(new NettyTransport())
+        .withTransport(new CopycatGrpcTransport(
```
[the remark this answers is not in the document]
ggezer: This is to support gRPC auth scenarios fully, in case we need to plug a copyat client in a non-master process. I think it's good to at least internally separate client/server configuration/user. 

Will implement a convenience constructor that supplies existing one with duplicates.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
ggezer: Nice!

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,439 @@
...
+
+    // Get a new request Id.
+    long requestId = mRequestCounter.incrementAndGet();
+
+    // Serialize the request and send it over to target.
+    try {
+      mTargetObserver.onNext(CopycatMessage.newBuilder()
+          .setRequestHeader(CopycatRequestHeader.newBuilder().setRequestId(requestId))
+          .setMessage(
+              ByteString.copyFrom(future.getContext().serializer().writeObject(request).array()))
```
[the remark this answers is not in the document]
ggezer: Nice!

ggezer:
Jenkins, test this please

ggezer:
@bf8086, Updated the PR. PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Copycat transport {@link Client} implementation that uses Alluxio gRPC.
+ */
+public class CopycatGrpcClient implements Client {
+  private static final Logger LOG = LoggerFactory.getLogger(CopycatGrpcClient.class);
+
+  /** Alluxio configuration. */
```
calvinjia: Could these be final?

ggezer:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
```diff
@@ -0,0 +1,134 @@
...
+  private AlluxioConfiguration mConf;
+  /** Authentication user. */
+  private UserState mUserState;
+
+  /** Created channels. */
+  private Map<Address, GrpcChannel> mChannels;
```
calvinjia: Do we expect this client to connect to make multiple connections to the same addresses or make multiple connections to different addresses? It seems more intuitive for each client to be to one address and share channels between clients.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,552 @@
...
+      // Don't allow request if connection is closed.
+      if (mClosed) {
+        future.completeExceptionally(new IllegalStateException("Connection closed"));
+        return future;
+      }
+
+      // Get a new request Id.
+      long requestId = mRequestCounter.incrementAndGet();
+      // Register request future.
+      mResponseFutures.put(requestId, future);
```
calvinjia: Do we need to do this if `fireAndForget` is true?

ggezer: That's for when `mTargetObserver.onNext()` takes longer than the timeout.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
ggezer: Copycat try to utilize single client for making connections to multiple servers. So pooling here is in line with their usage and as a result we will be authenticating each high-level Copycat client.

Also, actual managed channel pooling is still in effect when using GrpcChannelBuilder. So there is only 1 physical connection is opened to same target host.

ggezer:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatMessageServiceClientHandler.java
bf8086: I see.

core/server/common/src/test/java/alluxio/master/journal/raft/transport/CopycatGrpcTransportTest.java
```diff
@@ -0,0 +1,347 @@
...
+    Consumer<Connection> connectionListener = new CopycatTransportTestListener((connection) -> {
+      // Set connection as established.
+      connectionEstablished.set(true);
+    });
+
+    // Catalyst thread context for managing client/server.
+    ThreadContext connectionContext = createSingleThreadContext("ClientServerCtx");
+
+    // Address for binding transport server.
+    Address address = new Address("localhost", 10000);
```
bf8086: Will this fail if the tests are ran concurrently?

core/server/common/src/test/java/alluxio/master/journal/raft/transport/CopycatGrpcTransportTest.java
```diff
@@ -0,0 +1,347 @@
...
+      failed = true;
+    }
+    Assert.assertTrue(failed);
+  }
+
+  /**
+   * Creates and binds transport server.
+   *
+   * @param context catalyst context
+   * @param serverAddress bind address
```
bf8086: Is this used?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,552 @@
...
+          mContext.executor().execute(responseAction);
+        }
+      });
+    }
+  }
+
+  /**
+   * Sends the response to target.
+   *
+   * Note: It sends an error response if given object it {@link Throwable}.
```
calvinjia: `it` -> `is a`

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
```diff
@@ -0,0 +1,552 @@
...
+    LOG.debug("Sending response of type: {} for request: {} from: '{}'",
+        responseObjectType(responseObject), requestId, mConnectionOwner);
+    // Create response message.
+    CopycatMessage.Builder messageBuilder =
+        CopycatMessage.newBuilder().setResponseHeader(CopycatResponseHeader.newBuilder()
+            .setRequestId(requestId).setFailed(responseObject instanceof Throwable));
+    // Serialize and embed response object if provided.
+    if (responseObject != null) {
+      messageBuilder.setMessage(UnsafeByteOperations
+          .unsafeWrap(context.serializer().writeObject(responseObject).array()));
```
calvinjia: What are the requirements for using `unsafeWrap`, is it safe in our case?

core/server/common/src/test/java/alluxio/master/journal/raft/transport/CopycatGrpcTransportTest.java
ggezer: nice catch.

core/server/common/src/test/java/alluxio/master/journal/raft/transport/CopycatGrpcTransportTest.java
ggezer: Does mvn run concurrently within the test class?
Assigned different ports now in case it does.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcConnection.java
ggezer: @bf8086 ?

bf8086: unsafe meaning that it will reuse the byte array instead of copy it to a private variable. In our case the byte array is not used anywhere else so it will not be a problem.

core/server/common/src/test/java/alluxio/master/journal/raft/transport/CopycatGrpcTransportTest.java
```diff
@@ -0,0 +1,346 @@
...
+    Consumer<Connection> connectionListener = new CopycatTransportTestListener((connection) -> {
+      // Set connection as established.
+      connectionEstablished.set(true);
+    });
+
+    // Catalyst thread context for managing client/server.
+    ThreadContext connectionContext = createSingleThreadContext("ClientServerCtx");
+
+    // Address for binding transport server.
+    Address address = new Address("localhost", 10001);
```
calvinjia: Should we registered these with `ReservedPorts`?

ggezer: `PortCoordination` and `ReservedPort` utilities are for minicluster tests.
Changed static ports to binding dynamically.

core/server/common/src/test/java/alluxio/master/journal/raft/transport/CopycatGrpcTransportTest.java
ggezer: Changed it to bind ports dynamically.

ggezer:
@bf8086 Updated PR. PTAL.

ggezer:
Jenkins, test this please

calvinjia:

PR #9569, opened 2019-07-26
Change default TTLAction to null instead of DELETE for SetAttribute

LuQQiu:
TTL action in Alluxio create file/directory, set attribute are all default to `TtlAction.DELETE`.
Even after the file is created and users manually set ttl action to Free, some unrelated to setTtl set attribute operations will change the TtlAction back to DELETE.

In the workflow
```
createFile(path) .  <-- TtlAction = DELETE
setAttribute(TtlAction=FREE) .   
setAttribute(owner=new Owner) .  <-- TtlAction will change to DELETE because that's the default!
```

An example of `setAttributeOption` is 
```
SetAttributeOptions{commonOptions=CommonOptions{syncIntervalMs=-1}, pinned=null, ttl=null, ttlAction=DELETE, persisted=null, owner=NewOwner, group=null, mode=-1, recursive=false, operationTimeMs=1564102662737, ufsFingerprint=}
```

This PR change the default ttl action in setAttributeOptions to null instead of DELETE.

The issue only exist in 1.8. Master and 2.0 don't have this problem because we directly use `SetAttributePOptions` in `DefaultFileSytemMaster.setAttributeSingleFile`.
The `TtlAction` default in `SetAttributePOptions` is `null` (In 1.8, the default is `DELETE`). 

We will only set the `TtlAction` back to `DELETE` when users explicitly set the `ttlAction=DELETE` in `SetAttributePOptions` when calling the `fileSystemMasterClient.setAttribute(path, SetAttributePOptions.setTtlAction(DELETE)`.

PR #8511, opened 2019-03-04
Check for null TtlAction in DefaultFileSystemMaster

ZacBlanco:
TtlAction needs to check for null in DefaultFIleSystemMaster

The setAttribute function should only change the TtlAction if it's specifically set within the options. Otherwise, we should avoid the set. Setting the TtlAction with a null param results in the Ttl Action being set to DELETE. If the default is FREE, then a user may be inadvertently setting the action to DELETE by calling SetAttribute.
    
A test is added which fails before this fix.

PR #9570, opened 2019-07-26
Revert "Check for null TtlAction in DefaultFileSystemMaster"

yuzhu:
This reverts commit 7a8ed03367672e00579d11d4707b20527ce2263a.


yuzhu:
@ZacBlanco after looking at `branch-1.8` more closely, I think it is not necessary to cherry-pick your PR #8511 because we do not support setting `ttlaction` independent of `ttl` in `branch-1.8` and ttlaction defaults to a non-null value anyways, so we would not have a null ttlAction value. 

@LuQQiu this should fix the test failure you saw as well. 

PTAL

ZacBlanco:
@yuzhu Are we still planning on merging this?

yuzhu:
fixed via #9569

PR #9582, opened 2019-07-29
Validate only last line for leader command test

ggezer:
This is to deflake leader command integration test.

The reason for failure is that error output contains some netty error trace that we can't control. When netty receives a command for a stream that is closed or being closed, such traces are written to sys::err. In this case, it's the latter, that some stream is about to be closed when leader command hits the server. This flake is introduced by gRPC long polling. Reason for that is now that servers keep an open stream, closing a server attempts to close those as well. However that takes some time after actual complete call is called on the stream. Hence it increased exposure to receive cancelled/closed stream errors from netty.

### DIFF
--- a/tests/src/test/java/alluxio/client/cli/fs/command/LeaderCommandIntegrationTest.java
+++ b/tests/src/test/java/alluxio/client/cli/fs/command/LeaderCommandIntegrationTest.java
@@ -33,7 +33,15 @@ public final class LeaderCommandIntegrationTest extends AbstractFileSystemShellT
   public void leaderAddressNotAvailable() throws Exception {
     mLocalAlluxioCluster.stopMasters();
     mFsShell.run("leader");
-    String expected = "The leader is not currently serving requests.\n";
-    Assert.assertEquals(expected, mErrOutput.toString());
+    String expected = "The leader is not currently serving requests.";
+    Assert.assertEquals(expected, lastLine(mErrOutput.toString()));
+  }
+
+  private String lastLine(String output) {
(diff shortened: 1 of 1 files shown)


tests/src/test/java/alluxio/client/cli/fs/command/LeaderCommandIntegrationTest.java
```diff
@@ -33,7 +33,15 @@ public void leader() {
...
+    String expected = "The leader is not currently serving requests.";
+    Assert.assertEquals(expected, lastLine(mErrOutput.toString()));
+  }
+
+  private String lastLine(String output) {
+    String[] lines = output.split("\n");
+    if (lines.length > 0) {
+      return lines[lines.length - 1];
+    }
+    return "";
```
ZacBlanco: Eh it's only a test, but maybe a better way to retrieve the last line is to do

```java
int idx = output.lastIndexOf("\n");
idx = idx == -1 ? 0 : idx+1;
return output.substring(idx);
```

`output.substring(output.indexOf("\n")

ZacBlanco: Actually come to think you could just do 

```java

ggezer: After consideration, I think`String.contains()` should suffice.

ggezer:
@calvinjia PTAL. Thanks!

calvinjia:

PR #9584, opened 2019-07-29
Throw an exception on interrupt for DynamicResourcePool

calvinjia:
Await checks for the interrupted flag which will lead to an infinite loop.

### DIFF
--- a/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
+++ b/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
@@ -342,7 +342,7 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
             throw new TimeoutException("Acquire resource times out.");
           }
         } catch (InterruptedException e) {
-          throw new IOException("Thread interrupted while acquiring client.");
+          throw new IOException("Thread interrupted while acquiring client in pool: " + this);
         }
       }
     } finally {
--- a/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
+++ b/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
@@ -342,7 +342,7 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
             throw new TimeoutException("Acquire resource times out.");
           }
         } catch (InterruptedException e) {
-          throw new IOException("Thread interrupted while acquiring client in pool: " + this);
+          throw new IOException("Thread interrupted while acquiring client from pool: " + this);
         }
       }
     } finally {
--- a/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
+++ b/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
@@ -302,6 +302,7 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
    * @param unit the unit to use for time
    * @return a resource taken from the pool
    * @throws TimeoutException if it fails to acquire because of time out
+   * @throws IOException if the thread is interrupted while acquiring the resource
    */
   @Override
   public T acquire(long time, TimeUnit unit) throws TimeoutException, IOException {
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
```diff
@@ -341,8 +342,7 @@ public T acquire(long time, TimeUnit unit) throws TimeoutException, IOException
             throw new TimeoutException("Acquire resource times out.");
           }
         } catch (InterruptedException e) {
-          // Restore the interrupt flag so that it can be handled later.
-          Thread.currentThread().interrupt();
+          throw new IOException("Thread interrupted while acquiring client.");
```
gpang: should we add `this` or some distinguishing identifier so we know which pool this is happening in?

calvinjia: Unfortunately we can't get the type of resource without instantiating one. I added this to the message, but that will not be very helpful (just let us know if multiple pools are getting interrupted or just one).

apc999:
Jenkins, test this please

apc999:

apc999:

apc999:

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
sopel39: Please chain `InterruptedException` to newly created `IOException`?

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
```diff
@@ -302,6 +302,7 @@ public T acquire() throws IOException {
    * @param unit the unit to use for time
    * @return a resource taken from the pool
    * @throws TimeoutException if it fails to acquire because of time out
+   * @throws IOException if the thread is interrupted while acquiring the resource
```
kokosing: Why not to throw `InterruptedException` directly out of this method? That sounds natural for a blocking method.

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
```diff
@@ -341,8 +342,7 @@ public T acquire(long time, TimeUnit unit) throws TimeoutException, IOException
             throw new TimeoutException("Acquire resource times out.");
           }
         } catch (InterruptedException e) {
-          // Restore the interrupt flag so that it can be handled later.
-          Thread.currentThread().interrupt();
+          throw new IOException("Thread interrupted while acquiring client from pool: " + this);
```
findepi: Whenever you react to thread interrupt and throw something else than `InterruptedException`, you should restore interrupt flag:
```
Thread.currentThread().interrupt();
```

Otherwise, the thread interrupt status gets lost and caller will likely not stop what's it's doing. 
For example, you can retry on some generic IOException, but you should not retry on InterruptedException. Likewise, you should not retry when thread is interrupted state.

hustnn: Should still restore the flag as @findepi suggested like this?

```
          Thread.currentThread().interrupt();
          throw new IOException("Thread interrupted while acquiring client from pool: " + this);
```

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
calvinjia: You have a good point, and I think that is the right approach. This was a quick patch and we would like to follow up with refactoring in the master branch. See https://github.com/Alluxio/alluxio/issues/9679 
Adding `InterruptedException` would have impacted a large part of the codebase which is why we settled on this implementation.

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
calvinjia: Yeah I think this is the best practice when attempting to handle an interrupt. 
As a permanent fix, I prefer propagating the exception.

PR #9588, opened 2019-07-30
Remove missing test call for marking cluster as success

ggezer:
Jenkins was keeping an artifact for these tests even when they were successful.


tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTest.java
```diff
@@ -172,7 +172,6 @@ public void restart() throws Exception {
     assertTrue(fs.exists(testDir));
     restartMasters();
     assertTrue(fs.exists(testDir));
-    mCluster.saveWorkdir();
```
calvinjia: Was this previously a bug/left for debugging purposes?

ggezer: Most probably left from a debugging session. It exists in the original commit for adding the test.

calvinjia:

PR #9597, opened 2019-07-30
Implement a ConcurrentIdentityHashMap

### LINKED ISSUE #9449: Implement a concurrent hashmap with identity semantics
**Summary**
gRPC data marshaller has a dependency on an old netty package for `ConcurrentIdentityHashMap' implementation. Replacing that with in-house implementation or importing from a more lightweight package could strip alluxio code of all netty dependencies.

ZacBlanco:
- Implementation of a ConcurrentIdentityHashMap to avoid using
Netty dependencies

- Removed the `io.netty` dependency in the core-common
module. The rest of the dependency cleanup throughout the POMs
is occurring in https://github.com/Alluxio/alluxio/pull/9455

Closes #9449

### DIFF
--- a/core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
+++ b/core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
@@ -1,0 +1,347 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.collections;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+
+public class ConcurrentIdentityHashMap<K, V> implements ConcurrentMap<K, V> {
+
+  private final ConcurrentHashMap<IdentityObject<K>, V> mInternalMap;
+
+  /**
+   * Creates a new, empty map with the default initial table size (16).
+   */
+  public ConcurrentIdentityHashMap() {
+    mInternalMap = new ConcurrentHashMap<>();
+  }
+
+  /**
+   * Creates a new, empty map with an initial table size
+   * accommodating the specified number of elements without the need
+   * to dynamically resize.
+   *
+   * @param initialCapacity The implementation performs internal
+   * sizing to accommodate this many elements.
+   * @throws IllegalArgumentException if the initial capacity of
+   * elements is negative
+   */
+  public ConcurrentIdentityHashMap(int initialCapacity) {
+    mInternalMap = new ConcurrentHashMap<>(initialCapacity);
+  }
+
+  /**
+   * Creates a new, empty map with an initial table size based on
+   * the given number of elements ({@code initialCapacity}) and
+   * initial table density ({@code loadFactor}).
+   *
+   * @param initialCapacity the initial capacity. The implementation
+   * performs internal sizing to accommodate this many elements,
+   * given the specified load factor.
+   * @param loadFactor the load factor (table density) for
+   * establishing the initial table size
+   * @throws IllegalArgumentException if the initial capacity of
+   * elements is negative or the load factor is nonpositive
+   */
+  public ConcurrentIdentityHashMap(int initialCapacity, float loadFactor) {
+    this(initialCapacity, loadFactor, 1);
+  }
+
+  /**
+   * Creates a new, empty map with an initial table size based on
+   * the given number of elements ({@code initialCapacity}), table
+   * density ({@code loadFactor}), and number of concurrently
+   * updating threads ({@code concurrencyLevel}).
+   *
+   * @param initialCapacity the initial capacity. The implementation
+   * performs internal sizing to accommodate this many elements,
+   * given the specified load factor.
+   * @param loadFactor the load factor (table density) for
+   * establishing the initial table size
+   * @param concurrencyLevel the estimated number of concurrently
+   * updating threads. The implementation may use this value as
+   * a sizing hint.
+   * @throws IllegalArgumentException if the initial capacity is
+   * negative or the load factor or concurrencyLevel are
+   * nonpositive
+   */
+  public ConcurrentIdentityHashMap(int initialCapacity,
+      float loadFactor, int concurrencyLevel) {
+    mInternalMap = new ConcurrentHashMap<>(initialCapacity, loadFactor, concurrencyLevel);
+  }
+
+  @Override
+  public boolean containsKey(Object k) {
+    return mInternalMap.containsKey(new IdentityObject<>(k));
+  }
+
+  @Override
+  public V put(K k, V v) {
+    return mInternalMap.put(new IdentityObject<>(k), v);
+  }
+
+  @Override
+  public V putIfAbsent(K k, V v) {
+    return mInternalMap.putIfAbsent(new IdentityObject<>(k), v);
+  }
+
+  @Override
+  public boolean remove(Object k, Object v) {
+    return mInternalMap.remove(new IdentityObject<>(k), v);
+  }
+
+  @Override
+  public V remove(Object k) {
+    return mInternalMap.remove(new IdentityObject<>(k));
+  }
+
+  @Override
+  public boolean replace(K k, V v1, V v2) {
+    return mInternalMap.replace(new IdentityObject<>(k), v1, v2);
+  }
+
+  @Override
+  public V replace(K k, V v) {
+    return mInternalMap.replace(new IdentityObject<>(k), v);
+  }
+
+  @Override
+  public V get(Object k) {
+    return mInternalMap.get(new IdentityObject<>(k));
+  }
+
+  @Override
+  public Collection<V> values() {
+    return mInternalMap.values();
+  }
+
+  @Override
+  public void clear() {
+    mInternalMap.clear();
+  }
+
+  /**
+   * Return a set representing the entries of this map at the time of calling.
+   *
+   * Note, this method will create a copy of the map's entry set at creation time. Standard
+   * behavior in the {@link ConcurrentHashMap} says that the entry set backed by the map is
+   * updated when the underlying map is updated. This is not the case for this class.
+   *
+   * @return The set of entries in the map at time of calling.
+   */
+  @Override
+  public Set<Map.Entry<K, V>> entrySet() {
+    return mInternalMap.entrySet().stream().map(IdentityEntry<K, V>::new)
+        .collect(Collectors.toSet());
+  }
+
+  @Override
+  public void putAll(Map<? extends K, ? extends V> map) {
+    map.forEach((k, v) -> mInternalMap.put(new IdentityObject<>(k), v));
+  }
+
+  @Override
+  public boolean isEmpty() {
+    return mInternalMap.isEmpty();
+  }
+
+  @Override
+  public int size() {
+    return mInternalMap.size();
+  }
+
+  @Override
+  public boolean containsValue(Object value) {
+    return mInternalMap.containsValue(value);
+  }
+
+  /**
+   * Retrieves a set representing all keys contained within the map at time of calling.
+   *
+   * Note this method departs slightly from the standard {@link ConcurrentHashMap} implementation
+   * by providing its own implementation of a Set for {@link IdentityObject}. Not all methods
+   * have been implemented. Namely the {@code toArray} and {@code add*}.
+   *
+   * @return The set of keys in the map.
+   */
+  @Override
+  public Set<K> keySet() {
+    Set<IdentityObject<K>> set = mInternalMap.keySet();
+    return new Set<K>() {
+      @Override
+      public int size() {
+        return set.size();
+      }
+
+      @Override
+      public boolean isEmpty() {
+        return set.isEmpty();
+      }
+
+      @Override
+      public boolean contains(Object o) {
+        return set.contains(new IdentityObject<>(o));
+      }
+
+      @Override
+      public Iterator<K> iterator() {
+        return set.stream().map(IdentityObject::get).iterator();
+      }
+
+      @Override
+      public Object[] toArray() {
+        throw new UnsupportedOperationException("toArray not supported in identity keySet");
+      }
+
+      @Override
+      public <T> T[] toArray(T[] a) {
+        throw new UnsupportedOperationException("toArray not supported in identity keySet");
+      }
+
+      @Override
+      public boolean add(K k) {
+        throw new UnsupportedOperationException("add not supported in keySet");
+      }
+
+      @Override
+      public boolean remove(Object o) {
+        return set.remove(new IdentityObject<>(o));
+      }
+
+      @Override
+      public boolean containsAll(Collection<?> c) {
+        return set.containsAll(c);
+      }
+
+      @Override
+      public boolean addAll(Collection<? extends K> c) {
+        throw new UnsupportedOperationException("addAll not supported in keySet");
+      }
+
+      @Override
+      public boolean retainAll(Collection<?> c) {
+        return set.retainAll(c);
+      }
+
+      @Override
+      public boolean removeAll(Collection<?> c) {
+        return set.retainAll(c);
+      }
+
+      @Override
+      public void clear() {
+        set.clear();
+      }
+    };
+  }
+
+
+  private class IdentityObject<T> {
+    private T mObj;
+
+    public IdentityObject(T o) {
+      this.mObj = o;
+    }
+
+    public T get() {
+      return mObj;
+    }
+
+    @Override
+    public int hashCode() {
+      // Robert Jenkins 32-bit integer hash
+      int a = System.identityHashCode(mObj);
+      a = (a+0x7ed55d16) + (a<<12);
+      a = (a^0xc761c23c) ^ (a>>19);
+      a = (a+0x165667b1) + (a<<5);
+      a = (a+0xd3a2646c) ^ (a<<9);
+      a = (a+0xfd7046c5) + (a<<3);
+      a = (a^0xb55a4f09) ^ (a>>16);
+      return a;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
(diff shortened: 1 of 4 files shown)


core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,347 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+
+public class ConcurrentIdentityHashMap<K, V> implements ConcurrentMap<K, V> {
```
apc999: javadoc what this IdentityHashMap is expected to do on top of ConcurrentMap

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,347 @@
...
+
+  private final ConcurrentHashMap<IdentityObject<K>, V> mInternalMap;
+
+  /**
+   * Creates a new, empty map with the default initial table size (16).
+   */
+  public ConcurrentIdentityHashMap() {
+    mInternalMap = new ConcurrentHashMap<>();
```
apc999: `this(16)` to be explicit on the initial table size?

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
ZacBlanco: Will do! Still have to fix the style errors 🙃

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,347 @@
...
+
+    public T get() {
+      return mObj;
+    }
+
+    @Override
+    public int hashCode() {
+      // Robert Jenkins 32-bit integer hash
+      int a = System.identityHashCode(mObj);
+      a = (a+0x7ed55d16) + (a<<12);
```
ggezer: Why do we need to hash output of `System.identityHashCode()`?

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
ZacBlanco: I want to keep the implementation equivalent to the `ConcurrentHashMap`. I think that just leaving this no-arg constructor the same as the `ConcurrentHashMap()` implementation makes more sense?

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
ZacBlanco: This is similar to the behavior that the netty version of this class implemented. I'm not 100% sure why, but having this here vs. not using the jenkins hash I think will depend on how much we trust the `System.identityHash` to be a "good" hash.

ZacBlanco: Here's the relevant code: 

https://github.com/netty/netty/blob/3.10/src/main/java/org/jboss/netty/util/internal/ConcurrentIdentityHashMap.java#L112-L142

ggezer: I see. In a nutshell this is to regularize bit positions of hash value for efficient bucketing. Lets mention the reason in comments.

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,347 @@
...
+    return mInternalMap.values();
+  }
+
+  @Override
+  public void clear() {
+    mInternalMap.clear();
+  }
+
+  /**
+   * Return a set representing the entries of this map at the time of calling.
```
apc999: nit: returns

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+public class ConcurrentIdentityHashMapTest {
+
+
```
apc999: nit: extra empty line

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,347 @@
...
+
+  /**
+   * Creates a new, empty map with the default initial table size (16).
+   */
+  public ConcurrentIdentityHashMap() {
+    mInternalMap = new ConcurrentHashMap<>();
+  }
+
+  /**
+   * Creates a new, empty map with an initial table size
```
apc999: nit: are you wrapping the linewidth too early? it does not look like 100 char for me

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,347 @@
...
+   * elements is negative
+   */
+  public ConcurrentIdentityHashMap(int initialCapacity) {
+    mInternalMap = new ConcurrentHashMap<>(initialCapacity);
+  }
+
+  /**
+   * Creates a new, empty map with an initial table size based on
+   * the given number of elements ({@code initialCapacity}) and
+   * initial table density ({@code loadFactor}).
```
apc999: with default concurrency level (1)?

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+    String k = new String("test");
+    String k2 = new String("test");
+    m.put(k, true);
+    m.put(k2, false);
+    assertFalse(m.isEmpty());
+    assertEquals(2, m.size());
+    assertTrue(m.containsKey(k));
+    assertTrue(m.containsKey(k2));
+    assertEquals(2, m.entrySet().size());
+    for(Map.Entry e : m.entrySet()) {
```
apc999: iterate and assert the exact key value pairs -> (k, true) and (k2, false)

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+      assertTrue(e.getKey() == k || e.getKey() == k2);
+    }
+    m.remove("test"); // Should not remove, because it doesn't have the correct obj ref
+    assertEquals(2, m.size());
+    m.remove(k); // remove with correct identity ref
+    assertEquals(1, m.size());
+    m.remove(k); // Remove twice should not work
```
apc999: assert null?

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertEquals(2, m.size());
+    assertTrue(m.containsKey(k));
+    assertTrue(m.containsKey(k2));
+    assertEquals(2, m.entrySet().size());
+    for(Map.Entry e : m.entrySet()) {
+      assertTrue(e.getKey() == k || e.getKey() == k2);
+    }
+    m.remove("test"); // Should not remove, because it doesn't have the correct obj ref
+    assertEquals(2, m.size());
+    m.remove(k); // remove with correct identity ref
```
apc999: assert the val?

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertEquals(1, m.size());
+    m.remove(k); // Remove twice should not work
+    m.remove(k2);
```
apc999: assert the val?

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+    assertEquals(1, mMap.size());
+    assertEquals("v1", mMap.get(t));
+    assertFalse(mMap.containsKey("test"));
+  }
+
+  @Test
+  public void keySet() {
+    String x = new String("x");
+    String xx = new String("x");
+    mMap.put(x, "x");
```
apc999: assert null?

core/base/src/test/java/alluxio/collections/ConcurrentIdentityHashMapTest.java
```diff
@@ -0,0 +1,142 @@
...
+    mMap.put(xx, "x2");
```
apc999: assert null?

apc999: and other operations in the following tests

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
apc999: javadoc here why we want to update the default identityHashCode

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
ZacBlanco: I borrowed some of these descriptions from ConcurrentMap constructor. I will change them

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,383 @@
...
+   * Note, this method will create a copy of the map's entry set at creation time. Standard
+   * behavior in the {@link ConcurrentHashMap} says that the entry set backed by the map is
+   * updated when the underlying map is updated. That is not the case for this class.
+   *
+   * @return The set of entries in the map at time of calling
+   */
+  @Override
+  public Set<Map.Entry<K, V>> entrySet() {
+    return mInternalMap.entrySet().stream().map(IdentityEntry<K, V>::new)
+        .collect(Collectors.toSet());
```
calvinjia: Will this be much more expensive, why can't we use the default implementation?

ggezer: JFYI, we'll only be using `put`, `remove` and `values` methods.

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
```diff
@@ -0,0 +1,383 @@
...
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (!(o instanceof IdentityObject)) {
+        // If it isn't an instance of an identity object, compare our internal pointer with the arg
+        // A form of "auto unboxing" when performing comparisons
+        return mObj == o;
+      }
```
ggezer: This looks dangerous. It may hide some issues in the future if we change the implementation. Do we really need unboxing?

ZacBlanco: Not necessary but I thought it was kind of neat when writing that I could write this such that I could compare an `IdentityObject<T>` with an instance of `T` to check if they were equal without having to do  `IdentityObject<T>.equals(new IdentityObject<T>(other))`

ZacBlanco: I can change this to simply return false

core/base/src/main/java/alluxio/collections/ConcurrentIdentityHashMap.java
ZacBlanco: The internal map's entry set returns type `Set<Map.Entry<IdentityObject<K>, V>>`. To have a compatible return type I needed to this. Unfortunately I couldn't find a way to automatically unbox the `K` type from the `IdentityObject<K>` so I settled with this slower implementation for now. The caveats are noted in the javadoc of the class and on the method.


I think it would be possible to implement an auto-unboxing class internally, but I felt that because we don't use `entrySet()` elsewhere in the codebase that this would be unnecessary work. If you feel having an auto-unboxing implementation is necessary, then I can probably add one (similar to what I did for `keySet()`) but it will add a good bit of implementation

calvinjia: Thanks for the explanation.

ZacBlanco:
jenkins, test this please

ZacBlanco:
A note to hold off on merging this as I still want to remove the org.jboss netty dependency from the POM files in this PR as well.

ggezer:
@ZacBlanco If you'll replace the dependency in this PR it would be good to have proper test validation prior to merge for verifying performance is not impacted.

gpang:
@ZacBlanco the description has `<placeholder for description of changes to netty dependencies>`. Did you want to update that?

ZacBlanco:
@gpang I still haven't made the changes necessary to update the description for that placeholder

apc999:
is this pr ready to merge? @ZacBlanco

ZacBlanco:
@apc999 no I still want to push one more commit

ggezer:
@bf8086 Any advise on how to put `DataMessageMarshaller`'s buffer map under stress ?

ZacBlanco:
@ggezer I removed the netty dependency in the core-common module POM in the most recent commit. I'll leave the rest of the work removing the dependency throughout the codebase to you in https://github.com/Alluxio/alluxio/pull/9455

Still looking into stress testing. I might also try a few synthetic tests with the new HashMap vs the old one just to get a relative idea about performance

ZacBlanco:
If we are good with the final change to the POM in this PR is can now be merged

ZacBlanco:
Here is the code I used to run some synthetic benchmarks:

```java

  @Test
  public void testConcurrentHashMap() throws Exception {
    List<Map<Long, Long>> maps = new ArrayList<>();
    maps.add(new ConcurrentIdentityHashMap<>());
    maps.add(new alluxio.collections.ConcurrentIdentityHashMap<>());
    long seed = new Random().nextLong();
    Integer[] i = { 10_000, 100_000, 500_000 };
    Integer[] p = { 4, 8, 16, 32 };
    List<Integer> itemTest = Arrays.asList(i);
    List<Integer> parallelismVals = Arrays.asList(p);

    int trials = 50;
    for (int numItems : itemTest) {
      for (int parallelism : parallelismVals) {
        System.out.println(String.format("Trials for %s map items with %s parallelism", numItems,
         parallelism));
        ExecutorService svc = Executors.newFixedThreadPool(parallelism);
        maps.forEach(map -> serialMapPerfTest(map, seed, trials, numItems));
        maps.forEach(map -> parallelMapPerfTest(map, seed, trials, numItems, parallelism, svc));
        maps.forEach(m -> m.clear());
        svc.shutdown();
        System.gc();
        // System.gc() is more like a suggestion to the JVM to gc so sleep this thread
        // a few seconds to let the JVM gc if it decides to
        Thread.sleep(2000);
      }
    }
  }

  public void serialMapPerfTest(Map<Long, Long> map, long seed, long trials, int numItems) {
    Random r = new Random(seed);
    List<Long> putTimes = new ArrayList<>();
    List<Long> getTimes = new ArrayList<>();


    for (int i = 0; i < trials; i++) {
      map.clear();
      for (int j = 0; j < numItems; j++) {
        Long l = r.nextLong();
        map.put(l, l);
      }
      long end = System.currentTimeMillis();
      putTimes.add(end-start);
      start = System.currentTimeMillis();
      for (Long k : map.values()) {
        map.get(k);
      }
      end = System.currentTimeMillis();
      getTimes.add(end-start);
    }
    Double putTime = ((double) putTimes.stream().reduce((l1, l2) -> l1+l2).get()) / trials;
    System.out.println(String.format("Serial mean put time for %s: %s",
        map.getClass().getCanonicalName(),
        putTime));

    Double getTime = ((double) getTimes.stream().reduce((l1, l2) -> l1+l2).get()) / trials;
    System.out.println(String.format("Serial mean get time for %s: %s",
        map.getClass().getCanonicalName(),
        getTime));
  }

  public void parallelMapPerfTest(Map<Long, Long> map, long seed,
      long trials, int numItems, int parallelism, ExecutorService service) {
    map.clear();
    Random r = new Random(seed);
    AtomicInteger counter = new AtomicInteger(0);

    List<Long> putTimes = new ArrayList<>();
    List<Long> getTimes = new ArrayList<>();

    // Run for X trials
    for (int x = 0; x < trials; x++) {
      map.clear();
      counter.set(0);
      // Fill map with items and time it
      runNThreads(() -> {
        while (counter.getAndIncrement() < numItems) {
          Long l = r.nextLong();
          map.put(l, l);
        }
      }, parallelism, service);
      long end = System.currentTimeMillis();
      putTimes.add(end-start);
      // Get the list of all keys in the map and shuffle them into a queue
      ArrayList<Long> l = new ArrayList<>(map.keySet().stream().collect(Collectors.toList()));
      Collections.shuffle(l);
      ConcurrentLinkedQueue<Long> q = new ConcurrentLinkedQueue<>(l);
      start = System.currentTimeMillis();
      // Time how long it takes flush the queue while doing a containsKey() across all items in the
      // map
      runNThreads(() -> {
        while (!q.isEmpty()) {
          map.containsKey(q.poll());
        }
      }, parallelism, service);
      end = System.currentTimeMillis();
      getTimes.add(end-start);
    }

    Double putTime = ((double) putTimes.stream().reduce((l1, l2) -> l1+l2).get()) / trials;
    System.out.println(String.format("Parallel mean put time for %s: %s",
        map.getClass().getCanonicalName(),
        putTime));

    Double getTime = ((double) getTimes.stream().reduce((l1, l2) -> l1+l2).get()) / trials;
    System.out.println(String.format("Parallel mean get time for %s: %s",
        map.getClass().getCanonicalName(),
        getTime));
  }

  /**
   * run N tasks across an executor service until all have finished successfully.
   */
  private void runNThreads(Runnable r, int parallelism, ExecutorService svc) {
    List<Future<?>> futures = new ArrayList<>();
    for (int i = 0; i < parallelism; i++) {
      futures.add(svc.submit(r));
    }
    waitForFutures(futures);
  }

  /**
   * Blocks until all futures in a list
   */
  private void waitForFutures(List<Future<?>> futures) {
    for (Future<?> f : futures) {
      try {
        f.get();
      } catch (Exception e) {
        throw new RuntimeException(String.format("Got exception while running: %s", e));
      }
    }
  }
```
Results for parallel benchmark, on 2015 MBP w/ an Intel Core i7 (4C/8T), 16GB DDR3  1660MHz RAM. Java `build 1.8.0_191-b12`
Graphs:
image

Table:
image

apc999:
LGTM

calvinjia:

PR #9607, opened 2019-07-30
Do selective tracing for client stream errors

ggezer:
This is popping up quite often while bringing a cluster online and it's not very intuitive for users.
Stream errors are received asynchronous to channel building, which makes them harder to debug. Ignoring it for common case of UNAVAILABLE should be safe.

### DIFF
--- a/core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
+++ b/core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
@@ -111,22 +111,16 @@ public class SaslStreamClientDriver implements StreamObserver<SaslMessage> {
 
   @Override
   public void onError(Throwable throwable) {
+    // Fail handshake if still active.
     mHandshakeFuture.setException(throwable);
-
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/security/authentication/SaslStreamClientDriver.java
```diff
@@ -111,8 +111,26 @@ public void onNext(SaslMessage saslMessage) {
...
+      }
+    }
+
+    String errorMsg = String.format("Received error on client driver for channel: %s. Error: %s",
+        mChannelId, throwable);
+    if (traceDebug) {
+      LOG.debug(errorMsg);
+    } else {
+      LOG.warn(errorMsg);
+    }
```
calvinjia: Could we simplify this block of code by doing the conditional check after we already created the error message?

ggezer: Done

calvinjia:

LuQQiu:

PR #9614, opened 2019-07-31
Add CLI options for listing last access time

bf8086:
Added some options for selecting which timestamp should be displayed when running `ls` command.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
@@ -65,6 +65,31 @@ public final class LsCommand extends AbstractFileSystemCommand {
       + LS_FORMAT_GROUP_NAME + LS_FORMAT_FILE_SIZE + LS_FORMAT_PERSISTENCE_STATE
       + LS_FORMAT_TIMESTAMP + LS_FORMAT_ALLUXIO_STATE + " " + LS_FORMAT_FILE_PATH + "%n";
 
+  private static final Map<String, Comparator<URIStatus>> SORT_FIELD_COMPARATORS = new HashMap<>();
+
+  static {
+    SORT_FIELD_COMPARATORS.put("creationTime",
+        Comparator.comparingLong(URIStatus::getCreationTimeMs));
+    SORT_FIELD_COMPARATORS.put("inMemoryPercentage",
+        Comparator.comparingLong(URIStatus::getInMemoryPercentage));
+    SORT_FIELD_COMPARATORS.put("lastAccessTime",
+        Comparator.comparingLong(URIStatus::getLastAccessTimeMs));
+    SORT_FIELD_COMPARATORS.put("lastModificationTime",
+        Comparator.comparingLong(URIStatus::getLastModificationTimeMs));
+    SORT_FIELD_COMPARATORS.put("name",
+        Comparator.comparing(URIStatus::getName, String.CASE_INSENSITIVE_ORDER));
+    SORT_FIELD_COMPARATORS.put("path", Comparator.comparing(URIStatus::getPath));
+    SORT_FIELD_COMPARATORS.put("size", Comparator.comparingLong(URIStatus::getLength));
+  }
+
+  private static final Map<String, Function<URIStatus, Long>> TIMESTAMP_FIELDS = new HashMap<>();
+
+  static {
+    TIMESTAMP_FIELDS.put("creation", URIStatus::getCreationTimeMs);
+    TIMESTAMP_FIELDS.put("access", URIStatus::getLastAccessTimeMs);
+    TIMESTAMP_FIELDS.put("modification", URIStatus::getLastModificationTimeMs);
+  }
+
   private static final Option FORCE_OPTION =
       Option.builder("f")
           .required(false)
@@ -105,8 +130,9 @@ public final class LsCommand extends AbstractFileSystemCommand {
           .required(false)
           .longOpt("sort")
           .hasArg(true)
-          .desc("sort statuses by the given field "
-                  + "{size|creationTime|inMemoryPercentage|lastModificationTime|name|path}")
+          .desc("sort statuses by the given field {"
+              + String.join("|", SORT_FIELD_COMPARATORS.keySet())
+              + "}")
           .build();
 
   private static final Option REVERSE_SORT_OPTION =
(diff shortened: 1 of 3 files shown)


bf8086:
@gpang PTAL.

bf8086:
Jenkins, test this please.

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -116,13 +116,29 @@
               .desc("reverse order while sorting")
               .build();
 
+  private static final Option USE_ACCESS_TIME_OPTION =
```
gpang: instead of having 2 options for all these timestamps, we should have a single option like "--timestamp-format", which takes a string, and can be `m`, `c`, or `a`. 
- `m`: use modified time
- `c`: use creation time
- `a`: use access time

And the default value should be `m`, or whatever the actual default is. Would this work?

apc999: How about `--timestamp <modified | created | accessed>`?
in general I'm prefer using longer and meaningful option names rather than a shorter but meaningless one-char options

gpang: yeah that works. The main point is to have a single option that "picks" which timestamp to use for the time column.

bf8086: I agree we should use a single option. I am not sure how to implement that with apache common CLI APIs.

Regarding the option name, both HDFS and Unix use `-u` for listing access time and Unix use `-U` for listing creation time. Shall we keep the same convention to avoid mental overhead for user to remember different parameter scheme?

apc999: I don't think `-u` is commonly known any way. Plus IMO, HDFS and Unix are not good examples to follow w.r.t. CLI syntax. Anyway, my point is we don't have to follow their approach for detailed option names.

gpang: `-u` and `-U` seems confusing. For example, what happens if you use both at the same time? It seems like having a single option which controls the timestamp type is more intuitive.

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -116,13 +116,24 @@
...
+          .build();
+
   private static final Map<String, Comparator<URIStatus>> SORT_FIELD_COMPARATORS = new HashMap<>();
 
   static {
     SORT_FIELD_COMPARATORS.put("creationTime",
         Comparator.comparingLong(URIStatus::getCreationTimeMs));
     SORT_FIELD_COMPARATORS.put("inMemoryPercentage",
         Comparator.comparingLong(URIStatus::getInMemoryPercentage));
+    SORT_FIELD_COMPARATORS.put("lastAccessTime",
```
gpang: `lastAccessTime` should be added to the description of `SORT_OPTION`

bf8086: Updated.

bf8086:
Jenkins, test this please.

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -53,17 +53,42 @@
...
+    SORT_FIELD_COMPARATORS.put("name",
+        Comparator.comparing(URIStatus::getName, String.CASE_INSENSITIVE_ORDER));
+    SORT_FIELD_COMPARATORS.put("path", Comparator.comparing(URIStatus::getPath));
+    SORT_FIELD_COMPARATORS.put("size", Comparator.comparingLong(URIStatus::getLength));
+  }
+
+  private static final Map<String, Function<URIStatus, Long>> TIMESTAMP_FIELDS = new HashMap<>();
+
+  static {
+    TIMESTAMP_FIELDS.put("creation", URIStatus::getCreationTimeMs);
```
apc999: since we already have similar fields in `SORT_FIELD_COMPARATORS`, let's use `creationTime`, `lastAccessTime`, `lastModificationTime` as the options to be consistent in `TIMESTAMP_FIELDS` too?

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -226,14 +255,15 @@ public Options getOptions() {
    * @param sortField sort the result by this field
    */
   private void ls(AlluxioURI path, boolean recursive, boolean forceLoadMetadata, boolean dirAsFile,
-                  boolean hSize, boolean pinnedOnly, String sortField, boolean reverse)
+                  boolean hSize, boolean pinnedOnly, String sortField, boolean reverse,
```
apc999: nit: is the indent correct?

tests/src/test/java/alluxio/client/cli/fs/command/LsCommandIntegrationTest.java
```diff
@@ -290,6 +290,52 @@ public void lsWithSortByPath() throws IOException, AlluxioException {
...
+
+    FileSystemTestUtils.loadFile(mFileSystem, oldFileRecentlyAccessed);
+    mFsShell.run("ls", "--sort", "lastAccessTime", "/testRoot");
+    checkOutput(
+        "             10   NOT_PERSISTED .+ .+ 100% " + oldFileName,
+        "             10   NOT_PERSISTED .+ .+ 100% " + oldFileRecentlyAccessed);
+  }
+
+  /**
+   * Tests ls command with sort by access time option.
```
apc999: incorrect javadoc. Actually typically we don't javadoc test function any more, unless the test is very tricky

apc999: and can you add test for `--timestamp`?

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -211,6 +239,7 @@ public Options getOptions() {
         .addOption(LIST_DIR_AS_FILE_OPTION)
         .addOption(LIST_HUMAN_READABLE_OPTION)
         .addOption(LIST_PINNED_FILES_OPTION)
+        .addOption(TIMESTAMP_OPTION)
```
apc999: NIT: Keep the options sorted alphabetically.

apc999:
Left some minor comments. Please also update `getUsage()` method and the CLI doc?

tests/src/test/java/alluxio/client/cli/fs/command/LsCommandIntegrationTest.java
bf8086: Done

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
bf8086: Checkstyles doesn't seem to complain about it.

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
bf8086: Sorted.

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
bf8086: I usually prefer shorter arguments, but I can also see that we want to be consistent between arguments. I am fine either way.

bf8086:
Thanks @apc999! I updated the PR. PTAL.

apc999:

PR #9624, opened 2019-08-01
Create and Close copycat transport clients/servers asynchronously

ggezer:
Returning a completed future from connection creation was causing copycat to do a premature cleanup, which in turn was causing a race condition at copycat side. This PR makes relevant calls truly async.


ggezer:
@calvinjia PTAL. Thanks!

calvinjia:

PR #9625, opened 2019-08-01
Prevent persist job from updating last access time

bf8086:
Persist job uses `FileSystem.open` for reading file content from worker, which causes the access time to be updated. This change add an option to open file API to allow internal clients to indicate that access time should not be updated.

### DIFF
--- a/core/transport/src/grpc/file_system_master.proto
+++ b/core/transport/src/grpc/file_system_master.proto
@@ -64,6 +64,7 @@ message OpenFilePOptions {
   optional ReadPType readType = 1;
   optional int32 maxUfsReadConcurrency = 2;
   optional FileSystemMasterCommonPOptions commonOptions = 3;
+  optional bool updateLastAccessTime = 4 [default = true];
 }
 
 message CreateDirectoryPResponse {}
@@ -145,6 +146,7 @@ message GetStatusPOptions {
   optional LoadMetadataPType loadMetadataType = 1;
   optional FileSystemMasterCommonPOptions commonOptions = 2;
   optional Bits accessMode = 3;
+  optional bool updateTimestamps = 4 [default = true];
 }
 message GetStatusPRequest {
   /** the path of the file or directory */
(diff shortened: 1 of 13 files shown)


bf8086:
@gpang PTAL.

gpang:
Jenkins, test this please

calvinjia:
@bf8086 Why is it bad to update persist time in this case?

bf8086:
@calvinjia Some internal services might schedule persist jobs based on file timestamps. Updating access time on internal persist job execution can cause unwanted side effect.

core/transport/src/grpc/file_system_master.proto
```diff
@@ -64,6 +64,7 @@ message OpenFilePOptions {
   optional ReadPType readType = 1;
   optional int32 maxUfsReadConcurrency = 2;
   optional FileSystemMasterCommonPOptions commonOptions = 3;
+  optional bool updateLastAccessTime = 4 [default = true];
```
calvinjia: Should this be a common option? Is last access time updated on other calls like getStatus?

bf8086: lastAccessTime is only updated on content access(e.g. read, create, list children). It is not updated on other `getStatus` calls.

bf8086:
Jenkins, test this please

core/transport/src/grpc/file_system_master.proto
```diff
@@ -145,6 +146,7 @@ message GetStatusPOptions {
   optional LoadMetadataPType loadMetadataType = 1;
   optional FileSystemMasterCommonPOptions commonOptions = 2;
   optional Bits accessMode = 3;
+  optional bool updateTimestamps = 4 [default = true];
```
gpang: why is this called `updateTimestamps` and the other is called `updateLastAccessTime`? Also, which timestamps is this referring to?

bf8086: `OpenFilePOptions` is only for reading files and only last access time is relevant; `GetStatusPOptions` is used in `getStatus` call which is called for read, write and other operations, so it can be other mutable timestamps as well. It is not specific to a certain timestamp type but more to indicate that the call is for an internal operation that does not imply any user interactions.

gpang:

PR #9626, opened 2019-08-02
Support custom options provider for FileSystem implementations

ggezer:
Required for supporting transparent URI as each outgoing request should contain an auxiliary flag for master to decide whether to allow or deny the request for non-Alluxio scheme.


calvinjia:
@ggezer Instead of having an option, could the master always accept non-alluxio schemes?

PR #9630, opened 2019-08-03
Make copycat connections own dedicated gRPC channel

ggezer:
Earlier, gRPC channel cache in `CopycatGrpcClient` was not checking channels' health status. This was causing unhealthy/revoked channels to be reused sometimes for targets that have failed over. Restarted node was not remembering existing channels' authentication Id, so no new connections could be accepted.

Rather than disposing unhealthy channels from cache during connection building, this PR makes each connection own its own gRPC channel. There are two reasons for that:

- As a simplification of code base as copycat already does connection pooling. So channels will be cached implicitly by Copycat.
- When cluster membership changes, each copycat server resets its connection pool, which in turn closes each outstanding connection. There is no place to receive this event from transport layer, so caching gRPC channels outside causes old channels to linger around after their target is lost.

A trade off, imposed by copycat's connection reset procedure, is that when cluster membership changes, n^2 gRPC authentication will take place among servers. (n being participating node count). This is because new connections will be given their own brand new channels. This is not expected during normal operation and could happen only during bootup/failover and explicit membership change.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/transport/AbstractCopycatGrpcConnection.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/transport/AbstractCopycatGrpcConnection.java
@@ -44,10 +44,11 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 
 /**
- * Base copycat transport {@link Connection} implementation that uses Alluxio gRPC.
+ * Abstract copycat transport {@link Connection} implementation that uses Alluxio gRPC.
  */
-public class CopycatGrpcBaseConnection implements Connection, StreamObserver<CopycatMessage> {
-  private static final Logger LOG = LoggerFactory.getLogger(CopycatGrpcBaseConnection.class);
+public abstract class AbstractCopycatGrpcConnection
+    implements Connection, StreamObserver<CopycatMessage> {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractCopycatGrpcConnection.class);
 
   /** Exception listeners for this connection.  */
   private final Listeners<Throwable> mExceptionListeners;
@@ -67,9 +68,9 @@ public class CopycatGrpcBaseConnection implements Connection, StreamObserver<Cop
   private final AtomicLong mRequestCounter;
 
   /** Map of request handlers. */
-  private final Map<Class, CopycatGrpcBaseConnection.HandlerHolder> mHandlers;
+  private final Map<Class, AbstractCopycatGrpcConnection.HandlerHolder> mHandlers;
   /** Map of pending requests. */
-  private final Map<Long, CopycatGrpcBaseConnection.ContextualFuture> mResponseFutures;
+  private final Map<Long, AbstractCopycatGrpcConnection.ContextualFuture> mResponseFutures;
 
   /** Thread context of creator of this connection. */
   private final ThreadContext mContext;
@@ -98,8 +99,8 @@ public class CopycatGrpcBaseConnection implements Connection, StreamObserver<Cop
    * @param context copycat thread context
    * @param requestTimeoutMs timeout in milliseconds for requests
    */
-  public CopycatGrpcBaseConnection(ConnectionOwner connectionOwner, ThreadContext context,
-                                   long requestTimeoutMs) {
+  public AbstractCopycatGrpcConnection(ConnectionOwner connectionOwner, ThreadContext context,
+      long requestTimeoutMs) {
     mConnectionOwner = connectionOwner;
     mContext = context;
     mRequestTimeoutMs = requestTimeoutMs;
@@ -152,8 +153,8 @@ public class CopycatGrpcBaseConnection implements Connection, StreamObserver<Cop
       Assert.notNull(request, "request");
 
       // Create a contextual future for the request.
-      CopycatGrpcBaseConnection.ContextualFuture<U> future =
-              new CopycatGrpcBaseConnection.ContextualFuture<>(System.currentTimeMillis(),
+      AbstractCopycatGrpcConnection.ContextualFuture<U> future =
+              new AbstractCopycatGrpcConnection.ContextualFuture<>(System.currentTimeMillis(),
                       ThreadContext.currentContextOrThrow());
 
       // Don't allow request if connection is closed.
@@ -208,7 +209,7 @@ public class CopycatGrpcBaseConnection implements Connection, StreamObserver<Cop
       if (mClosed) {
         throw new IllegalStateException("Connection closed");
       }
-      mHandlers.put(type, new CopycatGrpcBaseConnection.HandlerHolder(handler,
+      mHandlers.put(type, new AbstractCopycatGrpcConnection.HandlerHolder(handler,
           ThreadContext.currentContextOrThrow()));
       return null;
     }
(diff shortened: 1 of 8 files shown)


ggezer:
Jenkins, test this please.

ggezer:
Jenkins, test this please

ggezer:
@calvinjia PTAL. Thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
```diff
@@ -98,8 +98,8 @@
    * @param context copycat thread context
    * @param requestTimeoutMs timeout in milliseconds for requests
    */
-  public CopycatGrpcConnection(ConnectionOwner connectionOwner, ThreadContext context,
-      long requestTimeoutMs) {
+  public CopycatGrpcBaseConnection(ConnectionOwner connectionOwner, ThreadContext context,
+                                   long requestTimeoutMs) {
```
calvinjia: formatting

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
```diff
@@ -360,37 +360,37 @@ private String responseObjectType(Object responseObject) {
...
+        // Close pending requests.
+        failPendingRequests(new ConnectException("Connection closed."));
 
-      // Call close listeners.
-      for (Listener<Connection> listener : mCloseListeners) {
-        listener.accept(this);
+        // Call close listeners.
+        for (Listener<Connection> listener : mCloseListeners) {
+          listener.accept(this);
+        }
```
calvinjia: Unrelated to this PR - with the way we close the connection, is it possible that we get lingering requests that do not get failed due to them being submitted after `failPendingRequests`?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
```diff
@@ -44,10 +44,10 @@
 import java.util.function.Function;
 
 /**
- * Copycat transport {@link Connection} implementation that uses Alluxio gRPC.
+ * Base copycat transport {@link Connection} implementation that uses Alluxio gRPC.
  */
-public class CopycatGrpcConnection implements Connection, StreamObserver<CopycatMessage> {
-  private static final Logger LOG = LoggerFactory.getLogger(CopycatGrpcConnection.class);
+public class CopycatGrpcBaseConnection implements Connection, StreamObserver<CopycatMessage> {
```
calvinjia: Should this be an abstract class now?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
```diff
@@ -101,29 +90,14 @@ public CopycatGrpcClient(AlluxioConfiguration conf, UserState userState) {
...
-
-    LOG.debug("Closing copycat transport client with {} gRPC channels.", mChannels.size());
+  public CompletableFuture<Void> close() {
+    LOG.debug("Closing copycat transport client with {} connections.", mConnections.size());
     // Close created connections.
     List<CompletableFuture<Void>> connectionCloseFutures = new ArrayList<>(mConnections.size());
-    for (Connection connection : mConnections) {
-      connectionCloseFutures.add(connection.close());
+    for (Connection connectionPair : mConnections) {
+      connectionCloseFutures.add(connectionPair.close());
```
calvinjia: Why is this a `connectionPair` now?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcServerConnection.java
```diff
@@ -0,0 +1,31 @@
...
+   * Creates a connection object for server.
+   *
+   * Note: {@link #setTargetObserver} should be called explicitly before using the connection.
+   *
+   * @param context copycat thread context
+   * @param requestTimeoutMs timeout in milliseconds for requests
+   */
+  public CopycatGrpcServerConnection(ThreadContext context, long requestTimeoutMs) {
+    super(ConnectionOwner.SERVER, context, requestTimeoutMs);
+  }
```
calvinjia: Do we need to shutdown server channels on close?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
ggezer: Early in `close()` a state lock is held exclusively and connection is marked as closed. After that new requests always check for the state and should simply fail with `ConnectionClosed`. So I think we're safe against this particular scenario.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
ggezer: Because of 1-1 mapping, I don't need separate collections but could put them alongside.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcServerConnection.java
ggezer: gRPC doesn't expose server-side channels. So there is nothing to close other than the connections themselves.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
ggezer: It's functional by itself. I can mark it abstract for restricting its use.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
calvinjia: Could new requests already be past the state check but not put into the pending requests yet?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
calvinjia: All connections should be either client or server right? If there is a use case for the base connection its fine. I would prefer to be more restrictive if possible.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcBaseConnection.java
ggezer: State lock is held in shared mode by clients for the whole duration of call.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/AbstractCopycatGrpcConnection.java
```diff
@@ -44,10 +44,11 @@
 import java.util.function.Function;
 
 /**
- * Base copycat transport {@link Connection} implementation that uses Alluxio gRPC.
+ * Abstract copycat transport {@link Connection} implementation that uses Alluxio gRPC.
  */
-public class CopycatGrpcBaseConnection implements Connection, StreamObserver<CopycatMessage> {
-  private static final Logger LOG = LoggerFactory.getLogger(CopycatGrpcBaseConnection.class);
+public abstract class AbstractCopycatGrpcConnection
```
calvinjia: Could we call this `CopycatGrpcConnection`?

ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

calvinjia:

PR #9641, opened 2019-08-05
Add NoSuchObjectException for no object found exception

HelloHorizon:
e.g. exception for getTable() (no such table), getDatabase() (no such database)


HelloHorizon:
@yuzhu please check this

HelloHorizon:
jenkins, test this please

yuzhu:
@gpang for classes like additional exceptions, should we move these into the `experimental` module?

apc999:
we typically don't add new exceptions. is this required?
If possible, we want to standardize on `AlluxioStatusException`

apc999:
Suggest to return `alluxio.exception.status.NotFoundException`

HelloHorizon:
@apc999 Good suggestion! I will try to reuse the alluxio.exception.status.NotFoundException.

PR #9643, opened 2019-08-05
Fix NPE during RaftJournal shut down

ggezer:
This issue is encountered while authoring a new integration test for growing embedded journal cluster. Stopping a standby master was failing with NPE

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -503,7 +503,9 @@ public final class RaftJournalSystem extends AbstractJournalSystem {
   @Override
   public synchronized void stopInternal() throws InterruptedException, IOException {
     LOG.info("Shutting down raft journal");
-    mRaftJournalWriter.close();
+    if (mRaftJournalWriter != null) {
+      mRaftJournalWriter.close();
+    }
     try {
       mServer.shutdown().get(ServerConfiguration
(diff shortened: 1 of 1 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -503,7 +503,9 @@ public synchronized void startInternal() throws InterruptedException, IOExceptio
   @Override
   public synchronized void stopInternal() throws InterruptedException, IOException {
     LOG.info("Shutting down raft journal");
-    mRaftJournalWriter.close();
+    if (mRaftJournalWriter != null) {
```
gpang: Do you know in what scenario the writer would be `null` but `stopInternal ` is called?

gpang:
@ggezer could you add a bit more info to the description, maybe what the scenario is that this could happen?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ggezer: When the journal is closed while it's secondary.
Writer is initialized when it's primary.

gpang: Would this happen when a secondary is promoted to a primary?

ggezer: During promotion writer is instantiated.

gpang:

LuQQiu:

PR #9644, opened 2019-08-05
Close journal system properly when stopping Alluxio processes

ggezer:
This issue is encountered while authoring a new integration test for growing embedded journal cluster. Restarting an in-process master showed that journal system was not being closed during stopping standby masters.

### DIFF
--- a/minicluster/src/main/java/alluxio/master/LocalAlluxioJobCluster.java
+++ b/minicluster/src/main/java/alluxio/master/LocalAlluxioJobCluster.java
@@ -44,6 +44,8 @@ public final class LocalAlluxioJobCluster {
   private Thread mMasterThread;
   private Thread mWorkerThread;
 
+  private boolean mIsRunning = false;
+
   /**
    * Creates a new instance of {@link LocalAlluxioJobCluster}.
    */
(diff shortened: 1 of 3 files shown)


ggezer:
Jenkins, test this please

gpang:
@ggezer could you add some information in the description of the PR? Does this fix a bug, or is this for improving unittests?

minicluster/src/main/java/alluxio/master/LocalAlluxioJobCluster.java
```diff
@@ -71,9 +74,12 @@ public void start() throws IOException, ConnectionFailedException {
    * @throws Exception when the operation fails
    */
   public void stop() throws Exception {
-    LOG.info("Stop Alluxio job service");
-    mWorker.stop();
-    mMaster.stop();
+    if (mIsRunning) {
```
gpang: what happens if you call stop when it is not running?

ggezer: Then it's a no-op as it should be. Or at least it's safe for it to be a no-op when not started.

gpang:

LuQQiu:

Issue #9295, opened 2019-06-14
Web UI error when authentication is SIMPLE but authorization is disabled


bf8086:
**Alluxio Version:**
Alluxio 1.8
**Describe the bug**

**To Reproduce**
set auth type to SIMPLE and disable authorization, web UI file browser will show error.
**Expected behavior**
Web UI should be usable.
**Urgency**
high, it makes webUI file browsing unusable

**Additional context**
![IMG_5343](https://user-images.githubusercontent.com/29311643/59482781-2822a400-8e1f-11e9-98fe-c7fbfe4d87fa.JPG)

Issue #9294, opened 2019-06-14
Make audit log not depend on authentication type


bf8086:
**Alluxio Version:**
1.8+

**Describe the bug**
If audit log is enabled and authentication is NOSASL, any operation will fail with "authentication not enabled" error. 
**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)

**Expected behavior**
Authentication should not be a dependency for audit log. If a user is not identified, audit log should log "N/A".
**Urgency**
medium
**Additional context**
Can be worked around by enabling authentication but disable authorization.

PR #9645, opened 2019-08-05
Avoid exception when audit log is used with NOSASL authentication

bf8086:
Fixes Alluxio/alluxio#9294
Fixes Alluxio/alluxio#9295

Update audit logic to be more flexible when acquiring user information.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -4520,8 +4520,8 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
           String primaryGroup = CommonUtils.getPrimaryGroupName(user, ServerConfiguration.global());
           ugi = user + "," + primaryGroup;
         } catch (IOException e) {
-          LOG.warn("Failed to get primary group for user {}.", user);
-          ugi = user;
+          LOG.debug("Failed to get primary group for user {}.", user);
+          ugi = user + ",N/A";
         }
       }
       AuthType authType =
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4504,19 +4504,25 @@ private Metrics() {} // prevent instantiation
...
+        user = AuthenticatedClientUser.getClientUser(ServerConfiguration.global());
+      } catch (AccessControlException e) {
+        ugi = "N/A";
+      }
+      if (user != null) {
+        try {
+          String primaryGroup = CommonUtils.getPrimaryGroupName(user, ServerConfiguration.global());
+          ugi = user + "," + primaryGroup;
+        } catch (IOException e) {
+          LOG.warn("Failed to get primary group for user {}.", user);
```
apc999: this will spamming the logs in this case, do you want to put it in warn level?

apc999: will this show up once or on each rpc?

bf8086: It can show up on each RPC, which might be too chatty. Do you want to make it a debug log?

apc999: When are we supposed to see the exception to thrown? It looks like we are throwing it from `alluxio.security.group.CachedGroupMapping#getGroups` which we may return an empty list instead.
If we expect one exception per RPC in the target use case, this is very very expensive and Master process will become very slow.

If we don't throw exception, maybe use just use "N/A" in ugi.group to indicate the failure to get group?

gpang:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
bf8086: makes sense. done.

bf8086:
Thanks @apc999! I updated the PR. PTAL.

bf8086:
Jenkins, test this please.

bf8086:
Jenkins, test this please.

gpang:
Jenkins, test this please

gpang:

PR #9652, opened 2019-08-06
Ensure tags are copied when overriding metrics

### LINKED ISSUE #9442: Metrics "Under Filesystem Read" does not update
**Alluxio Version:**
v2.0.0

**Describe the bug**

**To Reproduce**
```bash
$ bin/alluxio-start.sh local -f
$ bin/alluxio fsadmin report metrics
Total IO: 
    Short-circuit Read                                         0B
    Short-circuit Read (Domain Socket)                         0B
    From Remote Instances                                      0B
    Under Filesystem Read                                      0B
    Alluxio Write                                              0B
    Alluxio Write (Domain Socket)                              0B
    Under Filesystem Write                                     0B

Total IO Throughput (Last Minute): 
    Short-circuit Read                                         0B
    Short-circuit Read (Domain Socket)                         0B
    From Remote Instances                                      0B
    Under Filesystem Read                                      0B
    Alluxio Write                                              0B
    Alluxio Write (Domain Socket)                              0B
    Under Filesystem Write                                     0B

$ bin/alluxio fs cat /sample_tweets_1m.csv > /dev/null

$ bin/alluxio fsadmin report metrics
Total IO: 
    Short-circuit Read                                         0B
    Short-circuit Read (Domain Socket)                         0B
    From Remote Instances                                      0B
    Under Filesystem Read                                      0B
    Alluxio Write                                              0B
    Alluxio Write (Domain Socket)                              0B
    Under Filesystem Write                                     0B

Total IO Throughput (Last Minute): 
    Short-circuit Read                                         0B
    Short-circuit Read (Domain Socket)                         0B
    From Remote Instances                                      0B
    Under Filesystem Read                                171.72KB
    Alluxio Write                                              0B
    Alluxio Write (Domain Socket)                              0B
    Under Filesystem Write                                     0B
```
**Expected behavior**

The total IO of Under Filesystem Read is never incremeted

**Urgency**
Median

**Additional context**

Similar issue: #9354

calvinjia:
Fixes #9442 

We need to consider tags in the uniqueness of our metrics, otherwise we may incorrectly override metrics (usually to 0) leading to inaccurate numbers.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileOutStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileOutStream.java
@@ -26,8 +26,8 @@ import alluxio.exception.ExceptionMessage;
 import alluxio.exception.PreconditionMessage;
 import alluxio.exception.status.UnavailableException;
 import alluxio.grpc.CompleteFilePOptions;
+import alluxio.metrics.ClientMetrics;
 import alluxio.metrics.MetricsSystem;
-import alluxio.metrics.WorkerMetrics;
 import alluxio.resource.CloseableResource;
 import alluxio.util.CommonUtils;
 import alluxio.util.FileSystemOptions;
@@ -315,7 +315,7 @@ public class FileOutStream extends AbstractOutStream {
   @ThreadSafe
   private static final class Metrics {
     private static final Counter BYTES_WRITTEN_UFS =
-        MetricsSystem.counter(WorkerMetrics.BYTES_WRITTEN_UFS);
+        MetricsSystem.counter(ClientMetrics.BYTES_WRITTEN_UFS);
 
     private Metrics() {} // prevent instantiation
   }
--- a/core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
+++ b/core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.annotation.concurrent.GuardedBy;
@@ -147,6 +148,9 @@ public class MetricsStore {
         double oldVal = oldMetric == null ? 0.0 : oldMetric.getValue();
         Metric newMetric = new Metric(metric.getInstanceType(), metric.getHostname(),
             metric.getMetricType(), metric.getName(), oldVal + metric.getValue());
+        for (Map.Entry<String, String> tag : metric.getTags().entrySet()) {
+          newMetric.addTag(tag.getKey(), tag.getValue());
+        }
         metricSet.removeByField(FULL_NAME_INDEX, metric.getFullMetricName());
         newMetrics.add(newMetric);
       } else {
--- a/core/server/master/src/test/java/alluxio/master/metrics/MetricsStoreTest.java
+++ b/core/server/master/src/test/java/alluxio/master/metrics/MetricsStoreTest.java
@@ -66,4 +66,18 @@ public class MetricsStoreTest {
         Metric.from("client.192_1_1_1:B.metric1", 15, MetricType.GAUGE)),
         mMetricStore.getMetricsByInstanceTypeAndName(MetricsSystem.InstanceType.CLIENT, "metric1"));
   }
+
+  @Test
+  public void putTaggedMetrics() {
+    String name = "test";
+    Metric metric1 =
+        new Metric(MetricsSystem.InstanceType.WORKER, "host", MetricType.COUNTER, name, 1.0);
+    metric1.addTag("Tag", "1");
+    Metric metric2 =
+        new Metric(MetricsSystem.InstanceType.WORKER, "host", MetricType.COUNTER, name, 2.0);
+    metric2.addTag("Tag", "2");
+    mMetricStore.putWorkerMetrics("host", Lists.newArrayList(metric1, metric2));
+    assertEquals(mMetricStore.getMetricsByInstanceTypeAndName(MetricsSystem.InstanceType.WORKER,
+        name), Sets.newHashSet(metric1, metric2));
+  }
 }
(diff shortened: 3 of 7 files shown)


core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -147,6 +148,9 @@ private static void putReportedMetrics(IndexedSet<Metric> metricSet, String inst
         double oldVal = oldMetric == null ? 0.0 : oldMetric.getValue();
         Metric newMetric = new Metric(metric.getInstanceType(), metric.getHostname(),
             metric.getMetricType(), metric.getName(), oldVal + metric.getValue());
+        for (Map.Entry<String, String> tag : metric.getTags().entrySet()) {
+          newMetric.addTag(tag.getKey(), tag.getValue());
+        }
```
ZacBlanco: So, if my understanding is correct, the extra tags contain values which the master will aggregate and display in the UI/Metrics report?

core/server/master/src/test/java/alluxio/master/metrics/MetricsStoreTest.java
```diff
@@ -66,4 +66,18 @@ public void putClientMetrics() {
...
+    Metric metric1 =
+        new Metric(MetricsSystem.InstanceType.WORKER, "host", MetricType.COUNTER, name, 1.0);
+    metric1.addTag("Tag", "1");
+    Metric metric2 =
+        new Metric(MetricsSystem.InstanceType.WORKER, "host", MetricType.COUNTER, name, 2.0);
+    metric2.addTag("Tag", "2");
+    mMetricStore.putWorkerMetrics("host", Lists.newArrayList(metric1, metric2));
+    assertEquals(mMetricStore.getMetricsByInstanceTypeAndName(MetricsSystem.InstanceType.WORKER,
+        name), Sets.newHashSet(metric1, metric2));
+  }
```
ZacBlanco: This test specifically tests that the added metrics contain the tags, but in light of fixing #9442, I think it would also be appropriate to write a new or test add assertions in an already existing test which makes sure UFS read/write values are updated.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
calvinjia: For UFS read/write specifically, the master will aggregate all the values regardless of the tags (debatable whether or not we want to expose per UFS metrics).

core/server/master/src/test/java/alluxio/master/metrics/MetricsStoreTest.java
calvinjia: Before we would not consider these two metrics as distinct. In the case of UFS metrics, that often meant we 0-ed them out.

ZacBlanco: Do you think that it isn't appropriate to add a test that verifies that UFS metrics are updated?

calvinjia: I think we should add the test, I was clarifying what this test validates.

core/client/fs/src/main/java/alluxio/client/file/FileOutStream.java
```diff
@@ -315,7 +315,7 @@ private void handleCacheWriteException(Exception e) throws IOException {
   @ThreadSafe
   private static final class Metrics {
     private static final Counter BYTES_WRITTEN_UFS =
-        MetricsSystem.counter(WorkerMetrics.BYTES_WRITTEN_UFS);
+        MetricsSystem.counter(ClientMetrics.BYTES_WRITTEN_UFS);
```
calvinjia: Note this is unrelated to the PR. However, this is require to make the unit test effective. Although we incorrectly track metrics from the worker, we also have client metrics in our test environment which remedies the problem.

A separate question is whether or not we want to keep client metrics for operations which will be tracked by the server (this can be useful if the client itself is reporting to an external metrics server) and how to coordinate between the client and server to ensure metrics are not double counted.

ZacBlanco: If my understanding I think `BYTES_WRITTEN_UFS` is only applicable on the client if the user is using an implementation of `LocalUnderFileSystem`(i.e. local development or NFS)? Otherwise, the client will always need to send the bytes through the worker for writing to the UFS.

For this test, if you set ` alluxio.user.short.circuit.enabled=false` can we get away without this change? 

For the most part I feel `BYTES_WRITTEN_UFS` doesn't make sense to keep track of as a client metric.

calvinjia: This tracks the number of bytes we write if we are using a sync persist mode. It will duplicate the metric (but without tags, which is why the test will pass with or without the actual fix).

tests/src/test/java/alluxio/client/rest/AlluxioMasterRestApiTest.java
```diff
@@ -110,11 +121,35 @@ public void getLostWorkers() throws Exception {
...
   }
 
+  @Test
+  public void getUfsMetrics() throws Exception {
+    int len = 100;
+    FileSystemTestUtils.createByteFile(mResource.get().getClient(), "/f1", WritePType.THROUGH, len);
+    CommonUtils.waitFor("Metrics to be updated correctly", () -> {
+      try {
+        return FormatUtils.getSizeFromBytes(len).equals(getMetrics(NO_PARAMS)
+            .get("totalBytesWrittenUfs"));
```
ZacBlanco: I tried to find a constant with this name somewhere in the codebase but didn't have any luck. It would probably be better than hard-coding the string in this test.

calvinjia: Yeah its based on the variable name in `MasterWebUIMetrics` (we convert this to a json blob). Did you want a variable in the test or across the codebase, afaik we don't use the string form in other places.

ZacBlanco: I think it would be nice if we used some kind of constant variable with the name across the codebase. But I think if we do that for the total "totalBytesWrittenUfs" we should apply it to other metrics in the JSON blob. I would say just don't worry about it.

I was thinking in MasterWebUIMetrics would have a set of constant variables representing the names in the JSON blob, so tha

```java
// in MasterWebUIMetrics.java
public static final String TOTAL_BYTES_WRITTEN_UFS = "totalBytesWrittenUfs";

// in AlluxioMasterRestApiTest.java
then in the test you could write
FormatUtils.getSizeFromBytes(len).equals(getMetrics(NO_PARAMS)
            .get(MasterWebUIMetrics.TOTAL_BYTES_WRITTEN_UFS));
```

calvinjia:

PR #9653, opened 2019-08-06
Add integration test for growing embedded journal cluster

ggezer:


ggezer:
@LuQQiu PTAL. Thanks!

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTest.java
```diff
@@ -149,6 +162,120 @@ public void resizeCluster() throws Exception {
...
+    fs.createDirectory(testDir);
+    Assert.assertTrue(fs.exists(testDir));
+
+    // Validate current quorum size.
+    Assert.assertEquals(2,
+        mCluster.getJournalMasterClientForMaster().getQuorumInfo().getServerInfoList().size());
+
+    // Create and start a new master to join to existing cluster.
+    // Get new master address.
+    MasterNetAddress newMasterAddress = new MasterNetAddress(
```
LuQQiu: Is it possible to implement the create new master logics in the MultiProcessCluster and expose related methods? Since embedded journal is the default journal type now, we may use the create new master logics in more tests.

ggezer: I tried it and it's basically requires a complete rewrite of `MultiProcessCluster`.
This PR hack it by making it accept external address, so it's aware of external masters.

ggezer:
@gpang PTAL. Thanks!

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -497,6 +497,16 @@ public synchronized void updateMasterConf(PropertyKey key, @Nullable String valu
     mMasters.forEach(master -> master.updateConf(key, value));
   }
 
+  /**
+   * Updates internal master list with an external address.
```
gpang: are there any guidelines on when to use this method? This seems somewhat hack-y.

ggezer: It's used if the cluster contains more masters than it itself created.
There is no scenario other than this test to use that method.

ggezer: As I mentioned, `MultiProcessCluster` implementation is too complex to support this scenario without. The use of internal master list is not consistent and it assumes cluster size is fixed when managing the masters, but it assumes variable length when creating clients/contexts/etc.. So that's why I think extending internal master list is not as hacky as it looks, as we are just updating the internal master list with the master created external to this cluster.

gpang: Ok, that sounds good. Could you maybe add a little bit of context in this javadoc? Like, when/how this method is useful?

gpang:

gpang:

PR #9655, opened 2019-08-06
Add flag to skip generating the webui when running generate tarball

### LINKED ISSUE #9594: can generate-tarballs exclude webui

tag v2.0.0

webui building cost much time

**Urgency**
Explain why the feature is important

**Additional context**

ns1123:
Fixes #9594

### DIFF
--- a/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
+++ b/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
@@ -40,7 +40,7 @@ var (
 	hadoopDistributionFlag string
 	targetFlag             string
 	mvnArgsFlag            string
-	skipUIFlag            bool
+	skipUIFlag             bool
 )
 
 func init() {
--- a/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
+++ b/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
@@ -40,7 +40,7 @@ var (
 	hadoopDistributionFlag string
 	targetFlag             string
 	mvnArgsFlag            string
-	buildUIFlag               string
+	skipUIFlag            bool
 )
 
 func init() {
@@ -49,7 +49,7 @@ func init() {
 		fmt.Sprintf("an optional target name for the generated tarball. The default is alluxio-%v.tar.gz. The string %q will be substituted with the built version. "+
 			`Note that trailing ".tar.gz" will be stripped to determine the name for the Root directory of the generated tarball`, versionMarker, versionMarker))
 	cmdSingle.Flags.StringVar(&mvnArgsFlag, "mvn-args", "", `a comma-separated list of additional Maven arguments to build with, e.g. -mvn-args "-Pspark,-Dhadoop.version=2.2.0"`)
-	cmdSingle.Flags.StringVar(&buildUIFlag, "build-ui", "true", `set this flag to false to skip building the webui`)
+	cmdSingle.Flags.BoolVar(&skipUIFlag, "skip-ui", false, `set this flag to skip building the webui`)
 }
 
 func single(_ *cmdline.Env, _ []string) error {
@@ -293,7 +293,7 @@ func generateTarball(hadoopClients []string) error {
 	}
 
 	mvnArgs := getCommonMvnArgs(hadoopVersion)
-	if buildUIFlag == "false" {
+	if skipUIFlag {
 		mvnArgsNoUI := append(mvnArgs, "-pl", "!webui")
 		run("compiling repo without UI", "mvn", mvnArgsNoUI...)
 	} else {
@@ -325,7 +325,7 @@ func generateTarball(hadoopClients []string) error {
 	run("adding Alluxio FUSE jar", "mv", fmt.Sprintf("integration/fuse/target/alluxio-integration-fuse-%v-jar-with-dependencies.jar", version), filepath.Join(dstPath, "integration", "fuse", fmt.Sprintf("alluxio-fuse-%v.jar", version)))
 	run("adding Alluxio checker jar", "mv", fmt.Sprintf("integration/checker/target/alluxio-checker-%v-jar-with-dependencies.jar", version), filepath.Join(dstPath, "integration", "checker", fmt.Sprintf("alluxio-checker-%v.jar", version)))
 
-	if buildUIFlag != "false" {
+	if !skipUIFlag {
 		masterWebappDir := "webui/master"
 		run("creating webui master webapp directory", "mkdir", "-p", filepath.Join(dstPath, masterWebappDir))
 		run("copying webui master webapp build directory", "cp", "-r", filepath.Join(masterWebappDir, "build"), filepath.Join(dstPath, masterWebappDir))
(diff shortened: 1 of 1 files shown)


dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -48,6 +49,7 @@ func init() {
 		fmt.Sprintf("an optional target name for the generated tarball. The default is alluxio-%v.tar.gz. The string %q will be substituted with the built version. "+
 			`Note that trailing ".tar.gz" will be stripped to determine the name for the Root directory of the generated tarball`, versionMarker, versionMarker))
 	cmdSingle.Flags.StringVar(&mvnArgsFlag, "mvn-args", "", `a comma-separated list of additional Maven arguments to build with, e.g. -mvn-args "-Pspark,-Dhadoop.version=2.2.0"`)
+	cmdSingle.Flags.StringVar(&buildUIFlag, "build-ui", "true", `set this flag to false to skip building the webui`)
```
ZacBlanco: I would make this a boolean instead of a string var. Then we can simply call the flag `skip-build-ui` or `skip-ui` and all they have to do is provide it without an argument.

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -40,6 +40,7 @@ var (
 	hadoopDistributionFlag string
 	targetFlag             string
 	mvnArgsFlag            string
+	buildUIFlag               string
```
ZacBlanco: Please fix formatting

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
ns1123: Fixed.

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
ns1123: +1

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -291,7 +293,13 @@ func generateTarball(hadoopClients []string) error {
 	}
 
 	mvnArgs := getCommonMvnArgs(hadoopVersion)
-	run("compiling repo", "mvn", mvnArgs...)
+	if skipUIFlag {
+		mvnArgsNoUI := append(mvnArgs, "-pl", "!webui")
+		run("compiling repo without UI", "mvn", mvnArgsNoUI...)
```
ZacBlanco: is it really necessary to add this to the output string? Just seems bad to have the same command with a different string. If you really want it then,

```golang
var output
if skipUIFlag {
  output = "compiling repo without UI"
else {
  output = "compiling repo"
}
run(output, mvnArgs...)
```

ns1123: The mvnArgs are different in each case, so it's not just printing a different output. There's a second string array mvnArgsNoUI which is passed if skip-ui is set.

ZacBlanco: I think this would be better then?

```golang
var output
if skipUIFlag {
  mvnArgs = append(mvnArgs, "-pl", "!webui")
  output = "compiling repo without UI"
else {
  output = "compiling repo"
}
run(output, mvnArgs...)
```

ns1123: The issue is if you append the -pl !webui to the mvnArgs, then all the UFS modules attempt to build with those arguments as well. Since webui doesn't exist in the underfs module, it throws a reactor error. This was a clean way for me to only skip the webui when building the parent project without affecting the mvnArgs for subsequent modules.

ZacBlanco: I see. I didn't realize that the mvnArgs object was used down below

ZacBlanco:
Will verify and test on my local first

gpang:
@ns1123 wouldn't we want all the generated tarballs to include the webui?

ns1123:
@gpang one of the github issues (#9594) references some developers looking to cut down on the time to build a tarball for dockerized deployments. There might be a better way to do this, but this flag cuts down the build time from 4:30 to 2:00 minutes and could also be useful for quickly testing a fix in a non-local environment.

ns1123:
@gpang I tested with a noUI tarball. We can still run CLI commands and access the /metrics/json endpoint. All of our REST API's are serving, just the UI is unavailable.

gpang:
@ns1123 then I think this can make sense if we also change the tarball name to have "nowebui" in the name.

ns1123:
@gpang done.

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -48,6 +49,7 @@ func init() {
 		fmt.Sprintf("an optional target name for the generated tarball. The default is alluxio-%v.tar.gz. The string %q will be substituted with the built version. "+
 			`Note that trailing ".tar.gz" will be stripped to determine the name for the Root directory of the generated tarball`, versionMarker, versionMarker))
 	cmdSingle.Flags.StringVar(&mvnArgsFlag, "mvn-args", "", `a comma-separated list of additional Maven arguments to build with, e.g. -mvn-args "-Pspark,-Dhadoop.version=2.2.0"`)
+	cmdSingle.Flags.BoolVar(&skipUIFlag, "skip-ui", false, `set this flag to skip building the webui`)
```
gpang: can we include the implications of this? Something like `set this flag to skip building the webui which will speed up build times, but the web ui will not be available`

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
```diff
@@ -291,11 +293,20 @@ func generateTarball(hadoopClients []string) error {
...
+	} else {
+		run("compiling repo", "mvn", mvnArgs...)
+	}
+
 	// Compile ufs modules for the main build
 	buildModules(srcPath, "underfs", "hdfs", ufsModulesFlag, version, ufsModules, mvnArgs)
 
 	tarball := strings.Replace(targetFlag, versionMarker, version, 1)
+	if skipUIFlag {
+		tarball = strings.Replace(tarball, version, version+"-noUI", 1)
```
gpang: I'm concerned with this replace being potentially ambiguous, so should we do something like:
```golang
versionString := version
if skipUIFlag {
  versionString = versionString + "-noUI"
}
tarball := strings.Replace(targetFlag, versionMarker, versionString, 1)
```

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
ns1123: done

dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go
ns1123: done

ns1123:
Jenkins, test this please

ZacBlanco:

PR #9665, opened 2019-08-07
Implement ShimFileSystem

ggezer:
`ShimFileSystem` is an alternative to existing `AlluxioFileSystem` where foreign URIs are supported on client side. URIs that are sent to master as-is need to be translated to internal Alluxio URIs. This is done by a reverse lookup on master mount table. For `ShimFileSystem` to work properly, the file system that it has been attached to must be mounted on Alluxio.

Closes https://github.com/Alluxio/alluxio/issues/9231

### DIFF
--- a/core/base/src/main/java/alluxio/exception/ExceptionMessage.java
+++ b/core/base/src/main/java/alluxio/exception/ExceptionMessage.java
@@ -259,6 +259,7 @@ public enum ExceptionMessage {
       "Mount path {0} shadows an existing path {1} in the parent underlying filesystem"),
   MOUNT_READONLY("A write operation on {0} under a readonly mount point {1} is not allowed"),
   UFS_PATH_DOES_NOT_EXIST("Ufs path {0} does not exist"),
+  FOREIGN_URI_NOT_MOUNTED("Foreign URI: {0} is not found on Alluxio mounts."),
 
   // key-value
   KEY_VALUE_TOO_LARGE("Unable to put key-value pair: key {0} bytes, value {1} bytes"),
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -179,7 +179,6 @@ public final class FileSystemContext implements Closeable {
   public static FileSystemContext create(ClientContext clientContext) {
     FileSystemContext ctx = new FileSystemContext();
     ctx.init(clientContext, MasterInquireClient.Factory.create(clientContext.getClusterConf()));
-    ctx.mUriValidationEnabled = clientContext.getUriValidationEnabled();
     return ctx;
   }
 
@@ -235,6 +234,7 @@ public final class FileSystemContext implements Closeable {
     if (mMetricsEnabled) {
       MetricsHeartbeatContext.addHeartbeat(getClientContext(), masterInquireClient);
     }
+    mUriValidationEnabled = ctx.getUriValidationEnabled();
   }
 
   /**
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -393,7 +393,7 @@ public final class FileSystemContext implements Closeable {
   /**
    * @return {@code true} if URI validation is enabled
    */
-  public boolean getUriValidationEnabled() {
+  public synchronized boolean getUriValidationEnabled() {
     return mUriValidationEnabled;
   }
 
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -144,6 +144,9 @@ public final class FileSystemContext implements Closeable {
    */
   private volatile FileSystemContextReinitializer mReinitializer;
 
+  /** Whether to disable URI scheme validation for file systems using this context.  */
+  private boolean mDisableUriValidation = false;
+
   /**
    * Creates a {@link FileSystemContext} with a null subject.
    *
@@ -176,6 +179,7 @@ public final class FileSystemContext implements Closeable {
   public static FileSystemContext create(ClientContext clientContext) {
     FileSystemContext ctx = new FileSystemContext();
     ctx.init(clientContext, MasterInquireClient.Factory.create(clientContext.getClusterConf()));
+    ctx.mDisableUriValidation = clientContext.getDisableSchemeValidation();
     return ctx;
   }
 
@@ -384,6 +388,13 @@ public final class FileSystemContext implements Closeable {
    */
   public synchronized InetSocketAddress getMasterAddress() throws UnavailableException {
     return mMasterClientContext.getMasterInquireClient().getPrimaryRpcAddress();
+  }
+
+  /**
+   * @return {@code true} if URI validation is disabled
+   */
+  public boolean getDisableUriValidation() {
+    return mDisableUriValidation;
   }
 
   private FileSystemMasterClient acquireMasterClient() {
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -144,8 +144,8 @@ public final class FileSystemContext implements Closeable {
    */
   private volatile FileSystemContextReinitializer mReinitializer;
 
-  /** Whether to disable URI scheme validation for file systems using this context.  */
-  private boolean mDisableUriValidation = false;
+  /** Whether to do URI scheme validation for file systems using this context.  */
+  private boolean mUriValidationEnabled = true;
 
   /**
    * Creates a {@link FileSystemContext} with a null subject.
@@ -179,7 +179,7 @@ public final class FileSystemContext implements Closeable {
   public static FileSystemContext create(ClientContext clientContext) {
     FileSystemContext ctx = new FileSystemContext();
     ctx.init(clientContext, MasterInquireClient.Factory.create(clientContext.getClusterConf()));
-    ctx.mDisableUriValidation = clientContext.getDisableSchemeValidation();
+    ctx.mUriValidationEnabled = clientContext.getUriValidationEnabled();
     return ctx;
   }
 
@@ -391,10 +391,10 @@ public final class FileSystemContext implements Closeable {
   }
 
   /**
-   * @return {@code true} if URI validation is disabled
+   * @return {@code true} if URI validation is enabled
    */
-  public boolean getDisableUriValidation() {
-    return mDisableUriValidation;
+  public boolean getUriValidationEnabled() {
+    return mUriValidationEnabled;
   }
 
   private FileSystemMasterClient acquireMasterClient() {
--- a/core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
@@ -114,7 +114,7 @@ public final class RetryHandlingFileSystemMasterClient extends AbstractMasterCli
       final CheckConsistencyPOptions options) throws AlluxioStatusException {
     return retryRPC(() -> {
       List<String> inconsistentPaths = mClient.checkConsistency(CheckConsistencyPRequest
-          .newBuilder().setPath(path.getPath()).setOptions(options).build())
+          .newBuilder().setPath(path.toString()).setOptions(options).build())
           .getInconsistentPathsList();
       List<AlluxioURI> inconsistentUris = new ArrayList<>(inconsistentPaths.size());
       for (String inconsistentPath : inconsistentPaths) {
@@ -129,7 +129,7 @@ public final class RetryHandlingFileSystemMasterClient extends AbstractMasterCli
       final CreateDirectoryPOptions options) throws AlluxioStatusException {
     retryRPC(
         () -> mClient.createDirectory(CreateDirectoryPRequest.newBuilder()
-            .setPath(path.getPath()).setOptions(options).build()),
+            .setPath(path.toString()).setOptions(options).build()),
         "CreateDirectory");
   }
 
@@ -138,7 +138,7 @@ public final class RetryHandlingFileSystemMasterClient extends AbstractMasterCli
       throws AlluxioStatusException {
     return retryRPC(
         () -> new URIStatus(GrpcUtils.fromProto(mClient.createFile(
-            CreateFilePRequest.newBuilder().setPath(path.getPath()).setOptions(options).build())
+            CreateFilePRequest.newBuilder().setPath(path.toString()).setOptions(options).build())
             .getFileInfo())), "CreateFile");
   }
 
@@ -146,20 +146,20 @@ public final class RetryHandlingFileSystemMasterClient extends AbstractMasterCli
   public void completeFile(final AlluxioURI path, final CompleteFilePOptions options)
       throws AlluxioStatusException {
     retryRPC(() -> mClient.completeFile(CompleteFilePRequest.newBuilder()
-        .setPath(path.getPath()).setOptions(options).build()), "CompleteFile");
+        .setPath(path.toString()).setOptions(options).build()), "CompleteFile");
   }
 
   @Override
   public void delete(final AlluxioURI path, final DeletePOptions options)
       throws AlluxioStatusException {
-    retryRPC(() -> mClient.remove(DeletePRequest.newBuilder().setPath(path.getPath())
+    retryRPC(() -> mClient.remove(DeletePRequest.newBuilder().setPath(path.toString())
         .setOptions(options).build()), "Delete");
   }
 
   @Override
   public void free(final AlluxioURI path, final FreePOptions options)
       throws AlluxioStatusException {
-    retryRPC(() -> mClient.free(FreePRequest.newBuilder().setPath(path.getPath())
+    retryRPC(() -> mClient.free(FreePRequest.newBuilder().setPath(path.toString())
         .setOptions(options).build()), "Free");
   }
 
@@ -173,7 +173,7 @@ public final class RetryHandlingFileSystemMasterClient extends AbstractMasterCli
   public URIStatus getStatus(final AlluxioURI path, final GetStatusPOptions options)
       throws AlluxioStatusException {
     return retryRPC(() -> new URIStatus(GrpcUtils
-        .fromProto(mClient.getStatus(GetStatusPRequest.newBuilder().setPath(path.getPath())
+        .fromProto(mClient.getStatus(GetStatusPRequest.newBuilder().setPath(path.toString())
             .setOptions(options).build()).getFileInfo())),
         "GetStatus");
   }
@@ -190,15 +190,14 @@ public final class RetryHandlingFileSystemMasterClient extends AbstractMasterCli
       throws AlluxioStatusException {
     return retryRPC(
         () -> mClient
-            .getNewBlockIdForFile(GetNewBlockIdForFilePRequest.newBuilder().setPath(path.getPath())
+            .getNewBlockIdForFile(GetNewBlockIdForFilePRequest.newBuilder().setPath(path.toString())
                 .setOptions(GetNewBlockIdForFilePOptions.newBuilder().build()).build())
             .getId(),
         "GetNewBlockIdForFile");
   }
 
   @Override
-  public Map<String, alluxio.wire.MountPointInfo> getMountTable()
-      throws AlluxioStatusException {
+  public Map<String, alluxio.wire.MountPointInfo> getMountTable() throws AlluxioStatusException {
     return retryRPC(() -> {
       Map<String, alluxio.wire.MountPointInfo> mountTableWire = new HashMap<>();
       for (Map.Entry<String, alluxio.grpc.MountPointInfo> entry : mClient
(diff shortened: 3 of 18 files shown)


core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
```diff
@@ -132,16 +132,16 @@ public void checkConsistency(CheckConsistencyPRequest request,
...
-              new AlluxioURI(path), CheckConsistencyContext.create(options.toBuilder()));
-          List<String> uris = new ArrayList<>(inconsistentUris.size());
-          for (AlluxioURI uri : inconsistentUris) {
-            uris.add(uri.getPath());
-          }
-          return CheckConsistencyPResponse.newBuilder().addAllInconsistentPaths(uris).build();
-        }, "CheckConsistency", "path=%s, options=%s", responseObserver, path, options);
+    RpcUtils.call(LOG, (RpcUtils.RpcCallableThrowsIOException<CheckConsistencyPResponse>) () -> {
+      List<AlluxioURI> inconsistentUris =
+          mFileSystemMaster.checkConsistency(mFileSystemMaster.translateUri(path),
```
calvinjia: Instead of doing this inline, could we do it as part of declaring `path`?

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
```diff
@@ -381,6 +382,36 @@ public Resolution resolve(AlluxioURI uri) throws InvalidPathException {
...
+        if (uriStr.startsWith(ufsUriStr)) {
+          return new AlluxioURI(
+              PathUtils.concatPath(mountEntry.getKey(), uriStr.substring(ufsUriStr.length())));
+        }
+      }
+      // No mount found for given path.
+      throw new InvalidPathException(
+          String.format("Foreign URI: %s is not found on Alluxio mounts.", uriStr));
+    }
+  }
```
calvinjia: Could you add tests for this method?

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
```diff
@@ -114,7 +114,7 @@ protected void afterConnect() {
       final CheckConsistencyPOptions options) throws AlluxioStatusException {
     return retryRPC(() -> {
       List<String> inconsistentPaths = mClient.checkConsistency(CheckConsistencyPRequest
-          .newBuilder().setPath(path.getPath()).setOptions(options).build())
+          .newBuilder().setPath(path.toString()).setOptions(options).build())
```
calvinjia: What are the new requirements for paths sent over the wire, is there any time we should use `getPath` now?

core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
ggezer: Done. I took the chance to do some refactoring that I wanted to do. 
In case it's confusing, I done below additionally:
- Simplify lambda syntax
- Unify RPC description by printing request only.

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
ggezer: It looks like `getPath()` is more efficient than `toString()`. So I wrote a helper to call `getPath()` when the URI has no scheme, which we assume to be Alluxio path. And call `toString()` when it's transparent URI.

core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
ggezer: Please note that, I don't translate every URI coming to this handler. Specifically for `mount` / `umount` / `start-sync` / `stop-sync` / `schedule-persist`.

If you see any problem with that within transparent URI scenario, I could still do the translation. However I don't expect to see non-Alluxio URIs coming in for these RPCs as they are not used by HDFS client.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -144,6 +144,9 @@
    */
   private volatile FileSystemContextReinitializer mReinitializer;
 
+  /** Whether to disable URI scheme validation for file systems using this context.  */
+  private boolean mDisableUriValidation = false;
```
apc999: nit: typically we use `enableUriValidation` or `uriValidationEnabled` rather than `disableUriValidation` in this codebase, to simplify the logic slightly.

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
```diff
@@ -312,4 +310,20 @@ public void updateUfsMode(final AlluxioURI ufsUri,
...
+
+  /**
+   * Gets the path that will be transported to master.
+   * Scheme-less URIs are assumed to be Alluxio paths,
+   * and getPath() is used to avoid string conversion.
+   *
+   * @param uri uri
+   * @return transport path
+   */
+  private String getTransportPath(AlluxioURI uri) {
```
apc999: It looks useful. How about make this a member method of `AlluxioURI` rather than a util method here (and if it stays here, I would even make it static method).

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -329,6 +332,13 @@ public boolean setReplication(Path path, short replication) throws IOException {
     }
   }
 
+  /**
+   * @param isShimFs set shim mode for this fs
+   */
+  public void setShimFs(boolean isShimFs) {
```
apc999: it is tricky to have the setter for `isShimFs`, which may not prevent users to change this bool flag after a `FileSystem` instance created. If you see the usage of this method, it is always after `super()`. I would then just make `mIsShimFs` a `final` member variable, and take this argument on constructor to prevent others to change it afterwards.

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
ggezer: https://github.com/Alluxio/alluxio/pull/9665/commits/ed9d1ca09901dfa3939ebdd1d0c2b71befbe8944

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
ggezer: I couldn't find a generic use for this in `AlluxioURI`. Made it static instead.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
ggezer: 57dd443ea7110da95b1d3be4add7597f40f7a1a5

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ggezer: b87bceb301b8afec00dee8cdc6872ca8b486ef65

ggezer:
@calvinjia @apc999 Updated the PR. PTAL.

ggezer:
Closing for now until Presto has a fix for hard-coded `fs.impl` directive.

core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioShimFileSystem.java
```diff
@@ -0,0 +1,57 @@
...
+   * @param conf Hadoop configuration
+   * @throws URISyntaxException if <code>uri</code> has syntax error
+   */
+  AlluxioShimFileSystem(final URI uri, final Configuration conf)
+      throws IOException, URISyntaxException {
+    super(uri, new FileSystem(), conf, Constants.SCHEME, false);
+  }
+
+  @Override
+  public int getUriDefaultPort() {
```
madanadit: how is this used?

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -745,4 +774,19 @@ private void ensureExists(AlluxioURI path) throws IOException {
...
   }
+
+  /**
+   * Returns original path if shim mode is enabled.
+   * Returns URI path component for when shim mode is not enabled.
+   *
+   * @param path the input path
+   * @return the Alluxio path
+   */
+  private String getAlluxioPath(Path path) {
```
madanadit: not sure I understand why both AbstractFileSystem and RetryHandlingFileSystemMasterClient need to convert paths to string representations. does the first strip out the scheme for an alluxio path only? whereas the second conversion is more of a performance optimization?

core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioShimFileSystem.java
ggezer: Don't know exactly. The shim wouldn't be different from existing `AlluxioFileSystem` though.

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
```diff
@@ -312,4 +310,20 @@ public void updateUfsMode(final AlluxioURI ufsUri,
             .setUfsPath(ufsUri.getRootPath()).setOptions(options).build()),
         "UpdateUfsMode");
   }
+
+  /**
+   * Gets the path that will be transported to master.
+   * Scheme-less URIs are assumed to be Alluxio paths,
+   * and getPath() is used to avoid string conversion.
```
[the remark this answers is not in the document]
ggezer: Moved last 2 lines to code section.
Don't we always put line break after the method comment?

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
ggezer: `AbstractFileSystem` will either pass the `Path` as full path or it'll strip scheme&authority from it. This is something that `AbstractFileSystem` knows based on whether it's a shim or not. 

`RetryHandlingFileSystemMasterClient` converts it to because paths are simple strings on transport layer. 

Also, `AbstractFileSystem` converts from `Path`, while `RetryHandlingFileSystemMasterClient` converts from `AlluxioURI`.

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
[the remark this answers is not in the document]
madanadit: i was talking about the early line break after `paths,` (2 second line before)

core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
ggezer: Discussed with @madanadit offline. I'll unify all these, here and at `RetryHandlingFileSystemMasterClient`

ggezer: Actually `mount/unmount` can't be modified as translating a mount path will simply fail. Now with these mandatory exclusions I feel it's OK to not manipulate paths for RPCs that are not accessed via hadoop. @madanadit ?

ggezer: we can still unify path manipulation logic in`RetryHandlingFileSystemMasterClient` though.

core/client/hdfs/src/main/java/alluxio/hadoop/ShimFileSystem.java
```diff
@@ -57,11 +62,28 @@ public String getScheme() {
...
+    return Constants.NO_SCHEME;
   }
 
   @Override
   protected boolean isZookeeperMode() {
     return mFileSystem.getConf().getBoolean(PropertyKey.ZOOKEEPER_ENABLED);
   }
+
+  @Override
+  // Gets the full path to send to Alluxio.
```
madanadit: nit: send to Alluxio server?

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -524,8 +513,29 @@ public synchronized void initialize(URI uri, org.apache.hadoop.conf.Configuratio
...
+  }
+
+  /**
+   * Validates given FS base URI for scheme and authority.
+   *
+   * @param fsUri FS Uri
+   * @throws IOException
+   * @throws IllegalArgumentException
+   */
+  protected void validateFsUri(URI fsUri) throws IOException, IllegalArgumentException {
```
madanadit: maybe move this to `alluxio.hadoop.FileSystem` instead of a default implementation?

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -109,8 +110,10 @@
 
   /**
    * Constructs a new {@link AbstractFileSystem} instance.
+   *
```
madanadit: nit: revert

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -109,8 +110,10 @@
...
    */
-  AbstractFileSystem() {}
+  AbstractFileSystem() {
```
madanadit: nit: revert

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
ggezer: Changed this further by @madanadit's feedback. https://github.com/Alluxio/alluxio/pull/9665/commits/cea90cf29590e4818a08ead26e6d9a90e13e57d2

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -474,16 +480,15 @@ public synchronized void initialize(URI uri, org.apache.hadoop.conf.Configuratio
 
     // HDFS doesn't allow the authority to be empty; it must be "/" instead.
     String authority = uri.getAuthority() == null ? "/" : uri.getAuthority();
-    mAlluxioHeader = getScheme() + "://" + authority;
+    // Use URI's scheme for shim mode.
+    String scheme = getScheme();
+    // Extract the scheme from URI if FS implementation is not bound to a scheme.
+    if (scheme.equals(Constants.NO_SCHEME)) {
```
apc999: It looks to me we are testing which class we are (`FileSystem` or `ShimFileSystem`), and then make decision to create `mAlluxioHeader`.  If this is the case, I think it will cleaner to move this logic further to the implementing class rather in the base Abstract class. In other words, the logic of the base Abstract class should not depend on `Constants.NO_SCHEME` which is the local information in the implementing class `ShimFileSystem`

core/client/hdfs/src/main/java/alluxio/hadoop/ShimFileSystem.java
```diff
@@ -0,0 +1,89 @@
...
+    return Constants.NO_SCHEME;
+  }
+
+  @Override
+  protected boolean isZookeeperMode() {
+    return mFileSystem.getConf().getBoolean(PropertyKey.ZOOKEEPER_ENABLED);
+  }
+
+  @Override
+  // Gets the full path to use when contacting Alluxio server.
```
apc999: nit:  don't put javadoc between `@Override` and method signature

core/client/hdfs/src/main/java/alluxio/hadoop/ShimFileSystem.java
```diff
@@ -0,0 +1,89 @@
...
+  protected String getAlluxioPath(Path path) {
+    return path.toString();
+  }
+
+  @Override
+  // Gets UFS path as native FS path.
+  protected Path getFsPath(URIStatus fileStatus) {
```
apc999: same

core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
```diff
@@ -216,6 +216,16 @@ long getNewBlockIdForFile(AlluxioURI path) throws FileDoesNotExistException, Inv
    */
   Map<String, MountPointInfo>  getMountTable();
 
+  /**
+   * Translates given URI to an Alluxio URI.
+   * If given URI does not contain a scheme, then it is regarded as Alluxio URI.
+   *
+   * @param uriStr uri
+   * @return URI in Alluxio namespace
+   * @throws InvalidPathException
```
apc999: complete this line?

core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
```diff
@@ -216,6 +216,16 @@ long getNewBlockIdForFile(AlluxioURI path) throws FileDoesNotExistException, Inv
...
+   */
+  AlluxioURI translateUri(String uriStr) throws InvalidPathException;
```
apc999: nit: `translateUri` is not very clear what are we translating the URI to. How about `reverseLookup` as you commented below?

core/server/master/src/test/java/alluxio/master/file/meta/MountTableTest.java
```diff
@@ -390,6 +391,35 @@ public void getMountInfo() throws Exception {
     Assert.assertEquals(null, mMountTable.getMountInfo(4L));
   }
 
+  /**
+   * Tests {@link MountTable#translate(String)}.
+   */
+  @Test
+  public void translate() throws Exception {
```
apc999: nit: split UT into smaller cases , one UT case for a single test case (e.g., successful translate, unknown translation). perhaps also the case without scheme (treated as alluxio path as documented in FSM interface)

core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioShimFileSystem.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.DelegateToFileSystem;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * The Alluxio implementation of Hadoop AbstractFileSystem. The implementation delegates to the
```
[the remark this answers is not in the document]
ggezer: I think we still need it as Shim is different than regular FileSytem. However I was not doing the delegating correctly in the constructor. Fixed that.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -502,9 +507,21 @@ public synchronized void initialize(URI uri, org.apache.hadoop.conf.Configuratio
     LOG.info("Initializing filesystem with connect details {}",
         Factory.getConnectDetails(alluxioConf));
 
-    mFileSystem = FileSystem.Factory.create(ClientContext.create(subject, alluxioConf));
+    // Create FileSystem for accessing Alluxio.
+    // Disable URI validation for non-Alluxio schemes.
+    mFileSystem = FileSystem.Factory.create(ClientContext.create(subject, alluxioConf)
+        .setUriValidationEnabled(uri.getScheme().equals(Constants.SCHEME)));
```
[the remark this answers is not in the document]
ggezer: nice catch. I think hadoop always sends with scheme but it's good to be prepared.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -745,4 +761,14 @@ private void ensureExists(AlluxioURI path) throws IOException {
...
     }
   }
+
+  /**
+   * Used to convert hadoop path to Alluxio path.
+   *
+   * @param path the input path
+   * @return the Alluxio path
+   */
+  protected String getAlluxioPath(Path path) {
```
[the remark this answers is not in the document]
ggezer: Agreed to all. Fixed.

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
ggezer: Agreed.

core/server/master/src/test/java/alluxio/master/file/meta/MountTableTest.java
ggezer: Actually, `MountTable#translate` requires foreign URI. So can't test scheme-less URIs here.
Split the UT and changed method name to reverseLookup which makes more sense for mount table.

core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
ggezer: There is actually a translation here because it supports both Alluxio and non-Alluxio URIs. Updated the comment for clarity. 

MountTable#translate on the other hand doesn't do translation but do strict reverse lookup. Changed method name there.

core/base/src/main/java/alluxio/exception/ExceptionMessage.java
```diff
@@ -259,6 +259,7 @@
       "Mount path {0} shadows an existing path {1} in the parent underlying filesystem"),
   MOUNT_READONLY("A write operation on {0} under a readonly mount point {1} is not allowed"),
   UFS_PATH_DOES_NOT_EXIST("Ufs path {0} does not exist"),
+  FOREIGN_URI_NOT_MOUNTED("Foreign URI: {0} is not found on Alluxio mounts."),
```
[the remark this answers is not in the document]
ggezer: It's used all over `MountTableTest`. I merely choose to be consistent within the same file.

ggezer:
Jenkins, test this please

core/base/src/main/java/alluxio/exception/ExceptionMessage.java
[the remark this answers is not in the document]
apc999: That's fine for this PR to be consistent with other part of the file. I created an issue https://github.com/Alluxio/alluxio/issues/9788 to address this issue independently. I will leave this up to you.

apc999:

PR #9667, opened 2019-08-08
Refactor process launching in alluxio-start.sh

### LINKED ISSUE #9389: Provide a way to start Alluxio processes in foreground for process supervisors

In many Linux-based distributions such as Debian, Ubuntu, CentOS, RHEL, etc.. process supervisors like `systemd` are utilized to launch processes and track the life of a process, and automatically restart the process if it fails.

To function properly these process supervisors expect a program which runs in the foreground when launched in a terminal. This is because supervisors like `systemd` take care of daemonizing and tracking the status of the process.

Currently, our start script only starts processes via a command in the form`(nohup ${args} 2>&1 ${log_path}) &`. We provide no options which leaves out the `&` that forces the process into the background.

It would be useful to add an option to our `alluxio-start.sh` script that will force the Alluxio java process to run in the foreground. This would allow process supervisors to properly track Alluxio (job) masters, (job) workers, and possibly restart, or report when these processes crash or fail.

It might also be useful to provide a "default" set of system service files so that users can install Alluxio as a service through systemd.

None so far

**Urgency**

Low

**Additional context**

ZacBlanco:
This change abstracts out the logic for launching Alluxio processes.
For most users it is still recommended to use the alluxio-start.sh
script. The change here adds a new "alluxio-process" script which
is the launch point for any Alluxio server processes. When used
directly it launches the processes into the foreground of the
terminal using the bash built-in "exec". The reason why this is
needed is so that Alluxio can properly function under the
supervision of process supervisors such as SystemD, SysVInit,
and the userspace program supervisord.

Note that these changes preserve the function and arguments
of alluxio-start.sh in its entirety. This is an internal-logic refactor
which exposes the ability to start processes without using nohup.

The usage of alluxio-process is simple. Simply provide a process
name and the optional argument "-c" which will print the logging
output from the server to stdout.

In the future, the docker/K8s may be able to use these scripts instead
of the custom ones in the integration/docker folder. I am leaving that
as future work.

Closes #9389


ZacBlanco:
@Xenorith FYI

ZacBlanco:
jenkins, test this please

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
...
+
+  local logger_var_name
+  local logger_var_value
+  local console_logging_enabled
+  local process
+
+  process=${1}
+  shift
+
+  CONSOLE_LOG="false"
```
madanadit: where is this used?

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
...
+#
+
+set -e
+
+BIN=$(cd "$( dirname "$( readlink "$0" || echo "$0" )" )"; pwd)
+
+USAGE="Usage: alluxio-process <process name> [-c]
+
+alluxio-process launches one of the Alluxio processes in the foreground.
+By default it will only log to the configured log file. To configure the
```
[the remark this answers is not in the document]
ZacBlanco: The configured log file is in `${ALLUXIO_HOME}/conf/log4j.properties`

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
...
+  local process_java_opts_name
+  local main_class
+
+  process_java_opts_name=${1}
+  main_class=${2}
+  # Set the java opts variable equal to "ALLUXIO_JAVA_OPTS" if it isn't set yet
+  if [[ -z "${!process_java_opts_name}" ]]; then
+    eval ${process_java_opts_name}=${ALLUXIO_JAVA_OPTS}
+  fi
+  exec ${JAVA} -cp ${ALLUXIO_SERVER_CLASSPATH} ${!process_java_opts_name} "${main_class}" ${@:3}
```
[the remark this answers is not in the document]
ZacBlanco: It resolves the value based on the value stored in the variable - it is variable indirection: https://www.tldp.org/LDP/abs/html/bashver2.html#EX78

Another example

```console
$ BAR="Hello, World!"
$ FOO="BAR"
$ echo "${!FOO}"
Hello, World!
```

bin/alluxio-start.sh
```diff
@@ -242,38 +220,11 @@ start_master() {
   fi
 
   if [[ ${ALLUXIO_MASTER_SECONDARY} == "true" ]]; then
-    if [[ -z ${ALLUXIO_SECONDARY_MASTER_JAVA_OPTS} ]]; then
```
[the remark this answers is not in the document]
ZacBlanco: Most of the other processes used this kind of thing as well. I've added a which executed this conditional for all processes.

https://github.com/Alluxio/alluxio/pull/9667/files#diff-180f4e21375ebc2d3e7fece5e77d276eR49-R51

bin/alluxio-start.sh
```diff
@@ -192,21 +178,14 @@ start_job_masters() {
...
-   alluxio.worker.AlluxioJobWorker > ${ALLUXIO_LOGS_DIR}/job_worker.out 2>&1) &
+  (nohup ${BIN}/alluxio-process job_worker > ${ALLUXIO_LOGS_DIR}/job_worker.out 2>&1) &
   ALLUXIO_JOB_WORKER_JAVA_OPTS+=" -Dalluxio.job.worker.rpc.port=0 -Dalluxio.job.worker.web.port=0"
   local nworkers=${ALLUXIO_JOB_WORKER_COUNT:-1}
   for (( c = 1; c < ${nworkers}; c++ )); do
     echo "Starting job worker #$((c+1)) @ $(hostname -f). Logging to ${ALLUXIO_LOGS_DIR}"
-    (nohup ${JAVA} -cp ${CLASSPATH} \
-     ${ALLUXIO_JOB_WORKER_JAVA_OPTS} \
-     alluxio.worker.AlluxioJobWorker > ${ALLUXIO_LOGS_DIR}/job_worker.out 2>&1) &
+    (ALLUXIO_JOB_WORKER_JAVA_OPTS=${ALLUXIO_JOB_WORKER_JAVA_OPTS} \
```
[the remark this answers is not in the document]
ZacBlanco: In `alluxio-start.sh` there is an `nworkers` variable that gets set: https://github.com/Alluxio/alluxio/blob/6243193f3307a79c590d726ed76c0ddad148f7ad/bin/alluxio-start.sh#L183-L189

I set this here in order to pass down the Alluxio job worker port configuration to start the extra worker processes.

bin/alluxio-start.sh
```diff
@@ -221,9 +200,8 @@ start_logserver() {
     fi
 
     echo "Starting logserver @ $(hostname -f)."
-    (nohup "${JAVA}" -cp ${CLASSPATH} \
-     ${ALLUXIO_LOGSERVER_JAVA_OPTS} \
-     alluxio.logserver.AlluxioLogServer "${ALLUXIO_LOGSERVER_LOGS_DIR}" > ${ALLUXIO_LOGS_DIR}/logserver.out 2>&1) &
+    (ALLUXIO_LOGSERVER_LOGS_DIR="${ALLUXIO_LOGSERVER_LOGS_DIR}" \
```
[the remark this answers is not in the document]
ZacBlanco: similarly, the `ALLUXIO_LOGSERVER_LOGS_DIR` must be set in the `alluxio-process` script before being run. It is resolved and created beofrehand if necessary. I didn't want that logic inside of the `alluxio-process` script, so the value is simply passed down.

bin/alluxio-start.sh
```diff
@@ -221,9 +200,8 @@ start_logserver() {
...
+    nohup ${BIN}/alluxio-process logserver > ${ALLUXIO_LOGS_DIR}/logserver.out 2>&1) &
```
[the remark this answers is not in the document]
ZacBlanco: it was the same before. For the rest of the command we use `logserver`. I elected to keep the same name for consistency.

bin/alluxio-start.sh
```diff
@@ -242,38 +220,11 @@ start_master() {
...
-    contains "${ALLUXIO_MASTER_JAVA_OPTS}" "Xmx"
-    if [[ $? -eq 0 ]]; then
-      ALLUXIO_MASTER_JAVA_OPTS+=" -Xmx8g "
-    fi
-
     echo "Starting master @ $(hostname -f). Logging to ${ALLUXIO_LOGS_DIR}"
-    (nohup "${JAVA}" -cp ${CLASSPATH} \
-     ${ALLUXIO_MASTER_JAVA_OPTS} \
-     alluxio.master.AlluxioMaster > ${ALLUXIO_LOGS_DIR}/master.out 2>&1) &
+    (JOURNAL_BACKUP="${journal_backup}" nohup ${BIN}/alluxio-process master > ${ALLUXIO_LOGS_DIR}/master.out 2>&1) &
```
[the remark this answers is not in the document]
ZacBlanco: The `JOURNAL_BACKUP` variable is needed when starting a master: https://github.com/Alluxio/alluxio/pull/9667/files#diff-180f4e21375ebc2d3e7fece5e77d276eR57-R59

It adds in the Alluxio property. We set it here to make sure it gets passed down to the `alluxio-process` script

bin/alluxio-process
ZacBlanco: it isn't. Good catch

ZacBlanco:
jenkins, test this please

bin/alluxio-start.sh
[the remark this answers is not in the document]
madanadit: got it, there's no export so it isn't passed down w/o this.

bin/alluxio-start.sh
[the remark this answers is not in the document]
madanadit: but where does `${journal_backup}` gets set? seems independent of this PR though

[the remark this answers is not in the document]
ZacBlanco: https://github.com/Alluxio/alluxio/blob/6243193f3307a79c590d726ed76c0ddad148f7ad/bin/alluxio-start.sh#L354-L365

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
...
+      launch_proxy
+      ;;
+    *)
+      print_help 1
+      ;;
+  esac
+
+}
+
+main $@
```
apc999: newline needed

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env bash
```
apc999: `bin/alluxio-process.sh` to be consistent with other util scripts under bin/?

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
...
+"
+
+# Launches an Alluxio process
+#
+# Arguments:
+#         1: name of the *_JAVA_OPTS variable to use when launching
+#         2: the main class to launch
+#         3.... Additional arguments to the java process
+launch_process() {
+
```
apc999: nit: remove unnecessary empty lines, also for lin 55 and 152

bin/alluxio-process
```diff
@@ -0,0 +1,252 @@
...
+    -c                Provide this flag to also log directly to stdout
+
+    -h                Print this help text to stdout
+"
+
+# Launches an Alluxio process
+#
+# Arguments:
+#         1: name of the *_JAVA_OPTS variable to use when launching
```
apc999: it's uncommon to take the var name as the argument and use it using "reflection"?,
can we quote the list of opts to pass as one single arg?

bin/alluxio-process
ZacBlanco: I do my best when writing shell to follow google style when writing these script, and the guide is ambivalent about extensions

https://google.github.io/styleguide/shell.xml#File_Extensions

Additionally, the decision to leave the extension off of the script was intentional. The script results in the PID of the shell process being replaced with actual PID of the Alluxio process I felt it was appropriate to leave the extension off as this is script is more like a "real executable" than a script which kicks off some daemon processes (like `alluxio-start.sh`).


I also am curious about your opinion on the name. I just named it `alluxio-process` to start. But maybe `launch-process` or `alluxio-launch` could also work?

bin/alluxio-process
ZacBlanco: Variable indirection is also a well documented feature of bash variables. I would not call it "reflection". Is there a technical limitation, or something else you don't like about it?

Either way, without having the variable value of the variable passed as an arg, then the following conditional check would not be possible without duplication:

https://github.com/Alluxio/alluxio/blob/3c3f36b2360d47aed24c5e308e376b64a6fc873b/bin/alluxio-process#L51-L53

bin/alluxio-process
apc999: SG.
`launch-process` or `process-launcher`?

ZacBlanco: renamed to `launch-process`

bin/alluxio-process
ZacBlanco: After offline discussion, we decide that the behavior for setting all `ALLUXIO_*_JAVA_OPTS` variables is done within `./libexec/alluxio-config.sh` and it should be this script's responsibility to do so.

With that assumption, we can remove all of these checks before starting the processes. It has been documented within the script.

ZacBlanco:
jenkins, test this please

bin/launch-process
```diff
@@ -0,0 +1,282 @@
...
+#                                 directory where the logserver will
+#                                 store logs.
+launch_logserver() {
+  launch_process "${ALLUXIO_LOGSERVER_JAVA_OPTS}" \
+                 "alluxio.logserver.AlluxioLogServer" \
+                 "${ALLUXIO_LOGSERVER_LOGS_DIR}"
+}
+
+# returns 1 if "$1" contains "$2", 0 otherwise.
+contains() {
```
calvinjia: Define this at the top?

bin/launch-process
```diff
@@ -0,0 +1,282 @@
...
+launch_master() {
+  if [[ -n ${JOURNAL_BACKUP} ]]; then
+    ALLUXIO_MASTER_JAVA_OPTS+=" -Dalluxio.master.journal.init.from.backup=${JOURNAL_BACKUP}"
+  fi
+
+  # use a default Xmx value for the master
+  contains "${ALLUXIO_MASTER_JAVA_OPTS}" "Xmx"
+  if [[ $? -eq 0 ]]; then
+    ALLUXIO_MASTER_JAVA_OPTS+=" -Xmx8g "
+  fi
```
calvinjia: There is a new option introduced in #9697 , could you add that as well?

bin/launch-process
ZacBlanco: done

bin/launch-process
ZacBlanco: done

ZacBlanco:
jenkins, test this please

calvinjia:
alluxio-bot, check this please.

gpang:
This passed, but the bot is not picking up the latest comment.

gpang:

PR #9668, opened 2019-08-08
Improve sync cache to consider recursive sync of ancestors

### LINKED ISSUE #9545: UfsSyncPathCache does not cache descendants when recursive is used
**Alluxio Version:**
alluxio version 2.0.0

**Describe the bug**
After running  `ls -Rf dir_path`, `UfsSyncPathCache` only cache `dir_path`, but not the file under `dir_path` recursively.

As a result, after setting `alluxio.user.file.metadata.sync.interval=3hour` and start MR job to read the files inside the dir, 
the method `shouldSyncPath` of `UfsSyncPathCache` will get `null` from `mCache`, and then alluxio will recursively load the metadata once again for every file in the `dir_path`. It will process slowly.

**Expected behavior**
`UfsSyncPathCache` uses guava cache to cache time for path.
When `ls -Rf dif_path`, `UfsSyncPathCache` will cache the dir_path not for every recursive file,
and alluxio really has the metadata for the dir_path, but not in the `UfsSyncPathCache`.
If using the longest prefix match to get the key，and get time from cache with the lpm key, It will achieve the same effect. This way can also save memory for alluxio.

**Urgency**
very urgency

**Additional context**
Related method calls
1. DefaultFileSystemMaster.syncMetadataInternal()
2. UfsSyncPathCache.notifySyncedPath()
3. UfsSyncPathCache.shouldSyncPath

gpang:
fixes #9545

This enables the ufs sync cache to inspect the ancestors of the path. This allows the sync to be skipped if an ancestor was recently sync recursively. This will potentially increase the time to check if sync should be performed, since the path's parents have to be computed.


gpang:
@yuzhu please take a look

tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
```diff
@@ -605,6 +606,73 @@ public void recursiveSync() throws Exception {
...
+    writeUfsFile(ufsPath(fileB), 1);
+
+    FileSystemMasterCommonPOptions longinterval =
+        FileSystemMasterCommonPOptions.newBuilder().setSyncIntervalMs(Constants.HOUR_MS).build();
+
+    // Should not exist, since no loading or syncing
+    assertFalse(mFileSystem.exists(new AlluxioURI(alluxioPath(fileA)), ExistsPOptions.newBuilder()
+        .setCommonOptions(PSYNC_NEVER).build()));
+
+    try {
```
yuzhu: is the standard practice to use the `ExpectedException` rule here?

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
```diff
@@ -65,11 +78,62 @@ public boolean shouldSyncPath(String path, long intervalMs) {
...
+    SyncTime lastSync = mCache.getIfPresent(path);
+    if (lastSync != null && (System.currentTimeMillis() - lastSync.getLastSyncMs()) < intervalMs) {
+      // Sync is not necessary for this path.
+      return false;
+    }
+
+    // sync should be done on this path, but check all ancestors to determine if a recursive sync
+    // had been performed (to avoid a sync again).
+    String currPath = path;
+    while (!currPath.equals(AlluxioURI.SEPARATOR)) {
```
yuzhu: perhaps use `isRoot`? also i think we also need to check if root was actually synced as well.

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
```diff
@@ -65,11 +78,62 @@ public boolean shouldSyncPath(String path, long intervalMs) {
...
+      try {
+        currPath = PathUtils.getParent(currPath);
+        lastSync = mCache.getIfPresent(currPath);
+        if (lastSync != null
+            && (System.currentTimeMillis() - lastSync.getLastRecursiveSyncMs()) < intervalMs) {
```
yuzhu: because we initialize `syncMs` and `recursiveSyncMs` to -1, it kind of works out here. but could we make it more explicit? like if make this check in a method, and check if `syncMs` or `recursiveSyncMs` is less than 0, we fail the check. because technically `intervalMs` can be set to a very large number, larger than `currentTimeMillis`.

tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
gpang: It was, but then it was relaxed at a later time. I don't really like that rule, since you can only use 1 expected exception per method.

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
gpang: Well, `path` comes in as a `String` and I didn't want to keep converting to and from `AlluxioURI`. I think that could be too expensive. Also, root will be checked, because within the loop, the parent of the path is checked. So, the root will be checked on the previous iteration.

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
gpang: Done

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
yuzhu: oh that makes sense

tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
yuzhu: sounds good

gpang:

liuzx32:
@gpang Thanks for the issue #9545, Great way!

liuzx32:
@gpang @yuzhu Think it over carefully, this PR solve cache problem for the cmd "ls -R" , but it doesn't solve cache problem for the cmd "ls". So, there will be some miscalculation sometimes.

yuzhu:
@liuzx32 Not sure I completely understand your concern here. In that case, since the `syncCache` has the path as the key, we are still able to look up the SyncTime right? it does not matter what the descendantType is.

liuzx32:
> @liuzx32 Not sure I completely understand your concern here. In that case, since the `syncCache` has the path as the key, we are still able to look up the SyncTime right? it does not matter what the descendantType is.

@yuzhu The case from MR job starting, not only for "ls -R". Because the process of MR job starting is slow for load metadata, users need to execute "ls -Rf" first. If the dir is the last layer of directories, we think "ls dir" is the same result but the logic from pr is not ok.

liuzx32:
@yuzhu @gpang I add a variable parentLevel at https://github.com/Alluxio/alluxio/pull/9714.

Issue #9648, opened 2019-08-06
startSync failed


sunny19930321:
ALLUXIO ENVIRONMENT:
(1) ALLUXIO 2.0.0
(2) HADOOP-2.8.2(HDFS)
(3) JAVA 1.8
bin/alluxio fs startSync /syncdir

 ActiveSyncManager - Fatal error: Failed to apply syncpoint_path: "/syncdir"
Does anybody know why？

sunny19930321:
java.lang.RuntimeException: mountId for the syncPoint /user/liyang28 not found in the filterMap
        at alluxio.master.file.activesync.ActiveSyncManager.apply(ActiveSyncManager.java:427)
        at alluxio.master.file.activesync.ActiveSyncManager.applyAndJournal(ActiveSyncManager.java:256)
        at alluxio.master.file.DefaultFileSystemMaster.startSyncAndJournal(DefaultFileSystemMaster.java:3734)
        at alluxio.master.file.DefaultFileSystemMaster.startSync(DefaultFileSystemMaster.java:3757)
        at alluxio.master.file.FileSystemMasterClientServiceHandler.lambda$startSync$17(FileSystemMasterClientServiceHandler.java:400)
        at alluxio.RpcUtils.call(RpcUtils.java:193)
        at alluxio.RpcUtils.call(RpcUtils.java:166)
        at alluxio.master.file.FileSystemMasterClientServiceHandler.startSync(FileSystemMasterClientServiceHandler.java:399)
        at alluxio.grpc.FileSystemMasterClientServiceGrpc$MethodHandlers.invoke(FileSystemMasterClientServiceGrpc.java:2126)
        at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:171)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
        at io.grpc.ForwardingServerCallListener$SimpleForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:40)
        at alluxio.security.authentication.ClientIpAddressInjector$1.onHalfClose(ClientIpAddressInjector.java:52)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
[...]

PR #9670, opened 2019-08-08
Add more debugging to activeSync startSync and stopSync

yuzhu:
Currently when a user's request to startSync and stopSync fails, very little information is printed because the exception printing happens after a fatal error. Adding this to help with debugging in the future.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3757,7 +3757,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
       try {
         mSyncManager.startSyncPostJournal(uri);
       } catch (Throwable e) {
-        LOG.warn("Start sync failed because {}", ExceptionUtils.getStackTrace(e));
+        LOG.warn("Start sync failed on {}", uri, e);
         // revert state;
         RemoveSyncPointEntry removeSyncPoint =
             File.RemoveSyncPointEntry.newBuilder()
@@ -3809,7 +3809,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
         long mountId = resolution.getMountId();
         mSyncManager.stopSyncPostJournal(lockedInodePath.getUri());
       } catch (Throwable e) {
-        LOG.warn("Stop sync failed because {}", ExceptionUtils.getStackTrace(e));
+        LOG.warn("Stop sync failed on {}", uri, e);
         // revert state;
         AddSyncPointEntry addSyncPoint =
             File.AddSyncPointEntry.newBuilder()
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -185,7 +185,6 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import io.grpc.ServerInterceptors;
 import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.exception.ExceptionUtils;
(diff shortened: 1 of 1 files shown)


yuzhu:
Currently when a user's request to startSync and stopSync fails, very little information is printed because the exception printing happens after a fatal error.  Adding this to help with debugging in the future. 

@gpang PTAL, thanks.

yuzhu:
adding debug information to help root causing issue #9648

yuzhu:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3756,6 +3757,7 @@ private void startSyncAndJournal(RpcContext rpcContext, AlluxioURI uri)
       try {
         mSyncManager.startSyncPostJournal(uri);
       } catch (Throwable e) {
+        LOG.info("Start sync failed because {}", ExceptionUtils.getStackTrace(e));
```
gpang: This should not be an `info` log, right? Also, will this be spammy?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3807,6 +3809,7 @@ private void stopSyncAndJournal(RpcContext rpcContext,
         long mountId = resolution.getMountId();
         mSyncManager.stopSyncPostJournal(lockedInodePath.getUri());
       } catch (Throwable e) {
+        LOG.info("Stop sync failed because {}", ExceptionUtils.getStackTrace(e));
```
gpang: This should not be an `info` log, right? Also, will this be spammy?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
yuzhu: this should not be spammy because startSync should rarely fail.
what level do you think is appropriate?  
I was actually debating between `warn` and `debug`, and eventually settled on `info`

gpang:
@yuzhu if these will not be spammy, then maybe `warn` is more appropriate than `info`

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3756,6 +3757,7 @@ private void startSyncAndJournal(RpcContext rpcContext, AlluxioURI uri)
       try {
         mSyncManager.startSyncPostJournal(uri);
       } catch (Throwable e) {
+        LOG.warn("Start sync failed because {}", ExceptionUtils.getStackTrace(e));
```
gpang: I think if you just add the exception as the last parameter, the logging library will log the exception and stack trace. Also, we should include the uri.

like `LOG.warn("Start sync failed on {}", uri, e); `

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3807,6 +3809,7 @@ private void stopSyncAndJournal(RpcContext rpcContext,
         long mountId = resolution.getMountId();
         mSyncManager.stopSyncPostJournal(lockedInodePath.getUri());
       } catch (Throwable e) {
+        LOG.warn("Stop sync failed because {}", ExceptionUtils.getStackTrace(e));
```
gpang: should include the uri, like:

`LOG.warn("Stop sync failed on {}", uri, e);`

yuzhu:
@gpang PTAL

LuQQiu:
@gpang Can we merge this PR and cherry-pick to branch 2.0?

gpang:

gpang:

LuQQiu:
Thanks!

PR #9676, opened 2019-08-09
Improve sync cache to consider recursive sync of ancestors

### LINKED ISSUE #9545: UfsSyncPathCache does not cache descendants when recursive is used

gpang:
fixes #9545

This enables the ufs sync cache to inspect the ancestors of the path.
This allows the sync to be skipped if an ancestor was recently sync
recursively. This will potentially increase the time to check if sync
should be performed, since the path's parents have to be computed.

pr-link: Alluxio/alluxio#9668
change-id: cid-bd0cfe062c02c919585f5506bbe970d553dcc526

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3336,7 +3336,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
                     .mergeFrom(LoadMetadataPOptions.newBuilder().setCreateAncestors(true)
                         .setLoadDescendantType(GrpcUtils.toProto(syncDescendantType))));
 
-            mUfsSyncPathCache.notifySyncedPath(inodePath.getUri().getPath());
+            mUfsSyncPathCache.notifySyncedPath(inodePath.getUri().getPath(), syncDescendantType);
           } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


gpang:
@yuzhu the cherry-pick to 2.0 (no merge conflicts)

gpang:

PR #9680, opened 2019-08-09
Restore interrupted flag in DynamicResourcePool

calvinjia:
Follow up to #9584 

Restoring the interrupted state can be helpful to callers if they catch IOException and check the interrupted flag (ie. a retry handler).
We can reset the interrupted flag because we break out of the while loop afterward.


calvinjia:
@findepi @hustnn Thanks for catching this

calvinjia:

PR #9678, opened 2019-08-09
Allow getBlockInfo command to handle errors

gpang:
The getBlockInfo command may hit errors when contacting the block master or file system master. However, the command may still be able to print out useful information, even if some of the RPCs fail. Therefore, the RPC exceptions should be handled and ignored and print out any available information.

PR #9684, opened 2019-08-09
Allow getBlockInfo command to handle errors

gpang:
The getBlockInfo command may hit errors when contacting the block master
or file system master. However, the command may still be able to print
out useful information, even if some of the RPCs fail. Therefore, the
RPC exceptions should be handled and ignored and print out any available
information.

pr-link: Alluxio/alluxio#9678
change-id: cid-0deb4a77549f7b44bcef8827ccf74dee0614e0da


gpang:

PR #9692, opened 2019-08-10
Prevent excessive flushing while applying a backup

ggezer:
Use a buffer to batch more entries at a time to apply and journal, while applying a backup.

### DIFF
--- a/core/common/src/main/java/alluxio/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/PropertyKey.java
@@ -1012,6 +1012,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_BACKUP_ENTRY_BUFFER_COUNT =
+      new Builder(Name.MASTER_BACKUP_ENTRY_BUFFER_COUNT)
+          .setDefaultValue("10000")
+          .setDescription("How many journal entries to buffer during a back-up.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_BIND_HOST =
       new Builder(Name.MASTER_BIND_HOST)
           .setDefaultValue("0.0.0.0")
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/PropertyKey.java
```diff
@@ -1012,6 +1012,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_BACKUP_ENTRY_BUFFER_COUNT =
```
calvinjia: this sounds more like a batch size?

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -86,15 +90,48 @@ public void initFromBackup(InputStream is) throws IOException {
...
+      JournalEntry entry;
+      while (true) {
+        // Read next entry.
+        entry = reader.readEntry();
+        // Buffer next entry.
+        if (entry != null) {
+          entryBuffer.add(entry);
+        }
+        // Process the buffer if it is full or the input stream has no more entries.
+        if (entry == null || entryBuffer.size() >= bufferLimit) {
```
calvinjia: `entry == null` only when we hit the end of the stream right?

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -86,15 +90,48 @@ public void initFromBackup(InputStream is) throws IOException {
...
+            String masterName = JournalEntryAssociation.getMasterForEntry(bufferedEntry);
+            Master master = mastersByName.get(masterName);
+            // Apply entry to master.
+            master.processJournalEntry(bufferedEntry);
+            // Create journal context if master is seens the first time.
+            if (!contextMap.containsKey(masterName)) {
+              contextMap.put(masterName, master.createJournalContext());
+            }
+            // Apply entry to journal.
+            contextMap.get(masterName).append(bufferedEntry);
```
calvinjia: Can we assume all of these operations succeed? Otherwise it might be risky to apply the change to internal master state before the journal?

core/server/common/src/main/java/alluxio/master/BackupManager.java
ggezer: Correct.

core/server/common/src/main/java/alluxio/master/BackupManager.java
ggezer: If applying a backup throws, then masters won't come up.

gpang:

ggezer:
@gpang master already has this fix in a different form. This is back-ported from master.

gpang:
@ggezer o ok, LGTM

gpang:

Issue #9658, opened 2019-08-07
Failover Triggers an NPE because threads aren't shutdown correctly


ns1123:
**Alluxio Version:**
1.8.1

**Describe the bug**
When Alluxio is set up to use ZooKeeper for leadership election, during a failover we see the following error message.

2019-08-04 01:30:25,058 INFO  UfsJournalReader - Reading journal file hdfs://hanameservice:8020/alluxio/journal/LicenseMaster/v1/logs/0xd1-0xd2.
2019-08-04 01:30:25,070 INFO  UfsJournalReader - Reading journal file hdfs://hanameservice:8020/alluxio/journal/LicenseMaster/v1/logs/0xd2-0xd3.
2019-08-04 01:30:26,283 INFO  UfsJournalCheckpointThread - BlockMaster: Journal checkpoint thread started.
2019-08-04 01:30:26,283 INFO  UfsJournalCheckpointThread - CallHomeMaster: Journal checkpoint thread started.
2019-08-04 01:30:26,286 INFO  UfsJournalReader - Reading journal file hdfs://hanameservice:8020/alluxio/journal/BlockMaster/v1/checkpoints/0x0-0x22d6f62.
2019-08-04 01:30:26,304 INFO  UfsJournalLogWriter - Marking hdfs://hanameservice:8020/alluxio/journal/FileSystemMaster/v1/logs/0xcd4688b-0x7fffffffffffffff as complete with log entries within [215246987, 215252063).
2019-08-04 01:30:26,304 INFO  UfsJournalLogWriter - Completing log hdfs://hanameservice:8020/alluxio/journal/FileSystemMaster/v1/logs/0xcd4688b-0x7fffffffffffffff with next sequence number 215252063
2019-08-04 01:30:26,313 ERROR HeartbeatThread - Uncaught exception in heartbeat executor, Heartbeat Thread shutting down
java.lang.NullPointerException: Cannot call isRootId() before initializeRoot()
    at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:208)
    at alluxio.master.file.meta.InodeTree.isRootId(InodeTree.java:1168)
    at alluxio.master.file.meta.InodeTree.computePathForInode(InodeTree.java:441)
    at alluxio.master.file.meta.InodeTree.lockFullInodePath(InodeTree.java:376)
    at alluxio.master.file.LostFileDetector.heartbeat(LostFileDetector.java:48)
[...]

PR #9693, opened 2019-08-11
Handle interruptions on various heart beats

ggezer:
`HeartbeatExecutor#heaertbeat` should throw `InterruptedException` by contract. This PR enforces it for several implementations.

Fixes https://github.com/Alluxio/alluxio/issues/9658


ggezer:
These are the heartbeats that popped up. Also they are the only ones that we a chance to enforce interruptions. Others are single calls to fs master, so I depend on FSMaster to do the right cleanup.

gpang:

PR #9694, opened 2019-08-11
Manage thrift thread-pool and interrupt it during shutdown

ggezer:
Thrift server doesn't interrupt its `ExecutorService` during shutdown. This PR makes it use an external `ExecutorService` that we own, and interrupt it after shutting down Thrift.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
+++ b/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
@@ -50,6 +50,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 
@@ -102,6 +105,9 @@ public class AlluxioMasterProcess implements MasterProcess {
 
   /** The RPC server. */
   private TServer mThriftServer;
+
+  /** The ExecutorService for RPC server. */
+  protected ExecutorService mExecutorService;
 
   /** The start time for when the master started. */
   private final long mStartTimeMs = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


gpang:

Also, could you add a description?

ggezer:
@gpang This fix is not applicable to master.

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
```diff
@@ -382,16 +388,22 @@ protected void startServingRPCServer() {
     } catch (TTransportException e) {
       throw new RuntimeException(e);
     }
+    // create executor service for thrift.
+    // use shutdown timeout as keep-alive.
+    int stopTimeoutMs = (int) Configuration.getMs(PropertyKey.MASTER_THRIFT_SHUTDOWN_TIMEOUT);
+    mExecutorService = new ThreadPoolExecutor(mMinWorkerThreads,
```
gpang: is this similar to how thrift would have created its own executor service internally?

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
```diff
@@ -382,16 +388,22 @@ protected void startServingRPCServer() {
     } catch (TTransportException e) {
       throw new RuntimeException(e);
     }
+    // create executor service for thrift.
+    // use shutdown timeout as keep-alive.
+    int stopTimeoutMs = (int) Configuration.getMs(PropertyKey.MASTER_THRIFT_SHUTDOWN_TIMEOUT);
```
gpang: NIT: rename to `shutdownTimeoutMs`

ggezer: I think. stop would make more sense as it's what thrift calls it. Changed the doc.

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
ggezer: It's functionally equivalent to how thrift creates internally.

ggezer:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
```diff
@@ -406,7 +418,11 @@ protected void startServingRPCServer() {
   protected void stopServing() throws Exception {
     if (mThriftServer != null) {
       mThriftServer.stop();
+      // Thrift does not interrupt outstanding tasks.
+      // Call shutdownNow() to interrupt them.
+      mExecutorService.shutdownNow();
```
gpang: @ggezer  Is there any way to detect if all the threads are "done" or not?

calvinjia: Can we wait for a period of time to ensure all threads are stopped and the threadpool is no longer accepting tasks? If there are lingering threads still I think it is safe to crash.

ggezer: Thrift already calls shutdown and waits for a configured timeout before returning. Problem is that it doesn't to anything for lingering tasks. So this call will just make sure to terminate terminate them after grace period.

gpang: If there is a way to detect if threads are still alive, then we should "exit", right? That would ensure that we start the server without any lingering threads.

ggezer: @gpang Are you suggesting a second level of timeout? At the point we called `mExecutor.shutdownNow()` we already waited upto `stopTimeout` for outstanding tasks to finish. `shutdownNow` just interrupts if there any existing threads. After calling `shutdownNow` there is no way to block on completion. We could only see if there are any before invoking it.

gpang: @ggezer what about `awaitTermination` ? https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#awaitTermination(long,%20java.util.concurrent.TimeUnit)

would we be able to detect with that?

ggezer:
When I say thrift awaits after calling `shutdown`, it's by calling `awaitTermination`. After `shutdownNow` is called, `awaitTermination` returns immediately.

gpang:
@ggezer The `ThreadPoolExecutor#shutdownNow()` javadoc says: `This method does not wait for actively executing tasks to terminate.  Use awaitTermination to do that.` Doesn't that mean we would still have to call `awaitTermination()` after `shutdownNow()` to ensure running threads are completed?

gpang:

PR #9699, opened 2019-08-12
Improve ufs journal parsing efficiency

### LINKED ISSUE #9698: Improve ufs journal reader buffer usage
**Alluxio Version:**
2.1.0-SNAPSHOT

**Describe the bug**
We allocate unnecessary buffers when parsing JournalEntries through the UFS journal reader.

**To Reproduce**
Switch journal type to UFS and read a journal. 

**Expected behavior**
We use 1 buffer per Journal reader that expands on demand.

**Urgency**
Medium, this can cause us to allocate many more objects which can be problematic if the journal is large.

calvinjia:
Fixes #9698 

In the journal entry stream reader we keep a buffer to reuse for reading entries. Because this buffer might not be fully utilized, we cannot use protobuf's default `parseFrom(byte[])` and instead wrap the valid portion of the buffer in a ByteArrayStream. However this causes additional buffer allocations. We can instead use the lower level parser API to parse a slice of the existing buffer.

This does not affect the embedded journal implementation because we allocate a fresh buffer for each entry (exact size). Also I did not apply this optimization for the v0 journals as those are no longer in use aside for migrations.

The overall performance improvement is about 75% for a 10MB edit log file stored on local disk, this also reduces the number of new objects allocated which can reduce GC pressure.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/JournalEntryStreamReader.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/JournalEntryStreamReader.java
@@ -76,9 +76,7 @@ public class JournalEntryStreamReader implements Closeable {
           totalBytesRead);
       return null;
     }
-
-    JournalEntry entry = JournalEntry.parseFrom(new ByteArrayInputStream(mBuffer, 0, size));
-    return entry;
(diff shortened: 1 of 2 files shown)


calvinjia:

PR #9706, opened 2019-08-12
Use Concurrent Mark Sweep garbage collector for master by default

bf8086:
It is crucial to minimize full garbage collection pause in Alluxio master which could cause various issues including master fail over. This change switch to use Concurrent Mark Sweep garbage collector by default so it is less likely to enter a long full GC pause on master.

### DIFF
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -276,6 +276,11 @@ start_master() {
       ALLUXIO_MASTER_JAVA_OPTS+=" -XX:MetaspaceSize=256M "
     fi
 
+    # use a CMS garbage collector for the master
+    if ! [[ "${ALLUXIO_MASTER_JAVA_OPTS}" =~ (UseConcMarkSweepGC|UseSerialGC|UseG1GC|UseParallelGC|UseParallelOldGC|UseParNewGC) ]]; then
+      ALLUXIO_MASTER_JAVA_OPTS+=" -XX:+UseConcMarkSweepGC "
+    fi
+
     echo "Starting master @ $(hostname -f). Logging to ${ALLUXIO_LOGS_DIR}"
     (nohup "${JAVA}" -cp ${CLASSPATH} \
      ${ALLUXIO_MASTER_JAVA_OPTS} \
--- a/bin/launch-process
+++ b/bin/launch-process
@@ -120,6 +120,10 @@ launch_master() {
   if [[ $? -eq 0 ]]; then
     ALLUXIO_MASTER_JAVA_OPTS+=" -XX:MetaspaceSize=256M "
   fi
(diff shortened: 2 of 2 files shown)


bf8086:
@calvinjia PTAL.

bin/alluxio-start.sh
```diff
@@ -276,6 +276,11 @@ start_master() {
       ALLUXIO_MASTER_JAVA_OPTS+=" -XX:MetaspaceSize=256M "
     fi
 
+    # use a CMS garbage collector for the master
+    if ! [[ "${ALLUXIO_MASTER_JAVA_OPTS}" =~ (UseConcMarkSweepGC|UseSerialGC|UseG1GC|UseParallelGC|UseParallelOldGC|UseParNewGC) ]]; then
```
calvinjia: Maybe add something in our perf tuning docs about this? In case users want to use their own GC strategy, they can know to modify this line.

ZacBlanco: Shouldn't they just use alluxio-env.sh to set this java opt - and then the conditional won't add the CMS?

Also, is this something better set within `./libexec/alluxio-config.sh`?

ZacBlanco:
Additionally, should we consider tuning other parameters while putting this in?
From the page https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/cms.html:

> A concurrent collection also starts if the occupancy of the tenured generation exceeds an initiating occupancy (a percentage of the tenured generation). The default value for this initiating occupancy threshold is approximately 92%, but the value is subject to change from release to release. This value can be manually adjusted using the command-line option `-XX:CMSInitiatingOccupancyFraction=<N>`, where `<N>` is an integral percentage (0 to 100) of the tenured generation size.

Should we also set a default value for `-XX:CMSInitiatingOccupancyFraction`? 92 seems pretty high as a default.

bin/alluxio-start.sh
bf8086: Regarding the doc, I will push a separate PR for tuning GC strategies so the patch is not blocked on it.

Regarding where this logic should be, so far the `alluxio-config.sh` is more about Alluxio configuration properties while the `alluxio-env.sh` does the setting of JVM options. I will follow the convention for now but I am definitely open to discussion for potential refactor of the scripts.

bf8086:
> Additionally, should we consider tuning other parameters while putting this in?
> Should we also set a default value for `-XX:CMSInitiatingOccupancyFraction`? 92 seems pretty high as a default.
@zac yeah we should look further into all the parameters affecting the GC behavior. I created an issue https://github.com/Alluxio/alluxio/issues/9716 to track the effort.

bf8086:
Jenkins, test this please.

calvinjia:

PR #9709, opened 2019-08-13
Reduce spammy logs

gpang:
These log messages can be very spammy, so use the sampling logger.


gpang:
@calvinjia

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3666,8 +3668,8 @@ private void setAttributeSingleFile(RpcContext rpcContext, LockedInodePath inode
         try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
           UnderFileSystem ufs = ufsResource.get();
           if (ufs.isObjectStorage()) {
-            LOG.debug("setOwner/setMode is not supported to object storage UFS via Alluxio. "
-                + "UFS: " + ufsUri + ". This has no effect on the underlying object.");
+            SAMPLING_LOG.debug("setOwner/setMode is not supported to object storage UFS "
+                + "via Alluxio. UFS: {}. This has no effect on the underlying object.", ufsUri);
```
calvinjia: Should we still sample this since it is already at a debug level?

gpang: I see. In different versions, this line specifically was at `warn` so it would make sense to sample. I guess the question is, do we want this to be at warn and sample, or at debug and not sample?

gpang: I just moved this back to LOG, but it is still on debug.

calvinjia:

PR #9714, opened 2019-08-13
Improve parent dir check in UfsSyncPathCache

liuzx32:
**Description**
Improving cache logic for https://github.com/Alluxio/alluxio/pull/9668. 
Please refer https://github.com/Alluxio/alluxio/pull/9715
#
If the dir is the last layer of directories, the result of "ls" is the same with "ls -R", but the descendantType of "ls" is "ONE" not "ALL", and the path cache will be not the same. 

**Improvement**
Add variable parentLevel for parent dir check and skip the mLastRecursiveSyncMs for direct parent.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
@@ -89,12 +89,14 @@ public final class UfsSyncPathCache {
 
     // sync should be done on this path, but check all ancestors to determine if a recursive sync
     // had been performed (to avoid a sync again).
+    int parentLevel = 0;
     String currPath = path;
     while (!currPath.equals(AlluxioURI.SEPARATOR)) {
       try {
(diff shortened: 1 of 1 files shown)


liuzx32:
The two steps of shouldSyncInternal() can merge one in the while cycle with parentLevel, but the logic of two steps should be more clear.

yuzhu:
@liuzx32 Thanks for the PR. 
I think there are three cases we need to address. 
In all cases, we need to check its parent, its parent of parent... until root to see if it is in the cache with recursive set to true. 
In all cases, we need to check itself to see if it is in the cache with recursive set to false or true. . 
if the path is a file, we need to check its parent to see if it is in the cache with recursive set to false.
Example of the third case is /dir is in the cache with recursive set to false, /dir/file should have already been synced, but /dir/dir2 is not synced. 

The current PR does not quite address the third case, and it misses checking the immediate parent with the recursive set to true.

liuzx32:
@yuzhu Yeah, you are right! I will add a method for that.

PR #9715, opened 2019-08-13
Improve parent dir check in UfsSyncPathCache

liuzx32:
**Description**
Improving cache logic for PR #9668 #9714 .

**Improvement**
Add SyncTimeLevel to record time for the immediate parent dir, the intermediate parent dir, and the cmd "ls" or "ls -R".

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -738,7 +738,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
     Metrics.GET_FILE_INFO_OPS.inc();
     long opTimeMs = System.currentTimeMillis();
     LockingScheme lockingScheme =
-        createLockingScheme(path, context.getOptions().getCommonOptions(), LockPattern.READ);
+        createLockingScheme(path, context.getOptions().getCommonOptions(), LockPattern.READ, true);
     try (RpcContext rpcContext = createRpcContext();
          LockedInodePath inodePath = mInodeTree
              .lockInodePath(lockingScheme.getPath(), lockingScheme.getPattern());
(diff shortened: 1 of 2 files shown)


gpang:
@liuzx32 I think files and directories have to be treated differently. For example, if a file's parent was synced, then the file does not have to be synced. However, if a directory's parent was only synced (but not recursively), then that directory should be still be synced.

Also, I think this change can be simplified. Currently, there are 2 cases that are checked:
- check the path itself (if it was synced)
- check ancestors of the path (if they were recursively synced)

I think there needs to be just 1 more case:
- if the path is a file, check the direct parent (if it was synced in any way)

liuzx32:
> @liuzx32 I think files and directories have to be treated differently. For example, if a file's parent was synced, then the file does not have to be synced. However, if a directory's parent was only synced (but not recursively), then that directory should be still be synced.
> 
> Also, I think this change can be simplified. Currently, there are 2 cases that are checked:
> 
> * check the path itself (if it was synced)
> * check ancestors of the path (if they were recursively synced)
> 
> I think there needs to be just 1 more case:
> 
> * if the path is a file, check the direct parent (if it was synced in any way)

@gpang Yeah, however, if a directory's parent was only synced (but not recursively), then that directory should be still be synced. And the directory which just like a file should not be in the Cache.

liuzx32:
@gpang There are many cases need to be considered. If files and directories have to be treated differently, It will be more complex. Example of /dir1/dir2/dir3/*, we need to think about dir1, dir2,  dir3, and each one is recursive or not, not only for the direct parent. Just for cache simple data, we need to process complex sync logic.
#
I have thinked about checking the direct parent and commit the https://github.com/Alluxio/alluxio/pull/9714 which can't include other cases for example of the middle dir.

gpang:
> @gpang There are many cases need to be considered. If files and directories have to be treated differently, It will be more complex. Example of /dir1/dir2/dir3/*, we need to think about dir1, dir2, dir3, and each one is recursive or not, not only for the direct parent. Just for cache simple data, we need to process complex sync logic.
> 
> I have thinked about checking the direct parent and commit the #9714 which can't include other cases for example of the middle dir.

The current code already looks for the recursive flag for all ancestors of the path. So, for a path `/dir1/dir2/item`, it will already check for `/dir1/dir2/`, then `/dir1/`, then `/` to see if any of them were synced recursively. This should handle most cases.

However, the one case that this does not handle is for a file `/dir1/dir2/file` and `/dir1/dir2/` was already synced (not recursively). That is the 3rd case I was referring to earlier.

liuzx32:
@gpang, Yeah, the 3rd case is for a file /dir1/file and /dir1/dir2, but for /dir1/dir2, the shouldSync 
for "ls /dir1/dir2" should be true, for "getStatus of /dir1/dirs" should be false.

yuzhu:
@liuzx32 could you add some brief description of what you have done in this PR? I thought the only case we did not handle before this PR is the third case where we have to differentiate between directories and files. I don't understand why `adjustSyncTimeLevel` is necessary?  what specific cases are you addressing with this PR?  Thanks

apc999:
ping @liuzx32

PR #9719, opened 2019-08-14
Fix bugs related to the journal recovering from a UFS failure.

calvinjia:

1. If we fail to flush to the journal, do not try to recover, instead flag the stream as `needsRecovery` and handle the recovery on next operation
3. When recovering, create a new log file starting with the latest committed journal entry (instead of our current one)
4. When recovering, fail if the latest journal entry recovered from our buffer is not equal to our expected next sequence number - 1
5. When recovering, do not consider incomplete logs when inferring the last persisted sequence number from the file name (it will be `INTEGER_MIN`)


gpang:
does this not apply to 2.0/master branch?

ggezer:
@calvinjia LGTM. 
Should we target this fix for master branch?

calvinjia:
Good point, seems like this is also a problem in master. I will open another PR

Issue #9493, opened 2019-07-18
Improve the error message when there is no alluxio worker


ns1123:
**Alluxio Version:**
What version of Alluxio are you using?
Affects all current versions

**Describe the bug**
Currently, if there is no worker registered to the master, when the user performs a write operation, they see the below error.
"There is no worker with enough space for a new block of size {0}"

Similarly for reads, users see:
"Block {0} is unavailable in both Alluxio and UFS."

**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)
Start an Alluxio cluster with no workers and attempt a read/write operation

**Expected behavior**
A new error message should be thrown on any non-metadata operation clearly detailing that there are no workers registered to the cluster.

**Urgency**
Describe the impact and urgency of the bug.
High

**Additional context**

PR #9721, opened 2019-08-14
Improve error message when there is no worker

uronce9:
fixes Alluxio/alluxio#9493


gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
`alluxio.client.block.AlluxioBlockStoreTest` test fails.

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
```diff
@@ -170,24 +169,22 @@ public BlockInStream getInStream(long blockId, InStreamOptions options,
...
-      if (workerPool.isEmpty()) {
-        throw new UnavailableException(
-            "No Alluxio worker available. Check that your workers are still running");
-      }
     } else {
-      workerPool = locations.stream().map(BlockLocation::getWorkerAddress).collect(toSet());
-      if (workerPool.isEmpty()) {
-        throw new NotFoundException(
+      if (locations.isEmpty()) {
+        throw new UnavailableException(
```
gpang: `getEligibleWorkers` could be expensive, so maybe we should only do this in the specific error cases? For example, here, we could check all the workers to see if any exist. So, it could look something like:

```java
if (locations.isEmpty()) {
  if (getEligibleWorkers().isEmpty()) {
    throw new UnavailableException(ExceptionMessage.NO_WORKER_AVAILABLE.getMessage());
  }
  throw new UnavailableException(ExceptionMessage.BLOCK_UNAVAILABLE.getMessage(info.getBlockId()));
  workerPool = locations.stream().map(BlockLocation::getWorkerAddress).collect(toSet());
}
```

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
```diff
@@ -170,24 +169,22 @@ public BlockInStream getInStream(long blockId, InStreamOptions options,
...
+      throw new UnavailableException(ExceptionMessage.NO_WORKER_AVAILABLE.getMessage());
+    }
     // Initial target workers to read the block given the block locations.
     Set<WorkerNetAddress> workerPool;
     // Note that, it is possible that the blocks have been written as UFS blocks
     if (options.getStatus().isPersisted()
         || options.getStatus().getPersistenceState().equals("TO_BE_PERSISTED")) {
-      blockWorkerInfo = getEligibleWorkers();
       workerPool = blockWorkerInfo.stream().map(BlockWorkerInfo::getNetAddress).collect(toSet());
-      if (workerPool.isEmpty()) {
```
gpang: I think this previous check was fine, right? I am just concerned with calling `getEligibleWorkers` for every code path.

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
```diff
@@ -298,6 +295,9 @@ public BlockOutStream getOutStream(long blockId, long blockSize, WorkerNetAddres
    */
   public BlockOutStream getOutStream(long blockId, long blockSize, OutStreamOptions options)
       throws IOException {
+    if (getEligibleWorkers().isEmpty()) {
```
gpang: can we do this check when we know there are no locations? https://github.com/Alluxio/alluxio/blob/master/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java#L317

uronce9:
@gpang [getEligibleWorkers](https://github.com/Alluxio/alluxio/blob/master/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java#L120) will only fetch workers information lazily, when the workers information exists and it's not the time to refresh the information, then this is a noop.

gpang:
@uronce9 
Yes it is lazy, but the first call will definitely fetch it, right? So, if we are reading a MUST_CACHE file, in the old code, we would not call `getEligibleWorkers`, but in the new code, we would call `getEligibleWorkers`, and perform an RPC?

uronce9:
@gpang agree, i've updated the code, thanks.

gpang:

PR #9726, opened 2019-08-14
Fix bugs related to the journal recovering from a UFS failure

### LINKED ISSUE #9722: Address edge cases in UFS journal recovery in case of UFS failure
**Alluxio Version:**
2.1.0-SNAPSHOT

**Describe the bug**
When we fail to flush journal entries to the UFS, our internal state may become out of sync with the external journal files if using UFS journal.

**To Reproduce**
Fail to flush to the UFS and then successfully flush to the UFS.

**Expected behavior**
The journal files are repaired and operation resumes as usual after the UFS is available.

calvinjia:
Fixes #9722

1. If we fail to flush to the journal, do not try to recover, instead
flag the stream as needsRecovery and handle the recovery on next
operation
3. When recovering, create a new log file starting with the latest
committed journal entry (instead of our current one)
4. When recovering, fail if the latest journal entry recovered from our
buffer is not equal to our expected next sequence number - 1
5. When recovering, do not consider incomplete logs when inferring the
last persisted sequence number from the file name (it will be
INTEGER_MIN)

Cherry-pick of existing commit.

pr-link: Alluxio/alluxio#9723
change-id: cid-63623a4e96d91f9619605dd9cad1b7da7679ac9c

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
@@ -159,8 +159,11 @@ final class UfsJournalLogWriter implements JournalWriter {
     }
 
     long lastPersistSeq = recoverLastPersistedJournalEntry();
+    if (lastPersistSeq == -1) {
+      throw new RuntimeException("Cannot find any journal entry to recover from.");
+    }
 
-    createNewLogFile();
+    createNewLogFile(lastPersistSeq + 1);
     if (!mEntriesToFlush.isEmpty()) {
       JournalEntry firstEntryToFlush = mEntriesToFlush.peek();
       if (firstEntryToFlush.getSequenceNumber() > lastPersistSeq + 1) {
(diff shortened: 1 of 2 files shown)


core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
```diff
@@ -263,9 +277,13 @@ private void maybeRotateLog() throws IOException {
...
   private void createNewLogFile() throws IOException {
+    createNewLogFile(mNextSequenceNumber);
+  }
+
+  private void createNewLogFile(long startSequenceNumber) throws IOException {
     URI newLog = UfsJournalFile
-        .encodeLogFileLocation(mJournal, mNextSequenceNumber, UfsJournal.UNKNOWN_SEQUENCE_NUMBER);
-    UfsJournalFile currentLog = UfsJournalFile.createLogFile(newLog, mNextSequenceNumber,
+        .encodeLogFileLocation(mJournal, startSequenceNumber, UfsJournal.UNKNOWN_SEQUENCE_NUMBER);
+    UfsJournalFile currentLog = UfsJournalFile.createLogFile(newLog, startSequenceNumber,
```
calvinjia: Lines 280-283 and 285-286 are added in this patch because of 1.8 specific issues.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
```diff
@@ -159,8 +159,11 @@ private void maybeRecoverFromUfsFailures() throws IOException, JournalClosedExce
     }
 
     long lastPersistSeq = recoverLastPersistedJournalEntry();
+    if (lastPersistSeq == -1) {
+      throw new RuntimeException("Cannot find any journal entry to recover from.");
+    }
 
-    createNewLogFile();
+    createNewLogFile(lastPersistSeq + 1);
```
calvinjia: Line 166 is modified because of 1.8 specific issues.

calvinjia:

PR #9727, opened 2019-08-14
Fix bugs related to the journal recovering from a UFS failure

### LINKED ISSUE #9722: Address edge cases in UFS journal recovery in case of UFS failure

calvinjia:
Fixes #9722

1. If we fail to flush to the journal, do not try to recover, instead
flag the stream as needsRecovery and handle the recovery on next
operation
3. When recovering, create a new log file starting with the latest
committed journal entry (instead of our current one)
4. When recovering, fail if the latest journal entry recovered from our
buffer is not equal to our expected next sequence number - 1
5. When recovering, do not consider incomplete logs when inferring the
last persisted sequence number from the file name (it will be
INTEGER_MIN)

Cherry-pick of existing commit.

pr-link: Alluxio/alluxio#9723
change-id: cid-63623a4e96d91f9619605dd9cad1b7da7679ac9c

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
@@ -159,6 +159,9 @@ final class UfsJournalLogWriter implements JournalWriter {
     }
 
     long lastPersistSeq = recoverLastPersistedJournalEntry();
+    if (lastPersistSeq == -1) {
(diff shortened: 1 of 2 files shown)


calvinjia:

PR #9736, opened 2019-08-15
Check for updates

madanadit:
This PR introduces the ability to detect the latest version of Alluxio available and print a message in `master.out` if the version does not match the deployed version.


madanadit:
jenkins, test this please.

core/common/src/main/java/alluxio/util/EnvironmentUtils.java
```diff
@@ -0,0 +1,50 @@
...
+public final class EnvironmentUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(EnvironmentUtils.class);
+
+  /**
+   * Utility to detect the docker deployment environment.
+   *
+   * @return true, if running on docker
+   */
+  @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
+  public static boolean isDocker() throws IOException {
```
gpang: do we want these status methods to throw an exception?

core/transport/src/proto/journal/journal.proto
```diff
@@ -24,6 +24,7 @@ message JournalEntry {
   optional AsyncPersistRequestEntry async_persist_request = 16;
   optional BlockContainerIdGeneratorEntry block_container_id_generator = 3;
   optional BlockInfoEntry block_info = 4;
+  optional string cluster_id = 42;
```
gpang: should this be something like a "ClusterInfo" type, which contains the string `cluster_id` and potentially other fields in the future, like start time or something?

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -0,0 +1,78 @@
...
+    // Create the GET request.
+    Joiner joiner = Joiner.on("/");
+    String path = joiner.join("v0", "version");
+    String url = new URL(new URL(ProjectConstants.UPDATE_CHECK_HOST), path).toString();
+
+    HttpGet post = new HttpGet(url);
+    post.setHeader("User-Agent", getEnvString(clusterID));
+    post.setHeader("Authorization", "Basic " + ProjectConstants.UPDATE_CHECK_AUTH_STRING);
+
+    // Fire off the upload request.
```
gpang: NIT: not sure what the upload means, so maybe "fire off the version check request"

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * @param clusterID the cluster ID
+   * @return a string representation of the user's environment in the format "key1:value1, key2:
+   *         value2".
+   */
+  private static String getEnvString(String clusterID) throws IOException {
+    Joiner joiner = Joiner.on(",");
+    return joiner.join(
+        "ClusterID:" + clusterID,
+      "AlluxioVersion:" + ProjectConstants.VERSION,
```
gpang: is this indent wrong?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -147,6 +149,61 @@ public Address getFieldValue(MasterInfo o) {
...
+     * @param ctx the journal context
+     * @param entry the clusterID journal entry
+     */
+    public void applyAndJournal(java.util.function.Supplier<JournalContext> ctx, String entry) {
+      applyAndJournal(ctx, Journal.JournalEntry.newBuilder().setClusterId(entry).build());
+    }
+
+    @Override
+    public void resetState() {
+      mClusterID = null;
```
gpang: Should we have a constant id which represents the reset one (instead of using `null`)?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -238,6 +296,26 @@ public void start(Boolean isPrimary) throws IOException {
             ThreadFactoryUtils.build("DailyMetadataBackup-%d", true)), mUfs);
         mDailyBackup.start();
       }
+      if (ServerConfiguration.getBoolean(PropertyKey.MASTER_UPDATE_CHECK_ENABLED)) {
+        if (mState.getClusterID() == null) {
```
gpang: I feel like a cluster id is generally useful, so could this be independent from the version check?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -238,6 +296,26 @@ public void start(Boolean isPrimary) throws IOException {
...
+          try (JournalContext context = createJournalContext()) {
+            String clusterID = java.util.UUID.randomUUID().toString();
+            mState.applyAndJournal(context, clusterID);
+            LOG.info("Created new cluster ID {}", clusterID);
+            try {
+              String latestVersion = UpdateCheck.getLatestVersion(clusterID);
+              if (!latestVersion.equals(ProjectConstants.VERSION)) {
+                System.out.println("The latest version (" + latestVersion + ") is not the same"
+                    + "as the current version. To upgrade visit https://www.alluxio.io/download/.");
```
gpang: print out the current version too

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -0,0 +1,78 @@
...
+    String path = joiner.join("v0", "version");
+    String url = new URL(new URL(ProjectConstants.UPDATE_CHECK_HOST), path).toString();
+
+    HttpGet post = new HttpGet(url);
+    post.setHeader("User-Agent", getEnvString(clusterID));
+    post.setHeader("Authorization", "Basic " + ProjectConstants.UPDATE_CHECK_AUTH_STRING);
+
+    // Fire off the upload request.
+    HttpClient client = HttpClientBuilder.create().build();
+    HttpResponse response = client.execute(post);
```
gpang: what happens if the endpoint is unreachable? Shouldn't there be some timeout, so servers don't wait too long?

core/common/src/main/java/alluxio/util/EnvironmentUtils.java
madanadit: we catch all exceptions in the metamaster update check code. i feel that's better than hiding the exception and choosing a default

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
madanadit: sgtm

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
madanadit: sgtm

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
madanadit: yes

core/transport/src/proto/journal/journal.proto
madanadit: sgtm

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
madanadit: let me explicitly define the retry / timeout params. it does fail when the endpoint is unreachable.

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -0,0 +1,78 @@
...
+
+    // Check the response code.
+    int responseCode = response.getStatusLine().getStatusCode();
+    if (responseCode != HttpURLConnection.HTTP_OK) {
+      throw new IOException("Update check request failed with code: " + responseCode);
```
haoyuan: Does this mean that if the HTTP request failed, the user would see exception? If yes, suggest to make it silent. 

@apc999 @madanadit

madanadit: There will be a debug log level message in `master.log`

madanadit:
@gpang PTAL

madanadit:
jenkins, test this please.

core/common/src/main/java/alluxio/util/EnvironmentUtils.java
```diff
@@ -0,0 +1,50 @@
...
+    return new File("/.dockerenv").exists();
```
apc999: `alluxio.util.io.FileUtils#exists` which does not throw IOException ?

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -0,0 +1,90 @@
...
+
+    return EntityUtils.toString(response.getEntity(), "UTF-8");
+  }
+
+  /**
+   * @param clusterID the cluster ID
+   * @return a string representation of the user's environment in the format "key1:value1, key2:
+   *         value2".
+   */
+  private static String getEnvString(String clusterID) throws IOException {
```
apc999: `getUserAgentString`?

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -0,0 +1,90 @@
...
+    Joiner joiner = Joiner.on(",");
+    return joiner.join(
```
apc999: how about following format here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -229,15 +296,41 @@ public void start(Boolean isPrimary) throws IOException {
...
+      if (mState.getClusterID() == INVALID_CLUSTER_ID) {
+        try (JournalContext context = createJournalContext()) {
+          String clusterID = java.util.UUID.randomUUID().toString();
+          mState.applyAndJournal(context, clusterID);
+          LOG.info("Created new cluster ID {}", clusterID);
+        }
+        if (ServerConfiguration.getBoolean(PropertyKey.MASTER_UPDATE_CHECK_ENABLED)) {
+          try {
+            String latestVersion = UpdateCheck.getLatestVersion(mState.getClusterID(),
+                FormatUtils.parseTimeSize("60sec"),
```
gpang: why are we using `FormatUtils.parseTimeSize`. Can't we just have a constant long for the time? Also, all these timeouts are a minute. Are they too long? Would this mean startup could take up to 3 additional minutes in the worst case?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -229,15 +296,41 @@ public void start(Boolean isPrimary) throws IOException {
...
+                FormatUtils.parseTimeSize("60sec"),
+                FormatUtils.parseTimeSize("60sec")
+            );
+            if (!latestVersion.equals(ProjectConstants.VERSION)) {
+              System.out.println("The latest version (" + latestVersion + ") is not the same"
```
gpang: this will need a space at the end of the string

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
madanadit: reduced each of the timeouts to 3 secs

madanadit:
@gpang @apc999 PTAL. Thanks for the comments

madanadit:
jenkins, test this please.

madanadit:
jenkins, test this please.

madanadit:
@apc999 @gpang PTAL

madanadit:
jenkins, test this please.

core/server/common/src/main/java/alluxio/master/journal/JournalEntryAssociation.java
```diff
@@ -52,7 +52,8 @@ public static String getMasterForEntry(JournalEntry entry) {
         || entry.hasBlockInfo()) {
       return Constants.BLOCK_MASTER_NAME;
     }
-    if (entry.hasPathProperties()
+    if (entry.hasClusterInfo()
```
apc999: it seems 2.0.1 journal will break journal backwards compatibility by throwing `IllegalStateException`

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -492,18 +582,25 @@ public CheckpointName getCheckpointName() {
...
   public Iterator<Journal.JournalEntry> getJournalEntryIterator() {
-    return mPathProperties.getJournalEntryIterator();
+    return com.google.common.collect.Iterators.concat(mPathProperties.getJournalEntryIterator(),
+        mState.getJournalEntryIterator());
   }
 
   @Override
   public boolean processJournalEntry(Journal.JournalEntry entry) {
-    return mPathProperties.processJournalEntry(entry);
+    return mState.processJournalEntry(entry) || mPathProperties.processJournalEntry(entry);
```
apc999: is this change related to this PR?

madanadit: yes, its to process `clusterInfo` in `mState`

madanadit:
Note: Do NOT cherry-pick to branch-2.0. A companion PR () directed to branch

madanadit:

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -229,15 +295,39 @@ public void start(Boolean isPrimary) throws IOException {
...
+          String clusterID = java.util.UUID.randomUUID().toString();
+          mState.applyAndJournal(context, clusterID);
+          LOG.info("Created new cluster ID {}", clusterID);
+        }
+        if (Boolean.valueOf(ProjectConstants.UPDATE_CHECK_ENABLED)
+            && ServerConfiguration.getBoolean(PropertyKey.MASTER_UPDATE_CHECK_ENABLED)) {
+          try {
+            String latestVersion =
+                UpdateCheck.getLatestVersion(mState.getClusterID(), 3000, 3000, 3000);
+            if (!latestVersion.equals(ProjectConstants.VERSION)) {
```
apc999: `ProjectConstants.VERSION.equals(latestVersion)` in case `lastestVersion` is null?

madanadit:
@apc999 PTAL

apc999:
one minor comment left. Otherwise, LGTM

apc999:

PR #9737, opened 2019-08-16
Update launch-process contains function so the script does not exit

### LINKED ISSUE #9740: Some Java options may cause Alluxio to not start
**Alluxio Version:**
2.1.0-SNAPSHOT

**Describe the bug**

Recently a new script `./bin/launch-process` was added which offloads the actual launching of Alluxio processes from `./bin/alluxio-start.sh`. Within `launch-process` there is a `contains` function which checks whether the java options contain a specific argument. The options that are checked vary from process to process.

However, in the case where the java options already contain the value (i.e. when someone adds `-Xmx2G` in their  `conf/alluxio-env.sh`), then the `launch-process` script never actually launches the process.

The reason this happens is the `contains` function in `launch-process` calls `return 1` when the variable already contains the value that it is expecting. However, at the top of the script we use `set -e` to exit the script if any commands fail. The `return 1` from contains is interpreted as a failure by the bash interpreter and the script exits.

We can get around this by either removing `set -e` or changing how `contains` returns values.

**To Reproduce**

put the following in `conf/alluxio-env.sh`

```bash
ALLUXIO_MASTER_JAVA_OPTS+=" -Xmx2G "
```

Then try to start the master with `./bin/launch-process master`

**Expected behavior**

The master should start successfully 

**Urgency**

High, people may not be able to configure some options with `alluxio-env.sh`

**Additional context**

ZacBlanco:
The launch-process script uses the `set -e` flag which tells bash to exit
if a command returns non-zero; In the case when the "contains"
function returned "1", (meaning that arg 1 contains arg 2 somewhere within
it), the script would exit.

This means that any users which set values through alluxio-env.sh could
possibly have their processes fail to start if setting things like Xmx
and MetaspaceSize that are checked in either the master or workers.

To remediate this issue the contains function was modified to
instead, echo "1" or "0". Callers then simply have to call the
function in a subshell and use the result in their conditional
checks. This approach offers a better and more consistent usage
with the rest of our scripts.

Fixes #9740

### DIFF
--- a/bin/launch-process
+++ b/bin/launch-process
@@ -120,7 +120,7 @@ launch_master() {
   fi
   # use a default MetaspaceSize value for the master
   res="$(contains "${ALLUXIO_MASTER_JAVA_OPTS}" "XX:MetaspaceSize")"
-  if [[ $? -eq 0 ]]; then
+  if [[ "${res}" -eq "0" ]]; then
     ALLUXIO_MASTER_JAVA_OPTS+=" -XX:MetaspaceSize=256M "
   fi
 
@@ -132,7 +132,7 @@ launch_master() {
 launch_secondary_master() {
   # use a default Xmx value for the master
   local res="$(contains "${ALLUXIO_SECONDARY_MASTER_JAVA_OPTS}" "Xmx")"
-  if [[ $? -eq 0 ]]; then
+  if [[ "${res}" -eq "0" ]]; then
     ALLUXIO_SECONDARY_MASTER_JAVA_OPTS+=" -Xmx8g "
   fi
   launch_process "${ALLUXIO_SECONDARY_MASTER_JAVA_OPTS}" \
@@ -149,13 +149,13 @@ launch_job_master() {
 launch_worker() {
   # use a default Xmx value for the worker
   local res="$(contains "${ALLUXIO_WORKER_JAVA_OPTS}" "Xmx")"
-  if [[ $? -eq 0 ]]; then
+  if [[ "${res}" -eq "0" ]]; then
     ALLUXIO_WORKER_JAVA_OPTS+=" -Xmx4g "
   fi
 
   # use a default MaxDirectMemorySize value for the worker
   res="$(contains "${ALLUXIO_WORKER_JAVA_OPTS}" "XX:MaxDirectMemorySize")"
-  if [[ $? -eq 0 ]]; then
+  if [[ "${res}" -eq "0" ]]; then
     ALLUXIO_WORKER_JAVA_OPTS+=" -XX:MaxDirectMemorySize=4g "
   fi
 
@@ -189,7 +189,7 @@ launch_logserver() {
 
 # Launch the desired process
 main() {
-  if [[ "$#" -gt "2" || "${#}" -eq "0" ]]; then
+  if [[ "${#}" -gt "2" || "${#}" -eq "0" ]]; then
     print_help 1
   fi
 
--- a/bin/launch-process
+++ b/bin/launch-process
@@ -34,10 +34,16 @@ unless you know what you are doing.
     -h                Print this help text to stdout
 "
 
-# echos 1 if "$1" contains "$2", 0 otherwise.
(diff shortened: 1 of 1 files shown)


bin/launch-process
```diff
@@ -111,12 +114,12 @@ launch_master() {
...
-  contains "${ALLUXIO_MASTER_JAVA_OPTS}" "Xmx"
-  if [[ $? -eq 0 ]]; then
+  local res="$(contains "${ALLUXIO_MASTER_JAVA_OPTS}" "Xmx")"
+  if [[ "${res}" -eq "0" ]]; then
     ALLUXIO_MASTER_JAVA_OPTS+=" -Xmx8g "
   fi
   # use a default MetaspaceSize value for the master
-  contains "${ALLUXIO_MASTER_JAVA_OPTS}" "XX:MetaspaceSize"
+  res="$(contains "${ALLUXIO_MASTER_JAVA_OPTS}" "XX:MetaspaceSize")"
   if [[ $? -eq 0 ]]; then
```
gpang: This condition has to be updated?

bin/launch-process
```diff
@@ -128,7 +131,7 @@ launch_master() {
 # Launch a secondary master process
 launch_secondary_master() {
   # use a default Xmx value for the master
-  contains "${ALLUXIO_SECONDARY_MASTER_JAVA_OPTS}" "Xmx"
+  local res="$(contains "${ALLUXIO_SECONDARY_MASTER_JAVA_OPTS}" "Xmx")"
   if [[ $? -eq 0 ]]; then
```
gpang: This condition has to be updated?

bin/launch-process
```diff
@@ -145,13 +148,13 @@ launch_job_master() {
 # Launch a worker process
 launch_worker() {
   # use a default Xmx value for the worker
-  contains "${ALLUXIO_WORKER_JAVA_OPTS}" "Xmx"
+  local res="$(contains "${ALLUXIO_WORKER_JAVA_OPTS}" "Xmx")"
   if [[ $? -eq 0 ]]; then
```
gpang: This condition has to be updated?

bin/launch-process
```diff
@@ -145,13 +148,13 @@ launch_job_master() {
...
     ALLUXIO_WORKER_JAVA_OPTS+=" -Xmx4g "
   fi
 
   # use a default MaxDirectMemorySize value for the worker
-  contains "${ALLUXIO_WORKER_JAVA_OPTS}" "XX:MaxDirectMemorySize"
+  res="$(contains "${ALLUXIO_WORKER_JAVA_OPTS}" "XX:MaxDirectMemorySize")"
   if [[ $? -eq 0 ]]; then
```
gpang: This condition has to be updated?

bin/launch-process
ZacBlanco: done

bin/launch-process
ZacBlanco: done

bin/launch-process
ZacBlanco: done

bin/launch-process
ZacBlanco: done

ZacBlanco:
@gpang Tested with the most recent update and everything seems to work.

I found that with `echo`, a newline was appended to the output, which bash did not like when parsing the `${res}` values in the conditional. I could have used echo's `-n` option, but I've opted for `printf` which is an arguably more portable and safer alternative to echo as it does not rely on arguments which aren't in the UNIX specification

bin/launch-process
```diff
@@ -34,12 +34,15 @@ unless you know what you are doing.
     -h                Print this help text to stdout
 "
 
-# returns 1 if "$1" contains "$2", 0 otherwise.
+# echos 1 if "$1" contains "$2", 0 otherwise.
```
gpang: can you update this comment, and add a little bit more info as to why printf is required over echo?

ZacBlanco: done

gpang:
Jenkins, test this please

gpang:

PR #9738, opened 2019-08-16
Perform copycat connection building in a dedicated thread pool

ggezer:
Earlier, copycat catalyst `ThreadContext` was used to build connections for transport clients/servers. There are some cluster states that are managed by a single threaded `ThreadContext` and that could cause a node to loose contact with rest of cluster while a connection is being built. This PR offloads copycat connection building to a dedicated thread pool.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
 
 /**
  * Copycat transport {@link Client} implementation that uses Alluxio gRPC.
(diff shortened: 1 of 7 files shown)


ggezer:
@calvinjia PTAL. Thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcTransport.java
```diff
@@ -75,14 +81,16 @@ public CopycatGrpcTransport(AlluxioConfiguration clientConf, AlluxioConfiguratio
 
     mClients = new LinkedList<>();
     mServers = new LinkedList<>();
+    mExecutor = Executors
+        .newCachedThreadPool(ThreadFactoryUtils.build("copycat-transport-worker-%d", true));
```
calvinjia: Is cached threadpool the best choice, and can we share the executor between the client and server?

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcClient.java
```diff
@@ -86,16 +97,28 @@ public CopycatGrpcClient(AlluxioConfiguration conf, UserState userState) {
       } catch (Throwable e) {
         throw new RuntimeException(e);
       }
+    }, mExecutor);
+    // When connection is build, complete the connection future with it on a catalyst thread context
+    // for setting up the connection.
+    buildFuture.whenComplete((result, error) -> {
+      threadContext.execute(() -> {
```
calvinjia: What should we use the thread context executor for vs our own executor?

ggezer: End goal is to never use context executor within the transport layer for submitting work. Transport executor is for transport specific channel/connection building and termination.

core/server/common/src/main/java/alluxio/master/journal/raft/transport/CopycatGrpcTransport.java
ggezer: Since connection building/termination is only required at certain times such as boot-up/reconfigure/terminate, we want a thread pool that doesn't consume resources while we are not in those stages. So cached thread pool with a 0 core pool size is preferred. We could also put a max count to thread pool size by creating ThreadPoolExecutor directly. However I don't see how that would be necessary.

calvinjia:

Issue #9742, opened 2019-08-19
Fix inconsistent owner of file in Alluxio and HDFS


liuhongtong:
**Alluxio Version:**
2.0

**Describe the bug**
alluxio.user.file.writetype.default=CACHE_THROUGH
alluxio.security.login.impersonation.username=_HDFS_USER_
A new file‘s owner in Alluxio will be CurrentUser in UerGroupInformation, 
but the file's owner in HDFS will be User in UerGroupInformation.

gpang:
@liuhongtong Could you provide additional details on this issue? Specifically, what are the steps to reproduce the issue?

PR #9743, opened 2019-08-19
Fix inconsistent owner of file in Alluxio and HDFS #9742

liuhongtong:
Signed-off-by: liuhongtong <hongtongliu@126.com>

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
@@ -21,6 +21,7 @@ import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.PropertyKey;
 import alluxio.grpc.CreateFilePOptions;
 import alluxio.grpc.FileSystemMasterCommonPOptions;
+import alluxio.security.authentication.AuthenticationUserUtils;
 import alluxio.security.authorization.AccessControlList;
 import alluxio.security.authorization.Mode;
 import alluxio.util.CommonUtils;
@@ -130,7 +131,11 @@ public final class OutStreamOptions {
     mWriteTier = alluxioConf.getInt(PropertyKey.USER_FILE_WRITE_TIER_DEFAULT);
     mWriteType = alluxioConf.getEnum(PropertyKey.USER_FILE_WRITE_TYPE_DEFAULT, WriteType.class);
     try {
-      mOwner = context.getUserState().getUser().getName();
+      mOwner = AuthenticationUserUtils.getImpersonationUser(
+          context.getUserState().getSubject(), alluxioConf);
+      if (mOwner == null) {
+        mOwner = context.getUserState().getUser().getName();
+      }
       mGroup = CommonUtils.getPrimaryGroupName(mOwner, context.getClusterConf());
     } catch (IOException e) {
       mOwner = "";


liuhongtong:
@gpang @apc999 PTAL

liuhongtong:
@jiacheliu3 SaslClientHandlerPlain.SaslClientHandlerPlain() gets impersonation user and creates Sasl client.

```
/**
   * Creates {@link SaslClientHandler} instance for Plain/Custom.
   *
   * @param subject client subject
   * @param conf Alluxio configuration
   * @throws UnauthenticatedException
   */
  public SaslClientHandlerPlain(Subject subject, AlluxioConfiguration conf)
      throws UnauthenticatedException {
    if (subject == null) {
      throw new UnauthenticatedException("client subject not provided");
    }
    String connectionUser = null;
    String password = "noPassword";

    Set<User> user = subject.getPrincipals(User.class);
    if (user != null && !user.isEmpty()) {
      connectionUser = user.iterator().next().getName();
    }

    // Determine the impersonation user
    String impersonationUser = AuthenticationUserUtils.getImpersonationUser(subject, conf);

    mSaslClient = createSaslClient(connectionUser, password, impersonationUser);
  }
```

jiacheliu3:
@liuhongtong Do you mean this is the use case this change applies to?

If you mean `SaslClientHandlerPlain` is not giving the correct impersonated user, I don't see how the change in `OutStreamOptions` help, because this change in `mOwner` field doesn't change what `AuthenticationUserUtils.getImpersonationUser()` returns. Could you elaborate on this?

Still I suggest adding some unit tests to better illustrate what use cases the change is trying to solve. Thanks.

liuhongtong:
@jiacheliu3 Since SaslClientHandlerPlain uses impersonated user, should OutStreamOptions use the same user?

jiacheliu3:
@liuhongtong If we are impersonating, then `OutSteamOption` must to know the impersonated user, in order to talk to HDFS correctly. Does this answer your question?

i'm still very confused by your question though, what is the problem this PR is trying to solve?

liuhongtong:
@jiacheliu3 Yes.

jiacheliu3:
@liuhongtong Thanks now we have the use case. It makes sense to me. Can you please add the use case to the description? The description will be part of the commit when a PR is merged, so without it we can't recover what the PR is about later.

jiacheliu3:
@gpang This PR tries to carry the impersonation information in `OutStreamOptions` if any. The `mOwner` field is used in `OutStreamOption.getOwner()` in `GrpcDataWriter` so the data will be written as the impersonated user.

I'm not sure if this is the best place to make sure the impersonated user is given to the writer. My question is, since `OutStreamOptions` has `mOwner` `mGroup` and `mMode`, should the impersonation care about `mGroup` and `mMode`? I'm having this sense of inconsistency just because user, group and mode tend to be together as a unit, and it feels strange just to update part of this unit.

Do you think there are better ways to hand the impersonation info to the writer?

liuhongtong:
@jiacheliu3 I have updated comments for this PR.

apc999:
@gpang @jiacheliu3  any update on this pr?

jiacheliu3:
@liuhongtong Thanks for your patience. This PR makes sense to me. Can you please add the corresponding unit tests to `OutStreamOptions` for the change too?

Also I believe you probably have run into some related issues regarding impersonation and then fixed it with this change. Can you please link the related github issue if there is one? If not, please create one issue to describe what the error you initially ran into and how this change fixed the issue. Since impersonation is used by a number of Hadoop components that information will be very helpful for other users too. Thank you!

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -130,7 +131,11 @@ private OutStreamOptions(ClientContext context, AlluxioConfiguration alluxioConf
     mWriteTier = alluxioConf.getInt(PropertyKey.USER_FILE_WRITE_TIER_DEFAULT);
     mWriteType = alluxioConf.getEnum(PropertyKey.USER_FILE_WRITE_TYPE_DEFAULT, WriteType.class);
     try {
-      mOwner = context.getUserState().getUser().getName();
+      mOwner = AuthenticationUserUtils.getImpersonationUser(
+          context.getUserState().getSubject(), alluxioConf);
```
jiacheliu3: Please add unit tests which verify the impersonation user is got and set and later retrieved by `getOwner()` correctly. You can mock `AuthenticationUserUtils.getImpersonationUser()` in the unit tests.

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -130,7 +131,11 @@ private OutStreamOptions(ClientContext context, AlluxioConfiguration alluxioConf
...
+      if (mOwner == null) {
```
jiacheliu3: What if `mOwner` is not null while `AuthenticationUserUtils.getImpersonationUser()` returns you a user? I think we should output a WARNING log for the unexpected. Also I think it's worth logging in INFO or DEBUG when you are setting the impersonated user. If people run into impersonation issues they can find a trace.

PR #9747, opened 2019-08-19
Refactor gRPC channel/server building

ggezer:
Below refactoring items are done with this PR. Main motivation is to improve diagnosability of issues from traces by including precise source/target information for client channels.

- Extend tracing to contain more descriptive information on client side. https://github.com/Alluxio/alluxio/commit/cbdaf0e4d17bf1160f77f6439e8da2ccb597620b https://github.com/Alluxio/alluxio/commit/cdec65517eff980b982a6a94d34e5fb99100fe6e
- Improve server side tracing and simplify server building code. https://github.com/Alluxio/alluxio/commit/269b4013a2bd22db202d00797f72a3cf3d554f46 https://github.com/Alluxio/alluxio/commit/d850b80302c9bcfa0056ec2317ad23665f952273
- Miscellaneous improvements. https://github.com/Alluxio/alluxio/commit/e5ee04ceec8b5dd92566adbe4b4cc50caae97568

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
@@ -71,6 +71,17 @@ public final class GrpcChannelBuilder {
   }
 
   /**
+   * Sets human readable name for the channel's client.
+   *
+   * @param clientType client type
+   * @return the updated {@link GrpcChannelBuilder} instance
+   */
+  public GrpcChannelBuilder setClientType(String clientType) {
+    mChannelKey.setClientType(clientType);
+    return this;
+  }
+
+  /**
    * Sets {@link Subject} for authentication.
    *
    * @param subject the subject
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
@@ -29,11 +29,8 @@ import javax.security.auth.Subject;
  * building.
  */
 public final class GrpcChannelBuilder {
-  /** Key for acquiring the underlying managed channel. */
-  private GrpcManagedChannelPool.ChannelKey mChannelKey;
-
-  /** gRPC Server address. */
-  private GrpcServerAddress mServerAddress;
+  /** gRPC channel key. */
+  private GrpcChannelKey mChannelKey;
 
   /** Whether to use mParentSubject as authentication user. */
   private boolean mUseSubject;
@@ -51,11 +48,10 @@ public final class GrpcChannelBuilder {
   private AlluxioConfiguration mConfiguration;
 
   private GrpcChannelBuilder(GrpcServerAddress address, AlluxioConfiguration conf) {
-    mServerAddress = address;
     mConfiguration = conf;
-    mChannelKey = GrpcManagedChannelPool.ChannelKey.create(conf);
+    mChannelKey = GrpcChannelKey.create(conf);
     // Set default overrides for the channel.
-    mChannelKey.setAddress(address.getSocketAddress());
+    mChannelKey.setServerAddress(address);
     mChannelKey.setMaxInboundMessageSize(
         (int) mConfiguration.getBytes(PropertyKey.USER_NETWORK_MAX_INBOUND_MESSAGE_SIZE));
     mUseSubject = true;
@@ -187,7 +183,7 @@ public final class GrpcChannelBuilder {
    * @param strategy the pooling strategy
    * @return a new instance of {@link GrpcChannelBuilder}
    */
-  public GrpcChannelBuilder setPoolingStrategy(GrpcManagedChannelPool.PoolingStrategy strategy) {
+  public GrpcChannelBuilder setPoolingStrategy(GrpcChannelKey.PoolingStrategy strategy) {
     mChannelKey.setPoolingStrategy(strategy);
     return this;
   }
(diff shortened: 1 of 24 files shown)


ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -0,0 +1,316 @@
...
+
+/**
+ * Used to identify a unique {@link GrpcChannel}.
+ */
+public class GrpcChannelKey {
+  private static final Random RANDOM = new Random();
+
+  @IdentityField
+  @SuppressFBWarnings(value = "URF_UNREAD_FIELD")
+  Long mPoolKey = 0L;
```
calvinjia: Is the pool key only used for equality?

ggezer: Currently `mPoolKey` is a hack for working around pooling for block worker channels. Its value is used for both `hashCode` and `equals`. 
`GrpcChannelKey#setPoolingStrategy` would give you a better idea of its use.

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -0,0 +1,314 @@
...
+
+import com.google.common.base.MoreObjects;
+import io.netty.channel.Channel;
+import io.netty.channel.EventLoopGroup;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
```
gpang: what do all these new annotations do? The old code did not have them.

core/common/src/main/java/alluxio/grpc/GrpcServer.java
```diff
@@ -114,4 +115,10 @@ public void awaitTermination() {
   public boolean isServing() {
     return mStarted && !mServer.isShutdown() || !mServer.isTerminated();
   }
+
+  @Override
+  public String toString() {
+    // Netty server impl prints out its bind address.
+    return mServer.toString();
```
gpang: do we want more info in this string? (like it is a GrpcServer or auth server or anything else?) I don't know if any other info is useful...

core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
```diff
@@ -70,6 +70,17 @@ public static GrpcChannelBuilder newBuilder(GrpcServerAddress address,
...
   }
 
+  /**
+   * Sets human readable name for the channel's client.
+   *
+   * @param clientType client type
+   * @return the updated {@link GrpcChannelBuilder} instance
+   */
+  public GrpcChannelBuilder setClientType(String clientType) {
+    mChannelKey.setClientType(clientType);
```
gpang: if unset it is empty string? Is there a way to know if all channel keys set the client type?

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
ggezer: It's a new annotation defined below. It's easier to just tag fields to be included in identity calculation, rather than changing `hashCode` and `equals` implementations all the time. There has been some bugs in the past due to tediousness of prior way of maintaining the identity of channel.

core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
ggezer: We could assert that it's assigned. I saw it as an auxiliary field so didn't enforce it. Do you think we should always supply this field?

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -0,0 +1,314 @@
...
+import java.util.Optional;
+import java.util.Random;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Used to identify a unique {@link GrpcChannel}.
+ */
+public class GrpcChannelKey {
+  private static final Random RANDOM = new Random();
```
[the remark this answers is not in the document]
ggezer: I don't think this will cause contention as it's invoked only during physical channel building.

core/common/src/main/java/alluxio/grpc/GrpcServer.java
ggezer: Makes sense. Added authentication server info as well.

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+    return hashCodebuilder.toHashCode();
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other instanceof GrpcChannelKey) {
+      GrpcChannelKey otherKey = (GrpcChannelKey) other;
+      boolean areEqual = true;
+      for (Field field : this.getClass().getDeclaredFields()) {
```
gpang: will this type of reflection impact performance?

also, can we add unit tests for equals and hashcode? (do they already exist?)

core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
gpang: no, i don't think it has to be required.

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
ggezer: This will not be in hot path, only to acquire a physical channel. Since our channels are long lived I don't think we would notice any change in performance.

`GrpcManagedChannelPoolTest` does some testing on this area.

core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
ggezer: will leave as it is for now then.

gpang:

gpang:

Issue #9760, opened 2019-08-21
Failed to use RocksDB in docker


cheyang:
**Alluxio Version:**

2.1.0-SNAPSHOT

**Describe the bug**
Follow the guide: https://docs.alluxio.io/os/user/stable/en/deploy/Running-Alluxio-On-Docker.html  ,and try to change metadata store from `HEAP` to `RocksDB`. 

```
 docker run -d   -p 19999:19999   --net=alluxio_nw   --name=alluxio-master   --hostname=alluxio-master   -e ALLUXIO_JAVA_OPTS="-Dalluxio.master.hostname=alluxio-master alluxio.master.metastore=ROCKS "   -v journal:/opt/alluxio/journal   -v ufs:/opt/alluxio/underFSStorage   alluxio/alluxio:2.1.0-SNAPSHOT master
```

The logs  show:

```
/entrypoint.sh: line 81: /usr/lib/jvm/java-1.8-openjdk/jre/lib/security/java.security: Permission denied
```

**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)

**Expected behavior**

**Urgency**
Describe the impact and urgency of the bug.

**Additional context**

ZacBlanco:
@madanadit This looks like an issue with setting the network TTL cache. I believe it is unrelated to RocksDB

https://github.com/Alluxio/alluxio/blob/master/integration/docker/entrypoint.sh#L80-L82

madanadit:
hi @cheyang the default user in the image is `alluxio` please run as `root` using the `--user` option. i'll fix the image.

ZacBlanco:
I feel like this is something that should be done as part of the container build process rather than the entrypoint

apc999:
is the documentation in master branch is still up to date @madanadit ?

madanadit:
@apc999 the documentation is broken right now. is created a PR to fix this issue.

PR #9762, opened 2019-08-21
Disable DNS cache at docker build instead of entrypoint

madanadit:
Fixes: https://github.com/Alluxio/alluxio/issues/9760

The entry disabling DNS cache should be written at image build time instead run time (entrypoint). When the image is run as a non-root user (such as `alluxio`), the user may not have the permission to write to the desired location.


madanadit:

madanadit:

PR #9763, opened 2019-08-21
Update k8s doc to use the configMap file

madanadit:
The way to create the configMap was updated previously and docs are stale.


madanadit:

PR #9771, opened 2019-08-22
Improve parent dir check for UfsSyncPathCache

liuzx32:
**Description**
Improving cache logic for PR #9668 #9714 . from #9715, transfer to #9773

**Improvement**
Add SyncTimeLevel to record time for the immediate parent dir, the intermediate parent dir, and the cmd "ls" or "ls -R".

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -738,7 +738,7 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
     Metrics.GET_FILE_INFO_OPS.inc();
     long opTimeMs = System.currentTimeMillis();
     LockingScheme lockingScheme =
-        createLockingScheme(path, context.getOptions().getCommonOptions(), LockPattern.READ);
+        createLockingScheme(path, context.getOptions().getCommonOptions(), LockPattern.READ, false);
     try (RpcContext rpcContext = createRpcContext();
          LockedInodePath inodePath = mInodeTree
              .lockInodePath(lockingScheme.getPath(), lockingScheme.getPattern());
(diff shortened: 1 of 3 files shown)


yuzhu:
'alluxio-bot, check this please

yuzhu:

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4577,11 +4577,17 @@ public RpcContext createRpcContext() throws UnavailableException {
 
   private LockingScheme createLockingScheme(AlluxioURI path, FileSystemMasterCommonPOptions options,
       LockPattern desiredLockMode) {
+    return createLockingScheme(path, options, desiredLockMode, true);
+  }
+
+  private LockingScheme createLockingScheme(AlluxioURI path, FileSystemMasterCommonPOptions options,
+      LockPattern desiredLockMode, boolean directParentRecursive) {
```
yuzhu: can we change the name of the flag to something like `isGetFileInfo` to be more explicit?

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
```diff
@@ -68,9 +68,10 @@ public void notifySyncedPath(String path, DescendantType descendantType) {
   /**
    * @param path the path to check
    * @param intervalMs the sync interval, in ms
+   * @param directParentRecursive check the recursive time of direct parent or not
    * @return true if a sync should occur for the path and interval setting, false otherwise
    */
-  public boolean shouldSyncPath(String path, long intervalMs) {
+  public boolean shouldSyncPath(String path, long intervalMs, boolean directParentRecursive) {
```
yuzhu: same comment here, more explicit naming?

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
```diff
@@ -89,12 +90,14 @@ public boolean shouldSyncPath(String path, long intervalMs) {
...
     // had been performed (to avoid a sync again).
+    int parentLevel = 0;
     String currPath = path;
     while (!currPath.equals(AlluxioURI.SEPARATOR)) {
       try {
         currPath = PathUtils.getParent(currPath);
+        parentLevel++;
         lastSync = mCache.getIfPresent(currPath);
-        if (!shouldSyncInternal(lastSync, intervalMs, true)) {
+        if (!shouldSyncInternal(lastSync, intervalMs, parentLevel > 1 || directParentRecursive)) {
```
yuzhu: i think we should always check recursively all levels of ancestors?

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
```diff
@@ -89,12 +90,14 @@ public boolean shouldSyncPath(String path, long intervalMs) {
```
yuzhu: I think here, we can add something like 
```
SyncTime lastParentSync = mCache.getIfPresent(PathUtils.getParent(path));
if (isGetStatus && !shouldSyncInternal(lastParentSync, intervalMs, false)) {
   return false;
}
```
basically if we are only getting the status and the parent has been synced, then we return false.
The code path should remain the same for listStatus, because we can't optimize it until we know if it is a dir or file.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
liuzx32: Yeah, I get it.

core/server/master/src/main/java/alluxio/master/file/meta/UfsSyncPathCache.java
liuzx32: Ok

liuzx32:
@yuzhu Some things wrong with my PR-EMAIL, please check #9773

PR #9772, opened 2019-08-22
[WIP] Extract credential properties from configuration interface

jiacheliu3:
Currently the credential properties get the credential properties in the same way as all normal properties. There's security risk of a user unintentionally retrieves credentials in plaintext and leaks it in various ways. This change aims at refactoring the configuration interface that distinguishes how the credentials are accessed from non-credential fields. This will improve the user awareness and make the configuration API more fool-proof. Also the new interface makes it easier to locate all credential property access. So future implementations can be made to secure the credential properties.


madanadit:
@jiacheliu3 can this be closed?

jiacheliu3:
@madanadit Actually this change is still on the plate but never got enough priority. I plan to wrap it up in this coming few days (probably by next week).

Issue #8979, opened 2019-05-05
Show progress while taking a backup


ggezer:
**Is your feature request related to a problem? Please describe.**
Backup command takes a long time with no progress.

**Describe the solution you'd like**
Activate a progress thread by a flag in backup options and show how many entries are backed up for every 1-2 seconds.

PR #9776, opened 2019-08-22
Show progress for taking/applying backups

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/8979

### DIFF
--- a/shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
+++ b/shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
@@ -11,13 +11,15 @@
 
 package alluxio.cli.fsadmin.command;
 
+import alluxio.Constants;
 import alluxio.cli.CommandUtils;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.exception.status.InvalidArgumentException;
 import alluxio.grpc.BackupPOptions;
 import alluxio.grpc.BackupPOptions.Builder;
-import alluxio.wire.BackupResponse;
+import alluxio.util.CommonUtils;
 
+import alluxio.wire.BackupResponse;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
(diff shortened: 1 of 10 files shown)


shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
```diff
@@ -63,12 +65,21 @@ public int run(CommandLine cl) throws IOException {
       opts.setTargetDirectory(args[0]);
     }
     opts.setLocalFileSystem(cl.hasOption(LOCAL_OPTION.getLongOpt()));
-    BackupResponse resp = mMetaClient.backup(opts.build());
+    // Create progress thread for showing progress while backup is in progress.
```
LuQQiu: Currently the detailed backup progress is shown in leader master.log, looks like we are not able to print the detailed backup progress to users directly. Can we add some user-facing console message to tell them the detailed backup progress is in leader master master.log?

ggezer: We can, but details we can provide is not very interesting for a shell user. We cannot show percentage or estimation. We can only show how many entries applied so far, which is not indicative of anything, that's why we decided to go with simple dot-dot progress.

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterClient.java
```diff
@@ -35,7 +35,7 @@
    * @param options backup options
    * @return the server response
    */
-  BackupResponse backup(BackupPOptions options) throws IOException;
+  BackupPResponse backup(BackupPOptions options) throws IOException;
```
gpang: will this break client compatibility?

ggezer: Yes.

gpang: @apc999 are we ok with this client change?

ggezer:
Could someone please merge this PR? 
@gpang @apc999.

apc999:

PR #9792, opened 2019-08-26
Implement auto-mount for Transparent URIs

ggezer:
For handling foreign URIs that are not mounted to Alluxio. 
https://github.com/Alluxio/alluxio/issues/9231

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
@@ -218,14 +218,18 @@ public interface FileSystemMaster extends Master {
 
   /**
    * Translates given URI to an Alluxio URI.
+   *
    * If given URI does not contain a scheme, then it is regarded as Alluxio URI.
    * For other cases it will go through a reverse lookup against mount table.
+   *
+   * If reverse-lookup fails, this method will attempt to auto-mount UFS owning
+   * the foreign URI.
    *
    * @param uriStr uri
    * @return URI in Alluxio namespace
    * @throws InvalidPathException if foreign URI not found in Alluxio
    */
-  AlluxioURI translateUri(String uriStr) throws InvalidPathException;
+  AlluxioURI autoMountOrTranslateUri(String uriStr) throws InvalidPathException;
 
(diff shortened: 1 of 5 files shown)


core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
```diff
@@ -218,9 +218,13 @@ long getNewBlockIdForFile(AlluxioURI path) throws FileDoesNotExistException, Inv
 
   /**
    * Translates given URI to an Alluxio URI.
+   *
    * If given URI does not contain a scheme, then it is regarded as Alluxio URI.
    * For other cases it will go through a reverse lookup against mount table.
    *
+   * If reverse-lookup fails, this method will attempt to auto-mount UFS owning
```
gpang: Now that I think about it, I think this method should just do the translation, and the auto-mounting part should be an RPC option. I could imagine that doing a reverse lookup is a legitimate request, irrespective of shimfs.

ggezer: That's why I initially started with PR below to be able to provide shim specific options to master: https://github.com/Alluxio/alluxio/pull/9626. Without this work, I cannot isolate RPC options between shim and others. WDYT?

gpang: I don't think we need an options provider. Can the RPC itself just have an `auto-mount` option? By default it can be `false`, but in the shim, it will set it to `true`.

ggezer: `ShimFS` is just a wrapper over `AbstractFileSystem`, I should either make `AbstractFileSystem` aware that it's a shim and pass that flag along accordingly. Or, pass a provider from FileSystem implementations that knows fs specific defaults.

gpang: ahhh, i didnt realize `translate` was done implicitly for every RPC, and not just a simple rpc... I think this changes my assumptions now.

gpang:

PR #9799, opened 2019-08-27
Disallow foreign schemes on mounts that have nested mounts

ggezer:
Master should restrict foreign URI's access to its own mount only. To guarantee that this fix blocks foreign URI translation for mounts that have nested mounts.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1377,14 +1377,16 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
       // nested mounts as translation on one mount should not give visibility to another.
       if (mMountTable.containsMountPoint(resolution.getMountInfo().getAlluxioUri(), false)) {
         throw new IllegalStateException(
-            String.format("Foreign URI: %s is in a mount that has nested mounts.", uriStr));
+            String.format(
+                "Foreign URI: %s is reverse resolved to a mount :%s which has nested mounts.",
+                uriStr, resolution.getMountInfo().getAlluxioUri()));
       }
       return resolution.getUri();
     }
 
     // Resolve unsuccessful
     if (!ServerConfiguration.getBoolean(PropertyKey.MASTER_SHIMFS_AUTO_MOUNT_ENABLED)) {
-      throw new InvalidPathException(String.format("Could not reverse resolve path: %s", uriStr));
+      throw new InvalidPathException(String.format("Could not reverse resolve ufs path: %s", uriStr));
     }
 
     // Attempt to auto-mount.
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1371,16 +1371,23 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
     }
 
     // Reverse lookup mount table to find mount point that contains the path.
-    try {
-      return mMountTable.reverseLookup(uriStr);
-    } catch (InvalidPathException e) {
-      // Throw if auto-mount is not requested.
-      if (!ServerConfiguration.getBoolean(PropertyKey.MASTER_SHIMFS_AUTO_MOUNT_ENABLED)) {
-        throw e;
+    MountTable.ReverseResolution resolution = mMountTable.reverseResolve(uri);
+    if (resolution != null) {
+      // Before returning the resolved path, make sure mount point does not contain
+      // nested mounts as translation on one mount should not give visibility to another.
+      if (mMountTable.containsMountPoint(resolution.getMountInfo().getAlluxioUri(), false)) {
+        throw new IllegalStateException(
+            String.format("Foreign URI: %s is in a mount that has nested mounts.", uriStr));
       }
+      return resolution.getUri();
     }
 
-    // Reverse lookup failed. Attempt to auto-mount.
+    // Resolve unsuccessful
+    if (!ServerConfiguration.getBoolean(PropertyKey.MASTER_SHIMFS_AUTO_MOUNT_ENABLED)) {
+      throw new InvalidPathException(String.format("Could not reverse resolve path: %s", uriStr));
+    }
+
+    // Attempt to auto-mount.
     try {
       autoMountUfsForUri(uri);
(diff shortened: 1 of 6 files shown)


core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
```diff
@@ -389,19 +389,35 @@ public Resolution resolve(AlluxioURI uri) throws InvalidPathException {
...
+          foundEntry = mountEntry;
+          break;
         }
       }
+      if (foundEntry != null) {
+        // Validate the found mount does not have a nested mount.
+        // This is not allowed in order to restrict foreign URI schemes' scope.
+        for (Map.Entry<String, MountInfo> mountEntry : mState.getMountTable().entrySet()) {
+          if (mountEntry.getValue().getMountId() != foundEntry.getValue().getMountId()) {
+            if (mountEntry.getKey().startsWith(foundEntry.getKey())) {
```
gpang: I think this error should happen at a different layer. The mount table should just do the reverse lookup correctly, and some caller of the this lookup should determine if that is an error or not. I can imagine that just doing a reverse lookup is a legitimate operation, regardless of shimfs or not. This specific scenario we are trying to prevent is specific to shimfs.

ggezer: From outside it's not possible to do this without returning, along with the lookup result, the `MountEntry` that resolved it. So that it can be known if foreign URI is a root of some mount or is it stretching some mount to a nested mount.

Alternatively, I can hide this logic behind a flag. WDYT?

ggezer:
@gpang Updated as we discussed. PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
```diff
@@ -495,6 +471,33 @@ public long getMountId() {
     }
   }
 
+  /**
+   * This class represents a Alluxio path after resolution.
```
gpang: NIT:
```suggestion
   * This class represents a Alluxio path after a reverse resolution.
```

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
```diff
@@ -262,14 +262,19 @@ public String getMountPoint(AlluxioURI uri) throws InvalidPathException {
 
   /**
    * @param uri the Alluxio uri to check
-   * @return true if the specified uri is mount point, or has a descendant which is a mount point
+   * @param containsSelf cause method to return true when given uri itself is a mount point
+   * @return true if the given uri has a descendant which is a mount point [, or is a mount point]
    */
-  public boolean containsMountPoint(AlluxioURI uri) throws InvalidPathException {
+  public boolean containsMountPoint(AlluxioURI uri, boolean containsSelf)
```
gpang: can we add some unittests to `MountTableTest` for this method?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1371,24 +1371,36 @@ public AlluxioURI translateUri(String uriStr) throws InvalidPathException {
...
-      // Throw if auto-mount is not requested.
-      if (!ServerConfiguration.getBoolean(PropertyKey.MASTER_SHIMFS_AUTO_MOUNT_ENABLED)) {
-        throw e;
+    MountTable.ReverseResolution resolution = mMountTable.reverseResolve(uri);
+    if (resolution != null) {
+      // Before returning the resolved path, make sure mount point does not contain
+      // nested mounts as translation on one mount should not give visibility to another.
+      if (mMountTable.containsMountPoint(resolution.getMountInfo().getAlluxioUri(), false)) {
+        throw new IllegalStateException(
+            String.format("Foreign URI: %s is in a mount that has nested mounts.", uriStr));
```
gpang: can we include both the ufsUri and the reversed resolved alluxioUri in the exception message?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1371,24 +1371,36 @@ public AlluxioURI translateUri(String uriStr) throws InvalidPathException {
...
       }
+      return resolution.getUri();
     }
 
-    // Reverse lookup failed. Attempt to auto-mount.
+    // Resolve unsuccessful
+    if (!ServerConfiguration.getBoolean(PropertyKey.MASTER_SHIMFS_AUTO_MOUNT_ENABLED)) {
+      throw new InvalidPathException(String.format("Could not reverse resolve path: %s", uriStr));
```
gpang: NIT:
```suggestion
      throw new InvalidPathException(String.format("Could not reverse resolve ufs path: %s", uriStr));
```

core/server/master/src/main/java/alluxio/master/file/meta/MountTable.java
ggezer: MountTableTest#path does testing of reverseResolve. That's why I deleted old dupicates.

gpang: Are there any test cases for `containsMountPoint`?

ggezer: Added into `MountTableTest#pathNestedMount`.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1371,24 +1371,39 @@ public AlluxioURI translateUri(String uriStr) throws InvalidPathException {
...
-      if (!ServerConfiguration.getBoolean(PropertyKey.MASTER_SHIMFS_AUTO_MOUNT_ENABLED)) {
-        throw e;
+    MountTable.ReverseResolution resolution = mMountTable.reverseResolve(uri);
+    if (resolution != null) {
+      // Before returning the resolved path, make sure mount point does not contain
+      // nested mounts as translation on one mount should not give visibility to another.
+      if (mMountTable.containsMountPoint(resolution.getMountInfo().getAlluxioUri(), false)) {
+        throw new IllegalStateException(
+            String.format(
+                "Foreign URI: %s is reverse resolved to a mount :%s which has nested mounts.",
```
gpang: NIT
```suggestion
                "Foreign URI: %s is reverse resolved to a mount: %s which has nested mounts.",
```

ggezer:

PR #9803, opened 2019-08-28
Update generate-tarball script to support hadoop-3.2

chcn:


apc999:
@chcn Thanks for adding the support of Hadoop-3.2! I see this PR is making two efforts here:
(1) Add 3.2 to the tarball generation
(2) Change the default distribution version also to 3.2?

Can we limit the scope of this PR to be (1) (only changing `dev/scripts/src/alluxio.org/build-distribution/cmd/common.go`). @ns1123 will create a separate PR as we plan to have default version as hadoop 2.7 in the coming Alluxio 2.1 release

chcn:
@apc999 `generate-tarball.go` has to be changed because the hadoopDistributionFlag was never used, the defaultHadoopClient was always used. Also, the default version is already set to 2.7 in this PR.

apc999:
taken over by https://github.com/Alluxio/alluxio/pull/9889.  I will also create another PR to bump default hadoop version to 2.7

Issue #9240, opened 2019-06-08
Duplicated UnderFileSystem created by Master


apc999:
**Summary**
Multiple `UnderFileSystemFactory` is created for the same root mount: 
```
2019-06-01 14:22:18,558 DEBUG ExtensionFactoryRegistry - Factory implementation alluxio.underfs.s3a.S3AUnderFileSystemFactory@1cbd6fb0 is eligible for path s3://alluxio-quick-start/data
2019-06-01 14:22:18,558 DEBUG ExtensionFactoryRegistry - Factory implementation alluxio.underfs.s3a.S3AUnderFileSystemFactory@86819a1 is eligible for path s3://alluxio-quick-start/data
```

One is from `DefaultMetaMaster`:
```
        at alluxio.extensions.ExtensionFactoryRegistry.findAll(ExtensionFactoryRegistry.java:144)
        at alluxio.underfs.UnderFileSystemFactoryRegistry.findAll(UnderFileSystemFactoryRegistry.java:109)
        at alluxio.underfs.UnderFileSystem$Factory.create(UnderFileSystem.java:93)
        at alluxio.underfs.UnderFileSystem$Factory.createForRoot(UnderFileSystem.java:138)
        at alluxio.master.meta.DefaultMetaMaster.<init>(DefaultMetaMaster.java:186)
        at alluxio.master.meta.DefaultMetaMaster.<init>(DefaultMetaMaster.java:157)
        at alluxio.master.meta.MetaMasterFactory.create(MetaMasterFactory.java:51)
        at alluxio.master.meta.MetaMasterFactory.create(MetaMasterFactory.java:29)
        at alluxio.master.MasterUtils.lambda$createMasters$0(MasterUtils.java:49)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```
And the other is from `DefaultFileSystemMaster`
```
        at alluxio.extensions.ExtensionFactoryRegistry.findAll(ExtensionFactoryRegistry.java:144)
        at alluxio.underfs.UnderFileSystemFactoryRegistry.findAll(UnderFileSystemFactoryRegistry.java:109)
[...]

PR #9808, opened 2019-08-29
Move UfsManager to CoreMasterContext

apc999:
Fix https://github.com/Alluxio/alluxio/issues/9240

This moves provides `UfsManager` to master context so it can be shared across different masters. In this way, `DefaultMetaMaster` no longer needs to call UFS Factory to make UFS instances but consolidating to `UfsManager`


core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -243,13 +246,8 @@ public void resetState() {
...
 
-    if (URIUtils.isLocalFilesystem(ServerConfiguration
-        .get(PropertyKey.MASTER_MOUNT_TABLE_ROOT_UFS))) {
-      mUfs = UnderFileSystem.Factory
-          .create("/", UnderFileSystemConfiguration.defaults(ServerConfiguration.global()));
-    } else {
-      mUfs = UnderFileSystem.Factory.createForRoot(ServerConfiguration.global());
-    }
+    mRootUfsClient = masterContext.getUfsManager().getRoot();
+    mRootUfs = mRootUfsClient.acquireUfsResource();
```
calvinjia: Does it make sense to keep these as variables, or could we acquire them when necessary?

apc999: changed to acquire the resource when necessary. I think the code becomes more consistent and easier to reason.

apc999:
@calvinjia  PTAL

apc999:
ping @calvinjia

calvinjia:

PR #9817, opened 2019-09-02
Improve gRPC channel-pool concurrency

ggezer:
`Copycat` could invoke creation/shutdown of a connection concurrently. With gRPC transport, that translates to concurrently acquiring/releasing managed channels from `GrpcManagedChannelPool` for the same channel-key. `GrpcManagedChannelPool` was not handling this case properly and a bug was causing managed channel to remain shutdown. So a communication to a particular target was lost completely. (Other users of the channel pool were safe due to synchronous conventions they follow.)

This PR simplifies locking model for channel pooling which also fixes concurrency problems caused by copycat transport layer. Main changes are:
- the pool is never locked globally. Channels for different targets are built/destroyed without blocking each other.
- `acquire` and `release` acquires an exclusive lock per channel-key. (indirectly via `ConcurrentHashMap#compute()`)

One indirect result of above changes is that shutting down of managed channels is only serialized per channel-key. The reason we serialized channel shutdown was to circumvent some internal gRPC bug which was causing unceasing echoing of packets between client/server. It turns out serializing shutting down of channels per channel-key is sufficient.

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
@@ -12,10 +12,10 @@
 package alluxio.grpc;
 
 import alluxio.collections.Pair;
-import alluxio.resource.LockResource;
 import alluxio.util.CommonUtils;
 import alluxio.util.WaitForOptions;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Verify;
 import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
@@ -25,17 +25,15 @@ import io.netty.channel.EventLoopGroup;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
-import java.util.HashMap;
 import java.util.Optional;
-import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 /**
  * Used to maintain singleton gRPC {@link ManagedChannel} instances per process.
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
```diff
@@ -139,82 +76,54 @@ private boolean waitForChannelReady(ManagedChannel managedChannel, long healthCh
...
       }
-    }
-    try (LockResource lockExclusive = new LockResource(mLock.writeLock())) {
-      // Dispose existing channel if required.
-      int existingRefCount = 0;
-      if (shutdownExistingChannel && mChannels.containsKey(channelKey)
-          && mChannels.get(channelKey) == managedChannelRef) {
-        existingRefCount = managedChannelRef.getRefCount();
+      if (shutdownExistingChannel) {
+        existingRefCount = chHolder.getRefCount();
```
calvinjia: Why do we preserve the old ref count if we are shutting down the channel? I don't think the other users of the channel be able to pick up the new healthy channel we create?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
```diff
@@ -261,31 +170,78 @@ private ManagedChannel createManagedChannel(GrpcChannelKey channelKey) {
...
+          case TRANSIENT_FAILURE:
+          case SHUTDOWN:
+            return false;
+          case IDLE:
+          case CONNECTING:
+            return null;
+          default:
+            return null;
+        }
+      }, WaitForOptions.defaults().setTimeoutMs((int) healthCheckTimeoutMs));
```
calvinjia: Should we set the interval based on the timeout?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
```diff
@@ -261,31 +170,78 @@ private ManagedChannel createManagedChannel(GrpcChannelKey channelKey) {
...
+    managedChannel.shutdown();
+    try {
+      managedChannel.awaitTermination(shutdownTimeoutMs, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      // Allow thread to exit.
+    } finally {
+      managedChannel.shutdownNow();
+    }
+    Verify.verify(managedChannel.isShutdown());
```
calvinjia: Is this necessary?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
ggezer: Due to shutdown, their operations should start failing and eventually they'll release their channels. When they reacquire after release, that's when they'll get the new healthy channel.

calvinjia: So setting the old ref count is to account for the future `release` calls with the same channelKey on the now shutdown channel?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
ggezer: I think the default (20ms) is what we want as we'd like to get out out of this ASAP.

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
ggezer: Not really. removed it.

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
bf8086: Would it be better to pass in the managed channel in the release method so the release of an already shutdown channel becomes a noop instead of affecting the reference counting of the new channel?

ggezer: @calvinjia Correct.
@bf8086 That will require isolating the resource from the pool. I'm not sure what we'll gain from doing that.

bf8086: All the callers of the release method has access to the resource(managed channel) already so they can just pass in. It feels more clear to me when we can explicitly specify which actual managed channel we release so we don't have to worry about dereference new channel from users of the old channel.

ggezer: I agree this is not very clean. However I don't feel good about requiring both the key and the channel itself during `release`. Looks like it could be done clearly if we return a pool-handle from `acquire` call, rather than returning a channel.. Put a TODO for addressing this later.

ggezer: https://github.com/Alluxio/alluxio/issues/9831

calvinjia:

PR #9829, opened 2019-09-03
Remove hive metastore dependency on table client

ZacBlanco:
- Proto definitions for hive metastore
  - Proto definitions are based off of the hive metastore definitions at https://github.com/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift. They may be subject to change.
- `CatalogMasterClient` interface replaces the Hive thrift object types with Alluxio proto object types.


gpang:

PR #9837, opened 2019-09-04
Handle signals properly in Docker entrypoint

### LINKED ISSUE #9744: Wrong "ps" parameter in start/stop scripts when using alpine based docker image
In file **$ALLUXIO_HOME/bin/alluxio** in line 83 the call **ps -Aww** fails because busybox doesn't know the 'w' flag.
busybox is used when the environment is alpine (docker image!)

I suggest using pgrep instead of parsing the output of ps like this:

replace

`for pid in $(ps -Aww -o pid,command | grep -i "[j]ava" | grep ${keyword} | awk '{print $1}'); do`

with

`for pid in $(pgrep -f "[j]ava.*$keyword"); do`

ZacBlanco:
This PR improves the docker entrypoint script in a few ways:

- When launching single-process containers (i.e. worker-only, job worker,
master-only, proxy) the child process is now exec'd at the end of the
script so that the status of the container directly reflects the status
of the single alluxio process.
- Inside multi-process containers (master, worker) the child processes
will  now receive any signals which are sent to the parent shell
process.  This is achieved by using the `trap` feature of bash to
forward the signals. See the comments in the script for some more
details

The main caveat with this implementation is that the signal
handler for multi-process containers currently has a `wait` at the end of
the handler. This disallows the same signal to be sent multiple times.
Without the  `wait` the script will exit when the trap handler finishes. Not
all signals may cause Alluxio processes to exit, but without the wait, they
will forced to exit at any signal.

Resolves #9744


integration/docker/entrypoint.sh
```diff
@@ -113,41 +94,126 @@ function mountAlluxioRootFSWithFuseOption {
...
+  local signal
+  signal="${1}"
+
+  # background jobs don't respond to SIGINT
+  # Change to SIGHUP
+  if [ "${signal}" -eq "2" ]; then
+    signal="1"
+  fi
+
+  local procs="$(jobs -p)"
```
madanadit: how does this work? seems like you launch the background processes after setting up signals

integration/docker/entrypoint.sh
```diff
@@ -113,41 +94,126 @@ function mountAlluxioRootFSWithFuseOption {
...
+      processes+=("worker")
+      ;;
+    job-worker)
+      processes+=("job_worker")
+      ;;
+    proxy)
+      processes+=("proxy")
+      ;;
+    fuse)
+      mountAlluxioRootFSWithFuseOption
```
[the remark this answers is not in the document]
ZacBlanco: I was looking into it and was planning opn leaving it as a follow-up PR. However if you feel it's appropriate I can roll it into this PR. I might need to make some changes to the FUSE daemon script to get it working properly.

integration/docker/entrypoint.sh
```diff
@@ -113,41 +94,126 @@ function mountAlluxioRootFSWithFuseOption {
...
+  if [ -z "${processes}" ]; then
     printUsage
     exit 1
-    ;;
-esac
+  fi
+
+  # Only a single process is going to be started, simply exec and replace in the shell
+  if [ "${#processes[@]}" -eq 1 ]; then
+    exec ./bin/launch-process "${processes[0]}" -c
```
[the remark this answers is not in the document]
ZacBlanco: `exec` will not return to the calling process. Anything after this line will not be executed.

i.e. in the that `launch-process` fails, the exit code from `launch-process` will be returned to the parent shell which called the entrypoint script.

integration/docker/entrypoint.sh
ZacBlanco: Here's the general flow I had in my head.

```
- Must launch either 1 process or 2 processes
  - if 1, then simply exec and the signal handlers don't need to be set up. All signals will go straight to the process
- If 2 processes, then a signal must be forwarded to both background processes
  - To forward a signal to 2 processes, set up all the traps within the script first
  - Then launch the background processes
  - Any signals after this point get forwarded
```

ZacBlanco: Also,

calling `trap` is like setting up a signal handler. Basically the function doesn't get called _until_ the signal is sent.

so calling `trap "forward_signal 1" 1` only executes the `forward_signal 1` command when `1` is sent to the shell process.

integration/docker/entrypoint.sh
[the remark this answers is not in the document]
madanadit: you can open an issue for the fuse fix

integration/docker/entrypoint.sh
madanadit: got it, thx

integration/docker/entrypoint.sh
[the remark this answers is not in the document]
ZacBlanco: Issue created in https://github.com/Alluxio/alluxio/issues/9840

integration/docker/entrypoint.sh
```diff
@@ -113,41 +94,124 @@ function mountAlluxioRootFSWithFuseOption {
...
+}
+
+function main {
+  if [[ "$#" -lt 1 ]]; then
+    printUsage
+    exit 1
+  fi
+
+  local service="$1"
+  local options="$2"
```
bf8086: Will local `options` work for other functions?

ZacBlanco: Good catch! I will move options back to a global variable

ZacBlanco:

PR #9844, opened 2019-09-06
Handle signals properly in FUSE container

### LINKED ISSUE #9840: Update FUSE integration scripts to be able to be launched in the foreground

Recent the `./bin/launch-process` script was added to launch Alluxio processes in the foreground. This is now being used by `alluxio-start.sh` and the docker container entrypoint so that containerized Alluxio processes behave properly when signals are sent to the container.

However, the FUSE daemon launched from the script `./integration/fuse/bin` directory doesn't launch the process in the foreground, meaning that our containerized Alluxio FUSE daemon isn't a good citizen in the container world.

We should update the docker entrypoint and a`alluxio-fuse` script so that it can handle signals and terminate cleanly in docker or K8s.

Similar to `launch-process`, we'll need to use `exec` and limit use of `nohup` within the FUSE daemon script.

We could write a separate entrypoint for the fuse container, but it would still require changes to the fuse daemon script.

**Urgency**

medium

**Additional context**

N/A

ZacBlanco:
Resolves #9840


integration/docker/Dockerfile.fuse
```diff
@@ -21,7 +21,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends software-proper
     add-apt-repository -y ppa:openjdk-r/ppa && \
 	apt-get update && \
 	apt-get install -y --no-install-recommends \
-	openjdk-8-jdk openjdk-8-jre-headless libfuse-dev && \
+	openjdk-8-jdk openjdk-8-jre-headless fuse && \
```
madanadit: why is this change needed?

LuQQiu:
Docker and kubernetes require all processes running as foreground processes?

madanadit:
@LuQQiu The proper way in Kubernetes is to run processes in the foreground. Multiple processes are run in separate containers (potentially part of the same pod = the schedulable unit)

LuQQiu:
@madanadit  Got it! Thanks for the details!

integration/docker/Dockerfile.fuse
```diff
@@ -50,6 +50,8 @@ RUN if [ ${ALLUXIO_USERNAME} != "root" ] \
       mkdir -p /journal && \
       chown -R ${ALLUXIO_UID}:${ALLUXIO_GID} /opt/* /journal && \
       chmod -R g=u /opt/* /journal; \
+      mkdir /alluxio-fuse && \
```
LuQQiu: I am still curious why previously docker and kubernetes can launch Alluxio fuse containers without creating the alluxio-fuse folder?

ZacBlanco: Previously this worked because the container ran as `root` (uid 0), so the directory could be created at runtime. Now, because we're running as `alluxio`, permissions to `/` are denied. So we create the folder and `chown` at container creation time

ZacBlanco:
FYI I tested this locally and everything seemed to work. Only issue was that my master container ran as root instead of Alluxio. Due to the mismatch in users. I had to `chown` the local UFS volume as `alluxio` to allow files to be created on the fuse mount.

ZacBlanco:

Issue #9628, opened 2019-08-02
Metrics JSON endpoint not reporting correct UFS metrics


ns1123:
**Alluxio Version:**
1.8.2

**Describe the bug**
UFS metrics are not reported correctly. When creating two different mount points on the same HDFS cluster, operations on the second mount point are incorrectly reported to the first mount point.

**To Reproduce**
Mount two different directories from an HDFS cluster into the same Alluxio cluster. RunTests on the second mount directory. Check the /metrics/json/ endpoint and see the value of getBlockSizeBytes().

**Expected behavior**
Operations like getBlockSizeBytes should be reported based on which mount point they were run on instead of aggregating in the first mount point.

**Urgency**
Medium

**Additional context**
Was not able to reproduce in 2.x

gpang:
@ns1123 @chcn Do we know why is this not reproducible on master branch?

gpang:
fixed by https://github.com/Alluxio/alluxio/pull/9846 https://github.com/Alluxio/alluxio/pull/10343

PR #9846, opened 2019-09-06
Fix UFS RPC metrics aggregation

chcn:
Originally, the metrics are aggregated on UFS paths. But in scenario where two paths in the same UFS are mounted to different Alluxio mount points, since there is only one UFS client for the mount points, all metrics are aggregated to the first mount point even when operations are on the second mount point, which is confusing.
This PR aggregates UFS RPC metrics per UFS, instead of per UFS path.

fixes Alluxio/alluxio#9628

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
+++ b/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
@@ -625,7 +625,7 @@ public class UnderFileSystemWithLogging implements UnderFileSystem {
 
   private String getUfsTag() {
     AlluxioURI uri = new AlluxioURI(mPath);
-    uri = new AlluxioURI(uri.getScheme(), uri.getAuthority(), uri.getRootPath());
+    uri = new AlluxioURI(uri.getScheme(), uri.getAuthority(), uri.getPath());
     return MetricsSystem.escape(uri);
   }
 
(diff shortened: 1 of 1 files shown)


gpang:
@chcn This should target master branch first, then cherry-picked to other branches.

chcn:
@gpang according to @ns1123 , this issue was only observable in 1.8.

ns1123:
@gpang at the time I tested on 2.x I didn't realize that mounting the same HDFS cluster with different paths would piggyback on the same UFSclient. So I tested with two S3 URI's and potentially different mount options to stay in line with the original issue reported. This likely is an issue in 2.x but I would need to test again.

gpang:
@ns1123 @chcn If this affects master, then the fix should go to master, then branch-2.0, then branch-1.8

apc999:
@gpang @ns1123 any update?

gpang:
I think this fix should first target master branch, then cherry-picked to other branches.

yuzhu:
@chcn does this PR need to be merged? or did you address this some other way?

yuzhu:
@chcn @ZacBlanco this probably needs to be merged?

ZacBlanco:
This PR must be closed and a new one re-targeted against master

yuzhu:
@ZacBlanco @gpang reopened #10343 to replace this one.

Issue #9848, opened 2019-09-07
Support dumping ROCKS meta store in journal tool


ggezer:
**Is your feature request related to a problem? Please describe.**
Currently dumping a journal with `JournalTool`, only extracts ROCKS-db files as tarball. 

**Describe the solution you'd like**
`RocksInodeStore` can wrap the extracted db content to export all `inode`s.

**Describe alternatives you've considered**
A separate tool for dumping `inode`s in ROCKS-db tarball.

**Urgency**
Dumping the actual `inode` content, as in HEAP store, would be better for diagnostics.

**Additional context**

PR #9860, opened 2019-09-09
Dump ROCKS checkpoint in human readable form

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/9848


ggezer:
Jenkins, test this please

shell/pom.xml
```diff
@@ -75,6 +75,16 @@
...
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-core-server-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-core-server-master</artifactId>
```
LuQQiu: Alluxio client jar will be distributed to many different applications. For example, it may need to distribute to a hadoop cluster with hundreds node. We want to keep client jar light weight. Not sure is it acceptable to make alluxio client jar depends on the core-server module. Ideally, we should have a separate alluxio-examples jar (now we include many tools and tests in the client jar which is kind of strange).

shell/pom.xml
```diff
@@ -75,6 +75,16 @@
       <artifactId>alluxio-core-common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-core-server-common</artifactId>
```
gpang: why does `shell` need to depend on `server`?

shell/src/main/java/alluxio/master/journal/tool/AbstractJournalDumper.java
```diff
@@ -105,9 +108,23 @@ private void readCompoundCheckpoint(CheckpointInputStream checkpoint, Path path)
...
+      RocksInodeStore inodeStore = new RocksInodeStore(dbPath.toAbsolutePath().toString());
+      inodeStore.restoreFromCheckpoint(checkpoint);
+      // Dump entries.
+      final String ENTRY_SEPARATOR = Strings.repeat("-", 80);
+      for (InodeView inode : (Iterable<InodeView>) () -> inodeStore.iterator()) {
+        out.println(ENTRY_SEPARATOR);
+        out.println(inode.toProto());
+      }
+      // Remove the temp db directory.
+      FileUtils.deletePathRecursively(dbPath.toFile().getPath());
```
gpang: should this be in a `finally` block?

shell/src/main/java/alluxio/master/journal/tool/AbstractJournalDumper.java
```diff
@@ -105,9 +108,23 @@ private void readCompoundCheckpoint(CheckpointInputStream checkpoint, Path path)
...
 
   private void readRocksCheckpoint(CheckpointInputStream checkpoint, Path path) throws IOException {
-    TarballCheckpointFormat.TarballCheckpointReader reader =
-        new TarballCheckpointFormat.TarballCheckpointReader(checkpoint);
-    reader.unpackToDirectory(path);
+    // Create RocksInodeStore over checkpoint stream for extracting the inodes.
+    try (PrintStream out =
+        new PrintStream(new BufferedOutputStream(new FileOutputStream(path.toFile())))) {
+      // Create an empty dir for storing the db.
+      Path dbPath = Paths.get(path.toFile().getPath() + "db");
```
gpang: Is a "db" a sub directory of the path? Is there always a path separator before "db"?

Also, "db" should probably be more descriptive, like "rocksdb-checkpoint" or something.

ggezer:
@gpang @LuQQiu 
For achieving readability on ROCKS checkpoint, journal tool need to access core-server-master as meta-store is hosted in that module. If we don't want to bloat the client-assembly (not the client-jar as it's unaffected by this change), then we'd need to move journal tool to core-server-master instead.

shell/src/main/java/alluxio/master/journal/tool/AbstractJournalDumper.java
ggezer: `path` will be used to create a file. `dbPath` will create a dir along with `path` in order to temporarily store the db content.

shell/pom.xml
ggezer: Moved journal-tool to `core-server-master`.

shell/pom.xml
ggezer: Moved journal-tool to `core-server-master`.

ggezer:
@gpang @LuQQiu Moved journal-tool to `core-server-master` instead and resolved remaining comments. PTAL.

ggezer:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/journal/tool/AbstractJournalDumper.java
```diff
@@ -105,9 +108,24 @@ private void readCompoundCheckpoint(CheckpointInputStream checkpoint, Path path)
   }
 
   private void readRocksCheckpoint(CheckpointInputStream checkpoint, Path path) throws IOException {
-    TarballCheckpointFormat.TarballCheckpointReader reader =
-        new TarballCheckpointFormat.TarballCheckpointReader(checkpoint);
-    reader.unpackToDirectory(path);
+    // An empty dir for storing the db.
+    Path dbPath = Paths.get(path.toFile().getPath() + "rocks-db");
```
gpang:
```suggestion
    Path dbPath = Paths.get(path.toFile().getPath() + "-rocks-db");
```

ggezer:

Issue #9851, opened 2019-09-07
Seperate gRPC properties for copycat transport


ggezer:
**Is your feature request related to a problem? Please describe.**
Currently copycat's gRPC transport relies on client properties for channel settings. Client and copycat might need separate property settings.

**Describe the solution you'd like**
New set of EMBEDDED_JOURNAL_TRANSPORT_ family of properties for controlling gRPC transport's settings. 

Additionally, request timeout for copycat should have its own property rather than using leader-election-timeout value.

PR #9868, opened 2019-09-10
Use seperate gRPC properties for copycat transport

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/9851

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1316,6 +1316,46 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_NETWORK_REQUEST_TIMEOUT_MS =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_REQUEST_TIMEOUT_MS)
+          .setDefaultValue("5sec")
+          .setDescription("Timeout for communication between embedded journal masters.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_NETWORK_FLOWCONTROL_WINDOW =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_FLOWCONTROL_WINDOW)
+          .setDefaultValue("2MB")
+          .setDescription("The HTTP2 flow control window used by copycat connections. Larger "
+              + "value will allow more data to be buffered but will use more memory.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_NETWORK_KEEPALIVE_TIME_MS =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_KEEPALIVE_TIME_MS)
+          .setDefaultValue(Long.MAX_VALUE)
+          .setDescription(
+              "The amount of time for a copycat client (for block reads and block writes) "
+                  + "to wait for a response before pinging the server to see if it is still alive.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_NETWORK_KEEPALIVE_TIMEOUT_MS =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_KEEPALIVE_TIMEOUT_MS)
+          .setDefaultValue("30sec")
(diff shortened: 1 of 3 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1316,6 +1316,46 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_NETWORK_REQUEST_TIMEOUT_MS =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_REQUEST_TIMEOUT_MS)
+          .setDefaultValue("5sec")
```
LuQQiu: Should the default request timeout be longer? Looks like under 5sec, we are easy to get request timeout especially when reading backup/catching up.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1316,6 +1316,46 @@ public String toString() {
...
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_FLOWCONTROL_WINDOW)
+          .setDefaultValue("2MB")
+          .setDescription("The HTTP2 flow control window used by copycat connections. Larger "
+              + "value will allow more data to be buffered but will use more memory.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_NETWORK_KEEPALIVE_TIME_MS =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_NETWORK_KEEPALIVE_TIME_MS)
+          .setDefaultValue(Long.MAX_VALUE)
```
LuQQiu: Why set the default value to be that big? Will it cause some copycat client connections which should be timeout but keep waiting and use out resources like open file limits?

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Currently It's consistent with the default leader election timeout. Bigger value will cause leadership changes to stall.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Nice catch. I needed to define server side properties and enforce ping traffic from the server.
However, copycat's request timeout (MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT_MS) renders network level optimizations useless so I removed all network related props.

ggezer:

apc999:
FYI, @ggezer   @LuQQiu for public API (e.g., configuration change) , please add me as reviewer. Thanks!

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1316,6 +1316,20 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT_MS =
```
apc999: `MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT` rather than `MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT_MS` please. 
In 2.0 code base, we no longer use suffix `_MS` to indicate the unit

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3933,6 +3947,10 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.master.embedded.journal.shutdown.timeout";
     public static final String MASTER_EMBEDDED_JOURNAL_TRIGGERED_SNAPSHOT_WAIT_TIMEOUT =
         "alluxio.master.embedded.journal.triggered.snapshot.wait.timeout";
+    public static final String MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT_MS =
+        "alluxio.master.embedded.journal.transport.request.timeout.ms";
```
apc999: `alluxio.master.embedded.journal.transport.request.timeout` please (not using `.ms` as suffix)

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3933,6 +3947,10 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.master.embedded.journal.shutdown.timeout";
     public static final String MASTER_EMBEDDED_JOURNAL_TRIGGERED_SNAPSHOT_WAIT_TIMEOUT =
         "alluxio.master.embedded.journal.triggered.snapshot.wait.timeout";
+    public static final String MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT_MS =
```
apc999: also move `MASTER_EMBEDDED_JOURNAL_TRANSPORT_REQUEST_TIMEOUT` before `MASTER_EMBEDDED_JOURNAL_TRIGGERED_SNAPSHOT_WAIT_TIMEOUT ` to keep them sorted?

Issue #9845, opened 2019-09-06
Rapid leader transition causes web-server leak


ggezer:
**Alluxio Version:**
2.0+

**Describe the bug**
In a quick fail-over from follower -> leader -> follower, a web server could leak. So the node will never be able to become a leader gain due to not being able to bind to port.

**To Reproduce**
Cause a quick transition of master role as described.

**Expected behavior**
We should not leak any resources during a quick state transition.

PR #9872, opened 2019-09-10
Fix leaking rejecting-servers when primacy is unstable

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/9845

If leadership is lost during transitioning, `startMasters(false)` ends up being called back-to-back. That will cause earlier rejecting-server to be leaked and block the port for later. Fix ensures we don't recreate over a created rejecting-server.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/MasterProcess.java
+++ b/core/server/common/src/main/java/alluxio/master/MasterProcess.java
@@ -175,10 +175,14 @@ public abstract class MasterProcess implements Process {
   }
 
   protected void startRejectingServers() {
-    mRejectingRpcServer = new RejectingServer(mRpcBindAddress.getPort());
-    mRejectingRpcServer.start();
-    mRejectingWebServer = new RejectingServer(mWebBindAddress.getPort());
-    mRejectingWebServer.start();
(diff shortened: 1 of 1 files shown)


core/server/common/src/main/java/alluxio/master/MasterProcess.java
```diff
@@ -175,10 +175,14 @@ public boolean waitForReady(int timeoutMs) {
   }
 
   protected void startRejectingServers() {
-    mRejectingRpcServer = new RejectingServer(mRpcBindAddress.getPort());
-    mRejectingRpcServer.start();
-    mRejectingWebServer = new RejectingServer(mWebBindAddress.getPort());
-    mRejectingWebServer.start();
+    if (mRejectingRpcServer == null) {
```
LuQQiu: If the original rejecting server host the port, creating a new rejecting server with the same port will not fail or show errors?

ggezer: Judging by the current code for `RejectingServer` it'll fail and quit without us knowing it.

ggezer:
Jenkins, test this please

ggezer:

PR #9877, opened 2019-09-11
Fix NPE when adding un-partitioned tables

ZacBlanco:
We should do a check if `part.getValues()` may have returned `null`


gpang:

Issue #9852, opened 2019-09-07
Strip embedded journal of hardcoded timeouts


ggezer:
**Alluxio Version:**
2.0+

**Describe the bug**
Some timeouts for RaftJournalSystem are not configurable such as:
  - term submit
  - flush
  - checkpoint

PR #9902, opened 2019-09-17
Define a property for copycat write timeout

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/9852


LuQQiu:
We adding many more properties for embedded journal but looks like lack of any docs of what values should be given for those values. Could we add some docs as well, maybe in future PRs? If we are not quite sure about what values should be suggested to users, we may need more testing

ggezer:

Issue #9862, opened 2019-09-10
Make copycat batch size configurable


ggezer:
**Is your feature request related to a problem? Please describe.**
Currently it is hard coded to 32KB.

This may not be enough to utilize network bandwidth when, for example, installing a snapshot to another node. Additionally, due to our modification to prevent checkpoints on leader, upon loosing leadership, it'll need to be installed a snapshot from the next leader. So the behavior of installing big knowledge in batches is increased by our design.

For reference (From copycat):

```
abstract class AbstractAppender implements AutoCloseable {
  private static final int MAX_BATCH_SIZE = 1024 * 32;
```

PR #9911, opened 2019-09-18
Define prop for setting copycat append batch size

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/9862

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1289,6 +1289,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
                   + "(alluxio.master.embedded.journal.election.timeout).")
           .setDefaultValue("1s")
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE)
+          .setDescription(" Amount of data that is appended from leader to followers "
+              + "in single batch. Setting higher values might require increasing"
+              + "election timeout due to increased network delay.")
+          .setDefaultValue(1024 * 32)
+          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_PORT =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_PORT)
           .setDescription("The port to use for embedded journal communication with other masters.")
@@ -3944,6 +3951,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.embedded.journal.addresses";
     public static final String MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT =
         "alluxio.master.embedded.journal.election.timeout";
+    public static final String MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE =
+        "alluxio.master.embedded.journal.appender.batch.size";
     public static final String MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
         "alluxio.master.embedded.journal.heartbeat.interval";
     public static final String MASTER_EMBEDDED_JOURNAL_PORT =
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1289,6 +1289,13 @@ public String toString() {
                   + "(alluxio.master.embedded.journal.election.timeout).")
           .setDefaultValue("1s")
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE)
+          .setDescription(" Amount of data that is appended from leader to followers "
```
gpang: NIT: empty space prefix

also, what is the unit?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1289,6 +1289,13 @@ public String toString() {
...
+              + "in single batch. Setting higher values might require increasing"
+              + "election timeout due to increased network delay.")
```
gpang: can you also mention the other tradeoff when the value is too low?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1289,6 +1289,13 @@ public String toString() {
...
+          .setDefaultValue(1024 * 32)
```
gpang: this was the original default. Is there a better default?

ggezer: In a separate effort, I'll work on finding a sane default for this and other embedded journal properties.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: unit is BYTES. Updated the name to reflect that.

ggezer:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3951,7 +3952,7 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.master.embedded.journal.addresses";
     public static final String MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT =
         "alluxio.master.embedded.journal.election.timeout";
-    public static final String MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE =
+    public static final String MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE_BYTES =
```
gpang: we can still keep the original name, so we can use abbreviations like `32KB`.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1289,6 +1289,14 @@ public String toString() {
                   + "(alluxio.master.embedded.journal.election.timeout).")
           .setDefaultValue("1s")
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE_BYTES =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE_BYTES)
+          .setDescription("Amount of data that is appended from leader to followers "
+              + "in a single heartbeat. Setting higher values might require increasing"
```
gpang:
```suggestion
              + "in a single heartbeat. Setting higher values might require increasing "
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1289,6 +1289,14 @@ public String toString() {
...
+              + "election timeout due to increased network delay. Setting lower values "
+              + "might stall knowledge propagation between the leader and followers.")
+          .setDefaultValue(1024 * 32)
```
gpang:
```suggestion
          .setDefaultValue("32KB")
```

ggezer:

PR #9920, opened 2019-09-20
Implement ByteBufferReadable for HdfsFileInputStream

gengxin:
hdfs c api prefer to use direct read when opening hdfs file. [Code](https://github.com/apache/hadoop/blob/branch-2.7/hadoop-hdfs-project/hadoop-hdfs/src/main/native/libhdfs/hdfs.c#L1009)

However, Alluxio's HdfsDataInputStream don't implement interface ByteBufferReadable, so a lot of exception logs are seen when applications (like apache Impala) use hdfs c api to open hdfs files.

Exceptions logs are like below:

> UnsupportedOperationException: Byte-buffer read unsupported by input stream

This pull request fix this by implement ByteBufferReadable. In our tests,  it also benefits impala performance (probably because less exception handling and logs).

### DIFF
--- a/core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
+++ b/core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
@@ -12,6 +12,7 @@
 package alluxio.hadoop;
 
 import alluxio.AlluxioURI;
+import alluxio.Constants;
 import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileSystem;
 import alluxio.client.file.FileSystemContext;
@@ -20,6 +21,7 @@ import alluxio.exception.AlluxioException;
 import alluxio.exception.ExceptionMessage;
 import alluxio.exception.FileDoesNotExistException;
 
+import org.apache.hadoop.fs.ByteBufferReadable;
 import org.apache.hadoop.fs.FileSystem.Statistics;
 import org.apache.hadoop.fs.PositionedReadable;
 import org.apache.hadoop.fs.Seekable;
@@ -30,6 +32,7 @@ import java.io.EOFException;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
 
 import javax.annotation.concurrent.NotThreadSafe;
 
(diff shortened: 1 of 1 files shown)


apc999:
Thanks @gengxin for the improvement.  Do you know if this API is available in HDFS Client API since which version? I vaguely recall the reason not having this method is to make Alluxio client (and under storage) compile with  Hadoop 1.x, which is no more a constraint since 2.0. But might be good to double check.

gpang:
@gengxin Can we target this to `master` branch first, and then cherry-pick to other release branches?

gengxin:
> Do you know if this API is available in HDFS Client API since which version?

@apc999 Thanks for the reminding. ByteBufferReadable interface is supported from hadoop [2.0.2-alpha](https://issues.apache.org/jira/browse/HADOOP-8135), so this change will not build on hadoop 1.x version. I also tried to build this change on master branch for the following hadoop versions, all succeeded:

**Apache**
-Phadoop-2 -Dhadoop.version=2.2.0
-Phadoop-3 -Dhadoop.version=3.0.0

**Cloudera**
-Phadoop-2 -Dhadoop.version=2.0.0-cdh4.7.0

**Hortonworks**
-Phadoop-2 -Dhadoop.version=2.1.0.2.0.5.0-67

> which is no more a constraint since 2.0.

Seems [alluxio 2.0 still support hadoop 1.x](https://docs.alluxio.io/os/user/stable/en/contributor/Building-Alluxio-From-Source.html#hadoop-distribution-support), could I submit this change to master branch?

gengxin:
> @gengxin Can we target this to `master` branch first, and then cherry-pick to other release branches?

Sure, I will do it after the compatibility issue with hadoop 1.x is fixed. :)

apc999:
@gengxin thanks for the update. if I merge your PR now, it will break the build for Alluxio with Hadoop-1 profile. @calvinjia , is it time to split multiple hadoop modules now for hadoop 1, hadoop 2 and hadoop 3 on client side? I can see multiple benefits actually by having multiple versions

jiacheliu3:
@apc999 Came across this PR, do we still want to work on this?

apc999:
> @apc999 Came across this PR, do we still want to work on this?
@jiacheliu3 
already implemented  in #13406 . I will close this PR as this is duplicated.

PR #9923, opened 2019-09-20
Remove noisy log in BlockMaster

ns1123:
This line is too noisy and provides little to no valuable information. Will only get worse if default block size is reduced.

PR #9924, opened 2019-09-20
Remove noisy log in BlockMaster

ns1123:
This line is too noisy and provides little to no valuable information.
Will only get worse if default block size is reduced.

pr-link: #9923
change-id: cid-f2fb1d66ddb5601422d35a0a6c34e72e0cd24bc7

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
@@ -892,7 +892,7 @@ public final class DefaultBlockMaster extends AbstractMaster implements BlockMas
         continue;
       }
       synchronized (block) {
-        LOG.info("Block {} is removed on worker {}.", removedBlockId, workerInfo.getId());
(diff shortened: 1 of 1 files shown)


apc999:
@ns1123 should we go through master branch and cherrypick the change to 2.0 and 1.8 rather than starting from 1.8?

gpang:
@ns1123 yes, we have to target this to master branch first, then cherry-pick to other branches

ns1123:
@gpang this already made it to master. Auto-cherrypick wasn't an option. See the change-id and pr link in the description.

gpang:
I auto-cherry-picked it to branch-2.0. looks like 1.8 cannot auto cherry-pick

gpang:
@ns1123 however, you should use the `git cherry-pick` command to create this PR. Then, the change id will get used correctly.

gpang:
@ns1123 shouldn't there be a commit like the original commit? something like this? https://github.com/Alluxio/alluxio/pull/9927/commits

apc999:

Issue #9647, opened 2019-08-05
Improve object store persist performance by avoid extra rename


bf8086:
**Is your feature request related to a problem? Please describe.**
#9417 shows that a persist job on object store write to a temporary file first and the rename it. This is inefficient as object store rename is basically a copy and delete.

**Describe the solution you'd like**
Optimize the persist job so it will write to the final file on UFS directly for object stores. 

**Describe alternatives you've considered**

**Urgency**
Low to Medium, feature will improve overall performance and scalability of cluster.

**Additional context**

PR #9929, opened 2019-09-21
Make Persist jobs write to dest directly on object store

apc999:
Fix https://github.com/Alluxio/alluxio/issues/9647

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -670,21 +670,6 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
-  public static final PropertyKey UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED =
-      new Builder(Name.UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED)
-          .setDefaultValue(true)
-          .setDescription("When set to false, persisting files to object stores in background"
-              + "(e.g., writing files using ASYNC_THROUGH) will first create temporary objects "
-              + "suffixed by \".alluxio.TIMESTAMP.tmp\" in the object store which will then be "
-              + "renamed to the final UFS path. When set to true, the data will be "
-              + "written to the final UFS path directly without staging in the object store first "
-              + "with a temp suffix. Note that enabling the direct persist can significantly "
-              + "speed up the async persist jobs by making less data copy on object store, but "
-              + "may also run the risk of multiple clients racing to write to the same path in "
-              + "Alluxio before the files are fully persisted.")
-          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
-          .setScope(Scope.MASTER)
-          .build();
   public static final PropertyKey UNDERFS_OBJECT_STORE_MULTI_RANGE_CHUNK_SIZE =
       new Builder(Name.UNDERFS_OBJECT_STORE_MULTI_RANGE_CHUNK_SIZE)
           .setDefaultValue(String.format("${%s}", Name.USER_BLOCK_SIZE_BYTES_DEFAULT))
@@ -1907,6 +1892,22 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDefaultValue(true)
           .setDescription("Whether to check for update availability.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_UNSAFE_DIRECT_PERSIST_OBJECT_ENABLED =
+      new Builder(Name.MASTER_UNSAFE_DIRECT_PERSIST_OBJECT_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("When set to false, writing files using ASYNC_THROUGH or persist CLI "
+              + "with object stores as the UFS will first create temporary objects "
+              + "suffixed by \".alluxio.TIMESTAMP.tmp\" in the object store before "
+              + "committed to the final UFS path. When set to true, files will be "
+              + "put to the destination path directly in the object store without staging "
+              + "with a temp suffix. Enabling this optimization by directly persisting files "
+              + "can significantly improve write performance by making less data copy on object "
+              + "store as rename in object store can be slow, "
+              + "but also run the risk when multiple clients are racing to write to the "
+              + "same path in Alluxio but before the files are fully persisted to UFS.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
   public static final PropertyKey MASTER_WEB_BIND_HOST =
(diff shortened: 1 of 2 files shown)


apc999:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3970,11 +3970,18 @@ private void handleReady(long fileId) throws AlluxioException, IOException {
       try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
         // If previous persist job failed, clean up the temporary file.
         cleanup(ufsResource.get(), tempUfsPath);
+        // if the persist destination is on object store, let persist job copy files to destination
+        // directly
+        if (ServerConfiguration.getBoolean(PropertyKey.UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED)
+          && ufsResource.get().isObjectStorage()) {
```
calvinjia: indentation is incorrect here.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -670,6 +670,21 @@ public String toString() {
...
+      new Builder(Name.UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("When set to false, persisting files to object stores in background"
+              + "(e.g., writing files using ASYNC_THROUGH) will first create temporary objects "
+              + "suffixed by \".alluxio.TIMESTAMP.tmp\" in the object store and then renamed to "
+              + "the final UFS path in object store; when set to true, the data will be "
+              + "written to the final UFS path directly without staging in the object store first "
+              + "with a temp suffix. Note that, enabling the direct persist can significantly "
+              + "speed up the async persist jobs by making less data copy on object store, but "
+              + "may also have the risk for multiple clients racing to write files to the "
```
ns1123: "may also run the risk of"

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3970,11 +3970,18 @@ private void handleReady(long fileId) throws AlluxioException, IOException {
       try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
         // If previous persist job failed, clean up the temporary file.
         cleanup(ufsResource.get(), tempUfsPath);
+        // if the persist destination is on object store, let persist job copy files to destination
+        // directly
+        if (ServerConfiguration.getBoolean(PropertyKey.UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED)
```
ns1123: I think the conditional should encompass the cleanup() logic in the previous line. If this property assumes direct writes, we shouldn't be checking to for cleanup of tmp files.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -670,6 +670,21 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED =
+      new Builder(Name.UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("When set to false, persisting files to object stores in background"
+              + "(e.g., writing files using ASYNC_THROUGH) will first create temporary objects "
+              + "suffixed by \".alluxio.TIMESTAMP.tmp\" in the object store and then renamed to "
```
ns1123: "which will then be renamed to"

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -670,6 +670,21 @@ public String toString() {
...
+              + "the final UFS path in object store; when set to true, the data will be "
```
ns1123: Just use a "." instead of a ";"

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -670,6 +670,21 @@ public String toString() {
...
+              + "written to the final UFS path directly without staging in the object store first "
+              + "with a temp suffix. Note that, enabling the direct persist can significantly "
```
ns1123: "Note that enabling"

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
apc999: this should be good. the cleanup is to remove the left over from previous runs before scheduling a new persist job

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
apc999: fixed

apc999:
Jenkins, test this please

apc999:
Jenkins, test this please

gpang:
@apc999 I'm not sure why we want to introduce this unsafe mode, especially since the persist is async (not blocking the client).

apc999:
@gpang it is indeed an unsafe mode, but I think in reality the chance for a workload hitting the unsafe part may be low if users are running ELT workloads. In addition, it takes locking resource from master and requires S3 copy. So I view this is an optimization worth providing to users if they understand their workloads.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3806,6 +3821,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String UNDERFS_VERSION = "alluxio.underfs.version";
     public static final String UNDERFS_OBJECT_STORE_BREADCRUMBS_ENABLED =
         "alluxio.underfs.object.store.breadcrumbs.enabled";
+    public static final String UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED =
+        "alluxio.underfs.object.store.direct.persist.enabled";
```
gpang: This makes is sound like this is a performance optimization, but in reality, it is an unsafe operation. Maybe we should use the word "unsafe" in this parameter, and make the lack of safety more clear in the description.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -670,6 +670,21 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_OBJECT_STORE_DIRECT_PERSIST_ENABLED =
```
gpang: If this is an optimization (but unsafe), then why is it isolated to object stores? Why can't any ufs use this mode?

apc999: currently, we check the ufs type. in case it is object store and this flag is true, we skip the tmp files; otherwise temp files are still used. Are you suggesting to enable this regardless of ufs type?

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: I already put "unsafe" in the property description. do you feel strong about this @gpang ?

gpang: The description does not make it explicit that it is an unsafe setting. I don't think `risk of multiple clients racing` makes it obvious that this is unsafe and can cause inconsistencies.

core/common/src/main/java/alluxio/conf/PropertyKey.java
gpang: Yeah, I'm asking why this cannot be possible for any persist job, instead of only for object stores?

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: renamed to `alluxio.master.unsafe.direct.persist.object.enabled`

apc999:
@gpang @calvinjia change this property to be a master-side property for object store based on my discussion with Calvin.

calvinjia:

rastogiasr:
@apc999 - can this be merged now?

PR #9931, opened 2019-09-21
Add a list of files of Public APIs

apc999:


calvinjia:
@apc999 Do we automatically check if the file is consistent with the annotations?

apc999:
@calvinjia i was using this command `git grep "@PublicApi" | cut -d : -f 1 | sort -n | uniq > PublicApi.txt` to auto gen this list

calvinjia:

PR #9932, opened 2019-09-21
[WIP] Jobs are disappearing from the job master

ZacBlanco:
Run the FileOutStreamAsyncWriteIntegrationTest#badSpaceReserverTest

The test will trigger many jobs to run in the job service. It also sets
the max limit to only 500 jobs, but a retention of 0 seconds, so new jobs
coming in can run so long as there are jobs that have finished.

The issue we're seeing is that when we have many new jobs come in
Some will get scheduled, finish their tasks, and then be put into the
mFinishedJobs map with the time they finished. However, after extensive
testing I still see ResourceExhaustedExceptions being thrown, and when
they are thrown the state of the mIdToJobCoodinator map contains many
(hundreds) of jobs which say completed.

Logging has been added for the time when each job id gets added to
mFinished, removed from mFinished, and also removed from
mJobIdToCoordinator.

Essentially all the logs have shown me is that there are a number
of jobs which are created, run, finishes its task, gets added to
mFinished, but never logs a removal from the map.

The exception is only thrown when the map is empty. It seems that
we're having some kind of "ghost removal".

### DIFF
--- a/job/server/src/main/java/alluxio/master/job/JobCoordinator.java
+++ b/job/server/src/main/java/alluxio/master/job/JobCoordinator.java
@@ -24,6 +24,7 @@ import alluxio.master.job.command.CommandManager;
 import alluxio.wire.WorkerInfo;
 
 import com.google.common.base.Function;
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import org.slf4j.Logger;
@@ -153,10 +154,12 @@ public final class JobCoordinator {
    * @param taskInfoList List of @TaskInfo instances to update
    */
   public synchronized void updateTasks(List<TaskInfo> taskInfoList) {
-    for (TaskInfo taskInfo : taskInfoList) {
-      mJobInfo.setTaskInfo(taskInfo.getTaskId(), taskInfo);
+    synchronized (mJobInfo) {
+      for (TaskInfo taskInfo : taskInfoList) {
+        mJobInfo.setTaskInfo(taskInfo.getTaskId(), taskInfo);
+      }
+      updateStatus();
     }
-    updateStatus();
   }
 
   /**
@@ -164,6 +167,13 @@ public final class JobCoordinator {
    */
   public synchronized boolean isJobFinished() {
     return mJobInfo.getStatus().isFinished();
+  }
+
+  /**
+   * @return the id corresponding to the job
+   */
+  public long getJobId() {
+    return mJobInfo.getId();
   }
 
   /**
@@ -269,5 +279,24 @@ public final class JobCoordinator {
       taskResults.put(mTaskIdToWorkerInfo.get(taskInfo.getTaskId()), taskInfo.getResult());
     }
     return definition.join(mJobInfo.getJobConfig(), taskResults);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+
+    if (!(o instanceof JobCoordinator)) {
+      return false;
+    }
+
+    JobCoordinator other = (JobCoordinator) o;
+    return Objects.equal(mJobInfo, other.mJobInfo);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(mJobInfo);
   }
 }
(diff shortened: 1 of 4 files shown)


ZacBlanco:
Here's a more concrete example of what I'm talking about:

- Reference Bad Job ID: `1569098559715`
- Reference Good Job ID: ``

Search for the bad Job ID in the uploaded log. You'll find these messages:
```log
2019-09-21 13:42:52,700 INFO  command.CommandHandlingExecutor (CommandHandlingExecutor.java:run) - Received run task 0 for job 1569098559715 on worker 1569098549116
2019-09-21 13:42:52,700 INFO  task.TaskExecutorManager (TaskExecutorManager.java:executeTask) - Task 0 for job 1569098559715 started
2019-09-21 13:42:52,796 INFO  task.TaskExecutorManager (TaskExecutorManager.java:notifyTaskCompletion) - Task 0 for job 1569098559715 completed.
2019-09-21 13:42:53,693 ERROR job.JobMaster (JobMaster.java:lambda$run$0) - mFinished.put before - STATUS: COMPLETED - size: 1 - id: 1569098559715
2019-09-21 13:42:53,693 ERROR job.JobMaster (JobMaster.java:lambda$run$0) - mFinished.put after - STATUS: COMPLETED - size: 2 - id: 1569098559715
```

However, note that towards the end of a log, we throw a `ResourceExhaustedException` which indicates that the `mFinishedJobs` queue is empty. Also note, there was never a log for removal from `mFinishedJobs` nor `mIdToCoordinator`.

Now take the good Job ID provided:

```log
2019-09-21 13:42:52,700 INFO  command.CommandHandlingExecutor (CommandHandlingExecutor.java:run) - Received run task 0 for job 1569098559714 on worker 1569098549116
2019-09-21 13:42:52,700 INFO  task.TaskExecutorManager (TaskExecutorManager.java:executeTask) - Task 0 for job 1569098559714 started
2019-09-21 13:42:52,792 INFO  task.TaskExecutorManager (TaskExecutorManager.java:notifyTaskCompletion) - Task 0 for job 1569098559714 completed.
2019-09-21 13:42:53,692 ERROR job.JobMaster (JobMaster.java:lambda$run$0) - mFinished.put before - STATUS: COMPLETED - size: 0 - id: 1569098559714
2019-09-21 13:42:53,692 ERROR job.JobMaster (JobMaster.java:lambda$run$0) - mFinished.put after - STATUS: COMPLETED - size: 1 - id: 1569098559714
2019-09-21 13:42:54,089 ERROR job.JobMaster (JobMaster.java:run) - mFinished.remove removed jobId 1569098559714
2019-09-21 13:42:54,089 ERROR job.JobMaster (JobMaster.java:run) - mIdToJobCoord.remove job id: 1569098559714
```

Here, there are 7 lines, with the last two indicating their proper removal from tracking. I'm still trying to figure out why exactly this is happening.

job/server/src/main/java/alluxio/master/job/JobCoordinator.java
```diff
@@ -153,10 +154,12 @@ public synchronized void cancel() {
    * @param taskInfoList List of @TaskInfo instances to update
    */
   public synchronized void updateTasks(List<TaskInfo> taskInfoList) {
-    for (TaskInfo taskInfo : taskInfoList) {
-      mJobInfo.setTaskInfo(taskInfo.getTaskId(), taskInfo);
+    synchronized (mJobInfo) {
```
madanadit: this and following synchronization added is redundant? the previous way of synchronization is not fine grained but should be functionally correct by synchronizing on the JobCoordinator instance?

ZacBlanco:
Closing this as we've found the root cause of the issue to be that the implementors of `SortedSet` in the Java libraries utilize the `Comparable<T>` interface to determine equality of objects rather than `equals(Object)`.

What ended up happening was that when the job master was under high load with many small jobs, multiple jobs would come in and be marked as completed. However, when marking a job as completed we utilized the method `CommonUtils#getCurrentMs` to set the lastChangeMs time which did not provide good enough granularity to ensure that all jobs inserted into the finished set were unique.

That meant when inserting, some jobs would be inadvertently removed if the newly finished job had the same completion time. This resulted in the `compareTo` implementation resolving the two JobInfos as being equivalent, even though the references were not.

https://github.com/Alluxio/alluxio/pull/9934 Fixes this issue as well as trying to simplify how jobs are tracked, as well as improve error handling when near capacity.

PR #9942, opened 2019-09-23
Update outdated dependencies: jetty, jersey and others

witgo:


apc999:
@witgo thanks for updating the dependency. In general in this project, we update the version of an external dependency if we hit bugs, security or performance issues on the existing ones. Otherwise, updating the versions of other client-side dependencies (e.g., alluxio-client may depend on) is not typically encouraged because they may cause dependency conflicts on application side. On server-side, the requirement is less strict. 

Among the dependencies updated in this PR, do you have them in one of the above buckets?

witgo:
@apc999 

grpc =>  DoS vulnerability CVE-2019-9515 (SETTINGS flood)

rocksdb =>  ARM

jersey => jdk 11

core/server/master/pom.xml
```diff
@@ -70,6 +70,10 @@
       <groupId>org.glassfish.jersey.core</groupId>
       <artifactId>jersey-server</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.inject</groupId>
```
yuzhu: is this package used anywhere?

pom.xml
```diff
@@ -675,11 +691,10 @@
         <version>${hadoop.version}</version>
         <scope>test</scope>
       </dependency>
-      <!-- TODO(andrew): Update to a non-rc version before final Alluxio release -->
```
yuzhu: nice catch..oops

core/server/master/pom.xml
witgo: http://blog.supol.cz/?p=45

witgo:
@ZacBlanco Can you take a look?

job/server/pom.xml
```diff
@@ -52,6 +52,11 @@
       <artifactId>jersey-container-servlet-core</artifactId>
       <version>${jersey.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.jaxrs</groupId>
+      <artifactId>jackson-jaxrs-json-provider</artifactId>
+      <version>${jackson.version}</version>
+    </dependency>
```
ZacBlanco: this dependency and version should be specified in the root pom's `<dependencyManagement>` section, and the version can be left out here.

pom.xml
```diff
@@ -620,12 +629,12 @@
...
         <groupId>org.reflections</groupId>
         <artifactId>reflections</artifactId>
-        <version>0.9.10</version>
+        <version>0.9.11</version>
       </dependency>
       <dependency>
         <groupId>org.rocksdb</groupId>
         <artifactId>rocksdbjni</artifactId>
-        <version>5.15.10</version>
+        <version>6.6.4</version>
```
ZacBlanco: This dependency grows from 13MB to ~28MB according to this: https://mvnrepository.com/artifact/org.rocksdb/rocksdbjni

Just something to note

Additionally, upgrading a major version for rocksdb means there can be incompatibilities in the on-disk formats between versions and could result in backcompat issues for people using rocksdb. I think the highest version we can upgrade to is 5.18.3

pom.xml
```diff
@@ -128,24 +128,21 @@
...
+    <jersey.version>2.29.1</jersey.version>
+    <jetty.version>9.2.28.v20190418</jetty.version>
     <junit.version>4.12</junit.version>
     <log4j.version>1.2.17</log4j.version>
     <maven.version>3.3.9</maven.version>
     <metrics.version>3.1.2</metrics.version>
     <oshi.version>4.2.0</oshi.version>
     <powermock.version>1.6.2</powermock.version>
-    <prometheus.version>0.3.0</prometheus.version>
+    <prometheus.version>0.8.0</prometheus.version>
```
ZacBlanco: Are there any changes in the format in which metrics might be reported from 0.3 to 0.8? According to semver, Upgrades from `0.x` to `0.(x+1)` may not have a stable API

https://semver.org/#spec-item-4

witgo: Mainly new features and bug fixes, it looks like there are no compatibility issues

https://github.com/prometheus/client_java/releases

pom.xml
witgo: The new rocksdb version mainly adds support for arm and musl-libc. I changed the code to support a custom rocksdb version.

job/server/pom.xml
witgo: done

pom.xml
```diff
@@ -282,11 +280,6 @@
         <artifactId>commons-logging</artifactId>
         <version>1.1.1</version>
       </dependency>
-      <dependency>
-        <groupId>info.ganglia.gmetric4j</groupId>
-        <artifactId>gmetric4j</artifactId>
-        <version>${gmetric4j.version}</version>
-      </dependency>
```
ZacBlanco: Is this dependency not used in the codebase?

pom.xml
```diff
@@ -620,12 +635,12 @@
       <dependency>
         <groupId>org.reflections</groupId>
         <artifactId>reflections</artifactId>
-        <version>0.9.10</version>
+        <version>0.9.11</version>
```
ZacBlanco: looks like the latest is actually 0.9.12, can we bump that to the latest version?

pom.xml
```diff
@@ -668,11 +689,10 @@
         <version>${hadoop.version}</version>
         <scope>test</scope>
       </dependency>
-      <!-- TODO(andrew): Update to a non-rc version before final Alluxio release -->
       <dependency>
         <groupId>org.hamcrest</groupId>
         <artifactId>hamcrest</artifactId>
-        <version>2.1-rc4</version>
+        <version>2.1</version>
```
ZacBlanco: 2.2 seems to be the latest now. Can we bump this?

pom.xml
```diff
@@ -321,11 +314,23 @@
         <groupId>io.grpc</groupId>
         <artifactId>grpc-netty</artifactId>
         <version>${grpc.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>io.netty</groupId>
+            <artifactId>*</artifactId>
+          </exclusion>
+        </exclusions>
```
ZacBlanco: what's the reason for adding this now?

pom.xml
```diff
@@ -321,11 +314,23 @@
...
       <dependency>
         <groupId>io.grpc</groupId>
         <artifactId>grpc-protobuf</artifactId>
         <version>${grpc.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>com.google.api.grpc</groupId>
+            <artifactId>proto-google-common-protos</artifactId>
+          </exclusion>
+        </exclusions>
```
ZacBlanco: what's the reason for adding this now?

core/server/master/pom.xml
ZacBlanco: I think this link provides a better explanation: https://eclipse-ee4j.github.io/jersey.github.io/release-notes/2.26.html

pom.xml
witgo: There is an explicit dependency on `netty-all`, this is redundant.

pom.xml
witgo: alluxio does not use this jar related code

pom.xml
witgo: yes.

witgo:
The failed UT looks irrelevant

ZacBlanco:
jenkins, test this please

pom.xml
ZacBlanco: I think it is better to specifically exclude the explicit dependencies that we rely on, rather than just `*`?

What is your opinion?

ZacBlanco:

Issue #9910, opened 2019-09-18
Prevent accumulating status objects during listStatus


ggezer:
**Is your feature request related to a problem? Please describe.**
Currently, `listStatus` handler first receives the whole list from master and then sends it back to client in chunks. This approach will not be efficient for a very large namespace as it practically doubles the memory during the call.

**Describe the solution you'd like**
Make FileSystemMaster accepts an interface through which it could send chunks back to client without a need to serve the complete result as list.

**Urgency**
`listStatus` on a large namespace puts master to memory strain, potentially could crash it.

PR #9949, opened 2019-09-23
Prevent accumulating status objects on master during listStatus

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/9910


core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
```diff
@@ -140,6 +140,26 @@ FileInfo getFileInfo(AlluxioURI path, GetStatusContext context)
...
+   *
+   * @param path the path to get the {@link FileInfo} list for
+   * @param context the method context
+   * @param batchTracker the tracker that will receive batches of statuses
+   * @throws AccessControlException if permission checking fails
+   * @throws FileDoesNotExistException if the file does not exist
+   * @throws InvalidPathException if the path is invalid
+   */
+  void listStatus(AlluxioURI path, ListStatusContext context,
+      BatchTracker<ListStatusBatch> batchTracker) throws AccessControlException,
```
gpang: Why do we need an intermediate class like `ListStatusBatch` and not just use `List<FileInfo>` directly?

ggezer: Otherwise I'd need to pass both `BatchTracker<List<FileInfo>>` and an `Integer` for denoting the batch size. `ListStatusBatch` combines the data and the control.

core/server/master/src/main/java/alluxio/master/file/BatchTracker.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package alluxio.master.file;
+
+/**
+ * Used to receive batches from FileSystemMaster methods.
+ *
+ * @param <T> batch type
+ */
+public interface BatchTracker<T> {
```
gpang: what if the interface was just a streaming interface, and one such implementation is the batched version? So, we create something like:
```
interface ResultStream<T> {
  void submit(T item);
}
```

Then, we have a new `BatchedResultStream(batchSize)` which has a `List<T>`, and when it gets full, it sends it? Then, the whole batching concept is isolated in the `BatchedResultStream`, and the users don't have to decide whether or not to submit a batch.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -887,38 +887,40 @@ public PersistenceState getPersistenceState(long fileId) throws FileDoesNotExist
...
+   * descendantType. The result is returned via a list specified by statusList, in postorder
    * traversal order.
    *
    * @param rpcContext the context for the RPC call
    * @param currInodePath the inode path to find the status
    * @param auditContext the audit context to return any access exceptions
    * @param descendantType if the currInodePath is a directory, how many levels of its descendant
-   *                       should be returned
-   * @param statusList To be populated with the status of the files and directories requested
+   *        should be returned
```
gpang: can we undo this line change?

core/server/master/src/main/java/alluxio/master/file/ListStatusResultStream.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Used to define a single batch of listing.
+ */
+@ThreadSafe
+public class ListStatusResultStream implements ResultStream<FileInfo> {
+  /** List of file infos. */
+  private List<FileInfo> mInfos;
```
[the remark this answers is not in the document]
ggezer: It's possible but would not change the cost. Currently proto conversion is restricted to toProto() method, which makes it easier for when we remove the proto types after this PR.

ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please.

ggezer:
@alluxio-bot, merge this please

gpang:

PR #9951, opened 2019-09-23
Add explicit dependency on netty-native

calvinjia:
Presto imports netty packages individually, so we exclude netty-all. We could also explicitly import each package, but the only one we need over what is provided is the native module for domain socket.

### DIFF
--- a/experimental/table/client/pom.xml
+++ b/experimental/table/client/pom.xml
@@ -36,6 +36,10 @@
       <artifactId>netty-all</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-unix-common</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


experimental/table/client/pom.xml
```diff
@@ -35,6 +35,10 @@
       <groupId>io.netty</groupId>
       <artifactId>netty-all</artifactId>
```
yuzhu: so do we still need this `netty-all` dependency?

calvinjia: Yes, `netty-all` includes `netty-native`, but we need to explicitly add `netty-native` because we exclude `netty-all` in Presto. The way they define netty dependencies is each module individually (whereas we just include all), this means for the packages in `all` that are not included in Presto, we need to explicitly depend on them.

yuzhu: got it

calvinjia:

PR #9934, opened 2019-09-22
Fix tracking of finished jobs within JobMaster

ZacBlanco:
After some investigation it was found that the JobMaster was not properly
tracking finished jobs due to discrepancies in the implementation of the 
`compareTo` and `equals` methods of `JobInfo`. This led to jobs being
removed from the ConcurrentSkipListSet without ever actually calling
the `remove` function. The resulting behavior is that we would see the
master hit the maximum job capacity very frequently as there would be
un-flushable jobs in the `mIdToCoordinators` map.

This change fixes the original issue by altering the underlying structure
used to keep track of the finished jobs and by making more guarantees
about how the status of jobs may change.

First, with this change we no longer allow a JobInfo to change states after
it has been marked with a state where `JobInfo#isFinished` returns true.
This means all jobs states should now follow a DAG (RUNNING is an
optional state as a job may move from CREATED to any one of the finished
states):

```
CREATED -> (RUNNING) -> [FAILED|CANCELLED|COMPLETED]
```

Second, instead of utilizing a ConcurrentSkipListSet I've opted for a
simpler concurrent structure; the LinkedBlockingQueue. Using a FIFO
queue with the guarantee that once jobs are finished they cannot change
states, we gain the opportunity to use a simple queue that has O(1)
offer/poll operations. With this approach jobs that are marked as finished
may get added in different orders depending on how the internal locks
within the queue operate. This implementation no longer strictly guarantees
that objects in the queue are ordered by the lastStatusChangeMs time. The
amount of time difference between concurrent jobs being added should not
be so great that leaves jobs such that the retention time severely affects
the amount of jobs that can be evicted.

This implementation introduces the JobTracker class which is now
responsible for adding and evicting jobs from the job master. It encapsulates
what was the mIdToCoordinator map, and also houses the FIFO queue that
used to be the mFinishedJobs set.

Closes #9874

PR #9957, opened 2019-09-24
Fix bug for job info loss

liuhongtong:
mFinishedJobs in JobMaster is a ConcurrentSkipListSet to store finished
JobInfo. JobInfo implements the compareTo() function to compare the
JobInfos by mLastStatusChangeMs. For two different JobInfos, if the
mLastStatusChangeMs in is the same, there is only one JobInfo stored in
mFinishedJobs after the JobInfos added and one is lost.

mId in JobInfo is unique and can be used to compare JobInfos.

Signed-off-by: liuhongtong <hongtongliu@126.com>

### DIFF
--- a/job/common/src/main/java/alluxio/job/meta/JobInfo.java
+++ b/job/common/src/main/java/alluxio/job/meta/JobInfo.java
@@ -64,7 +64,7 @@ public final class JobInfo implements Comparable<JobInfo> {
    */
   @Override
   public synchronized int compareTo(JobInfo other) {
-    return Long.compare(mLastStatusChangeMs, other.getLastStatusChangeMs());
+    return Long.compare(mId, other.getId());
   }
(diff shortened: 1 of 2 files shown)


ZacBlanco:
@liuhongtong Thanks for trying to improve this!

One issue I see is that the `mFinishedJobs` set will no longer be ordered according to the oldest finished job, which the removal logic relies upon to purge as many jobs as possible when reaching capacity. While Job Ids are monotonically increasing, ordering by them is not the same guarantee as ordering by status change time.

We're working on a more comprehensive fix in https://github.com/Alluxio/alluxio/pull/9934 Which will likely be merged some time soon. This PR won't be necessary afterwards.

ZacBlanco:
I'm closing this as #9934 has been merged

PR #9966, opened 2019-09-25
Enable Alluxio to run as non-root in Kubernetes cluster

jiacheliu3:
This change will enable Alluxio to run as non-root user in Kubernetes cluster. 

The UFS local journal directory and the domain socket directory will have to be `chown` to `alluxio:alluxio`. This is done by adding `initContainers` which are executed before the containers. The `securityContext` are split up and moved to each container to provider security configuration in finer granularity.

Ref: [Kubernetes security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)

### DIFF
--- a/integration/kubernetes/helm/alluxio/templates/alluxio-master.yaml
+++ b/integration/kubernetes/helm/alluxio/templates/alluxio-master.yaml
@@ -10,6 +10,8 @@
 #
 
 {{ $root := . -}}
+{{- $isRoot := eq (int $root.Values.user) 0 }}
+{{- $ugi := printf "%v:%v" $root.Values.user $root.Values.group }}
 {{- $masterCount := int .Values.resources.masterCount }}
 {{- $isSingleMaster := eq $masterCount 1 }}
 {{- $isEmbedded := (eq .Values.journal.type "EMBEDDED") }}
@@ -17,7 +19,7 @@
 {{- $isUfsLocal := and (eq .Values.journal.type "UFS") (eq .Values.journal.ufsType "local") }}
 {{- $isSingleUfsLocal := and $isUfsLocal $isSingleMaster }}
 {{- $needJournalVolume := or $isEmbedded $isUfsLocal }}
-{{- if $isUfsLocal}}
+{{- if $isUfsLocal }}
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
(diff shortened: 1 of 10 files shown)


integration/kubernetes/helm/alluxio/templates/alluxio-master.yaml
```diff
@@ -70,9 +72,16 @@ spec:
         app: {{ $masterName }}
     spec:
       securityContext:
-        runAsUser: {{ $root.Values.user }}
-        runAsGroup: {{ $root.Values.group }}
         fsGroup: {{ $root.Values.fsGroup }}
+      {{- if and $isSingleUfsLocal (not $isRoot) }}
+      initContainers:
+      - name: journal-chown
+        image: busybox
```
madanadit: we should not create a dependency on a new image. modify the Dockerfile if needed to install `sudo` by `apk --no-cache --update add sudo`

integration/kubernetes/helm/alluxio/templates/alluxio-worker.yaml
```diff
@@ -26,9 +28,16 @@ spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       securityContext:
-        runAsUser: {{ .Values.user }}
-        runAsGroup: {{ .Values.group }}
         fsGroup: {{ .Values.fsGroup }}
+      {{- if not $isRoot }}
+      initContainers:
+        - name: socket-chown
+          image: busybox
```
madanadit: same here

integration/kubernetes/multiMaster-embeddedJournal/alluxio-worker.yaml.template
```diff
@@ -11,6 +11,7 @@
 # See the NOTICE file distributed with this work for information regarding copyright ownership.
 #
 
+
```
madanadit: get rid of this newline

integration/kubernetes/helm/alluxio/templates/alluxio-master.yaml
jiacheliu3: Turned out I cannot `sudo` in initContainer (maybe not even possible in container commands in general). So I had to specify root in initContainer as following.
```
securityContext:
  runAsUser: 0
```

madanadit:

PR #9971, opened 2019-09-25
Use better locking in JobInfo to avoid deadlocks

### LINKED ISSUE #9970: The Job master has locking inconsistencies which can cause deadlock
**Alluxio Version:**
This is in the master branch, 2.1.0-SNAPSHOT

**Describe the bug**
A deadlock can happen in the job master, because of the wrong locking order. Here are the stacks for 2 deadlocked threads.
```
"grpc-default-executor-12":
	at alluxio.master.job.JobCoordinator.updateStatus(JobCoordinator.java:228)
	- waiting to lock <0x00000006ccf6bdb8> (a alluxio.master.job.JobCoordinator)
	at alluxio.master.job.JobCoordinator.updateTasks(JobCoordinator.java:161)
	- locked <0x00000006ccf6bde0> (a alluxio.job.meta.JobInfo)
	at alluxio.master.job.JobMaster.workerHeartbeat(JobMaster.java:322)
"grpc-default-executor-7":
	at alluxio.job.meta.JobInfo.getId(JobInfo.java:92)
	- waiting to lock <0x00000006ccf6bde0> (a alluxio.job.meta.JobInfo)
	at alluxio.job.wire.JobInfo.<init>(JobInfo.java:50)
	at alluxio.master.job.JobCoordinator.getJobInfoWire(JobCoordinator.java:220)
	- locked <0x00000006ccf6bdb8> (a alluxio.master.job.JobCoordinator)
	at alluxio.master.job.JobMaster.getStatus(JobMaster.java:242)
```

**To Reproduce**
Unfortunately, the deadlock is difficult to reliably reproduce. I hit it randomly by running the PersistCommandTest.

**Expected behavior**
Avoid deadlock.

**Urgency**
Deadlocks can stall getStatus calls and worker heartbeats, which can stall persist jobs or replication jobs. Deadlocks must be avoid.

gpang:
JobInfo had unnecessary synchronized blocks which caused deadlocks. `final` variables do not need `synchronized` block, and converted some to concurrent data structures.

fixes #9970

### DIFF
--- a/job/common/src/main/java/alluxio/job/meta/JobInfo.java
+++ b/job/common/src/main/java/alluxio/job/meta/JobInfo.java
@@ -82,6 +82,7 @@ public final class JobInfo implements Comparable<JobInfo> {
     TaskInfo oldValue = mTaskIdToInfo.putIfAbsent(taskId,
         new TaskInfo().setJobId(mId).setTaskId(taskId).setStatus(Status.CREATED).setErrorMessage("")
             .setResult(null));
+    // the task is expected to not exist in the map.
     Preconditions.checkState(oldValue == null,
         String.format("JobId %d cannot add duplicate taskId %d", mId, taskId));
   }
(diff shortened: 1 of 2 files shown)


job/common/src/main/java/alluxio/job/meta/JobInfo.java
```diff
@@ -79,52 +78,54 @@ public synchronized int compareTo(JobInfo other) {
...
    */
-  public synchronized void addTask(int taskId) {
-    Preconditions.checkArgument(!mTaskIdToInfo.containsKey(taskId), "");
-    mTaskIdToInfo.put(taskId, new TaskInfo().setJobId(mId).setTaskId(taskId)
-        .setStatus(Status.CREATED).setErrorMessage("").setResult(null));
+  public void addTask(int taskId) {
+    TaskInfo oldValue = mTaskIdToInfo.putIfAbsent(taskId,
+        new TaskInfo().setJobId(mId).setTaskId(taskId).setStatus(Status.CREATED).setErrorMessage("")
+            .setResult(null));
+    Preconditions.checkState(oldValue == null,
```
ZacBlanco: Is this supposed to be `!=`? It will always be `oldValue == null` will always evaluate to `true` if the job doesn't exist yet

ZacBlanco:
```suggestion
    Preconditions.checkState(oldValue =! null,
```

ZacBlanco: Also, in the case this does happen, will the caller be able to handle the exception gracefully?

gpang: Well, we expect it to not exist in the map, so we expect the old value to be null.

Also, there was already a `checkArgument` that basically checked the same condition, so I didn't introduce a new precondition.

gpang:

gpang:

PR #9975, opened 2019-09-26
Adjust embedded journal properties

ggezer:
`HEARTBEAT_INTERVAL` to `3sec`: Heartbeat controls the frequency of communication between copycat nodes. So a larger value would stall entire communication.

`ELECTION_TIMEOUT` to `10sec` : Leader election timeout is effective during cluster startup as well.  A larger value would have been a bad user experience.

`APPENDER_BATCH_SIZE` to `512KB` : 512KB of data at every heartbeat (3sec) would utilize network bandwith better than `32KB` every `1sec` as before.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1280,7 +1280,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDescription(
               "The election timeout for the embedded journal. When this period elapses without a "
                   + "master receiving any messages, the master will attempt to become the primary.")
-          .setDefaultValue("5s")
+          .setDefaultValue("10s")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
@@ -1289,7 +1289,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               "The period between sending heartbeats from the embedded journal primary to "
                   + "followers. This should be less than half of the election timeout "
                   + "(alluxio.master.embedded.journal.election.timeout).")
-          .setDefaultValue("1s")
+          .setDefaultValue("3s")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE =
@@ -1298,7 +1298,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               + "in a single heartbeat. Setting higher values might require increasing "
               + "election timeout due to increased network delay. Setting lower values "
               + "might stall knowledge propagation between the leader and followers.")
-          .setDefaultValue("32KB")
+          .setDefaultValue("512KB")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_PORT =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1281,7 +1281,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               "The election timeout for the embedded journal. When this period elapses without a "
                   + "master receiving any messages, the master will attempt to become the primary."
                   + "Election timeout will be waited initially when the cluster is forming. "
-                  + "So larger values for election timeout will cause longer star-up time.")
+                  + "So larger values for election timeout will cause longer star-up time. "
+                  + "Smaller values might introduce instability to leadership.")
           .setDefaultValue("10s")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1280,23 +1280,26 @@ public String toString() {
           .setDescription(
               "The election timeout for the embedded journal. When this period elapses without a "
                   + "master receiving any messages, the master will attempt to become the primary.")
-          .setDefaultValue("5s")
+          .setDefaultValue("10s")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
```
gpang: shouldn't we also `.setScope(Scope.MASTER)` for all of these?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1280,23 +1280,26 @@ public String toString() {
...
           .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL)
           .setDescription(
               "The period between sending heartbeats from the embedded journal primary to "
                   + "followers. This should be less than half of the election timeout "
                   + "(alluxio.master.embedded.journal.election.timeout).")
```
gpang: can you add what the implications are for a short interval and a long interval?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1279,24 +1279,32 @@ public String toString() {
...
+          .setScope(Scope.MASTER)
           .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL)
           .setDescription(
               "The period between sending heartbeats from the embedded journal primary to "
                   + "followers. This should be less than half of the election timeout "
-                  + "(alluxio.master.embedded.journal.election.timeout).")
-          .setDefaultValue("1s")
+                  + "(alluxio.master.embedded.journal.election.timeout).Because the election "
```
gpang:
```suggestion
                  + "(alluxio.master.embedded.journal.election.timeout), because the election "
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1279,24 +1279,32 @@ public String toString() {
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT)
           .setDescription(
               "The election timeout for the embedded journal. When this period elapses without a "
-                  + "master receiving any messages, the master will attempt to become the primary.")
-          .setDefaultValue("5s")
+                  + "master receiving any messages, the master will attempt to become the primary."
+                  + "Election timeout will be waited initially when the cluster is forming. "
+                  + "So larger values for election timeout will cause longer start-up time.")
```
LuQQiu: Add "Shorter values may cause the master leadership to change more often." ?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1279,24 +1279,32 @@ public String toString() {
...
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL)
           .setDescription(
               "The period between sending heartbeats from the embedded journal primary to "
                   + "followers. This should be less than half of the election timeout "
-                  + "(alluxio.master.embedded.journal.election.timeout).")
-          .setDefaultValue("1s")
+                  + "(alluxio.master.embedded.journal.election.timeout), because the election "
+                  + "is driven by heart beats.")
+          .setDefaultValue("3s")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
```
LuQQiu: Add scope master

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1310,6 +1318,7 @@ public String toString() {
               + "losing state in case of power loss or host failure. Use MEMORY for "
               + "optimal performance, but no state persistence across cluster restarts.")
           .setDefaultValue("DISK")
+          .setScope(Scope.MASTER)
```
LuQQiu: Add consistency check level also be warn?

ggezer: I think we might want to support different storage tiers across cluster.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1279,24 +1279,34 @@ public String toString() {
...
           .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL)
           .setDescription(
               "The period between sending heartbeats from the embedded journal primary to "
                   + "followers. This should be less than half of the election timeout "
-                  + "(alluxio.master.embedded.journal.election.timeout).")
-          .setDefaultValue("1s")
+                  + "(alluxio.master.embedded.journal.election.timeout), because the election "
```
gpang: NIT: instead of hardcoding it, can we just use `String.format()` and programmatically include the actual name?

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Oh, i saw it as the scope of `MASTER_EMBEDDED_JOURNAL_PORT`, never mind,

ggezer:
@alluxio-bot, merge this please

PR #9976, opened 2019-09-26
[DOCFIX] Add doc on how to upgrade Alluxio in K8s

jiacheliu3:
This is a doc change adding doc on how to upgrade Alluxio in a K8s environment.

### DIFF
--- a/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
+++ b/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
@@ -347,24 +347,21 @@ containers:
 
 **Step 2: Stop running Alluxio master and worker Pods**
 
-Kill all running Alluxio worker Pods.
+Kill all running Alluxio worker Pods by deleting its DaemonSet.
 
 ```console
-$ kubectl delete -f alluxio-worker.yaml
+$ kubectl delete daemonset -l app=alluxio
 ```
 
-Then kill all running Alluxio master Pods by killing each StatefulSet and each Service.
+Then kill all running Alluxio master Pods by killing each StatefulSet and each Service with label `app=alluxio`.
 
 ```console
-$ let masterCount = {number of your masters minus 1}
-$ for i in {0..masterCount}; do
-> kubectl delete service alluxio-master-{i}
-> kubectl delete statefulset alluxio-master-{i}
-> done
+$ kubectl delete service -l app=alluxio
+$ kubectl delete statefulset -l app=alluxio
 ```
 
-The reason why you don't want to do `kubectl delete -f alluxio-master.yaml` is that will delete the Persistent Volume Claim in the `alluxio-master.yaml` if any.
-And that will result in your Persistent Volume getting released and the journal in it will be lost.
+The reason why you don't do `kubectl delete -f alluxio-master.yaml` is that will delete the Persistent Volume Claim in the `alluxio-master.yaml` if any.
+And that will result in your Persistent Volume getting released and the journal in it will be lost. The same logic applies to `alluxio-worker.yaml`.
 
 Make sure all the Pods have been terminated before you move on to the next step.
 
--- a/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
+++ b/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
@@ -323,8 +323,6 @@ $ kubectl delete -f alluxio-journal-volume.yaml
 #### Upgrade Alluxio
 
 This section will go over how to upgrade Alluxio in your Kubernetes cluster with `kubectl`.
-This section will only cover the basic procedure of how to kill all existing Pods, upgrade the version and bring the Alluxio master and worker Pods back up.
-More advanced rolling update procedure will be covered in the next section. 
 
 **Step 1: Upgrade the docker image version tag**
  
@@ -347,48 +345,53 @@ containers:
   ...
 ```
 
-**Step 2: Format journal if necessary**
+**Step 2: Stop running Alluxio master and worker Pods**
 
-Note that whether you need to re-format the journal and lose all file metadata depends on whether the journals are compatible.
-If you are upgrading from v1.8 to v2.x, you have to format the journal. And by doing that you lose all the journal and file metadata.
-If you are upgrading from v2.x to a higher minor version v2.y, or from v2.x.y to a higher minor version v2.x.z,
-the higher version Alluxio master will be able to read the lower version Alluxio master journal to recover all the journal and file metadata.
-
-If you need to format the journal, you can update the YAML file for Alluxio master containers to specify this on master startup time.
-Remove the `--no-format` option from Alluxio master startup arguments to force the re-format of journal as illustrated below.
-
-```yaml
-containers:
-- name: alluxio-master
-  command: ["/entrypoint.sh"]
-  # --no-format removed from the args
-  args: ["master-only"]
-  ...
-```
-
-**Step 3: Restart Alluxio master and worker Pods**
-
-Now that Alluxio masters and worker containers all use your desired version. Restart them to let it take effect.
-
-Kill the running Alluxio master and worker Pods first.
+Kill all running Alluxio worker Pods.
 
 ```console
 $ kubectl delete -f alluxio-worker.yaml
-$ kubectl delete service alluxio-master-0
-$ kubeclt delete statefulset alluxio-master-0
 ```
 
-If you have more alluxio-master Pods, delete them together with the corresponding Services.
+Then kill all running Alluxio master Pods by killing each StatefulSet and each Service.
 
 ```console
-$ kubectl delete service alluxio-master-1
-$ kubeclt delete statefulset alluxio-master-1
-$ kubectl delete service alluxio-master-2
-$ kubeclt delete statefulset alluxio-master-2
+$ let masterCount = {number of your masters minus 1}
+$ for i in {0..masterCount}; do
+> kubectl delete service alluxio-master-{i}
+> kubectl delete statefulset alluxio-master-{i}
+> done
 ```
 
 The reason why you don't want to do `kubectl delete -f alluxio-master.yaml` is that will delete the Persistent Volume Claim in the `alluxio-master.yaml` if any.
 And that will result in your Persistent Volume getting released and the journal in it will be lost.
+
+Make sure all the Pods have been terminated before you move on to the next step.
+
+**Step 3: Format journal and Alluxio storage if necessary**
+
+Check the Alluxio upgrade guide page for whether the Alluxio master journal has to be formatted.
+If no format is needed, you are ready to skip the rest of this section and move on to restart all Alluxio master and worker Pods.
+
+If you are running UFS journal, there is one Kubernetes [Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
+template that can be used for only formatting the master.
+You should make sure the Job runs with the same configuration with all your other Alluxio masters so it's able to find the journal persistent storage and format it.  
+
+```console
+$ kubectl apply -f alluxio-format-master.yaml
+```
+
+After the Job completes, it will be deleted by Kubernetes after the defined `ttlSecondsAfterFinished`.
+Then the clean journal will be ready for a new Alluxio master to start with.
+
+If you are running embedded journal and masters write journal each to its persistent storage,
+you should write your own Kubernetes Jobs that go to each of them and perform format.
+
+Once all the journals have been formatted, you are ready to restart the Alluxio master and worker Pods.
+
+**Step 4: Restart Alluxio master and worker Pods**
+
+Now that Alluxio masters and worker containers all use your desired version. Restart them to let it take effect.
 
 Now restart the Alluxio master and worker Pods from the YAML files.
 
@@ -397,54 +400,16 @@ $ kubectl create -f alluxio-master.yaml
 $ kubectl create -f alluxio-worker.yaml
 ```
 
-**Step 4: Verify the Alluxio master and worker Pods are back up**
+**Step 5: Verify the Alluxio master and worker Pods are back up**
 
-What a proper validation should be varies from case to case.
-The example is just verifying that Alluxio is running.
+You should verify the Alluxio Pods are back up in Running status.
 
 ```console
-# You should see Alluxio master and worker Pods in Running status
+# You should see all Alluxio master and worker Pods
 $ kubectl get pods
-# This runs a comprehensive sanity check
-$ kubectl exec <alluxio-master-pod> -- /opt/alluxio/bin/alluxio runTests
 ```
 
-#### Rolling Update Alluxio
-
-In a production environment, it is often more desirable to do rolling updates to maintain the availability.
-
-How the rolling update should be done depends on the journal type of Alluxio masters are configured and what storage Alluxio workers are using.
-We will discuss the scenarios separately.
-
-##### Alluxio Masters
-
-The behavior of Alluxio masters will differ depends on which journal mode they are using. 
-See the detailed discussion in [UFS Journal vs Embedded Journal]({{ '/en/operation/Journal.html#ufs-journal-vs-embedded-journal' | relativize_url }})
-
-###### Case A: Rolling Update with Embedded Journal
-
-We do not recommend doing rolling update with embedded journal. 
-
-The reason is Alluxio masters using embedded journal have to reach a consensus before each journaling to its own journal destination.
-The consensus logic of Alluxio masters can be changed in minor versions so if you have a group of Alluxio masters running different versions,
-the behavior will be undefined. 
-
-###### Case B: Rolling Update with UFS Journal
-
-With UFS journal, there is only one shared journal destination and only the one primary master writes. All the secondary masters only read from it.
-
-You are able to do rolling update among Alluxio masters as below.
-
-**Step 1: Update the secondary master Pods and restart them**
-
-**Step 2: Update the primary master Pods and restart them**
-
-In this way no lower version Alluxio master will read the journal written by a higher version master.
-
-##### Alluxio Workers 
-
-The rolling update of Alluxio workers can go parallel to that of Alluxio masters.
-Since Alluxio workers only communicate to masters, they can be updated individually.
+You can do more comprehensive verification following [Verify Alluxio]({{ '/en/deploy/Running-Alluxio-Locally.html?q=verify#verify-alluxio-is-running' | relativize_url }})
 
 ### Access the Web UI
 
--- a/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
+++ b/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
@@ -384,7 +384,11 @@ Then the clean journal will be ready for a new Alluxio master to start with.
 If you are running embedded journal and masters write journal each to its persistent storage,
 you should write your own Kubernetes Jobs that go to each of them and perform format.
 
-Once all the journals have been formatted, you are ready to restart the Alluxio master and worker Pods.
+If you are running Alluxio workers with [tiered storage]({{ '/en/advanced/Alluxio-Storage-Management.html#multiple-tier-storage' | relativize_url }}),
+and you have Persistent Volumes configured for Alluxio, the storage should be cleaned up too.
+You should delete and recreate the Persistent Volumes. 
+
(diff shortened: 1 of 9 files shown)


jiacheliu3:
alluxio-bot, test this please

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
 $ kubectl delete -f alluxio-journal-volume.yaml
 ```
 
+#### Upgrade Alluxio
+
+Upgrading Alluxio in Kubernetes is easy. You can just change the Alluxio docker image version to your desired upper version.
```
madanadit: - don't say that it is easy, it can  be frustrating for users if they are not able to do it right
- don't use "upper version" as that terminology isn't defined anywhere
Change this to 
Step 1: Upgrade the docker image version tag to the version you wish to upgrade to
Step 2: Restart the ... pod (specify the order to restart master / worker and commands to use)
...
Step n: Verify ...

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
...
+Then restart the Pod to make it take effect. You should make sure the Alluxio masters and workers are running on the same version for the best compatibility.
+
+For example, if you are currently running Alluxio v2.0.1 on Alluxio masters and you want to upgrade to the latest version.
```
madanadit: fo not hardcode 2.0.1, use doc macros instead

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
...
+Just change the `image` field of all the Alluxio containers to `image: alluxio/alluxio:latest` and restart them.
+
+```yaml
+  containers:
+  - name: alluxio-master
+    image: alluxio/alluxio:2.0.1
```
madanadit: the tag is not `latest` as per the line above

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
...
+    imagePullPolicy: IfNotPresent
+    command: ["/entrypoint.sh"]
```
madanadit: remove this line and the following. we don't want to maintain this as it might go stale

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
...
+    args: ["master-only", "--no-format"]
+    ...
+  - name: alluxio-job-master
+    image: alluxio/alluxio:2.0.1
+    imagePullPolicy: IfNotPresent
+    command: ["/entrypoint.sh"]
```
madanadit: remove this line as well

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
 $ kubectl delete -f alluxio-journal-volume.yaml
 ```
 
+#### Upgrade Alluxio
```
madanadit: We should have another section for "Updating Configuration"; can do that in a separate PR

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
...
+  - name: alluxio-job-master
+    image: alluxio/alluxio:2.0.1
+    imagePullPolicy: IfNotPresent
+    command: ["/entrypoint.sh"]
+    ...
+```
+
+Whether you need to re-format the journal and lose all file metadata depends on whether the journals are compatible.
+
+* If you are upgrading from v1.8 to v2.x, you have to format the journal. And by doing that you lose all the journal and file metadata.
```
madanadit: Add a bullet **Case A: Embedded Journal**, **Case B: UFS Journal**. keep embedded first as we will recommend that going fwd

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,53 @@ will be lost.
...
+the higher version Alluxio master will be able to read the lower version Alluxio master journal to recover all the journal and file metadata.
+We do not recommend rolling upgrade the masters with UFS journal, because there can be lower version masters reading journal written by higher version masters.
+And the behavior will be undefined.
+
+* If you are using embedded journal in Alluxio v2.x, the journal in each master will be lost on restart.
+
+In Kubernetes, [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) volumes have the same lifespan as the Pods they are on.
+So you should expect contents in *emptyDir* volumes to be lost after the upgrade.
+
+For example, the RAM disk below used by Alluxio workers will be wiped out after the upgrade.
```
madanadit: rather than just saying ramdisk will be wiped out, talk about what this means for the user. More like "any data written only to Alluxio storage and not persisted to the under storage will be lost. Metadata for those files will remain."

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: I'll do it in a separate PR. There's sth about live update configmap I wanna test and add too.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: I was trying to say "Find the `image` lines in the YAML below and change them to `latest`". But since it is more intuitive that the code snippet shows the desired resulting YAML, I'll update that.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: On second thought, the discussion on emptyDir and persistent vs nonpersistent storage is quite general. So it shouldn't be put in just update or rolling update section. I will move this discussion to the **Update Configuration** section in the PR yet to come., as that will cover general cases like restarting Pods. What do you think?

madanadit: sgtm, the upgrade section can point to that

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
...
+
+**Step 2: Format journal if necessary**
+
+Note that whether you need to re-format the journal and lose all file metadata depends on whether the journals are compatible.
+If you are upgrading from v1.8 to v2.x, you have to format the journal. And by doing that you lose all the journal and file metadata.
+If you are upgrading from v2.x to a higher minor version v2.y, or from v2.x.y to a higher minor version v2.x.z,
+the higher version Alluxio master will be able to read the lower version Alluxio master journal to recover all the journal and file metadata.
+
+If you need to format the journal, you can update the YAML file for Alluxio master containers to specify this on master startup time.
+Remove the `--no-format` option from Alluxio master startup arguments to force the re-format of journal as illustrated below.
```
madanadit: we should not suggest this way. every master restart will lead to a format, instead an init container should be used to force format

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
...
+```console
+$ kubectl delete -f alluxio-worker.yaml
+$ kubectl delete service alluxio-master-0
+$ kubeclt delete statefulset alluxio-master-0
+```
+
+If you have more alluxio-master Pods, delete them together with the corresponding Services.
+
+```console
+$ kubectl delete service alluxio-master-1
```
madanadit: write a bash for loop instead?

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
...
+$ kubectl create -f alluxio-master.yaml
+$ kubectl create -f alluxio-worker.yaml
+```
+
+**Step 4: Verify the Alluxio master and worker Pods are back up**
+
+What a proper validation should be varies from case to case.
+The example is just verifying that Alluxio is running.
+
+```console
```
madanadit: point to the verify section instead, do need to replicate commands here

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
...
+The consensus logic of Alluxio masters can be changed in minor versions so if you have a group of Alluxio masters running different versions,
+the behavior will be undefined. 
+
+###### Case B: Rolling Update with UFS Journal
+
+With UFS journal, there is only one shared journal destination and only the one primary master writes. All the secondary masters only read from it.
+
+You are able to do rolling update among Alluxio masters as below.
+
+**Step 1: Update the secondary master Pods and restart them**
```
madanadit: add instruction to identify the primary vs/ secondary (alluxio fs leader command) and kubectl commands to restart

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
...
+
+**Step 2: Update the primary master Pods and restart them**
+
+In this way no lower version Alluxio master will read the journal written by a higher version master.
+
+##### Alluxio Workers 
+
+The rolling update of Alluxio workers can go parallel to that of Alluxio masters.
```
madanadit: workers should be started after masters so that worker can register w/ new master after restart

madanadit:
Thanks @jiacheliu3. Added some more comments. In the absence of zookeeper in k8s,  the multi-master ufs journal rolling upgrade section seems unnecessary. Should we remove that altogether?

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
...
+  ...
+- name: alluxio-job-master
+  image: alluxio/alluxio:latest
+  imagePullPolicy: IfNotPresent
+  ...
+```
+
+**Step 2: Format journal if necessary**
+
+Note that whether you need to re-format the journal and lose all file metadata depends on whether the journals are compatible.
```
jiacheliu3: @madanadit According to [Upgrade from 1.x to 2.x](https://docs.alluxio.io/ee/user/stable/en/operation/2.x-Upgrade.html?q=upgrade), seems v2.x can pick up journal from 1.x?

The fact that we don't have one dedicated doc discussion what has to be formatted makes writing this part very hard actually...

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,132 @@ will be lost.
 $ kubectl delete -f alluxio-journal-volume.yaml
 ```
 
+#### Upgrade Alluxio
```
jiacheliu3: @madanadit Since there's no existing upgrade guide I can refer to, I had to create an issue instead. [Request for an upgrade guide](https://github.com/Alluxio/alluxio/issues/10002)

There are a few options how we can handle this:

1. Cover the things we need to cover (some may better belong elsewhere), remove them once the upgrade guide doc is ready.

2. Put this doc on hold, wait for the blocking transaction to finish and then continue.

3. Leave a placeholder link, either to an non-existent page or just blank link waiting to be updated.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,97 @@ will be lost.
...
+```
+
+The reason why you don't want to do `kubectl delete -f alluxio-master.yaml` is that will delete the Persistent Volume Claim in the `alluxio-master.yaml` if any.
+And that will result in your Persistent Volume getting released and the journal in it will be lost.
+
+Make sure all the Pods have been terminated before you move on to the next step.
+
+**Step 3: Format journal and Alluxio storage if necessary**
+
+Check the Alluxio upgrade guide page for whether the Alluxio master journal has to be formatted.
```
jiacheliu3: I'm taking the 3rd approach here. I plan to replace "Alluxio upgrade guide page" with the real page once it's ready. The only reason why I'm not taking the 1st option and provide the full picture is because I certainly is not the best person to write that. It may take a considerable time till I finish, and we eventually end up having a good deal of general Alluxio upgrade guide here in this PR.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,97 @@ will be lost.
...
+Kill all running Alluxio worker Pods.
+
+```console
+$ kubectl delete -f alluxio-worker.yaml
+```
+
+Then kill all running Alluxio master Pods by killing each StatefulSet and each Service.
+
+```console
+$ let masterCount = {number of your masters minus 1}
```
jiacheliu3: Jesus how can I do the range exclusive of the upper bound...So I can get rid of this `minus 1` here.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: In the updated doc I am pointing to a `alluxio-format-master.yaml` Job that will be added in a separate PR. I'm not including that in this PR because there's more than one thing to do to include that `alluxio-format-master.yaml`, I need to helm-template it and possible add something for formatting workers too.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: @madanadit The for loop has been replaced by using label selector `kubectl delete statefulset -l app=alluxio`. This is a cleaner solution (although somehow relies on masters being StatefulSet). We can discuss on better labeling.

The extra YAML changes are to apply the labels for doing this.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
madanadit: Option 1 sgtm. Include minimal info (k8s specific only when possible) in this doc

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,98 @@ will be lost.
...
+Make sure all the Pods have been terminated before you move on to the next step.
+
+**Step 3: Format journal and Alluxio storage if necessary**
+
+Check the Alluxio upgrade guide page for whether the Alluxio master journal has to be formatted.
+If no format is needed, you are ready to skip the rest of this section and move on to restart all Alluxio master and worker Pods.
+
+If you are running UFS journal, there is one Kubernetes [Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
+template that can be used for only formatting the master.
+You should make sure the Job runs with the same configuration with all your other Alluxio masters so it's able to find the journal persistent storage and format it.
```
madanadit: nit: same configuration = same configMap? add more details to clarify

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,98 @@ will be lost.
...
+And that will result in your Persistent Volume getting released and the journal in it will be lost. The same logic applies to `alluxio-worker.yaml`.
+
+Make sure all the Pods have been terminated before you move on to the next step.
+
+**Step 3: Format journal and Alluxio storage if necessary**
+
+Check the Alluxio upgrade guide page for whether the Alluxio master journal has to be formatted.
+If no format is needed, you are ready to skip the rest of this section and move on to restart all Alluxio master and worker Pods.
+
+If you are running UFS journal, there is one Kubernetes [Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
```
madanadit: one -> a single

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,98 @@ will be lost.
...
+You should make sure the Job runs with the same configuration with all your other Alluxio masters so it's able to find the journal persistent storage and format it.  
+
+```console
+$ kubectl apply -f alluxio-format-master.yaml
+```
+
+After the Job completes, it will be deleted by Kubernetes after the defined `ttlSecondsAfterFinished`.
+Then the clean journal will be ready for a new Alluxio master to start with.
+
+If you are running embedded journal and masters write journal each to its persistent storage,
```
madanadit: hmm, not clear what this means. can we add more detail here

integration/kubernetes/helm/alluxio/templates/alluxio-master.yaml
```diff
@@ -58,16 +58,19 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: {{ $masterName }}
+  labels:
+    app: alluxio
 spec:
   selector:
     matchLabels:
-      app: {{ $masterName }}
+      name: {{ $masterName }}
```
madanadit: why did we need this change?

jiacheliu3: Before we use `labels.app: {{ $masterName }}` for each StatefulSet. Now we use `labels.app: alluxio` for all resources so switched to use `labels.name` field as the selector.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,109 @@ will be lost.
...
+You should make sure the Job runs with the same configMap with all your other Alluxio masters so it's able to find the journal persistent storage and format it.  
+
+```console
+$ kubectl apply -f alluxio-format-master.yaml
+```
+
+After the Job completes, it will be deleted by Kubernetes after the defined `ttlSecondsAfterFinished`.
+Then the clean journal will be ready for a new Alluxio master to start with.
+
+**Case B: Emedded Journal**
```
jiacheliu3: @madanadit I tried to expand the discussion to 2 Cases and UFS journal in the front. Not quite sure about the double asterisk `**Case B**` since it's the same level as `**Step 3**`. Any suggestions?

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,109 @@ will be lost.
...
+
+If you are running embedded journal, each Alluxio master will write to its own journal destination defined by `alluxio.master.journal.folder`.
+In order to format the journals you have two options.
+
+1. Format all masters by running `alluxio formatMasters` in a Alluxio master Pod.
```
jiacheliu3: @madanadit PTAL if this is a valid option. If so then we can still use a single job that formats for all masters.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
madanadit: you can italicize instead with *Case A:...*

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
madanadit: that looks like a valid option

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -320,6 +320,110 @@ will be lost.
...
+If no format is needed, you are ready to skip the rest of this section and move on to restart all Alluxio master and worker Pods.
+
+How the journal should be formatted depends on the Alluxio master [journal type]({{ '/en/operation/Journal.html#ufs-journal-vs-embedded-journal' | relativize_url }}).
+
+*UFS Journal*
+
+If you are running UFS journal, there is only one place for the journal.
+There is a single Kubernetes [Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
+template that can be used for only formatting the master.
+The Job runs `alluxio formatMasters` and formats the journal for all masters.
```
jiacheliu3: @madanadit I will update the Job to use `alluxio formatMasters` so it can serve multiple cases

madanadit:

PR #9978, opened 2019-09-26
Add Mount Options

yuzhu:
This PR adds the ability for us to handle passing of mount options through `attachDB` calls. 
The format of the mount options is as follows: 
generic pass through options
`udb.mount-options.(scheme://authority).a.b.c=xyz`
special options
`udb.mount-options.(scheme://authority).readonly`
`udb.mount-options.(scheme://authority).shared`


yuzhu:
@ZacBlanco @gpang PTAL, thanks.

ZacBlanco:
@yuzhu I think putting `-` along with `.` in the property is pretty inconsistent with how the rest of our properties are defined. Can we convert everything to `.` for now so we're at least consistent with property keys?

I would also like to change the `udb-hive` properties at some point to something like `udb.hive` too...

gpang:
@ZacBlanco I intentionally introduced the `-` for the config. We can undo that decision, but my original reasoning was:
- I didn't need it to be the same as alluxio properties, since I doubt they would ever be specified in the same place at the same time
- I wanted the ability to have multi-word components in the hierarchy. I view the `.` as different levels in a hierarchy, but there are times where we use it to just separate a word boundary, even though that doesn't make sense in a hierarchy. For example `alluxio.master.embedded.journal.addresses` or `alluxio.master.daily.backup.enabled` probably shouldn't have to split up `embedded-journal`  or `daily-backup`.

ZacBlanco:
> I didn't need it to be the same as alluxio properties, since I doubt they would ever be specified in the same place at the same time

Is there any good reason to make it different though? What do we gain by adding `-` instead of having `.`?

> I wanted the ability to have multi-word components in the hierarchy. I view the . as different levels in a hierarchy, but there are times where we use it to just separate a word boundary, even though that doesn't make sense in a hierarchy. For example alluxio.master.embedded.journal.addresses or alluxio.master.daily.backup.enabled probably shouldn't have to split up embedded-journal or daily-backup.

From a user-facing perspective, I feel that most times I won't know which parts of a property would even constitute being part of the hierarchy vs being a multi-word component. Usually it's a decision the implementer will understand best, but from a user perspective it seems like it would just make it hard to memorize properties.

With `-` involved in properties (and the sheer magnitude of properties that we have) I feel that most users won't be able to memorize properties as easily because it introduces another construct to memorize: "Oh was this piece after `alluxio.master` `daily-backup` or `daily.backup`?"

Unless we can come up with some strict way of enforcing use of `-` in properties that is simple and intuitive for every property, I feel leaving it out is a better user-facing decision.

yuzhu:
@ZacBlanco @gpang changed `mount-options` to `mountoption` for now.

gpang:
@ZacBlanco sure, I guess the `-` doesn't provide much. We can get rid of that ability (in separate PR).

yuzhu:
jenkins, test this please

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
```diff
@@ -12,24 +12,69 @@
...
+    String mountPrefix = ConfigurationUtils.MOUNT_PREFIX;
+    Pattern datePatt = Pattern.compile("(\\(.*\\))\\.(.+?)");
+    for (Map.Entry<String, String> entry : values.entrySet()) {
+      if (entry.getKey().startsWith(mountPrefix)) {
+        String key = entry.getKey().substring(mountPrefix.length());
+        Matcher m = datePatt.matcher(key);
+        if (m.matches()) {
+          String resource = m.group(1);
+          // remove the bracket around the scheme://authority
+          resource = resource.substring(1, resource.length() - 1);
```
ZacBlanco: This feels a little sketchy. Regex might be a better option - but also could be overkill

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
```diff
@@ -12,24 +12,69 @@
...
   /**
    * Creates an instance.
    *
    * @param values the map of values
    */
   public UdbConfiguration(Map<String, String> values) {
     super(values);
+    mMountOptions = new HashMap<>(values.size());
+    String mountPrefix = ConfigurationUtils.MOUNT_PREFIX;
+    Pattern datePatt = Pattern.compile("(\\(.*\\))\\.(.+?)");
```
ZacBlanco: Make this `private static` so we don't compile every time?

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -132,22 +133,57 @@ public String getType() {
     }
   }
 
+  private void mountAlluxioPath(String tableName, AlluxioURI ufsPath, AlluxioURI tableUri)
+      throws IOException, AlluxioException {
+    // make sure the parent exists
+    mUdbContext.getFileSystem().createDirectory(tableUri.getParent(),
+        CreateDirectoryPOptions.newBuilder().setRecursive(true).setAllowExists(true).build());
+    Map<String, String> mountOptionMap = mConfiguration.getMountOption(
+        ufsPath.getScheme() + ";//" + ufsPath.getAuthority().toString());
```
ZacBlanco: colon `:` instead of `;`?

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -132,22 +133,57 @@ public String getType() {
...
+    MountPOptions.Builder option = MountPOptions.newBuilder();
+    for (Map.Entry<String, String> entry : mountOptionMap.entrySet()) {
+      if (entry.getKey().equals("readonly")) {
+        option.setReadOnly(Boolean.parseBoolean(entry.getValue()));
+      } else if (entry.getKey().equals("shared")) {
```
ZacBlanco: make `shared`/`readonly` constants somewhere?

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
yuzhu: i think Regex might be overkill in this case, we already extracted the substring using regex so it must begin with a bracket and end with a bracket.

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
yuzhu: done

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
yuzhu: done

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
yuzhu: done

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
```diff
@@ -29,6 +29,10 @@
 public class UdbConfiguration extends BaseConfiguration<UdbProperty> {
   private static final Logger LOG = LoggerFactory.getLogger(UdbConfiguration.class);
 
+  private static final Pattern CONFIG_PATTERN = Pattern.compile("(\\(.*\\))\\.(.+?)");
+  public static final String READ_ONLY_OPTION = "readyonly";
```
ZacBlanco: `read`?

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
ZacBlanco: What if someone made a typo

`udb.mountoption.([hdfs://host/path).mountoption.readonly`?

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
yuzhu: yikes, fixed..

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbConfiguration.java
yuzhu: i think in that case, the mountoption will not apply to the mount operation. I don't see how we can guard against user error in that case though.. we can only assume whatever in the bracket is the uri.

yuzhu:
jenkins, test this please

yuzhu:

PR #9982, opened 2019-09-26
Move configuration parsing out of looping scope

ggezer:
When FSM starts with many persist requests, parsing configuration values for every request has a high cost.


calvinjia:

PR #9984, opened 2019-09-26
Add check when mounting an S3 bucket at root to see if bucket exists

### LINKED ISSUE #9948: Alluxio allows mounting S3 buckets which don't exist
**Alluxio Version:**
Master branch

**Describe the bug**
Issuing a mount command on an S3 bucket that doesn't exist succeeds.

**To Reproduce**
Mount an S3 bucket that doesn't exist at the root of the S3 bucket.

**Expected behavior**
Some kind of warning/exception when mounting that the bucket is not accessible or doesn't exist.

**Urgency**
High. Users may make simple typos in mounting a bucket and will not be able to tell without further involved investigation that the mount itself is incorrect.

ns1123:
Fixes #9948

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
+++ b/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
@@ -81,7 +81,7 @@ public abstract class ObjectUnderFileSystem extends BaseUnderFileSystem {
   protected final Supplier<String> mRootKeySupplier =
       UnderFileSystemUtils.memoize(this::getRootKey);
 
-  /** Whether the bucket/root is listable */
+  /** Whether the bucket/root is listable. */
   protected final Supplier<Boolean> mIsRootListable =
       UnderFileSystemUtils.memoize(this::isRootListable);
 
--- a/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
+++ b/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
@@ -608,6 +608,7 @@ public abstract class ObjectUnderFileSystem extends BaseUnderFileSystem {
     try {
       return getObjectListingChunkForPath("", true) != null;
     } catch (Exception e) {
+      LOG.debug("Unable to list bucket root:", e);
       return false;
     }
   }
--- a/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
+++ b/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
@@ -608,7 +608,7 @@ public abstract class ObjectUnderFileSystem extends BaseUnderFileSystem {
     try {
       return getObjectListingChunkForPath("", true) != null;
     } catch (Exception e) {
-      LOG.debug("Unable to list root of bucket {}:",super.mUri.toString(), e.getMessage());
+      LOG.debug("Unable to list root of bucket {}:", super.mUri.toString(), e.getMessage());
       return false;
     }
   }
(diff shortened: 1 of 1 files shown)


gpang:
@ns1123 why is this check not enough? https://github.com/Alluxio/alluxio/blob/master/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java#L2642

ns1123:
@gpang the isDirectory() call always returns true if the URI isRoot(). However, the isRoot() check doesn't check if the user can actually access the root of the bucket.

gpang:
@ns1123 I see. Then shouldn't we change the object store implementation of `isDirectory` to fix this? It works as expected for other UFS implementations.

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -572,6 +573,32 @@ protected String getRootKey() {
     }
   }
 
+  @Override
+  public boolean isDirectory(String path) throws IOException {
```
gpang: why can't we do the logic in `ObjectUnderFileSystem` so all object stores are fixed?

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
```diff
@@ -594,6 +598,20 @@ public boolean isObjectStorage() {
...
+   * Checks if the root of the ObjectUnderFileSystem is listable. This is done once and cached
+   * per UFS to reduce time spent checking if the root is accessible.
+   *
+   * @return whether the root of the UFS is listable and return false if any exception is thrown
+   */
+  public boolean isRootListable() {
+    try {
+      return getObjectListingChunkForPath("", true) != null;
+    } catch (Exception e) {
+      return false;
```
calvinjia: Log an error here?

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
```diff
@@ -594,6 +598,21 @@ public boolean isObjectStorage() {
...
+   * Checks if the root of the ObjectUnderFileSystem is listable. This is done once and cached
+   * per UFS to reduce time spent checking if the root is accessible.
+   *
+   * @return whether the root of the UFS is listable and return false if any exception is thrown
+   */
+  public boolean isRootListable() {
+    try {
+      return getObjectListingChunkForPath("", true) != null;
+    } catch (Exception e) {
+      LOG.debug("Unable to list bucket root:", e);
```
gpang: Is there any more info you can add to this log message? Like the bucket name or anything else?

ns1123: done

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
```diff
@@ -594,6 +598,21 @@ public boolean isObjectStorage() {
...
+   * Checks if the root of the ObjectUnderFileSystem is listable. This is done once and cached
+   * per UFS to reduce time spent checking if the root is accessible.
+   *
+   * @return whether the root of the UFS is listable and return false if any exception is thrown
+   */
+  public boolean isRootListable() {
+    try {
+      return getObjectListingChunkForPath("", true) != null;
+    } catch (Exception e) {
+      LOG.debug("Unable to list root of bucket {}:", super.mUri.toString(), e.getMessage());
```
gpang: Is `e.getMessage()` used in the message? And, do we want the full stack trace of the exception, or do we just want the message?

ns1123: Good catch. There is no message for this and we do want the full exception.

ns1123:
Jenkins, test this please

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
ns1123: done

ns1123:
Jenkins, test this please

calvinjia:

Issue #9666, opened 2019-08-08
worker crash when copying the data to local


cheyang:
**Alluxio Version:**
2.1.0-SNAPSHOT

**Describe the bug**
Creating a large file which is more than than size of cache layer.Setting the file in alluxio as pin, then run `copyToLocal`. The worker is crash

**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)
1. Create the master and worker with docker

```
$ docker network create alluxio_nw
$ docker volume create ufs
$ docker volume create journal
$ docker run -d \
  -p 19999:19999 \
  --net=alluxio_nw \
  --name=alluxio-master \
  -e ALLUXIO_JAVA_OPTS="-Dalluxio.master.hostname=alluxio-master  " \
  -v journal:/opt/alluxio/journal \
  -v ufs:/opt/alluxio/underFSStorage \
  alluxio/alluxio:2.1.0-SNAPSHOT master
$ docker run -d \
  --net=alluxio_nw \
  --name=alluxio-worker \
  --shm-size=1G \
  -v ufs:/opt/alluxio/underFSStorage \
  -v journal:/opt/alluxio/journal \
  -e ALLUXIO_JAVA_OPTS="-Dalluxio.worker.memory.size=1G -Dalluxio.master.hostname=alluxio-master" \
  alluxio/alluxio:2.1.0-SNAPSHOT worker
```

2. jump into the worker, create a file which is 2048M

```
docker exec -it alluxio-worker bash
cd /opt/alluxio/underFSStorage/
dd if=/dev/zero of=hello.txt bs=2048M count=1
```

3.change to the master, sync the data

```
/opt/alluxio/bin/alluxio fs ls   -Dalluxio.user.file.metadata.sync.interval=0 /
-rw-r--r--  root           root                2147479552       PERSISTED 08-07-2019 23:41:21:315   0% /hello.txt
```

4.pin the file `/hello.txt`

```
/opt/alluxio/bin/alluxio fs pin /hello.txt
File '/hello.txt' was successfully pinned.
```

5. run `copyToLocal` inside the master

```
 /opt/alluxio/bin/alluxio fs copyToLocal /hello.txt /
Channel authentication failed. ChannelId: 9f4990e4-9e7c-4b03-a59f-5484db296d48, AuthType: SIMPLE, Target: 694fc63e18f3:29999, Error: alluxio.exception.status.UnavailableException: io exception
```

6. Check the workers log:

```
[...]

PR #10001, opened 2019-09-27
Fix local file writer client memory leak by cleaning up mmapped files

madanadit:
This PR addresses a memory leak in the Alluxio client for local file writes on the short-circuit data access path. The impact is most significant for long running processes, such as the job worker (with an embedded Alluxio client), as open files are not released until the JVM terminates and continue to occupy memory in the mean time.
    
In the issue reported, the worker and job worker run in the same container. Pinning a 2G file causes set replication = 1 and the replication checker creates a job for the job worker to load the file from the ufs. The job worker opens mmaped files for temporary blocks and does not cleanup temporary blocks correctly as reported by: 
```
bash-4.4# lsof +L1 | grep /dev/shm
...
166	/usr/lib/jvm/java-1.8-openjdk/jre/bin/java	/dev/shm/alluxioworker/.tmp_blocks/4/109d8b9683aba404-1000014 (deleted)
166	/usr/lib/jvm/java-1.8-openjdk/jre/bin/java	/dev/shm/alluxioworker/.tmp_blocks/917/1c977804f85aa395-1000017 (deleted)
```

When running `du` and based on the worker metadata, there is space available in `/dev/shm`. However when `df` is run, the job worker process occupies memory for unclosed files which have been deleted.

Fixes: https://github.com/Alluxio/alluxio/issues/9666

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
@@ -134,11 +134,8 @@ public final class LocalFileDataWriter implements DataWriter {
 
   @Override
   public void cancel() throws IOException {
-    boolean cancelStream = !(mStream.isClosed() || mStream.isCanceled());
     try {
-      if (cancelStream) {
-        mStream.cancel();
-      }
+      mStream.cancel();
     } catch (Exception e) {
       throw mCloser.rethrow(e);
     } finally {
@@ -151,13 +148,10 @@ public final class LocalFileDataWriter implements DataWriter {
 
   @Override
   public void close() throws IOException {
-    boolean closeStream = !(mStream.isClosed() || mStream.isCanceled());
-    if (closeStream) {
-      mCloser.register(() -> {
-        mStream.close();
-        mStream.waitForComplete(mDataTimeoutMs);
-      });
-    }
+    mCloser.register(() -> {
+      mStream.close();
+      mStream.waitForComplete(mDataTimeoutMs);
+    });
     mCloser.close();
   }
 
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
@@ -29,7 +29,6 @@ import io.netty.buffer.ByteBuf;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.Closeable;
 import java.io.IOException;
 
 import javax.annotation.concurrent.NotThreadSafe;
@@ -155,8 +154,8 @@ public final class LocalFileDataWriter implements DataWriter {
     boolean closeStream = !(mStream.isClosed() || mStream.isCanceled());
     if (closeStream) {
       mCloser.register(() -> {
-          mStream.close();
-          mStream.waitForComplete(mDataTimeoutMs);
+        mStream.close();
+        mStream.waitForComplete(mDataTimeoutMs);
       });
     }
     mCloser.close();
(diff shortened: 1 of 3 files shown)


core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
```diff
@@ -135,11 +135,11 @@ public void writeChunk(final ByteBuf buf) throws IOException {
...
   @Override
   public void cancel() throws IOException {
-    if (mStream.isClosed() || mStream.isCanceled()) {
-      return;
-    }
+    boolean cancelStream = !(mStream.isClosed() || mStream.isCanceled());
     try {
-      mStream.cancel();
+      if (cancelStream) {
+        mStream.cancel();
```
gpang: I'm curious. In the cancel path, we call cancel on the stream, but we don't add to the closer, so the stream doesn't get closed. Is that intentional?

madanadit:
Yes, that is accurate. I’ll update the PR title

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
madanadit: it seems so. the documentation for the grpc blocking stream says `No other methods on this class can be called after this method has been called.`(for cancel). 

@bf8086  what do you think?

madanadit:

madanadit:

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
bf8086: Yeah once we cancel the gRPC request it is effectively closed.

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
```diff
@@ -135,11 +134,11 @@ public void writeChunk(final ByteBuf buf) throws IOException {
 
   @Override
   public void cancel() throws IOException {
-    if (mStream.isClosed() || mStream.isCanceled()) {
-      return;
-    }
+    boolean cancelStream = !(mStream.isClosed() || mStream.isCanceled());
```
calvinjia: Perhaps extract this logic to another method?

calvinjia:
@madanadit Thanks for the fix! Is there a way we could validate this with a unit or integration test?

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
madanadit: i removed the check all together as the stream already has an `isOpen` check for both `close` and `cancel`. updated the javadoc to reflect that

madanadit:
@calvinjia added the unit test. PTAL

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataWriter.java
```diff
@@ -135,9 +134,6 @@ public void writeChunk(final ByteBuf buf) throws IOException {
 
   @Override
   public void cancel() throws IOException {
-    if (mStream.isClosed() || mStream.isCanceled()) {
-      return;
-    }
```
calvinjia: Should we do a similar pattern here and register `mStream.cancel()` with the closer and just close the closer?

madanadit: sgtm

madanadit: @calvinjia done

madanadit:
jenkins, test this please.

calvinjia:

madanadit:

madanadit:

PR #10006, opened 2019-09-30
Exclude ufs level information as default in listStatus results

ggezer:
This is in order to prevent mount table look-ups under `listStatus` call. Related fields are not utilized in `listStatus` call.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -781,15 +781,15 @@ public final class DefaultFileSystemMaster extends CoreMaster implements FileSys
 
   private FileInfo getFileInfoInternal(LockedInodePath inodePath)
       throws FileDoesNotExistException, UnavailableException {
-    return getFileInfoInternal(inodePath, true);
+    return getFileInfoInternal(inodePath, false);
   }
 
   /**
    * @param inodePath the {@link LockedInodePath} to get the {@link FileInfo} for
-   * @param includeUfsInformation whether to include ufs level information
+   * @param excludeUfsInfo whether to exclude ufs level information
    * @return the {@link FileInfo} for the given inode
    */
-  private FileInfo getFileInfoInternal(LockedInodePath inodePath, boolean includeUfsInformation)
+  private FileInfo getFileInfoInternal(LockedInodePath inodePath, boolean excludeUfsInfo)
       throws FileDoesNotExistException, UnavailableException {
     Inode inode = inodePath.getInode();
     AlluxioURI uri = inodePath.getUri();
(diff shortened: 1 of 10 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -801,15 +807,17 @@ private FileInfo getFileInfoInternal(LockedInodePath inodePath)
...
+    if (includeUfsInformation) {
+      MountTable.Resolution resolution;
+      try {
+        resolution = mMountTable.resolve(uri);
+      } catch (InvalidPathException e) {
+        throw new FileDoesNotExistException(e.getMessage(), e);
+      }
+      AlluxioURI resolvedUri = resolution.getUri();
+      fileInfo.setUfsPath(resolvedUri.toString());
+      fileInfo.setMountId(resolution.getMountId());
```
calvinjia: If these fields are not set, what will they default to? Should we have some handling on the client side `URIStatus` object to handle the unset fields if they are accessed (ie. throw some exception to tell them to use the other API)?

core/transport/src/grpc/file_system_master.proto
```diff
@@ -184,6 +184,7 @@ message ListStatusPOptions {
   optional LoadMetadataPType loadMetadataType = 2;
   optional FileSystemMasterCommonPOptions commonOptions = 3;
   optional bool recursive = 4;
+  optional bool includeUfsInfo = 5;
```
calvinjia: Do we want to default this to `true` for backward compatibility?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ggezer: Would we see any difference in these fields for files created as MUST_CACHE? Would they still resolve to a UFS path for instance?

core/transport/src/grpc/file_system_master.proto
ggezer: https://github.com/Alluxio/alluxio/pull/10006/commits/5551bc48c4ce46c8170c1c202af7f80e86d31abd

ggezer: We cannot set to true at proto layer.
However made a fix to interpret absence of this field properly for supporting older clients.

gpang: if `false` is default, then we should reverse this to be `excludeUfsInfo`

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: I don't think so, but it would be good to avoid having things like mount id = 0 or ufs path = blank.

core/transport/src/grpc/file_system_master.proto
ggezer: That would conform with the PR title as well.

ggezer:
Closing this in order to allow a generic fix on the main issue.
https://github.com/Alluxio/alluxio/issues/10038

PR #10030, opened 2019-10-02
Fix whitespace in K8s configMap YAML template

jiacheliu3:
This fix changes the end of line in `alluxio-configMap.yaml` to use `{{end}}` instead of `{{ end -}}` so the trailing newline whitespace doesn't get removed. If the trailing whitespace operation is not removed, comments on the new line will be merged into this line.


madanadit:

PR #10037, opened 2019-10-03
[DOCFIX] Require trailing slash when mounting whole s3 bucket

LuQQiu:
When backing up to root ufs with the following configuration
```
alluxio.master.mount.table.root.ufs=s3://bucket-name
```
The `fasdmin backup` command will fail because our Alluxio URI has problem processing uri with no path (without a trailing slash after authority).

This PR fixes the issue and allows s3://bucket-name as root ufs.

### DIFF
--- a/docs/en/ufs/S3.md
+++ b/docs/en/ufs/S3.md
@@ -45,6 +45,9 @@ Specify an **existing** S3 bucket and directory as the under storage system by m
 alluxio.master.mount.table.root.ufs=s3://S3_BUCKET/S3_DIRECTORY
 ```
 
+Note that if you want to mount the whole s3 bucket, add a trailing slash after the bucket name 
+(e.g. `s3://S3_BUCKET/`).
+
 Specify the AWS credentials for S3 access by setting aws.accessKeyId and aws.secretKey in
 `alluxio-site.properties`.
 


core/base/src/main/java/alluxio/AlluxioURI.java
```diff
@@ -110,7 +110,10 @@ public AlluxioURI(String scheme, Authority authority, String path, Map<String, S
    * @param child the child
    */
   public AlluxioURI(AlluxioURI parent, AlluxioURI child) {
-    mUri = URI.Factory.create(parent.mUri, child.mUri);
+    // Allow parent to be something similar to s3://bucket_name
+    URI parentUri = parent.toString().endsWith(AlluxioURI.SEPARATOR) ? parent.mUri :
+        new AlluxioURI(parent.toString() + AlluxioURI.SEPARATOR).mUri;
```
ZacBlanco: I'm not a huge fan of adding this to the `AlluxioURI` constructor. It adds extra string parsing because we call `new AlluxioURI` first which in turns calls another `URI.Factory.create` which can be slow.

Is there a better way we can support when the root UFS  doesn't have a trailing slash without `AlluxioURI` class?

LuQQiu: Changed to support in AlluxioURI, but I also feel strange becuase it also supports 
```
alluxio://host:port
hdfs://host:port
```
without trailing slash, do we also want to support those senarios?

core/base/src/main/java/alluxio/uri/URI.java
```diff
@@ -96,6 +96,9 @@ public static URI create(String uriStr) {
 
       // uri path is the rest of the string -- fragment not supported
       String path = uriStr.substring(start, uriStr.length());
+      if (!uriStr.equals("") && path.equals("")) {
```
ZacBlanco: What is the reason to check `uriString.equals("")` vs only checking `path.equals("")`?

core/base/src/main/java/alluxio/AlluxioURI.java
ZacBlanco: What does the `java.net.URI` class return for `getPath()` when we pass in `alluxio://host:port` or `hdfs://host:port`? Our URI class should be mirroring the behavior there

core/base/src/main/java/alluxio/uri/URI.java
LuQQiu: if the given path is "", we don't want to add a "/" to path, which cause the uri to be "/" instead of ""

LuQQiu:
jenkins, test this please

LuQQiu:
Tests in error: 
  Run 1: ZookeeperFailureIntegrationTest.zkFailure:78 » Timeout Timed out waiting for a...
  Run 2: ZookeeperFailureIntegrationTest.zkFailure:78 » Timeout Timed out waiting for a...

alluxio.server.ft.journal.MultiMasterJournalTest.testCheckpointReplay(alluxio.server.ft.journal.MultiMasterJournalTest)
  Run 2: MultiMasterJournalTest.testCheckpointReplay:64 » IO java.lang.RuntimeException...

  Run 2: MultiMasterJournalTest.before:53 » IO java.lang.RuntimeException: alluxio.exce...

LuQQiu:
jenkins, test this please

LuQQiu:
@ZacBlanco PTAL, thanks

LuQQiu:

PR #10039, opened 2019-10-03
Fix K8s Service label selector

jiacheliu3:
This change fixes the selector used by K8s Services to find the correct master Pods. Without this change the Services are not able to find the correct master Pods so the master Pods cannot be talked to. 

This bug was introduced by the previous change to use `app: alluxio` for all resource and `name: {{$masterName}}` for the master Pods. Only the selector for StatefulSet was updated, and the selector for Service was missed.


madanadit:

PR #9972, opened 2019-09-25
Set up initFromBackup correctly

apc999:
Fixes https://github.com/Alluxio/alluxio/issues/9969 by addressing the regression  in PR #9808

PR #10046, opened 2019-10-04
Fix restore from backup on local path

### LINKED ISSUE #9969: Alluxio does not initFromBackup correctly when specifying local path
**Alluxio Version:**
master branch

**Describe the bug**
Alluxio's restore from backup using the `alluxio-start.sh -i <Backup-URI>` does not work as intended. The master takes any user-specified URI and strips it of scheme/authority before checking the root UFS for the backup. This prevents users from specifying a local path using 'file:///'.

**To Reproduce**
Try to restore from a backup file in a non-root UFS URI

**Expected behavior**
Restoring from backup should allow either Root UFS URI or Local URI based on user specified scheme.

**Urgency**
HIGH

**Additional Context**
This used to work before the changes in PR #9808. Unit tests should be added here to prevent breaking changes like this in the future.

ZacBlanco:
There was a regression introduced that caused restoring from a
local backup file to stop working. Without this, all backups would
need to be stored in the root UFS which is non-ideal.

This PR fixes the regression and adds a test to ensure that we can
restore from backup when the root UFS is not local and the backup
path is on the local system.

- Regression originally introduced in #9808 
- Continuation of PR #9972 
- Fixes #9969


ZacBlanco:
@ns1123 primary, PTAL
@ggezer secondary, PTAL

@apc999 FYI

gpang:
Jenkins, test this please

ZacBlanco:
@ns1123 PTAL

ZacBlanco:

PR #10049, opened 2019-10-07
Support aliyun oss for multiple part download

cheyang:
1. Update aliyun_oss_sdk to 3.6.0
2. Support multi parts download to speed up
   a file's size is 136977679 can be complete  in
Single Part: 1982 ms
``` 
java -jar osscli-1.0-SNAPSHOT-jar-with-dependencies.jar imagenet-huabei5 images/train-00760-of-01024
Download the metadata: false
Getting Started with OSS SDK for Java

Downloading an object
Download time in ms = 1982
Content-Type: application/octet-stream
Size: 136977679
Dump to file:images/train-00760-of-01024
```
Multi part: 455 ms

```
java -cp osscli-1.0-SNAPSHOT-jar-with-dependencies.jar com.aliyun.osscli.DownloadOSSObject imagenet-huabei5 images/train-00760-of-01024 8 16
Use num: 8
Use part size: 16
Download time in ms = 455
D8F69CE8F96710BB6FB620B99D17E717-54
```

It can get 4 times speed up.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -766,7 +766,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
                   .setDefaultValue(false)
                   .setDescription("(Experimental) If true, using streaming download from OSS.")
                   .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
-                  .setScope(Scope.SERVER)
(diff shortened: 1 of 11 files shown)


calvinjia:
@cheyang Are you experimenting with the S3-like approach for input streams since OSS's streams are abortable? I think that would be a simpler solution.

cheyang:
I will create a new PR to address oss client upgrade. thanks.

PR #10054, opened 2019-10-07
Add metrics to BackupManager for backup and restore of the journal

ns1123:
Adding metrics for backup and restore from backup to track number of entries and time taken for backup.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/BackupManager.java
+++ b/core/server/common/src/main/java/alluxio/master/BackupManager.java
@@ -340,6 +340,11 @@ public class BackupManager {
     }
   }
 
+  /**
+   * Class that contains metrics for BackupManager.
+   * This class is public because the counter names are referenced in
+   * {@link alluxio.web.WebInterfaceAbstractMetricsServlet}.
+   */
   public static final class Metrics {
     private static final Timer BACKUP_ENTRIES_PROCESS_TIME
         = MetricsSystem.timer(MasterMetrics.BACKUP_ENTRIES_PROCESS_TIME);
--- a/core/server/common/src/main/java/alluxio/master/BackupManager.java
+++ b/core/server/common/src/main/java/alluxio/master/BackupManager.java
@@ -16,9 +16,13 @@ import alluxio.conf.ServerConfiguration;
 import alluxio.master.journal.JournalContext;
 import alluxio.master.journal.JournalEntryAssociation;
 import alluxio.master.journal.JournalEntryStreamReader;
+import alluxio.metrics.MasterMetrics;
+import alluxio.metrics.MetricsSystem;
 import alluxio.proto.journal.Journal.JournalEntry;
 import alluxio.util.ThreadFactoryUtils;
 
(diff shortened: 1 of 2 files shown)


ns1123:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -105,6 +109,10 @@ public void backup(OutputStream os, AtomicLong entryCount) throws IOException {
     // Whether buffering is still active.
     AtomicBoolean bufferingActive = new AtomicBoolean(true);
 
+    // Start the timer for backup metrics and reset the entry counter.
+    Metrics.LAST_BACKUP_ENTRIES_COUNT.dec(Metrics.LAST_BACKUP_ENTRIES_COUNT.getCount());
```
calvinjia: Do we do this because a process could restore from backup multiple times?

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -321,4 +339,23 @@ private void safeWaitTasks(Set<Future<?>> activeTasks, CompletionService<?> comp
...
+   */
+  public static final class Metrics {
+    private static final Timer BACKUP_ENTRIES_PROCESS_TIME
+        = MetricsSystem.timer(MasterMetrics.BACKUP_ENTRIES_PROCESS_TIME);
+    private static final Timer BACKUP_RESTORE_PROCESS_TIME
+        = MetricsSystem.timer(MasterMetrics.BACKUP_RESTORE_PROCESS_TIME);
+    private static final Counter LAST_BACKUP_ENTRIES_COUNT
+        = MetricsSystem.counter(MasterMetrics.LAST_BACKUP_ENTRIES_COUNT);
+    private static final Counter LAST_BACKUP_RESTORE_COUNT
+        = MetricsSystem.counter(MasterMetrics.LAST_BACKUP_RESTORE_COUNT);
```
calvinjia: Do the timers keep track of multiple cycles while the counters only keep track of the last one? If so should these be consistent?

core/server/common/src/main/java/alluxio/master/BackupManager.java
ns1123: I don't think we would run into a scenario where we restored from backup multiple times. The idea was to just keep this in line with the regular backup metric information. I can remove it if you feel that it is unnecessary.

core/server/common/src/main/java/alluxio/master/BackupManager.java
ns1123: Unfortunately yes. Having the entries metric report the entry count of the latest backup gives a good idea of the existing backup size. The rolling timer is useful in understanding how the backup cycle has been changing over time in duration. I'm open to suggestions on other metrics that we could collect for this, but I'm comfortable with these two metrics not being consistent because they serve different functions.

calvinjia:

ZacBlanco:

PR #10055, opened 2019-10-07
Remove root inode acl check

ns1123:
This check is not required and makes it hard to replay a backup from a different cluster.


gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1723,6 +1723,17 @@ public String toString() {
           .setIsHidden(true)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_SKIP_ROOT_ACL_CHECK =
+      new Builder(Name.MASTER_SKIP_ROOT_ACL_CHECK)
+          .setDefaultValue(false)
+          .setDescription("Skip root directory ACL check when restoring from backup. This is to "
```
gpang: Is this only intended for backup restore? The code change affects any journal replay (restart, secondary catchup, backup restore).

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1723,6 +1723,17 @@ public String toString() {
...
+              + "allow users to restore a backup from a different cluster onto their current one "
+              + "without having to recreate the different clusters owner user.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .setIsHidden(true)
+          .setIgnoredSiteProperty(true)
```
gpang: What does this do?

core/common/src/main/java/alluxio/conf/PropertyKey.java
ns1123: The only real use I see for this is in the event of restoring from a backup. However this can be helpful in other scenarios i.e. starting Alluxio as a different OS user if the old user had been deleted.

Also, I don't think this takes effect on secondary catchup since the check is done in isPrimary()

core/common/src/main/java/alluxio/conf/PropertyKey.java
ns1123: Prevents the property from being set in alluxio-site.properties. This property should be set explicitly in alluxio-env.sh or in the command line with -D since it is not expected to be used in standard operating procedure.

core/common/src/main/java/alluxio/conf/PropertyKey.java
gpang: I think this description is misleading, since this will actually affect all journal replay (restarts, secondaries, backup), not only backup. Can you update the description?

ns1123:
@gpang done

ZacBlanco:

PR #10056, opened 2019-10-07
Update table location in catalog service after transformation

chcn:
After transformation job is done, the new table location will be automatically updated in catalog service, so future queries will use the new table location transparently.


experimental/table/server/master/src/main/java/alluxio/master/catalog/DefaultCatalogMaster.java
```diff
@@ -111,6 +112,22 @@ public Table getTable(String dbName, String tableName) throws IOException {
     return mCatalog.readTable(dbName, tableName, constraint);
   }
 
+  @Override
+  public void transformTable(String dbName, String tableName, String type, String newTableLocation)
+      throws IOException {
+    Table table = mCatalog.getTable(dbName, tableName);
+    AlluxioURI tableLocation = new AlluxioURI(table.getUdbTable().getBaseLocation());
+    AlluxioURI transformedTableLocation = new AlluxioURI(tableLocation.getScheme(),
```
yuzhu: does `transformedTableLocation ` always have the same scheme and authority as the original table location? it shouldn't right?

experimental/table/server/master/src/main/java/alluxio/master/catalog/DefaultCatalogMaster.java
```diff
@@ -111,6 +112,22 @@ public Table getTable(String dbName, String tableName) throws IOException {
...
+        tableLocation.getAuthority(), newTableLocation);
+    table.getUdbTable().updateLocation(transformedTableLocation.toString());
+    for (alluxio.master.catalog.Partition partition : table.getPartitions()) {
+      AlluxioURI partitionLocation = new AlluxioURI(partition.getBaseLayout().getLocation());
+      AlluxioURI transformedPartitionLocation = transformedTableLocation.join(
```
yuzhu: is this going to work when original partition path is not a subdirectory of the table path?

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/util/PathTranslator.java
```diff
@@ -66,6 +68,10 @@ public PathTranslator addMapping(String alluxioPath, String ufsPath) {
    * @throws IOException if the ufs path is not mounted
    */
   public String toAlluxioPath(String ufsPath) throws IOException {
+    AlluxioURI uri = new AlluxioURI(ufsPath);
```
yuzhu: i think this method is called quite often, so it would be more efficient if we did not have to construct AlluxioURI for every ufsPath.  What about adding the mapping to the `PathTranslator` when we update the location of the `partition` and `udbTable`?

experimental/table/server/master/src/main/java/alluxio/master/catalog/DefaultCatalogMaster.java
chcn: currently, transformation will only transform a table from Alluxio and to Alluxio, so the scheme and authority should always be Alluxio.

experimental/table/server/master/src/main/java/alluxio/master/catalog/DefaultCatalogMaster.java
chcn: the current assumption of transformation is that partition directories are always subdirectories of the table path.

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/util/PathTranslator.java
chcn: changed to add the mapping in `HiveTable.updateLocation`

chcn:
@yuzhu thanks, PTAL

experimental/table/server/master/src/main/java/alluxio/master/catalog/DefaultCatalogMaster.java
yuzhu: so for now, can we use `newTableLocation` to construct the `transformedTableLocation` URI?

experimental/table/server/master/src/main/java/alluxio/master/catalog/DefaultCatalogMaster.java
yuzhu: you can use `layout.getSpec()` to access the partitionName and use that to construct the path. that might be more reliable.

yuzhu:
@chcn once #10079 is merged, you will have an easier time resolving this conflict.  Several changes that may affect this PR: 
1. Table no longer hold a reference to UdbTable. Table contains a copy of state from UdbTable, so we just need to update whatever is in Table object.  e.g. tableLocation in UdbTableInfo and Partition location in Partition. 
2. Layout construction now requires a factory. See #10079 for more details. 

We can discuss this further once #10079 is merged.

PR #10057, opened 2019-10-08
Add block size to information stored in UfsFileStatus

### LINKED ISSUE #10035: Wasteful UFS RPC calls for HDFS
**Alluxio Version:**
2.x and 1.8.x

**Describe the bug**
List on HDFS UFS issues a separate getBlockSizeBytes HDFS call for each file. This is a redundant RPC that can be removed.

**To Reproduce**
Mount an HDFS cluster, and open the Alluxio Metrics JSON endpoint. Run a list directory call on the Alluxio directory that has the mounted HDFS and then monitor the UFS metrics for getBlockSizeBytes RPC calls.

**Expected behavior**
No increase in getBlockSizeBytes since the available information is already provided via Hadoop API's getStatus.

**Urgency**
High

**Additional context**
Results in severe latency when bulk listing objects since the getBlockByteSize calls are done serially.

calvinjia:
Adds block size info to the `UfsFileStatus`. This mainly helps HDFS whos implementation of `getBlockSizeBytes` has 2(!) rpc calls.

This change should be fully backward compatible, but older servers using newer ufs jars will not work (because the constructor of UfsFileStatus is different). 

@apc999 A separate question is if we should enforce usage of UFS jars which are on the same version of the server so we don't need to maintain legacy code.

Fixes #10035

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UnderFileSystem.java
+++ b/core/common/src/main/java/alluxio/underfs/UnderFileSystem.java
@@ -329,9 +329,13 @@ public interface UnderFileSystem extends Closeable {
   /**
    * Gets the block size of a file in under file system, in bytes.
    *
+   * @deprecated block size should be returned as part of the {@link UfsFileStatus} from
+   * {@link #getFileStatus(String)} or {@link #getExistingFileStatus(String)}
+   *
    * @param path the file name
    * @return the block size in bytes
    */
+  @Deprecated
   long getBlockSizeByte(String path) throws IOException;
 
   /**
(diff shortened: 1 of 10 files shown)


calvinjia:
jenkins, test this please.

core/common/src/main/java/alluxio/underfs/UnderFileSystem.java
```diff
@@ -329,9 +329,13 @@ public int getValue() {
   /**
    * Gets the block size of a file in under file system, in bytes.
    *
+   * @deprecated block size should be returned as part of the {@link UfsFileStatus} from
```
apc999: not sure if you want to deprecate this API because other two APIs providing structures covering this piece of info?
e.g., in an empty UFS, what is the expected approach to get block size then?

calvinjia: My thoughts on why I want to deprecate this method -
This method is only useful if the cost of getting the block size is much less than getting the file status (this is true for certain UFSes, such as object stores) AND we use blocksize by itself often in our code (this is not true, we only use block size in one place and it is alongside getFileStatus).
Going forward, I don't think we will have a use case for only fetching the block size of a file from the UFS without fetching the rest of the file status, which is why I think we can deprecate this method.
When you say empty ufs, do you mean a ufs with no files? Block size is tied to a file (ie. it takes the file path as an argument).

calvinjia:

PR #10071, opened 2019-10-09
Write to the journal

yuzhu:
It journals attachDB operation and addTable operation, mount operation as part of the addTable is journaled by FileSystemMaster. 
- Separated UdbTable and Table so that table will only get its state from a udbTable when sync is called. This prevents the catalog service from automatically refreshing the state from udb upon restart. 
- Will add an update table journal entry when we have state changes (transformation) in the future, or when sync gets called.


experimental/table/server/master/src/main/java/alluxio/master/catalog/Database.java
```diff
@@ -29,10 +35,13 @@
  * The database implementation that manages a collection of tables.
  */
 public class Database {
```
gpang: Maybe `Database` should also implement the `Journaled` interface?

experimental/table/server/master/src/main/java/alluxio/master/catalog/Table.java
```diff
@@ -14,31 +14,62 @@
...
 public class Table {
-  private final String mName;
+  private static final Logger LOG = LoggerFactory.getLogger(Table.class);
+
+  private String mName;
   private final Database mDatabase;
-  private final Schema mSchema;
-  private final UdbTable mUdbTable;
+  private Schema mSchema;
+  private UdbTableInfo mTableInfo;
```
gpang: why is this not `UdbTable`?

experimental/table/server/master/src/main/java/alluxio/master/catalog/Table.java
```diff
@@ -14,31 +14,62 @@
...
   // TODO(gpang): this should be indexable by partition spec
-  private final ArrayList<Partition> mPartitions;
+  private List<alluxio.grpc.catalog.Partition> mPartitions;
```
gpang: why is this the grpc partition, and not the java class?

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveTable.java
```diff
@@ -72,17 +74,18 @@
...
       List<ColumnStatisticsInfo> statistics, List<FieldSchema> cols, List<Partition> partitions,
-      Table table) throws IOException {
-    // TODO(gpang): don't throw exception in constructor
+      Table table) {
     mHive = hive;
     mHiveDatabase = hiveDatabase;
+    mTable = table;
     mPathTranslator = pathTranslator;
+    mPartitions = partitions;
+    // The following field are always populated
```
gpang: I'm not sure what this means? the other fields also look populated?

core/transport/src/grpc/catalog/catalog_master.proto
```diff
@@ -202,6 +202,7 @@ message Layout {
   optional string layoutType = 1;
   optional LayoutSpec layoutSpec = 2;
   optional bytes layoutData = 3;
+  map<string, ColumnStatisticsInfo> stats = 4;
```
gpang: is `ColumnStatisticsInfo` a generalized structure, or specific to hive?

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbTable.java
```diff
@@ -49,4 +49,10 @@
    * @return returns a proto representing the table
    */
   UdbTableInfo toProto() throws IOException;
+
+  /**
+   *
+   * @return true if the udbtable is connected
+   */
+  boolean isConnected();
```
gpang: how is this used? When is something "not" connected?

experimental/table/server/master/src/main/java/alluxio/master/catalog/Table.java
yuzhu: my thinking was to decouple `UdbTable` from Table object so that we can't accidentally modify it. The only way to update a table object is to go through sync.

experimental/table/server/master/src/main/java/alluxio/master/catalog/Table.java
yuzhu: when I was constructing a table from the journal, it was not possible for me to construct a HiveLayout, because we don't have dependency on `alluxio.table.under.hive` module.. and therefore I can not construct a java partition class based on that HiveLayout class.

experimental/table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveTable.java
yuzhu: oops this comment is outdated. will remove

core/transport/src/grpc/catalog/catalog_master.proto
yuzhu: it came from hive, but i think it is quite generic.

experimental/table/server/common/src/main/java/alluxio/table/common/udb/UdbTable.java
yuzhu: also old comment, will remove

yuzhu:
@gpang replied to your comments, PTAL

yuzhu:
@gpang passed local tests, PTAL

yuzhu:
@gpang passed local tests, PTAL

gpang:

PR #10072, opened 2019-10-09
Optimize alluxio.resource.DynamicResourcePool

gengxin:
Optimize DynamicResourcePool by using ConcurrentLinkedDeque instead of (ArrayDeque + Lock)

JMH benchmark result shows about 4 times faster when simply calling acquire and release methods.

JMH test case:
https://gist.github.com/gengxin/4296a4c43b3cff9339df4b8100d5f1ee

Result before optimization:
> Benchmark                                              Mode  Cnt  Score   Error  Units
> DynamicResourcePoolBenchmark.acquireAndRelease         avgt   20  4.936 ± 0.429  us/op
> 
> ....[Thread state distributions]....................................................................
>  58.2%         RUNNABLE
>  22.5%         WAITING
>  19.3%         TIMED_WAITING

Result after optimization:
> Benchmark                                              Mode  Cnt  Score   Error  Units
> DynamicResourcePoolBenchmark.acquireAndRelease         avgt   20  1.378 ± 0.052  us/op
> 
> ....[Thread state distributions]....................................................................
>  77.2%         RUNNABLE
>  18.7%         TIMED_WAITING
>   4.2%         WAITING

Benchmark hardware:
Intel i5 (6 core)

### DIFF
--- a/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
+++ b/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
@@ -20,16 +20,16 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.time.Clock;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
-import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -212,7 +212,6 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
   // the most recently used resource).
   // These are the resources that acquire() will take.
   // This is always a subset of the other data structure mResources.
-  @GuardedBy("mLock")
   private final Deque<ResourceInternal<T>> mAvailableResources;
 
   // Tracks all the resources that are not closed.
@@ -226,6 +225,11 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
   private ScheduledExecutorService mExecutor;
   private ScheduledFuture<?> mGcFuture;
 
+  // It represents the total number of resources that have been created by this pool.
+  private AtomicInteger mCurrentCapacity;
+  // Number of threads waiting for resources
+  private AtomicInteger mWaitingThreadsNumber;
+
   protected Clock mClock = new SystemClock();
 
   /**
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
```diff
@@ -238,32 +241,37 @@ public DynamicResourcePool(Options options) {
...
+        if (shouldGc(queueTailResource)) {
+          resourcesToGc.add(queueTailResource.mResource);
+          remove(queueTailResource.mResource);
+          if (mCurrentCapacity.get() <= mMinCapacity) {
+            break;
+          }
+        } else {
+          // add the resource back the deque
+          mAvailableResources.offerLast(queueTailResource);
+          if (mWaitingThreadsNumber.get() > 0) {
```
gpang: Since the lock is no longer acquired, there is no longer consistent access between multiple variables. I see various places where decisions are made from one atomic int, that affects/mutates something else. Will that become a problem, since now those accesses are no longer consistent (within a lock)?

gengxin: > Since the lock is no longer acquired, there is no longer consistent access between multiple variables. I see various places where decisions are made from one atomic int, that affects/mutates something else. Will that become a problem, since now those accesses are no longer consistent (within a lock)?

Thanks @gpang, the purpose of this change is to make sure these variables are consistent without mLock when pool is not full or there is not waiting thread, so in most times it's lock-free. 

Combination of mCurrentCapacity and mAvailableResources is just like mCurrentCapacity and mResources in alluxio.resource.ResourcePool, so it's should be easier to understand.

I suspect your question is mainly about mWaitingThreadsNumber, which impacts whether to signal waiting threads. I think there are two corner cases when thread A calling accquire method runs to line 335 - 345.

1. A is not permitted to allocated new resource and has not yet increase mWaitingThreadsNumber.
2. A has increased mWaitingThreadsNumber, entered while loop, but has not called await.

In case 1, if another thread B releases resource, A will find the resource when running into the loop (line 341)
In case 2, if anothere thead B releases resource, A will not be notified because it's not waiting, so it will continue to call await and find the new resource in next loop. 

So the worst case is thread A throws TimeoutException even if B releases resource, and this case should be rare to happen.

-----------------------------

Below are details about the important variables and scenarios.

**variables**

- _mCurrentCapacity_ - Is AtomicInteger, used as a resource counter and permit token, only thread successfully increment this can allocate a new resource.
- _mAvailableResources_ - Is thread-safe it self, used to contain resource that are released and not yet gc.
- _mLock_ - when pool is full, this lock is used by user thread to await, or by user thread / gc thread to notify available resource.
- _mWaitingThreadsNumber_ - Is AtomicInteger, used as a counter of waiting threads, if it's 0, we don't need to notify when resource is available.

**status changing scenarios**

- _accquire_ - poll from mAvailableResources; if don't get a resource, trying to allocate new resource by trying to increase mCurrentCapacity; if not permitted, increase / release mWaitingThreadsNumber in a try / finally block and trying to await until be notified or timeout.
- _release_ - add resource to mAvailableResources, then if there is wating threads, call mLock to wake one thread
- _gc_ - remove timeout resources from mAvailableResources's queue end (oldest) and reduce mCurrentCapacity in a loop, until the queue is empty, or we removed a not-timeout resource, which should be added back to queue just like _release_ a resource and signal waiting thread if needed.

gpang:
Jenkins, test this please

apc999:
@gpang  are you good with this PR? shall I take a closer look at it ?

gpang:
@apc999 I haven't had a chance to look at it carefully yet...

Issue #10011, opened 2019-10-01
Merge helm chart b/w directories helm-chart and helm


madanadit:
**Is your feature request related to a problem? Please describe.**
We have 2 directories for helm which are out of sync after https://github.com/Alluxio/alluxio/pull/9933.

**Describe the solution you'd like**
We should merge the features supported in both and have a single helm chart

**Describe alternatives you've considered**
N/A

**Urgency**
Urgent

**Additional context**
Community contributed helm chart has diverged from the helm chart in the official Alluxio repo.

madanadit:
@cheyang created the issue to track this

madanadit:
The directory `integration/kubernetes/helm` and `integration/kubernetes/helm-chart` have diverged.

Changes from the following PRs in the directory `integration/kubernetes/helm` need to merged into the new modularized helm template structure in `integration/kubernetes/helm-chart`:
- HA including Embedded Journal: https://github.com/Alluxio/alluxio/pull/9832
- Generation of pre-defined deployment YAMLs from the helm templates: https://github.com/Alluxio/alluxio/pull/9898
- Dedicate alluxio-format job: https://github.com/Alluxio/alluxio/pull/10003
- Run as non-root user: https://github.com/Alluxio/alluxio/pull/9966
- Misc fixes: https://github.com/Alluxio/alluxio/pull/9961, https://github.com/Alluxio/alluxio/pull/10030, https://github.com/Alluxio/alluxio/pull/10039

cheyang:
Thanks @jiacheliu3 , I will also help review.

PR #9933, opened 2019-09-21
Enhance alluxio helm charts

cheyang:
To handle #9615

Features:
- Modularize the directory structure
- Java and docker natively integration(Consider memory, https://developers.redhat.com/blog/2017/03/14/java-inside-docker/)
- Two choices to decide short circuit: copying from directory directly or unix socket
- simplify tiered storage (Combine alluxio properties and Persistent volume)
- Use init container to do format /no format with intelligence
- Make more configurable, like node selector, tolerance
- Fuse daemonset

PR #10040, opened 2019-10-03
Use ALLUXIO_CLIENT_JAVA_OPTS in k8s fuse yaml to not override configMap

madanadit:
ALLUXIO_JAVA_OPTS is redundantly defined in the fuse container and the configMap which lead to configMap being ignored.

PR #10076, opened 2019-10-10
Merge helm chart

jiacheliu3:
This change merges [Enhance alluxio helm charts](https://github.com/Alluxio/alluxio/pull/9933) with the existing helm templates. This change will resolve https://github.com/Alluxio/alluxio/issues/10011

After the change all the helm templates will conform to the new structure defined in [Enhance alluxio helm charts](https://github.com/Alluxio/alluxio/pull/9933).


integration/kubernetes/helm-chart/alluxio/templates/_helpers.tpl
```diff
@@ -2,8 +2,12 @@
 {{/*
 Expand the name of the chart.
 */}}
+{{/*
+Bring the items back
+*/}}
+
 {{- define "alluxio.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
```
jiacheliu3: Please ignore the change in this file for now. (sth with template doesn't work in loops in other template files that refer here, so i disabled some logic.) This is will be reverted. This file will not be changed.

jiacheliu3:
@madanadit Changes are merged from `helm` into `helm-chart`. Do we now want to rename `helm-chart` back to `helm`?

integration/kubernetes/multiMaster-embeddedJournal/alluxio-format-journal-job.yaml.template
```diff
@@ -32,10 +38,12 @@ spec:
               cpu: 1
               memory: 1G
           command: ["/opt/alluxio/bin/alluxio"]
-          args: ["formatMasters"]
+          args: ["fsadmin", "report"]
+          # To format journal, uncomment the below args and comment the above!
```
jiacheliu3: Since we are going to change the way of creating these resources to `kubectl create -f dir/`, I don't want this job to be automatically created and wipe all the journal out. By making it run a report, the Job will just finish (or err) and die when the TTL past.

madanadit:
@jiacheliu3 i'm fine either way. the advantage of renaming it back to `helm` is that I won't have to fix the documentation.

integration/kubernetes/multiMaster-embeddedJournal/alluxio-format-journal-job.yaml.template
madanadit: this looks ugly. can we separate out the format job into a different directory instead? may the structure is `$dir/jobs/alluxio-format-journal-job`, `$dir/masters/...`, `$dir/workers/...`

jiacheliu3: Ok now the resources are put under `$dir/master` `$dir/worker` and `$dir/job`, config is directly under `$dir` as it applies to all resources

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -9,22 +9,30 @@
...
 #
 
+{{ $masterCount := int .Values.master.count }}
+{{ $defaultMasterName := "alluxio-master-0" }}
+{{- $isSingleMaster := eq $masterCount 1 }}
+{{- $isHaEmbedded := and (eq .Values.journal.type "EMBEDDED") (gt $masterCount 1) }}
+{{- $release := .Release }}
+{{- $name := include "alluxio.name" . }}
+{{- $fullName := include "alluxio.fullname" . }}
+{{- $chart := include "alluxio.chart" . }}
```
jiacheliu3: I have to use local variables like this because in `{{range}}{{end}}` loops the variable tree is different.

integration/kubernetes/helm-chart/alluxio/templates/fuse/client-daemonset.yaml
```diff
@@ -0,0 +1,59 @@
...
+      containers:
+        - name: alluxio-fuse-client
+          image: alpine:latest
+          resources:
+            requests:
+              cpu: "0.5"
+              memory: "500M"
+            limits:
+              cpu: "1"
+              memory: "500M"
```
jiacheliu3: So far we don't seem to parameterize fuse-client resource requests/limits. I'm fine it stays this way.

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -37,8 +37,8 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-fuse
     spec:
-      hostNetwork: true
-      dnsPolicy: ClusterFirstWithHostNet
+      hostNetwork: {{ .Values.fuse.hostNetwork }}
+      dnsPolicy: {{ .Values.fuse.dnsPolicy }}
```
jiacheliu3: Parameterized these two fields as we do have users that use different setting for these fields. Same applies to all other resources.

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -98,6 +90,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: status.hostIP
+          - name: ALLUXIO_CLIENT_JAVA_OPTS
+            value: " -Dalluxio.user.hostname=$(ALLUXIO_CLIENT_HOSTNAME) "
```
jiacheliu3: From https://github.com/Alluxio/alluxio/pull/10040

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -109,7 +103,7 @@ spec:
                 command: ["/opt/alluxio/integration/fuse/bin/alluxio-fuse", "unmount", "/alluxio-fuse"]
           envFrom:
           - configMapRef:
-              name: {{ template "alluxio.name" . }}-config
+              name: {{ template "alluxio.fullname" . }}-config
```
jiacheliu3: Changed to fullname or this will fail to match with the config. Same applies to all other resources.

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -9,238 +9,215 @@
...
+      {{- if $root.Values.master.nodeSelector }}
+{{ toYaml $root.Values.master.nodeSelector | trim | indent 8  }}
+      {{- else if $root.Values.nodeSelector }}
+{{ toYaml $root.Values.nodeSelector | trim | indent 8  }}
       {{- end }}
-      selector:
-        matchLabels:
-          app: {{ template "alluxio.name" . }}
-          role: alluxio-master
-          name: {{ template "alluxio.fullname" . }}-master
```
jiacheliu3: Shouldn't use `matchLabels` field under `.spec.template.spec`. Removed.

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -9,238 +9,215 @@
...
-                {{- end }}
-              {{- end }}
-            requests:
-              {{- if .Values.master.resources.requests }}
-                {{- if .Values.master.resources.requests.cpu  }}
-              cpu: {{ .Values.master.resources.requests.cpu }}
-              memory: {{ .Values.master.resources.requests.memory }}
-                {{- end }}
-              {{- end }}
-            {{- end }}
```
jiacheliu3: @cheyang I think `--no-format` option in Alluxio master container work as the journals are not formatted if they exist already. Please correct me if I'm wrong.

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -9,238 +9,215 @@
...
-            - configMapRef:
-                name: {{ template "alluxio.name" . }}-config
-          env:
-            - name: ALLUXIO_WORKER_HOSTNAME
-              valueFrom:
-                fieldRef:
-                  fieldPath: status.hostIP
-          volumeMounts:
-            - name: tz-config
-              mountPath: /etc/localtime
```
jiacheliu3: @cheyang  I chose to remove this localtime because these templates aim at giving users a good place to start with Alluxio. So we design the templates to serve the most basic use cases. `/etc/localtime` seems very specific to some use cases me so I chose to remove it for the simplicity.

Mounting `/etc/localtime` is definitely a good trick. Potentially if many users are hitting the time sync issue (for example, in logs), we should add this to the documentation or Alluxio K8s specific FAQ page.

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -9,238 +9,215 @@
...
-            {{- if .Values.master.resources  }}
-              {{- if .Values.master.resources.limits }}
-                {{- if .Values.master.resources.limits.cpu  }}
-              cpu: {{ .Values.master.resources.limits.cpu }}
-              memory: {{ .Values.master.resources.limits.memory }}
-                {{- end }}
-              {{- end }}
+              cpu: {{ $root.Values.master.resources.limits.cpu }}
+              memory: {{ $root.Values.master.resources.limits.memory }}
+            {{- end }}
```
jiacheliu3: I changed this part a bit (This applies to all resources too). I think it makes sense the user can very possibly have resource policies in real life environments, so it makes sense to have neither `requests` nor `limits`. But I think it's reasonable that the user should define both cpu and memory if they want to define some quotas. Checking the existence of both `cpu` and `memory` seems a little bit verbose to me.

This is just my personal opinion. Of course there are scenarios where the user only cares about cpu, not memory. It's just I chose the simpler way out. Open to opinions :)

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -9,238 +9,215 @@
...
             {{- end }}
+          {{- end }}
           command: ["/entrypoint.sh"]
           args: ["master-only", "--no-format"]
+          {{- if $isHaEmbedded }}
           env:
-          - name: ALLUXIO_WORKER_HOSTNAME
-            valueFrom:
-              fieldRef:
-                fieldPath: status.hostIP
```
jiacheliu3: This part was not correct so I fixed it.

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -35,61 +36,62 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-worker
     spec:
-      hostNetwork: true
-      dnsPolicy: ClusterFirstWithHostNet
+      hostNetwork: {{ .Values.worker.hostNetwork }}
+      dnsPolicy: {{ .Values.worker.dnsPolicy }}
```
jiacheliu3: Also, parameterizing these two

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -35,61 +36,62 @@ spec:
...
+      initContainers:
+      - name: socket-chown
+        securityContext:
+          runAsUser: 0
+        image: {{ .Values.image }}:{{ .Values.imageTag }}
+        command: ["/bin/chown","-R", {{ $ugi | quote }}, "/opt/domain"]
+        volumeMounts:
+        - name: alluxio-domain
+          mountPath: /opt/domain
+      {{- end }}
```
jiacheliu3: From https://github.com/Alluxio/alluxio/pull/9966

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -20,7 +20,7 @@ tieredstore:
     level: 0
     path: /dev/shm
     type: emptyDir
-    quota: 1GB
+    quota: 1G
```
jiacheliu3: K8s quota doesn't recognize "1GB", "B" should be removed (stupid)

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -34,12 +34,11 @@ domainHostPath: "/tmp/alluxio-domain"
 
 # properties for all the components
 properties:
-  alluxio.master.journal.type: UFS
-  alluxio.master.journal.folder: /journal
```
jiacheliu3: This is handled by options under `journal`

jiacheliu3:
Once the file names and paths are settled, the generate-tarball script can be updated accordingly

cheyang:
@jiacheliu3 , Thanks a lot for your PR. Can you upgrade the chart version to 0.5?

jiacheliu3:
@cheyang  the Chart version has been changed to 0.5

integration/kubernetes/alluxio-fuse-client.yaml.template
```diff
@@ -12,16 +14,29 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: alluxio-fuse-client
+  name: release-name-alluxio-fuse-client
```
madanadit: why do we have the prefix `release-name-`?

madanadit: same for all other templates

integration/kubernetes/alluxio-fuse.yaml.template
```diff
@@ -9,32 +11,64 @@
...
-              cpu: "1"
-              memory: "1G"
+          imagePullPolicy: IfNotPresent
           command: ["/entrypoint.sh"]
-          args: ["fuse", "--fuse-opts=allow_other"]
+          # args: ["fuse", "--fuse-opts=allow_other"]
+          args: 
+            - fuse
+            - --fuse-opts=allow_other
+
```
madanadit: nit: fix newline

integration/kubernetes/alluxio-fuse.yaml.template
```diff
@@ -53,7 +87,7 @@ spec:
                 command: ["/opt/alluxio/integration/fuse/bin/alluxio-fuse", "unmount", "/alluxio-fuse"]
           envFrom:
           - configMapRef:
-              name: alluxio-config
+              name: release-name-alluxio-config
```
madanadit: same here

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
madanadit: sgtm

integration/kubernetes/helm-chart/alluxio/templates/fuse/client-daemonset.yaml
madanadit: sgtm

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
madanadit: sgtm

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
madanadit: thx

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
madanadit: full name should not have the `release-name-` prefix

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
madanadit: sgtm

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
madanadit: @jiacheliu3 did you verify this is true for embedded as well?

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
madanadit: sgtm

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
madanadit: thx

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
madanadit: thx

madanadit:

PR #10087, opened 2019-10-11
Fix incorrect Helm chart file name upper/lower case

jiacheliu3:
This change fixes the incorrect upper/lower case in updated Helm chart files and `generate-tarball.go`. The file doesn't exist anymore so the tarball cannot be generated correctly.


madanadit:

PR #10091, opened 2019-10-11
[DOCFIX] Add a note for using SecondaryMaster with Embedded Journal

ggezer:


gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

ggezer:
@apc999 PTAL. Thanks!

docs/en/Architecture-DataFlow.md
```diff
@@ -105,6 +105,9 @@ the same server as the leading master to write journal checkpoints. Note that, t
 is not designed to provide high availability but offload the work from the leading master for fast
 recovery. Different from standby masters, a secondary master can never upgrade to a leading master.
 
+Note: Secondary Master is designed to work with UFS journals only. 
+`Do not launch when running with Embedded Journal.`
```
calvinjia: Do we have a check in our launch scripts for this?

ggezer: I didn't see any check.
Should we add this doc along with that or just launcher guard suffice?

ggezer: https://github.com/Alluxio/alluxio/pull/10142

gpang: Can we change the title also? `### Secondary Masters (for UFS journal)` so it is obvious it is only for ufs journal.

ggezer: I think after this section is useless after the launcher guard, because they can't launch anyway. Also updated the title and text to emphasize that this is for UFS journal only.

ggezer:
@calvinjia PTAL. Thanks.

calvinjia:

PR #10094, opened 2019-10-12
add debug info for Block Reader Handler

cheyang:


apc999:
discussed offline. This will not aim to be merged to upstream.

Issue #10043, opened 2019-10-04
Revisit the assumption of persist on rename 


apc999:
**Is your feature request related to a problem? Please describe.**

### Hive on MR
When running EMR Hive (version 2.x) using MR as the execution engine to write data to Alluxio, three renames are applied to the partition files before the final results are committed on the Hive side.

- First rename (on a partition file):
From
`/output/.hive-staging_hive_2019-10-04_16-48-19_644_1437496375846573990-1/_task_tmp.-ext-10000/dt=2019-08-14/step_check=false/_tmp.000000_0`
to `/output/.hive-staging_hive_2019-10-04_16-48-19_644_1437496375846573990-1/_tmp.-ext-10000/dt=2019-08-14/step_check=false/000000_0`

- Second rename (on a partition dir):
From
`/output/.hive-staging_hive_2019-10-04_16-48-19_644_1437496375846573990-1/_tmp.-ext-10000`
to `/output/.hive-s
taging_hive_2019-10-04_16-48-19_644_1437496375846573990-1/-ext-10000`

- Third rename (on the table dir):
From
`/output/.hive-staging_hive_2019-10-04_16-48-19_644_1437496375846573990-1/-ext-10000/dt=2019-08-14/step_check=false`
to `/output/dt=2019-08-14/step_check=false`

### Hive on Tez

Multiple renames on Tez too observed on similar queries.

- First rename from `/output/.hive-staging_hive_2019-10-07_21-36-05_239_8995819855931920865-1/_tmp.-ext-10000` to `/output/.hive-staging_hive_2019-10-07_21-36-05_239_8995819855931920865-1/-ext-10000`
- Second Rename frpm `/output/.hive-staging_hive_2019-10-07_21-36-05_239_8995819855931920865-1/-ext-10000/000000_0` to `/output/000000_0`

**Describe the solution you'd like**
When Hive is writing data with ASYNC_THROUGH with current persist on rename turned on, Alluxio will start to schedule persist jobs on the first rename, which can be inefficient as the later two renames can become on the critical path within a lock. When a table is large (e.g., a lot of partitions), the master can be very slow.

**Describe alternatives you've considered**
Ideally, Alluxio should persist files on the final commit.

**Urgency**
Medium

**Additional context**

PR #10111, opened 2019-10-15
Implement blacklist for persisting files

apc999:
Fix https://github.com/Alluxio/alluxio/issues/10043

Files containing strings in the blacklist will not be persisted. This can be useful for Alluxio to avoid persisting results too early for workloads like Hive or MapReduce as they typically create temp results first into temp directories, before the entire directory was renamed (as commit).

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1702,6 +1702,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
             + "persist) but does not affect CACHE_THROUGH writes. Users may want to specify "
             + "temporary files in the blacklist to avoid unnecessary I/O and errors. Some "
             + "examples are `.staging` and `.tmp`.")
+          .setScope(Scope.MASTER)
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
   public static final PropertyKey MASTER_REPLICATION_CHECK_INTERVAL_MS =
       new Builder(Name.MASTER_REPLICATION_CHECK_INTERVAL_MS)
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1697,7 +1697,11 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_PERSISTENCE_BLACKLIST =
       new Builder(Name.MASTER_PERSISTENCE_BLACKLIST)
-          .setDescription("Patterns to blacklist persist, comma separated, string match, no regex.")
+          .setDescription("Patterns to blacklist persist, comma separated, string match, no regex."
+            + " This affects any async persist call (including ASYNC_THROUGH writes and CLI "
+            + "persist) but does not affect CACHE_THROUGH writes. Users may want to specify "
+            + "temporary files in the blacklist to avoid unnecessary I/O and errors. Some "
+            + "examples are `.staging` and `.tmp`.")
           .build();
   public static final PropertyKey MASTER_REPLICATION_CHECK_INTERVAL_MS =
       new Builder(Name.MASTER_REPLICATION_CHECK_INTERVAL_MS)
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1985,6 +1985,21 @@ public void rename(AlluxioURI srcPath, AlluxioURI dstPath, RenameContext context
     }
   }
 
+  private boolean shouldPersistPath(String path) {
+    List<String> blacklist =
+        ServerConfiguration.isSet(PropertyKey.MASTER_PERSISTENCE_BLACKLIST)
+            ? ServerConfiguration.getList(PropertyKey.MASTER_PERSISTENCE_BLACKLIST, ",")
```
apc999: shall we make `ServerConfiguration.getList` outside this local method to avoid this string operation every time?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2070,6 +2087,35 @@ private void renameInternal(RpcContext rpcContext, LockedInodePath srcInodePath,
...
+        && shouldPersistPath(dstInodePath.toString())) {
+      LOG.debug("Schedule Async Persist on rename for Dir {}", dstInodePath.toString());
+      try (LockedInodePathList descendants = mInodeTree.getDescendants(srcInodePath)) {
+        for (LockedInodePath childPath : descendants) {
+          Inode childInode = childPath.getInode();
+          // TODO(apc999): Resolve the child path legitimately
+          if (childInode.isFile() && !childInode.isPersisted()
+              && shouldPersistPath(
+                  childPath.toString().substring(srcInodePath.toString().length()))) {
+            LOG.debug("Schedule Async Persist on rename for Child File {}", childPath.toString());
```
apc999: `childPath` should be sufficient rather than `childPath.toString()`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2056,7 +2071,9 @@ private void renameInternal(RpcContext rpcContext, LockedInodePath srcInodePath,
 
     // Check options and determine if we should schedule async persist. This is helpful for compute
     // frameworks that use rename as a commit operation.
-    if (context.getPersist() && srcInode.isFile() && !srcInode.isPersisted()) {
+    if (context.getPersist() && srcInode.isFile() && !srcInode.isPersisted()
+        && shouldPersistPath(dstInodePath.toString())) {
+      LOG.debug("Schedule Async Persist on rename for File {}", srcInodePath.toString());
```
apc999: nit: `srcInodePath` rather than `srcInodePath.toString()`?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1695,6 +1695,10 @@ public String toString() {
           .setDescription("How often the master schedules persistence jobs "
               + "for files written using ASYNC_THROUGH")
           .build();
+  public static final PropertyKey MASTER_PERSISTENCE_BLACKLIST =
+      new Builder(Name.MASTER_PERSISTENCE_BLACKLIST)
+          .setDescription("Patterns to blacklist persist, comma separated, string match, no regex.")
```
apc999: in description, can you give a few example, like "staging_hive" and etc and why this may help?

gpang: does this apply to all persistence concepts, include CACHE_THROUGH, as well as a user CLI initiated persist rpc?

Also, is this string match for any part of the full path, or just the last component? Can the string match span directories, like `o/b` for `/foo/bar/`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: This breaks unit tests unfortunately. I could add another method to reset it if you think the performance is nontrivial.

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: It does not apply to cache through, it applies to cli `persist`. String match is on the full string.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: Updated

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Updated the description

apc999:
LGTM

calvinjia:

PR #10115, opened 2019-10-15
Load libraries from lib/ directory

gpang:
Load libraries from the lib/ directory, instead of hard coding the dependency in the pom files.


gpang:

PR #10117, opened 2019-10-16
Fix the way we pass parameters to lerna so it is happy

yuzhu:
basically, maven puts a quote around each parameter, and forces npm to take the entire parameter as a single parameter.

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -2,7 +2,7 @@
   "name": "root",
   "private": true,
   "dependencies": {
-    "lerna": "^3.13.0"
+    "lerna": "~3.13.0"
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


yuzhu:
@Xenorith with this change, do we want to pin to a specific version?  Looking at other open source project, particularly, lerna itself. it seems like they use `^x.y.z` most commonly.

Xenorith:
hm in other places where we use lerna, we are using `~` to pin down the minor version, so i would prefer to maintain consistency, even though the usage/invocation pathway may be different

yuzhu:
@Xenorith fixed version.

yuzhu:

yuzhu:

PR #10119, opened 2019-10-16
Support structured data transformation

chcn:
From user's perspective:
`./bin/alluxio table transform default test 'write(hive).option(hive.num.files, 2)'` will compact/coalesce files under each partition of table `test` in database `default` into at most 2 files. The transformed data is stored in Alluxio internally managed directories. After transformation, new partition's metadata is updated automatically in the Alluxio table catalog, so Presto users can run queries against the transformed data transparently.

The current supported source format includes Parquet and CSV, the output format currently only supports Parquet. But the design takes into consideration of extending to other formats.

The transformation job architecture and the TransformManager subsystem in TableMaster are designed to support new types of transformations.


table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveLayout.java
```diff
@@ -100,14 +100,27 @@ public PartitionInfo getData() {
     return mPartitionStatsInfo;
   }
 
+  @Override
+  public String getLocation() {
+    return mPartitionInfo.getStorage().getLocation();
```
gpang: instead of having a layout return a location, shouldn't `getTransformPlan` construct the jobs (and the appropriate locations), so external classes don't need to know about locations? Locations might be a concept specific to individual layouts.

pom.xml
```diff
@@ -509,6 +510,16 @@
...
         <version>4.4.6</version>
       </dependency>
+      <dependency>
+        <groupId>org.apache.parquet</groupId>
+        <artifactId>parquet-avro</artifactId>
+        <version>${parquet.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.parquet</groupId>
+        <artifactId>parquet-cli</artifactId>
```
gpang: is using the "cli" artifact, correct?

table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveLayout.java
yuzhu: my two cents here.  I ran into a similar situation in updating table layout / location also because we have abstracted the location away from the layout.  IMO, location might mean different things to different layout but they all should have one? especially because `TransformContext` already exposes a location concept?

table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveLayout.java
```diff
@@ -100,14 +100,27 @@ public PartitionInfo getData() {
...
+  }
+
+  private HiveLayout transformLayout(AlluxioURI transformOutputPath) {
+    PartitionInfo info = mPartitionInfo.toBuilder()
+        .setStorage(mPartitionInfo.getStorage().toBuilder()
+            .setLocation(transformOutputPath.getPath())
```
yuzhu: should we be using `transformOutputPath.toString()`  to get the  scheme and authority as well?

chcn:
jenkins, test this please

conf/core-site.xml.template
```diff
@@ -11,6 +11,12 @@
 
 -->
 <configuration>
+  <!-- Must exist to transform structured tables. -->
```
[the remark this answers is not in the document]
chcn: `ParquetReader` uses parquet-mr library, which uses hadoop FileSystem, without this configuration, files in Alluxio cannot be read.

gpang:
@chcn checkstyle issues:
```
/usr/src/alluxio/table/server/master/src/main/java/alluxio/master/table/Table.java:60: error: Extra newline
/usr/src/alluxio/table/server/master/src/main/java/alluxio/master/table/transform/TransformManager.java:380:5: error: Missing a Javadoc comment.
```

table/server/master/src/main/java/alluxio/master/table/Table.java
```diff
@@ -45,6 +49,15 @@
   private String mOwner;
   private List<ColumnStatisticsInfo> mStatistics;
   private Map<String, String> mParameters;
+  private Map<String, Partition> mSpecToPartitions;
+
+  private void computeSpecToPartitions(PartitionScheme partitionScheme) {
```
gpang: I think this functionality should be owned by the `PartitionScheme`, so `PartitionScheme` can expose an api like `get(spec)`.

core/server/common/src/main/java/alluxio/master/journal/DelegatingGroupJournaled.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Convenience interface for classes which delegate journaling to a list of journaled components.
+ */
+public interface DelegatingGroupJournaled extends Journaled {
```
gpang: Hrmmm, I feel like this is a generalized version of `DelegatingJournaled`. Ideally, this should probably replace DelegatingJournaled`, be merge into master, as an independent commit, so that change is isolated.

gpang: I ported it to master with this PR: https://github.com/Alluxio/alluxio/pull/10151

table/server/master/src/main/java/alluxio/master/table/DefaultTableMaster.java
```diff
@@ -117,6 +131,7 @@ public String getName() {
   @Override
   public void start(Boolean isLeader) throws IOException {
     super.start(isLeader);
+    mTransformManager.startHeartbeat(getExecutorService(), mMasterContext.getUserState());
```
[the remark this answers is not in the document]
gpang: Will this be stopped correctly when needed?

table/server/master/src/main/java/alluxio/master/table/Table.java
chcn: added this interface API and abstracted out a `BasePartitionScheme` to provide default implementation

table/server/master/src/main/java/alluxio/master/table/DefaultTableMaster.java
```diff
@@ -135,8 +159,13 @@ public void close() throws IOException {
     super.close();
   }
 
+  @Override
```
gpang: we can get rid of this, since we are now using `DelegatingJournaled`, right?

table/server/master/src/main/java/alluxio/master/table/Table.java
```diff
@@ -46,6 +49,7 @@
   private List<ColumnStatisticsInfo> mStatistics;
   private Map<String, String> mParameters;
 
+
```
gpang: Remove extra blank line

chcn:
jenkins, test this please

gpang:
failed (maybe flaky): `alluxio.server.ft.journal.MultiMasterJournalTest.testCheckpointReplay`

gpang:
Jenkins, test this please

gpang:
failed (flaky): `alluxio.server.ft.journal.MultiMasterJournalTest.testTtl`

pom.xml
chcn: CsvReader depends on a utility class from cli.
In the future, if we implement our own CSV reader, we can get rid of this dependency.

table/server/master/src/main/java/alluxio/master/table/DefaultTableMaster.java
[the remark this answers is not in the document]
chcn: it will be stopped when master stops the executor service for heartbeat threads.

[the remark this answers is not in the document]
chcn: yes, added a check for leader

gpang:

table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveLayout.java
chcn: currently, job services need to know the physical location. In the future, we can modify job service to use protobuf serialization for job configurations, in that way, the input and output of some jobs can be the protobuf definition of Layout, then we can completely get rid of `getLocation`.

conf/core-site.xml.template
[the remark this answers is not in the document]
chcn: in Hadoop 2, this is no longer needed. As we do not target Hadoop 1, we can get rid of this.

table/server/under/hive/src/main/java/alluxio/table/under/hive/HiveLayout.java
chcn: yes, I now put scheme and authority into the location too

PR #10120, opened 2019-10-16
Update the option name in distributedLoad

apc999:
Consolidate the option name in `distributedLoad` and `persist` regarding parallelism. Using `--thread` is confusing as it is not indicating how many working threads on that machine but number of concurrent operations across an Alluxio cluster


calvinjia:

apc999:
@liuhongtong for the coming 2.1 release, I updated the option name for distributedLoad from `--thread` to `--parallelism` to be consistent with existing options in `persist` command. Just FYI

PR #10121, opened 2019-10-16
Improve distributedCp and distributedMv and Persist

apc999:
- Set readtype to `NO_CACHE` to `distributedCp` and `distributedMv` (in migration definition)
- Increase buffersize to 8 MB (from 8KB or less) for these jobs
- Make `distributedMv` carry writetype

### DIFF
--- a/job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
+++ b/job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
@@ -12,6 +12,7 @@
 package alluxio.job.migrate;
 
 import alluxio.AlluxioURI;
+import alluxio.Constants;
 import alluxio.client.WriteType;
 import alluxio.client.block.AlluxioBlockStore;
 import alluxio.client.block.BlockWorkerInfo;
@@ -26,6 +27,8 @@ import alluxio.exception.FileAlreadyExistsException;
 import alluxio.exception.FileDoesNotExistException;
 import alluxio.grpc.CreateFilePOptions;
 import alluxio.grpc.DeletePOptions;
+import alluxio.grpc.OpenFilePOptions;
+import alluxio.grpc.ReadPType;
 import alluxio.grpc.WritePType;
 import alluxio.job.AbstractVoidJobDefinition;
 import alluxio.job.RunTaskContext;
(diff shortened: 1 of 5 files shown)


gpang:
I'm curious, how much does the buffer size change affect loading a large file?

job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
```diff
@@ -308,10 +311,12 @@ private static void migrate(MigrateCommand command, WritePType writeType, boolea
 
     CreateFilePOptions createOptions =
         CreateFilePOptions.newBuilder().setWriteType(writeType).build();
-
-    try (FileOutStream out = fileSystem.createFile(new AlluxioURI(destination), createOptions)) {
-      try (FileInStream in = fileSystem.openFile(new AlluxioURI(source))) {
-        IOUtils.copy(in, out);
+    OpenFilePOptions openFileOptions =
+        OpenFilePOptions.newBuilder().setReadType(ReadPType.NO_CACHE).build();
```
calvinjia: I don't like hardcoding the read type, there could be some cases where the user wants to cache, could we pass it in as well?

apc999: do you mean put readtype  in `MigrateConfig`and set `NO_CACHE` in `distributed{Cp,Mv}`?

calvinjia: Ideally we would always use the user defined read/write mode and perhaps change the default read mode to `NO_CACHE` for cp/mv but allow the user to override that if they want.

apc999: I am for making Migrate job more flexible , 
but I am hesitate to provide more CLI tuning flags to users at this point.

apc999:
@gpang I was running a test of persisting a 100GB file to s3, with the previous tiny buffer I was not able to finish it; after changing to the large (8MB) buffer, it completes within a hour.

job/server/src/main/java/alluxio/job/migrate/MigrateDefinition.java
calvinjia: Sure, we can change this when we run into the case where `NO_CACHE` hardcoded is a problem.

calvinjia:

apc999:

apc999:

PR #10124, opened 2019-10-16
Decrease task priority slightly to deprioritize during contention

bradyoo:


bradyoo:
image

(The image is for when I was experimenting with the lowest priority.)

bradyoo:
@calvinjia

bradyoo:
It is unclear whether we want to do this.

calvinjia:
@bradyoo What impact does this have on resource sharing? Should the thread priority be configurable?

bradyoo:
It appears to have some small impact but it really depends on the
situation. For now, the intention for it to be slightly lower than all the
other things that could be running on the worker. We could make this
configurable but I would prefer not because we may want to reimplement
parts of this.

PR #10128, opened 2019-10-16
Propagate UserState to PollingMasterInquireClient

bf8086:
Currently `PollingMasterInquireClient` uses a new `Subject` when querying master. If the call needs authentication using credentials from `Subject`, it will fail the request because the credentials are missing. This change add parameters to APIs to allow passing in the required `Subject`.

Manually tested with external `Subject` from Hadoop client.

### DIFF
--- a/minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
+++ b/minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
@@ -41,7 +41,7 @@ import alluxio.master.ZkMasterInquireClient;
 import alluxio.master.journal.JournalType;
 import alluxio.multi.process.PortCoordination.ReservedPort;
 import alluxio.network.PortUtils;
-import alluxio.security.user.UserState;
+import alluxio.security.user.ServerUserState;
 import alluxio.util.CommonUtils;
 import alluxio.util.WaitForOptions;
 import alluxio.util.io.PathUtils;
@@ -674,9 +674,8 @@ public final class MultiProcessCluster {
             addresses.add(
                 InetSocketAddress.createUnresolved(address.getHostname(), address.getRpcPort()));
           }
-          UserState userState = UserState.Factory.create(ServerConfiguration.global());
           return new PollingMasterInquireClient(addresses, ServerConfiguration.global(),
-              userState);
+              ServerUserState.global());
         } else {
           return new SingleMasterInquireClient(InetSocketAddress.createUnresolved(
               mMasterAddresses.get(0).getHostname(), mMasterAddresses.get(0).getRpcPort()));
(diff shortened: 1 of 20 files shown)


bf8086:
@gpang @ggezer PTAL.

gpang:
This should pass in the `UserState`, instead of `Subject`.

bf8086:
Thanks @gpang! I updated the PR. PTAL.

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -673,7 +674,9 @@ public synchronized MasterInquireClient getMasterInquireClient() {
             addresses.add(
                 InetSocketAddress.createUnresolved(address.getHostname(), address.getRpcPort()));
           }
-          return new PollingMasterInquireClient(addresses, ServerConfiguration.global());
+          UserState userState = UserState.Factory.create(ServerConfiguration.global());
```
gpang: should this use `ServerUserState`?

tests/src/test/java/alluxio/server/ft/MasterFaultToleranceIntegrationTest.java
```diff
@@ -260,8 +260,9 @@ public void killStandby() throws Exception {
   public void queryStandby() throws Exception {
     List<InetSocketAddress> addresses = mMultiMasterLocalAlluxioCluster.getMasterAddresses();
     Collections.shuffle(addresses);
-    PollingMasterInquireClient inquireClient =
-        new PollingMasterInquireClient(addresses, ServerConfiguration.global());
+    PollingMasterInquireClient inquireClient = new PollingMasterInquireClient(addresses,
+        ServerConfiguration.global(),
+        FileSystemContext.create(ServerConfiguration.global()).getClientContext().getUserState());
```
gpang: is this allowed to use `ServerUserState`?

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
bf8086: Updated.

tests/src/test/java/alluxio/server/ft/MasterFaultToleranceIntegrationTest.java
bf8086: It should ideally use a "client" user state but for this test specifically it does not make any difference.

bf8086:
Thanks @gpang! PTAL.

gpang:

gpang:
@bf8086 does this need to be cherry-picked to `branch-2.0`?

bf8086:
@gpang yeah it looks like the same code path exists in `branch-2.0` as well

gpang:

gpang:
@bf8086 the auto-cherry-pick doesn't merge cleanly, so you will have to cherry-pick manually.

PR #10139, opened 2019-10-17
Remove HDFS 1.x from default built modules

ns1123:
Removing HDFS 1.x from the default modules we build in our tarball


calvinjia:

Issue #10122, opened 2019-10-16
Raft journal replay is slow even when formatted


chcn:
**Summary**
I formatted the journal, then restart master, `RaftJournalSystem - Caught up in 20032ms`, it's unreasonably slow.

**Urgency**
This slowness means there is some overhead in RaftJournalSystem during initialization.

PR #10140, opened 2019-10-17
Override EJ election timeout for single-master cluster

ggezer:
EmbeddedJournal cluster will take 2 election timeouts to elect a leader even when it's a single master. This change is to override election timeouts to lower values in single master cluster.

Fixes https://github.com/Alluxio/alluxio/issues/10122

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1258,14 +1258,6 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT =
-      new Builder(Name.MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT)
-          .setDescription(
-              "The election timeout to use for single master embedded journal cluster.")
-          .setDefaultValue("500ms")
-          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
-          .setScope(Scope.MASTER)
-          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL)
           .setDescription(
@@ -3997,8 +3989,6 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.embedded.journal.addresses";
     public static final String MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT =
         "alluxio.master.embedded.journal.election.timeout";
-    public static final String MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT =
-        "alluxio.master.embedded.journal.single.master.election.timeout";
     public static final String MASTER_EMBEDDED_JOURNAL_APPENDER_BATCH_SIZE =
         "alluxio.master.embedded.journal.appender.batch.size";
     public static final String MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1258,6 +1258,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT)
+          .setDescription(
+              "The election timeout to use for single master embedded journal cluster.")
+          .setDefaultValue("500ms")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL)
           .setDescription(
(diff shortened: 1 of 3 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -175,6 +175,15 @@
...
    */
   private RaftJournalSystem(RaftJournalConfiguration conf) {
+    // Override election/heartbeat timeouts for single master cluster.
+    if (conf.getClusterAddresses().size() == 1) {
+      long electionTimeout = ServerConfiguration
+          .getMs(PropertyKey.MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT);
+      LOG.debug("Overriding election timeout to {} for single master cluster.", electionTimeout);
+      conf.setElectionTimeoutMs(electionTimeout);
+      // Use the highest heartbeat internal relative to election timeout.
+      conf.setHeartbeatIntervalMs(Math.max(1, (electionTimeout / 2) - 1));
```
ZacBlanco: is there somewhere that specifies using this `Math.max(...)` as the guideline? Or another piece of code perhaps?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -175,6 +175,15 @@
    * @param conf raft journal configuration
    */
   private RaftJournalSystem(RaftJournalConfiguration conf) {
+    // Override election/heartbeat timeouts for single master cluster.
+    if (conf.getClusterAddresses().size() == 1) {
+      long electionTimeout = ServerConfiguration
+          .getMs(PropertyKey.MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT);
```
ZacBlanco: Are there consequences if someone were to add another master to the cluster?
- Should we just not support this?
- Should we update the docs to specify this behavior?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1258,6 +1258,14 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT)
+          .setDescription(
+              "The election timeout to use for single master embedded journal cluster.")
```
ZacBlanco: Does this fit on one line?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1258,6 +1258,14 @@ public String toString() {
...
+          .setDefaultValue("500ms")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
```
calvinjia: Is there a case when a user would want to configure this? I would like to avoid adding confs if possible.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3989,6 +3997,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.master.embedded.journal.addresses";
     public static final String MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT =
         "alluxio.master.embedded.journal.election.timeout";
+    public static final String MASTER_EMBEDDED_JOURNAL_SINGLE_MASTER_ELECTION_TIMEOUT =
+        "alluxio.master.embedded.journal.single.master.election.timeout";
```
ZacBlanco: Does this value really need to be configurable? I think more apt behavior would be to respect the original timeout if a user configures it, but if not set in configuration (i.e. typical for local testing), just use a saner, smaller value

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Agreed.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Agreed.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: N/A

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ggezer: updated the docs.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ggezer: After getting rid of prop for single-master timeout. I think it's not necessary to document this at all.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -122,6 +122,8 @@
 public final class RaftJournalSystem extends AbstractJournalSystem {
   private static final Logger LOG = LoggerFactory.getLogger(RaftJournalSystem.class);
 
+  private static final long SINGLE_MASTER_ELECTION_TIMEOUT = 500;
```
ZacBlanco: `_MS`?

docs/en/deploy/Running-Alluxio-On-a-HA-Cluster.md
```diff
@@ -351,6 +351,9 @@ Below are the steps to add a new master to a live cluster:
 
 Note: Adding to an already shut down cluster still requires adding only single master at a time.
 
+Note: When adding a master to a single master cluster, you should shut down and update configuration for the existing master.
```
ZacBlanco: Reference which properties should be configured/looked at? or even just another doc regarding configuring EJ

ZacBlanco:
jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -195,6 +197,23 @@ public static RaftJournalSystem create(RaftJournalConfiguration conf) {
     return system;
   }
 
+  private RaftJournalConfiguration processRaftConfiguration(RaftJournalConfiguration conf) {
+    // Override election/heartbeat timeouts for single master cluster
+    // if election timeout is not set explicitly.
+    // This is to speed up single master cluster boot-up.
+    if (conf.getClusterAddresses().size() == 1
+        && !ServerConfiguration.isSet(PropertyKey.MASTER_EMBEDDED_JOURNAL_ELECTION_TIMEOUT)) {
+      LOG.debug("Overriding election timeout to {} for single master cluster.",
```
calvinjia: `{}ms`?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -195,6 +197,23 @@ public static RaftJournalSystem create(RaftJournalConfiguration conf) {
...
+          SINGLE_MASTER_ELECTION_TIMEOUT_MS);
+      conf.setElectionTimeoutMs(SINGLE_MASTER_ELECTION_TIMEOUT_MS);
+      // Use the highest heartbeat internal relative to election timeout.
+      conf.setHeartbeatIntervalMs(Math.max(1, (SINGLE_MASTER_ELECTION_TIMEOUT_MS / 2) - 1));
```
calvinjia: We don't need this max anymore right?

ggezer: Right. don't need to guard against incorrectly supplied values.

ggezer:
Jenkins, test this please

calvinjia:

PR #10144, opened 2019-10-17
Add recently modified/failed jobs to the report

bradyoo:
Don't need this in 2.1.0 so it can wait

Examples:

example_01
example_02

### DIFF
--- a/shell/src/main/java/alluxio/cli/fsadmin/report/JobServiceMetricsCommand.java
+++ b/shell/src/main/java/alluxio/cli/fsadmin/report/JobServiceMetricsCommand.java
@@ -61,7 +61,7 @@ public class JobServiceMetricsCommand {
     }
 
     mPrintStream.println();
-    mPrintStream.println("Last 10 Activities:");
+    mPrintStream.println("10 Most Recently Modified Jobs:");
 
     List<JobInfo> lastActivities = jobServiceSummary.getRecentActivities();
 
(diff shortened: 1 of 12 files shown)


bradyoo:
@calvinjia

calvinjia:
@bradyoo Nice! Is it a large change to add the job type to `JobInfo`?

bradyoo:
I was going to look at some other things first. And if we wanted this for 2.1.0, we should put this in as is.
 
The things I wanted to look at before are:
1. generateTarballs taking significantly longer
2. I'm noticing that the master is having these phantom jobs that are not actually real. You can see above that it claims it has 48 running but I doubt they are actually running anywhere.

shell/src/main/java/alluxio/cli/fsadmin/report/JobServiceMetricsCommand.java
```diff
@@ -52,6 +59,31 @@ public int run() throws IOException {
...
+
+    for (JobInfo lastActivity : lastActivities) {
+      mPrintStream.print(String.format("Timestamp: %-30s",
+          CommonUtils.convertMsToDate(lastActivity.getLastStatusChangeMs(), mDateFormatPattern)));
+      mPrintStream.print(String.format("Job Id: %-20s", lastActivity.getJobId()));
+      mPrintStream.println(String.format("Status: %s", lastActivity.getStatus()));
+    }
+
+    mPrintStream.println();
+    mPrintStream.println("Last 10 Failures:");
```
calvinjia: This is not 100% accurate right, its more like `last 10 failures in recent history`?

bradyoo: Yeah I should say Last 10 **** in recent history for both.

bradyoo:
Addressed feedback + fixed test.

calvinjia:

PR #10146, opened 2019-10-17
[DOCFIX] Add vi/vim support in Fuse docs

LuQQiu:
Modify the doc to document it support vi/vim.

Other than vi/vim, other operations which do the append, modify or override still not work.
Cp when destination file exist will file.
Append to a file will fail.
Sed to replace content in a file will fail.

Vi/Vim may just be sepcial case that it internally do create a new file, delete old file, rename the new file to old name operation combinations.


LuQQiu:
jenkins, test this please

apc999:

apc999:

PR #10004, opened 2019-09-30
Add alluxio core client runtime shaded jar

LuQQiu:
Modify the Alluxio core client runtime module to publish alluxio core client runtime jar with dependencies when releasing Alluxio.

Fix https://github.com/Alluxio/alluxio/issues/9988

PR #10149, opened 2019-10-18
Replace runtime client module with shaded client

apc999:
Follow up of https://github.com/Alluxio/alluxio/pull/10004 to remove runtime client jar completely.


apc999:
@LuQQiu I did basic tests locally,it can gerenate the client locally correctly.
Can you help make some tests?

apc999:
@calvinjia , @LuQQiu , do you have preference of using a "new" module `alluxio-shaded-client` or re-use existing `alluxio-core-client-runtime` module for this purpose? I could see pros and cons for both approaches per my discussion with @ZacBlanco , tho I don't have a strong preference of each. Want to hear from you

calvinjia:
@apc999 I have a slight preference for a new module because the contents will not only be the core client.

LuQQiu:
I will prefer to put it under the `core/client/` instead of `shaded/` so that it's more obvious what content this jar contains. but since we already made a Presto PR, I think we don't have obvious problems to go by `shaded/client`.

apc999:

LuQQiu:

PR #10152, opened 2019-10-18
Change our startup script and mount script to support multiple ramdisks

yuzhu:


yuzhu:
@ZacBlanco PTAL, this fixes mount and startup script to support multiple ramdisks and multiple dirs for tier0 locations.

bin/alluxio-common.sh
```diff
@@ -0,0 +1,32 @@
...
+# either express or implied, as more fully set forth in the License.
+#
+# See the NOTICE file distributed with this work for information regarding copyright ownership.
+#
+
+# Utility functions for alluxio scripts 
+
+function get_ramdisk_array() {
+  local tier_path=$(${BIN}/alluxio getConf alluxio.worker.tieredstore.level0.dirs.path)
+  local medium_type=$(${BIN}/alluxio getConf alluxio.worker.tieredstore.level0.dirs.mediumtype)
```
ZacBlanco: `${BIN}` is never defined in this script?

bin/alluxio-mount.sh
```diff
@@ -117,7 +107,7 @@ function mount_ramfs_mac() {
 
 function umount_ramfs_mac() {
   TIER_PATH=${1}
-  local device=$(df -l | grep ${TIER_PATH} | cut -d " " -f 1)
+  local device=$(df -l | grep -w ${TIER_PATH} | cut -d " " -f 1)
```
ZacBlanco: why add `-w` to the greps?

bin/alluxio-common.sh
```diff
@@ -0,0 +1,32 @@
...
+  local patharray
+  local mediumtypearray
+  local oldifs=$IFS
+  IFS=','
+  read -ra patharray <<< "$tier_path"
+  read -ra mediumtypearray <<< "$medium_type"
+  
+  RAMDISKARRAY=()
```
ZacBlanco: I'm not a huge fan of storing the result in a variable the caller needs to know about, but returning arrays in bash is messy, so I'm ok with the simple solution.

bin/alluxio-common.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied, as more fully set forth in the License.
+#
+# See the NOTICE file distributed with this work for information regarding copyright ownership.
+#
+
+# Utility functions for alluxio scripts 
+
+function get_ramdisk_array() {
```
ZacBlanco: Add documentation on args, return values, dependencies on env variables, etc...

bin/alluxio-start.sh
```diff
@@ -100,21 +102,26 @@ check_mount_mode() {
...
-      if [[ $? -ne 0 ]]; then
-        echo "ERROR: Ramdisk ${tier_path} is not mounted with mount option NoMount. Use alluxio-mount.sh to mount ramdisk." >&2
-        echo -e "${USAGE}" >&2
+      for tier_path in "${RAMDISKARRAY[@]}"
+      do
+        is_ram_folder_mounted "${tier_path}"
+        if [[ $? -ne 0 ]]; then
+          echo "ERROR: Ramdisk ${tier_path} is not mounted with mount option NoMount. Use alluxio-mount.sh to mount ramdisk." >&2
+          echo -e "${USAGE}" >&2
         exit 1
```
ZacBlanco: indent?

bin/alluxio-start.sh
```diff
@@ -133,21 +140,24 @@ do_mount() {
...
-        MOUNT_FAILED=$?
-      fi
+      for tier_path in "${RAMDISKARRAY[@]}"
+      do
+        is_ram_folder_mounted "${tier_path}" # Returns 0 if already mounted.
+        if [[ $? -eq 0 ]]; then
+          echo "Ramdisk already mounted. Skipping mounting procedure."
+        else
+          echo "Ramdisk not detected. Mounting..."
+          ${LAUNCHER} "${BIN}/alluxio-mount.sh" $1
```
ZacBlanco: quote `"$1"`?

bin/alluxio-mount.sh
yuzhu: because a normal grep can return multiple entries.. when you grep for `Volumes/ramdisk` it returns both `/Volumes/ramdisk` and `/Volumes/ramdisk2`, and it confuses the subsequent script

bin/alluxio-common.sh
yuzhu: yeah.. exactly my thought as well

bin/alluxio-common.sh
yuzhu: done

bin/alluxio-start.sh
yuzhu: done

bin/alluxio-start.sh
yuzhu: done

bin/alluxio-common.sh
yuzhu: done

yuzhu:
@ZacBlanco comments addressed, PTAL

bin/alluxio-mount.sh
ZacBlanco: I think my main concern is that the `-w` is not an option specified for `grep` in the posix standard - so it makes the script less portable. I am hesitant about using this.

Do you think there is another way that has higher portability?

ZacBlanco: This is the latest spec for `grep`: https://pubs.opengroup.org/onlinepubs/9699919799/

yuzhu: changed to using `-E "(^|[^A-Za-z0-9_])${TIER_PATH}($|[^A-Za-z0-9_])"`

yuzhu: what about `grep -E "(^|[[:space:]])${TIER_PATH}($|[[:space:]])"`

yuzhu:
jenkins, test this please

ZacBlanco:
jenkins, test this please

yuzhu:
@ZacBlanco PTAL

yuzhu:

PR #10160, opened 2019-10-20
Fix K8s container runAsGroup typo

jiacheliu3:
The securityContext for K8s containers has a typo in `runAsGroup` field that runs in GID 10000 instead of 1000. Group `alluxio` has GID 1000. This change fixed the GID to the correct one for user `alluxio`.


madanadit:

PR #10174, opened 2019-10-21
Make client throw NotFoundException when service is not found

bf8086:
Currently if a service is not found on server, the client will throw `UnavailableException` which is confusing and causing unnecessary retries. This change makes it throw `NotFoundException` which is what the server is actually throwing. To differentiate it from other `NotFoundException` such as file not found, a `ServiceNotFoundException` is included as cause.

### DIFF
--- a/core/common/src/test/java/alluxio/AbstractClientTest.java
+++ b/core/common/src/test/java/alluxio/AbstractClientTest.java
@@ -78,7 +78,7 @@ public final class AbstractClientTest {
     }
   }
 
-  private static class TestSerivceNotFoundClient extends BaseTestClient {
+  private static class TestServiceNotFoundClient extends BaseTestClient {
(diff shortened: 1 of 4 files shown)


core/common/src/test/java/alluxio/AbstractClientTest.java
```diff
@@ -71,11 +73,27 @@ protected long getServiceVersion() {
     }
 
     @Override
-    protected long getRemoteServiceVersion() {
+    protected long getRemoteServiceVersion() throws AlluxioStatusException {
       return mRemoteServiceVersion;
     }
   }
 
+  private static class TestSerivceNotFoundClient extends BaseTestClient {
```
ggezer: NIT: Serivce -> Service

bf8086:

PR #10180, opened 2019-10-22
Fix K8s worker start arguments

jiacheliu3:
This change corrects start arguments for Alluxio worker containers, from `worker` to `worker-only`. Since we have worker and job-worker 2 separate processes, if we start worker with `worker` argument the ports will be occupied and the job-worker will not be able to start.

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
@@ -270,9 +270,9 @@ spec:
         - name: {{ .alias | lower }}
           emptyDir:
             medium: "Memory"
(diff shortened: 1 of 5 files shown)


integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -270,9 +270,9 @@ spec:
         - name: {{ .alias | lower }}
           emptyDir:
             medium: "Memory"
-                  {{- if .quota }} 
+                  {{- if .quota }}
```
jiacheliu3: These 2 trailing whitespace are removed so extra whitespace in templates are removed.

madanadit:

madanadit:

Issue #10066, opened 2019-10-08
Allow backups on any file system


ggezer:
**Is your feature request related to a problem? Please describe.**
Currently backup path should either be in root-ufs or local. This means if your backup resides anywhere else, you should first make a copy.

**Describe the solution you'd like**
Allow reading/writing backups from arbitrary file system schemes.

**Additional context**
https://github.com/Alluxio/alluxio/pull/10046

ggezer:
@ns1123 WDYT?

ns1123:
The easy fix for this is to mount the UFS that you want to configure to take backups to. Is there a scenario where this wouldn't be needed?

ggezer:
Is there a reason why we currently restrict backups to root&local ufs?

yuzhu:
@ggezer  do we plan to address this in the upcoming release?

ns1123:
I think this can be closed. Backup related issues have mostly been addressed. Scenarios with backing up to a different storage system haven't been requested.

yuzhu:
Closing this as discussed.

ggezer:
This restriction is too inconvenient

jja725:
@jenoudet do you think this is still a valid request? otherwise I would close this, thanks

jenoudet:
@jja725 It's a valid request but definitely not a priority.

jja725:
Will close it for now, feel free to reopen it and contact us if this is a strong request.

PR #10182, opened 2019-10-22
Create backups on stand-by masters

ggezer:
This is to add a capability for delegating backups to stand-by masters in HA cluster. Hence, avoiding the need to pause metadata changes on leader during the whole backup.


core/server/master/src/main/java/alluxio/master/backup/BackupLeaderRole.java
```diff
@@ -0,0 +1,293 @@
...
+        }
+        // Send backup request along with consistent journal sequences.
+        sendMessageBlocking(connection, new BackupRequestMessage(request, journalSequences));
+        // Delegation successful.
+        mBackupControl = new Pair<>(connection, null);
+        // Start abandon timer.
+        adjustAbandonTimeout(false);
+        return true;
+      } catch (Exception e) {
+        LOG.warn(String.format("Failed to delegate backup to worker: %s", connection), e);
```
LuQQiu: The same. in Alluxio logs, user may think it's Alluxio worker instead of backup worker.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,369 @@
...
+
+  /**
+   * Handler for backup request message. It's used in secondary master.
+   */
+  private void handleRequestMessage(BackupRequestMessage requestMsg) {
+    LOG.info("Received backup message: {}", requestMsg.toString());
+    Preconditions.checkState(!mBackupTracker.inProgress(), "Backup in progress");
+
+    // Cancel timeout timer.
+    mSuspendTimeoutFuture.cancel(true);
```
LuQQiu: Is it possible that this worker failed to receive the suspend message but receive the request message?
Should we add a preconditions.checkState or add a null check?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -303,4 +322,157 @@ public boolean isSnapshotting() {
...
+      Preconditions.checkState(mSuspended, "Not suspended");
+      Preconditions.checkState(mAdvanceFuture == null, "Advance future already set");
+      LOG.info("Advancing state machine to sequence: {}", sequence);
+      mAdvanceSequence = sequence;
+      // Complete the request if already at target sequence.
+      if (mLastAppliedSequence == mAdvanceSequence) {
+        return CompletableFuture.completedFuture(null);
+      }
+      // Create a future for completion.
+      mAdvanceFuture = new CompletableFuture<>();
```
LuQQiu: Looks like the advance behavior is synchronous, 
it will always be   
```
mAdvanceFuture = new CompletableFuture<>();
while loop to applyToMaster() -> mAdvanceFuture.complete(null);
return advanceFuture
```
Why do we need to use a completableFuture?

ggezer: It's synchronous only for the part that is in the buffer. If target sequence is not met, then future will not be completed until state machine receives it.

core/server/master/src/main/java/alluxio/master/backup/BackupLeaderRole.java
```diff
@@ -0,0 +1,293 @@
...
+    super.stop();
+  }
+
+  @Override
+  public Map<ServiceType, GrpcService> getRoleServices() {
+    Map<ServiceType, GrpcService> services = new HashMap<>();
+    services
+        .put(ServiceType.META_MASTER_BACKUP_MESSAGING_SERVICE,
+            new GrpcService(
+                ServerInterceptors.intercept(
```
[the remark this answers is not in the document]
ggezer: `ClientIpAddressInjector` is being injected to every call into`CopycatMessageServiceClientHandler` here. 

It's used mostly for traces. Otherwise It's very hard to say who's connecting to server in a gRPC bi-di stream.

core/server/master/src/main/java/alluxio/master/backup/BackupLeaderRole.java
```diff
@@ -0,0 +1,293 @@
...
+                    new CopycatMessageServiceClientHandler(
```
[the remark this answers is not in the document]
ggezer: I'll follow up with another PR for refactoring these classes for more generic use. Once this particular usage is locked then I can have better idea for refactoring.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1194,19 +1194,47 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_BIND_HOST =
-      new Builder(Name.MASTER_BIND_HOST)
-          .setDefaultValue("0.0.0.0")
-          .setDescription("The hostname that Alluxio master binds to.")
+  public static final PropertyKey MASTER_BACKUP_DELEGATION_ENABLED =
+      new Builder(Name.MASTER_BACKUP_DELEGATION_ENABLED)
+          .setDefaultValue(false)
```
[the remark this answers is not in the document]
ggezer: In non-HA cluster, we'll allow backups on leader due to below check.
` boolean delegateBackup =
      ServerConfiguration.getBoolean(PropertyKey.MASTER_BACKUP_DELEGATION_ENABLED)
            **&& ConfigurationUtils.isHaMode**(ServerConfiguration.global());`

Only in HA cluster, we won't allow blocking the leader by default.

That being said, we might still want to allow old way of backups in HA cluster. That's open to discussion.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
ggezer: Backup-leader don't send request before sending suspend message.

shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
```diff
@@ -41,6 +42,13 @@
           .desc("whether to write the backup to the master's local filesystem instead of the root"
               + " UFS")
           .build();
+  private static final Option ALLOW_LEADER_OPTION =
+      Option.builder()
+          .longOpt("allow-leader")
+          .required(false)
+          .hasArg(false)
+          .desc("whether to allow leader to take the backup when there are no secondaries.")
```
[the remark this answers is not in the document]
ggezer: @apc999?

shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
```diff
@@ -56,33 +64,64 @@ public String getCommandName() {
...
+    BackupStatus status = mMetaClient.backup(opts.build());
+    do {
+      // Backup could be after a fail-over.
+      if (status.getState() == BackupState.EMPTY) {
+        mPrintStream.printf("Backup lost. Check Alluxio logs.");
+        return -1;
+      }
+      if (status.getState() == BackupState.COMPLETED) {
+        break;
+      } else if (status.getState() == BackupState.FAULTED) {
```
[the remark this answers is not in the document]
ggezer: Proto doesn't allow it since`meta_master.proto` has another enum that defined `FAILED`. 
I can prefix all values with `Backup_`. Or just leave it as this if not that ugly.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1194,19 +1194,47 @@ public String toString() {
...
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_BACKUP_CONNECT_INTERVAL_MAX_MS =
+      new Builder(Name.MASTER_BACKUP_CONNECT_INTERVAL_MAX_MS)
+          .setDefaultValue("10sec")
+          .setDescription("Maximum delay between each connection attempt to leader backup master.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_BACKUP_ABANDON_TIMEOUT_MS =
```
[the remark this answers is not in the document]
ggezer: CONNECTION_TIMEOUT might not be a perfect fit as it's misleading in other ways. This property basically controls timing out of backup after not getting heart-beats for it.

ggezer:
@calvinjia @LuQQiu 
For reviewers' reference, this work will be followed by 3 PRs:

**1. Refactoring copycat transport:** Now that it's used by two separate features, it should be hosted by an appropriate component and refactored to generalize copycat specific names/traces etc...

**2. JournalStateMachine concurrency:** `JournalStateMachine` is a globally locked class and this PR doesn't aim to change that. It also added an additional contention when resuming journals. If locking during `resume()` takes a long time, then quorum could potentially be disrupted. I was not able to drive system to that point in manual tests but it's very much possible under heavy load.

**3. On-disk buffering while RAFT is suspended:** Currently entries are buffered on memory while journal is suspended. Buffer should be extended to disk.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
LuQQiu: Got it! Thanks

shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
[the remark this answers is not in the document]
LuQQiu: So strange, but simple is better, let's maintain FAULTED

core/server/master/src/main/java/alluxio/master/backup/BackupLeaderRole.java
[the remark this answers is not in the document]
LuQQiu: Got it, thanks

core/common/src/main/java/alluxio/conf/PropertyKey.java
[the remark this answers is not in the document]
LuQQiu: I try to understand how we backup in this PR:
(1) single master -> backup on local (allow_leader option is not needed)
(2) HA mode  -> delegation enabled -> backup on followers
                        -> delegation enabled -> but don't have follower connections -> check if backup allowed on leader -> Yes:backup on leader  No:error out
                        -> delegation not enabled   -> backup on leader

So the option allow_leader only affects HA cluster with delegation enabled but don't have any follower connections. 
I previously think allow-leader==false means we don't allow leader to take backup even in single master mode.  

```
  private static final Option ALLOW_LEADER_OPTION =
      Option.builder()
          .longOpt("allow-leader")
          .required(false)
          .hasArg(false)
          .desc("whether to allow leader to take the backup when there are no secondaries.")
          .build();
```
Can we change the description to let users know this option only affects high availability mode. In single master mode, backup is always taken in the master

core/common/src/main/java/alluxio/conf/PropertyKey.java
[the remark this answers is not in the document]
LuQQiu: Can you remove all the `_MS` from the property key? also for their names

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -217,11 +225,24 @@ public synchronized void start() throws IOException {
...
         "tailer thread must not be null in secondary mode");
-    mTailerThread.awaitTermination(true);
-    long nextSequenceNumber = mTailerThread.getNextSequenceNumber();
-    mTailerThread = null;
+
+    long nextSequenceNumber;
+    if (mSuspended) {
+      // Cancel if advancing is in progress.
+      if (mAdvanceFuture != null && !mAdvanceFuture.isDone()) {
+        // TODO(ggezer): Avoid re-applying sequences read while advancing.
```
LuQQiu: Did we stop backing up?

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+  }
+
+  /**
+   * Resumes the journals. Crashes the process if resume fails.
+   */
+  private void enforceResumeJournals() {
+    try {
+      mJournalSystem.resume();
+    } catch (IOException e) {
```
LuQQiu: catch throwable instead?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
ggezer: Yes, backup will be cancelled if backup-worker fail over to become a leader.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -215,6 +220,11 @@ private void handleRequestMessage(BackupRequestMessage requestMsg) {
           enforceResumeJournals();
         } else {
           LOG.info("Journal transition completed. Scheduling a backup.");
+          if (!mBackupTimeoutTask.cancel(true)) {
+            mBackupTracker
+                .updateError(new BackupException("Journal has ben resumed due to time-out."));
```
LuQQiu: When the advance future being cancelled, journal will be automatically resume?

ggezer: It should.
However it looks like CompletableFuture<?> does not support cancelling. So I'll need to change it in the signature to Future<?> and re-implement related parts.

ggezer:
@calvinjia @LuQQiu PTAL. Thanks!

ggezer:
Jenkins, test this please

LuQQiu:
@ggezer What's the major thing that you are fixing for delegation backup recently?  Because of the force push, I am not quite sure what's the major changes

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
ggezer: @LuQQiu  `CompletableFuture<?>` was not cancelleable. 

`JournalSystem#advance` now returns `AdvanceFuture` which allows for graceful termination of ongoing advancing.

core/server/common/src/main/java/alluxio/master/journal/AbstractAdvanceThread.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public abstract class AbstractAdvanceThread extends Thread {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractAdvanceThread.class);
+
+  @Override
+  public void run() {
+    try {
+      runAdvance();
+    } catch (Exception e) {
+      ProcessUtils.fatalError(LOG, e, "Advance thread is failed.");
```
LuQQiu: I try to list the workflow:
```
advancing
When BACKUP_ABORT_AFTER_TRANSITION_TIMEOUT_MS exceed
CompletableFuture will be cancel/interrupted
runAdvance will throw runtime exception
here will make it as a fatal error
system exit
```
Is it what's happening?

ggezer: `CompletableFuture` in `BackupWorkerRole#handleRequestMessage` is used to track `AdvanceFuture#waitTermimation`. Cancelling it won't interrupt advancing. 

Also `AdvanceFuture#cancel` itself won't interrupt thread but just mark for termination.

So timing out on advance won't trigger interrupt and will simply call `resume`.Which will cause advancing to finish prematurely.

I can implement `AdvanceFuture#waitTermination(long timeout)` variation to unclutter this piece of code but that itself will be complicated since `AdvanceFuture` can hold multiple `AdvanceFuture`s

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,293 @@
...
+  /** Journals managed by this applier. */
+  private final Map<String, RaftJournal> mJournals;
+  /** A supplier of journal sinks for this applier. */
+  private final Supplier<Set<JournalSink>> mJournalSinks;
+
+  /** The last sequence applied. */
+  private long mLastAppliedSequence = -1;
+
+  /** Whether this state machine is suspended. */
+  private boolean mSuspended = false;
```
calvinjia: Could we do with having an atomic boolean instead of using the locks?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,293 @@
...
+   * @return the future to track when applier reaches the target sequence
+   */
+  public AdvanceFuture advance(long sequence) {
+    try (LockResource stateLock = new LockResource(mStateLock)) {
+      Preconditions.checkState(mSuspended, "Not suspended");
+      Preconditions.checkState(!mResumeInProgress, "Resume in progress");
+      Preconditions.checkState(mAdvanceTask == null || !mAdvanceTask.isAlive(),
+          "Advance task in progress.");
+      Preconditions.checkState(sequence >= 0, "Invalid sequence");
+      Preconditions.checkState(mLastAppliedSequence <= sequence, "Can't advance to past.");
```
calvinjia: Print out the sequence numbers here?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,293 @@
...
+   * @param sequence target sequence
+   * @return the future to track when applier reaches the target sequence
+   */
+  public AdvanceFuture advance(long sequence) {
+    try (LockResource stateLock = new LockResource(mStateLock)) {
+      Preconditions.checkState(mSuspended, "Not suspended");
+      Preconditions.checkState(!mResumeInProgress, "Resume in progress");
+      Preconditions.checkState(mAdvanceTask == null || !mAdvanceTask.isAlive(),
+          "Advance task in progress.");
+      Preconditions.checkState(sequence >= 0, "Invalid sequence");
```
calvinjia: Print the invalid sequence number?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,293 @@
...
+  private long mLastAppliedSequence = -1;
+
+  /** Whether this state machine is suspended. */
+  private boolean mSuspended = false;
+
+  /** Whether a resume() call is in progress. */
+  private boolean mResumeInProgress = false;
+
+  /** Buffer to use during suspend period. */
+  private Queue<Journal.JournalEntry> mSuspendBuffer = new ConcurrentLinkedQueue();
```
calvinjia: Do we need to use the Concurrent object if we are managing the concurrency with locks? The size operation is more expensive.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -365,6 +369,40 @@ public synchronized void losePrimacy() {
...
+    // Suspend copycat snapshots.
+    mSnapshotAllowed.set(false);
+    mStateMachine.suspend();
+  }
+
+  @Override
+  public synchronized void resume() throws IOException {
+    mStateMachine.resume();
+    // Resume copycat snapshots.
+    mSnapshotAllowed.set(true);
```
calvinjia: Is it safe to allow snapshots?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -104,6 +106,15 @@
    */
   private UfsJournalCheckpointThread mTailerThread;
 
+  /** Whether the journal was suspended. */
```
calvinjia: `was` -> `is`?

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
```diff
@@ -0,0 +1,209 @@
...
+   * @param request the backup request
+   * @param entryCounter counter to receive written entry count
+   * @return URI of the backup
+   */
+  protected AlluxioURI takeBackup(BackupPRequest request, AtomicLong entryCounter)
+      throws IOException {
+    AlluxioURI backupUri;
+    // Acquire the UFS resource under which backup is being created.
+    try (CloseableResource<UnderFileSystem> ufsResource =
+        mUfsManager.getRoot().acquireUfsResource()) {
```
calvinjia: Are backups only written to the root ufs?

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,369 @@
...
+  private void handleSuspendJournalsMessage(BackupSuspendMessage suspendMsg) {
+    LOG.info("Received suspend message: {}", suspendMsg.toString());
+    Preconditions.checkState(!mBackupTracker.inProgress(), "Backup in progress");
+
+    try {
+      mJournalSystem.suspend();
+      LOG.info("Suspended journals for backup.");
+    } catch (IOException e) {
+      String failMessage = "Failed to suspended journals for backup.";
+      LOG.error(failMessage, e);
```
calvinjia: Do we need to log an error? We are already throwing an exception.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,369 @@
...
+              new BackupHeartbeatMessage(mBackupTracker.getCurrentStatus()));
+        } catch (Exception e) {
+          LOG.warn("Failed to send heartbeat to backup-leader: {}. Error: {}", mLeaderConnection,
+              e);
+        }
+        // Stop heartbeats if backup finished.
+        if (mBackupTracker.isFinished()) {
+          break;
+        }
+      } while (true);
```
calvinjia: Any reason not to use `do...while` instead of just `while`?

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
```diff
@@ -138,6 +139,40 @@
...
+  AdvanceFuture advance(Map<String, Long> journalSequences) throws IOException;
+
+  /**
+   * Used to get the current state from a leader journal system.
+   *
+   * Note: State changes to journals must have been suspended before calling this method.
+   *
+   * @return the current map of sequences for each master
+   */
+  Map<String, Long> getCurrentSequences();
```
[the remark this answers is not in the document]
ggezer: Yes. Do you have a suggestion for a better name?

core/server/common/src/main/java/alluxio/master/journal/AdvanceFuture.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  /**
+   * Creates a single future for tracking a list of advance futures.
+   *
+   * @param futures list of advance futures
+   * @return unified advance future
+   */
+  public static AdvanceFuture allOf(List<AdvanceFuture> futures) {
```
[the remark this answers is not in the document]
ggezer: I mimicked `CompletableFuture` in that. Static creator is to combine multiple instances together.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: `Resume` makes it complicated since it needs to actually lock.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: It's possible for `resume` and `processJournalEntry` to be accessing concurrently. This is until `resume` decides to take the lock in order to complete a long running catching up session.

ggezer: `AdvanceThread` also synchronizes with monitors only for getting notified when there are entries in the buffer, it doesn't lock while popping entries.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ggezer: The whole suspend/resume cycle is reserved for secondary role. So this is just restoring the original value after the backup.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
ggezer: Exceptions during message handling will show up only on the receiver side. This is to ease correlating.

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
ggezer: Nice catch, I think a fix for that regression was checked in after I started this.

ggezer: Oh, Looks like I still have that fix. In line 152 it overrides `ufs` resource with local factory if local backup is requested.

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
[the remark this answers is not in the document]
calvinjia: `getCurrentSequenceNumbers`?

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
calvinjia: Do we allow the case where it is a nonlocal, nonroot UFS?

ggezer: https://github.com/Alluxio/alluxio/issues/10066

core/common/src/main/java/alluxio/wire/BackupStatus.java
```diff
@@ -0,0 +1,225 @@
...
+import alluxio.grpc.BackupState;
+
+import com.google.common.base.MoreObjects;
+import com.google.protobuf.ByteString;
+import org.apache.commons.lang.SerializationUtils;
+
+/**
+ * Defines the status of a backup.
+ */
+public class BackupStatus {
```
calvinjia: Should this class be immutable?

core/common/src/main/java/alluxio/wire/BackupStatus.java
```diff
@@ -0,0 +1,225 @@
...
+   */
+  public boolean isCompleted() {
+    return mState == BackupState.COMPLETED;
+  }
+
+  /**
+   * @return {@code true} if the backup is failed
+   */
+  public boolean isFailed() {
+    return mState == BackupState.FAULTED;
```
calvinjia: Should the stated be `FAILED`?

core/server/common/src/main/java/alluxio/master/journal/AdvanceFuture.java
```diff
@@ -0,0 +1,82 @@
...
+package alluxio.master.journal;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Used to track and cancel journal advancing.
+ */
+public class AdvanceFuture {
```
calvinjia: Naming suggestion, does `catchup` make more sense here?

core/common/src/main/java/alluxio/wire/BackupStatus.java
ggezer: It's used as a wire type. I think it's OK to leave it mutable as with other types. Unless you have a particular reason?

core/server/common/src/main/java/alluxio/master/journal/AdvanceFuture.java
ggezer: For this class only or does you suggestion includes `JournalSystem.advance()` as well?

core/common/src/main/java/alluxio/wire/BackupStatus.java
ggezer: I could only make `Failed` as there is another enum in the same proto with value `FAILED`.

core/server/common/src/main/java/alluxio/master/journal/AdvanceFuture.java
calvinjia: Replace the concept of advance with catchup since we already have a concept of catchup.

core/common/src/main/java/alluxio/wire/BackupStatus.java
calvinjia: Oh I see, most projects get around this by naming the enums like `BackupState.BACKUP_STATE_FAILED`

core/common/src/main/java/alluxio/wire/BackupStatus.java
calvinjia: Oh I see, it would be nice to make our wire types immutable, but I'm fine with this since its the way we do it for our other ones.

core/common/src/main/java/alluxio/wire/BackupStatus.java
```diff
@@ -0,0 +1,225 @@
...
+
+import com.google.common.base.MoreObjects;
+import com.google.protobuf.ByteString;
+import org.apache.commons.lang.SerializationUtils;
+
+/**
+ * Defines the status of a backup.
+ */
+public class BackupStatus {
+  /** State of backup. */
```
calvinjia: Should there be an unique identifier for the backup?

core/common/src/main/java/alluxio/wire/BackupStatus.java
ggezer: I just didn't want to lengthen enum values. We can do that on the next 'Failed' value in the same file.

core/common/src/main/java/alluxio/wire/BackupStatus.java
ggezer: I was planning on getting unique IDs after uncovering some client side issue.
When run concurrently, backup shell commands could start picking up results for runs that they haven't launched.

ggezer: https://github.com/Alluxio/alluxio/pull/10182/commits/2ed4f85af8905341a0b972c38f62d49e921f3ce9

ggezer:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/AbstractAdvanceThread.java
ggezer: @LuQQiu Do you have any more concerns around this?

shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
```diff
@@ -97,12 +136,13 @@ public String getDescription() {
         + " directory to back up to can be overridden by specifying a directory here. The directory"
         + " path is relative to the root UFS. To write the backup to the local disk of the primary"
         + " master, use --local and specify a filesystem path. Backing up metadata"
-        + " requires a pause in master metadata changes, so use this command sparingly to"
-        + " avoid interfering with other users of the system.";
+        + " will be delegated to secondary masters in HA cluster. Use --allow-leader for"
```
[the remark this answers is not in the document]
ggezer: I think this should be stand-by master for consistency with other documents.

shell/src/main/java/alluxio/cli/fsadmin/command/BackupCommand.java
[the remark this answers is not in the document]
ggezer: I think this should be stand-by master for consistency with other documents.

ggezer:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    protected void runCatchup() {
+      // Spin for catching up until cancelled.
+      while (!mStopCatchingUp && mLastAppliedSequence < mCatchUpEndSequence) {
+        // Wait until notified for cancellation or more entries.
+        synchronized (mSuspendBuffer) {
+          while (!mStopCatchingUp && mSuspendBuffer.size() == 0) {
+            try {
+              mSuspendBuffer.wait();
```
LuQQiu: I am not sure how `ConcurrentLinkedQueue` work with `Synchronized`.
When running catchup, we 
```
        synchronized (mSuspendBuffer) {
          while (!mStopCatchingUp && mSuspendBuffer.size() == 0) {
            try {
              mSuspendBuffer.wait();
              throw new RuntimeException("Interrupted while catching up.");
            }
          }
```
In this situation, will other `mSuspendBuffer.notifyAll()` operations be blocked by not able to get the `synchronized (mSuspendBuffer)`?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,295 @@
...
+      mCatchupThread.cancel();
+      mCatchupThread.waitTermination();
+    }
+
+    /**
+     * Applies all buffered entries.
+     *
+     * It doesn't block state until
+     *   -> buffer contains few elements ( RESUME_LOCK_BUFFER_SIZE_WATERMARK )
+     *   -> was running for a long time  ( RESUME_LOCK_TIME_LIMIT_MS         )
```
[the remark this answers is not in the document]
ggezer: `processJournalEntry` could always cut in.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: Buffer could be accessed by `processJournalEntry`, `resume`, `catchup`.
`processJournalEntry` can be called concurrently with `resume` OR `catchup`. All three won't run in parallel since `resume` first waits until active `catchup` is done.

For when `processJournalEntry` and `catchup` is running concurrently. A monitor is required to get notified when the queue has an entry, so that `catchup` can continue. Note that resume locks only to check if there are entries, so the monitor won't get that hot. For the monitor I used the buffer itself, which will have no effect on concurrency of queue puts/gets.

For when, `processJournalEntry` and `resume` is running concurrently. We don't want to block anything until `resume` decides that it's time to block `processJournalEntry` and finish resuming.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+import java.time.Instant;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
+
+/**
+ * Used to control applying to masters.
```
calvinjia: Could you add some more info to the javadoc here (ie. we are applying journal entries and might need to buffer them if we are suspended)?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+  private boolean mResumeInProgress = false;
+
+  /** Buffer to use during suspend period. */
+  private Queue<Journal.JournalEntry> mSuspendBuffer = new ConcurrentLinkedQueue();
+
+  /** Used to store latest catch-up thread. */
+  private AbstractCatchupThread mCatchupThread;
+
+  /** Used to synchronize buffer state. */
+  private ReentrantLock mStateLock = new ReentrantLock(true);
```
calvinjia: Any reason for making this a fair lock?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+   * {@link #catchup(long)} is called.
+   *
+   * @throws IOException
+   */
+  public void suspend() throws IOException {
+    try (LockResource stateLock = new LockResource(mStateLock)) {
+      Preconditions.checkState(!mSuspended, "Already suspended");
+      LOG.info("Suspending state machine.");
+      mSuspended = true;
+      LOG.info("Suspended state machine at sequence: {}", mLastAppliedSequence);
```
calvinjia: The previous log message can be removed since there is trivial logic in between.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+   * Resumes the applier. This method will apply all buffered entries before returning.
+   *
+   * @throws IOException
+   */
+  public void resume() throws IOException {
+    try (LockResource stateLock = new LockResource(mStateLock)) {
+      Preconditions.checkState(mSuspended, "Not suspended");
+      Preconditions.checkState(!mResumeInProgress, "Resume in progress");
+      mResumeInProgress = true;
+      LOG.info("Resuming state machine.");
```
calvinjia: Log the last applied sequence number?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+    /**
+     * Applies all buffered entries.
+     *
+     * It doesn't block state until
+     *   -> buffer contains few elements ( RESUME_LOCK_BUFFER_SIZE_WATERMARK )
+     *   -> was running for a long time  ( RESUME_LOCK_TIME_LIMIT_MS         )
+     */
+    try {
+      // Mark resume start time.
+      Instant resumeStartTime = Instant.now();
```
calvinjia: Users have brought up performance issues with `Instant.now`, can we use `System.nanoTime`?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+      }
+
+      while (mSuspendBuffer.size() > 0) {
+        applyToMaster(mSuspendBuffer.remove());
+
+        // Check whether to lock the state now.
+        boolean lockSubmission = !mStateLock.isHeldByCurrentThread()
+            && (mSuspendBuffer.size() <= RESUME_LOCK_BUFFER_SIZE_WATERMARK
+                || Duration.between(Instant.now(), resumeStartTime)
+                    .compareTo(Duration.ofMillis(RESUME_LOCK_TIME_LIMIT_MS)) > 0);
```
calvinjia: Is there a downside to always grabbing the lock and flushing the entire queue? In the worst case scenario, this is what we revert to after waiting `RESUME_LOCK_TIME_LIMIT`

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -0,0 +1,298 @@
...
+  /** Resume will acquire exclusive lock when can't finish after this much time. */
+  private static final int RESUME_LOCK_TIME_LIMIT_MS = 30000;
+
+  /** Journals managed by this applier. */
+  private final Map<String, RaftJournal> mJournals;
+  /** A supplier of journal sinks for this applier. */
+  private final Supplier<Set<JournalSink>> mJournalSinks;
+
+  /** The last sequence applied. */
+  private long mLastAppliedSequence = -1;
```
calvinjia: Do we need to track one sequence number for each journal?

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,378 @@
...
+    mBackupTracker.update(new BackupStatus(requestMsg.getBackupId(), BackupState.Initiating));
+    mBackupTracker.updateHostname(NetworkAddressUtils.getLocalHostName(
+        (int) ServerConfiguration.global().getMs(PropertyKey.NETWORK_HOST_RESOLUTION_TIMEOUT_MS)));
+
+    // Start sending backup progress to leader.
+    startHeartbeatThread();
+
+    // Cancel timeout task created by suspend message handler.
+    if (!mBackupTimeoutTask.cancel(true)) {
+      mBackupTracker.updateError(new BackupException("Journal has ben resumed due to a time-out"));
```
calvinjia: `ben`->`been`

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,378 @@
...
+      }
+    });
+  }
+
+  /**
+   * Resumes the journals. Crashes the process if resume fails.
+   */
+  private void enforceResumeJournals() {
+    try {
+      mJournalSystem.resume();
```
calvinjia: Is there a timing issue where some journals could have been resumed and we cancel the task, or will that cause an exception to be thrown here?

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -0,0 +1,378 @@
...
+            "Initiating catching up of journals to consistent sequences before starting backup. {}",
+            requestMsg.getJournalSequences());
+        CatchupFuture catchupFuture = mJournalSystem.catchup(requestMsg.getJournalSequences());
+        CompletableFuture.runAsync(() -> catchupFuture.waitTermination())
+            .get(BACKUP_ABORT_AFTER_TRANSITION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      } catch (Exception e) {
+        mBackupTracker.updateError(new BackupException(
+            String.format("Failed to catch-up journals: %s", e.getMessage()), e));
+        // Don't resume journals if interrupted.
+        // Catch-up future can be interrupted only when master becomes primary during backup.
```
calvinjia: Is this true, don't we cancel the future if we lose connection to the master?

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterClient.java
```diff
@@ -12,30 +12,48 @@
...
+   *       be called for querying the status of the on-going backup.
+   *
+   * Note: When leader has no secondary in an HA cluster, it will reject backup. This could
+   *       be allowed by passing "AllowLeader" option in the request.
+   *
+   * @param backupRequest the backup request
+   * @return status of backup
+   * @throws IOException
+   */
+  BackupStatus backup(BackupPRequest backupRequest) throws IOException;
```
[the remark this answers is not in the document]
ggezer: Currently, backup will return complete BackupStatus unless you specify -async flag in the request. Are you suggesting that it returns UUID even when it was blocking?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: A guard against aggressive `processJournalEntry` calls by copycat.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: This is to not prevent `processJournalEntry` if buffer was overloaded during a long backup. Blocking it for a long time would render copycat server unreachable.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: This is used by `JournalStateMachine` which manages all journals with a single sequence. So not required at here as well.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
ggezer: Nice catch!

This statement is not correct:

First; the future that is being interrupted is not the `catchup-future` but the async task that is spawned to wait for `catchup-future`. (Waiting on catchup-future is not interruptable)
Second; we interrupt this code when connection to master is lost.

The reason, I didn't enforce resuming when interrupted was to quit from stopMasters() path ASAP during gain-primacy scenario. However the journal itself will still call it when gaining primacy, so being quick here does not really have a value. So removing this code completely.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
ggezer: There is no such window AFAICT.

Double resuming should be avoided by this line below and the fact that, gaining-primacy first stops this role...

`// Cancel timeout task created by suspend message handler.
    if (!mBackupTimeoutTask.cancel(true)) {
      mBackupTracker.updateError(new BackupException("Journal has been resumed due to a time-out"));
      return;
    }`

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterClient.java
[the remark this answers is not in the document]
calvinjia: Oh I see, that makes sense if its synchronous. Does this mean if there is an error, an exception will not be thrown but a backup status with an error will be returned?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
calvinjia: Could we still be in that situation after `RESUME_LOCK_TIME_LIMIT` is up?

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
calvinjia: `mBackupTimeoutTask` can return true if it interrupts `mJournalSystem.resume` while it is in progress right?

ggezer: Reverted try-catch condition in `enforceResumeJournal` to original.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: Right. Hitting a time limit is when the code stops being hopeful. Hitting a time limit is extremely unlikely because resuming is done on a thread, where as new entries will come in as part of copycat hearbeats.

ggezer:

PR #10187, opened 2019-10-22
Update README to include alluxio-shaded-client artifact

apc999:


README.md
```diff
@@ -82,29 +82,33 @@ to run a simple example with Alluxio.
...
-```
-libraryDependencies += "org.alluxio" % "alluxio-core-client-fs" % "2.0.1"
-```
+Alluxio publishes several client artifacts. Include the appropriate artifact in your project.
+
++ Artifact `alluxio-core-client-fs` provides a client for [Alluxio file system API](https://docs.alluxio.io/os/user/stable/en/api/FS-API.html#alluxio-java-api).  This is recommended for the best performance and access to all Alluxio functionality in a Java project.
++ Artifact `alluxio-core-client-hdfs` provides a client for [HDFS-Compatible file system API](https://docs.alluxio.io/os/user/stable/en/api/FS-API.html#hadoop-compatible-java-client).
++ Artifact `alluxio-shaded-client`, available in Alluxio 2.0.1 or later, includes both above
+  clients and all their dependencies but in a shaded form to prevent conflicts. This client jar is
+  self-contained but its size is also much bigger than the other two.
```
calvinjia: Could you add what this client is recommended to be used for?

apc999: done

apc999:
@calvinjia  PTAL

PR #10197, opened 2019-10-22
Add Job Config Description and Worker Hostname Information to job cli

bradyoo:


bradyoo:
@calvinjia: Added a comment for where I may want feedback from.

core/transport/src/proto/journal/job.proto
```diff
@@ -13,11 +13,12 @@ enum Status {
   COMPLETED = 5;
 }
 
-// next available id: 6
+// next available id: 8
```
calvinjia: Should this be 7?

job/common/src/main/java/alluxio/job/JobConfig.java
```diff
@@ -14,16 +14,31 @@
...
-  String getName();
+  public abstract String getName();
+
+  /**
+   * @return the description of the job
+   */
+  @Nullable
+  public String getDescription() {
+    String result = toString();
+    if (result.length() > MAX_DESCRIPTION_LENGTH) {
```
calvinjia: Any reason for limiting this here?

job/common/src/main/java/alluxio/job/meta/JobInfo.java
```diff
@@ -77,11 +79,11 @@ public int compareTo(JobInfo other) {
...
    *
    * @param taskId the task id
+   * @param workerInfo the workerInfo
    */
-  public void addTask(int taskId) {
+  public void addTask(int taskId, WorkerInfo workerInfo) {
     TaskInfo oldValue = mTaskIdToInfo.putIfAbsent(taskId,
-        new TaskInfo().setJobId(mId).setTaskId(taskId).setStatus(Status.CREATED).setErrorMessage("")
-            .setResult(null));
+        new TaskInfo(mId, taskId, workerInfo));
```
calvinjia: Can this be on one line now?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -136,6 +138,11 @@ public synchronized void executeTask(long jobId, int taskId, JobConfig jobConfig
     taskInfo.setJobId(jobId);
     taskInfo.setTaskId(taskId);
     taskInfo.setStatus(Status.RUNNING);
+    try {
+      taskInfo.setWorkerHost(InetAddress.getLocalHost().getHostName());
+    } catch (UnknownHostException e) {
+      // ignore
```
calvinjia: Why can we ignore this?

job/common/src/main/java/alluxio/job/JobConfig.java
```diff
@@ -14,16 +14,31 @@
...
+import javax.annotation.Nullable;
 import java.io.Serializable;
 
 /**
  * A job configuration. All the subclasses are both Java and JSON serializable.
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@type")
-public interface JobConfig extends Serializable {
+public abstract class JobConfig implements Serializable {
```
calvinjia: I would prefer this to be an interface if possible. I'm not convinced we need to limit the size of `getDescription` here, unless the bottleneck we are trying to avoid is too large of payload?

bradyoo: I took a look at when @gpang was debugging when he decided to print out this and it was in the 100,000s of characters but I can limit from the client instead.

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
bradyoo: This is only to stamp the worker host information so the worst case of this is to not have hostname information. But maybe, what I should do instead is to stamp them again from the master.

job/common/src/main/java/alluxio/job/JobConfig.java
calvinjia: Yeah if data transfer / serialization is not the concern and the concern is readability we should limit on the client side.

bradyoo:
oops I previously forgot to code that actually addresses the feedback.

bradyoo:
Forgot to write that feedback has been addressed.

bradyoo:
Failed tests: 
alluxio.collections.LockPoolTest.parallelInsertTest(alluxio.collections.LockPoolTest)
  Run 1: LockPoolTest.parallelInsertTest:117 expected:<16> but was:<15>
  Run 2: LockPoolTest.parallelInsertTest:117 expected:<16> but was:<15>

bradyoo:
jenkins, test this please.

bradyoo:
jenkins, test this please

bradyoo:
@calvinjia: ping?

bradyoo:
@calvinjia: Feedback addressed. Btw, I plan to delete that file you feedbacked on soon since we don't have journaling for job service.

bradyoo:
rebased from upstream/master

bradyoo:
  Run 1: FreeAndDeleteIntegrationTest.deleteDir:145 » TestTimedOut test timed out after...
  Run 2: FreeAndDeleteIntegrationTest.deleteDir:145 » TestTimedOut test timed out after...

bradyoo:
jenkins, test this please.

PR #10206, opened 2019-10-23
Prevent data loss when writing with ASYNC_THROUGH

ZacBlanco:
This PR aims to fix data loss issues that can occur when writing files
with the ASYNC_THROUGH WriteType. Before this change, data loss
may have occured in the  following situations:

**Case 1:**

- A user writes a file with `alluxio.user.file.persistence.initial.wait.time`
set to -1. This is typical when a user is attempting to use the "persist
on rename" feature.
- Any amount of data is written to the worker, so long as it exceeds
the high eviction watermark
- The file is closed, but not persisted.
- The completeFile operation is supposed to schedule the persist
job and change the inode state to `TO_BE_PERSISTED`, but because
the wait time is set as -1, this never happens. The file is then marked
as completed.
- SpaceReserver eviction occurs, but because the file isn't marked as
TO_BE_PERSISTED, the blocks of that file are available as eviction
candidates
- The blocks of the non-persisted file are evicted, and data is lost.

**Case 2:**

This case is more general as it only requires the write type to be
ASYNC_THROUGH, but is less likely to occur due to heartbeat thread
timing.

- A user writes a file with ASYNC_THROUGH, but doesn't yet close the
file. The worker containing all of the blocks is above the high eviction
watermark. Any blocks written up until this point are locked by the
local worker because they aren't committed until the file is closed.
- The user closes the file and blocks begin to get committed.
Subsequently, blocks are removed from the workers locked list and
added to the pin list in TieredBlockStore. While this is occurring, the
PinListSync occurs - this will wipe any blocks added from the
committed block list. By default PinListSync occurs every 1s.
- At this point the SpaceReserver thread is triggered (while the
file is still being closed and blocks being comitted), and then evict
any of the committed blocks which were wiped from the pin list
when the PinListSync previously occurred.

This PR aims to address both of these cases.

First, we introduce a new parameter on the CreateFileContext which
checks if we're writing with ASYNC_THROUGH. Upon creating will add
the inode immediately when PinListSync occurs, even if the file hasn't
been completed.

Second, We remove the async PinListSync and attach the RPC directly to
the SpaceReserver heartbeat. This ensures that when SpaceReserver runs
that it has the most up-to-date list of pinned inodes.

Additionally, I've added three integration tests for writing files with
ASYNC_THROUGH that previously failed for one or both of the cases
described above, that now all pass.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/contexts/CreateFileContext.java
+++ b/core/server/master/src/main/java/alluxio/master/file/contexts/CreateFileContext.java
@@ -11,8 +11,10 @@
 
 package alluxio.master.file.contexts;
 
+import alluxio.client.WriteType;
 import alluxio.conf.ServerConfiguration;
 import alluxio.grpc.CreateFilePOptions;
+import alluxio.grpc.WritePType;
 import alluxio.util.FileSystemOptions;
 
 import com.google.common.base.MoreObjects;
@@ -24,6 +26,7 @@ public class CreateFileContext
     extends CreatePathContext<CreateFilePOptions.Builder, CreateFileContext> {
 
   private boolean mCacheable;
+  private boolean mIsAsync;
 
   /**
    * Creates context with given option data.
@@ -33,6 +36,7 @@ public class CreateFileContext
   private CreateFileContext(CreateFilePOptions.Builder optionsBuilder) {
     super(optionsBuilder);
     mCacheable = false;
+    mIsAsync = optionsBuilder.getWriteType() == WritePType.ASYNC_THROUGH;
   }
 
   /**
@@ -69,6 +73,22 @@ public class CreateFileContext
   }
 
   /**
+   * @return whether this is a file created with {@link WriteType#ASYNC_THROUGH}
+   */
+  public boolean isAsync() {
+    return mIsAsync;
+  }
+
+  /**
+   * @param isAsync whether or not we are creating an async file
+   * @return the updated context
+   */
+  public CreateFileContext setIsAsync(boolean isAsync) {
+    mIsAsync = isAsync;
+    return getThis();
+  }
+
+  /**
    * @return true if file is cacheable
    */
   public boolean isCacheable() {
@@ -89,6 +109,7 @@ public class CreateFileContext
     return MoreObjects.toStringHelper(this)
         .add("PathContext", super.toString())
         .add("Cacheable", mCacheable)
+        .add("isAsync", mIsAsync)
         .toString();
   }
 }
--- a/core/server/master/src/main/java/alluxio/master/file/contexts/CreateFileContext.java
+++ b/core/server/master/src/main/java/alluxio/master/file/contexts/CreateFileContext.java
@@ -11,10 +11,8 @@
 
 package alluxio.master.file.contexts;
 
-import alluxio.client.WriteType;
 import alluxio.conf.ServerConfiguration;
 import alluxio.grpc.CreateFilePOptions;
-import alluxio.grpc.WritePType;
 import alluxio.util.FileSystemOptions;
 
 import com.google.common.base.MoreObjects;
@@ -26,7 +24,6 @@ public class CreateFileContext
     extends CreatePathContext<CreateFilePOptions.Builder, CreateFileContext> {
 
   private boolean mCacheable;
-  private boolean mIsAsync;
 
   /**
    * Creates context with given option data.
@@ -36,7 +33,6 @@ public class CreateFileContext
   private CreateFileContext(CreateFilePOptions.Builder optionsBuilder) {
     super(optionsBuilder);
     mCacheable = false;
-    mIsAsync = optionsBuilder.getWriteType() == WritePType.ASYNC_THROUGH;
   }
 
   /**
@@ -73,22 +69,6 @@ public class CreateFileContext
   }
 
   /**
-   * @return whether this is a file created with {@link WriteType#ASYNC_THROUGH}
-   */
-  public boolean isAsync() {
-    return mIsAsync;
-  }
-
-  /**
-   * @param isAsync whether or not we are creating an async file
-   * @return the updated context
-   */
-  public CreateFileContext setIsAsync(boolean isAsync) {
-    mIsAsync = isAsync;
-    return getThis();
-  }
-
-  /**
    * @return true if file is cacheable
    */
   public boolean isCacheable() {
@@ -109,7 +89,6 @@ public class CreateFileContext
     return MoreObjects.toStringHelper(this)
         .add("PathContext", super.toString())
         .add("Cacheable", mCacheable)
-        .add("isAsync", mIsAsync)
         .toString();
   }
 }
--- a/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
@@ -226,14 +226,14 @@ public final class DefaultBlockWorker extends AbstractWorker implements BlockWor
             (int) ServerConfiguration.getMs(PropertyKey.WORKER_TIERED_STORE_RESERVER_INTERVAL_MS),
             ServerConfiguration.global(), ServerUserState.global()));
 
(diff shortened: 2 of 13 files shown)


ZacBlanco:
@rastogiasr FYI 

@madanadit @gpang @calvinjia My attempt for this PR involves just setting `TO_BE_PERSISTED` immediately on the inodes instead of waiting for a persist to be scheduled. Previously, `TO_BE_PERSISTED` wasn't set until the persist was specifically scheduled. I'm not sure what the ramifications of setting this before `completeFile` being called are.

One other solution I considered was adding a new state like `TO_BE_SCHEDULED` or something of the sort to denote that the persist hasn't been scheduled, but should be at some point. Inodes with that state would be treated similar to the ones with `TO_BE_PERSISTED` (i.e. they are returned when calling `getPinIdList()` on FSMaster).

Let me know what your thoughts are. I expect I might need to iterate once or twice. I haven't dug around much in this code.

core/server/master/src/main/java/alluxio/master/file/contexts/CreateFileContext.java
```diff
@@ -24,6 +26,7 @@
     extends CreatePathContext<CreateFilePOptions.Builder, CreateFileContext> {
 
   private boolean mCacheable;
+  private boolean mIsAsync;
```
calvinjia: Instead of having an `isAsync` boolean, should we store the write type? This could be done in the Path context and also replace the other derived value (`isPersisted`)

core/server/worker/src/main/java/alluxio/worker/block/SpaceReserver.java
```diff
@@ -130,6 +130,8 @@ private synchronized void reserveSpace() {
 
   @Override
   public void heartbeat() {
+    // Sync the pin list before attempting to reserve space
+    mBlockWorker.syncPinList();
```
calvinjia: This might be too expensive to call on every eviction attempt. We want to keep the eviction loop fast and without dependencies on external services.

ZacBlanco: I think there is a trade-off here for speed vs correctness.

Additionally, the PinListSync is configured by default to run every `1s` and the SpaceReserver is also configured to run every `1s`. While the two could theoretically run in parallel, I don't see an advantage to having them be independent of one another if they have the same interval configured by default.

Do you see a case where you would want these values to be different?

calvinjia: The pin list being stale is still a problem even if we call it here, since the time to get the pin list can be non trivial.
An advantage to having the pin list sync run separately from the eviction process is when the pin list sync takes a non trivial amount of time to execute (ie. network, master slow). In this case eviction can continue using the stale pin list.

A stale pin list is a correctness issue which we address with the `pinOnCreate` flag (https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java#L305).

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -226,14 +226,14 @@ public void start(WorkerNetAddress address) throws IOException {
             (int) ServerConfiguration.getMs(PropertyKey.WORKER_TIERED_STORE_RESERVER_INTERVAL_MS),
             ServerConfiguration.global(), ServerUserState.global()));
 
+    // Start the pinlist syncer to perform the periodical fetching
```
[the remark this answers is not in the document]
ZacBlanco: It was a mistake

ZacBlanco:
@gpang @calvinjia I've addressed comments, PTAL

core/server/master/src/main/java/alluxio/master/file/contexts/CreateFileContext.java
ZacBlanco: Added write type on `CreatePathContext`

calvinjia:
@ZacBlanco did you want to keep the pin update in the same thread as eviction?

ZacBlanco:
I decided to leave them separate for now so that we can keep the eviction loop fast. For a system running at scale, the pin list sync could take a long amount of time depending on the number of inodes that are pinned/to be persisted. I would prefer clients to be able to evict the blocks using the stale list than wait a longer period of time to get the newest list.

pinned = ✅, not pinned = ❌ 
 
| worker | master | behavior |
|-|-|-|
| ✅ | ✅ | up to date pin list - blocks won't be evicted |
| ✅ | ❌ | stale pin list - ok, just limited on blocks to evict |
| ❌ | ✅ | blocks in this case shouldn't be evicted due to pin on create, but there still exists a small gap between commit blocks/complete file that pinListSync can occur. However, because we set TO_BE_PERSISTED early now, if the pinlist sync occurs at this time, it will contain the inode Id corresponding to the blocks which may have just been written with ASYNC_THROUGH. So we shouldn't lose the data
| ❌ | ❌ | up to date pin list  - blocks can be evicted |

I think we should revisit this pin list implementation because clusters with a large namespace or that writes with `ASYNC_PERSIST` often, will execute  1 RPC every second * number of workers. And the data transferred over the network could be large depending on how many files there are.

ZacBlanco:
Errored

```
Tests in error: 
  Run 2: MultiMasterJournalTest.testTtl:81 » IO java.lang.RuntimeException: alluxio.exc...
```

ZacBlanco:
jenkins, test this please

ZacBlanco:
Similar tests failing again:

```
Tests in error: 
  FileSystemUtilsIntegrationTest.beforeClass:65 » Runtime Timed out waiting for ...
  Run 1: FreeAndDeleteIntegrationTest.deleteDir:146 » TestTimedOut test timed out after...
  Run 2: FreeAndDeleteIntegrationTest.deleteDir:146 » TestTimedOut test timed out after...

  Run 2: MultiMasterJournalTest.testTtl:81 » IO java.lang.RuntimeException: alluxio.exc...

```

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
more flakes

```
Tests in error: 
alluxio.client.cli.JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal(alluxio.client.cli.JournalToolTest)
  Run 1: JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal:165->checkpointEmbeddedJournal:212->lambda$checkpointEmbeddedJournal$0:226 Runtime
  Run 2: JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal:165->checkpointEmbeddedJournal:212->lambda$checkpointEmbeddedJournal$0:226 Runtime

alluxio.client.cli.fsadmin.command.QuorumCommandIntegrationTest.quorumRemove(alluxio.client.cli.fsadmin.command.QuorumCommandIntegrationTest)
  Run 1: QuorumCommandIntegrationTest.quorumRemove:166 » TestTimedOut test timed out af...
  Run 2: QuorumCommandIntegrationTest.quorumRemove:135 » FileAlreadyExists /testDir alr...

Flaked tests: 
  Run 1: JobServiceFaultToleranceShellTest.before:52 » IO java.lang.RuntimeException: a...
  Run 2: PASS

alluxio.client.cli.fs.command.CpCommandIntegrationTest.copyFromLocalLarge(alluxio.client.cli.fs.command.CpCommandIntegrationTest)
  Run 1: CpCommandIntegrationTest.copyFromLocalLarge:418 » Internal http2 exception (Gr...
  Run 2: PASS

alluxio.client.cli.fs.command.UpdateMountCommandTest.updateMountWithWrongArgs(alluxio.client.cli.fs.command.UpdateMountCommandTest)
  Run 1: UpdateMountCommandTest>AbstractFileSystemShellTest.before:123 » Runtime Timed ...
  Run 2: UpdateMountCommandTest>AbstractFileSystemShellTest.after:132 » NullPointer
  Run 3: PASS
  Run 4: PASS

  Run 2: PASS

```

ZacBlanco:
jenkins, test this please

ZacBlanco:

ZacBlanco:

PR #10218, opened 2019-10-23
Improve and simplify attach cli command

gpang:


gpang:
failed: `alluxio.server.ft.journal.MultiMasterJournalTest.testTtl`

gpang:
Jenkins, test this please

table/shell/src/main/java/alluxio/cli/table/command/AttachDatabaseCommand.java
```diff
@@ -39,6 +39,14 @@
...
       .desc("options associated with this UDB")
       .build();
+  private static final Option DB_OPTION = Option.builder()
+      .longOpt("db")
+      .required(false)
+      .hasArg(true)
+      .numberOfArgs(1)
+      .argName("alluxio db name")
+      .desc("The name of the db in Alluxio. If unset, will use the udb db name.")
+      .build();
```
calvinjia: Should we use options to parse the uri and dbname? We can make `required` true?

gpang: The scheme I was following was that I tried to make required arguments as actual arguments to the command, and anything optional was a `--option`. So, all attach commands would require the type, uri, and udb name, and optionally we can use a different alluxio name.

calvinjia: I was thinking like attachDb --dbname=foo, but since its a CLI I'm fine with implicit parameters as well. Up to you.

gpang:

gpang:

PR #10222, opened 2019-10-23
Support selective partition transformation

chcn:
If new partitions are synced to a table, and the table is transformed with the same definition, then only the newly added partitions will be transformed.


core/transport/src/grpc/table/table_master.proto
```diff
@@ -61,9 +61,15 @@ message Layout {
...
 
+message Transformation {
+    optional Layout layout = 1;
+    optional string definition = 2;
+}
+
 message Partition {
     optional PartitionSpec partition_spec = 1;
     optional Layout layout = 2;
+    optional Transformation transformation = 3;
```
gpang: could this be?
```suggestion
    repeated Transformation transformations = 3;
```

table/server/master/src/main/java/alluxio/master/table/Partition.java
```diff
@@ -59,17 +114,29 @@ public Layout getBaseLayout() {
...
+  }
+
+  /**
+   * @param definition the transformation definition
+   * @return whether the latest transformation of Partition has the same definition
+   */
+  public boolean isTransformed(TransformDefinition definition) {
+    Transformation transformation = mTransformation.get();
+    return transformation != null
+        && transformation.getDefinition().equals(definition.getDefinition());
```
gpang: should we normalize the definition a little bit for comparing (like case-insensitive, whitespace, etc)?

chcn:
@gpang PTAL, thanks.

table/server/master/src/main/java/alluxio/master/table/Partition.java
```diff
@@ -20,14 +20,72 @@
...
 
 /**
  * The table partition class.
  */
 public class Partition {
   private final String mPartitionSpec;
   private final Layout mBaseLayout;
-  private Layout mTransformedLayout;
+  /** New transformation is pushed to the end of the queue. */
+  private final ConcurrentLinkedDeque<Transformation> mTransformations =
```
gpang: o, for now, we can just keep a single transformation. I just thought to make the RPC proto repeated in-case we want to add to it later, without having to change the rpc signature.

chcn:
@gpang PTAL

gpang:
@chcn Please resolve the merge conflicts

gpang:

gpang:

PR #10235, opened 2019-10-23
Add Longest Running Jobs into job service report

bradyoo:


calvinjia:
How big would the entire job info set be? Could we just send the entire thing over and do filtering on the client side? This would make it much easier to implement other filters or if the client wanted the last N jobs instead of just 10.

bradyoo:
The bound is rather large here. It's up to 10000 jobs each with up to
(currently) N tasks where currently N is the number of workers (but that's
not going to be the case very soon). Once we download all the parameter
information for all of these, it could theoretically be 100s of MBs.

calvinjia:
Would it be okay if we don't send the task infos?

bradyoo:
toString of Jobconfig is also very large because that may contain all of
the parameters of the tasks.

bradyoo:
What I may do is make an alternative JobInfo object for Summary that drops
all of the large datas. I may need to do this anyways because the job's
ListCommand has the same problem.

bradyoo:
@calvinjia: The biggest reason I want this is for debugging never ending jobs. I could keep cherry-picking this into various builds but I don't know if we want to keep doing that.

calvinjia:

PR #10237, opened 2019-10-24
Fix Alluxio helm chart deployment issue

cheyang:
- Move tiered store configuration to ALLUXIO_JAVA_OPTS
- Fix the user of alluxio worker


integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -32,13 +32,13 @@ metadata:
...
...nt}}{{ $fullName }}-master-{{$i}}:19200,{{end}}{{end}} {{ .Values.jvmOptions }} {{- range $key, $val := .Values.properties }} -D{{ $key }}={{ $val }} {{- end}} {{- if .Values.tieredstore }} -Dalluxio.worker.tieredstore.levels={{ len .Values.tieredstore.levels }} {{- range .Values.tieredstore.levels }} {{- if .alias }} -Dalluxio.worker.tieredstore.level{{ .level }}.alias={{ .alias }} {{- if .mediumtype }} -Dalluxio.worker.tieredstore.level{{ .level }}.dirs.mediumtype={{ .mediumtype }} {{- else}} -Dalluxio.worker.tieredstore.level{{ .level }}.dirs.mediumtype={{ .alias }} {{- end}} {{- end}} {{- if .path }} -Dalluxio.worker.tieredstore.level{{ .level }}.dirs.path={{ .path }} {{- end}} {{- if .quota }} -Dalluxio.worker.tieredstore.level{{ .level }}.dirs.quota={{ .quota }} {{- end}} {{- if .high }} -Dalluxio.worker.tieredstore.level{{ .level }}.watermark.high.ratio={{ .high }} {{- end}} {{- if .low }} -Dalluxio.worker.tieredstore.level{{ .level }}.watermark.low.ratio={{ .low }} {{- end}} {{- end}} {{- end}} {{- if eq .Values.worker.shortCircuitPolicy "uuid" }} -Dalluxio.worker.data.server.domain.socket.address=/opt/domain -Dalluxio.worker.data.server.domain.socket.as.uuid=true {{- end}}
```
madanadit: i don't think this belongs to common opts. why did you have to move this?

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -196,6 +196,9 @@ fuse:
   jvmOptions: " -XX:MaxDirectMemorySize=2g "
   hostNetwork: true
   dnsPolicy: ClusterFirstWithHostNet
+  user: 0
+  group: 0
+  fsGroup: 0
```
jiacheliu3: @madanadit I don't know much about fuse but does it only work with root access?

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -120,14 +123,14 @@ spec:
...
                     {{- else}}
             - mountPath: {{ .path }}
               name: {{ .alias | lower }}
                     {{- end}}
-                  {{- end}} 
+                  {{- end}}
                 {{- end}} 
-              {{- end}}
+              {{- end}} 
+            {{- end}}
```
jiacheliu3: Is there one extra `{{- end}}`? What does this match to?

integration/kubernetes/helm-chart/alluxio/values.yaml
madanadit: yes, fuse container need as root with sys_admin privileges. it actually modifies the mount table on the physical host (highly privileged)

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
cheyang: This matches {{- if .mediumtype }}. I think the number of {{- end}} is the same.

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -44,3 +44,4 @@ data:
   ALLUXIO_FUSE_JAVA_OPTS: |-
     -Dalluxio.user.hostname=${ALLUXIO_CLIENT_HOSTNAME} -Dalluxio.worker.hostname=${ALLUXIO_CLIENT_HOSTNAME} {{ .Values.fuse.jvmOptions }} {{- range $key, $val := .Values.fuse.properties }} -D{{ $key }}={{ $val }} {{- end}} 
   ALLUXIO_WORKER_TIEREDSTORE_LEVEL0_DIRS_PATH: /dev/shm
+
```
madanadit: fix newline

integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
```diff
@@ -26,3 +26,9 @@
 - Added multiple template directories for sample use cases
 - Merged helm-generate.sh which generates the templates 
 
+
+0.5.1
```
madanadit: please update https://github.com/Alluxio/alluxio/blob/master/docs/_config.yml#L21 as well. thx

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
cheyang: Thank you for reminding this, I think it may be caused the worker/daemonset.yaml also has ALLUXIO_JAVA_WORKER_OPTS. I will test.

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
cheyang: Done

madanadit:
Please also run `./integration/kubernetes/helm-generate.sh all` once the helm changes are done to update the pre-generated sample specifications. We are in the process of automating this step.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -215,6 +215,56 @@ secrets:
...
+***Examples: Memory Tiered Store Only***
+
+```properties
+tieredstore:
+  levels:
+  - alias: MEM
+    mediumtype: MEM
+    level: 0
+    path: /dev/shm
+    type: hostPath
```
madanadit: please mention that the 2 supported volume types right now are [hostPath, emptyDir]

jiacheliu3:
We use `./integration/kubernetes/helm-generate.sh all` to generate all the templates. You shouldn't need to manually edit any yaml templates in directories like `singleMaster-localJournal`. 

Also running the script will make sure the templates in all those directories keep aligned with the helm chart parameters and values.

integration/kubernetes/alluxio-fuse-client.yaml.template
```diff
@@ -17,24 +17,24 @@ metadata:
   name: alluxio-fuse-client
   labels:
     app: alluxio
-    chart: alluxio-0.5.0
-    release: release-name
+    chart: alluxio-0.5.1
+    release: RELEASE-NAME
```
jiacheliu3: Do we need to change the release name? I don't have a preference between lower or upper case though.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -265,6 +265,8 @@ tieredstore:
     low: 0.7
 ```
 
+> There 2 supported volume `type`: `hostPath` and `emptyDir`. `hostPath` mounts the directory from the host node’s filesystem into your Pod. The benifits are that data can be kept even the worker restarted. `emptyDir` is first created when a Pod is assigned to a Node, and exists as long as that Pod is running on that node. As the name says, it is initially empty. The benifit of empty dir is that it can set memory limit in Kubernetes level.
```
madanadit: "benifits" -> "benefits", "benifit -> "benefit"

madanadit:
@cheyang have both tieredstore and fuse been tested now? do you plan on adding docs for fuse in this PR as well? thank you!

integration/kubernetes/alluxio-fuse-client.yaml.template
madanadit: @cheyang Seems like helm might choose a different string based on version. So I've explicitly defined the release name in https://github.com/Alluxio/alluxio/pull/10245. Please rebase and run generate again. Thx.

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -84,8 +87,6 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: status.hostIP
-          - name: ALLUXIO_WORKER_JAVA_OPTS
-            value: {{ print " -Dalluxio.worker.hostname=$(ALLUXIO_WORKER_HOSTNAME) " | quote }}
```
jiacheliu3: @cheyang I don't think you should remove this `ALLUXIO_WORKER_JAVA_OPTS`. This handles `alluxio.worker.hostname` property, which can only be decided at runtime. Please restore these 2 lines.

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -182,16 +182,16 @@ spec:
...
                     {{- if eq .type "hostPath"}}
         - hostPath:
             path: {{ .path }}
             type: DirectoryOrCreate
           name: {{ .alias | lower }}
                     {{- end }}
-                  {{- end }} 
-                {{- end }} 
+                  {{- end }}
+                {{- end}}
```
jiacheliu3: `{{- end}}` -> `{{- end }}`

jiacheliu3: `{{- end}}` -> `{{- end }}`

integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -128,16 +128,16 @@ spec:
...
               mountPath: {{ .path }}
-                      {{- else }}
+                        {{- else }}
             # report error and exit
             {{ .name }} with {{ .type }} is not supported in shortCircuitPolicy local
-                      {{- end }} 
-                    {{- end }} 
+                        {{- end }} 
+                      {{- end }}
+                    {{- end}}
```
jiacheliu3: I think we can better reduce the number of nested levels by using `{{- with .Values.tieredstore.level }}{{ end -}}` instead of `{{if .Values.tieredstore}}{{if .Values.tieredstore.level}}{{end}}{{end}}`. 

Example:
```
            {{- with .Values.tieredstore.level }}
                {{- range . }}
                  {{- if .mediumtype }} ... {{end}}
               {{- end }}
            {{- end }}
```

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -168,8 +172,6 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: status.hostIP
-          - name: ALLUXIO_JOB_WORKER_JAVA_OPTS
-            value: {{ print " -Dalluxio.worker.hostname=$(ALLUXIO_WORKER_HOSTNAME) " | quote }}
```
jiacheliu3: Same. Please bring these 2 lines back.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: For `hostPath` do we require root privilege to write to it?

jiacheliu3: According to K8s doc:
> the files or directories created on the underlying hosts are only writable by root. You either need to run your process as root in a privileged Container or modify the file permissions on the host to be able to write to a hostPath volume

So if Alluxio is not run as root, things will be tricky. We either need another `initContainer` running as root to change the permission or explicitly state that the user need to watch out for the permission constraint.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: @cheyang I suggest keeping this explanation as short as possible and leave a link to [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) and emptyDir K8s doc. The more we explain here the more we possibly have to update in the future.

For example, let's not explain what emptyDir does.

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
cheyang: I think it should use  the ALLUXIO_JOB_WORKER_JAVA_OPTS in configmap. Otherwise it will lose a lot of other configurations. I've tested it in our environment.

https://github.com/Alluxio/alluxio/blob/master/integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml#L41

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
cheyang: `ALLUXIO_WORKER_JAVA_OPTS ` in configmap can also handle this  property. The ALLUXIO_WORKER_HOSTNAME can be determined in runtime.

cheyang:
> I tried to port your new `worker-daemonset.yaml.template` to my local but tiered store volumes are not generated as expected. Did you try to modify `values.yaml` and use `helm-generate.sh all` to update the templates? Do they have updated volumes? What if you change the local `$dir/config.yaml` to have more tiers?
> 
> I hope to make sure that after this change both deploying with helm and generated templates with `helm template` all keep working. It is sometimes tricky how helm renders templates.

Here is value.yaml I used:

```
master:
  count: 1 # For multiMaster mode increase this to >1

journal:
  type: "UFS"
  ufsType: "local"
  folder: "/journal"

imageTag: 2.1.0-SNAPSHOT

# Security Context
user: 0
group: 0
fsGroup: 0

tieredstore:
  levels:
  - alias: MEM
    mediumtype: MEM
    level: 0
    path: /dev/shm
    type: hostPath
    high: 0.95
    low: 0.7
```

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
madanadit: yes we do. we will move away from the limitation as part of using local persistent volume in a separate PR

madanadit:
I ran `helm-generate` on this PR locally and this change seems to be unintended: 
```
           volumeMounts:
             - name: alluxio-domain
               mountPath: /opt/domain
-            - mountPath: /dev/shm
-              name: mem
       restartPolicy: Always
       volumes:
         - name: alluxio-domain
           hostPath:
             path: /tmp/alluxio-domain
             type: DirectoryOrCreate
-        - name: mem
-          emptyDir:
-            medium: "Memory"
-            sizeLimit: 1G
```

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
jiacheliu3: @cheyang So I suggest we add something like this to the doc: "Note that `hostPath` files or directories are only writable by root. You should make sure Alluxio worker has access to the tiered storage."

jiacheliu3:
> I ran `helm-generate` on this PR locally and this change seems to be unintended:
> 
> ```
>            volumeMounts:
>              - name: alluxio-domain
>                mountPath: /opt/domain
> -            - mountPath: /dev/shm
> -              name: mem
>        restartPolicy: Always
>        volumes:
>          - name: alluxio-domain
>            hostPath:
>              path: /tmp/alluxio-domain
>              type: DirectoryOrCreate
> -        - name: mem
> -          emptyDir:
> -            medium: "Memory"
> -            sizeLimit: 1G
> ```

That happened to me too. So either something in the tiered storage nested `if` and `range` went wrong, or in the local `$dir/config.yaml` shaded your `.Values.tieredstore`.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -30,6 +30,15 @@ properties:
   # alluxio.user.metrics.collection.enabled: 'true'
   alluxio.security.stale.channel.purge.interval: 365d
 
+tieredstore:
```
madanadit: @cheyang this is redundant. scroll below https://github.com/Alluxio/alluxio/commit/3added565773609aabc71eccaa5cdeefff55c164#diff-c26bfed7e349bf27fd0554872241278eL124

The default config is already specified as ```# Tiered Storage
tieredstore:
  levels:
  - alias: MEM # Medium type
    level: 0
    path: /dev/shm
    type: emptyDir
    quota: 1G
    high: 0.95
    low: 0.7
```

**Comment by madanadit** [comment]
@cheyang @jiacheliu3 This diff seems to do the job: 
```
+++ b/integration/kubernetes/helm-chart/alluxio/values.yaml
@@ -123,7 +123,8 @@ journal:
 # Tiered Storage
 tieredstore:
   levels:
-  - alias: MEM # Medium type
+  - alias: MEM
+    mediumtype: MEM
```
Let's make `mediumtype` a required parameter and not overload alias to simplify the helm template. We can remove the else condition (when mediumtype is not present) and it seems hard to maintain. What do you think?

File: integration/kubernetes/helm-chart/alluxio/values.yaml (hunk 13 above)
Comment (cheyang) [reply integration/kubernetes/helm-chart/alluxio/values.yaml]: Yes, you're right.

File: integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml [hunk 14]
```diff
@@ -128,16 +128,16 @@ spec:
               name: {{ $val | lower }}-{{ $key | replace "_" "" }}
                           {{- end}}
                         {{- end}}
-                      {{- end}}
                       {{- else}}
-                      {{- if eq .type "hostPath"}}
-            - name: {{ .alias | lower }}
+                        {{- if eq .type "hostPath"}}
+            - name: {{ .mediumtype | lower }}
```
madanadit: is `,` allowed as part of `name`? when `mediumtype` is `MEM,SSD` what will this resolve to?

jiacheliu3: Great question. I don't think K8s accepts comma in the name. I can address this separately in the doc.

madanadit: It's not just about the doc, multiple paths/medium types in the same tier will not work then

madanadit: we can replace this with `mount<level>` to have a unique name

madanadit:

PR #10243, opened 2019-10-24
Update check to identify cluster

madanadit:


madanadit:
@Xenorith are we good with the api version or do we need to bump that up to v1?

PR #10249, opened 2019-10-24
Allow extra arguments as directories for ls command

### LINKED ISSUE #10217: ls command does not support multiple directories
ls command can totally support listing multiple directories at the same time, just like Linux.

Accept more arguments as extra directories to be listed

**Urgency**
Just from personal point of view.

tianrenz2:
Fixes #10217


apc999:
Thanks for the improvement! Can you also update the usage string to be?

```java
  @Override
  public String getUsage() {
    return "ls [-d|-f|-p|-R|-h|--sort=option|--timestamp=option|-r] <path> ...";
   }
```

apc999:
if possible, mind creating PRs to make the same improvement to other commands like `RmCommand`? Thanks

apc999:

PR #10262, opened 2019-10-24
Add a warning message to detach command

yuzhu:


yuzhu:
@gpang PTAL

yuzhu:
jenkins, test this please

yuzhu:
jenkins, test this please

table/shell/src/main/java/alluxio/cli/table/command/DetachDatabaseCommand.java
```diff
@@ -58,7 +61,13 @@ public String getDescription() {
 
   @Override
   public int run(CommandLine cli) throws AlluxioStatusException {
-    mClient.detachDatabase(cli.getArgs()[0]);
+    String dbName = cli.getArgs()[0];
+    if (mClient.detachDatabase(dbName)) {
+      AlluxioURI path = new AlluxioURI(PathUtils
+          .concatPath(mConf.get(PropertyKey.TABLE_CATALOG_PATH), dbName));
+      System.out.println("Please unmount and clean up the alluxio location at " + path
```
gpang: I think it may not be obvious that the command succeeded. Maybe we start the message as "Successfully detached db {dbName}. Please unmount ..."

yuzhu: done

ZacBlanco: Ideally, we could even provide the user with the exact commands they need to run?

gpang: but we don't know which mount points are under that directory. Are you suggesting the CLI gets the mount points and processes the paths?

ZacBlanco: Something like that. Just a suggestion

yuzhu: We would have to expose internal mount paths to CLI to suggest those commands. 
I think since this is a temporary solution anyways. This should be fine? 
in the next version, we will have proper reference count and unmount in detach and this would be removed.

yuzhu:
@gpang PTAL

yuzhu:

yuzhu:

PR #10267, opened 2019-10-24
Add db level locking to catalog

yuzhu:


yuzhu:
@gpang PTAL. I went with DB level locking instead of table level locking. Because we sync over a db, if we just lock table individually, two sync operations will interfere with each other.  I don't see this as a perf problem.

yuzhu:

yuzhu:

PR #10300, opened 2019-10-25
Replace AtomicReference by volatile

chcn:
Since only `get` and `set` are used in these `AtomicReference`s, `volatile` can be used to save the memory cost of the extra `AtomicReference` instances.


gpang:

gpang:

PR #10330, opened 2019-10-28
Add doc for transformation

chcn:
[skip ci]

### DIFF
--- a/docs/en/structured-data/transformation-service.md
+++ b/docs/en/structured-data/transformation-service.md
@@ -6,13 +6,14 @@ group: Structured Data
 priority: 1
 ---
 
+* Table of Contents
+{:toc}
+
 ## Data Transformations
 
-With job service and catalog service, Alluxio can transform a table 
-to a new table at partition level. The data of the original table is not modified,
-and the data of the new table is persisted in a new location managed by Alluxio.
-Once the transformation is done, Presto users can transparently query
-against the new data.
+With job service and catalog service, Alluxio can transform a table to a new table at partition level.
+The data of the original table is not modified, and the data of the new table is persisted in a new location managed by Alluxio.
+Once the transformation is done, Presto users can transparently query against the new data.
 
 There are two kinds of supported transformations:
 
@@ -37,7 +38,7 @@ $ ${ALLUXIO_HOME}/bin/alluxio table transform default test -d "write(hive).optio
 After running the above command, you'll see output like:
 
 ```console
-$ Started transformation job with job ID 1572296710137, you can monitor the status of the job with './bin/alluxio table transformStatus 1572296710137'.
+Started transformation job with job ID 1572296710137, you can monitor the status of the job with './bin/alluxio table transformStatus 1572296710137'.
 ```
 
 Now follow the instruction in the output to monitor status of the transformation:
@@ -64,3 +65,8 @@ You can find out the location of the transformed data with the following Presto 
 presto:default> select "$path" from test;
 ```
 
+You should see output like:
+
+```console
+alluxio://localhost:19998/catalog/default/tables/test/_internal_/part=0/20191024-213102-905-R29wf/part-0.parquet
+```
--- a/docs/en/structured-data/transformation-service.md
+++ b/docs/en/structured-data/transformation-service.md
@@ -6,11 +6,61 @@ group: Structured Data
 priority: 1
 ---
 
-* Table of Contents
-{:toc}
-
 ## Data Transformations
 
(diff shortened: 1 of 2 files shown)


docs/en/structured-data/transformation-service.md
```diff
@@ -6,11 +6,61 @@ group: Structured Data
 priority: 1
 ---
 
-* Table of Contents
-{:toc}
```
ZacBlanco: Is there a reason to remove the TOC?

docs/en/structured-data/transformation-service.md
```diff
@@ -6,11 +6,61 @@ group: Structured Data
...
-
 ## Data Transformations
 
+With job service and catalog service, Alluxio can transform a table 
+to a new table at partition level. The data of the original table is not modified,
```
ZacBlanco: new sentences should be on a new line:

https://docs.alluxio.io/os/user/stable/en/contributor/Documentation-Conventions.html#line-breaks

docs/en/structured-data/transformation-service.md
```diff
@@ -6,11 +6,61 @@ group: Structured Data
...
+The following command coalesce files under each partition of table "test" to one file.
+
+```console
+$ ${ALLUXIO_HOME}/bin/alluxio table transform default test -d "write(hive).option(hive.num.files, 1)"
+```
+
+After running the above command, you'll see output like:
+
+```console
+$ Started transformation job with job ID 1572296710137, you can monitor the status of the job with './bin/alluxio table transformStatus 1572296710137'.
```
ZacBlanco: for `console` type code blocks, you only need to put  `$` on lines which are commands that should be typed (or copy-pasted) in by the user. Output lines don't need to start with `$`

docs/en/structured-data/transformation-service.md
```diff
@@ -6,11 +6,61 @@ group: Structured Data
...
+job ID: 1572296710137
+job status: COMPLETED
 ```
+
+Since the transformation has completed, you can run your Presto queries on the transformed table transparently.
+
+You can find out the location of the transformed data with the following Presto query:
+
+```console
+presto:default> select "$path" from test;
```
ZacBlanco: is there an output that we want to show with this command?

table/shell/src/main/java/alluxio/cli/table/command/TransformTableCommand.java
```diff
@@ -72,7 +72,9 @@ public Options getOptions() {
   public String getDescription() {
     return "Transform files representing a structured table under an Alluxio directory."
         + "\n\n"
-        + "Files are coalesced and converted to parquet format.";
+        + "Files are coalesced and converted to parquet format."
+        + "\n\n"
+        + "Definition is in format like 'write(hive).option(hive.num.files, 100)'.";
```
ZacBlanco: Can we point users to some kind of regex about how to specify these options, or what the valid values are? I don't entirely understand the format based on the docs and usage string here.

chcn: due to the restriction of parser's implementation, currently, only a limited set of definition is supported, I added more explanation.

chcn:
@ZacBlanco PTAL

table/shell/src/main/java/alluxio/cli/table/command/TransformTableCommand.java
```diff
@@ -74,7 +74,9 @@ public String getDescription() {
         + "\n\n"
         + "Files are coalesced and converted to parquet format."
         + "\n\n"
-        + "Definition is in format like 'write(hive).option(hive.num.files, 100)'.";
+        + "Definition is in format like 'write(hive).option(hive.num.files, 100)',"
+        + "currently, the only modification you can do to the example definition is "
+        + "the number of files in the option."
```
ZacBlanco: I think this is better suited to live in the documentation page.  Maybe have a short reference section of some kind with all the available options - rather than putting this in the command output?

chcn: I think since this will change in future release, it's better to let it be here so it's easier to update.

ZacBlanco: All right - but if we start adding more I think it is better suited to live in the documentation, or have some kind of pretty-printed output that we can provide to users with the full list of options. Otherwise, no one will really know how to use it

ZacBlanco:

ZacBlanco:

Issue #10321, opened 2019-10-28
Helm install creates unnecessary resources


madanadit:
**Alluxio Version:**
2.1.0-SNAPSHOT

**Describe the bug**
Helm install creates everything under the directory `helm-chart/alluxio/templates/`. Some of these resources / jobs should not be created by default and made configurable.

1. Is this still needed `helm-chart/alluxio/templates/format/master.yaml` or can we remove it?
2. `helm install` run the fuse daemonset and client daemonset by default, both should be made configurable and false by default. 
```
fuse:
  enabled: true
  clientEnabled: true
```
3. `job/format-journal-job.yaml` is run by default when `helm install` is run. how should we disable it and invoke it on demand?
4. `master/journal-pv.yaml` and `master/journal-pvc.yaml` are created by default even when using configs which don't use it. This should also be disabled when not needed.

**To Reproduce**
Run `helm install` and look at the resources / jobs created using `kubectl`

**Expected behavior**
N/A

**Urgency**
N/A

**Additional context**
N/A

jiacheliu3:
Regarding the action items:

1. `helm-chart/alluxio/templates/format/master.yaml` is removed. We don't really need it in the template deployment.

2.(and 3. and 4.) As far as I found out the ONLY way to make `helm install` NOT create every single resource it can find is to make helm not see it, by excluding those files in `.helmignore`. As long as you have that file under the `templates/` directory I didn't think of an easy way to switch it ON/OFF. 
If `fuse/` and `job/` shouldn't be run they can be added to `.helmignore`.

To make matters worse, files defined in `.helmignore` will be ignored by `helm template`. We need `helm template` to generate templates in `job/` and `fuse/`. That's why I don't want to include these two dirs in `.helmignore`.

So as you see, `.helmignore` affects `helm install` and `helm template` together (even though the doc says this only affects the packaging step). We cannot make something ignored only for `helm install`. 

I lean towards NOT adding these 2 directories in `.helmignore` and explain in `Deploy using helm` how the user can make Helm not create the resource they don't need.

madanadit:
@jiacheliu3 @ZacBlanco not targeting this PR to 2.1.

calvinjia:
@madanadit Are we still planning to address this issue?

madanadit:
@calvinjia Yes, these issues are still present. We should prioritize these.

PR #10333, opened 2019-10-29
[WIP] Fix issue helm install command creating unnecessary resources

jiacheliu3:
This change fixes https://github.com/Alluxio/alluxio/issues/10321
Removes `templates/format/master.yaml`
Removes `apiVersion` key from `Chart.yaml`
Updates doc


integration/kubernetes/README.md
```diff
@@ -58,12 +58,14 @@ for documentation.
 Now that you have set up all that Helm needs, you can use the helper script `helm-generate.sh` to help you quickly generate the templates(overwriting the existing ones).
 The script takes only one argument, the target setup name.
 
+> Note: `helm-generate.sh` uses `helm template` command under the hood.
+Templates ignored with `.helmignore` file will not be generated.  
+
```
jiacheliu3: Added ref where contents added to `.helmignore` in Helm deployment will affect kubectl deployment

integration/kubernetes/helm-chart/alluxio/Chart.yaml
```diff
@@ -10,7 +10,6 @@
 #
 
 name: alluxio
-apiVersion: 2.2.0-SNAPSHOT
```
jiacheliu3: Helm will actually complain on this `apiVersion` for not being `v1` and helm commands will fail working on this repo. By the key `apiVersion` it really expects an K8s api version. The Alluxio version is fed with `.Values.imageTag` so we don't actually need this field in `Chart.yaml`. If we really need this field we should then rename it.

docs/en/deploy/Running-Alluxio-On-Kubernetes.md
```diff
@@ -289,13 +289,39 @@ tieredstore:
     low: 0.7
 ```
 
+#### Update Ignore List
```
madanadit: Hmm, this solution does not work if someone is using an existing helm repo? 

After we publish the chart to the stable helm repo, the only command needed to install Alluxio is `helm install alluxio` passing in the config yaml. How does the user choose to not install fuse/job; is there a way to pass in the ignore list to `helm install`?

jiacheliu3: I don't find any way to specify an include/exclude list on a repo so what's there gets installed.

madanadit: Otherwise we can wrap them up in an if condition with properties from config.yaml. 

For the generation of fuse / job sample templates we can commit a `config.yaml` in a sub-dir with those properties enabled and run `helm-generate` on it.

jiacheliu3:
WIP due to better ideas on how to make helm ignore certain resources.

ZacBlanco:
@jiacheliu3 are you still working on this?

jiacheliu3:
@ZacBlanco Yes we are postponing this until we have plans for Helm development in the next step. Thanks for checking :)

jiacheliu3:
Closed as updating helm ignore list is not a good approach. This shall be addressed in a separate PR.

PR #10343, opened 2019-10-29
Fix UFS RPC metrics aggregation

### LINKED ISSUE #9628: Metrics JSON endpoint not reporting correct UFS metrics

yuzhu:
Originally, the metrics are aggregated on UFS paths. But in scenario where two paths in the same UFS are mounted to different Alluxio mount points, since there is only one UFS client for the mount points, all metrics are aggregated to the first mount point even when operations are on the second mount point, which is confusing.
This PR aggregates UFS RPC metrics per UFS, instead of per UFS path.

fixes #9628

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
+++ b/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
@@ -965,7 +965,7 @@ public class UnderFileSystemWithLogging implements UnderFileSystem {
 
   private String getUfsTag() {
     AlluxioURI uri = new AlluxioURI(mPath);
-    uri = new AlluxioURI(uri.getScheme(), uri.getAuthority(), uri.getRootPath());
+    uri = new AlluxioURI(uri.getScheme(), uri.getAuthority(), uri.getPath());
     return MetricsSystem.escape(uri);
   }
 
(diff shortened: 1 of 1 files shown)


yuzhu:
@chcn ported #9846 to target master branch. PTAL

yuzhu:
@ZacBlanco @rastogiasr what is the verdict on this? to master but not to branch-2.1?

yuzhu:
@chcn PTAL

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
```diff
@@ -961,22 +963,26 @@ public boolean isSeekable() {
     return mUnderFileSystem.isSeekable();
   }
 
+  private String getUfsTag() {
+    AlluxioURI uri = new AlluxioURI(mPath);
+    uri = new AlluxioURI(uri.getScheme(), uri.getAuthority(), uri.getPath());
```
chcn: what's the purpose of this?

yuzhu: emm this is from your old PR...

yuzhu: #9846

PR #10364, opened 2019-10-30
Ignore all tests that use MultiMasterLocalAlluxioCluster

bradyoo:
I wrote a small script to see how many times MultiMasterJournalTest failed in the last few builds

➜  ~ python flakeTest.py MultiMasterJournalTest 6605 6628 (23 builds, doesn't include 6628)
12
➜  ~ python flakeTest.py MultiMasterJournalTest 6620 6628 (8 builds)
5

Seems like it fails on about half the builds (though the build may succeed because of retries). Ignore the test.

### DIFF
--- a/tests/src/test/java/alluxio/client/cli/fs/JobServiceFaultToleranceShellTest.java
+++ b/tests/src/test/java/alluxio/client/cli/fs/JobServiceFaultToleranceShellTest.java
@@ -25,6 +25,7 @@ import alluxio.testutils.IntegrationTestUtils;
 
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestName;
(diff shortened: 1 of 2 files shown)


bradyoo:
➜  ~ python flakeTest.py JobServiceFaultToleranceShellTest 6605 6628
21

Yeah.. :/

ggezer:
> FYI, every single test that uses the MultiMasterLocalAlluxioCluster is very flaky. Every one of them were ignored except for this one and JobServiceFaultToleranceShellTest and probably that one should be ignored too.

There are tests that uses `MultiMasterLocalAlluxioCluster` except these two.
Also these two tests provide important coverage. Silencing them is not favorable over flaky test pass IMO.

bradyoo:
All the other usages of MultiMasterLocalAlluxioCluster test are also
ignored.

bradyoo:
I think it is a valuable discussion to have on whether coverage is more
important than productivity.

ZacBlanco:
Frankly, we have poor test coverage across our codebase. Not to mention HA scenarios are more  important to test IMO. I would prefer not to give up test coverage.

ggezer:
> All the other usages of MultiMasterLocalAlluxioCluster test are also ignored.
There is still `MasterFailoverIntegrationTest`

bradyoo:
Hmm this one hasn't flaked recently, I'll take a look. But it seems like people like the coverage that this test maybe providing so will close.

PR #10386, opened 2019-11-01
Fix build and run in branch 2.1

apc999:
Fix building and running source code from branch 2.1


calvinjia:
This should be covered by #10392

PR #10397, opened 2019-11-01
Fix race condition building extension properties

bf8086:
There is a race condition when building multiple ufs configuration concurrently. Extension properties of the same name that are not built-in will be created as custom property but can throw due to duplication in the property registration.

Fixed by synchronizing registration of custom properties .


bf8086:
@madanadit @calvinjia PTAL.

calvinjia:

bf8086:

PR #10407, opened 2019-11-04
Replace FindBugs with SpotBugs maven plugin

### LINKED ISSUE #10406: Swap FindBugs plugin with SpotBugs
**Summary**

The original codehaus FindBugs plugin is deprecated and no longer under active maintenance - https://gleclaire.github.io/findbugs-maven-plugin/

The successor to findbugs is SpotBugs - https://spotbugs.github.io/ It contains more checks for poor patterns throughout the codebase and is actively maintained.

We should use the newer plugin. I made a small attempt at switching, and spotbugs has found nearly 200 issues which findbugs didn't. We should switch to this ASAP.

**Urgency**

Help reduce common bugs across our code base.

ZacBlanco:
Findbugs is deprecated. This PR switches the original FindBugs plugin
to the newer and better-maintained SpotBugs. Additionally, the original
list of bug exclusions has been revised, and a number of bugs fixed
throughout the codebase.

The original build flag `-Dfindbugs.skip` will still work after this change.

Fixes #10406

### DIFF
--- a/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
@@ -37,6 +37,8 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import javax.annotation.Nonnull;
+
 /**
  * Alluxio configuration.
  */
@@ -199,7 +201,7 @@ public class InstancedConfiguration implements AlluxioConfiguration {
    * @param value the value for the key
    * @param source the source of the the properties (e.g., system property, default and etc)
    */
-  public void set(PropertyKey key, Object value, Source source) {
+  public void set(@Nonnull PropertyKey key, @Nonnull Object value, @Nonnull Source source) {
     Preconditions.checkArgument(key != null && value != null && !value.equals(""),
         String.format("The key value pair (%s, %s) cannot be null", key, value));
     Preconditions.checkArgument(!value.equals(""),
--- a/core/common/src/main/java/alluxio/conf/Source.java
+++ b/core/common/src/main/java/alluxio/conf/Source.java
@@ -14,7 +14,6 @@ package alluxio.conf;
 import alluxio.annotation.PublicApi;
 
 import com.google.common.base.Objects;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
  * The source of a configuration property.
@@ -22,7 +21,6 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
  * The natural ordering of this class is not consistent with the {@link #equals(Object)} method.
  */
 @PublicApi
-@SuppressFBWarnings("EQ_COMPARETO_USE_OBJECT_EQUALS")
 public class Source implements Comparable<Source> {
   public static final Source UNKNOWN = new Source(Type.UNKNOWN);
   public static final Source DEFAULT = new Source(Type.DEFAULT);
@@ -98,6 +96,26 @@ public class Source implements Comparable<Source> {
   @Override
   public int compareTo(Source other) {
     return mType.compareTo(other.mType);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null) {
+      return false;
+    }
+
+    if (!(o instanceof Source)) {
+      return false;
+    }
+
+    Source other = (Source) o;
+
+    return compareTo(other) == 0;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(mType);
   }
 
   @Override
--- a/core/common/src/main/java/alluxio/conf/Source.java
+++ b/core/common/src/main/java/alluxio/conf/Source.java
@@ -100,8 +100,8 @@ public class Source implements Comparable<Source> {
 
   @Override
   public boolean equals(Object o) {
-    if (o == null) {
-      return false;
+    if (this == o) {
+      return true;
     }
 
     if (!(o instanceof Source)) {
@@ -133,14 +133,11 @@ public class Source implements Comparable<Source> {
 
     @Override
     public boolean equals(Object o) {
-      if (o == null) {
-        return false;
-      }
-
       if (this == o) {
         return true;
       }
-      if (!(this.getClass().equals(o.getClass()))) {
+
+      if (o == null || this.getClass() != o.getClass()) {
         return false;
       }
       SitePropertySource that = (SitePropertySource) o;
--- a/core/common/src/main/java/alluxio/conf/Source.java
+++ b/core/common/src/main/java/alluxio/conf/Source.java
@@ -133,10 +133,14 @@ public class Source implements Comparable<Source> {
 
     @Override
     public boolean equals(Object o) {
+      if (o == null) {
+        return false;
+      }
+
       if (this == o) {
         return true;
       }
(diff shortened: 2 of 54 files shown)


yuzhu:
```
Tests in error: 
  Run 1: JobServiceFaultToleranceShellTest.before:52 » IO java.lang.RuntimeException: a...
  Run 2: JobServiceFaultToleranceShellTest.before:52 » IO java.lang.RuntimeException: a...

Flaked tests: 
alluxio.client.cli.fs.command.PersistCommandTest.persistOnRenameDirectory(alluxio.client.cli.fs.command.PersistCommandTest)
  Run 1: PersistCommandTest.persistOnRenameDirectory:105->AbstractFileSystemShellTest.checkFilePersisted:306
  Run 2: PASS

  Run 2: PASS 
```

yuzhu:
jenkins, test this please

ZacBlanco:
I think before merging this PR we should re-evaluate the exclusions/inclusions within the `build/findbugs/findbugs-exclude.xml`

yuzhu:
@ZacBlanco could we remove all the bug code and see what the output looks like? so we can decide which bug code to keep?

wyykwan:
jenkins, test this pleaseeeee

wyykwan:
jenkins, test this please

ZacBlanco:
@gpang @yuzhu FYI

So I found while running this PR that we have a ton of locations that findbugs complains about with the error code [`REC_CATCH_EXCEPTION`](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#rec-exception-is-caught-when-exception-is-not-thrown-rec-catch-exception)

The short explanation is that there's are many places throughout our codebase where we do something like

```java
try {
  // Code that might throw IOException, NPE, NumberFormatException...
} catch (Exception e) { // This exception literally catches everything
 // Generic exception handler
}
```

Spotbugs' explanation of this is essentially there is rarely a time when it is desirable to catch/handle all exceptions, regardless of class. Most often you'll at least want to handle `RuntimeException` separately from `Exception` because they are different classes of errors.

Across the project we do a pretty poor job (according to Spotbugs) of exception handling as there are probably 30+ locations where there is simply `catch (Exception e)`. I tried to fix a lot of them, but I'm worried that I don't have enough context on how this could affect the reliability of the program, such as introducing crashes due to not catching the proper exception class. Nor do I properly know how the different classes of exceptions should be handled in all the different cases I encountered.

What I'm going to do is revert any changes related to `REC_CATCH_EXCEPTION` and file a ticket that can be addressed in a separate PR to deal with our poor exception handling.

gpang:
@ZacBlanco That sounds like a good plan. Let's get the simpler fixes/changes in first, and address the bigger issues separately. Thanks!

ZacBlanco:
Created https://github.com/Alluxio/alluxio/issues/10459 to track the debt for exception handling.

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
```diff
@@ -137,7 +137,10 @@ private UnderFileSystem getOrAdd(AlluxioURI ufsUri, UnderFileSystemConfiguration
         fs = new ManagedBlockingUfsForwarder(fs);
       }
 
-      mUnderFileSystemMap.putIfAbsent(key, fs);
+      if (mUnderFileSystemMap.putIfAbsent(key, fs) != null) {
+        // This shouldn't occur, warn if it
```
gpang: is this comment complete?

core/common/src/test/java/alluxio/util/IdUtilsTest.java
```diff
@@ -55,4 +57,13 @@ public void createRpcId() throws Exception {
     assertTrue(!second.isEmpty());
     assertNotEquals(first, second);
   }
+
+  @Test
+  public void testPositiveLong() {
```
gpang: why was this test added?

shell/src/main/java/alluxio/cli/ConfigurationDocGenerator.java
```diff
@@ -192,7 +192,7 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
       }
 
       LOG.info("YML files for description of Property Keys were created successfully.");
-    } catch (Exception e) {
+    } catch (Exception  e) {
```
gpang: NIT: extra space

shell/src/main/java/alluxio/cli/ConfigurationDocGenerator.java
```diff
@@ -114,7 +114,7 @@ public static void writeCSVFile(Collection<? extends PropertyKey> defaultKeys, S
       }
 
       LOG.info("Property Key CSV files were created successfully.");
-    } catch (Exception e) {
+    } catch (RuntimeException | IOException e) {
```
gpang: Undo this change for another PR?

core/common/src/main/java/alluxio/conf/Source.java
```diff
@@ -111,10 +133,14 @@ private SitePropertySource(String filename) {
...
     public boolean equals(Object o) {
+      if (o == null) {
+        return false;
+      }
+
       if (this == o) {
         return true;
       }
-      if (!(o instanceof SitePropertySource)) {
+      if (!(this.getClass().equals(o.getClass()))) {
```
gpang: is this because of the inheritance?

core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
```diff
@@ -199,7 +201,7 @@ public void set(PropertyKey key, Object value) {
    * @param value the value for the key
    * @param source the source of the the properties (e.g., system property, default and etc)
    */
-  public void set(PropertyKey key, Object value, Source source) {
+  public void set(@Nonnull PropertyKey key, @Nonnull Object value, @Nonnull Source source) {
```
gpang: this annotation is required?

core/common/src/test/java/alluxio/util/IdUtilsTest.java
ZacBlanco: outdated comment. Test already removed

shell/src/main/java/alluxio/cli/ConfigurationDocGenerator.java
ZacBlanco: outdatd comment, already fixed

core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
ZacBlanco: it's required or else spotbugs complains about possible null derefence.

AFAIK it can't determine that `Preconditions` does the null check

core/common/src/main/java/alluxio/conf/Source.java
ZacBlanco: yes

shell/src/main/java/alluxio/cli/ConfigurationDocGenerator.java
ZacBlanco: outdated, already fixed

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
ZacBlanco: updated

ZacBlanco:

PR #10412, opened 2019-11-06
Bring down runtime of S3ClientRestApiTest from 5 minutes to sub minute

bradyoo:


bradyoo:
I figured people will make meaning out of constants if I put in real constants.. There really is no meaning here other than, "Please look somewhere else."

bradyoo:
Because I accidentally flipped the comparison direction, I also changed the limit to 4KB instead of 5000 bytes.

bradyoo:
I have a bad feeling that this didn't actually make that big of a difference :/

ZacBlanco:
You won't see a huge decrease in time because there's lots of tests running in parallel. Making this one shorter will only allow some other tests to run on the process it was consuming a bit earlier. Any amount of change is good though

bradyoo:
Tests run: 37, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 370.645 sec - in alluxio.client.rest.S3ClientRestApiTest 

Yeah I didn't make a dent. My testing process was bad :(

bradyoo:
(even looking at this test individually)

PR #10418, opened 2019-11-06
Fix NPE in UfsDirectoryStatus

bf8086:
In some UFSes like S3, directories will have a null lastModifiedTimeMs indicating the timestamp is not available. We used `long` as constructor parameter type which will throw when null is passed in for a S3 directory.


bf8086:

bf8086:

PR #10419, opened 2019-11-06
Remove some String.format calls

yuzhu:


yuzhu:
@ZacBlanco possible improvement by converting `String.format` to concatenations.

ZacBlanco:
Is there really a measurable performance impact from any of these changes?

I know `String.format` is slower in most all cases, but to me it's a bit more readable.

yuzhu:
It is not clear whether we will trigger the specific calls easily with our current testing infrastructure. 
I only looked at those that are potentially called multiple times and not just for logging purposes. 

At the very least, we should remove the format call in every RPC call.

gpang:
@yuzhu Are there any experiments which show a difference?

core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -107,7 +107,12 @@ private RpcUtils() {} // prevent instantiation
       String methodName, boolean failureOk, String description, Object... args)
       throws StatusException {
     // avoid string format for better performance if debug is off
-    String debugDesc = logger.isDebugEnabled() ? String.format(description, args) : null;
+    String debugDesc;
+    if (logger.isDebugEnabled()) {
+      debugDesc = String.format(description, args);
+    } else {
```
chcn: this block can be removed by initializing debugDesc to null?

core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -157,7 +162,12 @@ private RpcUtils() {} // prevent instantiation
       String methodName, boolean sendResponse, boolean completeResponse,
       StreamObserver<T> responseObserver, String description, Object... args) {
     // avoid string format for better performance if debug is off
-    String debugDesc = logger.isDebugEnabled() ? String.format(description, args) : null;
+    String debugDesc;
+    if (logger.isDebugEnabled()) {
+      debugDesc = String.format(description, args);
+    } else {
```
chcn: this block can be removed by initializing debugDesc to null?

PR #10424, opened 2019-11-06
Refactor Job Service

bradyoo:
Changes Included:

Job = Plan | Task | (Workflow) to be introduced

What used to be "Job" in the "Job"/"Task" relationship is now called "Plan".

job_master.proto: The info object of PlanInfo and TaskInfo are merged to a single Info proto object. This is because of the lack of inheritance and WorkflowInfo will have children that is a collection of either PlanInfo or WorkflowInfo.

Almost all of the modules short of JobMaster and the JobClients and the ServiceHandlers have been moved to the plan subpackage. 

The Context modules are untouched for now but everything except for JobServerContext probably also belongs in the plan module.

CompositeConfig will be in the Workflow Module and will be a WorkflowConfig but that implementation is not in this PR.

TaskExecutorManager has been refactored to be using the wire object instead of the GRPC builder object.

Job Description and Task worker info has been added.


core/transport/src/grpc/job_master.proto
```diff
@@ -17,22 +17,32 @@ enum Status {
...
-  optional string errorMessage = 4;
-  optional bytes result = 5;
+enum JobType {
+  PLAN = 1;
+  TASK = 2;
 }
 
+message JobUnused {}
+
+// next available id: 12
```
calvinjia: This should be `14`?

job/client/src/main/java/alluxio/client/job/JobMasterClient.java
```diff
@@ -13,8 +13,8 @@
 
 import alluxio.Client;
 import alluxio.job.JobConfig;
-import alluxio.job.wire.JobInfo;
 import alluxio.job.wire.JobServiceSummary;
+import alluxio.job.wire.JobInfo;
```
calvinjia: Is the import order correct?

job/client/src/main/java/alluxio/client/job/JobMasterClient.java
```diff
@@ -66,15 +66,15 @@ public static JobMasterClient create(JobMasterClientContext conf) {
   JobServiceSummary getJobServiceSummary() throws IOException;
 
   /**
-   * @return the list of ids of all jobs
+   * @return the list of all plan ids
```
calvinjia: Is this wording because a Job can either be a task or plan and only top level plans are returned?

job/common/src/main/java/alluxio/job/plan/PlanConfig.java
```diff
@@ -9,22 +9,12 @@
...
-  /**
-   * Writes an entry to a checkpoint file.
-   *
-   * The entry should not have its sequence number set. This method will add the proper sequence
-   * number to the passed in entry.
-   *
-   * @param entry an entry to write to a journal checkpoint file
-   */
-  void writeJournalEntry(JournalEntry entry);
+public interface PlanConfig extends JobConfig {
```
calvinjia: Do we plan to have something here in the future?

job/common/src/main/java/alluxio/job/plan/meta/PlanInfo.java
```diff
@@ -9,12 +9,13 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.job.meta;
+package alluxio.job.plan.meta;
 
 import alluxio.job.JobConfig;
 import alluxio.job.wire.Status;
-import alluxio.job.wire.TaskInfo;
+import alluxio.job.plan.wire.TaskInfo;
```
calvinjia: Is the import order correct?

job/common/src/main/java/alluxio/job/plan/wire/PlanInfo.java
```diff
@@ -0,0 +1,220 @@
...
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.job.plan.wire;
```
calvinjia: Move this to `alluxio.job.wire`?

job/common/src/main/java/alluxio/job/plan/wire/TaskInfo.java
```diff
@@ -9,49 +9,96 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.job.wire;
+package alluxio.job.plan.wire;
```
calvinjia: Move this to `alluxio.job.wire`?

job/common/src/main/java/alluxio/job/workflow/composite/CompositeConfig.java
```diff
@@ -29,7 +30,7 @@
  * execute a set of jobs concurrently, each job can run run another set of sub jobs sequentially.
  */
 @ThreadSafe
-public final class CompositeConfig implements JobConfig {
+public final class CompositeConfig implements PlanConfig {
```
calvinjia: Will everything be a composite config? Does it make sense to have a single `workflow` package instead of `workflow.composite`?

job/common/src/main/java/alluxio/job/workflow/WorkflowConfig.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package alluxio.job.workflow;
+
+import alluxio.job.JobConfig;
+
+/**
+ * A workflow configuration.
+ */
+public interface WorkflowConfig extends JobConfig {
```
calvinjia: Is this interface necessary (currently unused)?
Also, are workflows subclasses of jobs, I thought the relationship was Job = Task | Plan

bradyoo:
@calvinjia: This is the one that contains everything.

job/common/src/main/java/alluxio/job/plan/meta/MasterWorkerInfo.java
```diff
@@ -9,7 +9,7 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.job.meta;
+package alluxio.job.plan.meta;
```
calvinjia: Why is this in the `plan` package?

job/common/src/main/java/alluxio/job/plan/meta/PlanInfo.java
```diff
@@ -9,12 +9,13 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.job.meta;
+package alluxio.job.plan.meta;
```
calvinjia: Does this need to be in the `meta` package, could we place it in `plan` directly?

job/client/src/main/java/alluxio/worker/job/JobMasterClient.java
```diff
@@ -15,7 +15,7 @@
 import alluxio.exception.AlluxioException;
 import alluxio.exception.ConnectionFailedException;
 import alluxio.grpc.JobCommand;
-import alluxio.grpc.TaskInfo;
+import alluxio.grpc.JobInfo;
```
calvinjia: This class should be in the `server` module instead of `client`. Though I think its not something this refactor introduced. Although it is a `client`, we use the `client` module for client applications to the servers and we use the server modules to hold clients that servers use to communicate with each other.

job/client/src/main/java/alluxio/job/plan/replicate/DefaultReplicationHandler.java
```diff
@@ -9,7 +9,7 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.job.replicate;
+package alluxio.job.plan.replicate;
```
calvinjia: Not something introduced by this PR - does this class belong here or as part of the `core/master` module?

job/common/src/main/java/alluxio/job/workflow/WorkflowConfig.java
bradyoo: Job is everything because you need to be able to submit everything through the JobService.

job/common/src/main/java/alluxio/job/plan/meta/MasterWorkerInfo.java
bradyoo: Because it is used by the PlanCoordinator.

job/common/src/main/java/alluxio/job/plan/meta/PlanInfo.java
bradyoo: PlanInfo has a wire representation and a meta representation. The existence of both probably makes it such that it's better to have a meta folder.

job/client/src/main/java/alluxio/client/job/JobMasterClient.java
bradyoo: No. This is probably me missing out on a rename. I'll change this one back.

job/common/src/main/java/alluxio/job/plan/meta/MasterWorkerInfo.java
calvinjia: It seems to be only be used by `JobMaster`?

job/common/src/main/java/alluxio/job/plan/PlanConfig.java
bradyoo: This is what I currently use to differentiate between using the PlanTracker versus the WorkflowTracker in the JobMaster.

I think code snippet of what it looks like will help:


```
  public synchronized void run(JobConfig jobConfig, long jobId)
      throws JobDoesNotExistException, ResourceExhaustedException {
    // This RPC service implementation triggers another RPC.
    // Run the implementation under forked context to avoid interference.
    // Then restore the current context at the end.
    Context forkedCtx = Context.current().fork();
    Context prevCtx = forkedCtx.attach();
    try {
      if (jobConfig instanceof PlanConfig) {
        mPlanTracker.run((PlanConfig) jobConfig, mCommandManager, mJobServerContext,
            getWorkerInfoList(), jobId);
        return;
      } else if (jobConfig instanceof WorkflowConfig) {
        mWorkflowTracker.run((WorkflowConfig) jobConfig, jobId);
        return;
      }
      throw new JobDoesNotExistException(
          ExceptionMessage.JOB_DEFINITION_DOES_NOT_EXIST.getMessage(jobConfig.getName()));
    } finally {
      forkedCtx.detach(prevCtx);
    }
  }
```

The Plan part is the same but the Workflow part looks something like:

```
  public void run(WorkflowConfig workflowConfig, long jobId)
      throws JobDoesNotExistException, ResourceExhaustedException {
    WorkflowExecutionFactory<WorkflowConfig> executionFactory =
        WorkflowExecutionFactoryRegistry.INSTANCE.getExecutionFactory(workflowConfig);

    WorkflowExecution workflowExecution = executionFactory.create(workflowConfig);

    mWorkflows.put(jobId, workflowExecution);

    // Some other code to start off the workflow
    ...
    ....
  }
```

job/client/src/main/java/alluxio/worker/job/JobMasterClient.java
bradyoo: That is correct. This file doesn't have much of a business being in the client module.

job/client/src/main/java/alluxio/job/plan/replicate/DefaultReplicationHandler.java
bradyoo: Probably? But not something I can answer.

job/common/src/main/java/alluxio/job/plan/meta/MasterWorkerInfo.java
bradyoo: Hmm.. Ideally, I would want to make JobMaster more Worker agnostic but that's fair. Let me move it out.

job/common/src/main/java/alluxio/job/workflow/composite/CompositeConfig.java
bradyoo: If all dependencies will be based around a predefined set of jobs, then it would be the case. However, I don't expect that to be the case for long. For example, I plan to rewrite MigrateDefinition's parallelism to be defined by some file system iteration config with specified depth. Which would be a workflow but with a very different implementation.

bradyoo:
Addressed the package feedbacks. Though I'm noticing that not all comments are actually making it to this main page...

bradyoo:
Here is what the current Workflow Implementation on top of this refactor currently looks like: https://github.com/bradyoo/alluxio/pull/3/files

job/common/src/main/java/alluxio/job/workflow/composite/CompositeConfig.java
calvinjia: Should `CompositeConfig` implement `WorkloadConfig`?

job/server/src/main/java/alluxio/master/job/plan/PlanTracker.java
```diff
@@ -96,20 +94,19 @@ public JobTracker(long capacity, long retentionMs, long maxJobPurgeCount) {
     mCoordinators = new ConcurrentHashMap<>(0,
         0.95f, ServerConfiguration.getInt(PropertyKey.MASTER_RPC_EXECUTOR_PARALLELISM));
     mFinished = new LinkedBlockingQueue<>();
-    mJobIdGenerator = new JobIdGenerator();
   }
 
-  private void statusChangeCallback(JobInfo jobInfo) {
-    if (jobInfo == null) {
+  private void statusChangeCallback(PlanInfo planInfo) {
```
calvinjia: Consider passing in the new status instead of having to derive it by inspecting `PlanInfo`

calvinjia:
@bradyoo I think this is good to go. A few things we noted but aren't addressing here:

1. Move JobWorker clients from client to server
2. Look into replicate definition and see if it makes sense for this to be in the job module or the core/master module
3. Consider improving the status change callback by passing in the new status directly

bradyoo:
Should I do a test merge to enterprise of this to see what that looks like? I assume it's going to have many merge conflicts.

calvinjia:

PR #10433, opened 2019-11-08
Enable jacoco code coverage plugin

gpang:


ZacBlanco:
have you tested locally on what the impact of this is for build time. Are there any github integrations we can add to display a code coverage report?

pom.xml
```diff
@@ -915,7 +916,7 @@
           <artifactId>maven-surefire-plugin</artifactId>
           <version>2.19.1</version>
           <configuration>
-            <argLine>-Djava.net.preferIPv4Stack=true</argLine>
+            <argLine>@{argLine} -Djava.net.preferIPv4Stack=true</argLine>
```
ZacBlanco: what's the reason for needing this?

ZacBlanco:
The last run took ~32 minutes. It's a bit longer than normal but could be due to contention on the same jenkins worker. I'll run it a few times to get an 

jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
We also might want to modify the jenkins PR build jenkins file to collect the code coverage artifacts?

gpang:
The default for the profile is off, so it doesn't run code coverage. Later, I will use this profile to display it elsewhere.

pom.xml
gpang: This is needed because both surefire AND jacoco change the argLine so this is how you are supposed to configure it in order to allow both to update argLine

ZacBlanco:
Oh, that's my bad. I missed the `<profile>` section

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:

PR #10434, opened 2019-11-08
Exit from lost-file detector hearbeat when can't progress

ggezer:


core/server/master/src/main/java/alluxio/master/file/LostFileDetector.java
```diff
@@ -65,7 +65,8 @@ public void heartbeat() throws InterruptedException {
       } catch (FileDoesNotExistException e) {
         LOG.debug("Exception trying to get inode from inode tree", e);
       } catch (UnavailableException e) {
-        LOG.warn("Failed to run lost file detector: {}", e.toString());
+        LOG.warn("Failed to run lost file detector. Quiting heart-beat.", e);
```
apc999: who will throw the `UnavailableException`? not clear to me why there will be no progress after this exception is thrown.

ggezer: Nevermind. https://github.com/Alluxio/alluxio/pull/10489 will render this fix redundant.

PR #10447, opened 2019-11-12
Clean up block master service handler

calvinjia:
Removes some unnecessary typing and also fixes the name of some methods so we collect the correct metrics.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterClientServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterClientServiceHandler.java
@@ -64,8 +64,9 @@ public final class BlockMasterClientServiceHandler
       StreamObserver<GetBlockInfoPResponse> responseObserver) {
     long blockId = request.getBlockId();
     GetBlockInfoPOptions options = request.getOptions();
-    RpcUtils.call(LOG, () -> GetBlockInfoPResponse.newBuilder()
-        .setBlockInfo(GrpcUtils.toProto(mBlockMaster.getBlockInfo(blockId))).build(),
+    RpcUtils.call(LOG,
+        () -> GetBlockInfoPResponse.newBuilder()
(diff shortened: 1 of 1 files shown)


calvinjia:
@ZacBlanco Good call, I've updated the format for the methods which do not require a multi-line closure. For the multi-line one I kept it as is since I think its more readable.

core/server/master/src/main/java/alluxio/master/block/BlockMasterClientServiceHandler.java
```diff
@@ -62,113 +62,95 @@ public BlockMasterClientServiceHandler(BlockMaster blockMaster) {
...
-
-    RpcUtils.call(LOG, (RpcUtils.RpcCallableThrowsIOException<GetBlockInfoPResponse>) () -> {
-
-      return GetBlockInfoPResponse.newBuilder()
-          .setBlockInfo(GrpcUtils.toProto(mBlockMaster.getBlockInfo(blockId))).build();
-    }, "CreateFile", "blockId=%s, options=%s", responseObserver, blockId, options);
+    RpcUtils.call(LOG,
+        () -> GetBlockInfoPResponse.newBuilder()
+            .setBlockInfo(GrpcUtils.toProto(mBlockMaster.getBlockInfo(blockId))).build(),
+        "GetBlockInfo", "blockId=%s, options=%s", responseObserver, blockId, options);
```
apc999: getBlockInfo?

calvinjia: I changed it all to capitalized to be consistent with our other logging.

apc999:

apc999:
@calvinjia I view this actually a bug fix .  do we target this to branch 2.1?

calvinjia:
@apc999 that makes sense

calvinjia:

PR #10457, opened 2019-11-12
Update test docker scripts to support sonar reporting

gpang:


gpang:
Jenkins, test this please

dev/jenkins/run_docker.sh
```diff
@@ -38,6 +38,13 @@ function main {
     run_args+=" -e ALLUXIO_GIT_CLEAN=true"
   fi
 
+  if [ -n "${ALLUXIO_SONAR_ARGS}" ]
+  then
+    # write out to a file, in case there are spaces in the args
+    echo "ALLUXIO_SONAR_ARGS=${ALLUXIO_SONAR_ARGS}" > /tmp/.env
```
ZacBlanco: you probably want this?

```
     echo "ALLUXIO_SONAR_ARGS=\"${ALLUXIO_SONAR_ARGS}\"" > /tmp/.env
```

dev/jenkins/build.sh
```diff
@@ -30,3 +30,8 @@ then
   git clean -fdx
 fi
 mvn -Duser.home=/home/jenkins -T 4C clean install -Pdeveloper -Dmaven.javadoc.skip -Dsurefire.forkCount=${ALLUXIO_BUILD_FORKCOUNT} $@
+
+if [ -n "${ALLUXIO_SONAR_ARGS}" ]
```
ZacBlanco: semicolon?
```suggestion
if [ -n "${ALLUXIO_SONAR_ARGS}" ];
```

dev/jenkins/run_docker.sh
gpang: That doesn't seem to work. The quotes get included, which is not what I want. I think Docker already takes care if there are spaces in the values.

dev/jenkins/build.sh
gpang: I just copied whatever we do in this file and the other file, and none of them use the semicolon.

ZacBlanco: It might be if the `fi` is on the same line

gpang:

PR #10461, opened 2019-11-13
Use different session IDs for async cache requests

apc999:
- Differentiate `ASYNC_CACHE_WORKER_SESSION_ID` with `ASYNC_CACHE_UFS_SESSION_ID`: the former session is caching from a remote worker and the latter session is caching from UFS
- The goal of this PR is to enable more advanced allocator which can leverage the information: caching from UFS or caching for remote worker to make subtle decisions.

### DIFF
--- a/core/server/worker/src/main/java/alluxio/Sessions.java
+++ b/core/server/worker/src/main/java/alluxio/Sessions.java
@@ -33,12 +33,10 @@ import javax.annotation.concurrent.ThreadSafe;
 public final class Sessions {
   private static final Logger LOG = LoggerFactory.getLogger(Sessions.class);
 
-  public static final int DATASERVER_SESSION_ID = -1;
-  public static final int CHECKPOINT_SESSION_ID = -2;
   public static final int MIGRATE_DATA_SESSION_ID = -3;
   public static final int MASTER_COMMAND_SESSION_ID = -4;
-  public static final int ACCESS_BLOCK_SESSION_ID = -5;
-  public static final int ASYNC_CACHE_SESSION_ID = -7;
+  public static final int ASYNC_CACHE_WORKER_SESSION_ID = -7;
+  public static final int ASYNC_CACHE_UFS_SESSION_ID = -8;
 
   // internal session id base should be smaller than all predefined session ids
   public static final long INTERNAL_SESSION_ID_BASE = -8;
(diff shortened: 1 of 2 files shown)


core/server/worker/src/main/java/alluxio/Sessions.java
```diff
@@ -33,12 +33,10 @@
 public final class Sessions {
   private static final Logger LOG = LoggerFactory.getLogger(Sessions.class);
 
-  public static final int DATASERVER_SESSION_ID = -1;
-  public static final int CHECKPOINT_SESSION_ID = -2;
   public static final int MIGRATE_DATA_SESSION_ID = -3;
   public static final int MASTER_COMMAND_SESSION_ID = -4;
-  public static final int ACCESS_BLOCK_SESSION_ID = -5;
-  public static final int ASYNC_CACHE_SESSION_ID = -7;
+  public static final int ASYNC_CACHE_WORKER_SESSION_ID = -7;
```
gpang: are these changes backwards compatible?

apc999: Yes, this is backwards compatible

apc999:
@gpang I split this PR into multiple smaller PRs. PTAL

core/server/worker/src/main/java/alluxio/worker/block/AsyncCacheRequestManager.java
```diff
@@ -91,8 +91,12 @@ public void submitRequest(AsyncCacheRequest request) {
       mAsyncCacheExecutor.submit(() -> {
         boolean result = false;
         try {
+          boolean isSourceLocal = mLocalWorkerHostname.equals(request.getSourceHost());
```
gpang: How does this distinguish if the block is coming from an alluxio worker, or the ufs?

It looks like it is checking the source of the request of with the local worker hostname, and I'm not sure how if it is UFS or not can be determined by that?

apc999: discussed offline

gpang: O, seems like this code already existed before, and was just moved up to this location.

apc999:

PR #10463, opened 2019-11-13
Use block size to request cache space

apc999:


apc999:
@calvinjia this PR changes the initial tmp block size on caching UFS blocks. I'm thinking that given the current default block size is much smaller (256MB->64MB), shall we actually just request the default block size on both creating new blocks in addition to caching blocks from UFS or remote workers?

calvinjia:
@apc999 I think for new blocks, it should be `min(10 MB, blocksize)`. For making the default reservation equal to block size, I think that requires some more consideration.

ZacBlanco:
@calvinjia @apc999 is this PR still on the table? If not, can we close it?

apc999:
@calvinjia  I actually think this PR is still valid at least for UFS blocks (which we know the size in ahead). This can strictly reduce the overhead to pressure worker at least from UFS reads.

PR #10466, opened 2019-11-13
Fix logging calls with wrong arguments

apc999:


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3241,9 +3241,9 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
         executorService.invokeAll(callables);
       }
     } catch (InvalidPathException e) {
-      LOG.warn("InvalidPathException during active sync {}", e);
+      LOG.warn("InvalidPathException during active sync {}", e.getMessage());
```
calvinjia: `e.toString` is better than `e.getMessage`

apc999: Oh, I was following guide here: https://docs.alluxio.io/os/user/stable/en/contributor/Code-Conventions.html?q=getMessage#warn-log-level. Do we want to update the guide too?

calvinjia: Yeah we should. The issue with `.getMessage` is sometimes the message is not populated well, and `.toString` is generally just the exception type + message.

apc999:
@calvinjia  PTAL

calvinjia:

apc999:

PR #10475, opened 2019-11-14
Optimize FileOutStream integration tests

ZacBlanco:
This improves and drops the times for a number of tests which
originally inherited from the AbstractFileOutStreamIntegrationTest
class.

Instead of starting a new cluster for each test, we re-use the same
cluster. Some tests needed more modifications to continue passing.

Depending on the number of tests per file, improvements in test
runtime ranged anywhere from 20%-1000%

### DIFF
--- a/tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
+++ b/tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
@@ -11,28 +11,19 @@
 
 package alluxio.client.fs;
 
-import alluxio.AlluxioURI;
-import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileOutStream;
 import alluxio.client.file.FileSystem;
-import alluxio.client.file.URIStatus;
 import alluxio.conf.PropertyKey;
-import alluxio.conf.ServerConfiguration;
-import alluxio.grpc.CreateFilePOptions;
-import alluxio.grpc.OpenFilePOptions;
-import alluxio.grpc.ReadPType;
 import alluxio.master.LocalAlluxioJobCluster;
 import alluxio.testutils.BaseIntegrationTest;
 import alluxio.testutils.LocalAlluxioClusterResource;
-import alluxio.underfs.UnderFileSystem;
-import alluxio.util.io.BufferUtils;
 
 import org.junit.After;
-import org.junit.Assert;
+import org.junit.AfterClass;
 import org.junit.Before;
-import org.junit.Rule;
-
-import java.io.InputStream;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.rules.TestRule;
 
 /**
  * Abstract classes for all integration tests of {@link FileOutStream}.
@@ -41,27 +32,44 @@ public abstract class AbstractFileOutStreamIntegrationTest extends BaseIntegrati
   protected static final int MIN_LEN = 0;
   protected static final int MAX_LEN = 255;
   protected static final int DELTA = 32;
-  protected static final int BUFFER_BYTES = 100;
-  protected LocalAlluxioJobCluster mLocalAlluxioJobCluster;
-  protected static final int BLOCK_SIZE_BYTES = 1000;
+  protected static final int BUFFER_BYTES = 512;
+  protected static final int BLOCK_SIZE_BYTES = 1024;
+  protected static LocalAlluxioJobCluster sLocalAlluxioJobCluster;
 
-  @Rule
-  public LocalAlluxioClusterResource mLocalAlluxioClusterResource =
+  @ClassRule
+  public static LocalAlluxioClusterResource sLocalAlluxioClusterResource =
       buildLocalAlluxioClusterResource();
 
-  protected FileSystem mFileSystem = null;
+  @ClassRule
+  public static TestRule sResetRule = sLocalAlluxioClusterResource.getResetResource();
+
+  protected static FileSystem sFileSystem = null;
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    sLocalAlluxioJobCluster = new LocalAlluxioJobCluster();
+    sLocalAlluxioJobCluster.setProperty(PropertyKey.JOB_MASTER_WORKER_HEARTBEAT_INTERVAL, "25ms");
+    sLocalAlluxioJobCluster.start();
+    sFileSystem = sLocalAlluxioClusterResource.get().getClient();
+  }
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+    if (sLocalAlluxioJobCluster != null) {
+      sLocalAlluxioJobCluster.stop();
+    }
+    sFileSystem.close();
+  }
 
   @Before
   public void before() throws Exception {
-    mLocalAlluxioJobCluster = new LocalAlluxioJobCluster();
-    mLocalAlluxioJobCluster.start();
-    mFileSystem = mLocalAlluxioClusterResource.get().getClient();
+    sFileSystem = sLocalAlluxioClusterResource.get().getClient();
   }
 
   @After
   public void after() throws Exception {
-    if (mLocalAlluxioJobCluster != null) {
-      mLocalAlluxioJobCluster.stop();
+    if (sFileSystem != null) {
+      sFileSystem.close();
     }
   }
 
@@ -69,124 +77,18 @@ public abstract class AbstractFileOutStreamIntegrationTest extends BaseIntegrati
    * Override this method in a test in order to customize the {@link LocalAlluxioClusterResource}.
    * @param resource an AlluxioClusterResource builder
    */
-  protected void customizeClusterResource(LocalAlluxioClusterResource.Builder resource) {
+  protected static void customizeClusterResource(LocalAlluxioClusterResource.Builder resource) {
     resource.setProperty(PropertyKey.USER_FILE_BUFFER_BYTES, BUFFER_BYTES)
         .setProperty(PropertyKey.USER_FILE_REPLICATION_DURABLE, 1)
+        .setProperty(PropertyKey.MASTER_PERSISTENCE_SCHEDULER_INTERVAL_MS, "100ms")
+        .setProperty(PropertyKey.MASTER_PERSISTENCE_CHECKER_INTERVAL_MS, "100ms")
+        .setProperty(PropertyKey.WORKER_TIERED_STORE_RESERVER_INTERVAL_MS, "100ms")
         .setProperty(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, BLOCK_SIZE_BYTES);
   }
 
-  private LocalAlluxioClusterResource buildLocalAlluxioClusterResource() {
+  private static LocalAlluxioClusterResource buildLocalAlluxioClusterResource() {
     LocalAlluxioClusterResource.Builder resource = new LocalAlluxioClusterResource.Builder();
     customizeClusterResource(resource);
     return resource.build();
-  }
-
-  /**
-   * Helper to write an Alluxio file with stream of bytes of increasing byte value.
-   *
-   * @param filePath path of the tmp file
-   * @param fileLen length of the file
-   * @param op options to create file
-   */
-  protected void writeIncreasingBytesToFile(AlluxioURI filePath, int fileLen, CreateFilePOptions op)
-      throws Exception {
-    writeIncreasingBytesToFile(mFileSystem, filePath, fileLen, op);
-  }
-
-  /**
-   * Helper to write an Alluxio file with stream of bytes of increasing byte value.
-   *
-   * @param fs the FileSystemClient to use
-   * @param filePath path of the tmp file
-   * @param fileLen length of the file
-   * @param op options to create file
-   */
-  protected void writeIncreasingBytesToFile(FileSystem fs, AlluxioURI filePath, int fileLen,
-      CreateFilePOptions op)
-      throws Exception {
-    try (FileOutStream os = fs.createFile(filePath, op)) {
-      for (int k = 0; k < fileLen; k++) {
-        os.write((byte) k);
-      }
-    }
-  }
-
-  /**
-   * Helper to write an Alluxio file with one increasing byte array, using a single
-   * {@link FileOutStream#write(byte[], int, int)} invocation.
-   *
-   * @param filePath path of the tmp file
-   * @param fileLen length of the file
-   * @param op options to create file
-   */
-  protected void writeIncreasingByteArrayToFile(AlluxioURI filePath, int fileLen,
-      CreateFilePOptions op) throws Exception {
-    try (FileOutStream os = mFileSystem.createFile(filePath, op)) {
-      os.write(BufferUtils.getIncreasingByteArray(fileLen));
-    }
-  }
-
-  /**
-   * Helper to write an Alluxio file with one increasing byte array, but using two separate
-   * {@link FileOutStream#write(byte[], int, int)} invocations.
-   *
-   * @param filePath path of the tmp file
-   * @param fileLen length of the file
-   * @param op options to create file
-   */
-  protected void writeTwoIncreasingByteArraysToFile(AlluxioURI filePath, int fileLen,
-      CreateFilePOptions op) throws Exception {
-    try (FileOutStream os = mFileSystem.createFile(filePath, op)) {
-      int len1 = fileLen / 2;
-      int len2 = fileLen - len1;
-      os.write(BufferUtils.getIncreasingByteArray(0, len1), 0, len1);
-      os.write(BufferUtils.getIncreasingByteArray(len1, len2), 0, len2);
-    }
-  }
-
-  /**
-   * Checks the given file exists in Alluxio storage and expects its content to be an increasing
-   * array of the given length.
-   *
-   * @param filePath path of the tmp file
-   * @param fileLen length of the file
-   */
-  protected void checkFileInAlluxio(AlluxioURI filePath, int fileLen) throws Exception {
-    URIStatus status = mFileSystem.getStatus(filePath);
-    Assert.assertEquals(fileLen, status.getLength());
-    try (FileInStream is = mFileSystem.openFile(filePath,
-        OpenFilePOptions.newBuilder().setReadType(ReadPType.NO_CACHE).build())) {
-      byte[] res = new byte[(int) status.getLength()];
-      Assert.assertEquals((int) status.getLength(), is.read(res));
-      Assert.assertTrue(BufferUtils.equalIncreasingByteArray(fileLen, res));
-    }
-  }
-
-  /**
-   * Checks the given file exists in under storage and expects its content to be an increasing
-   * array of the given length.
-   *
-   * @param filePath path of the tmp file
-   * @param fileLen length of the file
-   */
-  protected void checkFileInUnderStorage(AlluxioURI filePath, int fileLen) throws Exception {
-    URIStatus status = mFileSystem.getStatus(filePath);
-    String checkpointPath = status.getUfsPath();
-    UnderFileSystem ufs = UnderFileSystem.Factory.create(checkpointPath,
-        ServerConfiguration.global());
-
-    try (InputStream is = ufs.open(checkpointPath)) {
-      byte[] res = new byte[(int) status.getLength()];
-      int totalBytesRead = 0;
-      while (true) {
-        int bytesRead = is.read(res, totalBytesRead, res.length - totalBytesRead);
-        if (bytesRead <= 0) {
-          break;
-        }
-        totalBytesRead += bytesRead;
-      }
-      Assert.assertEquals((int) status.getLength(), totalBytesRead);
-      Assert.assertTrue(BufferUtils.equalIncreasingByteArray(fileLen, res));
-    }
   }
 }
--- a/tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
+++ b/tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
@@ -23,6 +23,8 @@ import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.rules.TestRule;
 
 /**
  * Abstract classes for all integration tests of {@link FileOutStream}.
@@ -38,6 +40,9 @@ public abstract class AbstractFileOutStreamIntegrationTest extends BaseIntegrati
   @ClassRule
   public static LocalAlluxioClusterResource sLocalAlluxioClusterResource =
       buildLocalAlluxioClusterResource();
+
+  @Rule
+  public TestRule mTestRule = sLocalAlluxioClusterResource.getResetResource();
 
   protected FileSystem mFileSystem = null;
 
@@ -67,11 +72,8 @@ public abstract class AbstractFileOutStreamIntegrationTest extends BaseIntegrati
     }
   }
 
-  /**
-   * Override this method in a test in order to customize the {@link LocalAlluxioClusterResource}.
-   * @param resource an AlluxioClusterResource builder
-   */
-  protected static void customizeClusterResource(LocalAlluxioClusterResource.Builder resource) {
+  private static LocalAlluxioClusterResource buildLocalAlluxioClusterResource() {
+    LocalAlluxioClusterResource.Builder resource = new LocalAlluxioClusterResource.Builder();
     resource.setProperty(PropertyKey.USER_FILE_BUFFER_BYTES, BUFFER_BYTES)
         .setProperty(PropertyKey.USER_FILE_REPLICATION_DURABLE, 1)
         .setProperty(PropertyKey.MASTER_PERSISTENCE_SCHEDULER_INTERVAL_MS, "10ms")
@@ -79,12 +81,8 @@ public abstract class AbstractFileOutStreamIntegrationTest extends BaseIntegrati
         .setProperty(PropertyKey.WORKER_TIERED_STORE_RESERVER_INTERVAL_MS, "10ms")
         .setProperty(PropertyKey.MASTER_WORKER_HEARTBEAT_INTERVAL, "10ms")
         .setProperty(PropertyKey.WORKER_BLOCK_HEARTBEAT_INTERVAL_MS, "10ms")
-        .setProperty(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, BLOCK_SIZE_BYTES);
-  }
-
-  private static LocalAlluxioClusterResource buildLocalAlluxioClusterResource() {
-    LocalAlluxioClusterResource.Builder resource = new LocalAlluxioClusterResource.Builder();
-    customizeClusterResource(resource);
+        .setProperty(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, BLOCK_SIZE_BYTES)
+    ;
     return resource.build();
   }
 }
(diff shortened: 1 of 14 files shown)


tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
```diff
@@ -24,40 +24,88 @@
...
+  }
+
+  @Before
+  public void before() throws Exception {
+    mFileSystem = sLocalAlluxioClusterResource.get().getClient();
+  }
+
+  @After
+  public void after() throws Exception {
+    mFileSystem.close();
```
gpang: are we supposed to close the clients we get from the local alluxio cluster?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
```diff
@@ -69,41 +117,45 @@ public void asyncWrite() throws Exception {
...
-    // check the file is completed but not persisted
-    URIStatus srcStatus = mFileSystem.getStatus(srcPath);
-    assertEquals(PersistenceState.TO_BE_PERSISTED.toString(), srcStatus.getPersistenceState());
-    Assert.assertTrue(srcStatus.isCompleted());
-
-    mFileSystem.rename(srcPath, dstPath);
-    CommonUtils.sleepMs(1);
-    checkPersistStateAndWaitForPersist(dstPath, 2);
+    ServerConfiguration.set(PropertyKey.USER_FILE_PERSIST_ON_RENAME, "true");
+    try (FileSystem fs = sLocalAlluxioClusterResource.get().getClient()) {
```
gpang: why can't we use `mFileSystem`?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
```diff
@@ -113,16 +165,18 @@ public void asyncWritePersistWithNoAutoPersist() throws Exception {
...
 
   @Test
   public void asyncWriteWithPersistWaitTime() throws Exception {
     AlluxioURI filePath = new AlluxioURI(PathUtils.uniqPath());
-    createTwoBytesFile(filePath, 2000);
+    createTwoBytesFile(mFileSystem, filePath, 2000);
 
-    CommonUtils.sleepMs(1000);
-    checkPersistStateAndWaitForPersist(filePath, 2);
+    CommonUtils.sleepMs(500);
```
gpang: if we shorten all the intervals for the cluster, can we shorten this?

Or, does the last line do the wait anyways? If so, does this line do anything?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
```diff
@@ -204,103 +218,55 @@ public void asyncPersistNoAutoPersistNoEvict() throws Exception {
...
-
-    FileSystemUtils.persistAndWait(fs, new AlluxioURI(p1), 0);
-    fstat = fs.listStatus(new AlluxioURI(p1)).get(0);
-
-    assertTrue(fstat.isPersisted());
-    assertEquals(0, mLocalAlluxioClusterResource.get().getLocalAlluxioMaster().getMasterProcess()
-        .getMaster(FileSystemMaster.class)
-        .getPinIdList().size());
-    assertTrue(getUsedWorkerSpace() < getClusterCapacity());
+    try (FileSystem fs = sLocalAlluxioClusterResource.get().getClient()) {
```
gpang: why not use `mFileSystem`?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
```diff
@@ -55,269 +56,282 @@
...
     // free for non-persisted file is no-op
     Assert.assertEquals(100, status.getInMemoryPercentage());
   }
 
   @Test
   public void freeAfterFilePersisted() throws Exception {
     URIStatus status = createAsyncFile();
-    IntegrationTestUtils.waitForPersist(mLocalAlluxioClusterResource, mUri);
-    mFileSystem.free(mUri);
+    System.out.println(status);
```
gpang: debugging?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
```diff
@@ -55,269 +56,282 @@
...
+    IntegrationTestUtils.waitForPersist(sLocalAlluxioClusterResource, mUri);
+    System.out.println("Finished waiting for persist");
```
gpang: debugging?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
```diff
@@ -55,269 +56,282 @@
...
+    sFileSystem.free(mUri);
+    System.out.println("Freed file");
```
gpang: debugging?

tests/src/test/java/alluxio/client/fs/FileOutStreamIntegrationTest.java
```diff
@@ -122,46 +124,45 @@ public void writeTwoByteArrays() throws Exception {
...
   @Test
   public void longWrite() throws Exception {
     AlluxioURI filePath = new AlluxioURI(PathUtils.uniqPath());
     final int length = 2;
-    try (FileOutStream os = mFileSystem.createFile(filePath, CreateFilePOptions.newBuilder()
+    try (FileOutStream os = sFileSystem.createFile(filePath, CreateFilePOptions.newBuilder()
         .setWriteType(mWriteType.toProto()).setRecursive(true).build())) {
       os.write((byte) 0);
-      Thread.sleep(Constants.SECOND_MS * 2);
+      Thread.sleep(ServerConfiguration.getMs(PropertyKey.WORKER_BLOCK_HEARTBEAT_INTERVAL_MS));
```
gpang: does this have to have a x2 ?

tests/src/test/java/alluxio/client/fs/FileOutStreamTestUtils.java
```diff
@@ -0,0 +1,131 @@
...
+import alluxio.grpc.OpenFilePOptions;
+import alluxio.grpc.ReadPType;
+import alluxio.underfs.UnderFileSystem;
+import alluxio.util.io.BufferUtils;
+
+import org.junit.Assert;
+
+import java.io.InputStream;
+
+public class FileOutStreamTestUtils {
```
gpang: for these methods, are they basically copied from other places?

tests/src/test/java/alluxio/client/fs/UfsFallbackFileOutStreamIntegrationTest.java
```diff
@@ -45,20 +53,50 @@
...
   @ClassRule
   public static ManuallyScheduleHeartbeat sManuallyScheduleEviction =
       new ManuallyScheduleHeartbeat(HeartbeatContext.WORKER_SPACE_RESERVER);
 
-  protected static final int WORKER_MEMORY_SIZE = 1500;
-  protected static final int BUFFER_BYTES = 100;
+  public static LocalAlluxioClusterResource sLocalAlluxioClusterResource =
+      new LocalAlluxioClusterResource.Builder()
+          .setProperty(PropertyKey.WORKER_MEMORY_SIZE, WORKER_MEMORY_SIZE)
+          .setProperty(PropertyKey.MASTER_PERSISTENCE_SCHEDULER_INTERVAL_MS, "100ms")
```
gpang: can these be lower, and if so, does that speed anything up?

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
```diff
@@ -12,44 +12,121 @@
...
+   *
+   * @param resource the local alluxio cluster resource
+   * @param path the path to check
+   * @param length expected size of the file
+   */
+  public static void checkPersistStateAndWaitForPersist(LocalAlluxioClusterResource resource,
+      FileSystem fs, AlluxioURI path, int length) throws Exception {
+    // check the file is completed but not persisted
+    URIStatus status = fs.getStatus(path);
+    assertEquals(PersistenceState.TO_BE_PERSISTED.toString(), status.getPersistenceState());
```
gpang: is it sometimes possible that the persist is so fast, that the TO_BE_PERSISTED state is skipped?

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
```diff
@@ -117,18 +195,24 @@ public static void waitForBlocksToBeFreed(final BlockWorker bw, final Long... bl
       throws TimeoutException {
     try {
       // Execute 1st heartbeat from worker.
-      HeartbeatScheduler.execute(HeartbeatContext.WORKER_BLOCK_SYNC);
+      Whitebox.getInternalState(bw, BlockMasterSync.class).heartbeat();
```
gpang: why do we have to whitebox these?

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
```diff
@@ -117,18 +195,24 @@ public static void waitForBlocksToBeFreed(final BlockWorker bw, final Long... bl
...
-        List<Long> blocksToRemove = Whitebox.getInternalState(reporter, "mRemovedBlocks");
-        return blocksToRemove.containsAll(Arrays.asList(blockIds));
+        if (blockIds.length == 0) {
+          return true;
+        }
+        Set<Long> blocks = bw.getBlockStore().getBlockStoreMetaFull()
+            .getBlockListByStorageLocation().values().stream().flatMap(List::stream)
+            .collect(Collectors.toSet());
+        List<Long> bids = Lists.newArrayList(blockIds);
+        return bids.stream().map(id -> !blocks.contains(id)).reduce((s1, s2) -> s1 && s2)
```
gpang: can we not do a set intersection instead of this?

tests/src/test/java/alluxio/testutils/PersistenceTestUtils.java
```diff
@@ -92,9 +93,13 @@ public static void pauseScheduler(LocalAlluxioClusterResource resource) {
...
     FileSystemMaster nestedMaster = getFileSystemMaster(resource);
-    BlackHoleMap<Long, ExponentialTimer> persistRequests =
-        Whitebox.getInternalState(nestedMaster, "mPersistRequests");
-    Whitebox.setInternalState(nestedMaster, "mPersistRequests", persistRequests.getInnerMap());
+    try {
+      BlackHoleMap<Long, ExponentialTimer> persistRequests =
+          Whitebox.getInternalState(nestedMaster, "mPersistRequests");
+      Whitebox.setInternalState(nestedMaster, "mPersistRequests", persistRequests.getInnerMap());
+    } catch (ClassCastException e) {
+      Whitebox.setInternalState(nestedMaster, "mPersistRequests", new ConcurrentHashMap<>());
```
gpang: I'm not sure what is going on here. Can you add a comment?

ZacBlanco:
jenkins, test this please

tests/src/test/java/alluxio/testutils/LocalAlluxioClusterResource.java
```diff
@@ -284,6 +286,17 @@ public void evaluate() throws Throwable {
             } else {
               resetCluster(fsm);
             }
+
+            // Reset any properties that have changed
```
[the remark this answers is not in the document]
ZacBlanco: Unfortunately, it's not quite the same as using the `LocalAlluxioResource.Config` annotation, but for any test that sets `ServerConfiguration` duing one of the tests, it's desirable to put the config after the test back to the state it was in before the test failed.

There's no good way to track the set of changes from a specific time, so I just take a copy of the config before

We can't just call `ServerConfiguration.reset()` because that will unset the port numbers for the Alluxio cluster.

tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
```diff
@@ -41,152 +32,65 @@
...
 
   /**
    * Override this method in a test in order to customize the {@link LocalAlluxioClusterResource}.
    * @param resource an AlluxioClusterResource builder
    */
-  protected void customizeClusterResource(LocalAlluxioClusterResource.Builder resource) {
+  protected static void customizeClusterResource(LocalAlluxioClusterResource.Builder resource) {
     resource.setProperty(PropertyKey.USER_FILE_BUFFER_BYTES, BUFFER_BYTES)
         .setProperty(PropertyKey.USER_FILE_REPLICATION_DURABLE, 1)
+        .setProperty(PropertyKey.MASTER_PERSISTENCE_SCHEDULER_INTERVAL_MS, "100ms")
```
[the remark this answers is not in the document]
ZacBlanco: I can try making it lower? I'm not sure how it will affect the tests

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
ZacBlanco: yes, otherwise it will result in resource leaks

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
ZacBlanco: We could, but all files created here need to be created with the `PERSIST_ON_RENAME` option, so it's easier to set it within `ServerConfiguration` and then create a client based on it so that all creates default to using it.

Also, mainly because the test uses the `createTwoByteFile` method which didn't expose the options previously. So, I just created a new client with the default option instead.

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
ZacBlanco: Not sure why the original wait time was 1000ms. I just dropped to 500 because all the heartbeat times are slightly less than that. I don't know what the lower bound is. I think it can just be removed?

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
ZacBlanco: There are tests which set some user configuration beforehand. I don't use `mFileSystem` because I want the client used to pick up any changes to configuration properties

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
ZacBlanco: whoops

tests/src/test/java/alluxio/client/fs/FileOutStreamIntegrationTest.java
ZacBlanco: Not sure. I can put it back to be safe? 

The original idea around this test was to make sure that the session id doesn't change.

tests/src/test/java/alluxio/client/fs/FileOutStreamTestUtils.java
ZacBlanco: yes I just extracted them from `FileOutStreamAsyncWrite` test

tests/src/test/java/alluxio/client/fs/UfsFallbackFileOutStreamIntegrationTest.java
ZacBlanco: I can try making them lower

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
ZacBlanco: Because the heartbeat threads have had their time interval reduced, calling `HeartbeatScheduler.execute` causes almost guaranteed deadlock.

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
ZacBlanco: good call

tests/src/test/java/alluxio/testutils/PersistenceTestUtils.java
ZacBlanco: updated

tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
```diff
@@ -41,152 +32,65 @@
...
+  protected static LocalAlluxioJobCluster sLocalAlluxioJobCluster;
 
-  @Rule
-  public LocalAlluxioClusterResource mLocalAlluxioClusterResource =
+  public static LocalAlluxioClusterResource sLocalAlluxioClusterResource =
       buildLocalAlluxioClusterResource();
 
-  protected FileSystem mFileSystem = null;
+  @ClassRule
+  public static RuleChain sRuleChain = RuleChain.outerRule(sLocalAlluxioClusterResource)
```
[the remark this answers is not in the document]
ZacBlanco: updated

tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
```diff
@@ -41,152 +32,65 @@
...
+      .around(sLocalAlluxioClusterResource.getResetResource());
+
+  protected static FileSystem sFileSystem = null;
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    sLocalAlluxioJobCluster = new LocalAlluxioJobCluster();
+    sLocalAlluxioJobCluster.setProperty(PropertyKey.JOB_MASTER_WORKER_HEARTBEAT_INTERVAL, "25ms");
+    sLocalAlluxioJobCluster.start();
+    sFileSystem = sLocalAlluxioClusterResource.get().getClient();
```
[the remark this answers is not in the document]
ZacBlanco: updated

tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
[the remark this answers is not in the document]
ZacBlanco: cahnged to 10ms

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteIntegrationTest.java
ZacBlanco: i removed this line

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
ZacBlanco: updated

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
ZacBlanco: removed

tests/src/test/java/alluxio/client/fs/FileOutStreamIntegrationTest.java
ZacBlanco: added the *2 back

tests/src/test/java/alluxio/client/fs/FileOutStreamAsyncWriteJobIntegrationTest.java
ZacBlanco: removed

tests/src/test/java/alluxio/client/fs/UfsFallbackFileOutStreamIntegrationTest.java
ZacBlanco: changed to 10ms

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
```diff
@@ -117,18 +195,21 @@ public static void waitForBlocksToBeFreed(final BlockWorker bw, final Long... bl
...
       CommonUtils.waitFor("blocks to be removed", () -> {
-        BlockHeartbeatReporter reporter = Whitebox.getInternalState(bw, "mHeartbeatReporter");
-        List<Long> blocksToRemove = Whitebox.getInternalState(reporter, "mRemovedBlocks");
-        return blocksToRemove.containsAll(Arrays.asList(blockIds));
+
+        Set<Long> blocks = bw.getBlockStore().getBlockStoreMetaFull()
+            .getBlockListByStorageLocation().values().stream().flatMap(List::stream)
+            .collect(Collectors.toSet());
+        blocks.retainAll(Sets.newHashSet(blockIds));
+        return blocks.size() <= 0;
```
gpang: is there an `isEmpty()` call?

tests/src/test/java/alluxio/testutils/PersistenceTestUtils.java
```diff
@@ -92,9 +94,14 @@ public static void pauseScheduler(LocalAlluxioClusterResource resource) {
...
-    BlackHoleMap<Long, ExponentialTimer> persistRequests =
-        Whitebox.getInternalState(nestedMaster, "mPersistRequests");
-    Whitebox.setInternalState(nestedMaster, "mPersistRequests", persistRequests.getInnerMap());
+    try {
+      BlackHoleMap<Long, ExponentialTimer> persistRequests =
+          Whitebox.getInternalState(nestedMaster, "mPersistRequests");
+      Whitebox.setInternalState(nestedMaster, "mPersistRequests", persistRequests.getInnerMap());
+    } catch (ClassCastException e) {
+      // This will occur is resume is called without first being paused
+      LOG.warn("Call to resumeScheduler failed: ", e);
```
gpang: If this happens, does this mean the test was written incorrectly? Should we just throw a real exception to kill the test?

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
gpang: hrmmm, how does that work? I thought if we manually scheduled heartbeats, the time intervals don't matter? We aren't using `ManuallyScheduleHeartbeat`?

tests/src/test/java/alluxio/testutils/PersistenceTestUtils.java
ZacBlanco: Not necessarily - if a test errors between pausing and resuming the scheduler, the test may not necessarily be written incorrectly.

If an Alluxio code changes causes an exception to be thrown in between pausing and resuming it might be possible for pause/resume to be called twice depending on how the next test in the suite is run.

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
ZacBlanco: 🤦‍♂️

tests/src/test/java/alluxio/client/fs/AbstractFileOutStreamIntegrationTest.java
```diff
@@ -41,152 +31,60 @@
...
-  protected static final int BUFFER_BYTES = 100;
-  protected LocalAlluxioJobCluster mLocalAlluxioJobCluster;
-  protected static final int BLOCK_SIZE_BYTES = 1000;
+  protected static final int BUFFER_BYTES = 512;
+  protected static final int BLOCK_SIZE_BYTES = 1024;
+  protected static LocalAlluxioJobCluster sLocalAlluxioJobCluster;
 
-  @Rule
-  public LocalAlluxioClusterResource mLocalAlluxioClusterResource =
+  @ClassRule
```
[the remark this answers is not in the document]
ZacBlanco: I had some issues before with the test failing to clean up but it might have been due to conflicting rules in the subclasses

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
ZacBlanco: The test where this caused failures was using a manual heartbeat, but I think there may have been other heartbeats other than the BlockMasterSync which took resources and starved/deadlocked this heartbeat thread.

I couldn't figure out the source of the deadlock but the program would hang indefinitely without this change...

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
ZacBlanco: It is possible that is the case. I think this logic was wrong before this change though.

tests/src/test/java/alluxio/testutils/IntegrationTestUtils.java
ZacBlanco: The deadlock occurs in `FileOutStreamAsyncWriteJobIntegrationTest#renameScheduleFree`

tests/src/test/java/alluxio/testutils/LocalAlluxioClusterResource.java
```diff
@@ -268,6 +274,7 @@ public Statement apply(Statement statement, Description description) {
       return new Statement() {
         @Override
         public void evaluate() throws Throwable {
+          AlluxioProperties props = ServerConfiguration.copyProperties();
```
gpang: instead of copying the whole thing, I tried using an annotation, similar to the existing one. This approach might be cleaner and more similar to what existed?
```java
  /**
   * An annotation for test methods that can be used to override any user configuration values. This
   * should not be used with server properties, since they may not take effect if the cluster is
   * a {@code @ClassRule} (single cluster per test file).
   */
  @Retention(RetentionPolicy.RUNTIME)
  public @interface UserConfig {
    String[] confParams() default {};
  }
```

Then, here, we examine the annotation and save the old values and set the new values:
```java
            Annotation configAnnotation = description.getAnnotation(UserConfig.class);
            if (configAnnotation != null) {
              UserConfig config = (UserConfig) configAnnotation;
              mOldConf.clear();
              mOldUnset.clear();
              // Override the configuration parameters with any configuration params
              for (int i = 0; i < config.confParams().length; i += 2) {
                PropertyKey prop = PropertyKey.fromString(config.confParams()[i]);
                if (ServerConfiguration.isSet(prop)) {
                  // save the existing conf value
                  mOldConf.put(prop, ServerConfiguration.get(prop));
                } else {
                  mOldUnset.add(prop);
                }
                // set the new conf value
                ServerConfiguration.set(prop, config.confParams()[i + 1]);
              }
            }
```

and in the `finally`, we set then back:
```java
            if (!mOldConf.isEmpty()) {
              // reset the old conf values
              for (Map.Entry<PropertyKey, String> entry : mOldConf.entrySet()) {
                ServerConfiguration.set(entry.getKey(), entry.getValue());
              }
              mOldConf.clear();
            }
            if (!mOldUnset.isEmpty()) {
              // reset the old unset conf values
              for (PropertyKey prop : mOldUnset) {
                ServerConfiguration.unset(prop);
              }
              mOldUnset.clear();
            }
```

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
One possible flake?
```
alluxio.client.fs.FileOutStreamIntegrationTest.cancelWrite[0](alluxio.client.fs.FileOutStreamIntegrationTest)
  Run 1: FileOutStreamIntegrationTest.cancelWrite:217 expected:<0> but was:<3584>
  Run 2: PASS
```

ZacBlanco:
Interesting. I haven't had it flake on me since I forced the block worker clear the block store. Let me push one more change that could fix it.

gpang:
Jenkins, test this please

gpang:
```
alluxio.client.fs.UfsFallbackFileOutStreamIntegrationTest  Time elapsed: 67.506 sec  <<< ERROR!
java.lang.RuntimeException: Timed out waiting for process Alluxio job master @ 410ec1fa8900:46480 to start
	at alluxio.client.fs.UfsFallbackFileOutStreamIntegrationTest.beforeClass(UfsFallbackFileOutStreamIntegrationTest.java:88)
```

gpang:
Jenkins, test this please

PR #10481, opened 2019-11-18
Improve reporting worker metrics concurrency [WIP]

LuQQiu:
Reduce the synchronized on the whole `workerMetrics` when updating worker metrics through heartbeat.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
+++ b/core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
@@ -95,9 +95,7 @@ public class MetricsStore {
     if (metrics.isEmpty()) {
       return;
     }
-    synchronized (mWorkerMetrics) {
-      putReportedMetrics(mWorkerMetrics, getFullInstanceId(hostname, null), metrics);
-    }
+    putReportedMetrics(mWorkerMetrics, getFullInstanceId(hostname, null), metrics);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -145,21 +140,30 @@ private static void putReportedMetrics(IndexedSet<Metric> metricSet, String inst
...
         Metric oldMetric = metricSet.getFirstByField(FULL_NAME_INDEX, metric.getFullMetricName());
-        double oldVal = oldMetric == null ? 0.0 : oldMetric.getValue();
-        Metric newMetric = new Metric(metric.getInstanceType(), metric.getHostname(),
-            metric.getMetricType(), metric.getName(), oldVal + metric.getValue());
-        for (Map.Entry<String, String> tag : metric.getTags().entrySet()) {
-          newMetric.addTag(tag.getKey(), tag.getValue());
+        if (oldMetric != null) {
+          metric.addValue(metric.getValue());
+        } else {
+          Metric newMetric = new Metric(metric.getInstanceType(), metric.getHostname(),
```
gpang: Why are the metrics updated by creating a new object for every one? That seems very costly...

I feel like we should be able to make Metric atomic and concurrent but using atomic data structures.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -145,21 +140,30 @@ private static void putReportedMetrics(IndexedSet<Metric> metricSet, String inst
...
+            }
+          }
         }
-        metricSet.removeByField(FULL_NAME_INDEX, metric.getFullMetricName());
-        newMetrics.add(newMetric);
       } else {
-        metricSet.removeByField(FULL_NAME_INDEX, metric.getFullMetricName());
-        newMetrics.add(metric);
+        synchronized (metricSet) {
+          metricSet.removeByField(FULL_NAME_INDEX, metric.getFullMetricName());
```
gpang: Do you know why the metrics are updated in this way (delete all metrics, then add all new metrics)? It seems inefficient, and we should probably be able to update existing data structures in an atomic way?

PR #10482, opened 2019-11-19
Fix NPE in UfsFallbackBlockWriteHandler

witgo:
UfsFallbackBlockWriteHandler#createRequestContext does not properly initialize the BlockWriteRequestContext instance

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
@@ -84,7 +84,7 @@ public final class UfsFallbackBlockWriteHandler
   @Override
   protected BlockWriteRequestContext createRequestContext(alluxio.grpc.WriteRequest msg)
       throws Exception {
-    BlockWriteRequestContext context = new BlockWriteRequestContext(msg, FILE_BUFFER_SIZE);
+    BlockWriteRequestContext context = mBlockWriteHandler.createRequestContext(msg);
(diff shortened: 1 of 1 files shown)


core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
```diff
@@ -84,7 +84,7 @@
   @Override
   protected BlockWriteRequestContext createRequestContext(alluxio.grpc.WriteRequest msg)
       throws Exception {
-    BlockWriteRequestContext context = new BlockWriteRequestContext(msg, FILE_BUFFER_SIZE);
+    BlockWriteRequestContext context = mBlockWriteHandler.createRequestContext(msg);
```
gpang: In `BlockWriteHandler#createRequestContext()` it looks like it calls `mWorker.createBlockRemote`, but in this method it can also call `mWorker.createBlockRemote`. Will that be a problem?

Instead, should we just initialize the counters correctly in this method?

witgo: Done

witgo:
ping

gpang:
@ZacBlanco Can you take a look at this PR?

ZacBlanco:
This PR actually uncovers some other issues with our UFS fallback handling - but I think we can address those in a separate PR. This should at least be able to fix the first NPE

gpang:

ZacBlanco:

PR #10483, opened 2019-11-19
Support session level connection management for leader election

ggezer:
Curator has 2 policies for managing zookeeper client state when connection errors happen.
- STANDARD; is finicky and treats leader state as dirty after any connection state change.
- SESSION; is for more stable leadership state as it leverages internal zookeeper session Ids.
Curator comes with STANDARD policy as default. 

This PR makes this setting configurable.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -536,7 +536,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY =
       new Builder(Name.ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY)
-          .setDefaultValue("STANDARD")
+          .setDefaultValue("SESSION")
           .setDescription("Connection error policy defines how errors on zookeeper connections "
               + "to be treated in leader election. "
               + "STANDARD policy treats every connection event as failure."
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -534,6 +534,17 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY =
+      new Builder(Name.ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY)
+          .setDefaultValue("STANDARD")
+          .setDescription("Connection error policy defines how errors on zookeeper connections "
+              + "to be treated in leader election. "
+              + "STANDARD policy treats every connection event as failure."
+              + "SESSION policy relies on zookeeper sessions for judging failures, "
+              + "helping leader to retain its status, as long as its session is protected.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
   /**
    * UFS related properties.
    */
@@ -3860,6 +3871,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     public static final String ZOOKEEPER_LEADER_PATH = "alluxio.zookeeper.leader.path";
     public static final String ZOOKEEPER_SESSION_TIMEOUT = "alluxio.zookeeper.session.timeout";
     public static final String ZOOKEEPER_AUTH_ENABLED = "alluxio.zookeeper.auth.enabled";
+    public static final String ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY =
+        "alluxio.zookeeper.leader.connection.error.policy";
     //
     // UFS related properties
     //
--- a/core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
+++ b/core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
@@ -22,6 +22,7 @@ import org.apache.curator.framework.recipes.leader.LeaderSelector;
 import org.apache.curator.framework.recipes.leader.LeaderSelectorListener;
 import org.apache.curator.framework.recipes.leader.Participant;
 import org.apache.curator.framework.state.ConnectionState;
+import org.apache.curator.framework.state.SessionConnectionStateErrorPolicy;
 import org.apache.curator.retry.ExponentialBackoffRetry;
 import org.apache.zookeeper.CreateMode;
 import org.slf4j.Logger;
@@ -53,6 +54,10 @@ public final class PrimarySelectorClient extends AbstractPrimarySelector
   private String mName;
   /** The address to Zookeeper. */
   private final String mZookeeperAddress;
+  /** The sessionID under which leadership is granted. */
+  private long mLeaderZkSessionId;
+  /** Configured connection error policy for leader election. */
+  private ZookeeperConnectionErrorPolicy mConnectionErrorPolicy;
 
   /**
    * Constructs a new {@link PrimarySelectorClient}.
@@ -69,6 +74,8 @@ public final class PrimarySelectorClient extends AbstractPrimarySelector
     } else {
       mLeaderFolder = leaderPath + AlluxioURI.SEPARATOR;
     }
+    mConnectionErrorPolicy = ServerConfiguration.getEnum(
+        PropertyKey.ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY, ZookeeperConnectionErrorPolicy.class);
 
     // Create a leader selector using the given path for management.
     // All participants in a given leader selection must use the same path.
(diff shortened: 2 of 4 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -534,6 +534,17 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY =
+      new Builder(Name.ZOOKEEPER_LEADER_CONNECTION_ERROR_POLICY)
+          .setDefaultValue("STANDARD")
```
calvinjia: Is there a downside to `SESSION`? Perhaps we should make that default?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -132,7 +139,55 @@ public void start(InetSocketAddress address) throws IOException {
...
   @Override
   public void stateChanged(CuratorFramework client, ConnectionState newState) {
-    setState(State.SECONDARY);
+    // Handle state change.
+    switch (newState) {
+      case CONNECTED:
+      case LOST:
+        setState(State.SECONDARY);
+        break;
+      case SUSPENDED:
```
calvinjia: Not sure if you want to fall through here since we later check on the state.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Only down-side is that it's not tested enough.

ggezer:
Jenkins, test this please

ggezer:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Switched to SESSION as default.

ggezer:
FYI, I run a fault tolerance test against this with no issues.

ggezer:
@gpang PTAL.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -200,4 +260,12 @@ private CuratorFramework getNewCuratorClient() {
     client.start();
     return client;
   }
+
+  /**
+   * Defines supported connection error policies for leader election.
+   */
+  protected enum ZookeeperConnectionErrorPolicy {
+    STANDARD, // Treat each state change as error.
+    SESSION   // Treat state change as error when state is lost.
```
gpang: is `state is lost` supposed to be `session is lost`?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -158,12 +212,14 @@ public void takeLeadership(CuratorFramework client) throws Exception {
...
     LOG.info("{} is now the leader.", mName);
     try {
+      mLeaderZkSessionId = client.getZookeeperClient().getZooKeeper().getSessionId();
       waitForState(State.SECONDARY);
     } finally {
       LOG.warn("{} relinquishing leadership.", mName);
       LOG.info("The current leader is {}", mLeaderSelector.getLeader().getId());
       LOG.info("All participants: {}", mLeaderSelector.getParticipants());
       client.delete().forPath(mLeaderFolder + mName);
+      mLeaderZkSessionId = -1;
```
gpang: can we have a constant for the `-1`, and set it in the constructor?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -158,12 +212,14 @@ public void takeLeadership(CuratorFramework client) throws Exception {
         .forPath(mLeaderFolder + mName);
     LOG.info("{} is now the leader.", mName);
     try {
+      mLeaderZkSessionId = client.getZookeeperClient().getZooKeeper().getSessionId();
```
gpang: Can we log what we think the zk session id is? Would that be useful info?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -158,12 +212,14 @@ public void takeLeadership(CuratorFramework client) throws Exception {
...
       waitForState(State.SECONDARY);
     } finally {
       LOG.warn("{} relinquishing leadership.", mName);
       LOG.info("The current leader is {}", mLeaderSelector.getLeader().getId());
       LOG.info("All participants: {}", mLeaderSelector.getParticipants());
```
gpang: Can we log what we think the zk session id was? Would that be useful info?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -132,7 +139,54 @@ public void start(InetSocketAddress address) throws IOException {
 
   @Override
   public void stateChanged(CuratorFramework client, ConnectionState newState) {
-    setState(State.SECONDARY);
+    // Handle state change.
+    switch (newState) {
```
gpang: This seems difficult to follow and maintain. I don't know if it makes sense to make subclasses, but maybe we can create 2 methods to handle the state change, one for each policy, and just do a single switch here on the policy?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -132,7 +139,54 @@ public void start(InetSocketAddress address) throws IOException {
...
+             * Do a sanity check when reconnected for a selector with "PRIMARY" mode.
+             * This is to ensure that curator reconnected with the same Id.
+             * Hence, guaranteeing Zookeeper state for this instance was preserved.
+             */
+            try {
+              long reconnectSessionId = client.getZookeeperClient().getZooKeeper().getSessionId();
+              if (mLeaderZkSessionId != reconnectSessionId) {
+                LOG.warn(String.format(
+                    "Curator reconnected under a different session. "
+                        + "Old sessionId: %x, New sessionId: %x",
```
gpang: are all session ids typically printed in hex?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -158,12 +212,14 @@ public void takeLeadership(CuratorFramework client) throws Exception {
         .forPath(mLeaderFolder + mName);
     LOG.info("{} is now the leader.", mName);
     try {
+      mLeaderZkSessionId = client.getZookeeperClient().getZooKeeper().getSessionId();
       waitForState(State.SECONDARY);
     } finally {
       LOG.warn("{} relinquishing leadership.", mName);
       LOG.info("The current leader is {}", mLeaderSelector.getLeader().getId());
```
gpang: NIT: For all log messages, would it make sense to also add `mName` to messages for debugging?

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -132,7 +139,54 @@ public void start(InetSocketAddress address) throws IOException {
...
+                    mLeaderZkSessionId, reconnectSessionId));
+                setState(State.SECONDARY);
+              } else {
+                LOG.info(String.format(
+                    "Retaining leader state after zookeeper reconnected with sessionId: %x.",
+                    reconnectSessionId));
+              }
+            } catch (Exception e) {
+              LOG.warn(String
+                  .format("Cannot query session Id after session is reconnected with Id: %x", e));
```
gpang: I think the actual session id is missing from the message. Also, can we standardize by using `sessionId` in the message?

gpang:
Jenkins, test this please

tests/src/test/java/alluxio/server/ft/ZookeeperFailureIntegrationTest.java
```diff
@@ -105,6 +107,63 @@ public void zkFailure() throws Exception {
...
+  }
+
+  /**
+   * Used to get primary master index with retries for when zk server is down.
+   */
+  private int getPrimaryMasterIndex() throws Exception {
+    AtomicInteger primaryIndex = new AtomicInteger();
+    CommonUtils.waitFor("Getting primary master index", () -> {
+      try {
+        primaryIndex.set(mCluster.getPrimaryMasterIndex(30000));
```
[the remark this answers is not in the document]
ggezer: This fails immediately when zk server is restarted. In that case only CommonUtils's wait is effective.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: What it was will be logged as I added a trace for that already.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: Yes. We not printing in hex was creating inconsistency between zk traces.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: Nice catch. There shouldn't be an id here as we couldn't query it.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: I'm not sure it makes sense to include `mName` anywhere except zookeeper path manipulation.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
ggezer: Agreed.

core/server/common/src/main/java/alluxio/master/PrimarySelectorClient.java
```diff
@@ -158,12 +226,15 @@ public void takeLeadership(CuratorFramework client) throws Exception {
         .forPath(mLeaderFolder + mName);
     LOG.info("{} is now the leader.", mName);
     try {
+      mLeaderZkSessionId = client.getZookeeperClient().getZooKeeper().getSessionId();
+      LOG.info("Taken leadership under session Id: {}", mLeaderZkSessionId);
```
gpang: use `sessionId` and use hex format?

ggezer:

ggezer:

gpang:
@ggezer does this have to be cherry-picked to 2.0 and 2.1?

ggezer:

ggezer:

PR #10484, opened 2019-11-19
Remove extraneous client RPC to get block info from master

chcn:
In the FileInfo retrieved from master in openFile, there should be the latest block information, so there is no need to issue an RPC to get the block info again.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
@@ -56,7 +56,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-import javax.annotation.Nonnull;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -180,7 +179,7 @@ public final class AlluxioBlockStore {
    * @param failedWorkers the map of workers address to most recent failure time
    * @return a stream which reads from the beginning of the block
    */
-  public BlockInStream getInStream(@Nonnull BlockInfo info, InStreamOptions options,
+  public BlockInStream getInStream(BlockInfo info, InStreamOptions options,
       Map<WorkerNetAddress, Long> failedWorkers) throws IOException {
     List<BlockLocation> locations = info.getLocations();
     List<BlockWorkerInfo> blockWorkerInfo = Collections.EMPTY_LIST;
--- a/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
@@ -14,18 +14,18 @@ package alluxio.client.block;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 
-import alluxio.client.block.util.BlockLocationUtils;
-import alluxio.collections.Pair;
-import alluxio.conf.PropertyKey;
 import alluxio.client.WriteType;
 import alluxio.client.block.policy.BlockLocationPolicy;
 import alluxio.client.block.policy.options.GetWorkerOptions;
 import alluxio.client.block.stream.BlockInStream;
 import alluxio.client.block.stream.BlockInStream.BlockInStreamSource;
 import alluxio.client.block.stream.BlockOutStream;
+import alluxio.client.block.util.BlockLocationUtils;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.options.InStreamOptions;
 import alluxio.client.file.options.OutStreamOptions;
+import alluxio.collections.Pair;
+import alluxio.conf.PropertyKey;
 import alluxio.exception.ExceptionMessage;
 import alluxio.exception.PreconditionMessage;
 import alluxio.exception.status.ResourceExhaustedException;
@@ -56,6 +56,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import javax.annotation.Nonnull;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -168,6 +169,19 @@ public final class AlluxioBlockStore {
              mContext.acquireBlockMasterClientResource()) {
       info = masterClientResource.get().getBlockInfo(blockId);
     }
+    return getInStream(info, options, failedWorkers);
+  }
+
+  /**
+   * {@link #getInStream(long, InStreamOptions, Map)}.
+   *
+   * @param info the block info
+   * @param options the options associated with the read request
+   * @param failedWorkers the map of workers address to most recent failure time
+   * @return a stream which reads from the beginning of the block
+   */
+  public BlockInStream getInStream(@Nonnull BlockInfo info, InStreamOptions options,
+      Map<WorkerNetAddress, Long> failedWorkers) throws IOException {
     List<BlockLocation> locations = info.getLocations();
     List<BlockWorkerInfo> blockWorkerInfo = Collections.EMPTY_LIST;
     // Initial target workers to read the block given the block locations.
--- a/core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
@@ -332,9 +332,12 @@ public class FileInStream extends InputStream implements BoundedStream, Position
     }
     // Create stream
     boolean isBlockInfoOutdated = true;
+    // blockInfo is "outdated" when all the locations in that blockInfo are failed workers,
+    // if there is at least one location that is not a failed worker, then it's not outdated.
     for (BlockLocation location : blockInfo.getLocations()) {
       if (!mFailedWorkers.containsKey(location.getWorkerAddress())) {
         isBlockInfoOutdated = false;
+        break;
       }
     }
     if (isBlockInfoOutdated) {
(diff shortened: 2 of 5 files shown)


core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
```diff
@@ -168,6 +169,19 @@ public BlockInStream getInStream(long blockId, InStreamOptions options,
...
+
+  /**
+   * {@link #getInStream(long, InStreamOptions, Map)}.
+   *
+   * @param info the block info
+   * @param options the options associated with the read request
+   * @param failedWorkers the map of workers address to most recent failure time
+   * @return a stream which reads from the beginning of the block
+   */
+  public BlockInStream getInStream(@Nonnull BlockInfo info, InStreamOptions options,
```
apc999: is our convention is to use `@NotNull` rather than `@Nonnull`?

core/common/src/main/java/alluxio/wire/FileInfo.java
```diff
@@ -555,7 +564,9 @@ public FileInfo setMountPoint(boolean mountPoint) {
    * @return the file information
    */
   public FileInfo setFileBlockInfos(List<FileBlockInfo> fileBlockInfos) {
-    mFileBlockInfos = new ArrayList<>(fileBlockInfos);
+    for (FileBlockInfo info : fileBlockInfos) {
```
apc999: is this change to avoid creating a new array list?

core/common/src/main/java/alluxio/client/file/URIStatus.java
```diff
@@ -59,6 +60,15 @@ public DefaultAccessControlList getDefaultAcl() {
     return mInfo.getDefaultAcl();
   }
 
+  /**
+   * @param blockId the block ID
+   * @return the corresponding block info or null
+   */
+  public BlockInfo getBlockInfo(long blockId) {
```
apc999: Annotate by `@Nullable`

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -323,8 +324,13 @@ private void updateStream() throws IOException {
     /* Create a new stream to read from mPosition. */
     // Calculate block id.
     long blockId = mStatus.getBlockIds().get(Math.toIntExact(mPosition / mBlockSize));
+    BlockInfo blockInfo = mStatus.getBlockInfo(blockId);
+    if (blockInfo == null) {
+      throw new IOException("Block " + blockId + " does not exist");
+    }
     // Create stream
-    mBlockInStream = mBlockStore.getInStream(blockId, mOptions, mFailedWorkers);
```
apc999: @calvinjia is this intentionally to contact master again rather than using existing info?
If yes, shall we have a flag to guard this?

apc999:
The PR looks good to me in general, please fix the failing integration tests.
adding @gpang as an another reviewer to get another look

core/common/src/main/java/alluxio/wire/FileInfo.java
```diff
@@ -63,7 +64,7 @@
   private int mMode;
   private String mPersistenceState = "";
   private boolean mMountPoint;
-  private ArrayList<FileBlockInfo> mFileBlockInfos = new ArrayList<>();
+  private Map<Long, FileBlockInfo> mFileBlockInfos = new HashMap<>();
```
gpang: do we have to update `serialVersionUID` if we change the field, and would there be any backwards compatibility issues?

core/common/src/main/java/alluxio/wire/FileInfo.java
```diff
@@ -251,7 +252,15 @@ public boolean isMountPoint() {
    * @return the list of file block descriptors
    */
   public List<FileBlockInfo> getFileBlockInfos() {
-    return mFileBlockInfos;
+    return new ArrayList<>(mFileBlockInfos.values());
```
gpang: would this be called often, and would the list copy be expensive?

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -323,8 +324,13 @@ private void updateStream() throws IOException {
     /* Create a new stream to read from mPosition. */
     // Calculate block id.
     long blockId = mStatus.getBlockIds().get(Math.toIntExact(mPosition / mBlockSize));
+    BlockInfo blockInfo = mStatus.getBlockInfo(blockId);
+    if (blockInfo == null) {
+      throw new IOException("Block " + blockId + " does not exist");
```
gpang: can we provide more information about this message, like the path and file id of the file (hopefully they are in `mStatus` already)?

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -323,8 +324,13 @@ private void updateStream() throws IOException {
...
+    }
     // Create stream
-    mBlockInStream = mBlockStore.getInStream(blockId, mOptions, mFailedWorkers);
+    mBlockInStream = mBlockStore.getInStream(blockInfo, mOptions,
```
gpang: Does this not fit on one line?

core/common/src/main/java/alluxio/wire/FileInfo.java
chcn: updated.
Since this is not persisted, we should not worry about incompatibility.

core/common/src/main/java/alluxio/wire/FileInfo.java
chcn: possibly. I keep both a list and a map now.

chcn:
@apc999 @gpang PTAL

gpang:
```
Tests in error: 
alluxio.server.ft.MultiWorkerIntegrationTest.readOneRecoverFromLostWorker(alluxio.server.ft.MultiWorkerIntegrationTest)
  Run 1: MultiWorkerIntegrationTest.readOneRecoverFromLostWorker:152 » Unavailable Targ...
  Run 2: MultiWorkerIntegrationTest.readOneRecoverFromLostWorker:152 » Unavailable Targ...

alluxio.server.ft.MultiWorkerIntegrationTest.readRecoverFromLostWorker(alluxio.server.ft.MultiWorkerIntegrationTest)
  Run 1: MultiWorkerIntegrationTest.readRecoverFromLostWorker:126 » Unavailable Target ...
  Run 2: MultiWorkerIntegrationTest.readRecoverFromLostWorker:126 » Unavailable Target ...

```

core/common/src/main/java/alluxio/wire/FileInfo.java
```diff
@@ -64,7 +64,8 @@
   private int mMode;
   private String mPersistenceState = "";
   private boolean mMountPoint;
-  private Map<Long, FileBlockInfo> mFileBlockInfos = new HashMap<>();
+  private ArrayList<FileBlockInfo> mFileBlockInfoList = new ArrayList<>();
+  private Map<Long, FileBlockInfo> mFileBlockInfoMap = new HashMap<>();
```
gpang: can you add a comment that this is like an index for `mFileBlockInfoList`?

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -323,8 +325,23 @@ private void updateStream() throws IOException {
...
     long blockId = mStatus.getBlockIds().get(Math.toIntExact(mPosition / mBlockSize));
+    BlockInfo blockInfo = mStatus.getBlockInfo(blockId);
+    if (blockInfo == null) {
+      throw new IOException("No BlockInfo for block(id=" + blockId + ") of file"
+          + "(id=" + mStatus.getFileId() + ", path=" + mStatus.getPath() + ")");
+    }
     // Create stream
-    mBlockInStream = mBlockStore.getInStream(blockId, mOptions, mFailedWorkers);
+    boolean isBlockInfoOutdated = true;
+    for (BlockLocation location : blockInfo.getLocations()) {
```
apc999: I'm confused by the logic. Shouldn't it be:
```java
boolean isBlockInfoOutdated = false;
for (BlockLocation location : blockInfo.getLocations()) {
if (mFailedWorkers.containsKey(location.getWorkerAddress())) {
        isBlockInfoOutdated = true;
        break;
      }
  }
```

gpang: hrmmm, yeah, I'm not sure what this is doing either. Is there a test case which covers this code?

chcn: blockInfo is "outdated" when all the locations in that blockInfo are failed workers, if there is at least one location that is not a failed worker, then it's not outdated.

chcn: without this logic, `alluxio.server.ft.MultiWorkerIntegrationTest.readRecoverFromLostWorker` and `alluxio.server.ft.MultiWorkerIntegrationTest.readOneRecoverFromLostWorker` will fail.

gpang: I see. Can you add that explanation as a comment? Also, once `isBlockInfoOutdated = false`, you can `break` out of the loop, right?

chcn:
@gpang PTAL

gpang:

apc999:

Issue #10470, opened 2019-11-14
Process worker metrics concurrently


ns1123:
**Is your feature request related to a problem? Please describe.**
The existing metrics system isn't great with concurrent updates from the worker heartbeats. Often times, processing of a large number of worker metrics ends up occurring serially which can then result in poor performance/latency.

The metrics store is serialized here:
https://github.com/Alluxio/alluxio/blob/v2.1.0/core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java#L98

**Describe the solution you'd like**
Improvements to how the master processes Worker heartbeat metrics. Ideally done in a way that is concurrent with minimal locking required.

**Describe alternatives you've considered**
Reducing the frequency of the worker heartbeats reduces the contention on the lock.

An external metrics system may be the preferred approach with each process sending metrics via heartbeats to an external server.

**Urgency**
Can cause performance degradation when running at large worker scale.

PR #10485, opened 2019-11-19
Process worker metrics asynchronously

LuQQiu:
Fixes https://github.com/Alluxio/alluxio/issues/10470
Previously we process worker metrics synchronously in the MasterWorkerInfo lock and also the WorkerMetrics lock. This will 
(1) Make each worker heartbeat takes longer. Worker heartbeat need to wait for the submitted metrics to be processed.
(2) Make worker heartbeats block each other. We can only process one worker metrics at a time.

Process worker metrics asynchronously will help unblock worker heartbeats.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
@@ -90,7 +90,7 @@ public class DefaultMetricsMaster extends CoreMaster implements MetricsMaster, N
             ServerConfiguration.global(), mMasterContext.getUserState());
     int numThreads = ServerConfiguration.getInt(PropertyKey.MASTER_METRICS_SERVICE_THREADS);
     mExecutorService = ExecutorServiceFactories.fixedThreadPool(
-        "alluxio-master-metrics-service", numThreads).create();
+        "alluxio-master-metrics-updater", numThreads).create();
   }
 
   @VisibleForTesting
--- a/core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
@@ -47,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
 
 /**
  * Default implementation of the metrics master.
@@ -57,6 +58,7 @@ public class DefaultMetricsMaster extends CoreMaster implements MetricsMaster, N
       new HashSet<>();
   private final MetricsStore mMetricsStore;
   private final HeartbeatThread mClusterMetricsUpdater;
+  private final ExecutorService mExecutorService;
 
   /**
    * Creates a new instance of {@link MetricsMaster}.
@@ -86,6 +88,9 @@ public class DefaultMetricsMaster extends CoreMaster implements MetricsMaster, N
             new ClusterMetricsUpdater(),
             ServerConfiguration.getMs(PropertyKey.MASTER_CLUSTER_METRICS_UPDATE_INTERVAL),
             ServerConfiguration.global(), mMasterContext.getUserState());
+    int numThreads = ServerConfiguration.getInt(PropertyKey.MASTER_METRICS_SERVICE_THREADS);
+    mExecutorService = ExecutorServiceFactories.fixedThreadPool(
+        "alluxio-master-metrics-service", numThreads).create();
   }
 
   @VisibleForTesting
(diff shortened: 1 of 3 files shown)


LuQQiu:
Jenkins, test this please

LuQQiu:
Jenkins, test this please

LuQQiu:
@gpang PTAL, thanks!

core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
```diff
@@ -188,10 +193,17 @@ public String getName() {
...
     if (isLeader) {
+      mMetricsStore.start();
       getExecutorService().submit(mClusterMetricsUpdater);
     }
   }
 
+  @Override
+  public void stop() throws IOException {
+    super.stop();
+    mMetricsStore.stop();
```
gpang: I have a high-level question. Should the queue and async nature be in the MetricsMaster, instead of the MetricsStore? It might make sense to make the MetricsStore simply a storing data structure, and this MetricsMaster just happens to update the store in an async way? I feel that allows the MetricsStore to simply handle storing/serving the metrics data as efficiently as possible.

What are you thoughts on that?

LuQQiu: Do you mean that we MetricsMaster have 
1) A queue storing all the worker/client metric submit through heartbeat
2) A thread pool (executor), if detect any element in the queue, poll from queue and submit to the executor which calls metricsStore.processMetrics

In this way, metricsStore is a storing data structure and it's multi-threaded instead of single-threaded. (More efficiency if we improve the Metric concurrency)

gpang: I think that is what I was proposing.

Yes, I was proposing that the MetricsMaster owns the threads, and executors, to implement the async nature of the updates.

I'm not sure what you mean by `metricsStore is a storing data structure and it's multi-threaded`. I was saying that the MetricsStore is a data structure, and the usage of that data structure (MetricsMaster) will call it in an async or multi-threaded way.

LuQQiu:
@gpang Move the async logics from `MetricsStore` to `DefaultMetricsMaster`. PTAL, thanks!

core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
```diff
@@ -86,6 +88,9 @@
             new ClusterMetricsUpdater(),
             ServerConfiguration.getMs(PropertyKey.MASTER_CLUSTER_METRICS_UPDATE_INTERVAL),
             ServerConfiguration.global(), mMasterContext.getUserState());
+    int numThreads = ServerConfiguration.getInt(PropertyKey.MASTER_METRICS_SERVICE_THREADS);
+    mExecutorService = ExecutorServiceFactories.fixedThreadPool(
+        "alluxio-master-metrics-service", numThreads).create();
```
gpang: maybe alluxio-master-metrics-updater ? I think "service" is a bit too general.

core/server/master/src/test/java/alluxio/master/metrics/MetricsMasterTest.java
```diff
@@ -86,14 +93,14 @@ public void testAggregator() {
         Metric.from("worker.192_1_1_2.metricA", 1, MetricType.GAUGE),
         Metric.from("worker.192_1_1_2.metricB", 2, MetricType.GAUGE));
     mMetricsMaster.workerHeartbeat("192_1_1_2", metrics2);
-    assertEquals(11L, getGauge("metricA"));
+    waitForAndCheckResult(() -> (Long) getGauge("metricA") == 11L);
```
gpang: why do only some of the asserts use this wait()? Is there a way to just wait for the update to happen, and then do the original asserts? It might be confusing to determine which assert or wait to call in which order.

LuQQiu: For now, I only use this wait() in the first assert check after worker's heartbeat(s). If one metric is updated, other metrics normally will also be updated.  I haven't come up with better ways to wait for the async metrics updater to finish. It will be better to use this waitAndCheck() function for all the assert checks to avoid the racing issues.

LuQQiu:
Tests in error: 
alluxio.client.cli.fs.command.PersistCommandTest.persistPartial(alluxio.client.cli.fs.command.PersistCommandTest)
  Run 1: PersistCommandTest.persistPartial:297->AbstractFileSystemShellTest.checkFilePersisted:307 » Timeout
  Run 2: PersistCommandTest.persistPartial:297->AbstractFileSystemShellTest.checkFilePersisted:307 » Timeout

LuQQiu:
Jenkins, test this please

gpang:

LuQQiu:
alluxio-bot, cherry-pick to branch-2.1 please

LuQQiu:

LuQQiu:

PR #10486, opened 2019-11-19
Report CPU Load Average of JobWorkers

bradyoo:
Package Used: oshi (https://github.com/oshi/oshi)
License: MIT

**Example**: (This load average is equivalent to the load average you would get from commands like w on the host itself.)

ec2-user@**AsyncWrite1-workers-2**:/tmp/alluxio ./bin/alluxio fsadmin report jobservice
Worker: **AsyncWrite1-workers-2**  Load Avg: 11.72, 3.06, 1.1
Worker: AsyncWrite1-workers-1  Load Avg: 11.08, 3.08, 1.1
...
(some other text containing report)

ec2-user@**AsyncWrite1-workers-2**:/tmp/alluxio w
 22:53:11 up 7 min,  1 user,  load average: 12.62, 3.39, 1.22
...
(some other text)

Changes:

workerHeartbeat's 1st parameter changed from:

long workerId -> JobWorkerHealth jobWorkerHealth

and contains other job worker's health information along with the id.

@ZacBlanco 

CC: @calvinjia


job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+  private final long mWorkerId;
+  private final ArrayList<Double> mLoadAverage;
+  private final long mLastUpdated;
+
+  private String mHostname;
+
+  /**
+   * Default constructor.
+   * @param workerId the worker id
+   * @param loadAverage the cpu load average
```
ZacBlanco: nit: Can you be more descriptive about this argument or refer to where else in the javadoc I can read about this param?

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+    mWorkerId = workerId;
+    mLoadAverage = Lists.newArrayList();
+    for (double l : loadAverage) {
+      mLoadAverage.add(l);
+    }
+    mLastUpdated = CommonUtils.getCurrentMs();
+  }
+
+  /**
+   * Constructor from the grpc representation.
```
ZacBlanco: nit: add a newline after the first description

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+public class JobWorkerHealth {
+
+  private final long mWorkerId;
+  private final ArrayList<Double> mLoadAverage;
+  private final long mLastUpdated;
+
+  private String mHostname;
+
+  /**
+   * Default constructor.
```
ZacBlanco: nit: Add a newline after first sentence description

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+    mLastUpdated = CommonUtils.getCurrentMs();
+  }
+
+  /**
+   * Constructor from the grpc representation.
+   * @param jobWorkerHealth grpc representation
+   */
+  public JobWorkerHealth(alluxio.grpc.JobWorkerHealth jobWorkerHealth) {
+    mWorkerId = jobWorkerHealth.getWorkerId();
+    mLoadAverage = Lists.newArrayList(jobWorkerHealth.getLoadAverageList());
```
ZacBlanco: What's the reason to copy the gRPC value into a new list?

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+    if (mHostname != null) {
+      builder.setHostname(mHostname);
+    }
+
+    return builder.build();
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(mWorkerId);
```
ZacBlanco: Does this mean that objects with the same worker id but different lists will hash to the same value? Maybe we could use the `lastUpdatedMs` to make the hashcode more evenly distributed?

job/server/pom.xml
```diff
@@ -30,6 +30,11 @@
 
   <dependencies>
     <!-- External dependencies -->
+    <dependency>
+      <groupId>com.github.oshi</groupId>
+      <artifactId>oshi-core</artifactId>
+      <version>${oshi.version}</version>
+    </dependency>
```
ZacBlanco: Assuming you went through the necessary hoops dealing with the license on this library?

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -319,6 +328,15 @@ public JobInfo getStatus(long jobId) throws JobDoesNotExistException {
     return new JobServiceSummary(jobInfos);
   }
 
+  /**
+   * @return health metrics for each of the job workers
+   */
+  public List<JobWorkerHealth> getAllWorkerHealth() {
+    ArrayList<JobWorkerHealth> result = Lists.newArrayList(mWorkerHealth.values());
+    Collections.sort(result, Comparator.comparingLong((a) -> a.getWorkerId()));
```
ZacBlanco: This could be relatively expensive depending on the size of the cluster. Maybe we could use a `ConcurrentSkipListSet/Map` instead?

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -368,12 +386,15 @@ public long registerWorker(WorkerNetAddress workerNetAddress) {
    * Updates the tasks' status when a worker periodically heartbeats with the master, and sends the
    * commands for the worker to execute.
    *
-   * @param workerId the worker id
+   * @param jobWorkerHealth the job worker info
```
ZacBlanco: job worker _health_ info?

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -388,6 +409,9 @@ public long registerWorker(WorkerNetAddress workerNetAddress) {
       worker.updateLastUpdatedTimeMs();
     }
 
+    jobWorkerHealth.setHostname(hostname);
```
ZacBlanco: Would it be possible to just include/set the hostname when we construct the health object instead of setting it over here?

I feel like it is more bug prone to have the object mostly immutable but then have a single field be able to be set like this. If there is good reason it's fine to keep it

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
```diff
@@ -0,0 +1,37 @@
...
+   */
+  public JobWorkerHealthReporter() {
+    mProcessor = new SystemInfo().getHardware().getProcessor();
+  }
+
+  /**
+   * @return the cpu load average of the worker
+   */
+  public double[] getCpuLoadAverage() {
+    return mProcessor.getSystemLoadAverage(3);
```
ZacBlanco: I'm not sure if Alluxio ever ran on Windows machines, but it looks like this method will prevent it from running there

http://oshi.github.io/oshi/apidocs/oshi/hardware/CentralProcessor.html#getSystemLoadAverage-int-

ZacBlanco: Also, when the 1/5/15 minute loads aren't available we get a negative number returned. Should we do any special handling for those cases?

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The job worker health information.
+ */
+public class JobWorkerHealth {
+
+  private final long mWorkerId;
+  private final ArrayList<Double> mLoadAverage;
```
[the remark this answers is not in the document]
bradyoo: The documentation around this on various systems is weak. It appears though that at least linux and mac's definition is 1 minute, 5 minute, 15 minute averages. 

I thought the definition would be available at places like the man page of w command which prints load averages but since they didn't have it, I decided to not have it. But I could.

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
bradyoo: See equals. Equals intentionalyl left out lastUpdatedMs.

ZacBlanco: Is there a reason that you don't want equals to use `lastUpdatedMs`?

bradyoo:
I needed to take advantage of it in the test for matching parameters for
verification and lastUpdatedMs is impossible to match because it's
currentMs at construction.

bradyoo:
Addressed some feedback and left comments about it on others.

And yes, there will be more metrics that are added here later.

bradyoo:

bradyoo:
Ping

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
ZacBlanco: I know we only print them now - but in the future do we want to simply return the negative numbers, or do any kind of processing here so that we don't have different logic for processing negative numbers?

i.e. in the case of any negative numbers we could return an empty array, set the negative values to 0, etc..

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
ZacBlanco: can we still update this?

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -0,0 +1,121 @@
...
+   * Default constructor.
+   * @param workerId the worker id
+   * @param loadAverage the cpu load average
+   */
+  public JobWorkerHealth(long workerId, double[] loadAverage) {
+    mWorkerId = workerId;
+    mLoadAverage = Lists.newArrayList();
+    for (double l : loadAverage) {
+      mLoadAverage.add(l);
+    }
```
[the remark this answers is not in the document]
ZacBlanco: Ah you're right. I didn't realize that it doesn't like primitives.

We can keep this, but another alternative is

```
DoubleStream.of(doublesArray).boxed().collect(Collectors.toList());
```

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
bradyoo: I don't think it's any different because if I use this for logic, I'll be comparing values to be greater than some value. I probably shouldn't change value of an api value until I need to.

job/server/src/main/java/alluxio/master/job/JobMaster.java
ZacBlanco: When the worker registers The master just uses the information provided in `WorkerNetAddress` - doesn't it report the hostname in there? Shouldn't the worker should have that information as well since it's the source of that information?

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
bradyoo: I don't expect this constructor to be called by anyone other than as output of CentralProcessor.getSystemLoadAverage so I guess I'll put that instead.

bradyoo:
You are right. But JobWorker doesn't actually store that information and
I'm a couple refactors away from actually getting that information readily
available on construction.

bradyoo:
Actually nevermind I have it. Let me stamp it better.

bradyoo:
Responded to comments.

job/server/src/main/java/alluxio/master/job/JobMaster.java
ZacBlanco: Have you been able to do a performance comparison? Sorting every time someone does an RPC sounds more expensive to me than the extra bookkeeping required for using a ConcurrentSkipList, especially once a cluster grows to a few hundred or even thousands of workers.

bradyoo:
This is only a result from hitting this through a report CLI.

bradyoo:
@ZacBlanco: merge?

ZacBlanco:

PR #10493, opened 2019-11-20
Improve concurrency for MetricsStore

LuQQiu:
Fixes https://github.com/Alluxio/alluxio/issues/10470
This PR aims to improve concurrency in MetricsStore when processing worker/client metrics.
It removes the synchronized lock on the metrics sets.
Atomically updates metric values instead of removing and constructing new objects.
This memoizes the Metrics.getFullMetricName(), since it is costly in time and allocations.

When Metric.getFullMetricName() runs normally, it bumps up the heartbeats can be processed per second from 1000-3000 heartbeats/second to 2000-5000 heartbeats/second.
When Metric.getFullMetricName() is unexpectedly slow, the performance improvement can be really big (when 100 worker heartbeats with 50 executor threads, before my change takes 11 min, after my change takes 18 second if getFullMetricName sleep for 15ms).

### DIFF
--- a/core/common/src/main/java/alluxio/metrics/Metric.java
+++ b/core/common/src/main/java/alluxio/metrics/Metric.java
@@ -13,6 +13,8 @@ package alluxio.metrics;
 
 import alluxio.grpc.MetricType;
 
+import alluxio.util.CommonUtils;
+
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
@@ -25,6 +27,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
 
 /**
  * A metric of a given instance. The instance can be master, worker, or client.
@@ -46,6 +49,14 @@ public final class Metric implements Serializable {
   // TODO(yupeng): consider a dedicated data structure for tag, when more functionality are added to
   // tags in the future
   private final Map<String, String> mTags;
+
+  /**
+   * The unique identifier to represent this metric.
+   * The pattern is instance.[hostname-id:instanceId.]name[.tagName:tagValue]*.
+   * Fetched once and assumed to be immutable.
+   */
+  private final Supplier<String> mFullMetricNameSupplier =
+      CommonUtils.memoize(this::constructFullMetricName);
 
   private AtomicDouble mValue;
 
@@ -197,6 +208,15 @@ public final class Metric implements Serializable {
    *         at the end
    */
   public String getFullMetricName() {
+    return mFullMetricNameSupplier.get();
+  }
+
+  /**
+   * @return the fully qualified metric name, which is of pattern
+   *         instance.[hostname-id:instanceId.]name[.tagName:tagValue]*, where the tags are appended
+   *         at the end
+   */
+  private String constructFullMetricName() {
     StringBuilder sb = new StringBuilder();
     sb.append(mInstanceType).append('.');
     if (mHostname != null) {
--- a/core/common/src/main/java/alluxio/metrics/Metric.java
+++ b/core/common/src/main/java/alluxio/metrics/Metric.java
@@ -40,12 +40,13 @@ public final class Metric implements Serializable {
   private final MetricsSystem.InstanceType mInstanceType;
   private final String mHostname;
   private final String mName;
-  private final Double mValue;
   private final MetricType mMetricType;
   private String mInstanceId;
   // TODO(yupeng): consider a dedicated data structure for tag, when more functionality are added to
   // tags in the future
   private final Map<String, String> mTags;
+
+  private Double mValue;
 
   /**
    * Constructs a {@link Metric} instance.
@@ -84,6 +85,33 @@ public final class Metric implements Serializable {
   }
 
   /**
+   * Add metric value delta to the existing value.
+   * This method should only be used by {@link alluxio.master.metrics.MetricsStore}
+   *
+   * @param value value to add
+   */
+  public void addValue(double value) {
+    mValue += value;
+  }
+
+  /**
+   * Set the metric value.
+   * This method should only be used by {@link alluxio.master.metrics.MetricsStore}
+   *
+   * @param value value to set
+   */
+  public void setValue(double value) {
+    mValue = value;
+  }
+
+  /**
+   * @return the metric value
+   */
+  public double getValue() {
+    return mValue;
+  }
+
+  /**
    * @return the instance type
    */
   public MetricsSystem.InstanceType getInstanceType() {
@@ -119,13 +147,6 @@ public final class Metric implements Serializable {
    */
   public String getName() {
     return mName;
-  }
-
-  /**
-   * @return the metric value
-   */
-  public double getValue() {
-    return mValue;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


gpang:
@LuQQiu Thanks for this improvement! I'm curious, why is this dependent on the async change? I was imagining that processing a metric update could be made more concurrent independent to any async change?

Also, do you have any numbers/experiments which show how much this improves the concurrency?

LuQQiu:
@gpang  I made the changes on top of the async change because no trivial changes has made on MetricsStore to deal with async change. If we put more logic on the MetricsMaster instead of MetricsStore, this PR can be independent. 
I haven't tested yet because I feel it will be better to at least merge the async PR and this PR and then do the test.

gpang:
I left a comment on the async PR, but if that PR implements all the async implementation in the MetricsMaster, then the MetricsStore changes would be independent.

LuQQiu:
@gpang I only change from overwriting original whole metric to add/set value in the existing metric. 
I created a single unit test against MetricsStore. The current change bumps up the performance about 4-8X. 
The time takes to processWorkerMetrics with 10 threads executor, 50 workers (each has 100 metrics),  50 heartbeats for each worker takes about 120 ms in my unit test in the current PR, compared to about 900 ms in master branch.

In reality, the metrics will be much more complicated and are expected to take more than when processing them.

I will continue to see if any low-hanging fruits can be achieved to continue bump up performance.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -141,25 +137,15 @@ private static void putReportedMetrics(IndexedSet<Metric> metricSet, String inst
...
-        Metric oldMetric = metricSet.getFirstByField(FULL_NAME_INDEX, metric.getFullMetricName());
-        double oldVal = oldMetric == null ? 0.0 : oldMetric.getValue();
-        Metric newMetric = new Metric(metric.getInstanceType(), metric.getHostname(),
-            metric.getMetricType(), metric.getName(), oldVal + metric.getValue());
-        for (Map.Entry<String, String> tag : metric.getTags().entrySet()) {
-          newMetric.addTag(tag.getKey(), tag.getValue());
-        }
-        metricSet.removeByField(FULL_NAME_INDEX, metric.getFullMetricName());
-        newMetrics.add(newMetric);
+      Metric oldMetric = metricSet.getFirstByField(FULL_NAME_INDEX, metric.getFullMetricName());
```
gpang: instead of reading first, then adding, can we just call `add` directly? The result of `add` will be `true` if it successfully add the new item, and `false` if it already exists. Then, if it didn't exist, we can "update" the metric value itself but setting it or incrementing it.

I think if we make these changes, we can even remove the `synchronized` block from the put methods above.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -141,25 +137,15 @@ private static void putReportedMetrics(IndexedSet<Metric> metricSet, String inst
...
+        metricSet.add(metric);
+      } else if (metric.getMetricType() == MetricType.COUNTER) {
+        oldMetric.addValue(metric.getValue());
       } else {
-        metricSet.removeByField(FULL_NAME_INDEX, metric.getFullMetricName());
-        newMetrics.add(metric);
+        oldMetric.setValue(metric.getValue());
       }
     }
-    metricSet.removeByField(ID_INDEX, instanceId);
```
gpang: Is there a case where a field will actually be removed? Looking at the javadoc for this method, it seems like there was a case where fields would be removed?

core/common/src/main/java/alluxio/metrics/Metric.java
```diff
@@ -40,13 +40,14 @@
...
   private final String mHostname;
   private final String mName;
-  private final Double mValue;
   private final MetricType mMetricType;
   private String mInstanceId;
   // TODO(yupeng): consider a dedicated data structure for tag, when more functionality are added to
   // tags in the future
   private final Map<String, String> mTags;
 
+  private Double mValue;
```
gpang: shouldn't this be an atomic boolean or a `volatile` variable for thread safety?

gpang:
Jenkins, test this please

core/common/src/main/java/alluxio/metrics/Metric.java
gpang: Actually, `volatile` won't work since there is an increment operation. It also doesn't seem like AtomicDouble exists in java, but there is one in guava.

core/server/master/src/test/java/alluxio/master/metrics/MetricsMasterTest.java
```diff
@@ -94,7 +94,7 @@ public void testAggregator() {
         Metric.from("worker.192_1_1_2.metricA", 3, MetricType.GAUGE));
     mMetricsMaster.workerHeartbeat("192_1_1_2", metrics3);
     assertEquals(13L, getGauge("metricA"));
-    assertEquals(20L, getGauge("metricB"));
+    assertEquals(22L, getGauge("metricB"));
```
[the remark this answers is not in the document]
LuQQiu: This is related to whether we want to keep the old metrics. For the previous implementation, if a metrics is not reported in the current heartbeat but reported in the previous heartbeats by a worker/client, it will be deleted.
In my implementation, even if a metric is not reported in the current heartbeat, we should also keep its old value to show the more correct metrics, especially for MetricType=VALUE.

I think the previous implementation helps remove old metrics reported by old clients, but doesn't make sense for worker metrics which is always stable and need to record old values(COUNTERS)

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
LuQQiu: This related to the MetricsMasterTest.testAggregater. For a worker or client A, if it reports metrics1 and metrics2 in previous heartbeats, but only reports metrics1 in the current heartbeat, the metric recording of the metric2 of this instance A will be deleted.

LuQQiu:
alluxio.client.cli.fs.command.CpCommandIntegrationTest.copyFromLocalLarge(alluxio.client.cli.fs.command.CpCommandIntegrationTest)
  Run 1: CpCommandIntegrationTest.copyFromLocalLarge:442 » Internal http2 exception (Gr...
  Run 2: CpCommandIntegrationTest.copyFromLocalLarge:442 » Internal http2 exception (Gr...

LuQQiu:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
In general, I think this looks good. I am just curious about the performance differences now.

LuQQiu:
Tests in error: 
alluxio.server.ft.ZookeeperFailureIntegrationTest.zkConnectionPolicy_Session(alluxio.server.ft.ZookeeperFailureIntegrationTest)
  Run 1: ZookeeperFailureIntegrationTest.zkConnectionPolicy_Session:139 » Timeout Timed...
  Run 2: ZookeeperFailureIntegrationTest.zkConnectionPolicy_Session:139 » Timeout Timed...

  Run 1: ZookeeperFailureIntegrationTest.zkFailure:80 » Timeout Timed out waiting for a...
  Run 2: ZookeeperFailureIntegrationTest.zkFailure:80 » Timeout Timed out waiting for a...

LuQQiu:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

gpang:

LuQQiu:
alluxio-bot, cherry-pick to branch-2.1 please

LuQQiu:

PR #10494, opened 2019-11-20
Change return type of selectExecutors for PlanDefinition

bradyoo:
From:

Map<WorkerInfo, Parameters> -> List<Pair<WorkerInfo, Parameters>>

This allows for each plans to submit more than one task perTaskWorker.

Other miscellaneous Change:

Pair's equal and hashcode was modified so that Pair accepts null as a valid input for First and/or Second.

### DIFF
--- a/job/server/src/main/java/alluxio/job/plan/PlanDefinition.java
+++ b/job/server/src/main/java/alluxio/job/plan/PlanDefinition.java
@@ -45,11 +45,11 @@ public interface PlanDefinition<T extends JobConfig, P extends Serializable,
    * @param config the job configuration
    * @param jobWorkerInfoList the list of available workers' information
    * @param selectExecutorsContext the context containing information used to select executors
-   * @return a map of selected workers to the parameters to pass along
+   * @return a list of pairs of selected workers to the parameters to pass along
    * @throws Exception if any error occurs
    */
   List<Pair<WorkerInfo, P>> selectExecutors(T config, List<WorkerInfo> jobWorkerInfoList,
-    SelectExecutorsContext selectExecutorsContext) throws Exception;
+      SelectExecutorsContext selectExecutorsContext) throws Exception;
 
   /**
    * Runs the task in the executor.
--- a/job/server/src/main/java/alluxio/job/plan/PlanDefinition.java
+++ b/job/server/src/main/java/alluxio/job/plan/PlanDefinition.java
@@ -20,6 +20,7 @@ import alluxio.wire.WorkerInfo;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * A job definition. A definition has two important parts: (1) a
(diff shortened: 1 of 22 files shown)


bradyoo:
rebased

bradyoo:
@ZacBlanco: do you want to look at this? Or should I give it to someone else?

job/server/src/main/java/alluxio/job/plan/PlanDefinition.java
```diff
@@ -44,10 +46,10 @@
    * @param config the job configuration
    * @param jobWorkerInfoList the list of available workers' information
    * @param selectExecutorsContext the context containing information used to select executors
-   * @return a map of selected workers to the parameters to pass along
+   * @return a set of pairs of selected workers to the parameters to pass along
    * @throws Exception if any error occurs
    */
-  Map<WorkerInfo, P> selectExecutors(T config, List<WorkerInfo> jobWorkerInfoList,
+  Set<Pair<WorkerInfo, P>> selectExecutors(T config, List<WorkerInfo> jobWorkerInfoList,
```
ZacBlanco: Is this really the best way to represent the output? It seems more natural to me to map the `WorkerInfo` to another collection, rather than by using `Pair`s

I'm not opposed - just wondering if you thought of other ways of representing it and why you landed on this

job/server/src/main/java/alluxio/job/plan/persist/PersistDefinition.java
```diff
@@ -36,7 +36,7 @@
 import alluxio.underfs.options.MkdirsOptions;
 import alluxio.wire.WorkerInfo;
 
-import com.google.common.collect.Maps;
+import com.beust.jcommander.internal.Sets;
```
ZacBlanco: same comment on this `Sets` library

job/server/src/main/java/alluxio/master/job/plan/PlanCoordinator.java
```diff
@@ -197,17 +200,24 @@ public void setJobAsFailed(String errorMessage) {
...
-      if (taskInfo.getStatus().isFinished()) {
-        return;
+
+      boolean statusChanged = false;
+      for (Long taskId : taskIds) {
+        TaskInfo taskInfo = mPlanInfo.getTaskInfo(taskId);
+        if (taskInfo.getStatus().isFinished()) {
+          return;
+        }
+        if (!mPlanInfo.getStatus().isFinished()) {
```
ZacBlanco: Does the condition above not guarantee that this is always true if we reach this statement?

job/server/src/main/java/alluxio/master/job/plan/PlanCoordinator.java
```diff
@@ -197,17 +200,24 @@ public void setJobAsFailed(String errorMessage) {
...
         return;
       }
-      TaskInfo taskInfo = mPlanInfo.getTaskInfo(taskId);
-      if (taskInfo.getStatus().isFinished()) {
-        return;
+
+      boolean statusChanged = false;
+      for (Long taskId : taskIds) {
+        TaskInfo taskInfo = mPlanInfo.getTaskInfo(taskId);
+        if (taskInfo.getStatus().isFinished()) {
```
ZacBlanco: is it always guaranteed that if one taskInfo is finished that we don't want to fail the rest?

job/server/src/main/java/alluxio/job/plan/PlanDefinition.java
bradyoo: Yes. The primary reason is because P can also be a collection.

job/server/src/main/java/alluxio/master/job/plan/PlanCoordinator.java
bradyoo: You are right. This is wrong.

bradyoo: I naively left the code but clearly that's not the right logic when there are multiple tasks

job/server/src/main/java/alluxio/master/job/plan/PlanCoordinator.java
bradyoo: Nope. The parent plan could have finished because one of its other tasks could have failed.

bradyoo:
Feedback addressed.

ZacBlanco:

PR #10496, opened 2019-11-21
Remove unnecessary lock to improve async journal writer concurrency

ggezer:
Because a semaphore is used to track all parties that requires a flush, ticket list need not be synchronized but could be accessed concurrently.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
@@ -11,6 +11,7 @@
 
 package alluxio.master.journal;
 
+import alluxio.collections.ConcurrentHashSet;
 import alluxio.concurrent.ForkJoinPoolHelper;
 import alluxio.concurrent.jsr.ForkJoinPool;
 import alluxio.conf.PropertyKey;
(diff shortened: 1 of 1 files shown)


core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
```diff
@@ -302,28 +292,24 @@ private void doFlush() {
...
-        try (LockResource lr = new LockResource(mTicketLock)) {
-          ListIterator<FlushTicket> ticketIterator = mTicketList.listIterator();
-          while (ticketIterator.hasNext()) {
-            FlushTicket ticket = ticketIterator.next();
-            if (ticket.getTargetCounter() <= mFlushCounter.get()) {
-              ticket.setCompleted();
-              ticketIterator.remove();
-            }
+        Iterator<FlushTicket> ticketIterator = mTicketSet.iterator();
+        while (ticketIterator.hasNext()) {
```
gpang: will this iterator be correct w.r.t. the concurrency?

ggezer: Whether doFlush should enter or not, is controlled by a semaphore. So as long as there is an RPC that needs a flush, doFlush will spin.

Because of that even if an active iteration of doFlush misses a ticket due to concurrency, it'll iterate one more because of the permit given by the ticket holder. And that guarantees the correctness.

gpang:
Jenkins, test this please

gpang:

ggezer:

PR #10498, opened 2019-11-22
Refine ufs.listStatus granularity for incremental sync

madanadit:
For incremental sync on large directories, a ufs listing at the sync point can be an expensive operation. This PR attempts to address the issue by calculating the lowest common ancestor of the changed file list reported by the ufs.

### DIFF
--- a/core/base/src/main/java/alluxio/util/io/PathUtils.java
+++ b/core/base/src/main/java/alluxio/util/io/PathUtils.java
@@ -150,9 +150,8 @@ public final class PathUtils {
         }
       }
     }
-    AlluxioURI lca = new AlluxioURI(AlluxioURI.SEPARATOR);
-    return new AlluxioURI(
-        PathUtils.concatPath(lca, matchedComponents.subList(0, matchedLen).toArray()));
+    return new AlluxioURI(PathUtils.concatPath(AlluxioURI.SEPARATOR,
+        matchedComponents.subList(0, matchedLen).toArray()));
   }
 
   /**
--- a/core/base/src/main/java/alluxio/util/io/PathUtils.java
+++ b/core/base/src/main/java/alluxio/util/io/PathUtils.java
@@ -20,6 +20,11 @@ import com.google.common.base.CharMatcher;
 import com.google.common.base.Preconditions;
 import org.apache.commons.io.FilenameUtils;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3213,7 +3213,14 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
 
     Map<AlluxioURI, UfsStatus> statusCache;
     try (RpcContext rpcContext = createRpcContext()) {
-      statusCache = populateStatusCache(path, DescendantType.ALL);
+      if (changedFiles == null) {
+        // full sync
```
gpang: should we just move these comments and statusCache population into the next set of `if` statements?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3213,7 +3213,14 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
...
+        statusCache = populateStatusCache(path, DescendantType.ALL);
+      } else {
+        // incremental sync
+        statusCache = populateStatusCache(PathUtils.findLowestCommonAncestor(changedFiles),
```
gpang: does `populateStatusCache` handle `null` paths?

core/base/src/main/java/alluxio/util/io/PathUtils.java
```diff
@@ -121,6 +126,35 @@ public static String concatPath(Object base, Object... paths) throws IllegalArgu
...
   }
 
+  /**
+   * @param paths the set of paths
+   * @return the lowest common ancestor, or null if paths is null or empty
+   */
+  public static AlluxioURI findLowestCommonAncestor(Collection<AlluxioURI> paths) {
+    if (paths == null || paths.isEmpty()) {
+      return null;
+    }
```
gpang: Do we know if it is common or rare to have only 1 path in `paths`?

core/base/src/main/java/alluxio/util/io/PathUtils.java
```diff
@@ -121,6 +126,35 @@ public static String concatPath(Object base, Object... paths) throws IllegalArgu
...
+      }
+
+      for (int i = 0; i < pathComp.length && i < matchedLen; ++i) {
+        if (!matchedComponents.get(i).equals(pathComp[i])) {
+          matchedLen = i;
+          break;
+        }
+      }
+    }
+    AlluxioURI lca = new AlluxioURI(AlluxioURI.SEPARATOR);
```
gpang: is there a way to avoid creating an extra AlluxioURI object for the root?

core/base/src/main/java/alluxio/util/io/PathUtils.java
```diff
@@ -121,6 +126,35 @@ public static String concatPath(Object base, Object... paths) throws IllegalArgu
...
+    return new AlluxioURI(
+        PathUtils.concatPath(lca, matchedComponents.subList(0, matchedLen).toArray()));
```
gpang: does `matchedComponents` not have an entry for the root?

core/common/src/test/java/alluxio/util/io/PathUtilsTest.java
```diff
@@ -119,6 +122,52 @@ public void concatPath() {
...
+   */
+  @Test
+  public void findLowestCommonAncestor() {
+    assertNull(PathUtils.findLowestCommonAncestor(null));
+    assertNull(PathUtils.findLowestCommonAncestor(Collections.EMPTY_LIST));
+
+    ArrayList<AlluxioURI> paths = new ArrayList<>();
+
+    paths.add(new AlluxioURI("/"));
+    assertTrue("/".equals(PathUtils.findLowestCommonAncestor(paths).getPath()));
```
gpang: can we not `assertEquals` throughout this test?

core/base/src/main/java/alluxio/util/io/PathUtils.java
madanadit: Not sure. I thought about about adding a property for a threshold below which we prefer individual listing over recursive (on lca). not sure what a good default would be (maybe 100). 

based on on my test setups, ufs.listStatus times:
- 1 file = 5ms
- recursive list on 1.5k files = 1.5s

The right number for the threshold depends on the number of ufs entries under the sync point / latency to the ufs.

madanadit: what do you think of such as property?

core/base/src/main/java/alluxio/util/io/PathUtils.java
madanadit: nope, "/".split -> returns an empty list

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
madanadit: nope, findLCA is never null though for non-empty change list (the incremental case)

madanadit:
@gpang PTAL

gpang:

madanadit:

madanadit:

madanadit:

PR #10499, opened 2019-11-22
Add JobMaster's ability to throttle tasks from running in JobWorker

bradyoo:
@calvinjia: Note that this is very different from the implementation we have talked about. This is primarily because:

1. There are other priorities being pushed and therefore, I took a bit of an easier path.
2. This is more extensible because it adds throttling to ExecutorService in general which is a common pattern we use.

Downsides of this implementation:

1. The queue can grow unbounded. In this case, it can grow up to Integer.MAX_VALUE.

2. Becuase the max number of threads of the thread pool executor is set to a very high number,
giving bound to the queue has unintended side effects where the thread pool executor will start launching new threads. I can change the implementation to account for this.

3. The current cancel implementation needs to be improved to where the task gets cancelled even when it's in the queue for the case when it gets popped off the queue. But I also don't think I will have much success removing items from this queue proactively.

In a very foreseeable Future:

Dynamically change this number based on WorkerHealth.


job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -35,10 +36,11 @@
    * @param loadAverage output of CentralProcessor.getSystemLoadAverage on the worker
    * @param hostname hostname of the worker
    */
-  public JobWorkerHealth(long workerId, double[] loadAverage, String hostname) {
+  public JobWorkerHealth(long workerId, double[] loadAverage, int taskPoolSize, String hostname) {
```
gpang: NIT: update the javadoc for this parameter

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -84,7 +94,8 @@ public String getHostname() {
    */
   public alluxio.grpc.JobWorkerHealth toProto() {
     alluxio.grpc.JobWorkerHealth.Builder builder = alluxio.grpc.JobWorkerHealth.newBuilder()
-        .setWorkerId(mWorkerId).addAllLoadAverage(mLoadAverage).setLastUpdated(mLastUpdated);
+        .setWorkerId(mWorkerId).addAllLoadAverage(mLoadAverage).setTaskPoolSize(mTaskPoolSize)
```
gpang: is there a unittest for this method?

tests/src/test/java/alluxio/job/master/JobMasterIntegrationTest.java
```diff
@@ -142,4 +148,14 @@ public void getAllWorkerHealth() throws Exception {
     assertNotNull(workerHealth.getHostname());
     assertEquals(3, workerHealth.getLoadAverage().size());
   }
+
+  @Test
+  public void testThreadPoolSize() throws Exception {
+    mJobMaster.setTaskPoolSize(0);
+    long jobId = mJobMaster.run(new SleepJobConfig(1));
+    CommonUtils.sleepMs(300);
```
gpang: why is the sleep necessary?

tests/src/test/java/alluxio/job/master/JobMasterIntegrationTest.java
```diff
@@ -76,6 +78,7 @@ public void after() throws Exception {
     mLocalAlluxioJobCluster.stop();
   }
 
+  @Ignore
```
gpang: why are all of these ignored?

job/server/src/main/java/alluxio/master/job/command/CommandManager.java
```diff
@@ -90,8 +88,27 @@ public synchronized void submitCancelTaskCommand(long jobId, long taskId, long w
...
+
+    JobCommand.Builder command = JobCommand.newBuilder();
+    command.setSetTaskPoolSizeCommand(setTaskPoolSizeCommand);
+    submit(workerId, command);
+  }
+
+  private void submit(long workerId, JobCommand.Builder command) {
     if (!mWorkerIdToPendingCommands.containsKey(workerId)) {
-      mWorkerIdToPendingCommands.put(workerId, Lists.<JobCommand>newArrayList());
+      mWorkerIdToPendingCommands.put(workerId, Lists.newArrayList());
```
gpang: Is this already protected by a lock?

job/server/src/main/java/alluxio/worker/job/task/ThrottleableThreadPoolExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+        mThrottleReduced.await();
+      }
+    } catch (InterruptedException e) {
+      t.interrupt();
+    } finally {
+      mThrottleLock.unlock();
+    }
+  }
+
+  private boolean check(Thread t) {
```
gpang: can we rename this to be more descriptive, maybe like "isRunnable" or if reversed "isThrottled"?

job/server/src/main/java/alluxio/worker/job/task/ThrottleableThreadPoolExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+      if (prevThrottlePercentage > mThrottlePercentage) {
+        mThrottleReduced.signalAll();
+      }
+    } finally {
+      mThrottleLock.unlock();
+    }
+  }
+
+  @Override
+  protected void beforeExecute(Thread t, Runnable r) {
```
gpang: Does this mean if a thread already started, it will still continue to run, even if the throttle percentage increases?

job/server/src/main/java/alluxio/worker/job/task/ThrottleableThreadPoolExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+    super.beforeExecute(t, r);
+    mThrottleLock.lock();
+    try {
+      while (!check(t)) {
+        mThrottleReduced.await();
```
gpang: could this cause some sort of deadlock where a running threads depends on a throttled thread?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -40,7 +42,9 @@
 public class TaskExecutorManager {
   private static final Logger LOG = LoggerFactory.getLogger(TaskExecutorManager.class);
 
-  private final ExecutorService mTaskExecutionService;
+  private static final int MAX_TASK_EXECUTOR_POOL_SIZE = 100;
```
gpang: should this be a configurable parameter? Also, should the default be larger?

tests/src/test/java/alluxio/job/master/JobMasterIntegrationTest.java
bradyoo: Oops :/

tests/src/test/java/alluxio/job/master/JobMasterIntegrationTest.java
bradyoo: The sleep is to show that this job never completes despite the waiting.

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
bradyoo: There is just an integration test for this. Let me add a unit test

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
bradyoo: This should have been fixed.

job/server/src/main/java/alluxio/worker/job/task/ThrottleableThreadPoolExecutor.java
bradyoo: This file was meant to be deleted. I used to use this but now I don't.

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
bradyoo: I can move this to something ridiculously large. There is no need to configure this because this can be very very large. I can move it to like a 1000 so that we wouldn't have to care.

job/server/src/main/java/alluxio/master/job/command/CommandManager.java
bradyoo: Every other public method caller to this is synchronized so since it doesn't make a difference, let me add synchronized to here as well.

bradyoo:
feedback addressed

bradyoo:
I changed the design of this and it's no longer working the way I think it should.

bradyoo:
Addressed some feedback. But more importantly, implementation changed quite a bit.

bradyoo:
changed description and fixed that failed test.

job/common/src/main/java/alluxio/job/wire/JobWorkerHealth.java
```diff
@@ -80,16 +106,55 @@ public String getHostname() {
   }
 
   /**
-   * @return proto representation of JobWorkerInfo
+   * @return proto representatx`ion of JobWorkerInfo
```
gpang: revert this change

tests/src/test/java/alluxio/job/master/JobMasterIntegrationTest.java
gpang: Is there a way to not depend on sleeps for testing? I would like for integration tests to complete in less time...

job/server/src/main/java/alluxio/worker/job/task/PausableThreadPoolExecutor.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * ThreadPoolExecutor that can be temporarily paused to prevent any threads from
+ * executing new tasks.
+ */
+public class PausableThreadPoolExecutor extends ThreadPoolExecutor {
+
+  private boolean mIsPaused;
+  private int mNumPaused;
```
gpang: can you comment on these variables? Especially, I think access has to be protected by the `mPauseLock`?

job/server/src/main/java/alluxio/worker/job/task/PausableThreadPoolExecutor.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  protected void beforeExecute(Thread t, Runnable r) {
+    super.beforeExecute(t, r);
+    mPauseLock.lock();
+    try {
+      mNumPaused++;
+      while (mIsPaused) {
+        mUnpaused.await();
+      }
+      mNumPaused--;
```
gpang: does this not have to be in the `finally` block?

job/server/src/main/java/alluxio/worker/job/task/PausableThreadPoolExecutor.java
```diff
@@ -0,0 +1,101 @@
...
+    mNumPaused = 0;
+    mPauseLock =  new ReentrantLock();
+    mUnpaused = mPauseLock.newCondition();
+  }
+
+  /**
+   * @return number of active threads subtracted by the number of tasks paused
+   */
+  public int getNumActiveTasks() {
+    return super.getActiveCount() - mNumPaused;
```
gpang: Is it correct that `mPauseLock` is not protecting `mNumPaused`?

bradyoo: Access to mNumPaused is not protected but that's okay. Because protect it or not, the value of mNumPaused can be inaccurate. But being slightly inaccurate here is okay.

bradyoo:
I took a look at the test times. It appears that the before block of launching is taking longer than the test itself. So it's better for me to spend time merging tests together then it is to try to remove sleeps here.

bradyoo:
Feedback addressed

bradyoo:
Ping?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -61,11 +64,43 @@ public TaskExecutorManager(int taskExecutorPoolSize, WorkerNetAddress address) {
...
+   */
+  public int getNumActiveTasks() {
+    return mTaskExecutionService.getNumActiveTasks();
+  }
+
+  /**
+   * @return task executor pool size
+   */
+  public int getTaskExecutorPoolSize() {
+    return mTaskExecutionService.getCorePoolSize();
```
calvinjia: This will always return `taskExecutorPoolSize` (# of core threads), is that what we want?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -61,11 +64,43 @@ public TaskExecutorManager(int taskExecutorPoolSize, WorkerNetAddress address) {
...
+
+  /**
+   * @param taskExecutorPoolSize number of threads in the task executor pool
+   */
+  public void setTaskExecutorPoolSize(int taskExecutorPoolSize) {
+    Preconditions.checkArgument(taskExecutorPoolSize >= 0);
+    Preconditions.checkArgument(taskExecutorPoolSize <= MAX_TASK_EXECUTOR_POOL_SIZE);
+
+    if (taskExecutorPoolSize == 0) {
+      // treat 0 as a schedule case because ThreadedTaskExecutorService can't seem to have 0 threads
```
calvinjia: `schedule` -> `special`?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -61,11 +64,43 @@ public TaskExecutorManager(int taskExecutorPoolSize, WorkerNetAddress address) {
...
+      mTaskExecutionService.pause();
+    } else {
+      mTaskExecutionService.resume();
```
calvinjia: Why do we need to call resume here?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
bradyoo: ?? Doesn't that make sense it's getTaskExecutorPoolSize.

job/server/src/main/java/alluxio/worker/job/task/PausableThreadPoolExecutor.java
gpang: Ok. Does it have to be `volatile` then? I'm not entirely sure how "stale" a non-volatile variable might be.

bradyoo:
addressed feedback

bradyoo:
[ERROR] Increment of volatile field alluxio.worker.job.task.PausableThreadPoolExecutor.mNumPaused in alluxio.worker.job.task.PausableThreadPoolExecutor.beforeExecute(Thread, Runnable) [alluxio.worker.job.task.PausableThreadPoolExecutor] At PausableThreadPoolExecutor.java:[line 96] VO_VOLATILE_INCREMENT
[ERROR] Increment of volatile field alluxio.worker.job.task.PausableThreadPoolExecutor.mNumPaused in alluxio.worker.job.task.PausableThreadPoolExecutor.beforeExecute(Thread, Runnable) [alluxio.worker.job.task.PausableThreadPoolExecutor] At PausableThreadPoolExecutor.java:[line 103] VO_VOLATILE_INCREMENT
[INFO] 

@gpang: You sure this is the right idea? Actually, I'll just use the lock.

bradyoo:
addressed feedback.

bradyoo:
alluxio-bot, test this please

bradyoo:
Can we merge this?

gpang:

PR #10507, opened 2019-11-23
Add UnfinishedTasks to JobWorkerHealth

bradyoo:
But only print UnfinishedTasks to jobservice report.

➜  alluxio git:(queue_length) ✗ ./bin/alluxio fsadmin report jobservice
Worker: 10.0.0.65   Unfinished Tasks: 1      Load Avg: 5.7861328125, 3.99169921875, 3.75390625

Status: CREATED   Count: 128
Status: CANCELED  Count: 0
Status: FAILED    Count: 0
Status: RUNNING   Count: 1
Status: COMPLETED Count: 1477

- Note here that unfinished task of a worker only contains jobs that are in "RUNNING" but not in "CREATED". 
- The Unfinished Tasks here is not very interesting because there is only one worker and not many running.

### DIFF
--- a/job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
+++ b/job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
@@ -12,20 +12,24 @@
 package alluxio.worker.job.command;
 
 import oshi.SystemInfo;
-import oshi.hardware.CentralProcessor;
+import oshi.hardware.HardwareAbstractionLayer;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.DoubleStream;
 
 /**
  * The job worker health reporter.
  */
 public class JobWorkerHealthReporter {
 
-  CentralProcessor mProcessor;
+  HardwareAbstractionLayer mHardware;
 
   /**
    * Default constructor.
    */
   public JobWorkerHealthReporter() {
-    mProcessor = new SystemInfo().getHardware().getProcessor();
+    mHardware = new SystemInfo().getHardware();
   }
 
   /**
@@ -34,7 +38,8 @@ public class JobWorkerHealthReporter {
    *
    * @return the system load average of the worker
    */
-  public double[] getCpuLoadAverage() {
-    return mProcessor.getSystemLoadAverage(3);
+  public List<Double> getCpuLoadAverage() {
+    return DoubleStream.of(mHardware.getProcessor().getSystemLoadAverage(3)).boxed()
+        .collect(Collectors.toList());
   }
 }
--- a/core/transport/src/grpc/job_master.proto
+++ b/core/transport/src/grpc/job_master.proto
@@ -65,6 +65,7 @@ message JobWorkerHealth {
   optional string hostname = 4;
   optional int32 taskPoolSize = 5;
   optional int32 numActiveTasks = 6;
+  optional int32 unfinishedTasks = 7;
 }
 
 message JobCommand {
--- a/core/transport/src/main/java/alluxio/grpc/JobMasterProto.java
+++ b/core/transport/src/main/java/alluxio/grpc/JobMasterProto.java
@@ -211,85 +211,85 @@ public final class JobMasterProto {
       "\0132\031.alluxio.grpc.job.JobInfo\0221\n\016recentFa" +
       "ilures\030\003 \003(\0132\031.alluxio.grpc.job.JobInfo\022" +
       "1\n\016longestRunning\030\004 \003(\0132\031.alluxio.grpc.j" +
-      "ob.JobInfo\"\215\001\n\017JobWorkerHealth\022\020\n\010worker" +
+      "ob.JobInfo\"\246\001\n\017JobWorkerHealth\022\020\n\010worker" +
       "Id\030\001 \001(\003\022\023\n\013loadAverage\030\002 \003(\001\022\023\n\013lastUpd" +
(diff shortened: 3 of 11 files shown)


bradyoo:
@ZacBlanco: This is probably another easy one.

bradyoo:
Ping?

job/common/src/main/java/alluxio/job/wire/NetworkHealth.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * The network health information.
+ */
+public class NetworkHealth {
+
+  private String mName;
+  private long mBytesRecv;
+  private long mBytesSent;
+  private long mSpeed;
```
ZacBlanco: can these all be final?

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
```diff
@@ -34,7 +41,17 @@ public JobWorkerHealthReporter() {
...
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * @return Network health for each of the network interfaces of the worker
+   */
+  public List<NetworkHealth> getNetworkHealths() {
+    return Arrays.stream(mHardware.getNetworkIFs()).map((networkIF) ->
+        new NetworkHealth(networkIF.getName(), networkIF.getBytesRecv(), networkIF.getBytesSent(),
+            networkIF.getSpeed())).collect(Collectors.toList());
```
ZacBlanco: Is this network speed the max connection speed or average?

My main concern is that I'm not sure if having this information is useful if we can't tell which network interface our traffic is traveling through.

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
```diff
@@ -34,7 +41,17 @@ public JobWorkerHealthReporter() {
    *
    * @return the system load average of the worker
    */
-  public double[] getCpuLoadAverage() {
-    return mProcessor.getSystemLoadAverage(3);
+  public List<Double> getCpuLoadAverage() {
+    return DoubleStream.of(mHardware.getProcessor().getSystemLoadAverage(3)).boxed()
+        .collect(Collectors.toList());
+  }
```
ZacBlanco: is there a reason you wanted to change this now?

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
bradyoo: My current thought is that I'm going to use the network interface with the highest speed/bytesRecv, etc.

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
bradyoo: Nope. I was realizing I'm using List objects in the other metric as well.

bradyoo:
Feedback addressed.

bradyoo:
Ping?

job/server/src/main/java/alluxio/worker/job/command/JobWorkerHealthReporter.java
ZacBlanco: How do you plan on "choosing" the network interface?

bradyoo: The way I'm considering this is that I'm not necessarily going to choose a network interface but rather just look at the fact that overall traffic is high or in a different case, look at the difference when there are no active tasks running.

ZacBlanco: how do you plan on tracking all of this information?

bradyoo: After discussing with Calvin, both me and him agree that in practice we only expect to see one network interface to be used. (By that, we mean that everything except for one will have very little bytes recv/sent). So it will be very easy to later filter out for actual logic.

ZacBlanco: I guess my main concern is that the metrics that this PR adds just don't seem useful in most situations.

Currently, it looks like the worker collects the bare minimum (if speed, bytes sent/recv) and then sends them to the master. But the master can't really do anything except relay the information to a user. And even then, it is unlikely that the information about the interface is really of any use to the user.

I guess this PR does at least set up the scaffolding for adding better metrics in the future. One thing that I think would be useful is to have 1/5/15 minute averages for bytes sent/sec recv/sec.

That can be left for another PR though

bradyoo:
Can you merge this for me? @ZacBlanco

bradyoo:
merging this with other commits is getting a bit annoying. 

@ZacBlanco: I'll remove every bit about the NetworkHealth from proto, in that case, are you willing to merge?

ZacBlanco:
Just re-generate the proto that this PR introduced and I will merge it

bradyoo:
There are a few other changes that conflict. (I know what they are.)

bradyoo:
Rebased and removed NetworkHealth from wire (but left them on Reporter)

ZacBlanco:
Can you update the title to accurately reflect the changes now that the network info is removed?

bradyoo:
I did but I will remove some more I guess.

bradyoo:
done

bradyoo:
Mind merging?

ZacBlanco:
whoops thought I did yesterday. My bad

ZacBlanco:

PR #10500, opened 2019-11-22
Make rejecting server listen on correct address

Hollyen:
image

Rejecting server is listening on 0.0.0.0, this is possibly cause some security issue.

PR #10508, opened 2019-11-24
Support using alluxio default block size when UFS is hadoop-compatible object storage filesystem

Hollyen:
### Background
Alluxio has 2 ways to use object storage as UFS.

 - Alluxio native interface: `alluxio.underfs.ObjectUnderFileSystem`
 - Hadoop-compatible FileSystem interface: `org.apache.hadoop.fs.FileSystem`

About block size, in `ObjectUnderFileSystem`, it is default using alluxio default block size (*alluxio.user.block.size.bytes.default*). In hadoop filesystem case, it is depend on different FileSystem implementation. For example, `S3AFileSystem` take configuration 'fs.s3a.block.size' as blocksize.
So when i use hadoop filesystem to access S3, i need configure *fs.s3a.block.size* not exact *alluxio.user.block.size.bytes.default* . 

This patch allows user config block size by *alluxio.user.block.size.bytes.default* even using `S3AFileSystem` as UFS.

### Config

in S3A case, ```fs.s3a.block.size=1```, `1` is magic number block size. which hint Alluxio to use `alluxio.user.block.size.bytes.default` as default block size.

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UfsFileStatus.java
+++ b/core/common/src/main/java/alluxio/underfs/UfsFileStatus.java
@@ -24,6 +24,9 @@ public class UfsFileStatus extends UfsStatus {
   public static final String INVALID_CONTENT_HASH = "";
   public static final long UNKNOWN_BLOCK_SIZE = -1;
 
+  // if block size equals to MAGIC_BLOCK_SIZE, use alluxio.user.block.size.bytes.default instread
+  public static final long MAGIC_BLOCK_SIZE = 1;
+
   protected final String mContentHash;
   protected final long mContentLength;
   protected final long mBlockSize;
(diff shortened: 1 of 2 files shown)


ZacBlanco:
Since you've signed the CLA in https://github.com/Alluxio/alluxio/pull/10500 I'll remove the label here

core/common/src/main/java/alluxio/underfs/UfsFileStatus.java
```diff
@@ -24,6 +24,9 @@
   public static final String INVALID_CONTENT_HASH = "";
   public static final long UNKNOWN_BLOCK_SIZE = -1;
 
+  // if block size equals to MAGIC_BLOCK_SIZE, use alluxio.user.block.size.bytes.default instread
+  public static final long MAGIC_BLOCK_SIZE = 1;
```
ZacBlanco: Since `1` is technically a valid number of bytes that someone could want to configure, what if we use `-1` or a negative value instead? It wouldn't make sense in any other case to configure a negative value, so I feel the magic number should be at least `<= 0`

Hollyen: please fine code in [S3AFileSystem](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java), `-1` is invalid number for s3 filesystem, the minimal number is `1`.

Another option is `Long.MAX_VALUE`, but still have same issue as you mentioned.

Hollyen: @ZacBlanco any suggestions?

PR #10513, opened 2019-11-25
Add CachingFileSystem

chcn:
1. in fuse openFile, a getStatus RPC is issued, but openFile will also issue a getStatus RPC internally, this redundant RCP is reduced in this PR.
2. when getting status of a file, or listing status of a directory, cache the status results so that future getStatus and listStatus calls do not need to issue RPCs to master.

This PR introduces a new class `CachingFileSystem extends BaseFileSystem` to cache path metadata.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
@@ -21,15 +21,15 @@ import alluxio.client.block.BlockWorkerInfo;
 import alluxio.client.file.FileSystemContextReinitializer.ReinitBlockerResource;
 import alluxio.client.file.options.InStreamOptions;
 import alluxio.client.file.options.OutStreamOptions;
-import alluxio.collections.Pair;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.PropertyKey;
 import alluxio.exception.AlluxioException;
 import alluxio.exception.DirectoryNotEmptyException;
-import alluxio.exception.ExceptionMessage;
 import alluxio.exception.FileAlreadyExistsException;
 import alluxio.exception.FileDoesNotExistException;
+import alluxio.exception.FileIncompleteException;
 import alluxio.exception.InvalidPathException;
+import alluxio.exception.OpenDirectoryException;
 import alluxio.exception.status.AlluxioStatusException;
 import alluxio.exception.status.AlreadyExistsException;
 import alluxio.exception.status.FailedPreconditionException;
@@ -77,9 +77,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -95,10 +92,8 @@ public class BaseFileSystem implements FileSystem {
   protected final FileSystemContext mFsContext;
   protected final AlluxioBlockStore mBlockStore;
   protected final boolean mCachingEnabled;
-  private final MetadataCache mMetadataCache;
-  private final ExecutorService mAccessTimeUpdater;
 
-  private volatile boolean mClosed = false;
+  protected volatile boolean mClosed = false;
 
   /**
    * @param context the {@link FileSystemContext} to use for client operations
@@ -127,15 +122,6 @@ public class BaseFileSystem implements FileSystem {
     mFsContext = fsContext;
     mBlockStore = AlluxioBlockStore.create(fsContext);
     mCachingEnabled = cachingEnabled;
-    if (mFsContext.getClusterConf().getBoolean(PropertyKey.USER_METADATA_CACHE_ENABLED)) {
-      int maxSize = mFsContext.getClusterConf().getInt(PropertyKey.USER_METADATA_CACHE_MAX_SIZE);
-      long expirationTimeMs = mFsContext.getClusterConf()
-          .getMs(PropertyKey.USER_METADATA_CACHE_EXPIRATION_TIME);
-      mMetadataCache = new MetadataCache(this, maxSize, expirationTimeMs);
-    } else {
-      mMetadataCache = null;
-    }
-    mAccessTimeUpdater = Executors.newCachedThreadPool();
   }
 
   /**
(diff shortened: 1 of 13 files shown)


core/client/fs/src/main/java/alluxio/client/file/MetadataCache.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.common.cache.CacheBuilder;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Cache for metadata of files.
+ */
+public final class MetadataCache {
```
apc999: Is this class thread safe? In case it is , annotate the class by `@ThreadSafe`.

apc999: also, can you add a unit test on this class?

core/client/fs/src/main/java/alluxio/client/file/MetadataCache.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * If file status is cached, return the cached status.
+   * Otherwise, issue an RPC to master to get and cache the status.
+   *
+   * @param file the file
+   * @param options the options
+   * @return the file status and a boolean indicating whether the status comes from the cache
+   */
+  public Pair<URIStatus, Boolean> getStatus(AlluxioURI file, GetStatusPOptions options)
```
apc999: any reason to return `Pair<URIStatus, Boolean>` rather than Nullable `URIStatus` or `Optional<URIStatus>`?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3306,6 +3306,40 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_DIRECT_MEMORY_IO_ENABLED =
```
apc999: not used. please remove

chcn:
@apc999 PTAL

chcn:
Refactored out `CachingFileSystem`, @apc999 @calvinjia PTAL

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -543,11 +579,30 @@ public void unmount(AlluxioURI path, UnmountPOptions options)
     });
   }
 
+  private void updateFileAccessTime(AlluxioURI path) throws IOException, AlluxioException {
```
calvinjia: This logic should not be in `BaseFileSystem`?

chcn: it's for the new `openFile(URIStatus)` API

calvinjia: This functionality doesn't feel like it belongs in the base client because the base file system shouldn't support the `openFile(URIStatus,...)` API. Instead, could we only add this new API to `CachingFileSystem`?

chcn: this API is added to the FileSystem interface to support the following calling pattern:
```
status = getStatus;
// do something with status
stream = openFile(status);
```
This pattern is seen in `AlluxioFuseFileSystem`, and it's orthogonal to caching.

calvinjia: I'm against introducing this API to the file system
1. We already have openFile which handles getting the file status, this sounds like an optimization for when the file status is already available (ie. cached)
2. It adds complexity to the base file system class by introducing async operations
3. This could easily be included in another file system type (that extends BaseFileSystem) and provide the same benefits to users

chcn: Refactored exception handling of `openFile(AlluxioURI)`, so that `AlluxioFuseFileSystem` gets rid of `openFile(Status)`.

chcn:
@calvinjia PTAL

chcn:
@calvinjia @apc999 PTAL

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -310,11 +311,37 @@ public URIStatus getStatus(AlluxioURI path)
...
+   * then update the path's last access time in the getStatus RPC.
+   *
+   * @param path the path
+   * @param options the open file options
+   * @return the status for path
+   * @throws FileDoesNotExistException when path does not exist
+   * @throws OpenDirectoryException when path is a directory
+   * @throws FileIncompleteException when path is a file and is not completed yet
+   */
+  protected URIStatus getStatus(AlluxioURI path, OpenFilePOptions options)
```
calvinjia: Do we need this variant of `getStatus` implemented in `BaseFileSystem` (or ever?), could the caller create the getStatus options and check for the two new exceptions?

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -400,30 +427,27 @@ public void persist(final AlluxioURI path, final ScheduleAsyncPersistencePOption
...
-      if (status.isFolder()) {
-        throw new FileDoesNotExistException(
-            ExceptionMessage.CANNOT_READ_DIRECTORY.getMessage(status.getName()));
-      }
-      OpenFilePOptions mergedOptions = FileSystemOptions.openFileDefaults(conf)
-          .toBuilder().mergeFrom(options).build();
-      InStreamOptions inStreamOptions = new InStreamOptions(status, mergedOptions, conf);
-      return new FileInStream(status, inStreamOptions, mFsContext);
-    });
+    URIStatus status = getStatus(path, options);
```
calvinjia: Should we check the result of `status.isFolder` and `status.isCompleted` here?

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -400,30 +427,27 @@ public void persist(final AlluxioURI path, final ScheduleAsyncPersistencePOption
...
+    return openFileInStream(path, status, options);
+  }
+
+  protected FileInStream openFileInStream(AlluxioURI path, URIStatus status,
```
calvinjia: I don't think we should break up the method in this class in order to better modularize for a class that inherits from this one.

chcn:
@calvinjia sure, they can all be implemented in CachingFileSystem, but then the logic in `getStatus(AlluxioURI, OpenFilePOptions)` and `openFileInStream` will be duplicated in both implementations. These functions can be viewed as utility functions exposed by `BaseFileSystem` for extended classes, what's your concern about exposing them?

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
chcn: it's just a utility function exposed for reducing duplicated codes

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
chcn: this function is for getStatus in openFile explicitly, otherwise, the exception checking code will be duplicated.

maybe can change the name to getAndCheckStatus though.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
chcn: this is just a utility function, without it, any extension to `BaseFileSystem` that wants to override `openFile` need to duplicate this logic.

chcn:
@calvinjia refactored `BaseFileSystem#openFile` so that `CachingFileSystem` does not need to override `openFile`, much cleaner.

chcn:
jenkins, test this please

calvinjia:

calvinjia:

PR #10518, opened 2019-11-26
Add metric master test [WIP]

LuQQiu:
Don't merge, for testing

### DIFF
--- a/tests/src/test/java/alluxio/server/health/MetricMasterIntegrityIntegrationTest.java
+++ b/tests/src/test/java/alluxio/server/health/MetricMasterIntegrityIntegrationTest.java
@@ -1,0 +1,108 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.server.health;
+
+import alluxio.master.LocalAlluxioCluster;
+import alluxio.master.metrics.MetricsMaster;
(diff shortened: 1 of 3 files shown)


tests/src/test/java/alluxio/server/health/MetricMasterIntegrityIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+    long mid = System.currentTimeMillis();
+    System.out.println("submit task takes " + (mid - start));
+    for (Future<?> future : futureList) {
+      future.get();
+    }
+    System.out.println("get result takes " + (System.currentTimeMillis() - mid));
+  }
+
+  @Test
+  public void testMultipleWorkersOneHeartbeat() throws Exception {
```
LuQQiu: @gpang For this test, I test with numWorkers = 50/200/1000/3000 cases and executor thread= 10/50.
For all the tests, in master branch, the throughput is 1000-3000 heartbeats/second. After my changes in https://github.com/Alluxio/alluxio/pull/10493, the throughput is 2300 to 5100 heartbeats/second. The throughput nearly doubled. 

I cannot get the real circumstances we faced before.  one of my guesses is that the bottleneck is memory other than the synchronization lock. @ZacBlanco Mentioned before that the `Metric.getFullMetricName` do many string concatenation and cause many memory copies. This function may use a large amount memory in reality.

LuQQiu:
```
  /**
   * @return the fully qualified metric name, which is of pattern
   *         instance.[hostname-id:instanceId.]name[.tagName:tagValue]*, where the tags are appended
   *         at the end
   */
  public String getFullMetricName() {
    StringBuilder sb = new StringBuilder();
    sb.append(mInstanceType).append('.');
    if (mHostname != null) {
      sb.append(mHostname);
      if (mInstanceId != null) {
        sb.append(ID_SEPARATOR).append(mInstanceId);
      }
      sb.append('.');
    }

    sb.append(mName);
    for (Entry<String, String> entry : mTags.entrySet()) {
      sb.append('.').append(entry.getKey()).append(TAG_SEPARATOR).append(entry.getValue());
    }
    return sb.toString();
  }
```

PR #10525, opened 2019-11-27
Parallelize MigrateDefinition within a worker

bradyoo:
Using some of the API changes from before to make distributedCp and distributedMv faster (and take more resources).
The same logic can be applied to PersistCommand as well.

Test scenario: (r4.xlarge)
1. 4000 Files each with 100KB.
2. 2 Workers.
3. Deployed https://github.com/bradyoo/alluxio/pull/4/files for testing to see how different parallel factors impact this.
4. Async Through

Results:

The last parameter is the parallel factor: Which is not in this PR because this PR simply defaults that to 10 where as in this result, if you don't provide a parameter, it defaults to 1 (old behavior)

This is with alluxio.job.worker.threadpool.size set to 20. (Default is 10.)

$ time ./bin/alluxio fs distributedCp /async-write2/ /async-write20/
.................Copied /async-write2 to /async-write20

real	0m37.183s
user	0m3.613s
sys	0m0.238s

$ time ./bin/alluxio fs distributedCp /async-write2/ /async-write21/ 10
...Copied /async-write2 to /async-write21

real	0m9.066s
user	0m3.365s
sys	0m0.230s

$ time ./bin/alluxio fs distributedCp /async-write2/ /async-write22/ 2
..........Copied /async-write2 to /async-write22

real	0m23.179s
user	0m3.779s
sys	0m0.233s

$ time ./bin/alluxio fs distributedCp /async-write2/ /async-write23/
.................Copied /async-write2 to /async-write23

real	0m37.229s
user	0m3.647s
sys	0m0.251s

$ time ./bin/alluxio fs distributedCp /async-write2/ /async-write24/ 15
...Copied /async-write2 to /async-write24

real	0m8.059s
user	0m3.366s
sys	0m0.235s

### DIFF
--- a/job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
+++ b/job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
@@ -203,8 +203,15 @@ public final class MigrateDefinition
 
       // split the list of MigrateCommands for a given worker into at most JOBS_PER_WORKER
       // equally sized lists
-      List<List<MigrateCommand>> splittedCommands = Lists.partition(assignment.getValue(),
-          Math.round(assignment.getValue().size() * 1.0f / JOBS_PER_WORKER + 0.5f));
+      ArrayList<MigrateCommand> migrateCommands = assignment.getValue();
+      ArrayList<ArrayList<MigrateCommand>> splittedCommands = Lists.newArrayList();
+
+      for (int i = 0; i < JOBS_PER_WORKER; i++) {
+        splittedCommands.add(Lists.newArrayList());
+      }
+      for (int i = 0; i < migrateCommands.size(); i++) {
+        splittedCommands.get(i % JOBS_PER_WORKER).add(migrateCommands.get(i));
+      }
 
       for (List<MigrateCommand> commands : splittedCommands) {
         if (!commands.isEmpty()) {
--- a/job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
+++ b/job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
@@ -202,22 +202,11 @@ public final class MigrateDefinition
     for (Map.Entry<WorkerInfo, ArrayList<MigrateCommand>> assignment : assignments.entrySet()) {
 
       // split the list of MigrateCommands for a given worker into JOBS_PER_WORKER.
+      List<List<MigrateCommand>> splittedCommands = Lists.partition(assignment.getValue(), JOBS_PER_WORKER);
 
-      ArrayList<MigrateCommand> migrateCommands = assignment.getValue();
-
-      ArrayList<ArrayList<MigrateCommand>> splittedCommands = Lists.newArrayList();
-
-      for (int i = 0; i < JOBS_PER_WORKER; i++) {
-        splittedCommands.add(Lists.newArrayList());
-      }
-
-      for (int i = 0; i < migrateCommands.size(); i++) {
-        splittedCommands.get(i % JOBS_PER_WORKER).add(migrateCommands.get(i));
-      }
-
-      for (ArrayList<MigrateCommand> commands : splittedCommands) {
+      for (List<MigrateCommand> commands : splittedCommands) {
         if (!commands.isEmpty()) {
-          result.add(new Pair<>(assignment.getKey(), commands));
+          result.add(new Pair<>(assignment.getKey(), Lists.newArrayList(commands)));
         }
       }
     }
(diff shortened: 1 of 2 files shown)


gpang:
Jenkins, test this please

bradyoo:
@calvinjia do you want to take a look at this?

calvinjia:
@bradyoo Thanks for the improvement. Should the parallelism be configurable? Also do you know why you only see ~4x improvement (does your machine only have 4 cores)?

bradyoo:
These were on a r4.xlarge instance which has a vCPU of 4.

bradyoo:
@calvinjia: I plan to ship with just 10 for now because I think this is a safe enough low-ish number with an idea of making it configurable (to let users go to a higher number).

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
```diff
@@ -197,7 +200,26 @@ private void checkMigrateValid(MigrateConfig config, FileSystem fs) throws Excep
...
+
+      ArrayList<ArrayList<MigrateCommand>> splittedCommands = Lists.newArrayList();
+
+      for (int i = 0; i < JOBS_PER_WORKER; i++) {
+        splittedCommands.add(Lists.newArrayList());
+      }
+
+      for (int i = 0; i < migrateCommands.size(); i++) {
+        splittedCommands.get(i % JOBS_PER_WORKER).add(migrateCommands.get(i));
+      }
```
calvinjia: Do you mind removing some of the white space? Also, would `Lists.partition` work here?

bradyoo: Oh I think Lists.partition is probably great.

bradyoo:
Addressed feedback

bradyoo:
Sorry I forgot to push. Pushed again.

calvinjia:
@bradyoo Note that this implementation is slightly different from the previous one, instead of evenly splitting the tasks, we will try to fully fill as many workers as possible, do you see any issues with this?

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
```diff
@@ -197,7 +200,15 @@ private void checkMigrateValid(MigrateConfig config, FileSystem fs) throws Excep
     Set<Pair<WorkerInfo, ArrayList<MigrateCommand>>> result = Sets.newHashSet();
 
     for (Map.Entry<WorkerInfo, ArrayList<MigrateCommand>> assignment : assignments.entrySet()) {
-      result.add(new Pair<>(assignment.getKey(), assignment.getValue()));
+
+      // split the list of MigrateCommands for a given worker into JOBS_PER_WORKER.
+      List<List<MigrateCommand>> splittedCommands = Lists.partition(assignment.getValue(), JOBS_PER_WORKER);
```
calvinjia: Checkstyle

bradyoo:
Yep saw it.

bradyoo:
failed test will address.

bradyoo:
Lists.partition turned out doesn't actually do the behavior it did in that the second parameter was about size of the partitioned lists and not number of lists. I did some divisions and rounding to closely mimic the behavior I wanted.

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
```diff
@@ -201,9 +201,9 @@ private void checkMigrateValid(MigrateConfig config, FileSystem fs) throws Excep
 
     for (Map.Entry<WorkerInfo, ArrayList<MigrateCommand>> assignment : assignments.entrySet()) {
 
-      // split the list of MigrateCommands for a given worker into JOBS_PER_WORKER.
+      // split the list of MigrateCommands for a given worker into at most equally sized JOBS_PER_WORKER lists
       List<List<MigrateCommand>> splittedCommands = Lists.partition(assignment.getValue(),
-          JOBS_PER_WORKER);
+          Math.round(assignment.getValue().size() * 1.0f / JOBS_PER_WORKER + 0.5f));
```
calvinjia: Is this equivalent to using a ceiling function? Using partition causes the size of the lists to be more skewed. For example if you had 12 tasks, instead of doing [2, 2, 1, 1, 1, ...], it will do [2, 2, 2, 2, 2, 2]

bradyoo:
I tried using Ceil. The really annoying thing about Ceil is that ceil takes
in a double and returns a double and I couldn't find a good way to convert
doubles to ints without triggering Findbugs.

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
bradyoo: I went back to the old implementation.

calvinjia:

PR #10531, opened 2019-11-27
Use Jacoco offline instrumentation to track PowerMock tests

gpang:
On-the-fly Jacoco instrumentation does not work for tests which use PowerMock. Therefore, offline instrumentation must be used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1320,7 +1320,7 @@
       <id>jacoco</id>
       <dependencies>
         <dependency>
-          <!-- must be on the classpath -->
+          <!-- to use instrument, jacoco must be on the classpath -->
           <groupId>org.jacoco</groupId>
           <artifactId>org.jacoco.agent</artifactId>
           <classifier>runtime</classifier>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1325,14 +1336,22 @@
...
-                <id>jacoco-report</id>
+                <id>jacoco-restore-instrumented-classes</id>
                 <phase>test</phase>
+                <goals>
+                  <goal>restore-instrumented-classes</goal>
+                </goals>
+              </execution>
+              <execution>
+                <id>jacoco-report</id>
+                <phase>verify</phase>
```
ZacBlanco: `report` was previously part of the `test` phase. What's the reason that you changed it?

gpang: hrmmm, at some point I needed it to specify it in a phase after `test`, but I think it got fixed by setting the `test` phase for `restore-instrumented-classes`. I just removed the phase and it still seems to work.

gpang:

PR #10532, opened 2019-11-27
Implement support for terminating cancelled RPCs

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/10544

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3051,6 +3051,10 @@ public final class DefaultFileSystemMaster extends CoreMaster
     if (context.getOptions().getRecursive()) {
       try (LockedInodePathList descendants = mInodeTree.getDescendants(inodePath)) {
         for (LockedInodePath childPath : descendants) {
+          // Fail if the client has cancelled the rpc.
+          if (context.isCancelled()) {
+            throw new RuntimeException("Call cancelled by the client.");
+          }
           setAclSingleInode(rpcContext, action, childPath, entries, replay, opTimeMs);
         }
       }
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -869,7 +869,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
       auditContext.setSrcInode(inode);
       DescendantType descendantTypeForListStatus =
           (context.getOptions().getRecursive()) ? DescendantType.ALL : DescendantType.ONE;
-      listStatusInternal(rpcContext, inodePath, auditContext, descendantTypeForListStatus,
+      listStatusInternal(context, rpcContext, inodePath, auditContext, descendantTypeForListStatus,
           resultStream, 0);
       auditContext.setSucceeded(true);
       Metrics.FILE_INFOS_GOT.inc();
@@ -889,18 +889,23 @@ public final class DefaultFileSystemMaster extends CoreMaster
    * descendantType. The result is returned via a list specified by statusList, in postorder
    * traversal order.
    *
+   * @param context call context
    * @param rpcContext the context for the RPC call
    * @param currInodePath the inode path to find the status
    * @param auditContext the audit context to return any access exceptions
    * @param descendantType if the currInodePath is a directory, how many levels of its descendant
-   *                       should be returned
+   *        should be returned
    * @param resultStream the stream to receive individual results
    * @param depth internal use field for tracking depth relative to root item
    */
-  private void listStatusInternal(RpcContext rpcContext, LockedInodePath currInodePath,
-      AuditContext auditContext, DescendantType descendantType, ResultStream<FileInfo> resultStream,
-      int depth) throws FileDoesNotExistException, UnavailableException, AccessControlException,
-      InvalidPathException {
+  private void listStatusInternal(ListStatusContext context, RpcContext rpcContext,
+      LockedInodePath currInodePath, AuditContext auditContext, DescendantType descendantType,
+      ResultStream<FileInfo> resultStream, int depth) throws FileDoesNotExistException,
+      UnavailableException, AccessControlException, InvalidPathException {
+    // Fail if the client has cancelled the rpc.
+    if (context.isCancelled()) {
+      throw new RuntimeException("Call cancelled by the client.");
+    }
     Inode inode = currInodePath.getInode();
     if (inode.isDirectory() && descendantType != DescendantType.NONE) {
       try {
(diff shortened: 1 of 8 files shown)


ggezer:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/contexts/GrpcCallTracker.java
```diff
@@ -0,0 +1,41 @@
...
+  private ServerCallStreamObserver<?> mStreamObserver;
+
+  /**
+   * Creates a call tracker for gRPC.
+   *
+   * @param streamObserver server stream observer
+   */
+  public GrpcCallTracker(StreamObserver<?> streamObserver) {
+    if (!(streamObserver instanceof ServerCallStreamObserver)) {
+      throw new IllegalStateException("Call tacking is supported for server streams.");
```
gpang: "tacking" -> "tracking"
"is supported" => "is only supported"

Also, can we also include the instance type of `streamObserver` in the message?

core/server/master/src/main/java/alluxio/master/file/contexts/GrpcCallTracker.java
```diff
@@ -0,0 +1,41 @@
...
+  /** Server-side gRPC stream observer. */
+  private ServerCallStreamObserver<?> mStreamObserver;
+
+  /**
+   * Creates a call tracker for gRPC.
+   *
+   * @param streamObserver server stream observer
+   */
+  public GrpcCallTracker(StreamObserver<?> streamObserver) {
+    if (!(streamObserver instanceof ServerCallStreamObserver)) {
```
gpang: Why is only stream observer supported? Is that the only one we can detect cancelation by the client?

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -22,14 +22,31 @@
...
 
   /**
    * Creates an instance with given proto message.
    *
    * @param optionsBuilder Internal proto message builder instance
    */
   public OperationContext(T optionsBuilder) {
+    this(optionsBuilder, null);
     mOptionsBuilder = optionsBuilder;
+    mCallTracker = () -> {
```
gpang: can we make this a static/constant noop tracker, to avoid re-creating it frequently?

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -22,14 +22,31 @@
 public class OperationContext<T extends com.google.protobuf.GeneratedMessageV3.Builder<?>> {
   // Proto message that is being wrapped
   private T mOptionsBuilder;
+  // Used to track client call status.
+  private CallTracker mCallTracker;
```
gpang: when will this be `null` and can this be `final`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -869,7 +869,7 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
       auditContext.setSrcInode(inode);
       DescendantType descendantTypeForListStatus =
           (context.getOptions().getRecursive()) ? DescendantType.ALL : DescendantType.ONE;
-      listStatusInternal(rpcContext, inodePath, auditContext, descendantTypeForListStatus,
+      listStatusInternal(context, rpcContext, inodePath, auditContext, descendantTypeForListStatus,
```
gpang: in general, why is the `isCancelled()` not part of the rpcContext? It would seem appropriate as part of an rpc context.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -889,18 +889,23 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
...
-      AuditContext auditContext, DescendantType descendantType, ResultStream<FileInfo> resultStream,
-      int depth) throws FileDoesNotExistException, UnavailableException, AccessControlException,
-      InvalidPathException {
+  private void listStatusInternal(ListStatusContext context, RpcContext rpcContext,
+      LockedInodePath currInodePath, AuditContext auditContext, DescendantType descendantType,
+      ResultStream<FileInfo> resultStream, int depth) throws FileDoesNotExistException,
+      UnavailableException, AccessControlException, InvalidPathException {
+    // Fail if the client has cancelled the rpc.
+    if (context.isCancelled()) {
+      throw new RuntimeException("Call cancelled by the client.");
```
gpang: Is `RuntimeException` the correct way to "cancel" an rpc thread? Also, for whichever exception we throw, can we include more in the message, like the type of call, and the path?

core/server/master/src/main/java/alluxio/master/file/contexts/GrpcCallTracker.java
ggezer: Correct. This is the only documented way.
Even unary calls send results using `StreamObserver`, so it's a way that covers all call types.

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
ggezer: In order to reduce amount of code change, I introduced another constructor here that sets the tracker. I can not make it final because of that.

It will be null, when contexts for calls that requires tracking is created with constructor that doesn't take tracker. (E.g unit tests, where we don't have a tracking facility).

Now, I created another static tracker for no-op tracker that always return false, and pass that instead of null. So this check is gone.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ggezer: `RpcContext` is a master side construct. I'd still need to pass `CallTracker` into FSM implementations then I could offload them into `RpcContext`. Since we already pass around op-contexts in most of the code, I am just calling them from there. This is the only part where we were not passing the op-context.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ggezer: What do you mean by correct? For the propagation, `RPCUtils` is prepared for `RuntimeException`s. Alternatively I'd define a new AlluxioStatusException type for cancelled calls and throw that instead.

ggezer: For adding more info to the messages, please check `RpcUtils`. It'll take care of putting all the details of the call.

gpang: I see, it logs a lot of useful things. I guess the actual exception message doesn't have to be descriptive, since the client will never get this information (since the client canceled and is gone)

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: I see. I thought it made sense for something called `RPCContext` to have a method like `isCancelled`, but we can always change it later.

gpang:
@ggezer just a minor comment, then this looks good

gpang:

ggezer:

PR #10536, opened 2019-12-01
Use correct properties for csv lines to skip

### LINKED ISSUE #10557: Transformation can fail when csv files has a header
**Alluxio Version:**
2.1.0

**Describe the bug**
When a table stores its file as a csv file and it has headers, its property `skip.header.line.count`  is not respected by the transformation and the transformation fails because of type mismatch. 

**To Reproduce**
attach a database with a table that has csv as its data and the csv has a header. 

**Expected behavior**
Expect the transformation to respect the header and the `skip.header.line.count` property

**Urgency**
fix before the next release

**Additional context**

yuzhu:
Previously we were looking for lines to skip in the wrong properties. (serde properties) now look for it in table properties. 

Fixes #10557

### DIFF
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
@@ -219,6 +219,7 @@ public class HiveDatabase implements UnderDatabase {
           .setTableName(tableName)
           .addAllDataCols(HiveUtils.toProto(table.getSd().getCols()))
           .setStorage(HiveUtils.toProto(table.getSd(), pathTranslator))
+          .putAllParameters(table.getParameters())
           // ignore partition name
           .build();
       Layout layout = Layout.newBuilder()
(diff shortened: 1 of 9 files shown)


yuzhu:
@chcn @gpang PTAL, thanks

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -219,6 +219,7 @@ public UdbTable getTable(String tableName) throws IOException {
           .setTableName(tableName)
           .addAllDataCols(HiveUtils.toProto(table.getSd().getCols()))
           .setStorage(HiveUtils.toProto(table.getSd(), pathTranslator))
+          .putAllParameters(table.getParameters())
```
gpang: This was missing from before? This is not related to the serde confusion, right? Is there any unittest that can cover this type of issue?

yuzhu: this is probably best tested using integration test / end to end test. Mocking a hive metaserver to test this has always had the problem of not being very accurate..

gpang:
@yuzhu if this is a bug, can you create an issue, and then link it in the PR description?

yuzhu:
@gpang issue created and referenced

yuzhu:

yuzhu:

PR #10537, opened 2019-12-02
Restore the hive client connection each time it is requested

yuzhu:
Slightly modified retry to support the following: 
- RetryException and CantRetryException
- When the function encounters RetryException, it retries. If it encounters CantRetryException, it unwraps the exception and throws it. 

- For concurrency protection, anytime we access it, we will grab a read lock, anytime we modify it we grab a write lock.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3782,7 +3782,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey TABLE_METASTORE_RETRY_TIMEOUT =
       new Builder(Name.TABLE_METASTORE_RETRY_TIMEOUT)
-          .setDescription("Retry period before Alluxio give up on connecting to metastore and exit.")
+          .setDescription("Retry period before Alluxio give up on connecting to"
+              + " metastore and exit.")
           .setDefaultValue("30 secs")
           .setScope(Scope.MASTER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3784,7 +3784,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       new Builder(Name.TABLE_METASTORE_RETRY_TIMEOUT)
           .setDescription("Retry period before Alluxio give up on connecting to"
               + " metastore and exit.")
-          .setDefaultValue("30 secs")
+          .setDefaultValue("30sec")
           .setScope(Scope.MASTER)
           .build();
 
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3780,6 +3780,12 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey TABLE_METASTORE_RETRY_TIMEOUT =
+      new Builder(Name.TABLE_METASTORE_RETRY_TIMEOUT)
+          .setDescription("Retry period before Alluxio give up on connecting to metastore and exit.")
+          .setDefaultValue("30 secs")
+          .setScope(Scope.MASTER)
+          .build();
 
   /**
    * @deprecated This key is used for testing. It is always deprecated.
@@ -4527,6 +4533,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.table.transform.manager.job.monitor.interval";
     public static final String TABLE_TRANSFORM_MANAGER_JOB_HISTORY_RETENTION_TIME =
         "alluxio.table.transform.manager.job.history.retention.time";
+    public static final String TABLE_METASTORE_RETRY_TIMEOUT =
+        "alluxio.table.metastore.retry.timeout";
 
     private Name() {} // prevent instantiation
   }
--- a/core/common/src/main/java/alluxio/retry/RetryUtils.java
+++ b/core/common/src/main/java/alluxio/retry/RetryUtils.java
@@ -56,7 +56,7 @@ public final class RetryUtils {
    * @param policy the retry policy to use
    * @return T return a value
    */
-  public static <T> T retry(String action, RunnableFunctionThrowsException<T> f, RetryPolicy policy)
+  public static <T> T retryFunction(String action, RunnableFunctionThrowsException<T> f, RetryPolicy policy)
       throws CantRetryException {
     RetryException e = null;
     while (policy.attempt()) {
--- a/core/common/src/main/java/alluxio/retry/RetryUtils.java
+++ b/core/common/src/main/java/alluxio/retry/RetryUtils.java
@@ -47,6 +47,32 @@ public final class RetryUtils {
   }
 
   /**
+   * Retries the given method until it doesn't throw an IO exception or the retry policy expires. If
+   * the retry policy expires, the last exception generated will be rethrown.
+   *
+   * @param action a description of the action that fits the phrase "Failed to ${action}"
+   * @param f the function to retry
+   * @param policy the retry policy to use
+   */
+  public static <T> T retry(String action, RunnableFunctionThrowsException<T> f, RetryPolicy policy)
+      throws CantRetryException {
+    RetryException e = null;
+    while (policy.attempt()) {
+      try {
+        return f.run();
+      } catch (RetryException ioe) {
+        e = ioe;
+        LOG.warn("Failed to {} (attempt {}): {}", action, policy.getAttemptCount(), e.toString());
+      }
+    }
+    if (e != null) {
+      throw new CantRetryException(e.getCause());
+    } else {
+      throw new CantRetryException();
+    }
+  }
+
+  /**
    * Gives a ClientRetry based on the given parameters.
    *
    * @param maxRetryDuration the maximum total duration to retry for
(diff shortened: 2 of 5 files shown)


yuzhu:
@gpang this PR reconnects the client everytime we get it from the cached object. Otherwise the client might be disconnected and not able to serve any requests..

yuzhu:

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -265,4 +263,19 @@ HiveMetaStoreClient getHive() throws IOException {
       Thread.currentThread().setContextClassLoader(currentClassLoader);
     }
   }
+
+  HiveMetaStoreClient getHive() throws IOException {
+    if (mHive != null) {
+      try {
+        mHive.getAllDatabases(); // test the connection
```
gpang: Is there a better way to detect if it is connected? Doing an extra RPC sounds inefficient.

An alternative can be to just do the desired operations with the existing cached client, and if it fails, try again with a re-connected client.

yuzhu: Unfortunately, `isConnected` is a private variable in the client. 
I thought about doing the second approach, but there is no way to encapsulate it in the `getHive` method though? That means for each call site of `getHive` we need to do this exception handling?

gpang: What was the scenario where this would happen? (I had never seen a client get disconnected)

I don't think calling an extra RPC just to check the connection state is what we want to do. Maybe there is a simple retry utility we can write?

yuzhu:
@gpang PTAL, implemented it as we discussed

yuzhu:
@gpang PTAL

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveMetaStoreRetryClient.java
```diff
@@ -0,0 +1,137 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A wrapper around HMS client that does caching and reconnection.
+ */
+public class HiveMetaStoreRetryClient implements MetaStoreClient {
+  private HiveMetaStoreClient mHive = null;
+  private String mConnectionUri;
```
gpang: can these 2 be `final`?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveMetaStoreRetryClient.java
```diff
@@ -0,0 +1,137 @@
...
+   * Constructor for the HiveMetaStoreRetryClient.
+   * @param connectionUri connection uri to the metastore
+   * @param hiveDbName the db name to access
+   */
+  public HiveMetaStoreRetryClient(String connectionUri, String hiveDbName) {
+    mConnectionUri = connectionUri;
+    mHiveDbName = hiveDbName;
+  }
+
+  private HiveMetaStoreClient getHive() throws IOException {
```
gpang: I think we are going to have to make this thread safe, by using a lock to protect changes to the cached client. Probably a RW lock, read to get the client, write to update it.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveMetaStoreRetryClient.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+
+  @Override
+  public List<String> getAllTables(String dbname) throws MetaException, IOException {
+    try {
+      return getHive().getAllTables(dbname);
+    } catch (TException e) {
+      mHive = newHiveClient();
+      return getHive().getAllTables(dbname);
```
gpang: instead of just calling it twice, can we have a retry loop?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveMetaStoreRetryClient.java
```diff
@@ -0,0 +1,137 @@
...
+      Thread.currentThread().setContextClassLoader(currentClassLoader);
+    }
+  }
+
+  @Override
+  public List<String> getAllTables(String dbname) throws MetaException, IOException {
+    try {
+      return getHive().getAllTables(dbname);
+    } catch (TException e) {
+      mHive = newHiveClient();
```
gpang: Instead of calling new directly, can we just set it to `null`, so that the next `getHive` call will just refresh it? We will need to do this in a thread-safe, concurrent way.

yuzhu:
@gpang  PTAL, thanks

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4527,6 +4534,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.table.transform.manager.job.monitor.interval";
     public static final String TABLE_TRANSFORM_MANAGER_JOB_HISTORY_RETENTION_TIME =
         "alluxio.table.transform.manager.job.history.retention.time";
+    public static final String TABLE_METASTORE_RETRY_TIMEOUT =
```
gpang: this is specific to the hive udb, so I think it should be part of the udb configuration.

core/common/src/main/java/alluxio/retry/RetryUtils.java
```diff
@@ -109,5 +137,61 @@ public static RetryPolicy defaultActiveSyncClientRetry(long activeUfsPollTimeout
     void run() throws IOException;
   }
 
+  /**
+   * Interface for methods which return a result and may throw IOException.
```
gpang: is this the correct comment for the exception part?

core/common/src/main/java/alluxio/retry/RetryUtils.java
```diff
@@ -109,5 +137,61 @@ public static RetryPolicy defaultActiveSyncClientRetry(long activeUfsPollTimeout
...
+     * Runs the runnable.
+     * @return T return an object of type T
+     */
+    T run() throws RetryException, CantRetryException;
+  }
+
+  /**
+   * Exceptions that can be retried.
+   */
+  public static class RetryException extends IOException {
```
gpang: can we call this `RetryableException`, and make it its own class (in this directory)?

core/common/src/main/java/alluxio/retry/RetryUtils.java
```diff
@@ -109,5 +137,61 @@ public static RetryPolicy defaultActiveSyncClientRetry(long activeUfsPollTimeout
...
+     */
+    public RetryException(String msg, Throwable cause) {
+      super(msg, cause);
+    }
+  }
+
+  /**
+   * Exceptions that can not be retried.
+   */
+  public static class CantRetryException extends Exception {
```
gpang: can we call this `NonRetryableException`, and make it its own class (in this directory)?

core/common/src/main/java/alluxio/retry/RetryUtils.java
```diff
@@ -46,6 +46,34 @@ public static void retry(String action, RunnableThrowsIOException f, RetryPolicy
...
+   * Retries the given method until it doesn't throw an IO exception or the retry policy expires. If
+   * the retry policy expires, the last exception generated will be rethrown.
+   *
+   * @param <T> type of the return value
+   * @param action a description of the action that fits the phrase "Failed to ${action}"
+   * @param f the function to retry
+   * @param policy the retry policy to use
+   * @return T return a value
+   */
+  public static <T> T retryFunction(String action, RunnableFunctionThrowsException<T> f,
```
gpang: Can we not just overload the `retry` name?

Also, can we write a unittest for this method?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -61,15 +59,16 @@
   /** the name of the hive db. */
   private final String mHiveDbName;
 
-  /** The cached hive client. This should not be used directly, but via {@link #getHive()}. */
-  private HiveMetaStoreClient mHive = null;
+  /** The cached hive client wrapper that tries to reconnect if it fails. */
+  private MetaStoreClient mHive = null;
```
gpang: Can this now be `final`?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -112,7 +111,7 @@ public String getName() {
   public List<String> getTableNames() throws IOException {
     try {
       return getHive().getAllTables(mHiveDbName);
-    } catch (MetaException e) {
+    } catch (TException e) {
```
gpang: Why is this different now?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveMetaStoreRetryClient.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * A wrapper around HMS client that does caching and reconnection.
+ */
+public class HiveMetaStoreRetryClient implements MetaStoreClient {
+  private HiveMetaStoreClient mHive = null;
+  private final String mConnectionUri;
+  private final String mHiveDbName;
+  private final ReadWriteLock mLock;
```
gpang: Please add a comment on what this is protecting and how it is used.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveMetaStoreRetryClient.java
```diff
@@ -0,0 +1,183 @@
...
+    mConnectionUri = connectionUri;
+    mHiveDbName = hiveDbName;
+    mLock = new ReentrantReadWriteLock();
+    mPolicy = RetryUtils.defaultClientRetry(
+        ServerConfiguration.getDuration(PropertyKey.TABLE_METASTORE_RETRY_TIMEOUT),
+        Duration.ofMillis(100), Duration.ofSeconds(5));
+  }
+
+  private HiveMetaStoreClient getHive() throws IOException {
+    if (mHive != null) {
```
gpang: These usages of `mHive` aren't protected by the lock?

gpang:
@yuzhu I just noticed this, but can we use `RetryingMetaStoreClient` from hive?

yuzhu:
opened #10577 instead to use Hive's retrying client

PR #10540, opened 2019-12-02
Refactoring SerializingExecutor related code in AbstractReadHandler class and AbstractWriteHandler class

witgo:


ZacBlanco:
@witgo thanks for the fix! Could you provide us with some more information about the refactor and why it's necessary?

witgo:
@ZacBlanco  In the previous version of the code, the mDataReaderExecutor#submit does not guarantee the Runnable#execution are executed in order. It may cause memory leaks.

ZacBlanco:
@witgo is this ready for review now?

witgo:
@ZacBlanco  yes.

ZacBlanco:

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
```diff
@@ -145,14 +148,56 @@ public void send(ReqT request) throws IOException {
...
     }
     try {
-      Object response = mResponses.poll(timeoutMs, TimeUnit.MILLISECONDS);
+      Object response;
+      if (request == null) {
+        response = mResponses.poll(timeoutMs, TimeUnit.MILLISECONDS);
+      } else {
+        Clock clock = Clock.systemUTC();
+        Duration maxWaitTime = Duration.ofSeconds(3);
+        Duration nextWaitTime = Duration.ofMillis(32);
```
ZacBlanco, line 178: Do you think the `maxWaitTime` or `nextWaitTime` should be configurable?

Also, have you considered using something from the `RetryUtils` class? It should make the code easier to read while keeping similar behavior.

ZacBlanco: Also, the default value passed for the `timeoutMs` is equal to the value `alluxio.user.streaming.data.timeout` which defaults to 30s. Since it is possible to configure the value higher or lower, I think rather than make it user-configurable I would rather just dynamically adjust the `maxWaitTime` based on the `timeoutMs` parameter.

What do you think?

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
```diff
@@ -145,14 +148,56 @@ public void send(ReqT request) throws IOException {
    * @throws IOException if any error occurs
    */
   public ResT receive(long timeoutMs) throws IOException {
+    return retryReceive(timeoutMs, null);
```
ZacBlanco, line 151: is there any point in having the `receive` method if it directly calls `retryReceive`? Or rather, could the implementation of `retryReceive` simply be `receive`?

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataMessageBlockingStream.java
```diff
@@ -58,10 +58,15 @@ public GrpcDataMessageBlockingStream(Function<StreamObserver<ResT>, StreamObserv
 
   @Override
   public ResT receive(long timeoutMs) throws IOException {
+    return retryReceive(timeoutMs, null);
+  }
+
+  @Override
+  public ResT retryReceive(long timeoutMs, ReqT request) throws IOException {
```
ZacBlanco, line 65: similar comment about whether a separate `retryReceive` is necessary if `receive` directly calls it.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataMessageBlockingStream.java
```diff
@@ -79,9 +84,25 @@ public ResT receive(long timeoutMs) throws IOException {
...
+   * or throw an exception if times out. Caller of this method must release the buffer after reading
+   * the data.
+   *
+   * @param timeoutMs maximum time to wait before giving up and throwing
+   *                  a {@link DeadlineExceededException}
+   * @param request   the request
+   * @return the response message with data buffer, or null if the inbound stream is completed
+   * @throws IOException if any error occurs
+   */
+  public DataMessage<ResT, DataBuffer> retryReceiveDataMessage(long timeoutMs, ReqT request)
```
ZacBlanco, line 101: similar comment about whether a separate `retryReceive` is necessary if `receive` directly calls it.

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
```diff
@@ -81,8 +81,8 @@ public static InstancedConfiguration defaults() {
     conf.put(PropertyKey.MASTER_METASTORE_DIR, PathUtils.concatPath(workDirectory, "metastore"));
 
     conf.put(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT, "1KB");
-    conf.put(PropertyKey.USER_BLOCK_REMOTE_READ_BUFFER_SIZE_BYTES, "64");
-    conf.put(PropertyKey.USER_STREAMING_READER_CHUNK_SIZE_BYTES, "64");
+    conf.put(PropertyKey.USER_BLOCK_REMOTE_READ_BUFFER_SIZE_BYTES, "64KB");
+    conf.put(PropertyKey.USER_STREAMING_READER_CHUNK_SIZE_BYTES, "64KB");
```
ZacBlanco, line 85: Any reason for changing these two?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -128,11 +122,8 @@ public void onNext(alluxio.grpc.ReadRequest request) {
       validateReadRequest(request);
       mContext.setPosToQueue(mContext.getRequest().getStart());
       mContext.setPosReceived(mContext.getRequest().getStart());
-      mDataReaderExecutor.submit(createDataReader(mContext, mResponseObserver));
+      mDataReaderExecutor.execute(createDataReader(mContext, mResponseObserver));
```
ZacBlanco, line 125: Is there an advantage to using `execute` vs `submit`?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
```diff
@@ -109,7 +107,7 @@ protected DataBuffer getDataBuffer(BlockReadRequestContext context, long offset,
       openBlock(context);
       BlockReader blockReader = context.getBlockReader();
       Preconditions.checkState(blockReader != null);
-      ByteBuf buf = PooledByteBufAllocator.DEFAULT.buffer(len, len);
+      ByteBuf buf = BufferAllocator.DEFAULT.buffer(len);
```
ZacBlanco, line 110: What are the advantages of using your `BufferAllocator` over the `PooledByteBufAllocator` from netty?

core/server/worker/src/main/java/alluxio/worker/grpc/BufferAllocator.java
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+      mAllocationLimit = maxCapacity;
+    }
+    mByteBufAllocator = PooledByteBufAllocator.DEFAULT;
+  }
+
+  /**
+   * Allocate a {@link ByteBuf} with the given capacity size.
+   * If it is a direct or heap buffer depends on the actual
+   * implementation.
```
ZacBlanco, line 58: Nit: can you re-format the javadoc for the 100 character line length limit?

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
```diff
@@ -45,16 +45,9 @@
   public static final ExecutorService BLOCK_READER_EXECUTOR =
       new ImpersonateThreadPoolExecutor(new ThreadPoolExecutor(THREADS_MIN,
           ServerConfiguration.getInt(PropertyKey.WORKER_NETWORK_BLOCK_READER_THREADS_MAX),
-          THREAD_STOP_MS, TimeUnit.MILLISECONDS, new SynchronousQueue<>(),
+          THREAD_STOP_MS, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>(),
```
ZacBlanco, line 48: Using a linked blocking queue will allow the worker to possibly become overloaded and requests could queue up for a while since the queue length is infinite.

Personally, I think it is better behavior from the user-experience side to get an immediate error back when the worker is overloaded with requests, rather than having to wait for a timeout.

That being said, I think it's ok to queue up requests, but I think there should be a bound on the queue where we start rejecting readers after a certain point in time. What are your thoughts?

core/server/worker/src/main/java/alluxio/worker/grpc/MutableWrappedByteBuf.java
```diff
@@ -0,0 +1,473 @@
...
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.ScatteringByteChannel;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * This is basically a complete copy of netty's DuplicatedByteBuf.
+ * We copy because we want to override
+ * some behaviors and make buffer mutable.
+ */
```
ZacBlanco, line 34: Nit: can you re-format the javadoc for the 100 character line-length limit?

core/server/worker/src/main/java/alluxio/worker/grpc/MutableWrappedByteBuf.java
```diff
@@ -0,0 +1,473 @@
...
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.ScatteringByteChannel;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * This is basically a complete copy of netty's DuplicatedByteBuf.
```
ZacBlanco, line 31: Can you use an `{@link <>}` to the netty class, or give the fully qualified class name?

core/server/worker/src/main/java/alluxio/worker/grpc/MutableWrappedByteBuf.java
```diff
@@ -0,0 +1,473 @@
...
+ * We copy because we want to override
+ * some behaviors and make buffer mutable.
```
ZacBlanco, line 33: Can you link the methods which we override the behavior for?

And can you also either in the class javadoc or method javadoc explain the difference in behavior?

ggezer:
@witgo Thank you for this fix. We'll prioritize taking a look at it. Meanwhile, I'd like to understand if we should expect any perf impact by this change?

ZacBlanco:
I also spent some time testing this patch. I found that under low load from a single worker that things worked appropriately. However, once I put more requests onto a single worker than it had configured threads it looks like the worker froze up and stopped accepting requests entirely, even after all of the initial requests were cancelled.

I think this PR is promising but we need to properly handle the case when the # of requests becomes too large to handle.

witgo:
I'm sorry, I've been busy with other things lately, and I haven't had time to get into this feature yet.
@ZacBlanco  Can you share the log with me？

witgo:
@ZacBlanco 
I did not reproduce the problem. Can you post the `jstack` output?

PR #10546, opened 2019-12-02
Improve error handling on creating Processes

apc999:
It turns out that our process creation can be quite heavy and we may see `java.lang.ExceptionInInitializerError` on constructors. Adding error handling to match how we are doing with alluxio master and alluxio job master on other processes. 

Prior to this change, the reasons of a failed process (e.g., a worker) may not be in `worker.log` but in `worker.out` instead, as the `ExceptionInInitializerError` is treated as a JVM runtime exception not logged. After this patch, users can tell the errors easier by looking at `{worker,proxy,job_worker}.log` similar as how they can reason at `{master,job_master}.log`


calvinjia:

apc999:

PR #10555, opened 2019-12-03
Include integration test results in code coverage

gpang:
Jacoco generates coverage data in .exec files, which are then used for the jacoco-report goal. However, integration tests are in a separate module from the source code they are testing. Therefore, all the coverage data is not included when reporting on the source code.

Therefore, to capture all the test coverage, the reporting has to be done in 2 steps:
- generate all coverage data (into the same data file)
- report the coverage data using the coverage data which includes all the tests

This is done with 2 separate profiles for the different steps:
- `jacoco` to generate the coverage data
- `jacocoReport` to generate the report using the coverage data

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,6 +146,7 @@
     <netty.version>4.1.30.Final</netty.version>
     <hadoop.version>2.7.3</hadoop.version>
     <hadoop-openstack.version>2.6.0</hadoop-openstack.version>
+    <jacoco.version>0.8.5</jacoco.version>
     <java.version>1.8</java.version>
     <jersey.version>2.22</jersey.version>
     <jetty.version>9.2.16.v20160414</jetty.version>
@@ -927,7 +928,8 @@
             <runOrder>alphabetical</runOrder>
             <systemPropertyVariables>
               <alluxio.test.mode>true</alluxio.test.mode>
-              <jacoco-agent.destfile>${project.basedir}/target/jacoco.exec</jacoco-agent.destfile>
+              <jacoco-agent.append>true</jacoco-agent.append>
+              <jacoco-agent.destfile>${build.path}/../target/jacoco-combined.exec</jacoco-agent.destfile>
             </systemPropertyVariables>
             <useSystemClassLoader>${surefire.useSystemClassLoader}</useSystemClassLoader>
           </configuration>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -927,7 +928,8 @@
             <runOrder>alphabetical</runOrder>
             <systemPropertyVariables>
               <alluxio.test.mode>true</alluxio.test.mode>
-              <jacoco-agent.destfile>${project.basedir}/target/jacoco.exec</jacoco-agent.destfile>
+              <jacoco-agent.append>true</jacoco-agent.append>
+              <jacoco-agent.destfile>${build.path}/../target/jacoco-combined.exec</jacoco-agent.destfile>
```
ZacBlanco: what does `${build.path}` normally resolve to? and how is it different from `${project.basedir}`?

tests/pom.xml
```diff
@@ -320,6 +320,40 @@
...
+
+    <profile>
+      <id>jacoco</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <version>${jacoco.version}</version>
+            <!-- integration tests shouldn't use PowerMock, so recommended prepare-agent goal can be used, instead of offline-instrumentation -->
```
ZacBlanco: Have you verified that none of them currently use PowerMock?

pom.xml
```diff
@@ -1333,27 +1335,60 @@
...
+                  in the project.
+               -->
+              <append>true</append>
+              <destFile>${build.path}/../target/jacoco-combined.exec</destFile>
+            </configuration>
             <executions>
               <execution>
                 <!-- offline instrumentation is required, for jacoco to work with powermock tests -->
-                <id>jacoco-instrument-classes</id>
+                <id>jacoco-unit-instrument-classes</id>
```
ZacBlanco: I couldn't find this id in the jacoco documentation. Do you have a pointer to it?

tests/pom.xml
```diff
@@ -320,6 +320,40 @@
...
+            <executions>
+              <execution>
+                <id>jacoco-unit-instrument-classes</id>
+                <configuration>
+                  <skip>true</skip>
+                </configuration>
+              </execution>
+              <execution>
+                <id>jacoco-unit-restore-instrumented-classes</id>
```
ZacBlanco: Should these this be `jacoco-unit-restore-instrumented-classes` or `jacoco-restore-instrumented-classes`?

pom.xml
gpang: Every sub-module/pom file already defines its own `build.path`, to point to a folder in the project root directory. I am reusing that to get to a different folder (`target`) in the project root directory. `project.basedir` is relative to the specific sub-module, so each sub-module ends up writing its own file, instead of appending to the shared one.

pom.xml
gpang: the `id` is something we can make up. It is the name of the execution.

The `goal` is the name of the specific goal of each plugin that is already predefined.

tests/pom.xml
gpang: Yes, there are no PowerMock usage in integration tests. There are usages of whitebox however, but I don't think those cause problems with the code coverage.

tests/pom.xml
gpang: the `id` is something we can make up.

pom.xml
ZacBlanco: I see. I hope that there is a way to verify that the `build.path` is correct for each module. I'm wondering if any are incorrect by chance. If they are it could impact the coverage report

gpang: I think if the `build.path` is incorrect, the actual compilation might fail, since that path was already being used. However, even after I ran it, I only found 1 instance of jacoco-combined.exec file, so in the current state, the `build.path` settings are all correct.

gpang:

PR #10574, opened 2019-12-04
Increase code coverage for backup delegation

ggezer:


ggezer:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -204,9 +204,9 @@ public CatchupFuture catchup(long sequence) {
       Preconditions.checkState(!mResumeInProgress, "Resume in progress");
       Preconditions.checkState(mCatchupThread == null || !mCatchupThread.isAlive(),
           "Catch-up task in progress.");
-      Preconditions.checkState(sequence >= 0, String.format("Invalid sequence: %d", sequence));
-      Preconditions.checkState(mLastAppliedSequence <= sequence, String.format(
-          "Can't catchup to past. Current: %d, Requested: %d", mLastAppliedSequence, sequence));
+      Preconditions.checkState(sequence >= 0, "Invalid sequence: %d", sequence);
```
gpang: Somewhere, can we add that it is invalid because it is less than 0? There are many ways for something to be invalid.

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
```diff
@@ -283,12 +283,13 @@ protected void runCatchup() {
             try {
               mSuspendBuffer.wait();
             } catch (InterruptedException e) {
+              Thread.currentThread().interrupt();
```
gpang: This was a bug? Was this discovered by one of the new tests?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -185,9 +185,9 @@ public void snapshot(SnapshotWriter writer) {
     try (SnapshotWriterStream sws = new SnapshotWriterStream(writer)) {
       writer.writeLong(snapshotId);
       JournalUtils.writeToCheckpoint(sws, getStateMachines());
-    } catch (Throwable t) {
-      ProcessUtils.fatalError(LOG, t, "Failed to snapshot");
-      throw new RuntimeException(t);
+    } catch (Exception e) {
+      ProcessUtils.fatalError(LOG, e, "Failed to snapshot");
```
gpang: can we include `snapshotId` in the message?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -208,8 +208,8 @@ public void install(SnapshotReader snapshotReader) {
     try (InputStream srs = new SnapshotReaderStream(snapshotReader)) {
       snapshotId = snapshotReader.readLong();
       JournalUtils.restoreFromCheckpoint(new CheckpointInputStream(srs), getStateMachines());
-    } catch (Throwable t) {
-      JournalUtils.handleJournalReplayFailure(LOG, t,
+    } catch (Exception e) {
+      JournalUtils.handleJournalReplayFailure(LOG, e,
           "Failed to install snapshot");
```
gpang: Can we include `snapshotId` in the message?

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
```diff
@@ -85,6 +85,12 @@ public void suspendCatchupResume() throws Exception {
 
     // Suspend follower journal system.
     mFollowerJournalSystem.suspend();
+    try {
+      mFollowerJournalSystem.suspend();
+    } catch (Exception e) {
+      // Expected to fail when suspending a suspended journal.
```
gpang: If this is expected to fail, then we should add an `assert.Fail()` after the `suspend` command.

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
```diff
@@ -186,9 +186,9 @@ protected AlluxioURI takeBackup(BackupPRequest request, AtomicLong entryCounter)
       } catch (IOException e) {
         try {
           ufs.deleteExistingFile(backupFilePath);
```
gpang: What is the difference between `backupFilePath` and `backupUri`?

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
```diff
@@ -186,9 +186,9 @@ protected AlluxioURI takeBackup(BackupPRequest request, AtomicLong entryCounter)
...
-        } catch (Throwable t) {
-          LOG.error("Failed to clean up failed backup at {}", backupFilePath, t);
-          e.addSuppressed(t);
+        } catch (Exception e2) {
+          LOG.error("Failed to clean up failed backup at {}", backupFilePath, e2);
+          e.addSuppressed(e2);
         }
         throw e;
```
gpang: Can we wrap this exception with more info, like backup path, and `entryCounter`?

core/server/common/src/main/java/alluxio/master/journal/raft/BufferedJournalApplier.java
ggezer: This is to silence sonarqube actually as it complained.
There is not much interrupting the thread that is going to terminate as soon as quitting this loop.

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
ggezer: `backupFilePath` is the given path+ generated file name.
`backupUri` is actual location that backup will be written based on other options. (i.e --local)

ggezer:

PR #10588, opened 2019-12-06
Change default worker heartbeat back to 1s

LuQQiu:
After the changes in https://github.com/Alluxio/alluxio/pull/10485 to make metrics processing async, worker heartheat has not related to the latency of metrics. 
Change default worker heartbeat back to 1s can improve RPC performance.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1526,7 +1526,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey MASTER_WORKER_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_WORKER_HEARTBEAT_INTERVAL)
           .setDefaultValue("1sec")
-          .setDescription("The interval between Alluxio master and worker heartbeats.")
+          .setDescription("The interval between Alluxio master heartbeats to detect lost workers "
+              + "and files based on updates from Alluxio workers.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
@@ -2111,7 +2112,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       new Builder(Name.WORKER_BLOCK_HEARTBEAT_INTERVAL_MS)
           .setAlias("alluxio.worker.block.heartbeat.interval.ms")
           .setDefaultValue("1sec")
-          .setDescription("The interval between block workers' heartbeats.")
+          .setDescription("The interval between block workers' heartbeats to update "
+              + "block status, storage health and other workers' information to Alluxio Master.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1525,6 +1525,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL =
       new Builder(Name.MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL)
+          .setAlias("alluxio.master.worker.heartbeat.interval")
           .setDefaultValue("10sec")
           .setDescription("The interval between Alluxio master detections to find lost workers "
               + "and files based on updates from Alluxio workers.")
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1523,10 +1523,10 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_WORKER_HEARTBEAT_INTERVAL =
-      new Builder(Name.MASTER_WORKER_HEARTBEAT_INTERVAL)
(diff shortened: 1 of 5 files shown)


LuQQiu:
Failed tests: 
alluxio.job.master.JobMasterIntegrationTest.throttleJobWorkerTasks(alluxio.job.master.JobMasterIntegrationTest)
  Run 1: JobMasterIntegrationTest.throttleJobWorkerTasks:203
  Run 2: JobMasterIntegrationTest.throttleJobWorkerTasks:203

LuQQiu:
jenkins, test this please

LuQQiu:
@gpang PTAL,thanks!

gpang:
@LuQQiu what was the difference between these 2 parameters? It is confusing.

LuQQiu:
master worker heartbeat is used for detecting lost workers and lost files.
worker block heartbeat is used for worker block sync, worekr pin list sync, worker storage health.
I also feel confusing for having two heartbeats which description don't help much to distinguish them

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1525,7 +1525,7 @@ public String toString() {
           .build();
   public static final PropertyKey MASTER_WORKER_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_WORKER_HEARTBEAT_INTERVAL)
-          .setDefaultValue("10sec")
+          .setDefaultValue("1sec")
           .setDescription("The interval between Alluxio master and worker heartbeats.")
```
gpang: can we update this description to be more accurate? It is quite confusing to have both parameters to represent different concepts.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1525,7 +1525,7 @@ public String toString() {
           .build();
   public static final PropertyKey MASTER_WORKER_HEARTBEAT_INTERVAL =
       new Builder(Name.MASTER_WORKER_HEARTBEAT_INTERVAL)
-          .setDefaultValue("10sec")
+          .setDefaultValue("1sec")
```
gpang: @calvinjia This was change to 10s a while ago, what was the reason for that, and do we want to change it back to 1s with this PR?

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Actually the biggest difference between these two properties is `MASTER_WOREKR_HEARTBEAT_INTERVAL` is used by AlluxioMaster. `WORKER_BLOCK_HEARTBEAT_INTERVAL` is used by AlluxioWorker.
What they actually do is not that important
Will update the description

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: I think it was in an attempt to change the worker heartbeat, but it only affects the lost file detector. I don't think we need to run the lost file detector so frequently, perhaps we should rename this key and keep 10 seconds.

LuQQiu: Previously our worker heartbeat takes a long time because of the sync metrics processing, so we update the `WORKER_BLOCK_HEARTBEAT_INTERVAL` which controls the interval of workerS calling worker heartbeat to 10s.

To be efficient, `MASTER_WORKER_HEARTBEAT_INTERVAL` should be similar to the value of `WORKER_BLOCK_HEARTBEAT_INTERVAL`. For example, `MASTER_WOREKR_HEARTBEAT_INTERVAL` controls the interval of finding lost workers and lost files which are updated by workerHeartbeat (controlled by `WORKER_BLOCK_HEARTBEAT_INTERVAL`)

LuQQiu: @calvinjia It also affects the lost worker detector, this part may be more critical

calvinjia: Oh right, though our worker timeout is 5 minutes so I think plus up to 10 seconds of additional delay is fine. Renaming it is mainly to avoid the confusion that it is the same as the actual heartbeat interval coming from the worker.

LuQQiu:
@gpang @calvinjia  I updated the description of the two interval properties and renamed the master worker heartbeat to be more concrete. PTAL, thanks!

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -64,7 +64,7 @@ private static String replacedSince(String version, String newProperty) {
       put("alluxio.master.connection.timeout", removedSince(V2_0_0));
       put("alluxio.master.file.async.persist.handler", removedSince(V2_0_0));
       put("alluxio.master.heartbeat.interval",
-          replacedSince(V2_0_0, PropertyKey.MASTER_WORKER_HEARTBEAT_INTERVAL.getName()));
+          replacedSince(V2_0_0, PropertyKey.MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL.getName()));
```
gpang: Yay, it looks like we already deprecated it!

LuQQiu: We change the name from `alluxio.master.heartbeat.interval` to `alluxio.master.worker.heartbeat.interval`.
Oh... may be I need to add the `alluxio.master.worker.heartbeat.interval` as the alias of this key or add it to the removedKeys

LuQQiu: Looks like the RemovedKey is updated in batch in release process. I will not update it for now.

LuQQiu:
@gpang I set the previous property key name `alluxio.master.worker.heartbeat.interval` as the alias name, PTAL

LuQQiu:
This PR should be cherry-picked to 2.1.1
for 1.8.3, will have a separate PR and leave the MASTER_WORKER_HEARTBEAT_INTERVAL NAME unchanged.

LuQQiu:

LuQQiu:

PR #10589, opened 2019-12-06
Optimize DistributedLoadCommand

bradyoo:
Premise: Each job has an overhead of at least 1~2 seconds because of it takes at least JobWorker heartbeats (once for the work to go from JobMaster to JobWorker and once more for the JobWorker work completion to be reported backed to JobMaster) so it is important that when there are a lot of jobs to be done, that many jobs get submitted at the same time.

What was done:
- Remove creating threads per job (only 1 thread)
- Increase number of default outgoing jobs from 10 -> 1000 (also configurable)
- Client round robins all of the outgoing active jobs, handles retries, and when they succeed, remove from the list to be filled with new jobs.

Test Environment: 

- Local machine
- 12000 Files
- Approximately 450 bytes each
- Ran fs free followed by fs distributedLoad, checked that distributedLoad was actually loading files and timed the run

New Code: 59 seconds
Old Code: 43 minutes 33 seconds

This is about a 45x speedup.

Memory Usage: Both the old version and new version memory leaked (the new version memory leaked much faster because it executes much faster) but the addition of https://github.com/Alluxio/alluxio/pull/10606 fixed the issue.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
@@ -12,16 +12,23 @@
 package alluxio.cli.fs.command;
 
 import alluxio.AlluxioURI;
+import alluxio.ClientContext;
 import alluxio.annotation.PublicApi;
 import alluxio.cli.CommandUtils;
 import alluxio.cli.fs.FileSystemShellUtils;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.URIStatus;
-import alluxio.client.job.JobGrpcClientUtils;
+import alluxio.client.job.JobMasterClient;
 import alluxio.exception.AlluxioException;
 import alluxio.exception.status.InvalidArgumentException;
+import alluxio.job.JobConfig;
 import alluxio.job.plan.load.LoadConfig;
 
+import alluxio.job.wire.JobInfo;
+import alluxio.retry.CountingRetry;
+import alluxio.retry.RetryPolicy;
+import alluxio.worker.job.JobMasterClientContext;
+import com.google.common.collect.Lists;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
@@ -30,16 +37,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CompletionService;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorCompletionService;
-import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 
 import javax.annotation.concurrent.ThreadSafe;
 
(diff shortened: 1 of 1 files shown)


bradyoo:
@chcn: This is probably testable already as is. What this is currently lacking is tests and better error handling.

bradyoo:
@chcn: I will but preferably not for the testing stage.

bradyoo:
Ready for review.

bradyoo:
@calvinjia: Want to look at this?

ggezer:
This change renders https://github.com/Alluxio/alluxio/issues/10608 relevant.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -61,20 +60,88 @@
...
+      }
+      return false;
+    }
+
+    /**
+     * Returns the status of the job attempt.
+     * @return True if finished successfully or cancelled, False if FAILED and should be retried,
+     *              null if the status should be checked again later
+     */
+    private Boolean check() {
```
calvinjia: Instead of returning a boolean should we just use the state, I'm not sure we are getting that much simplification?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -61,20 +60,88 @@
...
+    private final RetryPolicy mRetryPolicy;
+    private final JobMasterClient mClient;
+
+    private Long mJobId;
+
+    private JobAttempt(JobConfig jobConfig, RetryPolicy retryPolicy, ClientContext clientContext) {
+      mJobConfig = jobConfig;
+      mRetryPolicy = retryPolicy;
+      mClient = JobMasterClient.Factory.create(
+          JobMasterClientContext.newBuilder(clientContext).build());
```
calvinjia: +1 to pooling the clients, especially since these are mostly sharable (all going to job master), we could just fetch a resource from a pool each time we need to run an RPC.

bradyoo: pooling is very annoying here because the clientContext for each one of these is different. It's also no different from what it was before. (Sicne JobGrpcClientUtils... will create a new client every time)

bradyoo:
addressed feedback. Not doing pooling.

bradyoo:
Note that this is to be cherry-picked into 2.1.1

bradyoo:
alluxio-bot, test this please

bradyoo:
@calvinjia

calvinjia:

calvinjia:

bradyoo:
Urgh.. I'll do the manual cherry-pick in that case.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -130,29 +173,45 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
+          case RUNNING:
+            continue;
+          case CANCELED:
+          case COMPLETED:
+            removed = true;
+            jobAttempt.close();
+            iterator.remove();
+            continue;
+          case FAILED:
+            jobAttempt.run();
```
chcn: @bradyoo should we check whether the return value of `run` here? Otherwise, if the job keeps failing, the job cannot be removed from the queue, even when `run` returns `false`?

bradyoo: You are right. I should fix this.

apc999:
@bradyoo any reason to change the option name? we typically will not change the name unless it is confusing. Also we need to update the CLI description string.

bradyoo:
@apc999: The option name had to change because any previously used values for the option is woefully inappropriate for the new option. (Off by at least 2 orders of magnitude)

PR #10593, opened 2019-12-09
Prevent resource leaks in the filesystem.

witgo:


yuzhu:
@chcn PTAL

apc999:
@witgo can you provide more details of your PR in the PR description? e.g., when and how the leak happens in your case as well as how the fix helps. In this way, it will help us estimate the priority and speed up the review. thanks

PR #10596, opened 2019-12-09
Add collectInfo command

jiacheliu3:
Add Alluxio CollectInfo command set to Alluxio command line. This command will collect information on every host in the Alluxio cluster, tarball the information into a local tarball file, then gather all the tarballs back to where the command is issued.

This command is designed to help troubleshooting and collect information from Alluxio clusters. The collected information include logs, config files and it also executes a set of commands on each of the hosts and include outputs in the tarball.

**Added commands:**
This will run one or all checks on the local machine and write the output into files in `targetDir`. At the end of execution a tarball will be produced containing all the information collected, put in the `targetDir` as well.
`bin/alluxio collectInfo --local [all/collectConfig/collectEnv/collectLog/collectMetrics/collectAlluxioInfo] targetDir`

This will run the corresponding `collectInfo --local` command on all hosts in the Alluxio cluster defined in `conf/masters` and `conf/workers` via SSH. After that the tarballs generated by each individual `infoBundle` call will be SCPed into a local temp directory, and then put into one final tarball in the `targetDir`.
`bin/alluxio collectInfo [all/collectConfig/collectEnv/collectLog/collectMetrics/collectAlluxioInfo] targetDir`

**Phase 2:**
This PR is the Phase 1 of CollectInfo feature. This PR provides basic implementations and defines the structure.

Some features are not implemented in this PR and scheduled to a separate Phase 2 PR. These features include the following:
1. Skipping a command when pre-existing work are found. This is because if the cluster is large, running infoBundle can take a long time and fail in the middle. 
1. Add a `-f` option to force overwrite existing work.
1. Add a `-c` or `--components` option to copy only certain logs.
1. Add a `-h` or `--hosts` option to only invoke `collectInfo` on certain hosts.
1. Add security check to avoid copying `conf/` files that contain credentials.


shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+
+  @Override
+  protected Map<String, Command> loadCommands() {
+    // Give each command the configuration
+    Map<String, Command> commands = CommandUtils.loadCommands(
+            InfoCollector.class.getPackage().getName(),
+            new Class[] {FileSystemContext.class},
+            new Object[] {FileSystemContext.create(mConfiguration)});
+    System.out.println(String.format("Loaded commands %s", commands));
+    return commands;
```
bf8086: Is this used anywhere?

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+          System.out.println(String.format("Command failed on host %s", host));
+          System.out.println(c.getFormattedOutput());
+        } else {
+          System.out.println(String.format("Success on host %s", host));
+        }
+      }
+    }
+
+    // Collect all tarballs to local
+    File tempDir = AlluxioTestDirectory.createTemporaryDirectory(TEMP_DIR_NAME);
```
bf8086: It is a little strange to create a temporary directory named with `alluxio-tests`. You can probably use guava's ` Files.createTempDir()` instead.

shell/src/main/java/alluxio/cli/bundler/command/AbstractInfoCollectorCommand.java
```diff
@@ -0,0 +1,95 @@
...
+   * Creates the directory if it does not exist.
+   *
+   * @param cl the parsed {@link CommandLine}
+   * @return the directory path
+   * */
+  public String getWorkingDirectory(CommandLine cl) {
+    String[] args = cl.getArgs();
+    String baseDirPath = args[0];
+    String workingDirPath =  Paths.get(baseDirPath, this.getCommandName()).toString();
+    System.out.println(String.format("Command %s works in %s", this.getCommandName(),
```
bf8086: Does user need to know this? Should we just log a message instead?

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
```diff
@@ -0,0 +1,225 @@
...
+     * Creates an instance of {@link AlluxioCommand}.
+     *
+     * @param name name of this Alluxio operation
+     * @param alluxioPath where Alluxio can be found
+     * @param cmd Alluxio cmd to run
+     * @param alternative an alternative option if the Alluxio operation fails
+     *                    for reasons like the cluster being currently down
+     * */
+    public AlluxioCommand(String name, String alluxioPath, String cmd, ShellCommand alternative) {
+      super((alluxioPath + " " + cmd).split(" "));
```
bf8086: (nit) use `new String[] {alluxioPath, cmd}`?

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
```diff
@@ -0,0 +1,225 @@
...
+              tryAgainRes = String.format("Alternative command %s succeeded: %s", alternativeCmd,
+                      tryAgain.getFormattedOutput());
+              LOG.info(tryAgainRes);
+            }
+          } catch (IOException e) {
+            tryAgainRes = String.format("Alternative command %s failed with exception: %s",
+                    alternativeCmd, e.getMessage());
+            if (LOG.isDebugEnabled()) {
+              e.printStackTrace();
+            }
```
bf8086: Refactor line 145-162 into a helper method and reuse it here?

shell/src/main/java/alluxio/cli/bundler/command/AbstractInfoCollectorCommand.java
```diff
@@ -0,0 +1,95 @@
...
+  protected FileSystemContext mFsContext;
+  protected String mWorkingDirPath;
+
+  /**
+   * Creates an instance of {@link AbstractInfoCollectorCommand}.
+   *
+   * @param fsContext {@link FileSystemContext} the context to run in
+   * */
+  public AbstractInfoCollectorCommand(@Nullable FileSystemContext fsContext) {
+    if (fsContext == null) {
```
bf8086: Can this ever be null?

shell/src/main/java/alluxio/cli/bundler/command/CollectEnvCommand.java
```diff
@@ -0,0 +1,189 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Command to run a set of shell commands to get system information.
+ * */
+public class CollectEnvCommand extends AbstractInfoCollectorCommand {
```
bf8086: This command share a lot of common logic with `CollectAlluxioInfoCommand`. Can we move the shared logic to a base class and just define different shell commands in each sub class?

shell/src/main/java/alluxio/cli/bundler/command/CollectJvmInfoCommand.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    return 0;
+  }
+
+  private Map<String, String> getJps() throws IOException {
+    Map<String, String> procs = new HashMap<>();
+
+    // Get Jps output
+    String[] jpsCommand = new String[]{"jps"};
```
bf8086: `jps` and `jstack` will only work if the Java process is started by the current user unless you are running command as super user. Is it possible to try running it using sudo first?

shell/src/main/java/alluxio/cli/bundler/command/CollectJvmInfoCommand.java
```diff
@@ -0,0 +1,153 @@
...
+
+  @Override
+  public String getUsage() {
+    return null;
+  }
+
+  @Override
+  public String getDescription() {
+    return null;
+  }
```
bf8086: Add usage and description?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -0,0 +1,75 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Command to collect Alluxio logs.
+ * */
+public class CollectLogCommand  extends AbstractInfoCollectorCommand {
```
bf8086: Can we reuse the common code from `CollectConfigCommand`?

shell/src/main/java/alluxio/cli/bundler/command/CollectMetricsCommand.java
```diff
@@ -0,0 +1,127 @@
...
+
+      // Generate URL from config properties
+      String masterAddr;
+      try {
+        masterAddr = mFsContext.getMasterAddress().getHostName();
+      } catch (UnavailableException e) {
+        String noMasterMsg = "No Alluxio master available. Skip metrics collection.";
+        LOG.warn(noMasterMsg);
+        outputBuffer.write(noMasterMsg);
+        continue;
```
bf8086: We should probably break instead of continue which will just retry and get the same error?

shell/src/main/java/alluxio/cli/bundler/InfoCollector.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Class for collecting various information about the host instance.
+ */
+public class InfoCollector extends AbstractShell {
+  protected static final String TARBALL_NAME = "collectAll.tar.gz";
```
bf8086: (nit)`alluxio-info.tar.gz`?

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+    String targetDir = argv[1];
+
+    List<String> allHosts = new ArrayList<>(shellAll.getHosts());
+    System.out.println(String.format("Init thread pool for %s hosts", allHosts.size()));
+    shellAll.mExecutor = Executors.newFixedThreadPool(allHosts.size());
+
+    // Invoke infoBundle on each host
+    List<CompletableFuture<CommandReturn>> futureList = new ArrayList<>();
+    for (String host : allHosts) {
+      System.out.println(String.format("Execute InfoCollector on host %s", host));
```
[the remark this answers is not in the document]
jiacheliu3: By design the intention is, the command `alluxio infoBundle` (now `alluxio collectInfo`) can be directly used as a valid command. So I think it makes sense to say something like `Invoke alluxio collectInfo on host blah`. What do you think?

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+   * @return a set of hostnames in the cluster
+   * */
+  public Set<String> getHosts() {
+    // File
+    String confDirPath = mConfiguration.get(PropertyKey.CONF_DIR);
+    System.out.println(String.format("Looking for masters and workers in %s", confDirPath));
+    Set<String> hosts = new HashSet<>();
+    hosts.addAll(CommandUtils.readNodeList(confDirPath, "masters"));
+    hosts.addAll(CommandUtils.readNodeList(confDirPath, "workers"));
+    System.out.println(String.format("Found %s hosts: %s", hosts.size(), hosts));
```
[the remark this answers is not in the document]
jiacheliu3: On second thought, the hostnames are included in printout of each invocation. So the hostname list is removed here.

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+      System.out.println("Host size mismatch!");
+    } else {
+      for (int i = 0; i < allHosts.size(); i++) {
+        CommandReturn c = results.get(i);
+        String host = allHosts.get(i);
+        if (c.getExitCode() != 0) {
+          System.out.println(String.format("Command failed on host %s", host));
+          System.out.println(c.getFormattedOutput());
+        } else {
+          System.out.println(String.format("Success on host %s", host));
```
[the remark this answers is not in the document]
jiacheliu3: Changed to `Successfully collected information on %d/%d hosts` to keep a succeeded count.

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+
+      CompletableFuture<CommandReturn> future = CompletableFuture.supplyAsync(() -> {
+        String workDir = conf.get(PropertyKey.WORK_DIR);
+        String alluxioBinPath = Paths.get(workDir, "bin/alluxio")
+                .toAbsolutePath().toString();
+        System.out.println(String.format("host: %s, alluxio path %s", host, alluxioBinPath));
+
+        try {
+          String[] infoBundleArgs =
+                  (String[]) ArrayUtils.addAll(new String[]{alluxioBinPath, "infoBundle"}, argv);
```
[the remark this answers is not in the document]
jiacheliu3: Yea the RHS returns `Object[]` so the cast is needed.

shell/src/main/java/alluxio/cli/bundler/InfoCollector.java
```diff
@@ -0,0 +1,171 @@
...
+    // Execute the Collectors one by one
+    // Reduce the RPC retry max duration to fall earlier for CLIs
+    conf.set(PropertyKey.USER_RPC_RETRY_MAX_DURATION, "5s", Source.DEFAULT);
+    InfoCollector shell = new InfoCollector(conf);
+
+    // Determine the working dir path
+    if (argv.length < 2) {
+      throw new IOException(String.format("No target directory specified by args %s",
+              Arrays.toString(argv)));
+    }
```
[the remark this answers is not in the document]
jiacheliu3: On second thought I think throwing an InvalidArgumentException in `CommandUtils.checkNumOfArgsNoLessThan()` flavor is consistent to existing commands. What do you think? Do you have concerns about throwing an exception here?

shell/src/main/java/alluxio/cli/bundler/InfoCollector.java
```diff
@@ -0,0 +1,171 @@
...
+
+      if (childRet > ret) {
+        ret = childRet;
+      }
+    }
+
+    // TODO(jiacheng): phase 2 add an option to disable bundle
+    // Generate bundle
+    System.out.println(String.format("Archiving dir %s", targetDirPath));
+    System.out.println(String.format("Files to include: %s", filesToCollect));
```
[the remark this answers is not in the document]
jiacheliu3: Now prints an error message and generates an empty tarball.

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
```diff
@@ -0,0 +1,249 @@
...
+    System.out.println(String.format("Init thread pool for %s hosts", allHosts.size()));
+    shellAll.mExecutor = Executors.newFixedThreadPool(allHosts.size());
+
+    // Invoke infoBundle on each host
+    List<CompletableFuture<CommandReturn>> futureList = new ArrayList<>();
+    for (String host : allHosts) {
+      System.out.println(String.format("Execute InfoCollector on host %s", host));
+
+      CompletableFuture<CommandReturn> future = CompletableFuture.supplyAsync(() -> {
+        String workDir = conf.get(PropertyKey.WORK_DIR);
```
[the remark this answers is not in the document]
jiacheliu3: That's true. Do you think it's possible to determine the remote WORK_DIR? I think we have to make the assumption that the WORK_DIR should be the same or we won't be able to find Alluxio.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: That's a great idea but i'm a little not sure. For a few reasons these CollectConfigCommand and CollectLogCommand may diverge more in the implementation in the future:
1. Copying config files should be made more security-aware.
2. Copying logs may require more optimization and tuning as there can be a lot. (eg. 10MB for each log and 100 log files rotation)

shell/src/main/java/alluxio/cli/bundler/command/CollectEnvCommand.java
jiacheliu3: `CollectAlluxioInfoCommand` now extends `CollectEnvCommand`. PTAL at the new logic. Thanks!

shell/src/main/java/alluxio/cli/bundler/command/CollectJvmInfoCommand.java
jiacheliu3: Will this dump too many Java processes that are not necessarily relevant? For example there can be ZooKeeper, Yarn, HDFS and Spark (probably other apps).

My assumption is, this command is run by the one who started the Alluxio cluster. Do you think this is a valid assumption to make?

shell/src/main/java/alluxio/cli/bundler/TarUtils.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+/**
+ * Utilities for generating .tar.gz files.
+ *
+ * Ref: https://memorynotfound.com/java-tar-example-compress-decompress-tar-tar-gz-files/
```
jiacheliu3: @calvinjia Do you have general advice on how we quote public code? Should this be rewritten?

jiacheliu3:
Please ignore the conflict with HEAD now. I will rebase once we settle on the reviews.

shell/src/main/java/alluxio/cli/bundler/InfoCollector.java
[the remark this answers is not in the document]
bf8086: `checkNumOfArgsNoLessThan` is used in places where the `InvalidArgumentException` is caught and printed out nicely(alluxio/cli/AbstractShell.java:130):
```
    try {
      ...
      cmdline = command.parseAndValidateArgs(currArgs);
    } catch (InvalidArgumentException e) {
      // It outputs a prompt message when passing wrong args to CLI
      System.out.println(e.getMessage());
      System.out.println("Usage: " + command.getUsage());
      System.out.println(command.getDescription());
      LOG.error("Invalid arguments for command {}:", command.getCommandName(), e);
      return -1;
    }
```

shell/src/main/java/alluxio/cli/bundler/InfoCollectorAll.java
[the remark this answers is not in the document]
bf8086: @ns1123 Can we safely assume user of this command can run it in one of the server node or their alluxio client is in the same path as the server nodes?

If we need to deal with different remote path, a quick way is to allow a path argument that specify remote Alluxio path.

shell/src/main/java/alluxio/cli/bundler/command/CollectJvmInfoCommand.java
bf8086: It can be a different user than the alluxio service user. @ns1123 What do you think?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
bf8086: In that case then we can leave it as is.

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
```diff
@@ -0,0 +1,177 @@
...
+        ret = childRet;
+      }
+    }
+
+    // TODO(jiacheng): phase 2 add an option to disable bundle
+    // Generate bundle
+    System.out.format("Archiving dir %s", targetDirPath);
+    String tarballPath = Paths.get(targetDirPath, TARBALL_NAME).toAbsolutePath().toString();
+    if (filesToCollect.size() == 0) {
+      System.err.format("No files to add. Tarball %s will be empty!", tarballPath);
```
bf8086: Why do we need to generate an empty tarball?

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,259 @@
...
+    } catch (InterruptedException | ExecutionException e) {
+      System.err.format("Failed to collect the results. Error is %s", e.getMessage());
+      LOG.error("Error: %s", e);
+      return;
+    }
+
+    // Inspect all command results
+    if (results.size() != allHosts.size()) {
+      System.out.format(
+              "Error occurred while collection information on the following hosts: "
```
bf8086: My bad, should be "Error occurred while collecting information". Maybe just say "Error occurred while collection information on %d/%d hosts."

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.concurrent.Executors;
+import java.util.stream.Collectors;
+
+/**
+ * Class for collecting various information about all nodes in the cluster.
+ */
+public class CollectInfoAll extends AbstractShell {
+  private static final Logger LOG = LoggerFactory.getLogger(CollectInfoAll.class);
+  private static final String TEMP_DIR_NAME = "putAllRemoteTarballs";
+  private static final String FINAL_TARBALL_NAME =  "alluxio-cluster-info-{timestamp}.tar.gz";
```
bf8086: does {} work in `String.format` as argument placeholder?

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,259 @@
...
+  @Override
+  protected String getShellName() {
+    return "collectInfoAll";
+  }
+
+  @Override
+  protected Map<String, Command> loadCommands() {
+    // We do not load any commands here.
+    // All calls will be delegated to CollectInfo local calls on each host.
+    return new HashMap<>();
```
bf8086: You can just use Collections.EMPTY_MAP

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
```diff
@@ -0,0 +1,177 @@
...
+
+  @Override
+  public String getCommandName() {
+    return COMMAND_NAME;
+  }
+
+//  @Override
+//  public int run(CommandLine cl) throws AlluxioException, IOException {
+//    int ret = 0;
+//
```
bf8086: remove?

shell/src/main/java/alluxio/cli/bundler/command/CollectEnvCommand.java
bf8086: It would be more maintainable to create a common base class for the two commands, e.g. `ExecuteShellScriptCollectorCommand`. Having `CollectAlluxioInfoCommand` extend `CollectEnvCommand` makes it harder to change `CollectEnvCommand` without breaking `CollectAlluxioInfoCommand` .

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
jiacheliu3: On 2nd thought I think I was wrong, an empty tarball shouldn't be generated. At least 1 operation will be executed and at least 1 files will be generated. So having no files to tarball implies an error state.

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
jiacheliu3: Actually `new String[] {alluxioPath, cmd}` doesn't work. Seems `bin/alluxio` wants the commands to be a string array rather than one big concatenated string.

shell/src/main/java/alluxio/cli/bundler/InfoCollector.java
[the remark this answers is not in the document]
jiacheliu3: Gotcha! Changed this to reporting incorrect arg number then `printUsage()`. PTAL

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,300 @@
...
+    // Validate commands
+    if (argv.length < 2) {
+      shellAll.printUsage();
+      System.exit(-1);
+    }
+    String targetDir = argv[1];
+
+    List<String> allHosts = new ArrayList<>(shellAll.getHosts());
+    System.out.format("Init thread pool for %s hosts%n", allHosts.size());
+    shellAll.mExecutor = Executors.newFixedThreadPool(allHosts.size());
```
jiacheliu3: Is this a reasonable thread pool? What if the cluster is very big?

jiacheliu3:
Edit: Slightly improved the command names
`infoBundle` -> `collectInfo`
`infoBundleAll` -> `collectInfoAll`
`runAlluxioCheck` -> `collectAlluxioInfo`

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
bf8086: If no file is generated, then the empty tarball contains no useful information. Should we just print error message and not generate any tarball?

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
bf8086: probably okay for the execution stage where we are just waiting for commands to be executed on the remote servers. It might be too many concurrent connections trying to copy files back from large cluster.

shell/src/main/java/alluxio/cli/bundler/command/CollectMetricsCommand.java
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Command to probe Alluxio metrics for a few times.
+ * */
+public class CollectMetricsCommand extends AbstractCollectInfoCommand {
+  public static final String COMMAND_NAME = "collectMetrics";
+  private static final Logger LOG = LoggerFactory.getLogger(CollectMetricsCommand.class);
+  private static final int COLLECT_METRICS_INTERVAL = 3 * 1000;
+  private static final int COLLECT_METRICS_TIMES = 3;
+  private static final int COLLECT_METRICS_TIMEOUT = 5 * 1000;
+  private static final String METRICS_SERVLET_PATH = "/metrics/json/";
```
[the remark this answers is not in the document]
bf8086: Not on the client. For server code you can find it at `alluxio.metrics.sink.MetricsServlet#SERVLET_PATH`

shell/src/test/java/alluxio/cli/bundler/TarUtilsTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import alluxio.AlluxioTestDirectory;
+
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class TarUtilsTest {
```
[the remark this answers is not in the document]
bf8086: Should be okay. We can move it later if other features want to use it.

shell/src/main/java/alluxio/cli/bundler/command/CollectEnvCommand.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Command to run a set of bash commands to get system information.
+ * */
+public class CollectEnvCommand extends ExecuteShellCollectInfoCommand {
+  public static final String COMMAND_NAME = "collectEnv";
+  private static final Logger LOG = LoggerFactory.getLogger(CollectEnvCommand.class);
+
+  protected Map<String, ShellCommand> mCommands;
+  protected Map<String, ShellCommand> mCommandsAlt;
```
bf8086: This is already defined in the super class.

shell/src/main/java/alluxio/cli/bundler/command/CollectEnvCommand.java
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Creates a new instance of {@link CollectEnvCommand}.
+   *
+   * @param fsContext the {@link FileSystemContext} to execute in
+   * */
+  public CollectEnvCommand(FileSystemContext fsContext) {
+    super(fsContext);
+    mCommands = new HashMap<>();
+    mCommandsAlt = new HashMap<>();
```
bf8086: This is already initialized by super class.

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,300 @@
...
+      CompletableFuture<CommandReturn> future = CompletableFuture.supplyAsync(() -> {
+        // We make the assumption that the Alluxio WORK_DIR is the same
+        String workDir = conf.get(PropertyKey.WORK_DIR);
+        String alluxioBinPath = Paths.get(workDir, "bin/alluxio")
+                .toAbsolutePath().toString();
+        System.out.format("host: %s, alluxio path %s%n", host, alluxioBinPath);
+
+        try {
+          String[] infoBundleArgs =
+                  (String[]) ArrayUtils.addAll(new String[]{alluxioBinPath, "infoBundle"}, argv);
```
bf8086: Does `infoBundle ` need to be renamed?

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,300 @@
...
+          CommandReturn cr = ShellUtils.sshExecCommandWithOutput(host, infoBundleArgs);
+          return cr;
+        } catch (Exception e) {
+          LOG.error("Execution failed %s", e);
+          return new CommandReturn(1, e.toString());
+        }
+      }, shellAll.mExecutor);
+      sshFutureList.add(future);
+      System.out.format("Invoked infoBundle command on host %s%n", host);
```
bf8086: rename `infoBundle`?

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    // Collect SSH execution results
+    List<String> sshSucceededHosts =
+            shellAll.collectCommandReturnsFromHosts(sshFutureList, allHosts);
+
+    // If all executions failed, clean up and exit
+    if (sshSucceededHosts.size() == 0) {
+      System.err.println("Failed to invoke infoBundle command on all hosts!");
```
bf8086: rename infoBundle?

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
```diff
@@ -0,0 +1,175 @@
...
+   *
+   * @param alluxioConf Alluxio configuration
+   */
+  public CollectInfo(InstancedConfiguration alluxioConf) {
+    super(CMD_ALIAS, UNSTABLE_ALIAS, alluxioConf);
+  }
+
+  @Override
+  protected String getShellName() {
+    return "infoBundle";
```
bf8086: rename `infoBundle`?

shell/src/main/java/alluxio/cli/bundler/command/CollectMetricsCommand.java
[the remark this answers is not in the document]
jiacheliu3: Yea that agrees with what I find out

jiacheliu3:
`AlluxioTestDirectory` moving change has been reverted. One potential concern is I had to implement similar temp directory logic in `InfoCollectorTestUtils#createTemporaryDirectory`.

Another alternative was moving all these tests into integration tests under `alluxio/tests`. But I would say most of the tests involved are closer to unit tests.

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
jiacheliu3: Due to resource limitation I can only test with a small cluster with probably big log files (maybe 100 log files rotation). Personally I prefer to optimize this afterwards, with real feedback from users. @ns1123 what do you think?

gpang:
@jiacheliu3 have you considered having a single command , and just using an option like `--local` for just collecting the local info? It is confusing to have 2 commands `collectInfo` and `collectInfoAll`

core/common/src/main/java/alluxio/cli/CommandUtils.java
```diff
@@ -104,4 +111,33 @@ public static void checkNumOfArgsNoMoreThan(Command cmd, CommandLine cl, int n)
...
+    for (String line : lines) {
+      String node = line.trim();
+      if (node.startsWith("#") || node.length() == 0) {
+        continue;
+      }
+      nodes.add(node);
+    }
+
+    return nodes;
+  }
```
ZacBlanco: It's possible that many people don't set up their masters and workers file if they use an external system to bring up their processes. Do you think we could fall back, or cross-check with a list of masters/workers from the master to make the lives of users easier in case they don't configure their `masters` or `workers` file?

ns1123: @ZacBlanco I think we can expect users to have the masters/workers file populated for this scenario (at least for the first iteration). The idea was that this command would have the same requirements as './bin/alluxio-start.sh all'. Future versions may get smarter about this.

shell/src/main/java/alluxio/cli/bundler/CollectInfoAll.java
bf8086: Can we add an option to the command to limit the maximum thread pool size? Then here we can do something like `shellAll.mExecutor = Executors.newFixedThreadPool(Math.min(allHosts.size(), maxThreads))`

bf8086:
> `AlluxioTestDirectory` moving change has been reverted. One potential concern is I had to implement similar temp directory logic in `InfoCollectorTestUtils#createTemporaryDirectory`.
> 
> Another alternative was moving all these tests into integration tests under `alluxio/tests`. But I would say most of the tests involved are closer to unit tests.

Can we use `Files.createTempDir()` instead?

jiacheliu3:
> > `AlluxioTestDirectory` moving change has been reverted. One potential concern is I had to implement similar temp directory logic in `InfoCollectorTestUtils#createTemporaryDirectory`.
> > Another alternative was moving all these tests into integration tests under `alluxio/tests`. But I would say most of the tests involved are closer to unit tests.
> 
> @jiacheliu3 Can we use `Files.createTempDir()` instead?

@bf8086 `AlluxioTestDirectory#createTemporaryDirectory` is deleting the temp directory when the JVM exits. That is the reason it works well for tests. I'm adding this logic to `InfoCollectorTestUtils#createTemporaryDirectory` so it equals `Files.createTempDir() + delete temp dir`.

core/common/src/main/java/alluxio/cli/CommandUtils.java
jiacheliu3: @ZacBlanco the master should be able to give a list of current nodes(masters + workers) but that has downsides too. For example, if some workers are lost, do we want to collect information from those too? Also, when we want to run this command it's possible that the cluster is in an error state(or just down).

Considering the complexity and the level of correctness I think reading `conf/masters` and `conf/workers` is the easiest way. I can add more cross-checks as Zac recommends in future versions.

jiacheliu3: Also left a TODO on `CollectInfoAll.getHosts()`

ZacBlanco: My main use case I'm thinking of is EMR. We don't do any population of the workers files AFAIK, so it wouldn't be functional in that environment. IMO I think it should work regardless

jiacheliu3:
> @jiacheliu3 have you considered having a single command , and just using an option like `--local` for just collecting the local info? It is confusing to have 2 commands `collectInfo` and `collectInfoAll`

Good idea. Now `CollectInfoAll` has been merged into `CollectInfo` and added a `--local` option as the switch. @gpang PTAL at the new logic now. I may add more unit test coverage tmr.

gpang:
@jiacheliu3 Thanks! Can you update the PR title and description with the updated syntax?

jiacheliu3:
> @jiacheliu3 Thanks! Can you update the PR title and description with the updated syntax?

This is done. Thanks for the catch!

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
```diff
@@ -0,0 +1,97 @@
...
+    // TODO(jiacheng): a command to find lost blocks?
+    registerCommand("getConf",
+            new AlluxioCommand(mAlluxioPath, "getConf --master --source"), null);
+    registerCommand("fsadmin",
+            new AlluxioCommand(mAlluxioPath, "fsadmin report"), null);
+    registerCommand("mount",
+            new AlluxioCommand(mAlluxioPath, "fs mount"), null);
+    registerCommand("version",
+            new AlluxioCommand(mAlluxioPath, "version -r"), null);
+    registerCommand("job",
```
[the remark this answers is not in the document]
ZacBlanco: job ID's might not be immediately useful, but a string dump of the Job type and config for each would be the most useful IMO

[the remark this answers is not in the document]
bradyoo: FYI @ns1123 job ls in 2.2 prints Job Type

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
```diff
@@ -151,12 +376,64 @@ private AbstractCollectInfoCommand findCommand(String cmdName) {
...
+        CommandReturn c = results.get(i);
+        String host = hosts.get(i);
+        if (c.getExitCode() != 0) {
+          System.out.format("Execution failed on host %s%n", host);
+          System.out.println(c.getFormattedOutput());
+          continue;
+        }
+        successfulHosts.add(host);
+      }
+      System.out.format("Command successful on %d/%d hosts.%n",
```
bf8086: (nit) "Command executed successfully on %d/%d hosts."

shell/src/main/java/alluxio/cli/bundler/command/CollectEnvCommand.java
```diff
@@ -0,0 +1,79 @@
...
+   * @param fsContext the {@link FileSystemContext} to execute in
+   * */
+  public CollectEnvCommand(FileSystemContext fsContext) {
+    super(fsContext);
+    registerCommands();
+  }
+
+  @Override
+  protected void registerCommands() {
+    registerCommand("ps", new ShellCommand(new String[]{"ps", "-ef", "|grep alluxio*"}), null);
```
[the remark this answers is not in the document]
jiacheliu3: Added attempts to grab Spark/Yarn/Hdfs/Presto.

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
```diff
@@ -0,0 +1,97 @@
...
+            new AlluxioCommand(mAlluxioPath, "job ls"), null);
+    registerCommand("journal",
+            new AlluxioCommand(mAlluxioPath, String.format("fs ls -R %s",
```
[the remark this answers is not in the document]
jiacheliu3: Added logic to check if the journal dir is local or hdfs.
Local -> `ls -R journalDir`
Hdfs -> `hdfs dfs -ls -R journalDir`
Are there any more logic that can be helpful?

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
```diff
@@ -19,59 +19,284 @@
...
-    // Execute the Collectors one by one
-    // Reduce the RPC retry max duration to fall earlier for CLIs
+    // Reduce the RPC retry max duration to fail earlier for CLIs
     conf.set(PropertyKey.USER_RPC_RETRY_MAX_DURATION, "5s", Source.DEFAULT);
     CollectInfo shell = new CollectInfo(conf);
 
-    // If the args are not valid, return early
-    if (argv.length < 2) {
-      System.out.format("Command %s requires at least %s arguments (%s provided)%n",
-              2, argv.length);
```
bf8086: Why is the error output removed? It would be helpful for user to know what exactly is wrong with the argument.

shell/src/main/java/alluxio/cli/bundler/command/CollectJvmInfoCommand.java
jiacheliu3: Now the command attempts to do `sudo jps` and falls back to just `jps`.
When normal `jstack` fails on the pid, try `sudo jstack`. 
WRT to the cgroup constraint, I don't think that can be easily resolved in this. We can discuss further on this.

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
jiacheliu3: My bad, that's an overlook.

jiacheliu3:
@ns1123 PTAL now the `CollectJvmInfoCommand` attempts sudo.

jiacheliu3:
@bf8086 @ns1123 if you have great ideas on improvements that are not low-hanging fruits (like a smarter extra command option), I would propose we leave a mark there and I can address it in the phase 2 improvement on this command. This PR is getting a little too long to keep track and keep focused on all aspects.

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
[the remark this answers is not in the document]
ns1123: Cool.

shell/src/main/java/alluxio/cli/bundler/command/CollectAlluxioInfoCommand.java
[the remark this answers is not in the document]
ns1123: This pretty much covers all scenarios that we want to.

ns1123:
LGTM. Thanks @jiacheliu3 for doing this! looking forward to getting to use it.

calvinjia:

PR #10606, opened 2019-12-10
Avoid registering listener for channel state change

ggezer:
Calling `notifyWhenStateChanged()` on managed channel will cause accumulation of listeners and could potentially cause OOM in massive channel create/delete scenarios.

We can safely discard making this call since we keep channel's authentication stream active. Server state changes that effect authentication will reflect in authentication stream's `onError()` method as well. This provides more stable accounting as earlier tracking was sensitive to unrelated connection level state changes as well.


ggezer:
@bradyoo You might wanna try your scenario with this fix.

bradyoo:
Oh this looks potentially interesting!

calvinjia:

bradyoo:
If this were to be needed to be cherry-picked to 2.1.1 would that be fine? @calvinjia, @ggezer

ggezer:
Yes it should be compatible.

calvinjia:

PR #10625, opened 2019-12-11
Shutdown JobWorker on high load of CPU

bradyoo:
There are 2 components of this PR:

1. Throttle JobWorker on high load of CPU
2. The task is no longer considered Status.RUNNING until the task starts being processed. Therefore, the parent plan is not considered Status.RUNNING until one of its task starts being processed.


bradyoo:
@calvinjia:

Running this code locally, my distributedCp job that was taking 30~40 seconds was taking about a minute and a more stable CPU graph. So it's "working" though I don't know if this is exactly what we want. (Will need better graphs + experiments.)

But more importantly, because the amplab machines are always throttled, it seems very hard to run jobs with this throttling logic on. I may have to go to having at least 1 worker no matter how CPU throttled the machine is because or else, it's going to be hard to get any tests through.

calvinjia:
@bradyoo I think we can disable throttling during tests

bradyoo:
@calvinjia: What is the main method we use to figure out whether we are in the test environment?

calvinjia:
`PropertyKey.TEST_MODE` should be set

bradyoo:
jenkins, test this please.

job/server/src/main/java/alluxio/worker/job/command/CommandHandlingExecutor.java
```diff
@@ -79,11 +81,23 @@ public CommandHandlingExecutor(JobServerContext jobServerContext,
     mTaskExecutorManager = Preconditions.checkNotNull(taskExecutorManager, "taskExecutorManager");
     mMasterClient = Preconditions.checkNotNull(masterClient, "masterClient");
     mWorkerNetAddress = Preconditions.checkNotNull(workerNetAddress, "workerNetAddress");
-    mHealthReporter = new JobWorkerHealthReporter();
+    if (ServerConfiguration.getBoolean(PropertyKey.TEST_MODE)) {
+      mHealthReporter = new AlwaysHealthyJobWorkerHealthReporter();
```
calvinjia: Instead of doing a switch here, perhaps add logic in the `JobWorkerHealthReporter` to make the health factor configurable?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -101,13 +106,56 @@ public void setTaskExecutorPoolSize(int taskExecutorPoolSize) {
...
+  public synchronized void throttle() {
+    mThrottled = true;
+    setTaskExecutorPoolSize(0);
+  }
+
+  /**
+   * Unthrottle the TaskExecutorManager. This sets the internal thread pool size back to the
+   * default thread pool size.
+   */
+  public synchronized void unthrottle() {
```
calvinjia: Instead of having a binary throttle state, should we make it a sliding scale?

bradyoo: We absolutely could. But I decided not to this time around because it won't be used and if I need it later, the underlying mechanism will support it anyways.

job/server/src/main/java/alluxio/worker/job/command/CommandHandlingExecutor.java
bradyoo: I don't think I prefer that given that just mixes more logic inside JobWorkerHealthReporter file.

calvinjia: Wouldn't the `JobWorkerHealthReporter` benefit from being able to be configured? Is `1.0 * # of available processors` always the right threshold?

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
calvinjia: We can address this separately. I'm just concerned that a binary throttling mechanism will not be effective, have you seen it work well?

bradyoo: It works okay-ish for now. The thing is that the metric I'm trying to improve is a 1 minute average so even with spikes, it is an improvement.

job/server/src/main/java/alluxio/worker/job/command/CommandHandlingExecutor.java
bradyoo: I think configuration may be necessary later but even then, I think a class that always returns True for isHealthy is likely better.

bradyoo:
@calvinjia  There are certainly benefits of adding configuration but I expect I will need to change configuration often enough later that I don't want to add configuration yet.

job/server/src/main/java/alluxio/worker/job/command/CommandHandlingExecutor.java
calvinjia: I'm happy with a test version of the health reporter, but I don't like injecting it in the production code itself, it usually is handled though some kind of mocking/test framework.

bradyoo:
I think what I'm going to do here instead is to have a new configuration called whether throttling should be on or off.

bradyoo:
[INFO] --- copy-rename-maven-plugin:1.0:copy (copy-and-rename-file) @ alluxio-shaded-client ---
[INFO] Copied /usr/src/alluxio/shaded/client/target/alluxio-shaded-client-2.2.0-SNAPSHOT.jar to /usr/src/alluxio/client/alluxio-2.2.0-SNAPSHOT-client.jar
Build timed out (after 120 minutes). Marking the build as aborted.
Build was aborted
Archiving artifacts

Build timed out? Going to retry.

bradyoo:

calvinjia:

PR #10640, opened 2019-12-17
Add ability to cancel workflows

bradyoo:


bradyoo:
@calvinjia

job/server/src/main/java/alluxio/master/job/JobMaster.java
```diff
@@ -257,15 +257,17 @@ public synchronized void run(JobConfig jobConfig, long jobId)
   }
 
   /**
-   * Cancels a job.
+   * Cancels a job. Does nothing if the job doesn't exist.
```
calvinjia: Don't we throw `JobDoesNotExistException` still?

job/server/src/main/java/alluxio/master/job/workflow/WorkflowTracker.java
```diff
@@ -77,6 +77,27 @@ public synchronized void run(WorkflowConfig workflowConfig, long jobId)
...
+
+    for (Long child : children) {
+      try {
+        mJobMaster.cancel(child);
+      } catch (JobDoesNotExistException e) {
+        LOG.info("Tried to cancel jobId: {} but the job did not exist", child);
+      }
+    }
+    return true;
+  }
```
calvinjia: What are the guarantees for canceling a workflow in different cases, for example if some children have succeeded? Will we ever have partial job completion/cancellation?

bradyoo: That is correct. In fact, even if this job has already failed, there may be some tasks that are running in parallel which this would also cancel.

job/server/src/main/java/alluxio/master/job/JobMaster.java
bradyoo: Oh oops. It used to in the middle. And then, I went back so let me remove this.

job/server/src/main/java/alluxio/master/job/workflow/WorkflowTracker.java
calvinjia: In the case that the job has completed some tasks and we cancel the rest, will there be partial state, or are the guarantees up to each individual job to implement?

bradyoo: It cancels everything it "can". In other words, it goes to all of the leaves (tasks) and cancels those. 

Even for plan cancelling, it just cancels everything it has.

calvinjia:

PR #10650, opened 2019-12-18
Shutdown channel immediately if health-check failed

ggezer:
`awaitTermination()` on channel that has been shutdown gracefully will not return if a health check was triggered on it. This causes `PollingMasterInquireClient` to be blocked for the full channel shutdown timeout per master. (master count x channel shutdown timeout)

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
@@ -213,15 +213,16 @@ public class GrpcManagedChannelPool {
       managedChannel.shutdown();
     }
     try {
-      managedChannel.awaitTermination(shutdownTimeoutMs, TimeUnit.MILLISECONDS);
+      if (!managedChannel.awaitTermination(shutdownTimeoutMs, TimeUnit.MILLISECONDS)) {
+        LOG.warn("Timed out shutting down managed channel: {}. ", managedChannel);
+        // Forcefully shut down before returning if did not already.
+        if (!shutdownNow) {
+          managedChannel.shutdownNow();
+        }
+      }
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
       // Allow thread to exit.
-    } finally {
-      // Ensure shutdown if not not closed forcefully.
-      if (!shutdownNow) {
-        managedChannel.shutdownNow();
-      }
     }
   }
 
--- a/core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
@@ -94,6 +94,7 @@ public class GrpcManagedChannelPool {
         existingRefCount = chHolder.getRefCount();
         LOG.debug("Shutting down an existing unhealthy managed channel. "
             + "ChannelKey: {}. Existing Ref-count: {}", key, existingRefCount);
+        // Shutdown the channel forcefully as it's already unhealthy.
         shutdownManagedChannel(chHolder.get(), true, shutdownTimeoutMs);
       }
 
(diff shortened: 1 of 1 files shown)


ggezer:
Jenkins, test this please

gpang:
Jenkins, test this please

ggezer:
Jenkins, test this please

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
```diff
@@ -202,15 +202,24 @@ private boolean waitForChannelReady(ManagedChannel managedChannel, long healthCh
   /**
    * Shuts down the managed channel.
    */
-  private void shutdownManagedChannel(ManagedChannel managedChannel, long shutdownTimeoutMs) {
-    managedChannel.shutdown();
+  private void shutdownManagedChannel(ManagedChannel managedChannel, boolean shutdownNow,
+      long shutdownTimeoutMs) {
+    // Shutdown channel forcefully if requested.
+    if (shutdownNow) {
+      managedChannel.shutdownNow();
```
calvinjia: Do we need to run the following code if `shutdownNow` is true?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
```diff
@@ -94,7 +94,7 @@ public ManagedChannel acquireManagedChannel(GrpcChannelKey channelKey, long heal
         existingRefCount = chHolder.getRefCount();
         LOG.debug("Shutting down an existing unhealthy managed channel. "
             + "ChannelKey: {}. Existing Ref-count: {}", key, existingRefCount);
-        shutdownManagedChannel(chHolder.get(), shutdownTimeoutMs);
+        shutdownManagedChannel(chHolder.get(), true, shutdownTimeoutMs);
```
gpang: can you add a comment why this should use shutdownNow?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
```diff
@@ -115,7 +115,7 @@ public void releaseManagedChannel(GrpcChannelKey channelKey, long shutdownTimeou
       Preconditions.checkNotNull(chHolder, "Releasing nonexistent channel");
       if (chHolder.dereference() == 0) {
         LOG.debug("Released managed channel for: {}. Ref-count: {}", key, chHolder.getRefCount());
-        shutdownManagedChannel(chHolder.get(), shutdownTimeoutMs);
+        shutdownManagedChannel(chHolder.get(), false, shutdownTimeoutMs);
```
gpang: can you add a comment why this should not use shutdownNow?

core/common/src/main/java/alluxio/grpc/GrpcManagedChannelPool.java
ggezer: I guess we'd still want to guarantee complete termination before allowing another client to potentially open a channel to same target.

From `ManagedChannel#shutdownNow()` java doc:

   * Initiates a forceful shutdown in which preexisting and new calls are cancelled. Although
   * forceful, the shutdown process is still not instantaneous; {@link #isTerminated()} will likely
   * return {@code false} immediately after this method returns.

calvinjia: Makes sense, is the idea that `isTerminated` will return true when this method returns? We handle the two calls to `shutdownNow` in this method differently (we do not await termination again in finally)

ggezer: Yes. Updated the method to avoid multiple `shutdownNow()`s.

ggezer:
jenkins, test this please

ggezer:

ggezer:

PR #10658, opened 2019-12-19
Add SSH and SCP command utility and refactor shell utility

jiacheliu3:
This change adds a wrapper to SSH executing a command to ShellUtils, together with SCP commands.

`ShellUtils` class refactored with inner classed extracted to standalone classes.

PR https://github.com/Alluxio/alluxio/pull/10596 relies on this utility as it extensively invokes commands over SSH.

### DIFF
--- a/core/common/src/main/java/alluxio/shell/ScpCommand.java
+++ b/core/common/src/main/java/alluxio/shell/ScpCommand.java
@@ -15,14 +15,33 @@ import alluxio.util.ShellUtils;
 
 import javax.annotation.concurrent.NotThreadSafe;
 
+/**
+ * Object representation of a remote scp command.
+ * The scp command copies a file/dir from remote host to local.
+ */
 @NotThreadSafe
 public class ScpCommand extends ShellCommand {
   protected String mHostName;
 
-  private ScpCommand(String remoteHost, String fromFile, String toFile) {
+  /**
+   * Creates a remote scp command to copy a file to local.
+   *
+   * @param remoteHost remote hostname
+   * @param fromFile the remote file
+   * @param toFile target location to copy to
+   */
+  public ScpCommand(String remoteHost, String fromFile, String toFile) {
     this(remoteHost, fromFile, toFile, false);
   }
 
+  /**
+   * Creates a remote scp command to copy a file/dir to local.
+   *
+   * @param remoteHost the remote hostname
+   * @param fromFile the remote file/dir to copy
+   * @param toFile target path to copy to
+   * @param isDir if true, the remote file is a directory
+   */
   public ScpCommand(String remoteHost, String fromFile, String toFile, boolean isDir) {
     super(new String[]{});
 
@@ -38,7 +57,12 @@ public class ScpCommand extends ShellCommand {
     mHostName = remoteHost;
   }
 
-  protected String getHostName() {
+  /**
+   * Returns the remote target hostname.
+   *
+   * @return target hostname
+   */
+  public String getHostName() {
     return mHostName;
   }
 }
--- a/core/common/src/main/java/alluxio/shell/ScpCommand.java
+++ b/core/common/src/main/java/alluxio/shell/ScpCommand.java
@@ -21,7 +21,7 @@ import javax.annotation.concurrent.NotThreadSafe;
  */
 @NotThreadSafe
 public class ScpCommand extends ShellCommand {
-  protected String mHostName;
+  private final String mHostName;
 
   /**
    * Creates a remote scp command to copy a file to local.
@@ -43,17 +43,10 @@ public class ScpCommand extends ShellCommand {
    * @param isDir if true, the remote file is a directory
    */
   public ScpCommand(String remoteHost, String fromFile, String toFile, boolean isDir) {
-    super(new String[]{});
-
-    String template = "scp %s %s:%s localhost:%s";
-    if (isDir) {
-      template = "scp -r %s %s:%s localhost:%s";
-    }
-
-    // copy from hostname to local
-    mCommand = new String[]{"bash", "-c",
-            String.format(template, ShellUtils.COMMON_SSH_OPTS, remoteHost, fromFile, toFile)
-    };
+    super(new String[]{"bash", "-c",
+            String.format(isDir ? "scp -r %s %s:%s localhost:%s" : "scp %s %s:%s localhost:%s",
+                    ShellUtils.COMMON_SSH_OPTS, remoteHost, fromFile, toFile)
+    });
     mHostName = remoteHost;
   }
 
--- a/core/common/src/main/java/alluxio/shell/ScpCommand.java
+++ b/core/common/src/main/java/alluxio/shell/ScpCommand.java
@@ -1,0 +1,44 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.shell;
+
+import alluxio.util.ShellUtils;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+@NotThreadSafe
+public class ScpCommand extends ShellCommand {
+  protected String mHostName;
+
+  private ScpCommand(String remoteHost, String fromFile, String toFile) {
+    this(remoteHost, fromFile, toFile, false);
+  }
+
+  public ScpCommand(String remoteHost, String fromFile, String toFile, boolean isDir) {
+    super(new String[]{});
+
+    String template = "scp %s %s:%s localhost:%s";
+    if (isDir) {
+      template = "scp -r %s %s:%s localhost:%s";
+    }
+
+    // copy from hostname to local
+    mCommand = new String[]{"bash", "-c",
+            String.format(template, ShellUtils.COMMON_SSH_OPTS, remoteHost, fromFile, toFile)
+    };
+    mHostName = remoteHost;
+  }
+
+  protected String getHostName() {
+    return mHostName;
+  }
+}
--- a/core/common/src/main/java/alluxio/shell/ShellCommand.java
+++ b/core/common/src/main/java/alluxio/shell/ShellCommand.java
@@ -12,6 +12,7 @@
 package alluxio.shell;
 
 import alluxio.util.ShellUtils;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -35,7 +36,7 @@ public class ShellCommand {
   /**
    * Creates a ShellCommand object with the command to exec.
    *
-   * @param execString
+   * @param execString shell command
    */
   public ShellCommand(String[] execString) {
     mCommand = execString.clone();
--- a/core/common/src/main/java/alluxio/shell/ShellCommand.java
+++ b/core/common/src/main/java/alluxio/shell/ShellCommand.java
@@ -1,0 +1,145 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.shell;
+
+import alluxio.util.ShellUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.concurrent.NotThreadSafe;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+
+@NotThreadSafe
+public class ShellCommand {
+  private static final Logger LOG = LoggerFactory.getLogger(ShellCommand.class);
+
+  protected String[] mCommand;
+
+  public ShellCommand(String[] execString) {
+    mCommand = execString.clone();
+  }
+
+  /**
+   * Runs a command and returns its stdout on success.
+   *
+   * @return the output
+   * @throws ShellUtils.ExitCodeException if the command returns a non-zero exit code
+   */
+  public String run() throws ShellUtils.ExitCodeException, IOException {
+    Process process = new ProcessBuilder(mCommand).redirectErrorStream(true).start();
+
+    BufferedReader inReader =
+            new BufferedReader(new InputStreamReader(process.getInputStream(),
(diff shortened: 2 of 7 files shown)


core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -135,8 +136,8 @@ public static UnixMountInfo parseMountInfo(String line) {
   }
 
   @NotThreadSafe
-  private static final class Command {
-    private String[] mCommand;
+  private static class Command {
+    protected String[] mCommand;
```
jiacheliu3: Changed to `protected` to be inherited

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -135,8 +136,8 @@ public static UnixMountInfo parseMountInfo(String line) {
   }
 
   @NotThreadSafe
-  private static final class Command {
-    private String[] mCommand;
+  private static class Command {
```
jiacheliu3: Removed `final` to be inherited by SshCommand class. I don't see much benefit from stopping this class being inherited.

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -148,7 +149,7 @@ private Command(String[] execString) {
      * @return the output
      * @throws ExitCodeException if the command returns a non-zero exit code
      */
-    private String run() throws ExitCodeException, IOException {
+    protected String run() throws ExitCodeException, IOException {
```
jiacheliu3: Enable inheritance

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,101 @@ private String run() throws ExitCodeException, IOException {
...
       }
     }
+
+    /**
+     * Runs a command and returns its stdout, stderr and exit code.
+     * No matter it succeeds or not.
+     *
+     * @return {@link CommandReturn} object representation of stdout, stderr and exit code
+     */
+    protected CommandReturn runTolerateFailure() throws IOException {
```
jiacheliu3: This method is similar to `run()`.

The differences are:
1. Not redirecting stderr to stdout. 
2. Keeping exitCode, stderr and stdout separated and wrapped in an object CommandReturn.
3. Not throwing exception when exitcode is non-zero. Return all information instead.

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,101 @@ private String run() throws ExitCodeException, IOException {
...
+        } catch (IOException e) {
+          LOG.warn("Error while closing the error stream", e);
+        }
+        process.destroy();
+      }
+    }
+  }
+
+  @NotThreadSafe
+  private static class SshCommand extends Command {
```
jiacheliu3: This SshCommand automatically does SSH with common parameters. I currently don't see the needs for flexible configuration yet.

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -234,14 +330,113 @@ public String toString() {
...
+      return mExitCode;
+    }
+
+    /**
+     * Formats the object to more readable format.
+     * This is not done in toString() because stdout and stderr may be long.
+     *
+     * @return pretty formatted output
+     */
+    public String getFormattedOutput() {
```
jiacheliu3: Not overriding `toString()` cuz the stdout and stderr may have a good amount of info. I don't want people to unconsciously execute a lot of commands and print everything.

core/common/src/test/java/alluxio/util/ShellUtilsTest.java
```diff
@@ -110,4 +134,66 @@ public void getMountInfo() throws Exception {
...
+    mExceptionRule.expect(IOException.class);
+    mExceptionRule.expectMessage("No such file or directory");
+    String[] testCommandExcept = new String[]{"lsa",
+            String.format("%s", testDir.getAbsolutePath())};
+    // lsa is not a valid executable
+    ShellUtils.execCommandTolerateFailure(testCommandExcept);
+  }
+
+  @Test
+  public void sshExecuteCommandTolerateFailure() throws Exception {
```
jiacheliu3: It's worth debating how/whether we want to test SSH. Currently this relies on password-less loopback access.

jiacheliu3: Seems Jenkins doesn't have password-less SSH to localhost. So either we manual test SSH, or bring in a 3rd part implementation of invoking SSH calls.

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,84 @@ private String run() throws ExitCodeException, IOException {
...
+        if (process != null) {
+          process.destroy();
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new IOException(e);
+      } finally {
+        if (process != null) {
+          process.destroy();
+        }
```
jiacheliu3: Is it okay to destroy the process and let the streams be managed by the try block?

In https://github.com/Alluxio/alluxio/blob/5c2bab41150b0d981238f21c3484e063e55f241a/core/common/src/main/java/alluxio/util/ShellUtils.java#L179 seems we tried to manually close the stream before destroying the process. Seems that's specific to certain Java 7 versions. Is this still the case now?

core/common/src/main/java/alluxio/util/ShellUtils.java
gpang: does it make sense to make `Command` a real class with its own file?

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -234,14 +341,131 @@ public String toString() {
   }
 
   /**
-   * Static method to execute a shell command.
+   * Object representation of a command execution.
+   */
+  public static class CommandReturn {
```
gpang: this might make sense as a separate class, or a part of `Command` (which should be a separate class)

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,112 @@ private String run() throws ExitCodeException, IOException {
...
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new IOException(e);
+      } finally {
+        if (process != null) {
+          process.destroy();
+        }
+      }
+      return cr;
```
gpang: what does it mean if we return a `null`?

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,112 @@ private String run() throws ExitCodeException, IOException {
...
+        String outLine = inReader.readLine();
+        while (outLine != null) {
+          stdout.append(outLine);
+          stdout.append("\n");
+          outLine = inReader.readLine();
+        }
+
+        // read the stderr of the command
+        StringBuilder stderr = new StringBuilder();
+        String errLine = errReader.readLine();
```
gpang: will reading err after fully reading stdout work? I'm not sure how these work, but do both streams "end" at the same time?

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,112 @@ private String run() throws ExitCodeException, IOException {
...
+      return cr;
+    }
+  }
+
+  @NotThreadSafe
+  private static class SshCommand extends Command {
+    protected String mHostName;
+
+    private SshCommand(String[] execString) {
+      this("localhost", execString);
```
gpang: why is this `localhost` hardcoded?

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -193,6 +194,112 @@ private String run() throws ExitCodeException, IOException {
...
+
+  @NotThreadSafe
+  private static class ScpCommand extends Command {
+    protected String mHostName;
+
+    private ScpCommand(String remoteHost, String fromFile, String toFile) {
+      this(remoteHost, fromFile, toFile, false);
+    }
+
+    private ScpCommand(String remoteHost, String fromFile, String toFile, boolean isDir) {
```
gpang: can you javadoc these parameters? (and the other constructor too)

core/common/src/main/java/alluxio/util/ShellUtils.java
jiacheliu3: You are right, ssh to localhost doesn't make any sense (except for testing)

core/common/src/main/java/alluxio/util/ShellUtils.java
jiacheliu3: Yea the `return` should happen in the `try` block.

core/common/src/main/java/alluxio/util/ShellUtils.java
jiacheliu3: javadoc added after extracting these inner classes to separate class files and changed to `public`.

core/common/src/main/java/alluxio/util/ShellUtils.java
jiacheliu3: Done

core/common/src/main/java/alluxio/util/ShellUtils.java
jiacheliu3: This is a great question! In short it's wrong to read the 2 streams sequentially like what I did here. The correct way is to read them in two separated threads. 

On 2nd thought, the benefit of having stderr separated from stdout is out-weighted by the overhead of having another thread reading stderr (stdout can be read in the same thread cuz we are blocking anyway). So I'm now redirecting stderr into stdout. The corresponding `stderr` field in `CommandReturn` class has been removed cuz there's currently no need of it.

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -234,13 +173,71 @@ public String toString() {
...
    * @return the output of the executed command
+   * @throws IOException in various situations:
+   *  1. {@link ExitCodeException} when exit code is non-zero
+   *  2. when the executable is not valid, i.e. running ls in Windows
+   *  3. execution interrupted
+   *  4. other normal reasons for IOException
    */
   public static String execCommand(String... cmd) throws IOException {
-    return new Command(cmd).run();
+    return new ShellCommand(cmd).run();
```
gpang: What are the existing uses of this? The new behavior now has stdout and stderr both as the returned output, and will that affect anything negatively?

core/common/src/main/java/alluxio/shell/ShellCommand.java
```diff
@@ -0,0 +1,145 @@
...
+        // fd is recycled. the stream draining thread will attempt to
+        // drain that fd!! it may block, OOM, or cause bizarre behavior
+        // see: https://bugs.openjdk.java.net/browse/JDK-8024521
+        // issue is fixed in build 7u60
+        InputStream stdout = process.getInputStream();
+        synchronized (stdout) {
+          inReader.close();
+        }
+      } catch (IOException e) {
+        LOG.warn("Error while closing the input stream", e);
```
gpang: Can we add a bit more info to this message, like `Error closing input stream of process: %s`?

core/common/src/main/java/alluxio/shell/ShellCommand.java
```diff
@@ -0,0 +1,145 @@
...
+        stdout.append(outLine);
+        stdout.append("\n");
+        outLine = inReader.readLine();
+      }
+
+      // wait for the process to finish and check the exit code
+      int exitCode = process.waitFor();
+      if (exitCode != 0) {
+        // log error instead of throwing exception
+        LOG.warn(String.format("Non-zero exit code from command %s%nExit Code: %s",
```
gpang: Can we keep the error code in the same line, something like "Non-zero exit code (%d) from command: %s"

core/common/src/main/java/alluxio/shell/SshCommand.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Creates a SshCommand instance from the remote hostname and command.
+   *
+   * @param hostname the remote target hostname
+   * @param execString the command to execute
+   */
+  public SshCommand(String hostname, String[] execString) {
+    super(execString);
+    mHostName = hostname;
+    mCommand = new String[]{"bash", "-c",
```
gpang: Can we just use this for the `super` constructor? (and hopefully keep `mCommand` `final`)

core/common/src/main/java/alluxio/shell/ScpCommand.java
```diff
@@ -0,0 +1,68 @@
...
+  public ScpCommand(String remoteHost, String fromFile, String toFile, boolean isDir) {
+    super(new String[]{});
+
+    String template = "scp %s %s:%s localhost:%s";
+    if (isDir) {
+      template = "scp -r %s %s:%s localhost:%s";
+    }
+
+    // copy from hostname to local
+    mCommand = new String[]{"bash", "-c",
```
gpang: Can we use this when calling the `super` constructor?

core/common/src/main/java/alluxio/shell/ShellCommand.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /**
+   * Runs a command and returns its output and exit code.
+   * No matter it succeeds or not.
+   * Stderr is redirected to stdout.
+   *
+   * @return {@link CommandReturn} object representation of stdout, stderr and exit code
+   */
+  public CommandReturn runTolerateFailure() throws IOException {
```
gpang: I wonder if there is a better name for this, instead of "*TolerateFailure"? maybe "runWithResult", or maybe something else?

core/common/src/main/java/alluxio/shell/SshCommand.java
jiacheliu3: `mCommand` and `mHost` are now changed to `private final`

core/common/src/main/java/alluxio/util/ShellUtils.java
jiacheliu3: The behavior of this `run()` is not changed by this refactor. Only the `Command` class is renamed to `ShellCommand` (and now has a standalone class file) in this line. So there's no concern in the existing uses of this method.

For this `run()` method, stdout and stderr was always merged. And on non-zero exit code, there will be an exception thrown (which loses the output stdout and stderr). And this is exactly why I wanted to refactor and add another method preserving the output on error.

core/common/src/main/java/alluxio/shell/ShellCommand.java
jiacheliu3: Per discussion `runWithOutput` sounds decent enough

core/common/src/main/java/alluxio/util/ShellUtils.java
```diff
@@ -234,13 +173,71 @@ public String toString() {
...
+   * Preserves exit code. Stderr is redirected to stdout.
+   *
+   * @param cmd shell command to execute
+   * @return the output of the executed command
+   * @throws IOException in various situations:
+   *  1. when the executable is not valid, i.e. running ls in Windows
+   *  2. execution interrupted
+   *  3. other normal reasons for IOException
+   */
+  public static CommandReturn execCommandTolerateFailure(String... cmd) throws IOException {
```
gpang: rename these to "*execWithOutput`

jiacheliu3: Thanks for the catch. Turned out I missed a few `TolerateFailure` too.

gpang:

PR #10659, opened 2019-12-19
Add fsadmin metrics clear command

LuQQiu:
Add fsadmin metrics clear command to clear the leading master metrics or the whole cluster metrics.


LuQQiu:
Tests in error: 
alluxio.master.table.transform.TransformManagerTest.jobHistory(alluxio.master.table.transform.TransformManagerTest)
  Run 1: TransformManagerTest.jobHistory:223->heartbeat:328 » Runtime Timer for thread ...
  Run 2: TransformManagerTest.jobHistory:223->heartbeat:328 » Runtime Timer for thread ...

LuQQiu:
jenkins, test this please

LuQQiu:
Tests in error: 
alluxio.client.cli.fs.command.CopyFromLocalCommandIntegrationTest.copyFromLocalLarge(alluxio.client.cli.fs.command.CopyFromLocalCommandIntegrationTest)
  Run 1: CopyFromLocalCommandIntegrationTest.copyFromLocalLarge:175 » Internal http2 ex...
  Run 2: CopyFromLocalCommandIntegrationTest.copyFromLocalLarge:175 » Internal http2 ex...

LuQQiu:
jenkins, test this please

LuQQiu:
@gpang PTAL, thanks!

tests/src/test/java/alluxio/client/cli/fsadmin/metrics/MetricsCommandIntegrationTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+
+/**
+ * Integration tests for the metrics command.
+ */
+public final class MetricsCommandIntegrationTest extends AbstractFsAdminShellTest {
+  @Test
+  public void clearLeadingMasterMetrics() throws IOException {
```
gpang: can there be a test for the worker clear as well?

docs/en/operation/Admin-CLI.md
```diff
@@ -98,6 +98,22 @@ BlockInfo{id=16793993216, length=6, locations=[BlockLocation{workerId=8265394007
...
+`metrics clear` will clear all the metrics stored in Alluxio leading master.
+This command is useful when getting metrics information in short-term testing.
+It should be used sparingly as it may affect the current metrics recording and cause some metrics bias.
+
+If `--all` option is used, all the metrics stored in Alluxio leading master
+and active workers will be cleared.
+
+```console
+$ ./bin/alluxio fsadmin metrics clear
+$ ./bin/alluxio fsadmin metrics clear --all
```
gpang: Is there a scenario we would want to clear all the workers and not the master? Right now, the master is always cleared, and the flag has the ability to additionally clear the workers.

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
```diff
@@ -0,0 +1,123 @@
...
+  public int run(CommandLine cl) throws IOException {
+    System.out.print(String.format("Clearing metrics of Alluxio leading master ..."));
+    mMetricsClient.clearMetrics();
+    System.out.println("CLEARED！");
+    if (cl.hasOption(ALL_OPTION_NAME)) {
+      try (FileSystemContext context = FileSystemContext.create(mAlluxioConf)) {
+        AlluxioBlockStore store = AlluxioBlockStore.create(FileSystemContext.create(mAlluxioConf));
+        List<WorkerNetAddress> addressList = store.getEligibleWorkers().stream()
+            .map(BlockWorkerInfo::getNetAddress).collect(Collectors.toList());
+        for (WorkerNetAddress worker : addressList) {
```
gpang: should we do this in parallel?

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -73,12 +75,15 @@ private static String getFullInstanceId(String hostname, String id) {
     return str;
   }
 
+  // The lock for worker and client metrics set
```
gpang: Can you add a bit more details in this comment? I don't think it is obvious how to use the read and write lock, since the "put*" methods just take a read lock.

tests/src/test/java/alluxio/client/cli/fsadmin/metrics/MetricsCommandIntegrationTest.java
LuQQiu: Yeah, but there wasn't a good way to get and check the worker metrics in the integration tests.
The web UI is not enabled in local alluxio cluster in integration tests, so `<host>:<web_port>/metrics/json` cannot be accessed. The `fsadmin report metrics` doesn't show any metrics related to the worker metrics (doesn't show any aggregated worker metrics) because the current `MetaMasterClient.getMetrics()` only expose a limited set of master metrics.

I plan to merge this PR first, in the next PR, I will modify the current `MetaMasterClient.getMetrics`, modify this method, move it to metricsMaster.
After that, the `fsadmin report metrics` command will also be modified to expose aggregated cluster metrics.

docs/en/operation/Admin-CLI.md
LuQQiu: If a user wants to clear all the worker metrics, he has a large probability to clear the master metrics as well.

In some cases, a user may only want to clear the metrics of a specific worker and gain information about this worker.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
LuQQiu: Added more details! Thanks for pointing out

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
LuQQiu: Change to use executor service for parallism

docs/en/operation/Admin-CLI.md
LuQQiu: Changed to clear the whole cluster by default,
if --master is passed in, clear metrics of leading master only
if --workers is passed in, clear metrics of alluxio workers
if --node <worker_hostnames_separated_by_comma> is passed in, clear metrics of specified workers

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -552,6 +553,7 @@ public static synchronized void resetAllMetrics() {
       METRIC_REGISTRY.remove(timerName);
       METRIC_REGISTRY.timer(timerName);
     }
+    LAST_REPORTED_METRICS.clear();
```
gpang: Was this a bug? Is there a test case for this?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -440,7 +440,8 @@ public static Timer timer(String name) {
    * By sending only the difference since the last RPC, this method will allow the master to
    * track the metrics for a given worker, even if the worker is restarted.
    */
-  private static List<alluxio.grpc.Metric> reportMetrics(InstanceType instanceType) {
+  private static synchronized List<alluxio.grpc.Metric> reportMetrics(InstanceType instanceType) {
```
gpang: Why does this have to be `synchronized`?

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
```diff
@@ -0,0 +1,264 @@
...
+import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+
+/**
+ * Clear the leading master (and workers) metrics.
+ */
+public final class ClearCommand extends AbstractFsAdminCommand {
+  private static final String MASTER_OPTION_NAME = "master";
+  private static final String WORKERS_OPTION_NAME = "workers";
+  private static final String SPECIFIED_OPTION_NAME = "node";
```
gpang: I think this option might be confusing, since it is only workers. Ideally, an option like this would be called `--workers=a,b,c`, but there is already an option for `--workers`. Also, I'm not entirely sure we will need a clear-specific-workers functionality just yet. We could just stick with `--workers` and in the future, see if we can overload `--workers` to also accept a list of individual hostnames.

What do you think?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Yeah, it's a bug, we clears the metrics stored, but forget to clear the last_reported_metrics which may report negative value to the leading master (incorrectness).

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Because when clearing the metrics, `reportMetrics` are likely to be incorrect.

For example, we have a counter with value 4, its previous value which is stored in the LAST_REPORTED_METRICS is 3.

When the counter value is reset to 0, reportMetrics may report 0 - 3 = -3 to the leading master if we clean LAST_REPORTED_METRICS after counters.

If we clean counters after LAST_REPORTED_METRICS, it may occur that LAST_REPORTED_METRICS is 0, but counters are still 4, and delta value 4 - 0 = 4 is reported to the leading master which is also incorrect.

So when clearing the metrics, no `reportMetrics` should be going on. We can also use the same `readwriteLock` logic like `MetricsStore.clear()`: give `resetAllMetrics()` a write lock and give `reportMetrics` a read lock.

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
LuQQiu: Our current option system does not support `--workers a,b,c` and `--workers` at the same time. I also feel the option names are confusing. 

If users curious about all the metrics during a certain period, they may want to clear all the master and workers metrics. I cannot think of a senario that users want to clear all the workers metrics without clearing the leading master metrics. There may be no need for clearing all the workers metrics without clearing the leading master metrics. (no need for --workers flag)

For --master, user may only curious about the cluster metrics, so it makes sense to clear the metrics of leading master only.

I would prefer to keep `--workers a,b,c` and remove the `--workers` flag.
So the new usage would be:
by default, clear the metrics of the whole cluster.
if `--master`, clear the metrics of the leading master
if `--workers a,b,c`, clear the metrics of specific worker.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Added a MetricsSystemTest.testResetAllMetrics() test. If this line is not added, the test will fail.

LuQQiu:
jenkins, test this please

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
```diff
@@ -0,0 +1,259 @@
...
+      FileSystemContext context) throws IOException {
+    int workerNum = workers.size();
+    if (workerNum == 0) {
+      System.out.println("No worker metrics to clear.");
+      return true;
+    } else if (workerNum == 1) {
+      clearWorkerMetrics(workers.get(0), context);
+    } else {
+      List<Future<Void>> futures = new ArrayList<>();
+      int parallelism = Math.min(workerNum, DEFAULT_PARALLELISM);
```
gpang: can we add a `--parallelism` option for this?

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
```diff
@@ -0,0 +1,259 @@
...
+
+  /**
+   * Clears the worker metrics.
+   *
+   * @param worker the worker to clear metrics of
+   * @param context the file system context
+   */
+  private void clearWorkerMetrics(WorkerNetAddress worker,
+      FileSystemContext context) throws IOException {
+    BlockWorkerClient blockWorkerClient = context.acquireBlockWorkerClient(worker);
```
gpang: can we use try-with-resources here?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
gpang: Makes sense. Can you add a short comment in the javadoc of the method saying this is synchronized for correctness with `resetAllMetrics`

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
LuQQiu:
```
        BlockWorkerClient blockWorker = mContext.acquireBlockWorkerClient(worker);
        try {
          blockWorker.asyncCache(request);
          mLastBlockIdCached = blockId;
        } finally {
          mContext.releaseBlockWorkerClient(worker, blockWorker);
        }
```
In the client code, we always use `mContext.releaseBlockWorkerClient` instead of directly close the `blockWorkerClient`. This is for resources reuse and reduce the effort of creating/closing frequently-used resources

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Added, thanks!

docs/en/operation/Admin-CLI.md
```diff
@@ -98,6 +98,24 @@ BlockInfo{id=16793993216, length=6, locations=[BlockLocation{workerId=8265394007
...
+### metrics
+
+The `metrics` command provides operations for Alluxio metrics system.
+
+`metrics clear` will clear the metrics stored in the whole alluxio cluster.
+This command is useful when getting metrics information in short-term testing.
+It should be used sparingly as it may affect the current metrics recording and reporting which may lead to metrics incorrectness 
+and affect worker/client heartbeats with leading master.
+
+If `--master` option is used, all the metrics stored in Alluxio leading master will be cleared.
```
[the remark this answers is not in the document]
LuQQiu: They can be used at the same time, then will clear the master metrics and the specific worker metrics.

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+    // By default, metrics clear clears the metrics of the whole cluster
+    // If --master is passed in, clear metrics of the leading master
+    // If --workers a,b,c is passed in, clear metrics of specific workers
+    if (cl.hasOption(WORKERS_OPTION_NAME)) {
+      String workersValue = cl.getOptionValue(WORKERS_OPTION_NAME);
+      Set<String> workerSet = new HashSet<>(Arrays.asList(workersValue.split(",")));
+
+      try (FileSystemContext context = FileSystemContext.create(mAlluxioConf)) {
+        AlluxioBlockStore store = AlluxioBlockStore.create(FileSystemContext.create(mAlluxioConf));
```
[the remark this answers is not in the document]
LuQQiu: Fixed, thanks

shell/src/main/java/alluxio/cli/fsadmin/metrics/ClearCommand.java
LuQQiu: Added the parallelism option

LuQQiu:
alluxio.collections.LockPoolTest.parallelInsertTest(alluxio.collections.LockPoolTest)
  Run 1: LockPoolTest.parallelInsertTest:117 expected:<16> but was:<15>
  Run 2: LockPoolTest.parallelInsertTest:117 expected:<16> but was:<15>

LuQQiu:
Jenkins, test this please

LuQQiu:
@gpang Thanks for the review! Addressed the comments, PTAL, thanks

gpang:

PR #10669, opened 2019-12-20
Increase auth timeout to reduce flaky tests

gpang:
Occasionally, rpc auth takes longer than usual/expected, especially in a highly concurrent scenario (`ConcurrentFileSystemMasterCreateTest`). Anecdotally, I've seen some take over 4 seconds, as opposed to the usual sub-100ms auth. Therefore, 2s is sometimes too short for tests.


calvinjia:

PR #10679, opened 2019-12-24
Refactor FileSystem Interface

apc999:
Motivation
- Simplify the logic to implement a new FileSystem class. E.g., separating the logic to handle file system instance cache as well as reducing the set of methods to implement. 

Details: 
- Move some convenience methods from `BaseFileSystem` to `FileSystem` Interface as default implementation. So in the future, other implementing class only needs to implement the non-default methods to avoid code duplication.
- Move `FileSystem.Cache` outside the interface as `FileSystemCache` to prevent future confusion and also to make the interface thinner
- Rename `CachingFileSystem` to `MetadataCachingBaseFileSystem` to avoid confusion
- Consolidate the logic of Caching `FileSystem` Instances inside `FileSystemCache` and `InstanceCachingFileSystem`, rather than leaving cache insertion in `FileSystem` and cache removal in implementing classes like `BaseFileSystem`


core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -135,11 +126,8 @@ public synchronized void close() throws IOException {
     // TODO(zac) Determine the behavior when closing the context during operations.
     if (!mClosed) {
       mClosed = true;
-      if (mCachingEnabled) {
```
apc999: put this logic back to `FileSystem` interface so all classes implementing the same interface can benefit from the same logic.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -148,12 +132,6 @@ public boolean isClosed() {
     return mClosed;
   }
 
-  @Override
```
apc999: move this and following to interface as default methods

apc999:
@calvinjia  PTAL

core/client/fs/src/main/java/alluxio/client/file/InstanceCachingFileSystem.java
```diff
@@ -0,0 +1,42 @@
...
+  InstanceCachingFileSystem(FileSystem fs, FileSystemCache cache, FileSystemCache.Key key) {
+    super(fs);
+    mCache = cache;
+    mKey = key;
+  }
+
+  @Override
+  public void close() throws IOException {
+    super.close();
+    // TODO(binfan): is this the expected behavior? shouldn't we keep refcount in cache?
```
calvinjia: Yes, I think we need to keep a ref count, since users will use the filesystem without knowing about the cache, so close will potentially be called multiple times.

core/client/fs/src/main/java/alluxio/client/file/LocalCacheFileSystem.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A FileSystem implementation with a local cache.
+ */
+public class LocalCacheFileSystem extends DelegatingFileSystem {
+
+  /**
+   * @param fs a BaseFileSystem instance to query on cache miss
+   */
+  public LocalCacheFileSystem(BaseFileSystem fs) {
```
calvinjia: Does this need to be `BaseFileSystem`, does `FileSystem` work?

core/client/fs/src/main/java/alluxio/client/file/LocalCacheFileSystem.java
```diff
@@ -0,0 +1,34 @@
...
+    super(fs);
+  }
+
+  @Override
+  public FileInStream openFile(AlluxioURI path, OpenFilePOptions options) {
+    return null;
```
calvinjia: Add TODO here?

core/client/fs/src/main/java/alluxio/client/file/LocalCacheFileSystem.java
apc999: changed to `FileSystem`

core/client/fs/src/main/java/alluxio/client/file/LocalCacheFileSystem.java
apc999: added

core/client/fs/src/main/java/alluxio/client/file/InstanceCachingFileSystem.java
apc999: refcount added

core/client/fs/src/main/java/alluxio/client/file/InstanceCachingFileSystem.java
```diff
@@ -31,12 +34,21 @@
...
+  public void incrementCount() {
+    mRefCount.incrementAndGet();
   }
 
   @Override
   public void close() throws IOException {
-    super.close();
-    // TODO(binfan): is this the expected behavior? shouldn't we keep refcount in cache?
-    mCache.remove(mKey);
+    if (mRefCount.decrementAndGet() == 0) {
```
calvinjia: Is there a race condition where a client calls close and sees refcount 0, but before calling `mCache.remove` another client gets the filesystem?

calvinjia:
@apc999 Thanks for the update, had a question about the locking.

core/client/fs/src/main/java/alluxio/client/file/InstanceCachingFileSystem.java
apc999: fixed by serializing all `FileSystemCache.get` and `InstanceCachingFileSystem.close`

core/client/fs/src/main/java/alluxio/client/file/FileSystemCache.java
```diff
@@ -31,7 +32,7 @@
  */
 @ThreadSafe
 public class FileSystemCache {
-  @GuardedBy("itself")
+  @GuardedBy("this")
```
calvinjia: Make an explicit object to lock on?

calvinjia:
@apc999 thanks for the update, there seems to still be some concurrency issues based on the test results

apc999:
@calvinjia Yes fixing the issue now. But now I'm blocked by this question: what's the expected behavior after this sequence:
1. `fs1 = FileSystem.Factory.get()`
2. `fs2 = FileSystem.Factory.get()`
3. `fs1.close`
4. should `fs2` also be closed (which is our current behavior even before this PR---close fs1 is equivalent of closing fs2), or `fs2` should be __indepedent__ in terms of close status though it shares a delegating fs with `fs1`?

calvinjia:
@apc999 I'm not sure if `isClosed` is a useful API to have, but if we need to keep it, I would prefer choice c, where the bulk of the fs is shared but each instance may have a small amount of additional state to hide the fact that the fs is being cached.

calvinjia:

apc999:
alluxio-bot, cherry-pick this to master please

PR #10682, opened 2019-12-28
Set raft-client policy to connect to leader for journal writes

ggezer:
Default server selection policy will select a random server while connecting to cluster. When a connected server is not the leader, client requests will simply be forwarded to the leader.

We use raft clients for journal writes and writes are only allowed on leader. Forcing client to connect to leader will improve performance by eliminating extra hops and will make transport level traces less confusing for investigation.


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -274,6 +275,7 @@ private CopycatClient createAndConnectClient() {
   private CopycatClient createClient() {
     return CopycatClient.builder(getClusterAddresses(mConf))
         .withRecoveryStrategy(RecoveryStrategies.RECOVER)
+        .withServerSelectionStrategy(ServerSelectionStrategies.LEADER)
```
gpang: can you add a short comment about why we want this? (a short version of the PR description would be enough)

calvinjia:

ggezer:

PR #10691, opened 2020-01-02
Make SleepingUnderFileSystem to make use of managed blocking

ggezer:
Wrapping SleepingUnderFileSystem with a special forwarder to let master thread-pool know that threads will sleep on this UFS.

### DIFF
--- a/tests/src/test/java/alluxio/testutils/underfs/sleeping/SleepingUnderFileSystemFactory.java
+++ b/tests/src/test/java/alluxio/testutils/underfs/sleeping/SleepingUnderFileSystemFactory.java
@@ -12,6 +12,7 @@
 package alluxio.testutils.underfs.sleeping;
 
 import alluxio.AlluxioURI;
+import alluxio.conf.PropertyKey;
 import alluxio.underfs.UnderFileSystem;
 import alluxio.underfs.UnderFileSystemConfiguration;
 import alluxio.underfs.UnderFileSystemFactory;
@@ -46,6 +47,8 @@ public class SleepingUnderFileSystemFactory implements UnderFileSystemFactory {
 
   @Override
   public UnderFileSystem create(String path, UnderFileSystemConfiguration conf) {
+    // Managed blocking should be enabled when using Sleeping UFS.
+    conf.set(PropertyKey.MASTER_UFS_MANAGED_BLOCKING_ENABLED, true);
     if (mUfs == null) {
       Preconditions.checkArgument(path != null, "path may not be null");
       return new SleepingUnderFileSystem(new AlluxioURI(path), mOptions, conf);


ggezer:
Jenkins, test this please

tests/src/test/java/alluxio/testutils/underfs/sleeping/SleepingUnderFileSystemFactory.java
```diff
@@ -41,14 +42,17 @@ public SleepingUnderFileSystemFactory(SleepingUnderFileSystemOptions options) {
...
 
   @Override
   public UnderFileSystem create(String path, UnderFileSystemConfiguration conf) {
     if (mUfs == null) {
       Preconditions.checkArgument(path != null, "path may not be null");
-      return new SleepingUnderFileSystem(new AlluxioURI(path), mOptions, conf);
+      // Wrapping with a special forwarder to let master thread-pool know that threads will sleep
+      // on this UFS.
+      return new ManagedBlockingUfsForwarder(
+          new SleepingUnderFileSystem(new AlluxioURI(path), mOptions, conf));
```
calvinjia: Instead of directly wrapping, should we modify the conf to set managedblocking to true?

ggezer: You mean overriding the conf within `SleepingUnderFileSystemFactory`? Or in all tests that use `SleepingUnderFileSystem` ?

calvinjia: overriding the conf within `SleepingUnderFileSystemFactory` if it is not set

ggezer: I feel uneasy about introducing a special config override.  Could you please explain what do you mean by ufs not being automatically wrapped?

calvinjia: Currently the managed blocking flag is turned on if `isObjectStore` is true or if the managed blocking configuration value is set. I would prefer to use one mechanism to determine if we are using managed blocking.

What I mean is - if somewhere in our code we get a new S3AUnderFileSystem, it will not be wrapped in managed blocking (only if it is created through the UfsManager).

calvinjia: @gpang right, it would be set in the ufsconf, ideally we want to check if it has been set already (not sure if we can distinguish between set and default value though)

ggezer: Understood. PTAL at the change.

calvinjia:

PR #10692, opened 2020-01-03
Parallelize load definition within a worker

bradyoo:
Also removed active_jobs parameter from the command because it's needlessly confusing.

### DIFF
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -717,5 +717,27 @@ public final class CommonUtils {
     };
   }
 
+  /**
+   * Partitions a list into numLists many lists each with around list.size() / numLists elements.
+   *
+   * @param list the list to partition
+   * @param numLists number of lists to return
+   * @param <T> the object type
+   * @return partitioned list
+   */
+  public static <T> List<List<T>> partition(List<T> list, int numLists) {
+    ArrayList<List<T>> result = new ArrayList<>(numLists);
+
+    for (int i = 0; i < numLists; i++) {
+      result.add(new ArrayList<>(list.size() / numLists + 1));
+    }
+
+    for (int i = 0; i < list.size(); i++) {
+      result.get(i % numLists).add(list.get(i));
+    }
+
+    return result;
+  }
+
   private CommonUtils() {} // prevent instantiation
 }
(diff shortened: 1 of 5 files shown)


bradyoo:
@calvinjia

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -717,5 +717,27 @@ private synchronized T firstTime() {
...
+   * @param list the list to partition
+   * @param numLists number of lists to return
+   * @param <T> the object type
+   * @return partitioned list
+   */
+  public static <T> List<List<T>> partition(List<T> list, int numLists) {
+    ArrayList<List<T>> result = new ArrayList<>(numLists);
+
+    for (int i = 0; i < numLists; i++) {
+      result.add(new ArrayList<>(list.size() / numLists + 1));
```
calvinjia: Will this be a problem if `list.size` is smaller than `numLists + 1`?

job/server/src/main/java/alluxio/job/plan/load/LoadDefinition.java
```diff
@@ -54,6 +55,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(LoadDefinition.class);
   private static final int MAX_BUFFER_SIZE = 500 * Constants.MB;
 
+  private static final int JOBS_PER_WORKER = 10;
```
calvinjia: `CONCURRENT_TASKS_PER_WORKER`?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -148,11 +138,6 @@ public String getCommandName() {
     return "distributedLoad";
   }
 
-  @Override
-  public Options getOptions() {
-    return new Options().addOption(REPLICATION_OPTION).addOption(ACTIVE_JOBS_OPTION);
```
calvinjia: Should the replication options be removed?

apc999: please also update `getUsage ()`

job/server/src/main/java/alluxio/job/plan/load/LoadDefinition.java
```diff
@@ -54,6 +55,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(LoadDefinition.class);
   private static final int MAX_BUFFER_SIZE = 500 * Constants.MB;
```
apc999: nit: I would remove the empty line 57. We try to keep the src code concise and compact

core/common/src/main/java/alluxio/util/CommonUtils.java
bradyoo: 1. divide happens before + 1 so it's guaranteed to be at least 1 (as long as numLists is > 0).
2. But more importantly, new ArrayList<>(0) is fine because it's just initial capacity.

bradyoo:
feedback addressed.

apc999:

PR #10695, opened 2020-01-03
Make FileInStream an abstract class

calvinjia:
Changes FileInStream to AlluxioFileInStream and makes FileInStream an abstract class.


apc999:

PR #10698, opened 2020-01-05
Reduce client heartbeat frequency and heartbeat on context close

LuQQiu:
This PR reduces the client heartbeat frequency by enlarging the heartbeat interval and add the initial delay.
To prevent incomplete client metrics for short-life clients, add the last heartbeat when close the last FileSystemContext with the same connect details.


calvinjia:

PR #10700, opened 2020-01-06
Cache getFileSystem for path

bradyoo:
JobServerContext provides a FileSystem to be shared amongst its workers.
However, Hadoop Path has its own API for getting FileSystem which ends up creating too many filesystems that can be shared.

This extends Path to cache values of getFileSystem to prevent opening of too many FileSystems.

As a result, CsvReader now no longer closes the FileSystem

Known Issue (For future fix):

Multiple threads can get into creating the same FileSystem at the same time. The problem with this is that, not only there can be 9 extra copies, because only one actually makes it into the cache, there are 9 that just never get closed.

### DIFF
--- a/job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
+++ b/job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
@@ -1,0 +1,111 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.job.plan.transform.format;
+
+import alluxio.client.ReadType;
+import alluxio.client.WriteType;
+import alluxio.conf.InstancedConfiguration;
+import alluxio.conf.PropertyKey;
+
+import com.google.common.base.Objects;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.hadoop.util.StringUtils;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Implementation of {@link Path} that has a cache for getting file system.
+ */
+public class CachedPath extends Path {
+
+  private static final ConcurrentHashMap<FileSystemKey, FileSystem> CACHE =
+      new ConcurrentHashMap<>();
+
+  private static class FileSystemKey {
+
+    final String mScheme;
+    final String mAuthority;
+    final UserGroupInformation mUgi;
+    final ReadType mReadType;
+    final WriteType mWriteType;
+
+    public FileSystemKey(CachedPath path, Configuration conf) throws IOException {
+      URI uri = path.toUri();
+      mScheme = uri.getScheme() == null ? "" : StringUtils.toLowerCase(uri.getScheme());
+      mAuthority = uri.getAuthority() == null ? "" : StringUtils.toLowerCase(uri.getAuthority());
+      mUgi = UserGroupInformation.getCurrentUser();
+      mReadType = conf.getEnum(PropertyKey.USER_FILE_READ_TYPE_DEFAULT.getName(),
+          InstancedConfiguration.defaults().getEnum(
+              PropertyKey.USER_FILE_READ_TYPE_DEFAULT, ReadType.class));
+      mWriteType = conf.getEnum(PropertyKey.USER_FILE_WRITE_TYPE_DEFAULT.getName(),
+          InstancedConfiguration.defaults().getEnum(
+              PropertyKey.USER_FILE_WRITE_TYPE_DEFAULT, WriteType.class));
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(mScheme, mAuthority, mUgi, mReadType, mWriteType);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (o == null) {
+        return false;
+      }
+      if (this == o) {
+        return true;
+      }
+      if (!(o instanceof FileSystemKey)) {
+        return false;
+      }
+      FileSystemKey that = (FileSystemKey) o;
+      return Objects.equal(mScheme, that.mScheme)
+          && Objects.equal(mAuthority, that.mAuthority)
+          && Objects.equal(mUgi, that.mUgi)
+          && Objects.equal(mReadType, that.mReadType)
+          && Objects.equal(mWriteType, that.mWriteType);
+    }
+  }
+
+  /**
+   * Copy of the constructor in {@link Path}.
+   * @param scheme the scheme
+   * @param authority the authority
+   * @param path the path
+   */
+  public CachedPath(String scheme, String authority, String path) {
+    super(scheme, authority, path);
+  }
+
+  @Override
+  public FileSystem getFileSystem(Configuration conf) throws IOException {
+    try {
+      return CACHE.computeIfAbsent(new FileSystemKey(this, conf), (key) -> {
+        try {
+          return super.getFileSystem(conf);
+        } catch (IOException e) {
+          throw new RuntimeException(e);
+        }
+      });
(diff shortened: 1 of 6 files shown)


bradyoo:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project alluxio-core-base: ExecutionException The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Command was /bin/sh -c cd /usr/src/alluxio/core/base && /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -

bradyoo:

job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.hadoop.security.UserGroupInformation;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Implementation of {@link Path} that has a cache for getting file system.
+ */
+public class CachedPath extends Path {
```
gpang: Is this specific to `transform/format` or should this be more general to any job needing a "Path"?

Also, the name is a bit misleading, since the cache is not caching a path, but the fs client. Maybe rename to something like "JobPath"?

job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private static final ConcurrentHashMap<FileSystemKey, FileSystem> CACHE =
+      new ConcurrentHashMap<>();
+
+  private static class FileSystemKey {
+
+    final String mScheme;
+    final String mAuthority;
+    final UserGroupInformation mUgi;
+    final ReadType mReadType;
```
gpang: Are these the only user config we are interested in? Could there be issues for reusing a client when other user config values differ?

job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
```diff
@@ -0,0 +1,111 @@
...
+    final WriteType mWriteType;
+
+    public FileSystemKey(CachedPath path, Configuration conf) throws IOException {
```
gpang: How does this conf get set/created for the job?

gpang:
@bradyoo could you add a bit more information in the PR description?

job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
bradyoo: There definitely can be new user configs in the future that we may care that we currently don't. It's going to be impossible to have this be always up-to-date. For example, one that I can currently think of Replication.

job/server/src/main/java/alluxio/job/plan/transform/format/CachedPath.java
bradyoo: This is part of the Path.getFileSystem call.

bradyoo:
Feedback commented/addressed

bradyoo:
@gpang: false alarm on this making a big difference in runtime of Compaction. It doesn't seem to make that much of a difference.

gpang:
@bradyoo does this eliminate the noisy log messages about the new client initialization?

bradyoo:
You will get a few but much less than before.

job/server/src/main/java/alluxio/job/plan/transform/format/JobPath.java
```diff
@@ -0,0 +1,111 @@
...
+   * @param scheme the scheme
+   * @param authority the authority
+   * @param path the path
+   */
+  public JobPath(String scheme, String authority, String path) {
+    super(scheme, authority, path);
+  }
+
+  @Override
+  public FileSystem getFileSystem(Configuration conf) throws IOException {
```
gpang: Can we add a unittest for this, to verify the cache behavior?

job/server/src/main/java/alluxio/job/plan/transform/format/JobPath.java
```diff
@@ -0,0 +1,111 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Implementation of {@link Path} that has a cache for getting file system.
+ */
+public class JobPath extends Path {
+
+  private static final ConcurrentHashMap<FileSystemKey, FileSystem> CACHE =
```
gpang: should we use a guava `CacheBuilder` with a maximum size, to limit the unlimited growth?

bradyoo: For now, this is not much of a concern. mScheme, mAuthority, and mUgi are all basically always gonna be constant and even mReadType and mWriteType is very limited.

What's going to get really annoying once we limit size is I don't know which cache is going to need to close the FileSystem on removal.

job/server/src/main/java/alluxio/job/plan/transform/format/JobPath.java
bradyoo: Will add test.

bradyoo:
Feedback addressed

job/server/src/test/java/alluxio/job/plan/transform/format/JobPathTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    when(UserGroupInformation.getCurrentUser()).thenReturn(null);
+  }
+
+  @Test
+  public void testCache() throws Exception {
+    mockStatic(FileSystem.class);
+
+    Configuration conf = new Configuration();
+    JobPath jobPath = new JobPath("foo", "bar", "/baz");
```
gpang: I'm not sure how this test works. Shouldn't there be some verification that the same JobPath scheme and auth and conf get the same client back, and also, when, scheme, and/or auth, and/or conf is different, we get a different client back? We should also test that if a non-inspected conf is changed, it still returns a cached client.

job/server/src/test/java/alluxio/job/plan/transform/format/JobPathTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    FileSystem mockFileSystem = mock(FileSystem.class);
+
+    when(FileSystem.get(eq(jobPath.toUri()), any())).thenReturn(mockFileSystem);
```
gpang: does the test not work unless the fs is mocked?

job/server/src/test/java/alluxio/job/plan/transform/format/JobPathTest.java
bradyoo: The main reason I did it this way is because I had to mock FileSystem so instead of checking what I get back. I instead check the number of calls to the FileSystem.get

bradyoo: We also do test that is a non-inspected conf is changed, it doesn't call the underlying FileSystem.get which is the same thing.

bradyoo:
Improved test.

job/server/src/test/java/alluxio/job/plan/transform/format/JobPathTest.java
gpang: Got it. Can you also add a case where there is a new `JobPath` instance, with the same path (fs should be cached), and a new `JobPath` instance with different path (fs should not be cached)

bradyoo:
Realized I was missing path and also added a test case to the test.

gpang:
```
alluxio.job.plan.transform.format.JobPathTest
testCache(alluxio.job.plan.transform.format.JobPathTest)  Time elapsed: 1.515 sec  <<< FAILURE!
java.lang.AssertionError: Values should be different. Actual: Mock for FileSystem, hashCode: 1056953587
	at alluxio.job.plan.transform.format.JobPathTest.testCache(JobPathTest.java:82)
```

bradyoo:
there was a bug in the code (and not the test). Fixed.

job/server/src/test/java/alluxio/job/plan/transform/format/JobPathTest.java
```diff
@@ -0,0 +1,86 @@
...
+    assertEquals(newFileSystem, jobPath.getFileSystem(conf));
+    verifyStatic(times(2));
+    FileSystem.get(any(), any());
+
+    jobPath = new JobPath("foo", "bar", "/bar");
+    assertEquals(newFileSystem, jobPath.getFileSystem(conf));
+    verifyStatic(times(2));
+    FileSystem.get(any(), any());
+
+    jobPath = new JobPath("foo", "baz", "/bar");
```
gpang: can we add one more case with `new JobPath("foo2", "bar", "/baz");`

bradyoo:
Done.

bradyoo:
Mind merging this for me?

gpang:

PR #10705, opened 2020-01-07
Cache list status results

chcn:
Note that after caching the list status results, if there are files added to or deleted from the directory, then these updates might not be reflected until the cache is invalidated.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/CachingFileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/CachingFileSystem.java
@@ -27,7 +27,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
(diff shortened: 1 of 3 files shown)


core/client/fs/src/main/java/alluxio/client/file/CachingFileSystem.java
```diff
@@ -87,9 +87,10 @@ public URIStatus getStatus(AlluxioURI path, GetStatusPOptions options)
   public List<URIStatus> listStatus(AlluxioURI path, ListStatusPOptions options)
       throws FileDoesNotExistException, IOException, AlluxioException {
     checkUri(path);
-    List<URIStatus> statuses = super.listStatus(path, options);
-    for (URIStatus status : statuses) {
-      mMetadataCache.put(status.getPath(), status);
+    List<URIStatus> statuses = mMetadataCache.listStatus(path);
```
apc999: as discussed offline, let us selectively cache `listStatus` result only on non-recursive queries.
@chcn please also comment the pros/cons here for future references.

chcn: @apc999 updated, PTAL

apc999:

apc999:

Issue #10693, opened 2020-01-03
Add listening address for embedded journal


calvinjia:
**Is your feature request related to a problem? Please describe.**
Fine control over the embedded journal service's listening address is not available.

**Describe the solution you'd like**
Add a property to specify the listening address of the journal service, similar to how the master has a listening address for the rpc service.

**Additional context**
This allows users to more easily deploy Alluxio in K8s environments secured with istio.

ggezer:
Do copycat clients supposed to connect to listen port then? This means the membership address list config property should list host and listen addresses separately for each.

PR #10706, opened 2020-01-07
Add listening address for embedded journal

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/10693

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -240,9 +240,8 @@ public final class RaftJournalSystem extends AbstractJournalSystem {
     }
     mStateMachine = new JournalStateMachine(mJournals, () -> this.getJournalSinks(null));
     // Read external proxy configuration.
-    GrpcMessagingProxy serverProxy = null;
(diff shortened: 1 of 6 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -238,14 +239,21 @@ private synchronized void initServer() {
       mStateMachine.close();
     }
     mStateMachine = new JournalStateMachine(mJournals, () -> this.getJournalSinks(null));
+    // Read external proxy configuration.
+    GrpcMessagingProxy serverProxy = null;
+    if (mConf.getProxyAddress() != null) {
+      serverProxy = new GrpcMessagingProxy();
+      serverProxy.addProxy(getLocalAddress(mConf), new Address(mConf.getProxyAddress()));
+    }
```
calvinjia: Consider creating the `GrpcMessagingTransport` here? Then we won't have to do as many null checks here and in `GrpcMessagingTransport` since we can assume the proxy is only set if necessary.

ggezer: I was able to strip some null checks by enforcing valid proxy reference.

calvinjia:

ggezer:

PR #10709, opened 2020-01-08
Fix backup restore timer metrics

LuQQiu:
Previous we record the time of last backup/restore using Timer metrics.
Timer metrics has some problem of recording operation durations. 
In local backup restore tests, timer metrics cannot record the restore time correctly.
This PR change to use Gauge metrics.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/BackupManager.java
+++ b/core/server/common/src/main/java/alluxio/master/BackupManager.java
@@ -21,7 +21,6 @@ import alluxio.metrics.MetricsSystem;
 import alluxio.proto.journal.Journal.JournalEntry;
 import alluxio.util.ThreadFactoryUtils;
 
-import com.codahale.metrics.Timer;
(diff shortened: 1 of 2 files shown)


LuQQiu:
@calvinjia  PTAL, thanks!

core/server/common/src/main/java/alluxio/master/BackupManager.java
```diff
@@ -76,6 +75,9 @@
   private long mBackupEntriesCount;
   private long mRestoreEntriesCount;
 
+  private long mBackupTimeMs;
+  private long mRestoreTimeMs;
```
calvinjia: Should we set them to `-1` to indicate there was no backup or restore completed?

LuQQiu: Good idea! Set it to -1

calvinjia:
alluxio-bot, merge this please?

PR #10714, opened 2020-01-09
Release file and block client resources during interrupted exception

ZacBlanco:
This contains two separate, but connected commits that alleviate the leaking of resources on the client side. Mainly the behavior is seen on long-running clients (i.e. Presto) where queries may be killed and the client threads experience `InterruptedException`s

----

The first commit contains a fix that was actually committed ~1 year ago to `branch-1.8` which did not make it into master. It converted the `ResourcePool` underneath the `BlockMasterClientPool` and `FileSystemMasterClientPool` into a `DynamicResourcePool` which periodically GCs clients and closes them to reduce resource utilization among other things. The original author probably has the most context on why this change was made.

----

The second change deals with an issue in the `FileSystemContext#blockReinit`. This method prevents operations on the FileSystemContext from continuing while a reinitialization is in progress. The issue around leaking clients mainly comes from when the reinit is blocked and an interrupt exception is thrown. Originally the logic supposed that if the thread is interrupted that we should simply throw a runtime exception and not eventually release the resource because of the chance of partial re-initialization of the FSContext.

The fix changes the strategy as best effort to release the resource back to its respective pool even after the interrupted exception is thrown on `blockReinit`. This ensures that no resources are leaked on thread interrupts.

The solution is up for discussion as it seems to fix the issue, but I feel there may be better ways to address it.

----

There is an example test added in `FileInStreamIntegrationTest` at the bottom, which, before the final commit on this branch is able to reproduce the issue with hanging block master clients.


gpang:
```
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:36:8: error: Unused import - org.apache.log4j.Level.
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:37:8: error: Unused import - org.apache.log4j.LogManager.
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:38:8: error: Unused import - org.apache.log4j.Logger.
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:46:8: error: Unused import - org.slf4j.LoggerFactory.
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:57:8: error: Unused import - java.util.function.Supplier.
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:613:39: error: '(' is followed by whitespace.
/usr/src/alluxio/tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java:616: error: '}' have incorrect indentation level 2, expected level should be 6.
```

ZacBlanco:
This PR was more of a PoC. It is superseded by https://github.com/Alluxio/alluxio/pull/10719

Issue #10468, opened 2019-11-13
Presto queries get stuck infinitely while accessing files from alluxio.


apc999:
**Alluxio Version:**
Alluxio v2.1.0

**Describe the bug**
Originally reported from https://alluxio-community.slack.com/archives/CEXGGUBDK/p1573638510393600

we are facing some weird issue while accessing data from Alluxio using presto. Presto queries are getting stuck infinitely while accessing files from alluxio. I checked the thread dumps for presto worker and seems that threads are getting locked while accessing alluxio file. We were facing this issue with Alluxio 2.0.0, we were told that few improvements has done on client side in the latest version so we upgraded to latest version 2.1.0, but still we are getting the same issue. Could someone please help us to resolve this issue? I am adding alluxio worker logs  and presto thread dumps.
Note - This deployment has done using kubernetes on AWS

**To Reproduce**

**Expected behavior**

**Urgency**
High

**Additional context**
Thread dump
[thread_dumps.log](https://github.com/Alluxio/alluxio/files/3842826/thread_dumps.log)
Worker log: 
[alluxio_worker_log.txt](https://github.com/Alluxio/alluxio/files/3842827/alluxio_worker_log.txt)

surajc-cuelogic:
@apc999,
As per discussion on slack I tried by setting **alluxio.user.file.master.client.threads=128** this property but I still getting the same issue. I have captured the thread dumps and attaching here. I hope that will help to debug the issue.
[presto-worker-thread-dump-18-11.txt](https://github.com/Alluxio/alluxio/files/3857872/presto-worker-thread-dump-18-11.txt)

surajc-cuelogic:
Attaching files to understand the alluxio configurations we have on production

[alluxio-env-master.txt](https://github.com/Alluxio/alluxio/files/3858519/alluxio-env-master.txt)
[alluxio-env-worker.txt](https://github.com/Alluxio/alluxio/files/3858520/alluxio-env-worker.txt)

apc999:
@bf8086 can you help take a look?

bf8086:
@surajc-cuelogic thanks for reporting the issue. Do you still have the Alluxio worker log for node `soc-alluxio-worker-5.soc-alluxio-worker` around timestamp `2019-11-13 07:54`?

prashantk-cuelogic:
[...]

PR #10719, opened 2020-01-09
Release clients back into pools during even during context reinit

### LINKED ISSUE #9922: Client hang for Presto query
**Alluxio Version:**
2.0.1

**Describe the bug**

We repeatedly run a query and find it hang when running for several days. 

The query is 
```sql
select count(1) from a table. 
```

This table we already load in ahead from hdfs to alluxio 
During the running period, some alluxio workers may be down and some lost data may need to read directly from hdfs.

From `jstack`, find 30 threads (as we set presto concurrency to 30) stuck at 
```
"20190919_081107_00969_nmmwc.1.7-19-4862" #4862 prio=5 os_prio=0 tid=0x00007f9618009dd0 nid=0xa86a2 waiting on condition [0x00007f925c2f1000]
   java.lang.Thread.State: WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00007f9bf8c94008> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
        at alluxio.resource.ResourcePool.acquire(ResourcePool.java:132)
        at alluxio.resource.ResourcePool.acquire(ResourcePool.java:77)
        at alluxio.client.file.FileSystemContext.acquireBlockMasterClient(FileSystemContext.java:421)
        at alluxio.client.file.FileSystemContext.acquireBlockMasterClientResource(FileSystemContext.java:441)
        at alluxio.client.block.AlluxioBlockStore.getInStream(AlluxioBlockStore.java:169)
        at alluxio.client.file.FileInStream.positionedReadInternal(FileInStream.java:252)
        at alluxio.client.file.FileInStream.positionedRead(FileInStream.java:232)
        at alluxio.hadoop.HdfsFileInputStream.read(HdfsFileInputStream.java:128)
        at alluxio.hadoop.HdfsFileInputStream.readFully(HdfsFileInputStream.java:145)
        at org.apache.hadoop.fs.FSDataInputStream.readFully(FSDataInputStream.java:111)
        at org.apache.hadoop.fs.FSDataInputStream.readFully(FSDataInputStream.java:120)
        at io.prestosql.parquet.reader.MetadataReader.readFully(MetadataReader.java:313)
        at io.prestosql.parquet.reader.MetadataReader.readFooter(MetadataReader.java:92)
        at io.prestosql.plugin.hive.parquet.ParquetPageSourceFactory.createParquetPageSource(ParquetPageSourceFactory.java:161)
        at io.prestosql.plugin.hive.parquet.ParquetPageSourceFactory.createPageSource(ParquetPageSourceFactory.java:120)
        at io.prestosql.plugin.hive.HivePageSourceProvider.createHivePageSource(HivePageSourceProvider.java:162)
        at io.prestosql.plugin.hive.HivePageSourceProvider.createPageSource(HivePageSourceProvider.java:96)
        at io.prestosql.spi.connector.classloader.ClassLoaderSafeConnectorPageSourceProvider.createPageSource(ClassLoaderSafeConnectorPageSourceProvider.java:53)
        at io.prestosql.split.PageSourceManager.createPageSource(PageSourceManager.java:56)
        at io.prestosql.operator.TableScanOperator.getOutput(TableScanOperator.java:271)
        at io.prestosql.operator.Driver.processInternal(Driver.java:379)
        at io.prestosql.operator.Driver.lambda$processFor$8(Driver.java:283)
        at io.prestosql.operator.Driver$$Lambda$2810/201986911.get(Unknown Source)
        at io.prestosql.operator.Driver.tryWithLock(Driver.java:675)
        at io.prestosql.operator.Driver.processFor(Driver.java:276)
        at io.prestosql.execution.SqlTaskExecution$DriverSplitRunner.processFor(SqlTaskExecution.java:1075)
        at io.prestosql.execution.executor.PrioritizedSplitRunner.process(PrioritizedSplitRunner.java:163)
        at io.prestosql.execution.executor.TaskExecutor$TaskRunner.run(TaskExecutor.java:484)
        at io.prestosql.$gen.Presto_shopee_101_dirty__shopee102____20190918_090137_1.run(Unknown Source)
```

**To Reproduce**
Currently, not clear. 

**Expected behavior**
No hanging

**Urgency**
High

**Additional context**

ZacBlanco:
This removes a resource leak which may occur when a thread on a client is interrupted while performing any kind of operation.

Because it's possible for a fs context re-initialization to occur while the resource is
acquired this solution removes the `blockReinit` which throws a runtime exception during
an interrupt and is the root cause of the resource leak.

There are three different cases to which may occur during the release of the resource

1. Release while the context is re-initializing
   - The original mBlockMasterClientPool or mFileSystemMasterClientPool
   may be null, closed, or overwritten with a difference pool. The inner class here saves
   the original pool reference used to acquire the client initially. Releasing into the closed pool
   is harmless.
2. Release after the context has been re-initialized
   - Similar to the above scenario the original mBlockMasterClientPool or
   mFileSystemMasterClientPool are going to be using an entirely new pool. Since
   this method will save the original pool reference, this method would result in releasing
   into a closed pool which is harmless
3. Release before any re-initialization
   - This is the normal case. There are no special considerations

The issues mainly affected the FSMaster and BlockMaster clients, however upon inspection, BlockWorker clients have a similar issue, but seems to happen less frequently than the master clients. This PR also applies a similar strategy for BlockWorker clients to hold onto the references of the pool used during the acquire method to make a best-effort to release the resource when finished being used. 

Closes #9922 #10468

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -29,6 +29,8 @@ import alluxio.grpc.GrpcServerAddress;
 import alluxio.master.MasterClientContext;
 import alluxio.master.MasterInquireClient;
 import alluxio.resource.CloseableResource;
+import alluxio.resource.DynamicResourcePool;
+import alluxio.resource.ResourcePool;
 import alluxio.security.authentication.AuthenticationUserUtils;
 import alluxio.util.IdUtils;
 import alluxio.util.network.NettyUtils;
@@ -424,31 +426,7 @@ public final class FileSystemContext implements Closeable {
    * @return the acquired file system master client resource
    */
   public CloseableResource<FileSystemMasterClient> acquireMasterClientResource() {
-    try {
-      return new CloseableResource<FileSystemMasterClient>(mFileSystemMasterClientPool.acquire()) {
-        @Override
-        public void close() {
-          releaseMasterClient(get());
-        }
-      };
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-  }
-
-  private BlockMasterClient acquireBlockMasterClient() throws IOException {
-    try (ReinitBlockerResource r = blockReinit()) {
-      return mBlockMasterClientPool.acquire();
-    }
-  }
-
-  private void releaseBlockMasterClient(BlockMasterClient client) {
-    try (ReinitBlockerResource r = blockReinit()) {
-      if (!client.isClosed()) {
-        // The client might have been closed during reinitialization.
-        mBlockMasterClientPool.release(client);
-      }
-    }
+    return acquireClosableClientResource(mFileSystemMasterClientPool);
   }
 
   /**
@@ -458,11 +436,42 @@ public final class FileSystemContext implements Closeable {
    * @return the acquired block master client resource
    */
   public CloseableResource<BlockMasterClient> acquireBlockMasterClientResource() {
-    try {
-      return new CloseableResource<BlockMasterClient>(mBlockMasterClientPool.acquire()) {
+    return acquireClosableClientResource(mBlockMasterClientPool);
+  }
+
+  /**
+   * Acquire a client resource from {@link #mBlockMasterClientPool} or
+   * {@link #mFileSystemMasterClientPool}.
+   *
+   * Because it's possible for a context re-initialization to occur while the resource is
+   * acquired this method uses an inline class which will save the reference to the pool used to
+   * acquire the resource.
+   *
+   * There are three different cases to which may occur during the release of the resource
+   *
+   * 1. release while the context is re-initializing
+   *    - The original {@link #mBlockMasterClientPool} or {@link #mFileSystemMasterClientPool}
+   *    may be null, closed, or overwritten with a difference pool. The inner class here saves
+   *    the original pool from being GCed because it holds a reference to the pool that was used
+   *    to acquire the client initially. Releasing into the closed pool is harmless.
+   * 2. release after the context has been re-initialized
+   *    - Similar to the above scenario the original {@link #mBlockMasterClientPool} or
+   *    {@link #mFileSystemMasterClientPool} are going to be using an entirely new pool. Since
+   *    this method will save the original pool reference, this method would result in releasing
+   *    into a closed pool which is harmless
+   * 3. release before any re-initialization
+   *    - This is the normal case. There are no special considerations
+   *
+   * @param pool the pool to acquire from and release to
+   * @param <T> the resource type
+   * @return a {@link CloseableResource}
+   */
+  private <T> CloseableResource<T> acquireClosableClientResource(DynamicResourcePool<T> pool) {
+    try (ReinitBlockerResource r = blockReinit()) {
+      return new CloseableResource<T>(pool.acquire()) {
         @Override
         public void close() {
-          releaseBlockMasterClient(get());
+          pool.release(get());
         }
       };
     } catch (IOException e) {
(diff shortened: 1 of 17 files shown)


chcn:
@ZacBlanco do you have a test case to show what the issue is and this PR fixes the issue?

ZacBlanco:
@chcn There is a test here which I used to reproduce the issue: https://github.com/Alluxio/alluxio/pull/10714/files#diff-f570d0a32cd800fc55939b5f4175b3d8R610-R685

I haven't added it to the PR because it takes a bit of time to run locally, but essentially the test will hang indefinitely due to the leaked resources.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -471,25 +471,33 @@ public void close() {
    * @param workerNetAddress the network address of the channel
    * @return the acquired block worker
```
gpang: update comment about the resource

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -471,25 +471,33 @@ public void close() {
...
-        .acquire();
+    ClientPoolKey key = new ClientPoolKey(address, AuthenticationUserUtils
+            .getImpersonationUser(context.getSubject(), context.getClusterConf()));
+    return new CloseableResource<BlockWorkerClient>(mBlockWorkerClientPool.computeIfAbsent(key,
+        k -> new BlockWorkerClientPool(context.getSubject(), serverAddress,
+            context.getClusterConf().getInt(PropertyKey.USER_BLOCK_WORKER_CLIENT_POOL_SIZE),
+            context.getClusterConf(), mWorkerGroup))
+        .acquire()) {
+      // Save the reference to the original pool map.
+      ConcurrentHashMap<ClientPoolKey, BlockWorkerClientPool> mPoolMap = mBlockWorkerClientPool;
```
gpang: Hrmmm, can we avoid explicitly defining it, and just define a local final var for it? It will get captured like that right?

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: done

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: done

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -476,27 +473,36 @@ public void close() {
...
-    return mBlockWorkerClientPool.computeIfAbsent(key,
-        k -> new BlockWorkerClientPool(subject, serverAddress,
-            getClusterConf().getInt(PropertyKey.USER_BLOCK_WORKER_CLIENT_POOL_SIZE),
-            getClusterConf(), mWorkerGroup))
-        .acquire();
+    ClientPoolKey key = new ClientPoolKey(address, AuthenticationUserUtils
+            .getImpersonationUser(context.getSubject(), context.getClusterConf()));
+    final ConcurrentHashMap<ClientPoolKey, BlockWorkerClientPool> poolMap =
+        mBlockWorkerClientPool;
+    return new CloseableResource<BlockWorkerClient>(mBlockWorkerClientPool.computeIfAbsent(key,
```
gpang:
```suggestion
    return new CloseableResource<BlockWorkerClient>(poolMap.computeIfAbsent(key,
```

gpang:
@ggezer could you take a look at the block worker client changes?

ZacBlanco:

ZacBlanco:

PR #10726, opened 2020-01-10
Fix fs memory leak when concurrently getting fs in JobPath

bradyoo:
This code depends on https://github.com/Alluxio/alluxio/pull/10700

### DIFF
--- a/job/server/src/main/java/alluxio/job/plan/transform/format/JobPath.java
+++ b/job/server/src/main/java/alluxio/job/plan/transform/format/JobPath.java
@@ -31,6 +31,9 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public class JobPath extends Path {
 
+  private static final String ALLUXIO_HADOOP_FILESYSTEM_DISABLE_CACHE =
+      "fs.alluxio.impl.disable.cache";
+
   private static final ConcurrentHashMap<FileSystemKey, FileSystem> CACHE =
       new ConcurrentHashMap<>();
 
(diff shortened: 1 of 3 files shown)


bradyoo:
@gpang

bradyoo:
rebased

job/server/src/main/java/alluxio/job/plan/transform/format/ReadWriterUtils.java
```diff
@@ -51,7 +49,6 @@ public static Configuration readNoCacheConf() {
     Configuration conf = new Configuration();
     conf.setEnum(PropertyKey.USER_FILE_READ_TYPE_DEFAULT.getName(), ReadType.NO_CACHE);
     // The cached filesystem might not be configured with the above read type.
```
gpang: should this comment be removed?

job/server/src/main/java/alluxio/job/plan/transform/format/JobPath.java
```diff
@@ -93,20 +96,28 @@ public JobPath(String scheme, String authority, String path) {
...
+
+    FileSystemKey key = new FileSystemKey(this, conf);
+    FileSystem fileSystem = CACHE.get(key);
+    if (fileSystem != null) {
+      return fileSystem;
+    }
+
+    fileSystem = super.getFileSystem(conf);
+
+    synchronized (CACHE) {
```
gpang: I don't think you will need `synchronized`. Instead there should be a `putIfAbsent` which may be applicable, or a `compute` method.

PR #10728, opened 2020-01-10
Keep worker client open after creating data reader/writer

ZacBlanco:
This fix keeps the BlockWorkerClient open after returning the GrpcDataWriter.
Previously, we used a try-with-resources which would close the client before
being returned to the user. This could result in poor behavior if the writer
is used after the #create() call.

Also, to prevent a leak in the LocalFileDataReader, we attempt to close the
reader within the try block inside of its close method.

This is a follow-up to #10719


ZacBlanco:
alluxio-bot, test this please

ZacBlanco:
@bf8086 added a unit test

ZacBlanco:

ZacBlanco:
alluxio-bot, cherry-pick this into branch-2.1 please

ZacBlanco:
alluxio-bot, cherry-pick this into branch-2.1 please

PR #10732, opened 2020-01-13
Cleanup LocalCacheManager  Unit Test

apc999:
Improve unit test of `LocalCacheManager` by
- removing PowerMock
- instead of asserting on internal implementation (counting func calls), assert on cache state before and after.


calvinjia:

PR #10733, opened 2020-01-13
Implement client side cache evictor and metastore

bf8086:
Implement a basic metastore and LRU evictor for client side cache.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
@@ -28,7 +28,7 @@ public interface CacheEvictor {
    */
   static CacheEvictor create(AlluxioConfiguration conf) {
     return CommonUtils.createNewClassInstance(
-        conf.getClass(PropertyKey.USER_CLIENT_CACHE_EVICTOR_CLASS), new Class[0], new Object[0]);
+        conf.getClass(PropertyKey.USER_CLIENT_CACHE_EVICTOR_CLASS), null, null);
   }
 
   /**
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
@@ -21,7 +21,7 @@ public interface CacheEvictor {
    */
   static CacheEvictor create() {
     // return corresponding CacheEvictor impl
-    return null;
+    return new LRUCacheEvictor();
   }
 
   /**
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
@@ -11,6 +11,10 @@
 
 package alluxio.client.file.cache;
 
+import alluxio.conf.AlluxioConfiguration;
+import alluxio.conf.PropertyKey;
+import alluxio.util.CommonUtils;
+
 import javax.annotation.Nullable;
 
 /**
@@ -19,11 +23,12 @@ import javax.annotation.Nullable;
 public interface CacheEvictor {
 
   /**
+   * @param conf the alluxio configuration
    * @return a CacheEvictor instance
    */
-  static CacheEvictor create() {
-    // return corresponding CacheEvictor impl
-    return new LRUCacheEvictor();
+  static CacheEvictor create(AlluxioConfiguration conf) {
+    return CommonUtils.createNewClassInstance(
+        conf.getClass(PropertyKey.USER_CLIENT_CACHE_EVICTOR_CLASS), new Class[0], new Object[0]);
   }
 
   /**
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
@@ -13,43 +13,29 @@ package alluxio.client.file.cache;
 
 import alluxio.exception.PageNotFoundException;
 
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * The default implementation of a metadata store for pages stored in cache.
  */
 public class DefaultMetaStore implements MetaStore {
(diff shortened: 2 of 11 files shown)


bf8086:
@calvin @apc999 PTAL.

calvinjia:
@bf8086 overall LGTM, please resolve the merge conflicts

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
```diff
@@ -13,43 +13,29 @@
...
 import alluxio.exception.PageNotFoundException;
 
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * The default implementation of a metadata store for pages stored in cache.
  */
 public class DefaultMetaStore implements MetaStore {
+  private final Set<PageId> mPageMap = new HashSet<>();
```
calvinjia: Is this implementation actually exactly the same as the `HashSetMetaStore` in the test? If so maybe we only need 1 of these?

apc999: looks like so. Then we should dedup and keep one impl `DefaultMetaStore`.

bf8086: We can reuse the `DefaultMetaStore` in tests for now, but it is likely we will have to reintroduce the test implementation as the `DefaultMetaStore` become more complex.

core/client/fs/src/main/java/alluxio/client/file/cache/LRUCacheEvictor.java
```diff
@@ -0,0 +1,51 @@
...
+package alluxio.client.file.cache;
+
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * LRU client-side cache eviction policy.
+ */
+public class LRUCacheEvictor implements CacheEvictor {
```
apc999: is this class thread safe? If yes, please annotate it with `@ThreadSafe`; if not, currently `LocalCacheManager` does not provide any guard to protect concurrent access to this evictor from multiple threads.

core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
```diff
@@ -23,7 +23,7 @@
    */
   static CacheEvictor create() {
     // return corresponding CacheEvictor impl
-    return null;
+    return new LRUCacheEvictor();
```
apc999: I would suggest to add user-facing propertyKey like `alluxio.user.client.cache.evictor` and allow to select different evictors (thought currently we have only one). This will make it easier to switch cache replacement policy without recompiling the src code

core/client/fs/src/test/java/alluxio/client/file/cache/LRUCacheEvictorTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Sets up the instances.
+   */
+  @Before
+  public void before() {
+    mEvictor = new LRUCacheEvictor();
+  }
+
+  @Test
```
apc999: Shall we cover the following cases:
- `evict` before any other operations (just make sure we don't see NPE)
- `evict` after all pages are put and then deleted

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
calvinjia: if that's the case we can keep them separate.

core/client/fs/src/main/java/alluxio/client/file/cache/LRUCacheEvictor.java
bf8086: It is. Updated with annotation.

core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
bf8086: Added.

core/client/fs/src/test/java/alluxio/client/file/cache/LRUCacheEvictorTest.java
bf8086: Added.

core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
```diff
@@ -19,11 +23,12 @@
...
   /**
+   * @param conf the alluxio configuration
    * @return a CacheEvictor instance
    */
-  static CacheEvictor create() {
-    // return corresponding CacheEvictor impl
-    return new LRUCacheEvictor();
+  static CacheEvictor create(AlluxioConfiguration conf) {
+    return CommonUtils.createNewClassInstance(
+        conf.getClass(PropertyKey.USER_CLIENT_CACHE_EVICTOR_CLASS), new Class[0], new Object[0]);
```
calvinjia: I think you want `null` instead of empty arrays?

core/client/fs/src/main/java/alluxio/client/file/cache/LRUCacheEvictor.java
```diff
@@ -0,0 +1,56 @@
...
+    mLRUCache.put(pageId, UNUSED_MAP_VALUE);
+  }
+
+  @Override
+  public void updateOnDelete(PageId pageId) {
+    mLRUCache.remove(pageId, UNUSED_MAP_VALUE);
+  }
+
+  @Override
+  public PageId evict() {
```
apc999: can you mark this method and the interface def with `@Nullable`

bf8086: Done.

core/client/fs/src/main/java/alluxio/client/file/cache/CacheEvictor.java
bf8086: Updated.

apc999:

PR #10734, opened 2020-01-13
Implement basic version of getDatabase in catalog service

yuzhu:
Note here `getDatabase` returns an object that contains only the database name.
A subsequent PR will add more content into this database object and the associated Udb object to populate this database object.

### DIFF
--- a/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
+++ b/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
@@ -200,7 +200,7 @@ public class AlluxioCatalog implements Journaled {
    * Get Database by its name.
    *
    * @param dbName database name
-   * @return a databae object
+   * @return a database object
    */
   public alluxio.grpc.table.Database getDatabase(String dbName)  throws IOException {
     try (LockResource l = getLock(dbName, true)) {
--- a/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
+++ b/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
@@ -196,6 +196,13 @@ public class AlluxioCatalog implements Journaled {
     return new ArrayList<>(mDBs.keySet());
   }
 
+  public alluxio.grpc.table.Database getDatabase(String dbName)  throws IOException {
+    try (LockResource l = getLock(dbName, true)) {
+      Database db = getDatabaseByName(dbName);
+      return alluxio.grpc.table.Database.newBuilder().setDbName(db.getName()).build();
+    }
+  }
+
   private Database getDatabaseByName(String dbName) throws NotFoundException {
     Database db = mDBs.get(dbName);
     if (db == null) {
--- a/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
+++ b/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
@@ -196,6 +196,12 @@ public class AlluxioCatalog implements Journaled {
     return new ArrayList<>(mDBs.keySet());
   }
 
+  /**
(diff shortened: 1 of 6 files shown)


yuzhu:
@ZacBlanco PTAL. I implemented bare minimum here, but we probably want to have a more complete database object and Udb object with information about the database owner and params. That's for another PR.

table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java
```diff
@@ -196,6 +196,19 @@ private Table getTableInternal(String dbName, String tableName) throws IOExcepti
     return new ArrayList<>(mDBs.keySet());
   }
 
+  /**
+   * Get Database by its name.
+   *
+   * @param dbName database name
+   * @return a databae object
```
ZacBlanco: nit: typo

ZacBlanco:
Does our current database object only contain the name? I thought I saw some fields for description and location before

yuzhu:
> Does our current database object only contain the name? I thought I saw some fields for description and location before

The RPC proto contains description and location, but we don't store that information in our server side representation or our Udb object from Hive yet. Hopefully this is enough to get past the test for now. Changing Udb requires changes to journal as well,  so that's why I want to make that a separate PR.

ZacBlanco:
Also can we add some kind of simple UT for the the `getDatabase` call?

tests/src/test/java/alluxio/server/ft/journal/TableMasterJournalIntegrationTest.java
```diff
@@ -92,8 +96,12 @@ public void journalAttachDb() throws Exception {
     TableMaster tableMaster =
         mCluster.getLocalAlluxioMaster().getMasterProcess().getMaster(TableMaster.class);
 
+    mException.expect(IOException.class);
+    tableMaster.getDatabase(DB_NAME);
```
ZacBlanco: won't this stop the rest of the test from being run?

table/server/master/src/test/java/alluxio/master/table/AlluxioCatalogTest.java
```diff
@@ -100,6 +100,18 @@ public void detachDb() throws Exception {
     assertEquals(0, mCatalog.getAllDatabases().size());
   }
 
+  @Test
+  public void getDb() throws Exception {
+    String dbName = "testdb";
+    TestDatabase.genTable(1, 2);
+    mException.expect(NotFoundException.class);
+    mCatalog.getDatabase(dbName);
```
ZacBlanco: similarly, won't all lines after this not be executed?

tests/src/test/java/alluxio/server/ft/journal/TableMasterJournalIntegrationTest.java
yuzhu: doh.. why aren't we on Junit 5 yet?  I totally misunderstood the semantics of `expect`

table/server/master/src/main/java/alluxio/master/table/TableMaster.java
```diff
@@ -138,4 +139,12 @@ boolean detachDatabase(String dbName)
...
    */
   boolean syncDatabase(String dbName) throws IOException;
+
+  /**
+   * Gets a database object.
+   *
+   * @param dbName the database name
+   * @return a database object
+   */
+  Database getDatabase(String dbName) throws IOException;
```
ZacBlanco: i don't care that much but for the most part methods are in alphabetical order in this class. Should we try to keep it that way? the transform methods are obviously not in the right order, but the rest seem to be

yuzhu:

yuzhu:

PR #10737, opened 2020-01-14
Implement positioned read

calvinjia:
This PR also uses the newer API to do partial page reads correctly.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
@@ -249,7 +249,11 @@ public class LocalCacheFileInStream extends FileInStream {
    * Reads a page from external storage which contains the position specified. Note that this makes
    * a copy of the page.
    *
-   * This method is synchronized to ensure thread safety for positioned reads.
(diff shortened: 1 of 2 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -216,12 +249,14 @@ private FileInStream getExternalFileInStream(long pos) throws IOException {
    * Reads a page from external storage which contains the position specified. Note that this makes
    * a copy of the page.
    *
+   * This method is synchronized to ensure thread safety for positioned reads.
```
apc999: not clear to me what are we guarding. in addition, do we only provide thread-safety for positioned reads not normal reads?

calvinjia: Yeah, only for positioned reads (PositionedReadable). The read API (InputStream) is not thread safe.

calvinjia: I explained a bit more in the comment.

apc999:

PR #10739, opened 2020-01-14
Specify the Alluxio file system type with a class instead of a boolean

calvinjia:
This makes it cleaner to augment the filesystem with different functionality like metadata caching.

Also ran confDocGen so some unrelated docs are updated.


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3368,9 +3368,9 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
-  public static final PropertyKey USER_METADATA_CACHE_ENABLED =
-      new Builder(Name.USER_METADATA_CACHE_ENABLED)
-          .setDefaultValue(false)
+  public static final PropertyKey USER_FILESYSTEM_CLASS =
+      new Builder(Name.USER_FILESYSTEM_CLASS)
+          .setDefaultValue("alluxio.client.file.BaseFileSystem")
```
ZacBlanco: Do you think we should list out the different options this can be configured with?

core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
```diff
@@ -148,10 +149,10 @@ private static FileSystem create(FileSystemContext context, boolean cachingEnabl
...
         }
       }
-      if (context.getClusterConf().getBoolean(PropertyKey.USER_METADATA_CACHE_ENABLED)) {
-        return new CachingFileSystem(context, cachingEnabled);
-      }
-      return BaseFileSystem.create(context, cachingEnabled);
+      Class fsClass = context.getClusterConf().getClass(PropertyKey.USER_FILESYSTEM_CLASS);
+      Class[] ctorArgClasses = new Class[] {FileSystemContext.class, boolean.class};
+      Object[] ctorArgs = new Object[] {context, cachingEnabled};
+      return (FileSystem) CommonUtils.createNewClassInstance(fsClass, ctorArgClasses, ctorArgs);
```
ZacBlanco: I'm not a huge fan of creating the classes this way because it creates an implicit rather than explicit contract on all implementations to use the exact same constructor. I don't think it's the best approach.

Because we don't have many FS implementations I think I'm okay leaving it for now, but we might want to consider a factory-based pattern or something similar if we end up extending this more in the future.

calvinjia: I agree, even in our current code we don't have a standard way of creating the file system

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Good catch, I forgot the change the description.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3368,29 +3368,29 @@ public String toString() {
...
           .build();
-  public static final PropertyKey USER_METADATA_CACHE_ENABLED =
-      new Builder(Name.USER_METADATA_CACHE_ENABLED)
-          .setDefaultValue(false)
-          .setDescription("If this is enabled, metadata of paths will be cached. "
-              + "The cached metadata will be evicted when it expires after "
-              + Name.USER_METADATA_CACHE_EXPIRATION_TIME
-              + " or the cache size is over the limit of "
-              + Name.USER_METADATA_CACHE_MAX_SIZE + ".")
+  public static final PropertyKey USER_FILESYSTEM_CLASS =
```
apc999: nit: can you put this property in alphabet order?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4539,8 +4539,7 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.user.short.circuit.preferred";
     public static final String USER_WORKER_LIST_REFRESH_INTERVAL =
         "alluxio.user.worker.list.refresh.interval";
-    public static final String USER_METADATA_CACHE_ENABLED =
-        "alluxio.user.metadata.cache.enabled";
+    public static final String USER_FILESYSTEM_CLASS = "alluxio.user.filesystem.class";
```
apc999: nit: can you put this property in alphabet order?

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: updated, thanks. There are other keys which are not in order, but I won't address that in this PR.

calvinjia:
@apc999 Conflict resolution was nontrivial, please take a look at the latest commit

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3152,36 +3169,19 @@ public String toString() {
...
-              + Name.USER_METADATA_CACHE_MAX_SIZE + ".")
-          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
-          .setScope(Scope.CLIENT)
-          .build();
   public static final PropertyKey USER_METADATA_CACHE_MAX_SIZE =
       new Builder(Name.USER_METADATA_CACHE_MAX_SIZE)
           .setDefaultValue(100000)
-          .setDescription("Maximum number of paths with cached metadata.")
+          .setDescription("Maximum number of paths with cached metadata. Only valid if the "
+              + "filesystem is alluxio.client.file.CachingFileSystem.")
```
apc999: update

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3152,36 +3169,19 @@ public String toString() {
...
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
   public static final PropertyKey USER_METADATA_CACHE_EXPIRATION_TIME =
       new Builder(Name.USER_METADATA_CACHE_EXPIRATION_TIME)
           .setDefaultValue("10min")
           .setDescription("Metadata will expire and be evicted after being cached for this time "
-              + "period.")
+              + "period. Only valid if the filesystem is alluxio.client.file.CachingFileSystem.")
```
apc999: update

apc999:

PR #10748, opened 2020-01-14
Implement a User-Side Alluxio Local Cache

apc999:
Overview
====

The Alluxio Local Cache will be a lightweight caching library for compute frameworks. Different from a server-side Alluxio caching service, this caching library can be embedded into an application  as a part of Alluxio client jar as long as user sets `alluxio.user.client.cache.enabled=true`.

Pros/Cons
=====
Compare the client-side cache vs. server-side cache (e.g., current Alluxio service)
Pros
  - A light-weight cache to improve performance;
  - Easy to deploy and integrate (e.g., one can use the HDFS compatible API in Alluxio client jar to work with this library when `alluxio.user.client.cache.enabled=true`;

Cons
  - Potential data dup across different client-side application processes
  - Hard to achieve fine-grained caching control over the cluster 

Implementation Details
=====

We have introduced a few components including:
- `LocalCacheFileSystem`: as the one implementation of Alluxio `FileSystem` class for users to operate Alluxio Filesystem with a local cache embedded
- `LocalCacheFileInStream` is the entry point to retrieve a file in streaming reads or positioned reads. Note that, files are divided into small segments called "pages" (size controled by `alluxio.user.client.cache.page.size`) 
- `LocalCacheManager` provides a thread-safe cache abstraction on the local data. 
- `PageStore`: abstraction of a dump storage for pages. This implementation provides one implementation based on local file system (each page stored as a file) and one based on rocksdb (each page stored as a key-value pair)
- `CacheEvictor`: abstraction of cache replacement policies

Scope of this initial implementation
=====
- Only support reading static files
- Only a few selected cache policies initially (e.g., LRU)


core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -101,7 +101,7 @@
    *
    * @param fsContext file system context
    */
-  public BaseFileSystem(FileSystemContext fsContext) {
+  protected BaseFileSystem(FileSystemContext fsContext) {
```
calvinjia: @apc999 This needs to be public or reflection will not work

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,300 @@
...
+   * @param path path of the file
+   * @param options read options
+   * @param externalFs the external file system if a cache miss occurs
+   * @param cacheManager local cache manager
+   */
+  public LocalCacheFileInStream(AlluxioURI path, OpenFilePOptions options, FileSystem externalFs,
+      CacheManager cacheManager) {
+    mPageSize = externalFs.getConf().getBytes(PropertyKey.USER_CLIENT_CACHE_PAGE_SIZE);
+    mPath = path;
+    // Lazy init of status object
```
apc999: is this already a lazy init or just a TODO, @calvinjia ?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,300 @@
...
+    int bytesRead = 0;
+    long currentPosition = pos;
+    long lengthToRead = Math.min(len, mStatus.getLength() - pos);
+    // for each page, check if it is available in the cache
+    while (bytesRead < lengthToRead) {
+      long currentPage = currentPosition / mPageSize;
+      int currentPageOffset = (int) (currentPosition % mPageSize);
+      int bytesLeftInPage = (int) Math.min(mPageSize - currentPageOffset, lengthToRead - bytesRead);
+      PageId pageId = new PageId(mStatus.getFileId(), currentPage);
+      try (ReadableByteChannel cachedData = mCacheManager.get(pageId, currentPageOffset)) {
```
apc999: how shall we handle IOException here when reading cache? Assume we shall fallback to external stream? Currently the IOE will be escalated.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,300 @@
...
+      return -1;
+    }
+    int bytesRead = 0;
+    // for each page, check if it is available in the cache
+    while (bytesRead < len && mPosition < mStatus.getLength()) {
+      long currentPage = mPosition / mPageSize;
+      int currentPageOffset = (int) (mPosition % mPageSize);
+      int bytesLeftInPage = (int) Math.min(mPageSize - currentPageOffset, len - bytesRead);
+      PageId pageId = new PageId(mStatus.getFileId(), currentPage);
+      try (ReadableByteChannel cachedData = mCacheManager.get(pageId, currentPageOffset)) {
```
apc999: how shall we handle IOException here when reading cache? Assume we shall fallback to external stream? Currently the IOE will be escalated.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,300 @@
...
+              break;
+            }
+          }
+          Preconditions.checkState(buf.position() == buf.limit());
+          bytesRead += bytesLeftInPage;
+          mPosition += bytesLeftInPage;
+        } else { // cache miss
+          byte[] page = readExternalPage(mPosition);
+          if (page.length > 0) {
+            mCacheManager.put(pageId, page);
```
apc999: how shall we handle failed put with IOE? shall we silently swallow the exception (w/ warning)?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,300 @@
...
+            if (cachedData.read(buf) == -1) {
+              break;
+            }
+          }
+          Preconditions.checkState(buf.position() == buf.limit());
+          bytesRead += bytesLeftInPage;
+          currentPosition += bytesLeftInPage;
+        } else { // cache miss
+          byte[] page = readExternalPage(currentPosition);
+          mCacheManager.put(pageId, page);
```
apc999: handle IOException?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,249 @@
...
+ * </ul>
+ */
+@ThreadSafe
+public class LocalCacheManager implements CacheManager {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalCacheManager.class);
+
+  private static final int LOCK_SIZE = 1024;
+  private final int mPageSize;
+  private final long mCacheSize;
+  private final CacheEvictor mEvictor;
```
apc999: what's the invariant when access evictor in terms of locking?

core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
```diff
@@ -140,13 +149,13 @@ public static FileSystem create(FileSystemContext context) {
           LOG.debug("{}={} ({})", key.getName(), value, source);
         }
       }
-      Class fsClass = context.getClusterConf().getClass(PropertyKey.USER_FILESYSTEM_CLASS);
+      Class fsClass = conf.getClass(PropertyKey.USER_FILESYSTEM_CLASS);
```
ZacBlanco: should this be done through `getClusterConf()`?

core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
```diff
@@ -140,13 +149,13 @@ public static FileSystem create(FileSystemContext context) {
...
       Class[] ctorArgClasses = new Class[] {FileSystemContext.class};
       Object[] ctorArgs = new Object[] {context};
       FileSystem fs =
           (FileSystem) CommonUtils.createNewClassInstance(fsClass, ctorArgClasses, ctorArgs);
-      if (context.getClusterConf().getBoolean(PropertyKey.USER_LOCAL_CACHE_ENABLED)) {
-        return new LocalCacheFileSystem(fs);
+      if (conf.getBoolean(PropertyKey.USER_LOCAL_CACHE_ENABLED)) {
```
ZacBlanco: same question

core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
calvinjia: We previously were trying to decouple filesystem creation from the file system context, but I don't think that is necessary anymore, I've reverted the changes. See: https://github.com/Alluxio/alluxio/pull/10907

core/common/src/main/java/alluxio/wire/FileInfo.java
```diff
@@ -89,6 +90,13 @@ public long getFileId() {
     return mFileId;
   }
 
+  /**
+   * @return the file identifier
```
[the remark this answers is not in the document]
calvinjia: Added: https://github.com/Alluxio/alluxio/pull/10907

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,349 @@
...
+      return 0;
+    }
+    if (pos < 0 || pos >= mStatus.getLength()) { // at end of file
+      return -1;
+    }
+    int bytesRead = 0;
+    long currentPosition = pos;
+    long lengthToRead = Math.min(len, mStatus.getLength() - pos);
+    // for each page, check if it is available in the cache
+    while (bytesRead < lengthToRead) {
```
gpang: This looks like it shares a lot of code with `read()`. Can it be refactored/reused?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -0,0 +1,349 @@
...
+          System.arraycopy(page, currentPageOffset, b, off + bytesRead, bytesLeftInPage);
+          bytesRead += bytesLeftInPage;
+          currentPosition += bytesLeftInPage;
+          Metrics.BYTES_REQUESTED_EXTERNAL.inc(bytesLeftInPage);
+        }
+      }
+    }
+    Preconditions.checkState(
+        bytesRead == len || (bytesRead < len && currentPosition == mStatus.getLength()),
+        "Invalid number of bytes read - "
```
gpang: this looks like the same exact error message as `read()`. Can we make them distinct?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -773,6 +773,28 @@ public MetricKey build() {
           .setMetricType(MetricType.COUNTER)
           .setIsClusterAggregated(true)
           .build();
+  // Client local cache metrics
+  public static final MetricKey CLIENT_CACHE_BYTES_READ_CACHE =
+      new Builder(Name.CLIENT_CACHE_BYTES_READ_CACHE)
+          .setDescription("Total number of bytes read from the Local Cache.")
```
ZacBlanco: nit: should `Local Cache` be capitalized?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -773,6 +773,28 @@ public MetricKey build() {
...
+      new Builder(Name.CLIENT_CACHE_BYTES_READ_EXTERNAL)
+          .setDescription("Total number of bytes read from external storage due to a cache miss "
+              + "on the local cache.")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(false)
+          .build();
+  public static final MetricKey CLIENT_CACHE_BYTES_REQUESTED_EXTERNAL =
+      new Builder(Name.CLIENT_CACHE_BYTES_REQUESTED_EXTERNAL)
+          .setDescription("Total number of bytes the user requested to read which resulted in a "
+              + "cache miss. This number may be smaller than CacheBytesReadExternal due to chunk "
```
ZacBlanco: what is `CacheBytesReadExternal`? Should we reference the PK variable instead?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileSystem.java
```diff
@@ -0,0 +1,85 @@
...
+   * @param conf the configuration, only respected for the first call
+   */
+  @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(
+      value = "ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD",
+      justification = "write to static is made threadsafe")
+  public LocalCacheFileSystem(FileSystem fs, AlluxioConfiguration conf) {
+    super(fs);
+    // TODO(feng): support multiple cache managers
+    if (sCacheManager == null) {
+      synchronized (LocalCacheFileSystem.class) {
```
gpang: Would this be easier if using something like `AtomicReference`?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,257 @@
...
+@ThreadSafe
+public class LocalCacheManager implements CacheManager {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalCacheManager.class);
+
+  private static final int LOCK_SIZE = 1024;
+  private final long mPageSize;
+  private final long mCacheSize;
+  private final CacheEvictor mEvictor;
+  /** A readwrite lock pool to guard individual pages based on striping. */
+  private final ReadWriteLock[] mPageLocks = new ReentrantReadWriteLock[LOCK_SIZE];
```
gpang: Could this use `com.google.common.util.concurrent.Striped`?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,257 @@
...
+  private final PageStore mPageStore;
+  /** A readwrite lock to guard metadata operations. */
+  private final ReadWriteLock mMetaLock = new ReentrantReadWriteLock();
```
gpang: what are the locking rules for locking both a meta lock and a page lock?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,257 @@
...
+ */
+@ThreadSafe
+public class LocalCacheManager implements CacheManager {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalCacheManager.class);
+
+  private static final int LOCK_SIZE = 1024;
+  private final long mPageSize;
+  private final long mCacheSize;
+  private final CacheEvictor mEvictor;
+  /** A readwrite lock pool to guard individual pages based on striping. */
```
gpang: Is anything allowed to lock more than 1 page lock?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,257 @@
...
+  public boolean put(PageId pageId, byte[] page) throws IOException {
+    PageId victim = null;
+    PageInfo victimPageInfo = null;
+    boolean enoughSpace;
+
+    ReadWriteLock pageLock = getPageLock(pageId);
+    try (LockResource r = new LockResource(pageLock.writeLock())) {
+      try (LockResource r2 = new LockResource(mMetaLock.writeLock())) {
+        if (mMetaStore.hasPage(pageId)) {
+          LOG.debug("{} is already inserted before", pageId);
```
gpang: Can you make this a more descriptive debug message?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+        enoughSpace = mPageStore.bytes() + page.length <= mCacheSize;
+        if (enoughSpace) {
+          mMetaStore.addPage(pageId, new PageInfo(pageId, page.length));
+        } else {
+          victim = mEvictor.evict();
+          victimPageInfo = mMetaStore.getPageInfo(victim);
+        }
+      } catch (PageNotFoundException e) {
+        throw new IllegalStateException("we shall not reach here");
```
gpang: Can you improve this message?

core/client/fs/src/main/java/alluxio/client/file/cache/PageId.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.common.base.Objects;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * A class identifies a single cached page.
+ */
+@ThreadSafe
+public class PageId {
+  private final String mFileId;
```
gpang: what is a file id, and why is it a string?

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -0,0 +1,244 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * A page store implementation which utilizes rocksDB to persist the data.
+ */
+@NotThreadSafe
+public class RocksPageStore implements PageStore {
+  private static final Logger LOG = LoggerFactory.getLogger(RocksPageStore.class);
+  public static final int KEY_LEN = Long.BYTES * 2;
```
gpang: What is the format of the key?

core/client/fs/pom.xml
```diff
@@ -55,6 +55,10 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.rocksdb</groupId>
+      <artifactId>rocksdbjni</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
calvinjia: That makes sense to me.
@bf8086 @apc999 WDYT?

core/client/fs/src/main/java/alluxio/client/file/cache/evictor/LRUCacheEvictor.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Override
+  public void updateOnDelete(PageId pageId) {
+    mLRUCache.remove(pageId, UNUSED_MAP_VALUE);
+  }
+
+  @Nullable
+  @Override
+  public PageId evict() {
+    synchronized (mLRUCache) {
```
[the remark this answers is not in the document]
calvinjia: IMO finer grained synchronization is cleaner

core/common/src/main/java/alluxio/metrics/MetricKey.java
calvinjia: Updated: #10907

core/common/src/main/java/alluxio/metrics/MetricKey.java
calvinjia: Updated: #10907

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The default implementation of a metadata store for pages stored in cache.
+ */
+public class DefaultMetaStore implements MetaStore {
+  /** A map from PageId to page info. */
+  private final Map<PageId, PageInfo> mPageMap = new HashMap<>();
```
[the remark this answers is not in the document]
calvinjia: I think we are handling locking in LocalCacheManager, @bf8086

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
calvinjia: We weren't able to refactor much of it, it comes down to updating the instance variable `mPosition` vs a local variable. Do you have any suggestions? FWIW the other implementations of read/positionedread are also duplicated.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
calvinjia: Updated: #10907

core/client/fs/src/main/java/alluxio/client/file/cache/PageId.java
calvinjia: fileID will be used to key the pages, any unique identifier will be ok

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileSystem.java
calvinjia: Are you suggesting to use `getAndSet`, that requires creating a new instance of cachemanager each time?

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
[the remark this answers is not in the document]
bf8086: Yeah the `LocalCacheManager` manages the locks for metadata operations.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
bf8086: It is `PageIndex(long) + FileIdentifier(String)`. This constant here is only used for estimation of the overhead.

calvinjia:
alluxio-bot, feature-merge this please.

core/client/fs/pom.xml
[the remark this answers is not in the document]
apc999: marked as `provided` in shaded client module. so we can keep the original jar size

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -0,0 +1,257 @@
...
+    boolean enoughSpace;
+
+    ReadWriteLock pageLock = getPageLock(pageId);
+    try (LockResource r = new LockResource(pageLock.writeLock())) {
+      try (LockResource r2 = new LockResource(mMetaLock.writeLock())) {
+        if (mMetaStore.hasPage(pageId)) {
+          LOG.debug("{} is already inserted before", pageId);
+          return false;
+        }
+        enoughSpace = mPageStore.bytes() + page.length <= mCacheSize;
```
calvinjia: @bf8086 From what I understand, `mPageStore.bytes()` is not protected by `mMetaLock`, so if `enoughSpace` is true here, there is no guarantee that it will continue to hold true.

calvinjia: Instead of `mPageStore.bytes()` I think we need to use something like `mMetastore.bytes()`? Since we can guarantee that is not changing and we will reserve the required space while still holding the metalock

core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
```diff
@@ -0,0 +1,77 @@
...
+   * Writes a new page from a source channel with best effort. It is possible that this put
+   * operation returns without page written due to transient behavior not due to failures writing
+   * to disks.
+   *
+   * @param pageId page identifier
+   * @param page page data
+   * @throws IOException if error happens when writing the page to disk
+   * @return true on a successful put or false due to transient
+   */
+  boolean put(PageId pageId, byte[] page) throws IOException;
```
calvinjia: Since this is best effort, could make this API not throw exception, even on write failure?

core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Wraps the page in a channel or null if the queried page is not found in the cache.
+   *
+   * @param pageId page identifier
+   * @return a channel to read the page
+   * @throws IOException if error happens when reading the page
+   */
+  @Nullable
+  ReadableByteChannel get(PageId pageId) throws IOException;
```
calvinjia: Could we default to returning null from this API instead of throwing an exception, since cache hit is best effort?

PR #10749, opened 2020-01-14
Fix some bugs and add new properties for client-side cache

bf8086:
Fixed index out of bound exception for cache manager locking.
Fixed exception when reading from stream with oversized buffer.
Added properties to configure page store directory and type.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
@@ -100,10 +100,12 @@ public class LocalCacheFileInStream extends FileInStream {
 
   @Override
   public int read(byte[] b, int off, int len) throws IOException {
+    Preconditions.checkArgument(len >= 0, "length should be non-negative");
+    Preconditions.checkArgument(off >= 0, "offset should be non-negative");
     if (len == 0) {
       return 0;
     }
-    if (mPosition == mStatus.getLength()) { // at end of file
+    if (mPosition >= mStatus.getLength()) { // at end of file
       return -1;
     }
     int bytesRead = 0;
@@ -140,7 +142,9 @@ public class LocalCacheFileInStream extends FileInStream {
       }
     }
     Preconditions.checkState(bytesRead == len || (bytesRead < len && remaining() == 0),
-        "Invalid number of bytes read");
+        "Invalid number of bytes read - "
+            + "bytes to read = %d, actual bytes read = %d, bytes remains in file %d",
+        len, bytesRead, remaining());
     return bytesRead;
   }
 
@@ -167,13 +171,23 @@ public class LocalCacheFileInStream extends FileInStream {
 
   @Override
   public int positionedRead(long pos, byte[] b, int off, int len) throws IOException {
+    Preconditions.checkArgument(len >= 0, "length should be non-negative");
+    Preconditions.checkArgument(off >= 0, "offset should be non-negative");
+    Preconditions.checkArgument(pos >= 0, "position should be non-negative");
+    if (len == 0) {
+      return 0;
+    }
+    if (pos < 0 || pos >= mStatus.getLength()) { // at end of file
+      return -1;
+    }
     int bytesRead = 0;
     long currentPosition = pos;
+    long lengthToRead = Math.min(len, mStatus.getLength() - pos);
     // for each page, check if it is available in the cache
-    while (bytesRead < len) {
+    while (bytesRead < lengthToRead) {
       long currentPage = currentPosition / mPageSize;
       int currentPageOffset = (int) (currentPosition % mPageSize);
-      int bytesLeftInPage = (int) Math.min(mPageSize - currentPageOffset, len - bytesRead);
+      int bytesLeftInPage = (int) Math.min(mPageSize - currentPageOffset, lengthToRead - bytesRead);
       PageId pageId = new PageId(mStatus.getFileId(), currentPage);
       try (ReadableByteChannel cachedData = mCacheManager.get(pageId, currentPageOffset)) {
         if (cachedData != null) { // cache hit
@@ -199,7 +213,11 @@ public class LocalCacheFileInStream extends FileInStream {
         }
       }
     }
-    Preconditions.checkState(bytesRead == len, "Invalid number of bytes read");
+    Preconditions.checkState(
+        bytesRead == len || (bytesRead < len && currentPosition == mStatus.getLength()),
+        "Invalid number of bytes read - "
+            + "bytes to read = %d, actual bytes read = %d, bytes remains in file %d",
+        len, bytesRead, mStatus.getLength() - currentPosition);
     return bytesRead;
   }
 
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
@@ -100,9 +100,15 @@ public class LocalCacheFileInStream extends FileInStream {
 
   @Override
   public int read(byte[] b, int off, int len) throws IOException {
+    if (len == 0) {
+      return 0;
+    }
+    if (mPosition == mStatus.getLength()) { // at end of file
+      return -1;
+    }
     int bytesRead = 0;
     // for each page, check if it is available in the cache
-    while (bytesRead < len) {
+    while (bytesRead < len && mPosition < mStatus.getLength()) {
       long currentPage = mPosition / mPageSize;
       int currentPageOffset = (int) (mPosition % mPageSize);
       int bytesLeftInPage = (int) Math.min(mPageSize - currentPageOffset, len - bytesRead);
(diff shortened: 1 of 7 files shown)


bf8086:
@calvinjia @apc999 PTAL.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -100,9 +100,15 @@ public int read(byte[] b) throws IOException {
 
   @Override
   public int read(byte[] b, int off, int len) throws IOException {
+    if (len == 0) {
+      return 0;
+    }
+    if (mPosition == mStatus.getLength()) { // at end of file
```
apc999: `mPosition >= mStatus.getLength()`?

calvinjia:
jenkins, test this please.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
apc999: how about updating `positionedRead` https://github.com/Alluxio/alluxio/pull/10749/files#diff-dda4e9db6a53b7319a9d5c19177847bcR173 ?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -124,14 +130,17 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+            System.arraycopy(page, currentPageOffset, b, off + bytesRead, bytesLeftInPage);
+            bytesRead += bytesLeftInPage;
+            mPosition += bytesLeftInPage;
+          }
         }
       }
     }
-    Preconditions.checkState(bytesRead == len, "Invalid number of bytes read");
+    Preconditions.checkState(bytesRead == len || (bytesRead < len && remaining() == 0),
+        "Invalid number of bytes read");
```
apc999: please include numbers we are checking here in the error message, or it is hard to tell what's going wrong

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
calvinjia: Good catch, positioned read should be updated as well

core/client/fs/src/main/java/alluxio/client/file/cache/store/PageStoreOptions.java
```diff
@@ -11,11 +11,39 @@
...
  * Options used to instantiate a {@link alluxio.client.file.cache.PageStore}.
  */
 public abstract class PageStoreOptions {
 
+  /**
+   * Creates an instance of the {@link PageStoreOptions} based on configurations.
+   * @param conf the Alluxio configuration
+   * @return the created page store options
+   */
+  public static PageStoreOptions create(AlluxioConfiguration conf) {
```
apc999: Since this is simple enough, I would suggest to move the body this function into factory method `PageStore.create`, inside which we can switch the creation of options based on `PropertyKey.USER_CLIENT_CACHE_STORE_TYPE`.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -52,6 +53,15 @@ static PageStore create() {
     return create(new LocalPageStoreOptions());
   }
 
+  /**
+   * Creates a new instance of {@link PageStore} based on configuration.
+   * @param conf configuration
+   * @return the {@link PageStore}
+   */
+  static PageStore create(AlluxioConfiguration conf) {
```
apc999: do we still want to keep `create()` then?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3120,6 +3120,19 @@ public String toString() {
...
+      new Builder(Name.USER_CLIENT_CACHE_DIR)
+          .setDefaultValue("/tmp")
+          .setDescription("The directory where client-side cache is stored.")
+          .setScope(Scope.CLIENT)
+          .build();
+  public static final PropertyKey USER_CLIENT_CACHE_STORE_TYPE =
+      new Builder(Name.USER_CLIENT_CACHE_STORE_TYPE)
+          .setDefaultValue("LOCAL")
+          .setDescription("The type of page store to use for client-side cache. Can be either "
+              + "`LOCAL` or `ROCKS`.")
```
apc999: specify a bit more here what is `LOCAL` and what is `ROCKS`?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -100,9 +100,15 @@ public int read(byte[] b) throws IOException {
 
   @Override
   public int read(byte[] b, int off, int len) throws IOException {
+    if (len == 0) {
```
apc999: Add `Preconditions.check`  on `len >= 0` and `off >= 0`?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -103,7 +103,7 @@ public LocalCacheManager(FileSystemContext fsContext) {
    * @return the corresponding page lock
    */
   private ReadWriteLock getPageLock(PageId pageId) {
-    return mPageLocks[(int) (pageId.getFileId() + pageId.getPageIndex()) % LOCK_SIZE];
+    return mPageLocks[Math.floorMod((int) (pageId.getFileId() + pageId.getPageIndex()), LOCK_SIZE)];
```
apc999: will `Math.floorMod` handle overflow if `pageId.getFileId() + pageId.getPageIndex()` is larger than `LONG.MAX_VALUE`?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
bf8086: is `mPosition > mStatus.getLength()` possible?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3120,6 +3120,19 @@ public String toString() {
               + "`alluxio.client.file.cache.LRUCacheEvictor`.")
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_CLIENT_CACHE_DIR =
+      new Builder(Name.USER_CLIENT_CACHE_DIR)
+          .setDefaultValue("/tmp")
```
apc999: maybe something like `/tmp/alluxio_cache` to avoid creating too many dirs under `/tmp` directly?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
apc999: just to be safe and prevent hitting bugs elsewhere

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
bf8086: We don't need it to handle - if the sum is overflow it will become negative, but the floorMod will always return non-negative when the second argument is positive.

core/common/src/main/java/alluxio/conf/PropertyKey.java
bf8086: Done.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
bf8086: Done.

core/common/src/main/java/alluxio/conf/PropertyKey.java
bf8086: Done.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
bf8086: Done.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
bf8086: Done.

core/client/fs/src/test/java/alluxio/client/file/cache/LocalCacheFileInStreamTest.java
```diff
@@ -300,7 +325,7 @@ public ReadableByteChannel get(PageId pageId, int pageOffset) {
       }
       mPagesServed++;
       return Channels.newChannel(new ByteArrayInputStream(
-          Arrays.copyOfRange(mPages.get(pageId), pageOffset, PAGE_SIZE - pageOffset)));
+          Arrays.copyOfRange(mPages.get(pageId), pageOffset, PAGE_SIZE)));
```
calvinjia: Good catch

calvinjia:

PR #10756, opened 2020-01-16
Fail nested jobs correctly in all of the jobs for JobMaster failures

bradyoo:
For errors like ResourceExhaustedException and JobDoesNotExistException, the WorkflowTracker will only stamp Status.FAILED from a level above.

Reproduction: (see Test)

mJobMaster.run(new Composite..([new Composite(new DummyJob....

Expected:

Name: Composite
..
Status: FAILED
Message: The job definition for config bleh does not exist
Task 0
    Name: Composite
    ...
    Status: FAILED
    Message: The job definition for config bleh does not exist

Actual:

Name: Composite
..
Status: FAILED
Message: The job definition for config bleh does not exist
Task 0
    Name: Composite
    ...
    Status: RUNNING

### DIFF
--- a/job/server/src/main/java/alluxio/master/job/workflow/WorkflowTracker.java
+++ b/job/server/src/main/java/alluxio/master/job/workflow/WorkflowTracker.java
@@ -250,6 +250,7 @@ public class WorkflowTracker {
         mJobMaster.run(childJobConfig, childJobId);
       } catch (JobDoesNotExistException | ResourceExhaustedException e) {
         LOG.warn(e.getMessage());
+        workflowExecution.stop(Status.FAILED, e.getMessage());
         stop(jobId, Status.FAILED, e.getMessage());
       }
     }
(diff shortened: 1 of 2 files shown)


job/server/src/main/java/alluxio/master/job/workflow/WorkflowTracker.java
```diff
@@ -250,6 +250,7 @@ private synchronized void next(long jobId) {
         mJobMaster.run(childJobConfig, childJobId);
       } catch (JobDoesNotExistException | ResourceExhaustedException e) {
         LOG.warn(e.getMessage());
+        workflowExecution.stop(Status.FAILED, e.getMessage());
```
calvinjia: Should this logic be in the `stop` call? Does the bug happen for workflows which fail through `PlanTracker#onPlanStatusChange`?

bradyoo: No the bug doesn't happen for workflows which fail through PlanTracker#onPlanStatusChange because all logic that goes through onPlanStatusChange are not workflows and are Plans which are handled by the PlanTracker. This is also the reason why the stop logic looks at the parent. onPlanStatusChange is the normal case and this JobDoesNotExist + ResourceExhausted is the edge case.

calvinjia:

PR #10758, opened 2020-01-16
Add metrics for cache hit/miss

calvinjia:
Adds 3 metrics
number of bytes read from local cache
number of bytes read from external store
number of bytes requested from external store (may be less than bytes read because we read full pages)


calvinjia:
@apc999 Yeah we should add more metrics over time, I would prefer to merge this PR first so we can have an example of how metrics are added in code and consumed by users.

calvinjia:

LuQQiu:
@calvinjia  is number of bytes read from local cache + number of bytes read from external store = existing number of bytes read from alluxio

calvinjia:
@LuQQiu It's not quite equivalent, since we read additional data.

Local Cache + External Requested = # User requested bytes (through read api)
Local Cache + External Read = true # of bytes read

PR #10763, opened 2020-01-17
Add positioned readable HDFS ufs input stream

madanadit:
When HDFS is remote and slow, we use the positioned read API instead of seek with this PR. 

HDFS FSDataInputStream seek may be implemented as read and discard. Over remote HDFS, this may increase the amount of data read over the network many folds based on the workload. Using positioned read, the amount of data requested from the HDFS client matches the amount of data read from the HDFS datanode.


underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public int read() throws IOException {
+    byte[] buffer = new byte[1];
+    int bytesRead = ((FSDataInputStream) in).read(mPos, buffer, 0, buffer.length);
+    if (bytesRead > 0) {
+      mPos += bytesRead;
+    }
+    return bytesRead < 0 ? bytesRead : (int) buffer[0];
```
calvinjia: `BufferUtils.byteToInt`?

calvinjia:
@madanadit Thanks for the improvement! Is there a case for always using positioned read, or is using seek + read faster when the underlying storage is reasonably fast?

madanadit:
@calvinjia I'm actually running that experiment now, will update here once done.

madanadit:
@calvinjia IO and metadata tests look good with using the new API for local HDFS, so I've made it the default implementation instead of hiding it behind a flag.
Testing should be covered under `UnderFileSystemCommonOperations` so I'm not adding new tests in this PR.

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -0,0 +1,86 @@
...
+  private long mPos;
+
+  HdfsPositionedUnderFileInputStream(FSDataInputStream in, long pos) {
+    super(in);
+    mPos = pos;
+  }
+
+  @Override
+  public int available() throws IOException {
+    int available = in.available() + (int) (((FSDataInputStream) in).getPos() - mPos);
```
calvinjia: Could you explain the logic here?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public int read() throws IOException {
+    byte[] buffer = new byte[1];
+    int bytesRead = ((FSDataInputStream) in).read(mPos, buffer, 0, buffer.length);
+    if (bytesRead > 0) {
+      mPos += bytesRead;
+      return BufferUtils.byteToInt(buffer[0]);
+    }
+    Preconditions.checkArgument(bytesRead != 0,
+        "Expected a non-zero value if end of stream has not been reached");
```
calvinjia: Is it valid to hit EOS here and return `-1`?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
```diff
@@ -563,13 +563,8 @@ public InputStream open(String path, OpenOptions options) throws IOException {
...
       try {
         FSDataInputStream inputStream = hdfs.open(new Path(path));
-        try {
-          inputStream.seek(options.getOffset());
-        } catch (IOException e) {
-          inputStream.close();
-          throw e;
-        }
-        return new HdfsUnderFileInputStream(inputStream);
+        // For remote HDFS, use positionedRead instead of seek
```
calvinjia: Perhaps change this comment to say why positionedRead is better than seeking in some cases?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
madanadit: Calling seek in this stream doesn't change the wrapped FSDataInputStream and the number of bytes remaining would never be negative

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
madanadit: yes, `-1` is valid return value for this method

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
madanadit: in.available + in.getPos = estimate of total bytes remaining

madanadit:
```
Tests in error: 
alluxio.server.ft.journal.MultiMasterJournalTest.testCheckpointReplay(alluxio.server.ft.journal.MultiMasterJournalTest)
  Run 1: MultiMasterJournalTest.before:53 » IO java.util.concurrent.TimeoutException: T...
  Run 2: MultiMasterJournalTest.testCheckpointReplay:64 » IO java.util.concurrent.Timeo...
```

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
madanadit: done

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
madanadit: thx, fixed this method

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -0,0 +1,88 @@
...
+  public long getPos() throws IOException {
+    return mPos;
+  }
+
+  @Override
+  public int read() throws IOException {
+    byte[] buffer = new byte[1];
+    int bytesRead = ((FSDataInputStream) in).read(mPos, buffer, 0, buffer.length);
+    if (bytesRead > 0) {
+      mPos += bytesRead;
```
calvinjia: Consider using this class's `read(byte[])` method? Then we should only need to bookkeep `mPos` in one place

madanadit: thx, done

madanadit:
@calvinjia did you have more comments?

calvinjia:

PR #10766, opened 2020-01-20
Start sinks on client when client metrics are enabled

calvinjia:
This allows users to integrate client metrics with 3rd party metrics tools (similar to the servers)


calvinjia:

PR #10768, opened 2020-01-20
Add client config load before initializing metrics heartbeat

LuQQiu:
When client-side configuration is wrong or missing, client heartbeat may have connection issues (ie due to security settings).  This PR adds client side load configuration before initializign metrics heartbeat to solve the issue.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -229,6 +229,15 @@ public final class FileSystemContext implements Closeable {
     mClosed.set(false);
     mMasterClientContext = MasterClientContext.newBuilder(ctx)
         .setMasterInquireClient(masterInquireClient).build();
+    try {
+      InetSocketAddress masterAddr = masterInquireClient.getPrimaryRpcAddress();
+      mMasterClientContext.loadConf(masterAddr, true, true);
+    } catch (UnavailableException e) {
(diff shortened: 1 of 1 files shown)


core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -229,6 +229,15 @@ private synchronized void initContext(ClientContext ctx,
...
         .setMasterInquireClient(masterInquireClient).build();
+    try {
+      InetSocketAddress masterAddr = masterInquireClient.getPrimaryRpcAddress();
+      mMasterClientContext.loadConf(masterAddr, true, true);
+    } catch (UnavailableException e) {
+      LOG.error("Failed to get master address during initialization", e);
+    } catch (AlluxioStatusException ae) {
+      LOG.error("Failed to load configuration from "
+          + "meta master during initialization", ae);
+    }
```
calvinjia: Should we only do this if metrics are enabled?

LuQQiu: For the safest change, we should only do this if metrics are enabled.
Update client configuration beforehand in all cases may help acquire and work with block worker client and file system master client.

LuQQiu:
Jenkins, test this please

LuQQiu:
@calvinjia PTAL, thanks

calvinjia:

PR #10769, opened 2020-01-20
Implement parallel block eviction

ggezer:


ggezer:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2294,6 +2294,14 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_NETWORK_BLOCK_EVICTION_THREADS =
+      new Builder(Name.WORKER_NETWORK_BLOCK_EVICTION_THREADS)
+          .setDefaultSupplier(() -> Runtime.getRuntime().availableProcessors(),
```
calvinjia: Is this a reasonable default?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4352,6 +4360,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.worker.network.block.reader.threads.max";
     public static final String WORKER_NETWORK_BLOCK_WRITER_THREADS_MAX =
         "alluxio.worker.network.block.writer.threads.max";
+    public static final String WORKER_NETWORK_BLOCK_EVICTION_THREADS =
+        "alluxio.worker.network.block.eviction.threads";
```
calvinjia: Does this fit the `network` category?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -691,39 +703,62 @@ private void freeSpaceInternal(long sessionId, long availableBytes, BlockStoreLo
...
+                for (BlockStoreEventListener listener : mBlockStoreEventListeners) {
+                  listener.onMoveBlockByWorker(sessionId, blockId, moveResult.getSrcLocation(),
+                      newLocation);
+                }
+              }
+              return true;
             }
+          } catch (InvalidWorkerStateException e) {
+            // Evictor is not working properly
+            LOG.error("Failed to demote blockId {}, this is temp block", blockId);
```
calvinjia: Should this be warn or should we do something to handle this error?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -691,39 +703,62 @@ private void freeSpaceInternal(long sessionId, long availableBytes, BlockStoreLo
...
+          } catch (BlockAlreadyExistsException e) {
+            LOG.info("Failed to demote blockId {}, it could be already moved", blockId);
+          } catch (BlockDoesNotExistException e) {
+            LOG.info("Failed to demote blockId {}, it could be already deleted", blockId);
           }
-        }
+          return false;
+        });
```
calvinjia: This block of logic is pretty involved, perhaps make a separate method?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -691,39 +703,62 @@ private void freeSpaceInternal(long sessionId, long availableBytes, BlockStoreLo
...
+      }
+      // Execute move orders in thread pool.
+      try {
+        List<Future<Boolean>> moveResults = mEvictionExecutor.invokeAll(moveTasks);
+        long failCount = moveTasks.size() - moveResults.stream().filter(
+            (fut) -> {
+              try {
+                return fut.get();
+              } catch (Exception e) {
+                return false;
```
calvinjia: Do we need to distinguish between the future failures (which return false) and the future exceptions (which are caught here and will return false)?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -691,39 +703,62 @@ private void freeSpaceInternal(long sessionId, long availableBytes, BlockStoreLo
...
+              }
+            }).count();
+        LOG.info("Finished moving:{} blocks to tier: {}. Failure count: {} ", moveTasks.size(),
```
calvinjia: space between `:` and `{`?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -691,39 +703,62 @@ private void freeSpaceInternal(long sessionId, long availableBytes, BlockStoreLo
...
+            mStorageTierAssoc.getAlias(tierOrdinal), failCount);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new RuntimeException("Interrupted while executing eviction move orders.");
```
calvinjia: Should we attempt to cancel or interrupt the tasks running in the eviction executor? Also, is throwing runtime exception here handled properly by the callers?

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: I found this to yield min eviction duration.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: This is not expected so I'd leave it ERROR based on above comment.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: `invokeAll` cancels unfinished tasks if interrupted.

ggezer: Caller was crash-exiting. I declared `InterruptedException` along the call chain to gracefully handle it.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: `invokeAll` guarantees that all returned futures will be completed.
This is merely to satisfy method signature.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: not really. Changed to `WORKER_EVICTOR_BLOCK_TRANSFER_THREAD_COUNT`

ggezer:
jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Right, it pretty much allocates all resources on the machine to eviction if needed, that seems a bit aggressive?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
calvinjia: How does `InterruptedException` get handled further up the stack? If eviction is interrupted, should we just interpret that as out of space?

ggezer: Not throwing eviction will cause cascading eviction runs to continue until the heartbeat executor notices the interruption. And we'd need to properly store interruption status to heartbeat executor.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Should we leave one core out? Or half the core count is better?

calvinjia: Half of the cores seems a more moderate choice, you can add some more details in the description about how this parameter can be tuned.

ggezer:
jenkins, test this please

ggezer:
Closing for now...

PR #10770, opened 2020-01-21
Make FileOutStream an abstract class similar to FileInStream

calvinjia:
AbstractOutStream -> FileOutStream
FileOutStream -> AlluxioFileOutStream

This makes it symmetrical with FileInStream


core/client/fs/src/main/java/alluxio/client/file/FileOutStream.java
```diff
@@ -11,312 +11,40 @@
...
- * blocks in the file and abstracts the switching between streams. The backing streams can write to
- * Alluxio space in the local machine or remote machines. If the {@link UnderStorageType} is
- * {@link UnderStorageType#SYNC_PERSIST}, another stream will write the data to the under storage
- * system.
+ * An abstraction of the output stream API in Alluxio to write data to a file or a block. In
+ * addition to extending abstract class {@link OutputStream} as the basics, it also keeps counting
+ * the number of bytes written to the output stream, and extends {@link Cancelable} to abort the
+ * writes.
  */
-@PublicApi
```
apc999: should we keep this annotation?

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
```diff
@@ -0,0 +1,321 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Provides a streaming API to write a file. This class wraps the BlockOutStreams for each of the
+ * blocks in the file and abstracts the switching between streams. The backing streams can write to
+ * Alluxio space in the local machine or remote machines. If the {@link UnderStorageType} is
+ * {@link UnderStorageType#SYNC_PERSIST}, another stream will write the data to the under storage
+ * system.
+ */
+@PublicApi
```
apc999: should we still keep this annotation?

core/client/fs/src/main/java/alluxio/client/file/FileOutStream.java
calvinjia: Good point, updated.

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
calvinjia: Removed, thanks

calvinjia:

PR #10771, opened 2020-01-21
Exclude com.ibm package in shaded client jar to support ibm jvm

LuQQiu:
In LoginModuleConfigurationUtils.java, we use `com.ibm.security` for login in IBM java. This package should be unshaded in Alluxio client jar to avoid errors like `Unable to find LoginModule class: alluxio.shaded.client.com.ibm.security.auth.module.LinuxLoginModule`

### DIFF
--- a/shaded/client/pom.xml
+++ b/shaded/client/pom.xml
@@ -195,6 +195,8 @@
                   <excludes>
                     <exclude>**/pom.xml</exclude>
                     <!-- Exclude the packages belonging to JDK -->
+                    <exclude>com/ibm/*</exclude>
+                    <exclude>com/ibm/**/*</exclude>
                     <exclude>com/sun/tools/*</exclude>
                     <exclude>com/sun/javadoc/*</exclude>
                     <exclude>com/sun/security/*</exclude>
--- a/shaded/client/pom.xml
+++ b/shaded/client/pom.xml
@@ -195,9 +195,8 @@
                   <excludes>
                     <exclude>**/pom.xml</exclude>
                     <!-- Exclude the packages belonging to JDK -->
-                    <!-- TODO(lu) shade more detailed ibm JDK packages like com.ibm.security -->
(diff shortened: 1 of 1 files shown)


jiacheliu3:
Requested review from @calvinjia for veteran opinions on whether `com.ibm.security` or `com.ibm.**` makes more sense.

LuQQiu:
More info:
The code path is https://github.com/Alluxio/alluxio/blob/244ef3f7ba11d07cafa29c6e53c95fdc6ac11016/core/common/src/main/java/alluxio/security/login/LoginModuleConfigurationUtils.java#L47

The full error is:
```
scala> val fs = FileSystem.Factory.get
WARN BaseUserState: Subject login  failed: Failed to login: unable to find LoginModule class: alluxio.shaded.client.com.ibm.security.auth.module.LinuxLoginModule
```

calvinjia:
@LuQQiu It looks like we explicitly exclude packages for the `sun` distribution, so I think we should do something similar for `ibm`?

LuQQiu:
@calvinjia Agreed that it's better to shade packages more explicitly. The changes in this PR are verified in the user env that helps solve the issue but I failed to reproduce the error (so not able to narrow down the package range). I add the TODO comment to remind future developers.

jiacheliu3:
@LuQQiu Please use the below 2 lines for exclusion. This has been verified to work on IBM J9 JRE.
```
<exclude>com/ibm/security/*</exclude>
<exclude>com/ibm/security/**/*</exclude>
````

LuQQiu:
@jiacheliu3 Updated, do we want to merge this PR?

ZacBlanco:
If you want to test with different JVMs I highly recommend this tool I've been using recently: https://github.com/shyiko/jabba

jiacheliu3:
@LuQQiu i think so. LTGM. Thanks for making the change!

LuQQiu:

PR #10634, opened 2019-12-13
Fix JobMaster.cancel(jobId)

bradyoo:
JobMaster.cancel() previously cleaned up properly only if the JobWorker thread was actually executing the task. But in reality, almost all of the tasks are in the executor queue.

PR #10774, opened 2020-01-21
Fix cancel 2.1 (Port #10634 to branch-2.1)

bradyoo:
JobMaster.cancel() previously cleaned up properly only if the JobWorker
thread was actually executing the task. But in reality, almost all of
the tasks are in the executor queue.

pr-link: Alluxio/alluxio#10634
change-id: cid-a9af8ad612139b0540e8de972380f8f61c00bcb1


calvinjia:

PR #10775, opened 2020-01-21
Fix client leaks of copyFromLocal command

ggezer:
`copyFromLocal`command creates its own fs-context but it never gets closed. This fix lays necessary plumbing and properly closes the command.


calvinjia:

PR #10776, opened 2020-01-21
Allow local cache to recover from existing files

bf8086:
Add functionality to detect previously cached data on local storage and reuse them if the data is compatible.


core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -69,7 +85,64 @@ static PageStore create(AlluxioConfiguration conf) {
...
+            // check enough cache size
+            && cacheConf.getBytes(PropertyKey.USER_CLIENT_CACHE_SIZE)
+            <= conf.getBytes(PropertyKey.USER_CLIENT_CACHE_SIZE)
+            // check alluxio version
+            && cacheConf.get(PropertyKey.VERSION).equals(
+            conf.get(PropertyKey.VERSION))) {
+          LOG.info("Found recoverable local cache at {}", rootPath);
+          canLoad = true;
+        } else {
+          LOG.info("Found local cache at {} with incompatible configuration.", rootPath);
```
calvinjia: Would be nice to know why its incompatible.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -69,7 +85,64 @@ static PageStore create(AlluxioConfiguration conf) {
...
+      for (PropertyKey key : keys) {
+        properties.setProperty(key.getName(), conf.get(key));
+      }
+      try (FileOutputStream stream = new FileOutputStream(confPath.toString())) {
+        properties.store(stream, "Alluxio local cache configuration");
+      } catch (IOException e) {
+        throw new IllegalStateException(
+            String.format("failed to write cache configuration to file %s", confPath), e);
+      }
+    }
```
calvinjia: It feels kind of weird to not do the loading here. Also, could you put the clean up/load compatibility logic in another method, this method is doing a lot.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -111,17 +114,45 @@ private Path getFilePath(PageId pageId) {
...
-  @Override
-  public void close() {
+  /**
+   * @param path path of a file
+   * @return the corresponding page id, or null if the file name does not match the pattern
+   */
+  private PageId getPageId(Path path) {
+    Path parent = Preconditions.checkNotNull(path.getParent());
+    if (!Paths.get(mRoot).equals(parent.getParent())) {
+      return null;
```
calvinjia: Should we warn on files we were unable to recover? Also should we delete them?

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
bf8086: Added individual checks and log messages.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
bf8086: I agree it feels weird. However the load method needs to pass a list of page ids to the meta store which cannot be returned in this factory method.

bf8086: Extracted format and check to separate methods.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
bf8086: Added some warnings.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -73,6 +94,104 @@ static PageStore create(AlluxioConfiguration conf) {
...
+    if (!Files.exists(confPath)) {
+      return false;
+    }
+    Properties properties = ConfigurationUtils.loadPropertiesFromFile(confPath.toString());
+    if (properties == null) {
+      return false;
+    }
+    AlluxioProperties alluxioProperties = new AlluxioProperties();
+    alluxioProperties.merge(properties, Source.DEFAULT);
+    AlluxioConfiguration cacheConf = new InstancedConfiguration(alluxioProperties);
```
apc999: why we need to reload the cache conf from path? Can't we trust the argument passed in?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -91,6 +92,16 @@ public LocalCacheManager(FileSystemContext fsContext) {
     for (int i = 0; i < LOCK_SIZE; i++) {
       mPageLocks[i] = new ReentrantReadWriteLock();
     }
+    Collection<PageId> pages;
+    try {
+      pages = mPageStore.load();
+    } catch (IOException e) {
+      throw new IllegalStateException("failed to scan page cache", e);
```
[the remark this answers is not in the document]
bf8086: Added a factory method and added logic to catch exceptions and fallback to normal file stream.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -120,4 +239,12 @@ ReadableByteChannel get(PageId pageId, int pageOffset) throws IOException,
...
    */
   int size();
+
+  /**
+   * Loads page store from local storage and returns all page ids.
+   *
+   * @return collection of ids representing all pages loaded from disk
+   * @throws IOException if any error occurs
+   */
+  Collection<PageId> load() throws IOException;
```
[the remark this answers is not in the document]
bf8086: Both iterating files and rocksDB records require explicit call to `close()` method to release the resources. Returning an iterator hides this requirement and will lead to resource leak. Creating a copy of the list allow us to close the resource properly before this method returns. I would not concern too much about the memory consumption since we already store all page ids in meta store.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -111,17 +114,55 @@ private Path getFilePath(PageId pageId) {
...
+    }
     try {
-      FileUtils.deleteDirectory(new File(mRoot));
-    } catch (IOException e) {
-      LOG.warn("Failed to clean up local page store directory", e);
+      Path fileName = Preconditions.checkNotNull(path.getFileName());
```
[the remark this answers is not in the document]
bf8086: So far the encoding/decoding is rather store type specific and no logic is shared between different stores. I will create util classes if we later have more places using this encoding scheme.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
bf8086: Because configuration might changed between client restart and the data might not be compatible after the change. Here we just check the stored configuration and discard the data if the configuration is not compatible(e.g. page size changed, store type changed).

core/client/fs/src/main/java/alluxio/client/file/cache/store/PageStoreOptions.java
```diff
@@ -48,4 +63,46 @@ public void setRootDir(String rootDir) {
...
+
+  /**
+   * @return the size of the page
+   */
+  public long getPageSize() {
+    return mPageSize;
+  }
+
+  /**
+   * @param pageSize the size of the page
```
apc999: nit "pageSize the size of the page in bytes" (is it?)

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -106,22 +138,80 @@ public void delete(PageId pageId) throws PageNotFoundException {
...
   public void close() {
     mDb.close();
-    try {
-      FileUtils.deleteDirectory(new File(mRoot));
-    } catch (IOException e) {
-      LOG.warn("Failed to clean up rocksDB root directory.");
-    }
   }
 
   private byte[] getPageKey(PageId pageId) {
```
apc999: nits:
- since this helper method requires no instance state, possibly turned into a static method. so as the next method. 
- naming: `getKeyFromPageId` and `getPageIdFromKey` to be more explicit on the function? Leaving this up to you.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -106,22 +138,80 @@ public void delete(PageId pageId) throws PageNotFoundException {
...
+
+    @Override
+    public PageId next() {
+      PageId value = ensureValue();
+      mIter.next();
+      mValue = null;
+      return value;
+    }
+
+    private PageId ensureValue() {
```
apc999: still mark nullable

core/client/fs/src/test/java/alluxio/client/file/cache/store/PageStoreTest.java
```diff
@@ -101,6 +112,25 @@ public void getOffsetOverflow() throws Exception {
...
+  public void getPages() throws Exception {
+    mOptions.setRootDir(mTemp.getRoot().getAbsolutePath());
+    int len = 32;
+    int count = 16;
+    byte[] data = BufferUtils.getIncreasingByteArray(len);
+    Set<PageId> pages = new HashSet<>(count);
+    try (PageStore store = PageStore.create(mOptions)) {
+      for (int i = 0; i < count; i++) {
+        PageId id = new PageId(0, i);
+        store.put(id, data);
```
apc999: all page values are the same, can we use different page content here?

core/client/fs/src/main/java/alluxio/client/file/cache/store/PageStoreOptions.java
```diff
@@ -48,4 +63,46 @@ public void setRootDir(String rootDir) {
...
+   * @return the size of the cache
+   */
+  public long getCacheSize() {
+    return mCacheSize;
+  }
+
+  /**
+   * @param cacheSize the size of the cache
+   */
+  public void setCacheSize(long cacheSize) {
```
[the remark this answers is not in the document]
bf8086: Updated.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
bf8086: Done.

core/client/fs/src/test/java/alluxio/client/file/cache/store/PageStoreTest.java
bf8086: This test only focus on verifying the page ids. The content is verified in other tests.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -604,11 +604,11 @@ private void initializeLocalWorker() throws IOException {
   /**
    * @return the client side cache manager
    */
-  public CacheManager getCacheManager() {
+  public CacheManager getCacheManager() throws IOException {
     return getCacheManager(this);
   }
 
-  private static CacheManager getCacheManager(FileSystemContext fsContext) {
+  private static CacheManager getCacheManager(FileSystemContext fsContext) throws IOException {
```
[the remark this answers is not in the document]
bf8086: Updated to return null if failed to initialize and not retry afterwards.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileSystem.java
```diff
@@ -34,9 +44,15 @@ public LocalCacheFileSystem(FileSystem fs, FileSystemContext fsContext) {
...
+  public FileInStream openFile(AlluxioURI path, OpenFilePOptions options)
+      throws IOException, AlluxioException {
     // TODO(calvin): We should add another API to reduce the cost of openFile
-    return new LocalCacheFileInStream(path, options, mDelegatedFileSystem,
-        mFsContext.getCacheManager());
+    try {
+      return new LocalCacheFileInStream(path, options, mDelegatedFileSystem,
+          mFsContext.getCacheManager());
+    } catch (IOException e) {
+      SAMPLING_LOGGER.warn("Failed to create LocalCacheFileInStream {}", e.getMessage());
```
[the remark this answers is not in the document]
bf8086: Yes.
Removed since we are now logging at `FileSystemContext::getCacheManager` on the first initialization.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -69,10 +81,35 @@ static PageStore create(AlluxioConfiguration conf) {
...
+  /**
+   * Initialize a page store at the configured location.
+   * Data from different store type will be removed.
+   *
+   * @param rootPath root path of the page store
+   * @param storeType the page store type
+   */
+  static void initialize(String rootPath, PageStoreType storeType) throws IOException {
+    Path storePath = Paths.get(rootPath, storeType.name());
+    Files.createDirectories(storePath);
```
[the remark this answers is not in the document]
bf8086: It's probably preferable to allow page store to be created even some files are not removed. I am adding some logging to surface the error message.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -107,21 +134,65 @@ public void delete(PageId pageId) throws IOException, PageNotFoundException {
...
 
-  @Override
-  public void close() {
+  /**
+   * @param path path of a file
+   * @return the corresponding page id, or null if the file name does not match the pattern
+   */
+  @Nullable
+  private PageId getPageId(Path path) {
+    Path parent = path.getParent();
```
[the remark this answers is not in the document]
bf8086: Good idea! Done.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
bf8086: Renamed.

core/client/fs/src/main/java/alluxio/client/file/cache/store/PageStoreOptions.java
bf8086: Updated.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -43,14 +51,38 @@
...
    * @param options options for the local page store
+   * @throws IOException when fails to create a {@link LocalPageStore}
    */
-  public LocalPageStore(LocalPageStoreOptions options) {
+  public LocalPageStore(LocalPageStoreOptions options) throws IOException {
     mRoot = options.getRootDir();
+    mPageSize = options.getPageSize();
+    Path rootDir = Paths.get(mRoot);
+    mPagePattern = Pattern.compile(
+        String.format("%s/%d/(\\d+)/(\\d+)", rootDir.toString(), mPageSize));
```
apc999: `Pattern.quote(rootDir.toString())` to escape the root path?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -43,14 +51,38 @@
...
+          .filter(Files::isRegularFile)
+          .anyMatch(path -> {
+            if (getPageId(path) == null) {
+              return true;
+            }
+            mSize.incrementAndGet();
+            return false;
+          });
+      if (invalidPage || (long) mSize.get() * mPageSize > options.getCacheSize()) {
+        FileUtils.cleanDirectory(new File(mRoot));
```
apc999: `LOG.warn` this page?

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -20,23 +20,34 @@
...
+import java.nio.file.Paths;
+import java.util.Collection;
+import java.util.stream.Stream;
 
 /**
  * A simple abstraction on the storage to put, get and delete pages. The implementation of this
  * class does not need to provide thread-safety.
  */
 public interface PageStore extends AutoCloseable {
+  Logger LOG = LoggerFactory.getLogger(LocalPageStore.class);
```
apc999: LocalPageStore -> PageStore ?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileSystem.java
```diff
@@ -34,9 +37,14 @@ public LocalCacheFileSystem(FileSystem fs, FileSystemContext fsContext) {
   }
 
   @Override
-  public FileInStream openFile(AlluxioURI path, OpenFilePOptions options) {
+  public FileInStream openFile(AlluxioURI path, OpenFilePOptions options)
+      throws IOException, AlluxioException {
     // TODO(calvin): We should add another API to reduce the cost of openFile
+    CacheManager cacheManager = mFsContext.getCacheManager();
+    if (cacheManager == null) {
```
apc999: We should log warn the fallback. However, if you are concerned that the log is too frequent, we can check if `getCacheManager()` equals `null` at the factory method `alluxio.client.file.FileSystem.Factory#create`; in case yes, fallback to returning a normal filesystem instance even `USER_LOCAL_CACHE_ENABLED` is true and `LOG.warn` the fallback?
Then in here, we can simply `Precondition.checkNotNull(cacheManager)`.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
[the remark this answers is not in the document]
apc999: ack

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -69,10 +80,50 @@ static PageStore create(AlluxioConfiguration conf) {
...
+    Path storePath = getStorePath(storeType, rootPath);
+    Files.createDirectories(storePath);
+    LOG.info("Clean cache directory {}", rootPath);
+    try (Stream<Path> stream = Files.list(Paths.get(rootPath))) {
+      stream.filter(path -> !storePath.equals(path))
+          .forEach(path -> {
+            try {
+              FileUtils.deletePathRecursively(path.toString());
+            } catch (IOException e) {
+              LOG.warn("failed to delete {} in cache directory: {}", path, e.getMessage());
```
apc999: same case for https://docs.alluxio.io/os/user/stable/en/contributor/Code-Conventions.html#warn-log-level

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -602,22 +603,29 @@ private void initializeLocalWorker() throws IOException {
...
   private static CacheManager getCacheManager(FileSystemContext fsContext) {
     // TODO(feng): support multiple cache managers
     if (sCacheManager == null) {
       synchronized (FileSystemContext.class) {
         if (sCacheManager == null) {
-          sCacheManager = CacheManager.create(fsContext);
+          try {
+            sCacheManager = Optional.of(CacheManager.create(fsContext));
+          } catch (IOException e) {
+            LOG.warn("Failed to create CacheManager: {}", e.getMessage());
```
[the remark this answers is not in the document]
bf8086: Done.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
bf8086: Good catch! Updated.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
bf8086: Added.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
bf8086: Fixed.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
bf8086: Done.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileSystem.java
bf8086: In `FileSystemContext::getCacheManager`, we already log a warning when the cache manager fails to initialize. Would that be sufficient?

bf8086:
@apc999 Thanks! I updated the PR. PTAL.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -92,7 +93,7 @@
   /**
    * The local cache manager for the file system.
    */
-  private static volatile CacheManager sCacheManager;
+  private static volatile Optional<CacheManager> sCacheManager;
```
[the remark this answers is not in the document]
bf8086: Adding `Optional` here is to represent three different states:
- `null`: cache manager is not initialized
- `Optional.empty()`: cache manager failed to initialize. 
- `Optional.of(instance)`cache manager initialized successfully.

Having the extra state allow us to know that the cache manager initialization is attempted and avoid retrying if it already failed once.

bf8086:

bf8086:

apc999:

PR #10782, opened 2020-01-22
Add metricsMaster.getMetrics() and fix fsadmin report metrics

LuQQiu:
Expose all metrics stored in the leading master via MetricsMasterClient.getMetrics().
Change fsadmin report metrics command to show all metrics.


LuQQiu:
@calvinjia PTAL, thanks!
Move the `metaMaster.getMetrics()` to `metricsMaster.getMetrics()`.
Previously `getMetrics()` only expose a subset of the master instance metrics. Now it exposes all the metrics stored in the leading master including all the master instance metrics and cluster aggregated metrics.

shell/src/main/java/alluxio/cli/fsadmin/command/ReportCommand.java
```diff
@@ -149,7 +149,7 @@ public int run(CommandLine cl) throws IOException {
         break;
       case METRICS:
         MetricsCommand metricsCommand = new MetricsCommand(
-            mMetaClient, mPrintStream);
+            mMetricsClient, mPrintStream);
```
calvinjia: One line?

shell/src/main/java/alluxio/cli/fsadmin/report/MetricsCommand.java
```diff
@@ -120,40 +126,68 @@ public int run() throws IOException {
...
-    printMetric(MasterMetrics.FILES_COMPLETED, "Files Completed", false);
-    printMetric(MasterMetrics.FILES_CREATED, "Files Created", false);
-    printMetric(MasterMetrics.FILES_FREED, "Files Freed", false);
-    printMetric(MasterMetrics.FILES_PERSISTED, "Files Persisted", false);
-    printMetric(MasterMetrics.NEW_BLOCKS_GOT, "New Blocks Got", false);
-    printMetric(MasterMetrics.PATHS_DELETED, "Paths Deleted", false);
-    printMetric(MasterMetrics.PATHS_MOUNTED, "Paths Mounted", false);
-    printMetric(MasterMetrics.PATHS_RENAMED, "Paths Renamed", false);
-    printMetric(MasterMetrics.PATHS_UNMOUNTED, "Paths Unmounted", false);
+    printMetric(MetricsSystem.getMasterMetricName(MasterMetrics.DIRECTORIES_CREATED),
```
calvinjia: There is no easy way to iterate over these? Could we just output the values in the map, that would also avoid the need to expose `getMasterMetricName`?

LuQQiu: Previously we want `fsadmin report metrics` similar to the master web ui metrics page, that's why we add those complicate logic to add `nick name` to those metrics. I agreed that we can modify to directly output contents in the map and maybe sort the metrics according to prefix (master/cluster/worker)

LuQQiu:
@calvinjia I change the fsadmin metrics command to print original metric name instead of changing it to human-readable name to be consistent with web UI. PTAL, thanks!

tests/src/test/java/alluxio/server/ft/journal/TriggeredCheckpointTest.java
```diff
@@ -107,9 +107,9 @@ private void createFiles(MultiProcessCluster cluster, int numFiles)
     for (int i = 0; i < numFiles; i++) {
       fs.createFile(new AlluxioURI("/file" + i)).close();
     }
-    MetaMasterClient meta = cluster.getMetaMasterClient();
-    assertEquals(numFiles + 1,
-        meta.getMetrics().get("Master." + MasterMetrics.TOTAL_PATHS).getLongValue());
+    MetricsMasterClient metricsMasterClient = cluster.getMetricsMasterClient();
+    assertEquals(numFiles + 1, (long) metricsMasterClient
+        .getMetrics().get("Master." + MasterMetrics.TOTAL_PATHS).getDoubleValue());
```
calvinjia: is the cast here correct? Or should we make `numFiles + 1` a double (applies to all cases in tests)

shell/src/main/java/alluxio/cli/fsadmin/report/MetricsCommand.java
```diff
@@ -11,44 +11,44 @@
...
 /**
  * Prints Alluxio metrics information.
  */
 public class MetricsCommand {
+  private static final String BYTES_METRIC_IDENTIFIER = "Bytes";
+  private static final String THROUGHPUT_METRIC_IDENTIFIER = "Throughput";
   private static final DecimalFormat DECIMAL_FORMAT
       = new DecimalFormat("###,###.#####", new DecimalFormatSymbols(Locale.US));
-  private static final String INDENT = "    ";
+  private static final String INFO_FORMAT = "%s  (Type: %s, Value: %s)%n";
```
calvinjia: Do we want to use tabs or something for pretty alignment?

tests/src/test/java/alluxio/server/ft/journal/TriggeredCheckpointTest.java
LuQQiu: The reason that I make `double -> long` is that `assertEquals(double,double)` is deprecated and may have some problems in double comparison.

shell/src/main/java/alluxio/cli/fsadmin/report/MetricsCommand.java
LuQQiu: We have no categories in the new way of printing all the metrics. Previously indent is used for separating category name and its values. So I remove the indent.

LuQQiu:
@calvinjia I change the `fsadmin report metrics` command to show the alluxio metrics only, PTAL, thanks

calvinjia:

PR #10798, opened 2020-01-24
Enable tests after release

gpang:


Xenorith:
we don't need tests because all commits should've gone through master first... right? 😇 

/jk

yuzhu:

PR #10799, opened 2020-01-25
Fix directory index after disk failure

bf8086:
Worker uses directory index to find a directory in a tier. When a disk goes down we detect that and remove the dir, which mess up the index. To avoid the confusion, this fix makes the index absolute in the tier so removing any dir does not alter its index. As a side effect, `getDir(index)` can return null if the corresponding directory is removed. Evictors and Allocators have been updated to handle the null values.

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTierView.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTierView.java
@@ -14,8 +14,9 @@ package alluxio.worker.block.meta;
 import com.google.common.base.Preconditions;
 
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * This class is an abstract class for allocators and evictors to extend to provide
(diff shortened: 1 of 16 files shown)


bf8086:
jenkins, test this please.

bf8086:
jenkins, test this please.

bf8086:
@calvinjia PTAL.

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTierView.java
```diff
@@ -41,7 +42,7 @@ public StorageTierView(StorageTier tier) {
    * @return a list of directory views in this storage tier view
    */
   public List<StorageDirView> getDirViews() {
-    return Collections.unmodifiableList(mDirViews);
+    return new ArrayList<>(mDirViews.values());
```
yuzhu: Does this List need to be modifiable now?

bf8086: No. it's only changed because we now need to convert the value set of the map to a list. It is equivalent to the old unmodifiable list in the sense that user of this API won't be able to change the underlying directory collection stored in this view.

yuzhu: but the user of this API will not get an error if they try to modify this list. In fact they will successfully modify the returned list right?

bf8086: yeah I think the original purpose of wrapping the collection with `unmodifiableList` is to prevent mutation of the view rather than prevent the mutation of the return value after method returns. I don't feel it is necessary to add an unmodifiable wrapper on top of the list now that we already create a copy of the collection. Let me know if you think there is a strong reason for an additional wrapper otherwise.

yuzhu: sounds good

bf8086:

bf8086:

PR #10802, opened 2020-01-28
Fix NPE in GetDatabase

yuzhu:
Location and Comment are optional fields, so we must treat them as such..
Apparently, so is the Owner and OwnerType

### DIFF
--- a/table/server/common/src/main/java/alluxio/master/table/DatabaseInfo.java
+++ b/table/server/common/src/main/java/alluxio/master/table/DatabaseInfo.java
@@ -16,6 +16,7 @@ import alluxio.grpc.table.PrincipalType;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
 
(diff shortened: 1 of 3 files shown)


yuzhu:
jenkins, test this please

table/server/common/src/main/java/alluxio/master/table/DatabaseInfo.java
```diff
@@ -26,7 +26,9 @@
 public class DatabaseInfo {
   @Nullable
   private final String mLocation;
+  @Nullable
   private final String mOwnerName;
+  @Nullable
   private final PrincipalType mOwnerType;
```
ZacBlanco: should the default values in the constructor continue to be `""` and `PrincipalType.USER`? or should they be `null`? what is the difference between `""` and `null` for `mOwnerName`?

SImilar questions for the location and comment

yuzhu:

yuzhu:

PR #10704, opened 2020-01-07
Fix Distributed Load Retry Logic

bradyoo:
Hypothetical Environment:

3 nodes, each with a job worker and a worker.
Suppose a file x.txt has 10 blocks to load but one of its blocks (**Block 1**) is in the middle of being read on (**Worker A**), Attempting to load this file creates 10 tasks and the task to load **Block 1** is assigned to be in **Worker A**.

Previous behavior:

The task in an attempt to load **Block 1** in **Worker A** fails. This fails the job but the other 9 tasks are still continuing. The client notices the failure and triggers a retry. Retry notices that none of the 10 blocks are still loaded yet and shuffles the location of where they are getting loaded.

New behavior:

The task in an attempt to load **Block 1** in **Worker A** fails. This fails the job but the other 9 tasks are still continuing. The client waits for all of its 10 tasks to finish despite knowing that the job has already failed. After all the tasks are complete/failed, it reruns the job again. This time, the server knows that 9 of its blocks are already loaded and only one more needs to get loaded.

PR #10805, opened 2020-01-28
Port changes of [#10704] to branch-2.1

bradyoo:
Hypothetical Environment:

3 nodes, each with a job worker and a worker.
Suppose a file x.txt has 10 blocks to load but one of its blocks
(**Block 1**) is in the middle of being read on (**Worker A**),
Attempting to load this file creates 10 tasks and the task to load
**Block 1** is assigned to be in **Worker A**.

Previous behavior:

The task in an attempt to load **Block 1** in **Worker A** fails. This
fails the job but the other 9 tasks are still continuing. The client
notices the failure and triggers a retry. Retry notices that none of the
10 blocks are still loaded yet and shuffles the location of where they
are getting loaded.

New behavior:

The task in an attempt to load **Block 1** in **Worker A** fails. This
fails the job but the other 9 tasks are still continuing. The client
waits for all of its 10 tasks to finish despite knowing that the job has
already failed. After all the tasks are complete/failed, it reruns the
job again. This time, the server knows that 9 of its blocks are already
loaded and only one more needs to get loaded.

pr-link: #10704
change-id: cid-ac0708727c706624c6b30ede5165b60ed476b7e0

Also removed the active_jobs option because it was needlessly confusing.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
@@ -63,16 +63,6 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
           .build();
 
   private static final int DEFAULT_ACTIVE_JOBS = 1000;
-  private static final Option ACTIVE_JOBS_OPTION =
-      Option.builder()
-          .longOpt("active_jobs")
-          .required(false)
-          .hasArg(true)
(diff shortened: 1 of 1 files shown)


bradyoo:
Changed title and description. @calvinjia

bradyoo:
@calvinjia: Can you merge it for me?

apc999:

PR #10808, opened 2020-01-29
Do not issue persist job if it is already schedule to persist

### LINKED ISSUE #10699: Persist command persist file to s3 underfs second time.
**Alluxio Version:**
2.1.0

**Describe the bug**
I am using Alluxio server with s3 underfs on aws. I am writing output to Alluxio using spark job.
Alluxio server is configure with following setting as describe in doc.
```
# Writes data only to Alluxio before returning a successful write
# Does not persist the data automatically to the underlying storage, this is important because
# only the final committed data is necessary to persist
alluxio.user.file.persistence.initial.wait.time=-1
# Hints that Alluxio should treat renaming as committing data and trigger a persist operation
alluxio.user.file.persist.on.rename=true
# Determines the number of copies in Alluxio when files are not yet persisted, increase this to
# a larger number to ensure fault tolerance in case of Alluxio worker failures
alluxio.user.file.replication.durable=1
# blacklist files which contain the string "_temporary" anywhere in their path
alluxio.master.persistence.blacklist=_temporary
```
Once spark job is completed. Alluxio start writing to s3. Now before Alluxio complete writing to s3 if i issue persist command on ouput dir. It seems that Alluxio start uploading file which is not persisted yet but it is getting uploaded by async process. I have s3 notification enabled on bucket. and I am getting two s3 create "ObjectCreated:CompleteMultipartUpload" notification for the same object. 
I can see persist command writing file s3 second time in output. This issue only occur for file which is not persisted yet.

**To Reproduce**
Already mention in bug description.

**Expected behavior**
Alluxio should write file to s3 underfs only once.

**Urgency**

**Additional context**

pragnesh:
This change fixes #10699

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemUtils.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemUtils.java
@@ -153,8 +153,10 @@ public final class FileSystemUtils {
   public static void persistAndWait(final FileSystem fs, final AlluxioURI uri,
       long persistenceWaitTime, int timeoutMs) throws FileDoesNotExistException, IOException,
       AlluxioException, TimeoutException, InterruptedException {
-    fs.persist(uri, ScheduleAsyncPersistencePOptions
-        .newBuilder().setPersistenceWaitTime(persistenceWaitTime).build());
+    if (!fs.getStatus(uri).getPersistenceState().equals("TO_BE_PERSISTED")) {
+      fs.persist(uri, ScheduleAsyncPersistencePOptions
+              .newBuilder().setPersistenceWaitTime(persistenceWaitTime).build());
+    }
     CommonUtils.waitFor(String.format("%s to be persisted", uri) , () -> {
       try {
         return fs.getStatus(uri).isPersisted();


pragnesh:
PersistCommandTest testcase are passed locally, so i am not sure why it is failing on jenkin. with other test case FileOutStreamAsyncWriteIntegrationTest it seems that LocalAlluxioJobCluster for not scheduling async persist automatically.

core/client/fs/src/main/java/alluxio/client/file/FileSystemUtils.java
```diff
@@ -153,8 +153,10 @@ public static void persistAndWait(final FileSystem fs, final AlluxioURI uri,
   public static void persistAndWait(final FileSystem fs, final AlluxioURI uri,
       long persistenceWaitTime, int timeoutMs) throws FileDoesNotExistException, IOException,
       AlluxioException, TimeoutException, InterruptedException {
-    fs.persist(uri, ScheduleAsyncPersistencePOptions
-        .newBuilder().setPersistenceWaitTime(persistenceWaitTime).build());
+    if (!fs.getStatus(uri).getPersistenceState().equals("TO_BE_PERSISTED")) {
```
gpang: Should this check be done on the server side, since this will now require 2 RPCs for each persist request. Instead, maybe the server side should perform this check?

pragnesh: Thanks @gpang that make sense

yuzhu:
this fix is still open to race conditions. close for now

PR #10809, opened 2020-01-29
Refactor HiveConstant type checking

bradyoo:
Because of types like char(20) and what not, we use startswith instead of equals for some cases and this also prevents us from using switch statements. Standardize the types before checking to enable us to use switch statements.


gpang:

PR #10811, opened 2020-01-29
Remove unneeded logic in worker web

LuQQiu:
The dividend of logical operation and rpc operation should only exist in the master web metrics. For worker web metric, normal operation is enough.

Removed the getRpcInvocationMetrics.

### DIFF
--- a/core/common/src/main/java/alluxio/wire/WorkerWebUIMetrics.java
+++ b/core/common/src/main/java/alluxio/wire/WorkerWebUIMetrics.java
@@ -11,7 +11,6 @@
 
 package alluxio.wire;
 
-import com.codahale.metrics.Counter;
 import com.codahale.metrics.Metric;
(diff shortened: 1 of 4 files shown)


calvinjia:
@LuQQiu thanks for the improvement. 

Unrelated to this PR - the way we split logical and rpc operations seems fairly brittle?

calvinjia:

LuQQiu:
@calvinjia  Yeah, the way we spit logic operation and rpc operation doesn't sound right especially when we adding more counter metrics. 
I am not quite understand the categories (logical vs rpc) as well.
Like CreateDirectoryOps is RPC operation and DirectoriesCreated is logical operation.
They sound to me is more like how many requested and how many succeed. Maybe we can remove the `logical` vs `rpc` operation category and change to RPC operations and succeed operation.s.

PR #10813, opened 2020-01-30
Add a script to update versions for releases

ZacBlanco:
This removes all manual components for a release manager needing to
update versions throughout the codebase. In the future, any part of the
codebase which relies on the version be up-to-date should add a
section in this script that allows the developer to update the section
automatically.


dev/scripts/update-versions.sh
```diff
@@ -0,0 +1,119 @@
...
+    fi
+
+    update_libexec "$_old" "$_new"
+    update_docs "$_old" "$_new"
+    update_k8s "$_old" "$_new"
+
+    exit 0
+}
+
+main "$@"
```
gpang: NIT: newline at end

dev/scripts/update-versions.sh
```diff
@@ -0,0 +1,119 @@
...
+
+
+EOF
+)
+
+# Arguments:
+#  $1: old version
+#  $2: new version
+function update_poms() {
+    find . -name pom.xml | xargs -t -n 1 perl -pi -e "s/${1}/${2}/g"
```
gpang: is `perl` ok to assume? Is that more common than `sed`?

ZacBlanco: As far as my experience has gone, I've found perl to be installed on pretty much every operating system that we develop on (Amazon linux, CentOS, Ubuntu, MacOS) and the interface and arguments that I am using with perl don't vary from distro to distro.

The reason I decided against `sed` is that the default `sed` installed on MacOS has a difference interface than GNU `sed` which is installed on the Linux-based OSes (Ubuntu, CentOS, Amazon Linux) when it comes to in-place editing with the `-i` argument.

[`-i` is not a part of the POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html), so it seems the default used on MacOS doesn't behave the same as on the linux systems.

ZacBlanco: Here are two additional threads I read before deciding to stick with perl

- https://unix.stackexchange.com/questions/92895/how-can-i-achieve-portability-with-sed-i-in-place-editing
- https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

dev/scripts/update-versions.sh
ZacBlanco: fixed

ns1123:
Seems to be missing integrations/Dockerfile

ZacBlanco:
Aware. Still WIP. Have a bunch of changes that I haven't pushed since the last release

ZacBlanco:
Added update for dockerfile

I've used this for 3+ RC cuts now and it works well. I can't speak for final release though

As for editing the URL in the dockerfile - I think it makes more sense to use the SNAPSHOT url, even if it's invalid as it allows the script to actually update the URL properly when building the final release cut

ZacBlanco:
Since I've been able to use this for a couple RCs now, I'm confident that this covers all the cases of versions that need to be updated.

dev/scripts/update-versions.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# Debugging
+# set -x
+set -e # exit if a command fails
+set -u # unbound variable is an error
+
+HELP=$(cat <<EOF
+Usage: update-version.sh [-s] <old version> <new version>
+
+    <old version>    The version to replace throughout the codebase
```
gpang: can you provide an example version string?

ZacBlanco: updated

gpang:
Jenkins, test this please

ZacBlanco:

ZacBlanco:

PR #10816, opened 2020-01-31
Add thread-safe annotation to evictor interface

apc999:


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -221,6 +223,7 @@ public void delete(PageId pageId) throws IOException, PageNotFoundException {
       try (LockResource r1 = new LockResource(mMetaLock.writeLock())) {
         mMetaStore.removePage(pageId);
       }
+      mEvictor.updateOnDelete(pageId);
```
calvinjia: Should we abstract this logic to another method?

1. Grab metalock
2. Do metastore operation
3. Release metalock
4. Notify listeners

apc999: what's the benefit in this case?

calvinjia: Just makes sure that we do all the necessary steps, otherwise we might have cases where we forget to update listeners or grab locks

calvinjia:

PR #10821, opened 2020-01-31
Skip logging error on worker tmp folder

ns1123:
Filter out tmp blocks dir so worker doesn't throw error on start

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDir.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDir.java
@@ -16,6 +16,7 @@ import alluxio.conf.ServerConfiguration;
 import alluxio.exception.BlockAlreadyExistsException;
 import alluxio.exception.BlockDoesNotExistException;
 import alluxio.exception.ExceptionMessage;
+import alluxio.exception.InvalidPathException;
 import alluxio.exception.InvalidWorkerStateException;
 import alluxio.exception.WorkerOutOfSpaceException;
 import alluxio.util.io.FileUtils;
@@ -92,7 +93,8 @@ public final class StorageDir {
    */
   public static StorageDir newStorageDir(StorageTier tier, int dirIndex, long capacityBytes,
       String dirPath, String dirMedium)
-      throws BlockAlreadyExistsException, IOException, WorkerOutOfSpaceException {
+      throws BlockAlreadyExistsException, IOException, WorkerOutOfSpaceException,
+      InvalidPathException {
     StorageDir dir = new StorageDir(tier, dirIndex, capacityBytes, dirPath, dirMedium);
     dir.initializeMeta();
     return dir;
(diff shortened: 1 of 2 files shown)


ns1123:
alluxio-bot, retest this please

ns1123:
alluxio-bot, test this please

ns1123:
It's completely benign, but users are reporting the error as a concern since they're not sure what the implications might be. For the most part, it's essentially cosmetic.

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDir.java
```diff
@@ -125,7 +125,9 @@ private void initializeMeta() throws BlockAlreadyExistsException, IOException,
     }
     for (File path : paths) {
       if (!path.isFile()) {
-        LOG.error("{} in StorageDir is not a file", path.getAbsolutePath());
+        if (!path.getName().equals(ServerConfiguration.get(PropertyKey.WORKER_DATA_TMP_FOLDER))) {
```
ZacBlanco: Is the intention to _only_ filter out this tmp folder, or any blocks within it? Because as it, won't this only filter out just that single folder?

ns1123: There should really only be one object in that directory that isn't a file which is the .tmp_blocks directory. Listing the objects underneath is not needed since this directory will get deleted anyways. The main point is to avoid logging the ERROR since it has/will alarm certain users.

ZacBlanco: My last comment is how the value of the `WORKER_DATA_TMP_FOLDER` might match `path.getName()` - can the user specify a relative path? will it still match `path.getName()` if the config specifies a relative or absolute?

ns1123: WORKER_DATA_TMP_FOLDER must be a relative path. We don't allow an absolute path for this property. Path.getName() will also only get the last index of the path so this match should be correct.

ZacBlanco: couldn't someone specify `path1/path2` - configuration would return that, but then the `path.getName()` would only have `path2`?

ns1123: I think path.getName() would only have path1, but I see your concern. Is this something we reasonably see happening?

ZacBlanco: If the goal is to remove the message we need to account for the property being configured to something other than the default

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDir.java
```diff
@@ -109,11 +111,11 @@ public static StorageDir newStorageDir(StorageTier tier, int dirIndex, long capa
    * @throws WorkerOutOfSpaceException when metadata can not be added due to limited left space
    */
   private void initializeMeta() throws BlockAlreadyExistsException, IOException,
-      WorkerOutOfSpaceException {
+      WorkerOutOfSpaceException, InvalidPathException {
     // Create the storage directory path
     boolean isDirectoryNewlyCreated = FileUtils.createStorageDirPath(mDirPath,
         ServerConfiguration.get(PropertyKey.WORKER_DATA_FOLDER_PERMISSIONS));
-
+    String tmpDir = ServerConfiguration.get(PropertyKey.WORKER_DATA_TMP_FOLDER).split("/")[0];
```
ZacBlanco: can we use a method from `PathUtils` instead?

ns1123: Done as per your suggestion.

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTier.java
```diff
@@ -100,7 +100,7 @@ private void initStorageTier()
             dirMedium[mediumTypeindex]);
         totalCapacity += capacity;
         mDirs.put(i, dir);
-      } catch (IOException e) {
+      } catch (IOException | InvalidPathException e) {
```
ZacBlanco: is this still necessary?

ZacBlanco:

PR #10826, opened 2020-01-31
Replace Source Code Pro with Anonymous Pro font

ZacBlanco:
The reason I'm proposing replacing SCP with AP is because there is a
bug in  firefox and/or MacOS which causes this  font to render
incorrectly.

See:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1534503
- https://bugzilla.mozilla.org/show_bug.cgi?id=1520157

Anonymous Pro is a bit more lightweight and has at least  been recently
updated  by its maintainers (1.44.0 was 3 months ago). SCP is definitely
more ubiquitous  and well-maintained but I would rather have a font that
works everywhere. As an added bonus, this font is *much* more
lightweight than SCP.

Additionally, it seems the maintainers of the project have no desire to
address or look into what triggered the regression:
https://github.com/adobe-fonts/source-code-pro/issues/217

### DIFF
--- a/webui/common/package.json
+++ b/webui/common/package.json
@@ -8,6 +8,7 @@
     "@fortawesome/free-regular-svg-icons": "5.7.2",
     "@fortawesome/free-solid-svg-icons": "5.7.2",
     "@fortawesome/react-fontawesome": "0.1.4",
+    "@openfonts/anonymous-pro_all": "^1.44.0",
     "bootstrap": "4.3.1",
     "bootswatch": "4.3.1",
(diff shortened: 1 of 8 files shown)


ZacBlanco:
Here's a source with some samples of the font: https://www.marksimonson.com/fonts/view/anonymous-pro

ZacBlanco:
I also wanted to use this font, but there doesn't seem to be a well-maintained package on the npm registry: https://typeof.net/Iosevka/

calvinjia:

PR #10828, opened 2020-02-01
Fix gRPC exception handling and improve traces

ggezer:
Client level retries were sometimes ineffective due to narrow handling of exceptions.

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -94,6 +94,10 @@ public final class GrpcChannel extends Channel {
 
   /**
    * Shuts down the channel.
+   *
+   * Shutdown should be synchronized as it could be called concurrently due to:
+   *  - Authentication long polling
+   *  - gRPC messaging stream.
    */
   public synchronized void shutdown() {
     if (mAuthDriver != null) {
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -95,7 +95,7 @@ public final class GrpcChannel extends Channel {
   /**
    * Shuts down the channel.
    */
-  public void shutdown() {
+  public synchronized void shutdown() {
     if (mAuthDriver != null) {
       // Close authenticated session with server.
       mAuthDriver.close();
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -131,7 +131,7 @@ public final class GrpcChannel extends Channel {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper("")
+    return MoreObjects.toStringHelper(this)
         .add("ChannelKey", mChannelKey)
         .add("ChannelHealthy", isHealthy())
         .add("ChannelReleased", mChannelReleased)
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannel.java
@@ -127,7 +127,7 @@ public final class GrpcChannel extends Channel {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this)
+    return MoreObjects.toStringHelper("")
         .add("ChannelKey", mChannelKey)
         .add("ChannelHealthy", isHealthy())
         .add("ChannelReleased", mChannelReleased)
--- a/core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
@@ -288,7 +288,7 @@ public class GrpcChannelKey {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper("")
+    return MoreObjects.toStringHelper(this)
         .add("ClientType", getStringFromOptional(mClientType))
         .add("ServerAddress", mServerAddress)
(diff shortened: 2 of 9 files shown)


ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
please jenkins, test this please

ggezer:
thank you jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
@gpang This seems back to usual flakiness. I'll work more on some flakes, but I believe this is good to go for alleviating PR failures.

ggezer:
jenkins, test this please

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
```diff
@@ -127,7 +127,7 @@ public boolean isHealthy() {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this)
+    return MoreObjects.toStringHelper("")
```
gpang: why is this (and others) ""?

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
```diff
@@ -95,7 +95,7 @@ public void intercept(ClientInterceptor interceptor) {
   /**
    * Shuts down the channel.
    */
-  public void shutdown() {
+  public synchronized void shutdown() {
```
gpang: could you add a comment in the code to explain why this has to be synchronized?

core/common/src/main/java/alluxio/security/authentication/AbstractSaslClientHandler.java
```diff
@@ -85,10 +88,11 @@ public SaslMessage handleMessage(SaslMessage message) throws SaslException {
   }
 
   @Override
-  public void close() {
+  public synchronized void close() {
     if (mSaslClient != null) {
       try {
         mSaslClient.dispose();
+        mSaslClient = null;
```
gpang: do we want to keep this non-null if there is an issue with `dispose`? Basically, should we set it to null in finally or here?

core/common/src/main/java/alluxio/security/authentication/AbstractSaslClientHandler.java
```diff
@@ -47,7 +47,10 @@ public AbstractSaslClientHandler(ChannelAuthenticationScheme authScheme) {
   }
 
   @Override
-  public SaslMessage handleMessage(SaslMessage message) throws SaslException {
+  public synchronized SaslMessage handleMessage(SaslMessage message) throws SaslException {
```
gpang: can you leave a comment as to why this and others have to be synchronized? (seems like it is for `mSaslClient`?)

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
ggezer: When given an instance, the string generated will start with that class's name. (Something like "GrpcChannel { Host=....... }". Giving empty header makes the traces much cleaner.

core/common/src/main/java/alluxio/security/authentication/AbstractSaslClientHandler.java
ggezer: We don't want to keep it set if dispose fails. So in finally is better.

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
gpang: We want to remove the class name? I would have thought the class name is useful?

ggezer: Yeah, I think it's better for this particular one. For the rest, empty is better.

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -288,7 +288,7 @@ public boolean equals(Object other) {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this)
+    return MoreObjects.toStringHelper("")
```
gpang: I am confused when this is supposed to be "" or `this`?

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -305,7 +305,7 @@ public String toString() {
    * @return short representation of this channel key
    */
   public String toStringShort() {
-    return MoreObjects.toStringHelper("Channel")
+    return MoreObjects.toStringHelper("")
```
gpang: I am confused when this is supposed to be "" or `this`?

core/common/src/main/java/alluxio/grpc/GrpcServer.java
```diff
@@ -135,7 +135,7 @@ public boolean isServing() {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this)
+    return MoreObjects.toStringHelper("")
```
gpang: I am confused when this is supposed to be "" or `this`?

core/common/src/main/java/alluxio/grpc/GrpcServerAddress.java
```diff
@@ -86,7 +86,7 @@ public boolean equals(Object other) {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this)
+    return MoreObjects.toStringHelper("")
```
gpang: I am confused when this is supposed to be "" or `this`?

ggezer:
jenkins, test this please

ggezer:
alluxio-bot, test this please

ggezer:

PR #10834, opened 2020-02-03
Support compiling with Java 11

### LINKED ISSUE #9856: Support Java 11
Currently, Alluxio only supports Java 8

Java 11 is the next LTS release of Java.  

> The end of public updates for Java 8 is scheduled for January of 2019 for commercial use. For non-commercial use, the same is scheduled at an unspecified date in December of 2020.

Current Alluxio 2.1-SNAPSHOT does not compile with JDK 11.
We should fix the src code so it works with either JDK 8 and JDK 11

**Urgency**
Medium

**Additional context**

ZacBlanco:
Use maven 3.6+ with JDK 11 and the flag -Pjava11 to enable the Java 11
profile. However, maven should activate the profile by default if it detects
your default is JDK 11

There is a long list of changes and incompatibilities between Java 8
and Java 11, only a few affected us for the most part though.

- The `jdk.tools` system dependency which is transitively depended
upon by `org.apache.hbase:hbase-annotations` no longer exists in
Java 11, so we exclude `jdk.tools:jdk.tools` throughout the codebase.
- The javax annotations are no longer a part of the JDK distribution,
but they are available as a backwards-compatible maven dependency
so we add that if we're compiling with Java 11.
- The swagger maven plugin which is used to generate the REST API
docs depends on a JAXB class which isn't a part of the JDK in Java 11
so we add that dependency for the plugin to function properly.
- The maven javadoc plugin has incompatibilities for docs in a certain
format from Java 8. Configure the plugin with `<source>8</source>`
to tell the plugin the sources use some Java 8 formats. See this
OpenJDK bug for more info:
https://bugs.openjdk.java.net/browse/JDK-8212233

Partially addresses #9856

### DIFF
--- a/table/server/underdb/hive/pom.xml
+++ b/table/server/underdb/hive/pom.xml
@@ -50,6 +50,7 @@
           <groupId>org.apache.logging.log4j</groupId>
           <artifactId>log4j-slf4j-impl</artifactId>
         </exclusion>
+
       </exclusions>
     </dependency>
 
@@ -124,4 +125,32 @@
       </plugin>
     </plugins>
   </build>
+
+  <profiles>
+    <profile>
+      <id>java11</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.hive</groupId>
+          <artifactId>hive-metastore</artifactId>
+          <version>${hive-metastore.version}</version>
+          <scope>compile</scope>
+          <exclusions>
+            <exclusion>
+              <groupId>org.apache.logging.log4j</groupId>
+              <artifactId>log4j-1.2-api</artifactId>
+            </exclusion>
+            <exclusion>
+              <groupId>org.apache.logging.log4j</groupId>
+              <artifactId>log4j-slf4j-impl</artifactId>
+            </exclusion>
+            <exclusion>
+              <groupId>jdk.tools</groupId>
+              <artifactId>jdk.tools</artifactId>
+            </exclusion>
+          </exclusions>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
 </project>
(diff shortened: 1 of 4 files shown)


ZacBlanco:
I tested this on my local machine using a script to run the build in a docker container:

`build.sh`

```bash
#!/usr/bin/env bash
#
# The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
# (the "License"). You may not use this work except in compliance with the License, which is
# available at www.apache.org/licenses/LICENSE-2.0
#
# This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
# either express or implied, as more fully set forth in the License.
#
# See the NOTICE file distributed with this work for information regarding copyright ownership.
#

docker run -it --rm --name alluxio-java-11 \
  -v "$(pwd)":/usr/src/mymaven \
  -v ~/.m2:/root/.m2 \
  -w /usr/src/mymaven \
  maven:3.6.3-jdk-11 \
  mvn $@

```

table/server/underdb/hive/pom.xml
```diff
@@ -50,6 +50,7 @@
           <groupId>org.apache.logging.log4j</groupId>
           <artifactId>log4j-slf4j-impl</artifactId>
         </exclusion>
+
```
gpang: remove extra line?

ZacBlanco:
@gpang I've updated the description with the issues I hit and how they were resolved

table/server/underdb/hive/pom.xml
ZacBlanco: fixed

ZacBlanco:
Good catch. I'm not sure how much it matters. I was able to compile and run with java 11 without setting that property. I will add that to the profile though.

ZacBlanco:
jenkins, test this please

ZacBlanco:
Just want to add a note that if you build with java 11 locally, many of the unit and integration tests will *not* work - this simply gets the compilation working.

ZacBlanco:

PR #10835, opened 2020-02-04
Refactor for local cache

calvinjia:
Better division for the file system configuration and constructors.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
@@ -130,9 +130,17 @@ public interface FileSystem extends Closeable {
      * @return a new FileSystem instance
      */
     public static FileSystem create(FileSystemContext context) {
+      return create(context, context.getClusterConf());
+    }
+
+    /**
+     * @param context the FileSystemContext to use with the FileSystem
+     * @param conf the Alluxio configuration
+     * @return a new FileSystem instance
+     */
+    public static FileSystem create(FileSystemContext context, AlluxioConfiguration conf) {
       if (LOG.isDebugEnabled() && !CONF_LOGGED.getAndSet(true)) {
         // Sort properties by name to keep output ordered.
-        AlluxioConfiguration conf = context.getClusterConf();
         List<PropertyKey> keys = new ArrayList<>(conf.keySet());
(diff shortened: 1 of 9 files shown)


core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -491,9 +491,13 @@ public synchronized void initialize(URI uri, org.apache.hadoop.conf.Configuratio
     // Create FileSystem for accessing Alluxio.
     // Disable URI validation for non-Alluxio schemes.
     boolean disableUriValidation =
-        (uri.getScheme() != null) ? uri.getScheme().equals(Constants.SCHEME) : true;
-    mFileSystem = FileSystem.Factory.create(
-        ClientContext.create(subject, alluxioConf).setUriValidationEnabled(disableUriValidation));
+        (uri.getScheme() == null) || uri.getScheme().equals(Constants.SCHEME);
+    if (alluxioConf.getBoolean(PropertyKey.USER_LOCAL_CACHE_LIBRARY)) {
+      mFileSystem = FileSystem.Factory.create(null, alluxioConf);
```
bf8086: Would it be okay now that the file system client does not have the client user subject?

core/client/fs/src/main/java/alluxio/client/file/FileSystem.java
```diff
@@ -130,9 +130,17 @@ public static FileSystem create(ClientContext ctx) {
      * @return a new FileSystem instance
      */
     public static FileSystem create(FileSystemContext context) {
+      return create(context, context.getClusterConf());
```
bf8086: Would it be cleaner if we just have a subclass of `FileSystemContext` that override `getClusterConf` to return the client side configuration instead?

calvinjia: `FileSystemContext` wasn't written to be easily extended, and even if we had a subclass we would still need a switch somewhere to make that FileSystemContext version

core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
calvinjia: that shouldn't be necessary in library mode

calvinjia:

PR #10836, opened 2020-02-04
Update table metadata from changes in UDB

gpang:
This allows the sync command to update the metadata of existing partitions, not only new partitions. This means, if an existing partition is updated in hive, it will be detected and updated during a sync command.

For this functionality, table version numbers and partition version numbers have been added. The concurrency model was also updated for these changes. It is an optimistic concurrency control model, where multiple updates can be concurrent, but only 1 will succeed in updating the metadata to the next version.


gpang:
@yuzhu

gpang:

Issue #10842, opened 2020-02-04
Remove unnecessary OSS logging due to deleteFileIfExists


apc999:
**Alluxio Version:**
2.1.1

**Describe the bug**

Excessive unnecessary logging on checking object existence due to `deleteFileIfExists` using OSS

```
2020-02-04 22:02:03,816 WARN  OSSUnderFileSystem - Failed to get Object io-test/.alluxio_ufs_blocks.alluxio.0x1D91AC0E01AB0165.tmp/234377707520, return null
com.aliyun.oss.OSSException: Not Found
[ErrorCode]: NoSuchKey
[RequestId]: 5E39E9DBB766D33436D131A4
[HostId]: null
        at com.aliyun.oss.common.utils.ExceptionFactory.createOSSException(ExceptionFactory.java:105)
        at com.aliyun.oss.internal.OSSErrorResponseHandler.handle(OSSErrorResponseHandler.java:56)
        at com.aliyun.oss.common.comm.ServiceClient.handleResponse(ServiceClient.java:257)
        at com.aliyun.oss.common.comm.ServiceClient.sendRequestImpl(ServiceClient.java:140)
        at com.aliyun.oss.common.comm.ServiceClient.sendRequest(ServiceClient.java:70)
        at com.aliyun.oss.internal.OSSOperation.send(OSSOperation.java:83)
        at com.aliyun.oss.internal.OSSOperation.doOperation(OSSOperation.java:145)
        at com.aliyun.oss.internal.OSSObjectOperation.getObjectMetadata(OSSObjectOperation.java:496)
        at com.aliyun.oss.OSSClient.getObjectMetadata(OSSClient.java:599)
        at com.aliyun.oss.OSSClient.getObjectMetadata(OSSClient.java:589)
        at alluxio.underfs.oss.OSSUnderFileSystem.getObjectStatus(OSSUnderFileSystem.java:243)
        at alluxio.underfs.ObjectUnderFileSystem.isFile(ObjectUnderFileSystem.java:588)
        at alluxio.underfs.UnderFileSystemWithLogging$30.call(UnderFileSystemWithLogging.java:543)
        at alluxio.underfs.UnderFileSystemWithLogging$30.call(UnderFileSystemWithLogging.java:540)
        at alluxio.underfs.UnderFileSystemWithLogging.call(UnderFileSystemWithLogging.java:949)
[...]

PR #10843, opened 2020-02-04
Remove unnecessary logging in under storage

apc999:
Fix https://github.com/Alluxio/alluxio/issues/10842 but not only in OSS but a few other under storage to leave the server logs cleaner.

It is unnecessary to log again on non-existence of an object as the null value is already indicating the scenario.  Besides, the non-existence may be expected (e.g. when called from `deleteFileIfExists`


calvinjia:

PR #10845, opened 2020-02-05
Choose HDFS read API based on Client buffer size

madanadit:
This PR is a follow up to https://github.com/Alluxio/alluxio/pull/10763.

When the client buffer size is large ( > 2MB) and reads are guaranteed to be somewhat sequential, the `pread` API to HDFS is not as efficient as simple `read`. We introduce a heuristic to choose which API to use.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
@@ -234,9 +234,8 @@ public class AlluxioFileInStream extends FileInStream {
       return -1;
     }
 
-    LOG.info("AMDEBUG pread len {} ", len);
-    if (len < 2 * Constants.MB) {
-      LOG.info("AMDEBUG set for pread len {} ", len);
+    if (len < mContext.getPathConf(new AlluxioURI(mStatus.getPath()))
+        .getBytes(PropertyKey.USER_FILE_SEQUENTIAL_PREAD_THRESHOLD)) {
       mOptions.setPositionShort(true);
     }
     int lenCopy = len;
--- a/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
@@ -12,6 +12,7 @@
 package alluxio.client.file;
 
 import alluxio.AlluxioURI;
+import alluxio.Constants;
 import alluxio.annotation.PublicApi;
 import alluxio.client.ReadType;
 import alluxio.client.block.AlluxioBlockStore;
@@ -233,6 +234,9 @@ public class AlluxioFileInStream extends FileInStream {
       return -1;
     }
 
+    if (len > 1 * Constants.MB) {
+      mOptions.setSwitch(true);
+    }
     int lenCopy = len;
     CountingRetry retry = new CountingRetry(mBlockWorkerClientReadRetry);
     IOException lastException = null;
--- a/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
@@ -234,7 +234,9 @@ public class AlluxioFileInStream extends FileInStream {
       return -1;
     }
 
+    LOG.info("AMDEBUG pread len {} ", len);
     if (len > 1 * Constants.MB) {
+      LOG.info("AMDEBUG set switch for pread len {} ", len);
       mOptions.setSwitch(true);
     }
     int lenCopy = len;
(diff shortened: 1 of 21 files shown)


madanadit:
@calvinjia Early comments requested. The performance seems to workout for large buffer sizes with this PR. 

A couple of open questions:
1. Should the threshold be made a mount option? The threshold should depend on how remote the HDFS is, for slower HDFS I could imagine a lower threshold could work better. The disadvantage of this over the current client option is that we cannot change this on the fly (with path defaults)
2. If using a mount option, instead of comparing against the threshold on the client side, I would then pass in `pread_buffer_size` as part of `ReadRequest`, propagate to the UFS  and  evaluate the condition in the worker ufs

Let me know your thoughts.

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
```diff
@@ -233,6 +234,11 @@ private int positionedReadInternal(long pos, byte[] b, int off, int len) throws
       return -1;
     }
 
+    LOG.info("AMDEBUG pread len {} ", len);
+    if (len < 2 * Constants.MB) {
+      LOG.info("AMDEBUG set for pread len {} ", len);
+      mOptions.setPositionShort(true);
```
calvinjia: Does this option only apply to positioned reads or also reads through the `read` API?

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
```diff
@@ -233,6 +234,11 @@ private int positionedReadInternal(long pos, byte[] b, int off, int len) throws
       return -1;
     }
 
+    LOG.info("AMDEBUG pread len {} ", len);
+    if (len < 2 * Constants.MB) {
```
calvinjia: I think it is good to have our heuristic to cover the typical case, but should we check if it has been set to allow the user to override?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
```diff
@@ -563,10 +563,21 @@ public InputStream open(String path, OpenOptions options) throws IOException {
...
-        // pread API instead of seek is more efficient for FSDataInputStream.
-        // A seek on FSDataInputStream uses a skip op which is implemented as read + discard
-        // and hence ends up reading extra data from the datanode.
-        return new HdfsPositionedUnderFileInputStream(inputStream, options.getOffset());
+        if (options.getPositionShort()) {
+          LOG.info("AMDEBUG: using pread");
+          // pread API instead of seek is more efficient for FSDataInputStream.
+          // A seek on FSDataInputStream uses a skip op which is implemented as read + discard
+          // and hence ends up reading extra data from the datanode.
+          return new HdfsPositionedUnderFileInputStream(inputStream, options.getOffset());
```
calvinjia: Do we end up caching this stream anywhere, if so would it be a problem if it is a certain type?

calvinjia:
@madanadit I think a user side option is more flexible and we can have a server side default instead of the client side default if we want it to be more user friendly.

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
madanadit: only applies to `pread`. reads through the `read` API always are assumed to be sequential by this heuristic

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
madanadit: yep, i'll make this a client config which can be overridden by the user

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
madanadit: I just grepped our codebase, doesn't look like we do, so we should be fine

tests/src/test/java/alluxio/client/fs/FileInStreamIntegrationTest.java
```diff
@@ -435,6 +435,44 @@ public void positionedReadWithoutCaching() throws Exception {
...
+      String filename = mTestPath + "/file_" + MIN_LEN + "_" + op.hashCode();
+      AlluxioURI uri = new AlluxioURI(filename);
+
+      try (FileInStream is = mFileSystem.openFile(uri, FileSystemTestUtils.toOpenFileOptions(op))) {
+        byte[] ret = new byte[DELTA - 1];
+        Assert.assertEquals(DELTA - 1, is.positionedRead(MIN_LEN - DELTA + 1, ret, 0, DELTA));
+        Assert.assertTrue(BufferUtils.equalIncreasingByteArray(MIN_LEN - DELTA + 1, DELTA - 1, ret));
+      }
+    }
+  }
```
calvinjia: Is there a way to verify if the PRead path is utilized?

madanadit:
jenkins, test this please.

madanadit:

PR #10846, opened 2020-02-05
Update check periodically

LuQQiu:
Change update check to check every 7 days instead of once when master starts.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
@@ -306,7 +306,7 @@ public final class DefaultMetaMaster extends CoreMaster implements MetaMaster {
             ThreadFactoryUtils.build("DailyMetadataBackup-%d", true)), mUfsManager);
         mDailyBackup.start();
       }
-      if (mState.getClusterID() == INVALID_CLUSTER_ID) {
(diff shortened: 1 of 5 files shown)


core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -306,26 +306,18 @@ public void start(Boolean isPrimary) throws IOException {
...
-              System.out.println("The latest version (" + latestVersion + ") is not the same "
-                  + "as the current version (" + ProjectConstants.VERSION + "). To upgrade "
-                  + "visit https://www.alluxio.io/download/.");
-              mNewerVersionAvailable = true;
-            }
-          } catch (Exception e) {
-            LOG.debug("Unable to check for updates: {}", e.getMessage());
-          }
+          getExecutorService().submit(new HeartbeatThread(HeartbeatContext.MASTER_UPDATE_CHECK,
+              new UpdateChecker(this, mState.getClusterID()),
```
madanadit: `DefaultMetaMaster` already has a method `getClusterID`? we can remove the 2nd parameter to the constructor

LuQQiu: Good advice~!!!

madanadit:

PR #10847, opened 2020-02-05
Make page calculation in bytes on eviction

apc999:


core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
```diff
@@ -13,29 +13,39 @@
...
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * The default implementation of a metadata store for pages stored in cache.
  */
 public class DefaultMetaStore implements MetaStore {
-  private final Set<PageId> mPageMap = new HashSet<>();
+  /** A map from PageId to page size. */
+  private final Map<PageId, Long> mPageMap = new HashMap<>();
```
calvinjia: Do we want hashmap or just a page object that has the size?

calvinjia: Actually we may need to store more page information later, so having a page info kind of object makes sense, for now we can keep it a long.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -120,25 +119,31 @@ private ReadWriteLock getPageLock(PageId pageId) {
...
         alreadyCached = mMetaStore.hasPage(pageId);
-        if (!alreadyCached) {
-          needEvict = mPageStore.size() + 1 > mCacheSize;
+        if (alreadyCached) {
+          pageSize = mMetaStore.getPageSize(pageId);
+        } else {
+          needEvict = mPageStore.bytes() + page.length > mCacheSize;
           if (needEvict) {
             victim = mEvictor.evict();
+            victimSize = mMetaStore.getPageSize(victim);
```
calvinjia: I think we need a loop here where we evict data, since there is no guarantee one victim is sufficient?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -120,25 +119,31 @@ private ReadWriteLock getPageLock(PageId pageId) {
...
+      } catch (PageNotFoundException e) {
+        throw new IllegalStateException("we shall not reach here");
       }
       if (alreadyCached) {
         try {
-          mPageStore.delete(pageId);
+          mPageStore.delete(pageId, pageSize);
         } catch (PageNotFoundException e) {
           throw new IllegalStateException(
               String.format("Page store is missing page %s.", pageId), e);
```
calvinjia: Unrelated question - in some places we do operation first then evictor update and in others we do evictor update then operation, what is the reasoning?

apc999: that is unintentional. Let's fix that issue in https://github.com/Alluxio/alluxio/pull/10816/

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: As discussed, we will make best effort and evict at most one page to keep put concurrent

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: fixed in later commit. PTAL

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -120,38 +119,32 @@ private ReadWriteLock getPageLock(PageId pageId) {
...
       try (LockResource r2 = new LockResource(mMetaLock.writeLock())) {
-        alreadyCached = mMetaStore.hasPage(pageId);
-        if (!alreadyCached) {
-          needEvict = mPageStore.size() + 1 > mCacheSize;
-          if (needEvict) {
-            victim = mEvictor.evict();
-          } else {
-            mMetaStore.addPage(pageId);
-          }
+        if (mMetaStore.hasPage(pageId)) {
```
apc999: I changed the behavior here a bit: if page to insert exists, it returns false now

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -205,8 +204,9 @@ public ReadableByteChannel get(PageId pageId, int pageOffset)
       if (!hasPage) {
         return null;
       }
+      ReadableByteChannel ret = mPageStore.get(pageId, pageOffset);
```
apc999: now we consistently update page store first followed by updating evictor

apc999:
@calvinjia PTAL

core/client/fs/src/test/java/alluxio/client/file/cache/LocalCacheManagerTest.java
```diff
@@ -54,66 +51,148 @@
...
 
   @Before
   public void before() {
     mConf.set(PropertyKey.USER_CLIENT_CACHE_PAGE_SIZE, PAGE_SIZE_BYTES);
     mConf.set(PropertyKey.USER_CLIENT_CACHE_SIZE, CACHE_SIZE_BYTES);
-    mMetaStore = new HashSetMetaStore();
-    mPageStore = new HashMapPageStore();
+    mConf.set(PropertyKey.USER_CLIENT_CACHE_DIR, mTemp.getRoot().getAbsolutePath());
+    mMetaStore = MetaStore.create();
+    mPageStore = PageStore.create(mConf);
```
apc999: Let's test with real implementation of MetaStore and PageStore instead the mock. FYI

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
```diff
@@ -13,29 +13,39 @@
...
   @Override
   public boolean hasPage(PageId pageId) {
-    return mPageMap.contains(pageId);
+    return mPageMap.containsKey(pageId);
   }
 
   @Override
-  public boolean addPage(PageId pageId) {
-    return mPageMap.add(pageId);
+  public void addPage(PageId pageId, PageInfo pageInfo) {
```
apc999: addPage requires PageInfo now.

bf8086: Do we still need to pass in separate page id now that PageInfo has the id?

calvinjia:

PR #10853, opened 2020-02-06
Improve performance of MetricsHeartbeatContextTest

ZacBlanco:
The test currently runs for almost 3 minutes due to client retry
configuration. This change makes the PollingMasterInquireClient
respect the USER_RPC_RETRY* configuration values which
subsequently speeds up the test from ~3min to ~5sec

### DIFF
--- a/core/common/src/main/java/alluxio/master/PollingMasterInquireClient.java
+++ b/core/common/src/main/java/alluxio/master/PollingMasterInquireClient.java
@@ -23,8 +23,8 @@ import alluxio.grpc.GrpcChannelBuilder;
 import alluxio.grpc.GrpcServerAddress;
 import alluxio.grpc.ServiceType;
 import alluxio.grpc.ServiceVersionClientServiceGrpc;
-import alluxio.retry.ExponentialBackoffRetry;
 import alluxio.retry.RetryPolicy;
+import alluxio.retry.RetryUtils;
 import alluxio.security.user.UserState;
 import alluxio.uri.Authority;
 import alluxio.uri.MultiMasterAuthority;
@@ -61,7 +61,10 @@ public class PollingMasterInquireClient implements MasterInquireClient {
   public PollingMasterInquireClient(List<InetSocketAddress> masterAddresses,
       AlluxioConfiguration alluxioConf,
       UserState userState) {
-    this(masterAddresses, () -> new ExponentialBackoffRetry(20, 2000, 30),
+    this(masterAddresses, () -> RetryUtils.defaultClientRetry(
+        alluxioConf.getDuration(PropertyKey.USER_RPC_RETRY_MAX_DURATION),
+        alluxioConf.getDuration(PropertyKey.USER_RPC_RETRY_BASE_SLEEP_MS),
+        alluxioConf.getDuration(PropertyKey.USER_RPC_RETRY_MAX_SLEEP_MS)),
         alluxioConf, userState);
   }
 
--- a/core/client/fs/src/test/java/alluxio/client/metrics/MetricsHeartbeatContextTest.java
+++ b/core/client/fs/src/test/java/alluxio/client/metrics/MetricsHeartbeatContextTest.java
@@ -120,7 +120,9 @@ public class MetricsHeartbeatContextTest {
 
     ScheduledFuture<?> future = Mockito.mock(ScheduledFuture.class);
     when(future.cancel(any(Boolean.class))).thenReturn(true);
-    ClientContext ctx = ClientContext.create();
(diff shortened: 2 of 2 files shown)


ZacBlanco:
@gpang FYI more test improvements

ZacBlanco:
jenkins, test this please

ZacBlanco:
These are the difference in maven build times when running `mvn test -Dsurefire.forkCount=12` before and after this change:

Before:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Alluxio Core - Client 2.2.0-SNAPSHOT ............... SUCCESS [  2.902 s]
[INFO] Alluxio Core - Client - File System ................ SUCCESS [03:31 min]
[INFO] Alluxio Core - Client - HDFS 2.2.0-SNAPSHOT ........ SUCCESS [ 10.336 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 03:45 min
[INFO] Finished at: 2020-02-05T19:32:54-08:00
[INFO] ------------------------------------------------------------------------
```

After:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Alluxio Core - Client 2.2.0-SNAPSHOT ............... SUCCESS [  2.711 s]
[INFO] Alluxio Core - Client - File System ................ SUCCESS [ 40.926 s]
[INFO] Alluxio Core - Client - HDFS 2.2.0-SNAPSHOT ........ SUCCESS [ 11.513 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 55.316 s
[INFO] Finished at: 2020-02-05T19:27:15-08:00
[INFO] ------------------------------------------------------------------------
```

gpang:
hrmmm, on the PR build, it is still 3+ minutes. Do you know why that would be?
```
[INFO] Alluxio Core - Client - File System ................ SUCCESS [03:15 min]
```

ZacBlanco:
It's possible that the time to build the module is still high because during that phase of the build I believe there are 6-7 modules running concurrently which are probably in contention for CPU time.

Specifically, if you look for the test I set out to fix, it is much lower now:
```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 9.575 sec - in 
alluxio.client.metrics.MetricsHeartbeatContextTest
```

ZacBlanco:
jenkins, test this please

core/common/src/main/java/alluxio/master/PollingMasterInquireClient.java
```diff
@@ -61,7 +61,10 @@
   public PollingMasterInquireClient(List<InetSocketAddress> masterAddresses,
       AlluxioConfiguration alluxioConf,
       UserState userState) {
-    this(masterAddresses, () -> new ExponentialBackoffRetry(20, 2000, 30),
+    this(masterAddresses, () -> RetryUtils.defaultClientRetry(
+        alluxioConf.getDuration(PropertyKey.USER_RPC_RETRY_MAX_DURATION),
```
gpang: How does changing this affect other tests? Will it still behave similarly to the previous retry policy?

ZacBlanco: The old way has an exponential backoff max of 2s * 30 total attempts. It takes 7 retries before hitting the 2s cap. This results in a max wait time of 46s + ~3s = ~50s. The values from PK defaults result in a max of 2min before stopping.

This is the profile for time between retries for each

image

This is a graph for total time spent vs # of retries

image

The wait time maxes out much higher for the polling client when using the default RPC retries

Equations:
image


Most cases there won't be a difference. The worst thing that will happen is that it results in users needing to wait longer for the polling client to determine who the leader master is. I can add more PKs so that the retry profile for the polling client match more closely with the current values than using the generic RPC retry parameters.

@ggezer I requested your review for the reason stated above. Do you think there are other repercussions to changing the profile? Should we try to keep the values what they are now?

ggezer: As discussed offline, fail-over scenario is not effected as the tight-loop of finding a responding master address won't be done under retry policy. Quick swipe on code shows no usage of inquire clients outside of RPC context, so I think It's fine to reuse these parameters.

ZacBlanco:
@gpang any more comments?

ZacBlanco:

PR #10856, opened 2020-02-06
Allow domain to be configured in jmx

calvinjia:
Previously our domain was always `metrics` which is not very helpful, it now defaults to `org.alluxio`

### DIFF
--- a/core/common/src/main/java/alluxio/metrics/sink/JmxSink.java
+++ b/core/common/src/main/java/alluxio/metrics/sink/JmxSink.java
@@ -23,6 +23,8 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 @ThreadSafe
 public final class JmxSink implements Sink {
+  private static final String DOMAIN = "org.alluxio";
+
   private JmxReporter mReporter;
 
   /**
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/metrics/sink/JmxSink.java
```diff
@@ -32,7 +34,14 @@
    * @param registry the metric registry to register
    */
   public JmxSink(Properties properties, MetricRegistry registry) {
-    mReporter = JmxReporter.forRegistry(registry).build();
+    JmxReporter.Builder builder = JmxReporter.forRegistry(registry);
+    String domain = properties.getProperty("domain");
```
ZacBlanco: Will we ever need to know if the "`domain`" key will need to be used anywhere else? Should we make it a constant?

calvinjia: For metrics configuration we just hardcode the values (ie. https://github.com/Alluxio/alluxio/blob/master/core/common/src/main/java/alluxio/metrics/MetricsSystem.java#L166). This one seems 1 time use, aside from maybe testing?

ZacBlanco: I am ok leaving it for now. Maybe in the future we can just integrate metrics configuration with `alluxio-site.properties`....

calvinjia:

PR #10859, opened 2020-02-06
Add client cache dry run mode

bf8086:
Added a dry run mode which bypass all client caching but still collect metrics as if the cache is enabled.


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -72,6 +74,10 @@
   public LocalCacheFileInStream(AlluxioURI path, OpenFilePOptions options, FileSystem externalFs,
       CacheManager cacheManager) {
     mPageSize = externalFs.getConf().getBytes(PropertyKey.USER_CLIENT_CACHE_PAGE_SIZE);
+    CacheMode dryrun = externalFs.getConf()
```
calvinjia: `dryrun` -> `cacheMode`?

PR #10860, opened 2020-02-06
Set default logger level for InternalParquetRecordReader/Writer to WARN

bradyoo:
InternalParquetRecordReader/Writer gets very noisy when row group size of the parquet file writing/reading is small.


gpang:
@bradyoo Doesn't WARN also get written to the log by default?

bradyoo:
the default logger level for this is INFO.

bradyoo:
(This PR is intending to remove the INFO logs)

gpang:

PR #10864, opened 2020-02-07
Disable authentication in client for versionService

ns1123:
Client shouldn't be creating an authenticated channel for an unauthenticated service.


core/common/src/main/java/alluxio/master/PollingMasterInquireClient.java
```diff
@@ -131,7 +131,8 @@ private InetSocketAddress getAddress() {
   private void pingMetaService(InetSocketAddress address) throws AlluxioStatusException {
     GrpcChannel channel =
```
gpang: Please add a comment that we are disabling authentication on purpose.

gpang:
@ns1123 Since this is a bug, could you create an issue describing the bug and behavior, and link to it from this PR? Thanks!

gpang:
@ns1123 can we close this now?

ns1123:
Closing as this is addressed by #10878

PR #10865, opened 2020-02-07
Enable parallel udb sync

gpang:
This change sync a table in a thread, instead of sequentially. This should reduce the response times of databases with a large number of tables.

### DIFF
--- a/table/server/master/src/main/java/alluxio/master/table/Database.java
+++ b/table/server/master/src/main/java/alluxio/master/table/Database.java
@@ -11,6 +11,7 @@
 
 package alluxio.master.table;
 
+import alluxio.Constants;
 import alluxio.exception.ExceptionMessage;
 import alluxio.exception.status.NotFoundException;
 import alluxio.grpc.table.FileStatistics;
@@ -22,6 +23,7 @@ import alluxio.proto.journal.Journal;
 import alluxio.table.common.udb.UdbContext;
 import alluxio.table.common.udb.UdbTable;
 import alluxio.table.common.udb.UnderDatabase;
+import alluxio.util.CommonUtils;
 
 import com.google.common.collect.Iterators;
 import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -177,31 +182,46 @@ public Table createTable(String tableName, Schema schema) {
    * external synchronization.
    *
    * @param context journal context
-   * @return true if the database changed as a result of fullSync
+   * @param service executor service for parallezing the sync
+   * @return true if the database changed as a result of syncing with UDB
    */
-  public boolean sync(JournalContext context) throws IOException {
-    boolean updated = false;
+  public boolean sync(JournalContext context, ExecutorService service) throws IOException {
```
yuzhu: this `ExecutorService` is not used?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HiveClientPool.java
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Creates a new hive client client pool.
+   *
+   * @param connectionUri the connect uri for the hive metastore
+   * @param hiveDbName the db name in hive
+   */
+  public HiveClientPool(String connectionUri, String hiveDbName) {
+    super(Options.defaultOptions()
+        .setMinCapacity(16)
```
yuzhu: do you think 16 may be too high a number? we are assuming there will be at least 16 concurrent udb connections?

table/server/master/src/main/java/alluxio/master/table/Database.java
gpang: ooops, I forgot to update `invokeAll`. Thanks!

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HiveClientPool.java
gpang: In a way, yes. If a database has more than 16 tables, it would easily need more concurrent hive clients. For example, tpcds has over 20 tables.

gpang:

PR #10873, opened 2020-02-10
Add integration test for transform journaling

yuzhu:


yuzhu:
jenkins, test this please

tests/src/test/java/alluxio/server/ft/journal/TableMasterJournalIntegrationTest.java
```diff
@@ -45,18 +56,17 @@
 
   private static final String DB_NAME = TestDatabase.TEST_UDB_NAME;
 
-  @Before
-  public void before() throws Exception {
-    TestDatabase.genTable(1, 2);
-  }
+  @ClassRule
+  public static ManuallyScheduleHeartbeat sManuallySchedule = new ManuallyScheduleHeartbeat(
```
gpang: is there a way we can avoid this rule? This doesn't interact well with `LocalAlluxioClusterResource` especially when `LocalAlluxioClusterResource` is a class rule (which is ideal, so it only creates the cluster once per class)

gpang: Is there a way we can make the `LocalAlluxioClusterResource` a `ClassRule`, and not per test?

yuzhu: If we make it a ClassRule, does that mean the state will be kept between each test? that way, tests will interfere with each other right?

gpang: Yes, so we just have to clear the necessary state in Alluxio. The reason I want to go in that direction is that there is a large startup cost to creating a cluster.

yuzhu: what is the alternative to ManuallyScheduleHeartbeat? i need to manually trigger a heartbeat in a test. 

Changed it to use classrule.

gpang:

PR #10877, opened 2020-02-10
Fix meter bug to calculate throughput value at master directly

LuQQiu:
Alluxio clients are usually short-lived especially those created through running Alluxio CLI. Alluxio throughput is recorded as `MetricType:Meter`. `Meter.getOneMinuteRate()` can be zero when the data is first marked since the one minute rate is a one-minute exponentially-weighted moving average rate. When a client is shutting down, it will report metrics the last time to leader master. Because they are too short-lived, the reported `meter.getOneMinuteRate()` is always zero, client metrics can be considered as lost in this scenario.

In this PR, cluster throughput values are not aggregated value of client/worker reported meter one minute rate anymore. Instead, cluster throughput value is calculated when requested by dividing cluster bytes counter to the metrics uptime (the last time that the metrics store is cleared).

Because of this change, only counter value needs to be reported to the leading master, many logics in the leading master metrics store to store the individual metrics of types other than counter can be removed. In addition, when the should report counters are all zero, client heartbeats to report metrics will not be triggered. A large amount of client heartbeat overhead is reduced.


LuQQiu:
jenkins, test this please

LuQQiu:
Jenkins, test this please

LuQQiu:
alluxio.client.cli.JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal(alluxio.client.cli.JournalToolTest)
  Run 1: JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal:165->checkpointEmbeddedJournal:206 » Internal
  Run 2: JournalToolTest.dumpHeapCheckpointFromEmbeddedJournal:154 » Alluxio Fatal erro...

alluxio.client.cli.fs.command.CopyFromLocalCommandIntegrationTest.copyFromLargeLocalDirectory(alluxio.client.cli.fs.command.CopyFromLocalCommandIntegrationTest)
  Run 1: CopyFromLocalCommandIntegrationTest.copyFromLargeLocalDirectory:353 » TestTimedOut
  Run 2: CopyFromLocalCommandIntegrationTest.copyFromLargeLocalDirectory:353 » TestTimedOut

LuQQiu:
Tests in error: 
  Run 1: JobServiceFaultToleranceShellTest.before:52 » IO java.util.concurrent.TimeoutE...
  Run 2: JobServiceFaultToleranceShellTest.before:52 » IO java.util.concurrent.TimeoutE...

LuQQiu:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3221,7 +3203,7 @@ public String toString() {
   public static final PropertyKey USER_METRICS_HEARTBEAT_INTERVAL_MS =
       new Builder(Name.USER_METRICS_HEARTBEAT_INTERVAL_MS)
           .setAlias("alluxio.user.metrics.heartbeat.interval.ms")
-          .setDefaultValue("1min")
+          .setDefaultValue("10sec")
```
ZacBlanco: What's the reasoning for using 10 seconds?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -668,7 +668,7 @@ public MetricKey build() {
           .setDescription("Bytes read throughput from Alluxio storage "
               + "via domain socket by this worker")
           .setMetricType(MetricType.METER)
-          .setIsClusterAggregated(true)
+          .setIsClusterAggregated(false)
```
ZacBlanco: why are these all set to false now?

core/common/src/main/java/alluxio/conf/PropertyKey.java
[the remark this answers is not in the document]
LuQQiu: It's hard to choose the default value for this interval. 
We want it to be big enough to make sure master can handle that many client connections. --> reduce client connections overhead on the leading master.
Also we want it to be small enough that we can see more updated changes in the leading master metrics page/sink.

So it will be good to choose the smallest number under the criteria that leading master can handles.
I expect that we can handle 1000 workers with heartbeat 1sec to the leading master. Then I feel we are likely to handle 10,000 workers with heartbeat interval 10sec.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -132,45 +76,34 @@ public void putWorkerMetrics(String hostname, List<Metric> metrics) {
...
-   * Put the metrics from a client with a hostname and a client id. If all the old metrics
-   * associated with this instance will be removed and then replaced by the latest.
+   * Put the metrics from a client with a hostname.
    *
    * @param hostname the hostname of the client
-   * @param clientId the id of the client
    * @param metrics the new metrics
    */
-  public void putClientMetrics(String hostname, String clientId, List<Metric> metrics) {
+  public void putClientMetrics(String hostname, List<Metric> metrics) {
```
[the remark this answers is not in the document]
LuQQiu: There is still some discussion needed for the PR, it would be better to deal with them separately.

core/common/src/main/java/alluxio/metrics/MetricKey.java
LuQQiu: Because they are throughput meter. Instead of sending meter one minute rate to the leading master, directly calculating throughput value in the leading master is more correct and easier to understand. The details are written in the PR description

LuQQiu:

PR #10890, opened 2020-02-11
Allow client name to be configurable

calvinjia:
If a user sets the appid, we can use that instead of the hostname.

### DIFF
--- a/core/common/src/main/java/alluxio/metrics/MetricsSystem.java
+++ b/core/common/src/main/java/alluxio/metrics/MetricsSystem.java
@@ -12,6 +12,7 @@
 package alluxio.metrics;
 
 import alluxio.AlluxioURI;
+import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.InstancedConfiguration;
 import alluxio.conf.PropertyKey;
 import alluxio.grpc.MetricType;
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -341,12 +342,17 @@ public static String getJobWorkerMetricName(String name) {
    * @return the metric registry name
    */
   private static String getMetricNameWithUniqueId(InstanceType instance, String name) {
+    AlluxioConfiguration conf = new InstancedConfiguration(ConfigurationUtils.defaults());
+    String identifier;
+    if (conf.isSet(PropertyKey.USER_APP_ID)) {
+      identifier = conf.get(PropertyKey.USER_APP_ID);
+    } else {
+      identifier = NetworkAddressUtils.getLocalHostMetricName(sResolveTimeout);
```
LuQQiu: In the getLocalHostMetricName, we have the caching logic, maybe better to put this logic with caching.

In addition, this function is also used by instanceType=WORKER but the changes may relate to clients only.

I still no sure is it right to change it.

The `metrics.Metric.java` class may also need to be changed so that metric names can be processed correctly with the new format. Metric name is so tricky.....!

PR #10892, opened 2020-02-12
Use source instead of hostname and client id in metrics

LuQQiu:
Use source instead of hostname and client id to identify metrics source


LuQQiu:
@calvinjia PTAL, I only modify a small set of files. Most of the files are auto-generated proto code, no sure why that many. I may directly run -Pgenerate on the master and create a PR to minmize those auto-generated coded.

LuQQiu:
Need to merge https://github.com/Alluxio/alluxio/pull/10893 before this PR

calvinjia:
@LuQQiu could you address the merge conflicts? #10893 has been merged

LuQQiu:
@calvinjia Addressed the merge conflicts, PTAL, thanks

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -73,6 +76,10 @@
   // A flag telling whether metrics have been reported yet.
   // Using this prevents us from initializing {@link #SHOULD_REPORT_METRICS} more than once
   private static boolean sReported = false;
+  // The source of the metrics in this metrics system.
+  // It can be set through property keys based on process types.
+  // Local hostname will be used if no related property key founds.
+  private static String sSourceName;
```
calvinjia: Why does this need to be a static variable? It seems like we are using it as a local variable?

LuQQiu: All the methods calling it are static methods.

LuQQiu: All methods in the `MetricsSystem` class are static methods, MetricsSystem is initialized per-jvm and does not allow explicit initialization.

calvinjia: I mean a local variable in the method, not an instance variable

LuQQiu: Oh, I understand, this is used for memory the source value since this method `getMetricNameWithUniqueId` is called really often.

I added the null check

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -341,12 +347,38 @@ public static String getJobWorkerMetricName(String name) {
    * @return the metric registry name
    */
   private static String getMetricNameWithUniqueId(InstanceType instance, String name) {
+    if (sSourceName == null) {
```
calvinjia: Could you add the appropriate here to ensure `sSourceName` is only set once? Also do we assume each JVM will only have 1 instance type?

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -69,16 +70,16 @@ public synchronized void heartbeat() {
     // Currently we only support JVM-level metrics. A list is used here because in the near
     // future we will support sending per filesystem client-level metrics.
     List<alluxio.grpc.ClientMetrics> fsClientMetrics = new ArrayList<>();
-    String hostname = NetworkAddressUtils.getClientHostName(mConf);
+    String source = mConf.isSet(PropertyKey.USER_APP_ID)
+        ? mConf.get(PropertyKey.USER_APP_ID) : NetworkAddressUtils.getClientHostName(mConf);
```
calvinjia: Consider making this an instance variable?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -158,6 +164,37 @@ public static void startSinks(String metricsConfFile) {
     startSinksFromConfig(config);
   }
 
+  /**
+   * Initialize the {@link MetricsSystem}.
+   * This should be called after setting the {@link CommonUtils#PROCESS_TYPE}
+   * and before creating any metric in the current process.
+   */
+  public static void init() {
```
calvinjia: Could you make this synchronized and have it be a no-op if `sSourceName` has already been set? The reason is the client may make multiple calls to init when multiple filesystem objects are created (ie. for different users).
Also, where is this called for a client?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Yeah, that's why assumed, each jvm will only have one instance type depends on the process type

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Sounds good. I maked it a no-op if already been set and maked it synchronized.
Added the call of init() to FileSystemContext.initContext() and a bunch of unit tests.

LuQQiu:
Failed tests: 
alluxio.server.ft.journal.TableMasterJournalIntegrationTest.journalTransformDb(alluxio.server.ft.journal.TableMasterJournalIntegrationTest)
  Run 1: TableMasterJournalIntegrationTest.journalTransformDb:198 expected:<COMPLETED> but was:<FAILED>
  Run 2: TableMasterJournalIntegrationTest.journalTransformDb:198 expected:<COMPLETED> but was:<FAILED>

LuQQiu:
Jenkins, test this please

LuQQiu:
Tests in error: 
alluxio.server.ft.ZookeeperFailureIntegrationTest.zkConnectionPolicy_Session(alluxio.server.ft.ZookeeperFailureIntegrationTest)
  Run 1: ZookeeperFailureIntegrationTest.zkConnectionPolicy_Session:139 » Timeout Timed...
  Run 2: ZookeeperFailureIntegrationTest.zkConnectionPolicy_Session:139 » Timeout Timed...

LuQQiu:
Jenkins, test this please

core/common/src/main/java/alluxio/metrics/Metric.java
```diff
@@ -321,26 +277,15 @@ public static Metric from(String fullName, double value, MetricType metricType)
...
-        String[] ids = pieces[len - 1].split(ID_SEPARATOR);
-        hostname = ids[0];
-        id = ids[1];
-      } else {
-        hostname = pieces[len - 1];
-      }
-      name = pieces[1];
+    if (!pieces[0].equals(MetricsSystem.InstanceType.MASTER.toString())
+        && !pieces[0].equals(MetricsSystem.CLUSTER)) {
+      hostname = pieces[len - 1];
```
ZacBlanco: `hostname` isn't really a descriptive variable name for this anymore, right?

core/common/src/main/java/alluxio/metrics/Metric.java
```diff
@@ -321,26 +277,15 @@ public static Metric from(String fullName, double value, MetricType metricType)
...
       tagEndIndex = len - 1;
     }
     MetricsSystem.InstanceType instance = MetricsSystem.InstanceType.fromString(pieces[0]);
-    Metric metric = new Metric(instance, hostname, id, metricType, name, value);
+    Metric metric = new Metric(instance, hostname, metricType, pieces[1], value);
```
ZacBlanco: `pieces[1]` seems kind of ambiguous compared to the previous `name` variable. Can we add a comment on what it is, or use the old variable?

LuQQiu:

Issue #10812, opened 2020-01-30
Alluxio CLI doesn't show right mount info (readonly/shared)


LuQQiu:
**Alluxio Version:**
Tested in both branch-2.1 and master branch 

**Describe the bug**
CLI doesn't show right mount info (readonly/shared)

Start alluxio with 
```
alluxio.master.mount.table.root.ufs=${alluxio.work.dir}/underFSStorage
alluxio.master.mount.table.root.readonly=true
alluxio.master.mount.table.root.shared=true
```
but mount info shows not read-only and not shared
```
$ ./bin/alluxio fs mount
/Users/alluxio/alluxioFolder/alluxio/underFSStorage  on  /  (local, capacity=465.72GB, used=-1B(0%), not read-only, not shared, properties={})
$ ./bin/alluxio fsadmin report ufs
Alluxio under storage system information:
/Users/alluxio/alluxioFolder/alluxio/underFSStorage  on  /  (local, capacity=465.72GB, used=-1B(0%), not read-only, not shared, properties={})
```

Try to delete file but find the mount point should be read-only
```
$ ./bin/alluxio fs rm /LICENSE
Failed to delete /LICENSE (Operation WRITE not allowed on ufs path /Users/alluxio/alluxioFolder/alluxio/underFSStorage/LICENSE under maintenance mode READ_ONLY) from the under file system
```

**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)

**Expected behavior**
Mount should show read-only in my case

**Urgency**
Describe the impact and urgency of the bug.

**Additional context**

jiacheliu3:
@LuQQiu Actually when creating the root mount as readonly and shared, I don't think it really takes effect. For example I started Alluxio with readonly root mount using your config properties, and then I can put files in the root mount. 
```
$bin/alluxio fs copyFromLocal ./LICENSE /LICENSE
Copied file:///Users/jiachengliu/Documents/Alluxio/alluxio-jiacheng/alluxio/LICENSE to /LICENSE
```

Do you know if readonly works at all?

jiacheliu3:
Confirmed that readonly doesn't work for root mount but seems to work for nested mount. See different behavior below. 
```
$ ./bin/alluxio fs mount --shared --readonly /readonly_folder /Users/jiachengliu/Documents/Alluxio/alluxio-jiacheng/alluxio/readonly
[...]

PR #10895, opened 2020-02-12
Pass readonly option on root mount

jiacheliu3:
This resolves issue https://github.com/Alluxio/alluxio/issues/10812#event-3012406507

Correctly propagate root mount options `alluxio.master.mount.table.root.readonly` to MountTable

After the change:
```
$ bin/alluxio fs mount                                                                                     
/xxx/alluxio/underFSStorage  on  /  (local, capacity=420.03GB, used=-1B(0%), read-only, not shared, properties={})
$ bin/alluxio getConf --master | grep readonly
alluxio.master.mount.table.root.readonly=true
alluxio.underfs.object.store.mount.shared.publicly=false
$ bin/alluxio fs copyFromLocal ./LICENSE /LICENSE                                                          
A write operation on /LICENSE under a readonly mount point / is not allowed
```

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -456,14 +456,8 @@ public final class DefaultFileSystemMaster extends CoreMaster
 
   private static MountInfo getRootMountInfo(MasterUfsManager ufsManager) {
     try (CloseableResource<UnderFileSystem> resource = ufsManager.getRoot().acquireUfsResource()) {
-      boolean shared;
-      if (resource.get().isObjectStorage()) {
-        shared = ServerConfiguration.getBoolean(
-                PropertyKey.UNDERFS_OBJECT_STORE_MOUNT_SHARED_PUBLICLY);
-      } else {
-        shared = ServerConfiguration.getBoolean(
-                PropertyKey.MASTER_MOUNT_TABLE_ROOT_SHARED);
-      }
+      boolean shared = resource.get().isObjectStorage()
+              && ServerConfiguration.getBoolean(PropertyKey.UNDERFS_OBJECT_STORE_MOUNT_SHARED_PUBLICLY);
       boolean readonly = ServerConfiguration.getBoolean(
               PropertyKey.MASTER_MOUNT_TABLE_ROOT_READONLY);
       String rootUfsUri = ServerConfiguration.get(PropertyKey.MASTER_MOUNT_TABLE_ROOT_UFS);
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -457,9 +457,9 @@ public final class DefaultFileSystemMaster extends CoreMaster
   private static MountInfo getRootMountInfo(MasterUfsManager ufsManager) {
     try (CloseableResource<UnderFileSystem> resource = ufsManager.getRoot().acquireUfsResource()) {
       boolean shared = resource.get().isObjectStorage()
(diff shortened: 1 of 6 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -456,13 +456,22 @@ public DefaultFileSystemMaster(BlockMaster blockMaster, CoreMasterContext master
...
+        shared = ServerConfiguration.getBoolean(
+                PropertyKey.UNDERFS_OBJECT_STORE_MOUNT_SHARED_PUBLICLY);
+      } else {
+        shared = ServerConfiguration.getBoolean(
+                PropertyKey.MASTER_MOUNT_TABLE_ROOT_SHARED);
+      }
+      boolean readonly = ServerConfiguration.getBoolean(
+              PropertyKey.MASTER_MOUNT_TABLE_ROOT_READONLY);
       String rootUfsUri = ServerConfiguration.get(PropertyKey.MASTER_MOUNT_TABLE_ROOT_UFS);
-      boolean shared = resource.get().isObjectStorage()
```
jiacheliu3: @gpang Does the original implementation look like a bug to you?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -456,13 +456,22 @@ public DefaultFileSystemMaster(BlockMaster blockMaster, CoreMasterContext master
...
   private static MountInfo getRootMountInfo(MasterUfsManager ufsManager) {
     try (CloseableResource<UnderFileSystem> resource = ufsManager.getRoot().acquireUfsResource()) {
+      boolean shared;
+      if (resource.get().isObjectStorage()) {
+        shared = ServerConfiguration.getBoolean(
+                PropertyKey.UNDERFS_OBJECT_STORE_MOUNT_SHARED_PUBLICLY);
+      } else {
+        shared = ServerConfiguration.getBoolean(
+                PropertyKey.MASTER_MOUNT_TABLE_ROOT_SHARED);
+      }
```
jiacheliu3: I tried to update `shared` to this logic which seems more correct. And subsequently found issue https://github.com/Alluxio/alluxio/issues/10896
The issue related to `shared` seems deeper than I thought. I will revert the change regarding `shared` field and focus on `readonly` in this PR. I will attempt on the `shared` field in a separate PR later.

jiacheliu3:
@gpang I tried to add a test but wasn't able to.

There's no existing test for `DefaultFileSystemMaster` and I think the workload creating one deserves a separate PR already...

And I don't think the unit/integration tests for either MountCommand or ReportCommand are good places to test this.

Do you have a better idea where the tests can be?

jiacheliu3:
@gpang PTAL there's one integration test added on readonly root mount. Thanks!

tests/src/test/java/alluxio/client/fs/FileSystemReadonlyTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.util.Map;
+
+/**
+ * Test behavior of {@link FileSystemMaster}, when the paths are readonly.
+ */
+public class FileSystemReadonlyTest {
+  @Rule
```
gpang: Can we make this a `ClassRule` so we dont have to re-create the cluster for each test? Starting a cluster is costly. (I think this also means the variable will have to be static)

jiacheliu3: Done. BTW most integration tests are still using `@Rule`. We might want to batch-fix that in the future.

gpang: Yes, I know. That is why I'm trying to make new usages use classrule. Thanks!

gpang:

PR #10897, opened 2020-02-12
Improve error handling for attach db and sync

gpang:
This change adds an `attachdb` option `--ignore-sync-errors` to keep the database attached even if a sync failure occurred.

PR #10900, opened 2020-02-12
Enable configuration of ignored tables

gpang:
This allows ignoring a certain set of tables when attaching a database.


gpang:
this depends on Alluxio/alluxio#10897

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -58,20 +60,23 @@
...
+      CatalogConfiguration config) {
     mContext = context;
     mType = type;
     mName = name;
     mTables = new ConcurrentHashMap<>();
     mUdb = udb;
-    mConfig = configMap;
+    mConfig = config;
+    mIgnoreTables = Sets.newHashSet(Splitter.on(",").trimResults().omitEmptyStrings()
+        .split(mConfig.get(CatalogProperty.DB_IGNORE_TABLES)));
```
ZacBlanco: This looks like the same implementation from `InstancedConfiguration`, but maybe we can extract this into a function in `FormatUtils` or `CommonUtils` so we don't have the code in both places?

gpang: I extract that part.

core/common/src/test/java/alluxio/util/ConfigurationUtilsTest.java
```diff
@@ -142,6 +143,15 @@ public void testMasterNotConfiguredMessage() {
...
   }
 
+  @Test
+  public void parseAsList() {
+    assertEquals(Lists.newArrayList("a"), ConfigurationUtils.parseAsList("a", ","));
+    assertEquals(Lists.newArrayList("a", "b", "c"), ConfigurationUtils.parseAsList("a,b,c", ","));
+    assertEquals(Lists.newArrayList("a", "b", "c"),
+        ConfigurationUtils.parseAsList(" a , b , c ", ","));
+    assertEquals(Lists.newArrayList("a,b,c"), ConfigurationUtils.parseAsList("a,b,c", ";"));
+  }
```
ZacBlanco: Can we add a case for `a,,b` to make sure empty strings are omitted?

core/common/src/main/java/alluxio/util/ConfigurationUtils.java
```diff
@@ -358,10 +357,8 @@ public static boolean isHaMode(AlluxioConfiguration conf) {
 
     List<ConfigProperty> configs = new ArrayList<>();
     List<PropertyKey> selectedKeys =
-        conf.keySet().stream()
-            .filter(key -> GrpcUtils.contains(key.getScope(), scope))
-            .filter(key -> key.isValid(key.getName()))
-            .collect(toList());
+        conf.keySet().stream().filter(key -> GrpcUtils.contains(key.getScope(), scope))
+            .filter(key -> key.isValid(key.getName())).collect(toList());
```
ZacBlanco: Not sure if you updated all this by hand or automatically formatted, but I think the old version of this is easier to read

core/common/src/test/java/alluxio/util/ConfigurationUtilsTest.java
gpang: Good call. Added.

core/common/src/main/java/alluxio/util/ConfigurationUtils.java
gpang: It was auto-formatted. I reverted back.

gpang:
I am updating this with tests and improvements

gpang:
@yuzhu please take a look

table/server/master/src/main/java/alluxio/master/table/CatalogProperty.java
```diff
@@ -26,6 +26,7 @@ private CatalogProperty(String name, String description, String defaultValue) {
     super(name, description, defaultValue);
   }
 
-  public static final CatalogProperty DB_TYPE =
```
yuzhu: why delete this? do we not need it for `GLUE`?

table/server/master/src/main/java/alluxio/master/table/CatalogProperty.java
```diff
@@ -26,6 +26,7 @@ private CatalogProperty(String name, String description, String defaultValue) {
...
-      new CatalogProperty("db.type", "The type of database. Possible values are: hive", "");
+  public static final CatalogProperty DB_IGNORE_TABLES =
+      new CatalogProperty("catalog.db.ignore.tables",
+          "The comma-separated list of table names to ignore from the UDB.", "");
```
yuzhu: does this option affect deletion? should we have a comment about that?

gpang: For now, it is only related to getting tables from udb via udb sync.

gpang:

Issue #10858, opened 2020-02-06
Upgrade cycle with RocksDB results in BlockStore errors


ns1123:
**Alluxio Version:**
2.1.2

**Describe the bug**
When upgrading from an older version of Alluxio to a newer one, failing to copy the ./metastore directory that is generated for RocksDB results in errors with certain blocks missing from the blockstore. This causes cached blocks to be considered orphaned when the master starts up.

**To Reproduce**
Repro is intermittent. General steps involve using Embedded and RocksDB and then creating/discovering directories from the UFS. After blockID's have been assigned, load the data and then stop the cluster. Then delete the ./metadata folder and restart the masters. (So far I have been unable to repro on a single master embedded cluster). In certain cases, some files will report missing BlockId Exceptions when trying to run ./bin/alluxio fs stat.

**Expected behavior**
Either replay of the journal rebuilds the BlockStore in its entirety in RocksDB, or instructions need to be updated to explicitly mention the need to preserve the directory in the event of an upgrade.

**Urgency**
High. Users are losing cached block metadata which can be a painful process.

PR #10901, opened 2020-02-13
Fix iteration for RocksBlockStore

ggezer:
This is to get rocks db out of prefix seek mode, under which iterators seekToFirst/Last APIs are ineffective. See https://github.com/facebook/rocksdb/wiki/Prefix-Seek-API-Changes for more context.

Fixes https://github.com/Alluxio/alluxio/issues/10858

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1518,7 +1518,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE =
       new Builder(Name.MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE)
-          .setDefaultValue("10000")
+          .setDefaultValue("4MB")
           .setDescription("The read-ahead size for block iterator.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1516,6 +1516,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE =
+      new Builder(Name.MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE)
+          .setDefaultValue("10000")
+          .setDescription("The read-ahead size for block iterator.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_METASTORE_INODE_INHERIT_OWNER_AND_GROUP =
       new Builder(Name.MASTER_METASTORE_INODE_INHERIT_OWNER_AND_GROUP)
           .setDefaultValue("true")
@@ -4234,6 +4241,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.metastore.inode.iteration.crawler.count";
     public static final String MASTER_METASTORE_INODE_ENUMERATOR_BUFFER_COUNT =
         "alluxio.master.metastore.inode.enumerator.buffer.count";
+    public static final String MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE =
+        "alluxio.master.metastore.block.iterator.readahead.size";
     public static final String MASTER_METASTORE_INODE_INHERIT_OWNER_AND_GROUP =
         "alluxio.master.metastore.inode.inherit.owner.and.group";
     public static final String MASTER_PERSISTENCE_CHECKER_INTERVAL_MS =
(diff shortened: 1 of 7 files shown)


ZacBlanco:
@ggezer can you update the description to describe how this PR fixes the issue?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1516,6 +1516,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE =
+      new Builder(Name.MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE)
+          .setDefaultValue("10000")
+          .setDescription("The read-ahead size for block iterator.")
```
gpang: Can you describe this more, and the implications of a small one and a large one?

Also, what is "size"? rows? bytes?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -67,10 +70,11 @@
   public RocksBlockStore(String baseDir) {
     RocksDB.loadLibrary();
     mDisableWAL = new WriteOptions().setDisableWAL(true);
+    mIteratorOption = new ReadOptions().setReadaheadSize(
+        ServerConfiguration.getInt(PropertyKey.MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE));
```
gpang: below, it uses `setPrefixSameAsStart()`. We don't need that anymore?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -67,10 +70,11 @@
...
     ColumnFamilyOptions cfOpts = new ColumnFamilyOptions()
         .setMemTableConfig(new HashLinkedListMemTableConfig())
-        .setCompressionType(CompressionType.NO_COMPRESSION)
-        .useFixedLengthPrefixExtractor(8); // We always search using the initial long key
```
gpang: why is this removed?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1516,6 +1516,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE =
+      new Builder(Name.MASTER_METASTORE_BLOCK_ITERATOR_READAHEAD_SIZE)
+          .setDefaultValue("10000")
```
gpang: Should we make this larger?

core/server/master/src/test/java/alluxio/master/metastore/rocks/RocksBlockStoreTest.java
```diff
@@ -0,0 +1,60 @@
...
+  public void testIterator() throws Exception {
+    final int blockCount = 3;
+    RocksBlockStore blockStore = new RocksBlockStore(mFolder.newFolder().getAbsolutePath());
+    for (int i = 0; i < blockCount; i++) {
+      blockStore.putBlock(i, Block.BlockMeta.newBuilder().setLength(i).build());
+    }
+
+    Iterator<BlockStore.Block> iter = blockStore.iterator();
+    for (int i = 0; i < blockCount; i++) {
+      assertTrue(iter.hasNext());
```
gpang: after the loop, we should check that the iterator is done?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
ggezer: This will put rocks-db to prefix-seek mode. Please see description for more details.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: bytes

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: I was under impression that it was count, but turns out it's bytes. Made it 4MB to start with.

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
ggezer: This one uses `RocksIterator.seek()` for which this is fine.

ggezer:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksInodeStore.java
```diff
@@ -245,7 +245,8 @@ public boolean hasChildren(InodeDirectoryView inode, ReadOption option) {
    * @return an iterator over stored inodes
    */
   public Iterator<InodeView> iterator() {
-    return new RocksInodeViewIterator(db().newIterator(mInodesColumn.get()));
+    return RocksUtils.createIterator(db().newIterator(mInodesColumn.get()),
+        (iter) -> getMutable(Longs.fromByteArray(iter.key()), ReadOption.defaults()).get());
```
gpang: should this also use the read option like the other iterator?

core/common/src/main/java/alluxio/conf/PropertyKey.java
gpang: I see. Could you update the description to mention bytes, and maybe some of the implications of the size? For example, if we have a billion blocks, is reading 4mb at a time efficient enough? Should this be a larger size?

ggezer: I guess 4MB suffices, as it'd prefetch about 10K inodes and 100K block-Ids. This should be enough as iterator content are ultimately written back to disk.

ggezer:

PR #10908, opened 2020-02-13
Add metrics tests

LuQQiu:
This PR does some cleaning jobs of metrics:
(1) Add time loggings for metrics related methods
(2) Reduce the heavy string operations by memorizing the result of escapePath
(3) Reduce the call for METRIC_REGISTRY.getGauges() since each getGauges call traverse the whole metrics and make a full copy of gauges.
(4) Add the MetricsStoreTest and MetricsSystemTest


core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -203,13 +205,16 @@ public void init() {
...
   public void clear() {
+    long start = System.currentTimeMillis();
     try (LockResource r = new LockResource(mLock.writeLock())) {
       for (Counter counter : mClusterCounters.values()) {
         counter.dec(counter.getCount());
       }
       mLastClearTime = System.currentTimeMillis();
       MetricsSystem.resetAllMetrics();
     }
+    LOG.info("Clearing the metrics store and metrics system in {} ms",
```
ZacBlanco: nit: 

```suggestion
    LOG.info("Cleared the metrics store and metrics system in {} ms",
```

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
```diff
@@ -57,6 +59,8 @@
   private final UnderFileSystem mUnderFileSystem;
   private final UnderFileSystemConfiguration mConf;
   private final String mPath;
+  private final Supplier<String> mEscapedPath =
+      CommonUtils.memoize(this::escapePath);
```
ZacBlanco: what does memoizing do for us here?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -718,6 +731,8 @@ public static synchronized void resetAllMetrics() {
       METRIC_REGISTRY.timer(timerName);
     }
     LAST_REPORTED_METRICS.clear();
+    LOG.info("Reset all metrics in the metrics system in {}ms",
+        System.currentTimeMillis() - startTime);
```
ZacBlanco: I feel like this will be a spammy log at `INFO` level

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -395,7 +397,9 @@ public static String stripInstanceAndHost(String metricsName) {
    * @return the string representing the escaped URI
    */
   public static String escape(AlluxioURI uri) {
-    return uri.toString().replace("%", "%25").replace("/", "%2F").replace(".", "%2E");
+    return CACHED_ESCAPED_PATH.computeIfAbsent(uri,
+        u -> u.toString().replace("%", "%25")
+            .replace("/", "%2F").replace(".", "%2E"));
```
ZacBlanco: If you're looking to escape into URI-compatible paths I would look somewhere in apache commons utils

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: `ResetAllMetrics()` is used very often in tests, but should be called really rarely in production. It will only be called when the Master starts and when the `fsadmin metrics clear` is called. Since it may affect the metrics recording and delays some metrics heartbeat, logging is useful.

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
LuQQiu: So that when MetricsSystem record the metrics of any UFS operation,  the escaped path can be memorized without constructing Alluxio URI and call metho in MetricsSystem.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: I haven't found existing utils doing exactly the same thing. If there is a util method, that would be really convenient.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
ZacBlanco: fair enough

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
ZacBlanco: I don't know what the original intention of this code was (if it is doing a URL encoding on the metric name or not, but it looks like it to me.

We should be using this: https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
ZacBlanco: Isn't the escaped path always the same though because `mPath` is final?

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: I think for the first time `MetricsSystem.escape` created for Metric name processing. Like we use `.` for distinguish metric name components like (`InstanceType.MetricName.Tag:TagValue`), so the given url should not have `.`.

This use case does not fit in `UTLEncoder` because it will `The special characters ".", "-", "*", and "_" remain the same.`

ZacBlanco: Since the code was already here I guess it's ok - but this seems very sketchy to me...we should improve it in the future.

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
LuQQiu: Make sense, Changed to initialize directly in the constructor.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
LuQQiu: Yeah, it's sketchy, will reconsider how to improve it in the future

LuQQiu:
Tests in error: 
  Run 1: FreeAndDeleteIntegrationTest.deleteDir:146 » TestTimedOut test timed out after...
  Run 2: FreeAndDeleteIntegrationTest.deleteDir:148 » TestTimedOut test timed out after...

alluxio.job.master.JobMasterIntegrationTest.cancel(alluxio.job.master.JobMasterIntegrationTest)
  Run 1: JobMasterIntegrationTest.cancel:246 » Timeout Timed out waiting for job 158163...
  Run 2: JobMasterIntegrationTest.cancel:246 » Timeout Timed out waiting for job 158163...

LuQQiu:
Jenkins, test this please

LuQQiu:
Jenkins, test this please

LuQQiu:

PR #10911, opened 2020-02-13
Add an experimental API to URIStatus

calvinjia:


core/common/src/main/java/alluxio/client/file/URIStatus.java
```diff
@@ -299,6 +299,16 @@ public String getUfsFingerprint() {
...
+  /**
+   * This is an experimental API. The returned {@link FileInfo} object does not have a stable API.
+   * Make modifications to the returned file info object at your own risk.
+   *
+   * @return the underlying file info object
+   */
+  public FileInfo getFileInfo() {
+    return mInfo;
+  }
+
```
ZacBlanco: Is there a `@Experimental` annotation or something similar we can use?

calvinjia: I tried to look for it, but I think we've done away with that annotation

calvinjia:

PR #10912, opened 2020-02-13
Fix LocalUfs.mkdir() for when parent dir is null

ggezer:


ggezer:

PR #10916, opened 2020-02-14
Exclude dependency on rocksdb from client jar

apc999:
Tested with my local build, client jar size reduced from 40MB to 26MB


calvinjia:

PR #10919, opened 2020-02-14
Allow superuser to stopActiveSync w/o a lock

madanadit:
A full scan for active sync acquires a write lock on the syncPoint which prevents the permission check for stopSync to proceed. This PR skips the read lock needed on the inode for the permission check if the user is a super user. This allows us to terminate a long-running full scan early.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -149,6 +149,7 @@ import alluxio.util.CommonUtils;
 import alluxio.util.IdUtils;
 import alluxio.util.ModeUtils;
 import alluxio.util.SecurityUtils;
+import alluxio.util.SleepUtils;
 import alluxio.util.UnderFileSystemUtils;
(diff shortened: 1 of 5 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3817,21 +3817,35 @@ public void startSync(AlluxioURI syncPoint)
...
       try {
-        mPermissionChecker.checkParentPermission(Mode.Bits.WRITE, inodePath);
+        mPermissionChecker.checkSuperUser();
       } catch (AccessControlException e) {
-        auditContext.setAllowed(false);
-        throw e;
+        isSuperUser = false;
+      }
+      if (isSuperUser) {
+        // Stop sync w/o acquiring an inode lock to terminate an initial full scan (if running)
```
gpang: can we add a TODO saying we can remove this once a sync doesn't require a write lock during the sync?

madanadit: done

madanadit:

PR #10922, opened 2020-02-14
Remove previous users from the hadoop subject

gpang:


core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -542,9 +542,12 @@ private Subject getHadoopSubject() throws IOException {
       LOG.warn("Hadoop subject does not exist. Creating a fresh subject for Alluxio client");
       subject = new Subject(false, new HashSet<>(), new HashSet<>(), new HashSet<>());
     }
-    if (subject.getPrincipals(CurrentUser.class).isEmpty() && ugi != null) {
-      subject.getPrincipals().add(new CurrentUser(ugi.getShortUserName(), mUri.toString()));
-    }
+
```
bf8086: Shall we synchronize on subject to avoid race conditions?

PR #10924, opened 2020-02-15
Change the behavior of UFS sync to delete unpersisted files when syncing

yuzhu:


yuzhu:
@ns1123 FYI

gpang:
@yuzhu can you describe what this is for? Also, can you add some tests for this?

yuzhu:
closing this because we do not want to change the existing sync behavior right now

PR #10930, opened 2020-02-18
Add more debug or info log to local cache

apc999:


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -76,6 +76,7 @@ public static LocalCacheManager create(AlluxioConfiguration conf) throws IOExcep
     PageStore pageStore = PageStore.create(conf);
     try {
       Collection<PageInfo> pageInfos = pageStore.getPages();
+      LOG.info("Loading {} existing pages", pageInfos.size());
```
gpang: Should this be something like "Creating LocalCacheManager with {} existing pages" ?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -143,6 +144,7 @@ private ReadWriteLock getPageLock(PageId pageId) {
 
   @Override
   public boolean put(PageId pageId, byte[] page) throws IOException {
+    LOG.debug("put(pageId={},page={} bytes)", pageId, page.length);
```
gpang: will the `PageId` instance already print out `PageId`? Will `pageId=` be redundant in the log?

apc999:
@gpang  @calvinjia 
PTAL

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -205,6 +208,7 @@ public boolean put(PageId pageId, byte[] page) throws IOException {
         mEvictor.updateOnPut(pageId);
       }
     }
+    LOG.debug("put({},{} bytes) exits with evicting ({})", pageId, page.length, victimPageInfo);
```
bf8086: Shall we differentiate when there is not enough space and the new page fails to add?

apc999: fixed . good catch

apc999:

apc999:

PR #10933, opened 2020-02-18
Fix metrics throughput to be per minute value

LuQQiu:
Change the throughput calculation to be per minute value instead of per second and use clock for easy testing


shell/src/main/java/alluxio/cli/fsadmin/report/MetricsCommand.java
```diff
@@ -74,8 +74,7 @@ public int run() throws IOException {
           // Bytes long can be transformed to human-readable format
           strValue = FormatUtils.getSizeFromBytes((long) doubleValue);
           if (name.contains(THROUGHPUT_METRIC_IDENTIFIER)) {
-            // throughput is calculated as one-minute exponentially-weighted moving average rate
-            strValue = strValue + "/min";
+            strValue = strValue + "/SEC";
```
ZacBlanco: nit: is there a reason to make this all caps?

LuQQiu: Because `FormatUtils.getSizeFromBytes()` will return size in caps like `B, MB, GB`

LuQQiu:
@ZacBlanco PTAL, thanks!

shell/src/main/java/alluxio/cli/fsadmin/report/MetricsCommand.java
ZacBlanco: also, is every gauge at the `sec` granularity? Is that something we even want?

I feel like there will be times when a cluster is not under a high load and many of the gauges will report as 0 - especially for people who scrape metrics on a periodic basis, second-level granularity seems like it would cause lots of information to be missed.

LuQQiu: All the cluster throughput is at `sec` now. Throughput can be of sec or minute. Yeah, maybe use `MIN` make more sense since alluxio is usually long-running cluster.

LuQQiu:
@ZacBlanco  PTAL, thanks!

gpang:
@LuQQiu can you update the PR description? It is not clear what changed, since it is per minute throughput before and after this PR?

LuQQiu:
@gpang Previously the throughput is per second value but show as per minute value in metrics command. This PR change it to be actual per minute value. Thanks

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -209,7 +223,7 @@ public void clear() {
       for (Counter counter : mClusterCounters.values()) {
         counter.dec(counter.getCount());
       }
-      mLastClearTime = System.currentTimeMillis();
+      mLastClearTime = mClock.millis();
```
ZacBlanco: what's the advantage to using `mClock` vs `System.currentTimeMillis`?

core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
```diff
@@ -141,9 +141,11 @@ protected void registerThroughputGauge(String counterName, String throughputName
...
           @Override
           public Object getValue() {
-            long uptime = (System.currentTimeMillis() - mMetricsStore.getLastClearTime())
-                / Constants.SECOND_MS;
-            return uptime == 0 ? 0 : MetricsSystem.counter(counterName).getCount() / uptime;
+            long uptime = (mClock.millis() - mMetricsStore.getLastClearTime())
+                / Constants.MINUTE_MS;
+            long value = MetricsSystem.counter(counterName).getCount();
+            // The value is bytes per minute
+            return uptime == 0 ? value : value / uptime;
```
ZacBlanco: hmm if uptime is 0 should we be returning `value`? or `0`?

In the off chance of some kind of error, what if uptime is negative?

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
LuQQiu: They are nearly the same, but mClock is good for testing, I can manually modify the time in unit tests instead of waiting for time to passed.

core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
LuQQiu: Oh, that should not happen. mClock.mills() should always be equal or bigger than the mMetricsStore.getLastClearTime()). 
I feel returning value in the first minute actually makes more sense. I am fine with both 0 or value just feeling that using value can even provide users information in the first minute so it's better.

LuQQiu: wait, actually uptime can be negative..... 
Let me modify the code

LuQQiu: Changed the code to make sure uptime is positive. Thanks for finding out that!!

core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java
```diff
@@ -141,9 +141,13 @@ protected void registerThroughputGauge(String counterName, String throughputName
         new Gauge<Object>() {
           @Override
           public Object getValue() {
-            long uptime = (System.currentTimeMillis() - mMetricsStore.getLastClearTime())
-                / Constants.SECOND_MS;
-            return uptime == 0 ? 0 : MetricsSystem.counter(counterName).getCount() / uptime;
+            // Divide into two lines so uptime is always zero or positive
+            long lastClearTime = mMetricsStore.getLastClearTime();
+            long uptime = (mClock.millis() - lastClearTime)
```
ZacBlanco: I'm not sure if this will still fix the issue?

A java `Clock` object can have an offset added to it (See https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html). If the duration is negative it could cause this to return negative for an extended period of time.

I don't think we use the `offset` method anywhere, but it would be good to guard against it. I think keeping the code on one line it fine.

Maybe instead we can simply update the `return uptime == 0 ? ` to `return uptime <= 0 ? value : value / uptime`

or actually - since we should simply return the value within the first minute, maybe it should be 

```java
return uptime <= 60 * Constants.SECOND_MS ? value : value / uptime
```

LuQQiu: Clock.offset can be set, but usually it's in tests only. 
Changed to set uptime <= 0

LuQQiu:
@ZacBlanco PTAL, thanks, update the uptime <= 0 condition

LuQQiu:

LuQQiu:

PR #10936, opened 2020-02-18
Make the addition of client metrics heartbeats asynchronous

LuQQiu:
When client metrics is enabled, the master starts up takes four more minutes. The four more minutes are taken by `PollingMasterInquireClient.getPrimaryRpcAddress`  in `FileSystemContext.initContext`. `getPrimaryRpcAddress()` retries 44 times and takes 2 minutes each call which heavily delays the master starts up. The primary rpc address is used for load configuration from the leading master and initialize metrics heartbeat.

This PR put the load configuration logics in the metrics heartbeat thread, so that client metrics logic in FileSystemContext will not block actual operations.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -231,15 +231,6 @@ public class FileSystemContext implements Closeable {
         .setMasterInquireClient(masterInquireClient).build();
     mMetricsEnabled = getClusterConf().getBoolean(PropertyKey.USER_METRICS_COLLECTION_ENABLED);
     if (mMetricsEnabled) {
-      try {
-        InetSocketAddress masterAddr = masterInquireClient.getPrimaryRpcAddress();
-        mMasterClientContext.loadConf(masterAddr, true, true);
-      } catch (UnavailableException e) {
-        LOG.error("Failed to get master address during initialization", e);
-      } catch (AlluxioStatusException ae) {
-        LOG.error("Failed to load configuration from "
-            + "meta master during initialization", ae);
-      }
       MetricsSystem.startSinks(getClusterConf().get(PropertyKey.METRICS_CONF_FILE));
       MetricsHeartbeatContext.addHeartbeat(getClientContext(), masterInquireClient);
     }
--- a/core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
@@ -71,7 +71,14 @@ public final class ClientMasterSync {
    */
   public synchronized void heartbeat() {
     if (mMasterClient == null) {
-      loadConfAndInitClient();
+      if (loadConf()) {
+        mMasterClient = new RetryHandlingMetricsMasterClient(MasterClientContext
+            .newBuilder(mContext)
+            .setMasterInquireClient(mInquireClient)
+            .build());
+      } else {
+        return; // not heartbeat when failed to load conf
+      }
     }
     // TODO(zac): Support per FileSystem instance metrics
     // Currently we only support JVM-level metrics. A list is used here because in the near
@@ -105,21 +112,22 @@ public final class ClientMasterSync {
   }
 
   /**
-   * Loads configuration and initialize metrics master client.
+   * Loads configuration.
+   *
+   * @return true if successfully loaded configuration
    */
-  private void loadConfAndInitClient() {
+  private boolean loadConf() {
     try {
       InetSocketAddress masterAddr = mInquireClient.getPrimaryRpcAddress();
-      mContext.loadConf(masterAddr, true, true);
+      mContext.loadConf(masterAddr, true, false);
     } catch (UnavailableException e) {
       SAMPLING_LOG.error("Failed to get master address during initialization: {}", e.toString());
+      return false;
     } catch (AlluxioStatusException ae) {
       SAMPLING_LOG.error("Failed to load configuration from "
           + "meta master during initialization: {}", ae.toString());
+      return false;
     }
-    mMasterClient = new RetryHandlingMetricsMasterClient(MasterClientContext
-        .newBuilder(mContext)
-        .setMasterInquireClient(mInquireClient)
-        .build());
+    return true;
   }
 }
--- a/core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
@@ -11,9 +11,13 @@
 
 package alluxio.client.metrics;
 
+import alluxio.ClientContext;
 import alluxio.Constants;
-import alluxio.conf.AlluxioConfiguration;
+import alluxio.exception.status.AlluxioStatusException;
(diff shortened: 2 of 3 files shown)


LuQQiu:
@calvinjia PTAL, thanks

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -231,14 +232,16 @@ private synchronized void initContext(ClientContext ctx,
...
     if (mMetricsEnabled) {
-      try {
-        InetSocketAddress masterAddr = masterInquireClient.getPrimaryRpcAddress();
-        mMasterClientContext.loadConf(masterAddr, true, true);
-      } catch (UnavailableException e) {
-        LOG.error("Failed to get master address during initialization", e);
-      } catch (AlluxioStatusException ae) {
-        LOG.error("Failed to load configuration from "
-            + "meta master during initialization", ae);
+      if (CommonUtils.PROCESS_TYPE.get() != CommonUtils.ProcessType.MASTER) {
```
calvinjia: Does it make sense to report client metrics at all from the master?

LuQQiu: Yeah, it makes sense to not report client metrics at all from the master. If not reporting, we may lose some of the client metrics, but client metrics in the master process is wrong no matter what. This is because of the client metrics in master may record as `Master.Client.metricName` which will not be reported. Will add a TODO to solve master client metrics in the future.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -230,8 +231,11 @@ private synchronized void initContext(ClientContext ctx,
     mMasterClientContext = MasterClientContext.newBuilder(ctx)
         .setMasterInquireClient(masterInquireClient).build();
     mMetricsEnabled = getClusterConf().getBoolean(PropertyKey.USER_METRICS_COLLECTION_ENABLED);
-    if (mMetricsEnabled) {
+    if (mMetricsEnabled && CommonUtils.PROCESS_TYPE.get() != CommonUtils.ProcessType.MASTER) {
+      // TODO(lu) report clients metrics in master process
       try {
+        // clients may be used in the master process before master starts.
+        // Retry getting master addresses in this circumstance may block master starting up
         InetSocketAddress masterAddr = masterInquireClient.getPrimaryRpcAddress();
```
ZacBlanco: I don't understand why this would take longer if the process is the master. Is there some kind of race condition?

ZacBlanco: I think that we shouldn't make the master startup depend upon having Alluxio clients available. If anything, I think this should be a change that we make on the catalog and WebServer side to make sure that a client doesn't need to connect before the RPC server starts?

gpang:
Jenkins, test this please

LuQQiu:
Discussed with @ZacBlanco, Metrics should not block any actual operations. Is working on the make metrics logics in the `FileSystemContext.initContext` async.

LuQQiu:
@calvinjia @ZacBlanco PTAL, this PR changed to process add/remove client metrics heartbeat async so that it will not block normal operations, thanks

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -177,13 +206,43 @@ public static synchronized void addHeartbeat(ClientContext ctx,
           sAppId, PropertyKey.Name.USER_APP_ID);
     }
 
+    if (MetricsSystem.getNumSinks() == 0) {
+      MetricsSystem.startSinks(ctx.getClusterConf().get(PropertyKey.METRICS_CONF_FILE));
+    }
```
calvinjia: It seems out of place to start the metric system in the heartbeat.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -231,17 +230,7 @@ private synchronized void initContext(ClientContext ctx,
...
-        mMasterClientContext.loadConf(masterAddr, true, true);
-      } catch (UnavailableException e) {
-        LOG.error("Failed to get master address during initialization", e);
-      } catch (AlluxioStatusException ae) {
-        LOG.error("Failed to load configuration from "
-            + "meta master during initialization", ae);
-      }
-      MetricsSystem.startSinks(getClusterConf().get(PropertyKey.METRICS_CONF_FILE));
-      MetricsHeartbeatContext.addHeartbeat(getClientContext(), masterInquireClient);
+      MetricsHeartbeatContext.asyncAddHeartbeat(getClientContext(), masterInquireClient);
```
calvinjia: Is it necessary to start the metrics system asynchronously? Only the heartbeat needs the master address right? Instead of computing master address, could we compute it lazily during the heartbeat call (and thus use the heartbeat thread, making the submission of the heartbeat thread lightweight)?

core/client/fs/src/main/java/alluxio/client/metrics/MetricsHeartbeatContext.java
```diff
@@ -52,6 +59,7 @@
 @ThreadSafe
 public class MetricsHeartbeatContext {
   private static final Logger LOG = LoggerFactory.getLogger(MetricsHeartbeatContext.class);
+  private static final long SHUTDOWN_TIMEOUT_MS = 10 * Constants.SECOND_MS;
```
ZacBlanco: 10 seconds seems like a long shutdown timeout to me. For metrics, I would prefer it to have as little impact as possible on the shutdown of the JVM. Maybe 2-3 seconds is better?

core/client/fs/src/main/java/alluxio/client/metrics/ClientMasterSync.java
```diff
@@ -85,4 +94,32 @@ public synchronized void heartbeat() {
...
+    try {
+      InetSocketAddress masterAddr = mInquireClient.getPrimaryRpcAddress();
+      mContext.loadConf(masterAddr, true, true);
+    } catch (UnavailableException e) {
+      SAMPLING_LOG.error("Failed to get master address during initialization: {}", e.toString());
+    } catch (AlluxioStatusException ae) {
+      SAMPLING_LOG.error("Failed to load configuration from "
+          + "meta master during initialization: {}", ae.toString());
+    }
+    mMasterClient = new RetryHandlingMetricsMasterClient(MasterClientContext
```
calvinjia: Should we still create `mMasterClient` when we fail above? Since heartbeat will be called periodically, it might be easier to just no-op this heartbeat and try again next time?

LuQQiu: Make sense, now retry load conf if fail and make heartbeat no-op if load conf failed

LuQQiu:
@calvinjia Addressed the comments, PTAL, thanks

ZacBlanco:
@LuQQiu can you update the original PR description to more accurately reflect the change in this PR. At the end you can note that the original issue was that the master was starting slowly due to needing to loadConf before the RPC server started.

LuQQiu:

LuQQiu:

PR #10939, opened 2020-02-19
Add worker client relogin

bf8086:
Currently only master client does relogin when needed. There is a slight chance that worker client will run into credential expiration therefore the same relogin logic should be also in the worker client. In the future we should unify the two clients.


gpang:
Jenkins, test this please

bf8086:
> @bf8086 Thanks! I'm assuming this is similar to what the other client does for retries?
Yes it is a shrinked down version for only relogin purpose. It should be refactored to use AbstractClient logic in the future.

bf8086:

bf8086:

PR #10948, opened 2020-02-19
Fix coalesce max files behavior

bradyoo:
The compression ratio for the compact job was not applied correctly, resulting in unexpected file sizes.

### DIFF
--- a/job/server/src/main/java/alluxio/job/plan/transform/CompactDefinition.java
+++ b/job/server/src/main/java/alluxio/job/plan/transform/CompactDefinition.java
@@ -96,14 +96,15 @@ public final class CompactDefinition
     Map<WorkerInfo, ArrayList<CompactTask>> assignments = Maps.newHashMap();
     int maxNumFiles = config.getMaxNumFiles();
     long groupMinSize = config.getMinFileSize();
-    if (totalFileSize / groupMinSize > maxNumFiles) {
-      groupMinSize = Math.round(totalFileSize / maxNumFiles);
-    }
 
     if (!files.isEmpty() && config.getPartitionInfo() != null) {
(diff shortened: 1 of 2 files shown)


job/server/src/main/java/alluxio/job/plan/transform/CompactDefinition.java
```diff
@@ -96,16 +96,17 @@ private boolean shouldIgnore(URIStatus status) {
...
-      groupMinSize = Math.round(totalFileSize / maxNumFiles);
-    }
 
     if (!files.isEmpty() && config.getPartitionInfo() != null) {
       // adjust the group minimum size for source compression ratio
       groupMinSize *= COMPRESSION_RATIO.get(
           config.getPartitionInfo().getFormat(files.get(0).getName()));
     }
 
+    if (totalFileSize / groupMinSize > maxNumFiles) {
```
yuzhu: Wouldn't this override the `groupMinSize` variable and discard the compression ratio? 
what was the observed symptom in testing?

bradyoo: The problem with the current behavior is that when you have a CSV input and you say I want 100 files with groupMinSize small enough that I should be able to get 100 files, you instead get 20 files.

bradyoo: compressionRatio should not be downsizing maxNumFiles.

yuzhu: ah makes sense.

gpang:

gpang:

PR #10953, opened 2020-02-20
Make metastore to track pagestore size

apc999:
This PR has two major changes:
- Instead of querying pagestore for size to make eviction decisions, querying metastore
- Update & simplify restoring logic accordingly. Now only one pass is needed


apc999:
Jenkins, test this please

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -50,70 +45,44 @@
 @NotThreadSafe
 public class RocksPageStore implements PageStore {
   private static final Logger LOG = LoggerFactory.getLogger(RocksPageStore.class);
-  public static final int KEY_LEN = Long.BYTES * 2;
   private static final byte[] CONF_KEY = "CONF".getBytes();
+  // TODO(feng): consider making the overhead ratio configurable
+  // We assume 20% overhead using Rocksdb as a page store, i.e., with 1GB space allocated, we
+  // expect no more than 800MB logical data stored
```
calvinjia: This comment is not consistent with the calculation (ie. 1000 / 1.2 = 833)

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -179,25 +146,31 @@ private static PageId getPageIdFromKey(byte[] key) {
...
+          mDb.close();
+          return false;
+        }
+      }
+      mDb.put(CONF_KEY, pOptions.toByteArray());
+      try (RocksIterator iter = mDb.newIterator()) {
+        return Streams.stream(new PageIterator(iter)).allMatch(initFunc::test);
+      }
+    } catch (RocksDBException | IOException e) {
+      LOG.error("Failed to restore RocksPageStore:", e);
```
bf8086: We should close the db whenever restore returns false.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -77,24 +77,27 @@
...
-      } catch (Exception ex) {
-        e.addSuppressed(ex);
-      }
-      throw new IOException("failed to create local cache manager", e);
+      LOG.error("Failed to restore PageStore", e);
+    }
+    if (!restored) {
+      metaStore.reset();
+      evictor.reset();
+      pageStore.initialize(options);
```
bf8086: RocksPageStore will be in bad state after restore then initialize.

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -170,27 +136,16 @@ ReadableByteChannel get(PageId pageId, int pageOffset) throws IOException,
...
-   * Gets all page ids.
+   * Restores the page store from a previous run.
    *
-   * @return collection of ids representing all pages loaded from disk
+   * @param initFunc function to apply during restore process
+   * @return true if successfully restored from previous state
    * @throws IOException if any error occurs
    */
-  Collection<PageInfo> getPages() throws IOException;
+  boolean restore(Predicate<PageInfo> initFunc) throws IOException;
```
bf8086: We need to indicate that this function must be called right after the constructor. The state of the page store is undetermined before this function is called.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
apc999: done

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: initialize will remove all files. will it still be a problem?

bf8086: The issue is that initialize is generic and it does not initialize resource specific to an implementation. In RocksDB page store, the mDB will be stale (depends on the restore result) and unusable after the initialize clean up the directory.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -180,25 +147,29 @@ private static PageId getPageIdFromKey(byte[] key) {
...
+      if (confData != null) {
+        Cache.PRocksPageStoreOptions persistedOptions =
+            Cache.PRocksPageStoreOptions.parseFrom(confData);
+        if (!persistedOptions.equals(pOptions)) {
+          mDb.close();
+          throw new IOException("Inconsistent configuration for RocksPageStore");
+        }
+      }
+      mDb.put(CONF_KEY, pOptions.toByteArray());
+      return Streams.stream(new PageIterator(mDb.newIterator()));
```
calvinjia: What happens when we call close on this stream, is it a no-op?

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -171,27 +183,16 @@ ReadableByteChannel get(PageId pageId, int pageOffset) throws IOException,
...
-  long pages();
-
-  /**
-   * @return the total size of pages stored in bytes
-   */
-  long bytes();
-
-  /**
-   * Gets all page ids.
+   * Gets a stream of all pages from the page store. This stream needs to be closed as it may
```
bf8086: (nit) Given all other stream usages do not require explicit closing, it might be hard for people to remember to close it.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -77,24 +77,19 @@
...
-        e.addSuppressed(ex);
-      }
-      throw new IOException("failed to create local cache manager", e);
+      LOG.error("Failed to restore PageStore", e);
+    }
+    if (pageStore == null) {
+      metaStore.reset();
+      evictor.reset();
+      pageStore = PageStore.create(options);
     }
```
bf8086: (nit) Would it be simpler if we just do it in the previous catch block?

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -45,51 +43,66 @@
...
-        options = new RocksPageStoreOptions();
+        pageStore = new RocksPageStore(options.toOptions());
         break;
       default:
-        throw new IllegalArgumentException(String.format("Unrecognized store type %s",
-            storeType.name()));
+        throw new IllegalArgumentException(
+            "Incompatible PageStore " + options.getType() + " specified");
+    }
+    if (metaStore != null && evictor != null) {
```
bf8086: Would it be more clear if we just move the logic before this line to `create(options)` and just call it first?

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
```diff
@@ -180,25 +147,29 @@ private static PageId getPageIdFromKey(byte[] key) {
...
-      return Streams.stream(new PageIterator(iter)).collect(Collectors.toList());
+  public Stream<PageInfo> getPages() throws IOException {
+    try {
+      byte[] confData = mDb.get(CONF_KEY);
+      Cache.PRocksPageStoreOptions pOptions = mOptions.toProto();
+      if (confData != null) {
+        Cache.PRocksPageStoreOptions persistedOptions =
+            Cache.PRocksPageStoreOptions.parseFrom(confData);
+        if (!persistedOptions.equals(pOptions)) {
+          mDb.close();
```
bf8086: This is a pretty strong side effect from a function that looks like a getter. Shall we either rename this function or make it protected and close the db upon exception from `PageStore.create()`?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -77,24 +77,19 @@
...
-      return new LocalCacheManager(conf, metaStore, pageStore, evictor);
+      pageStore = PageStore.create(options, metaStore, evictor);
     } catch (Exception e) {
-      try {
-        pageStore.close();
-      } catch (Exception ex) {
-        e.addSuppressed(ex);
-      }
-      throw new IOException("failed to create local cache manager", e);
+      LOG.error("Failed to restore PageStore", e);
```
bf8086: After restoration failed, we will need to clean the store directory before creating a new store.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
apc999: make sense. Moved the conf verification part to factory method

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
apc999: correct

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: in this case, passing `init=true` to create PageStore in the following.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: i feel that is not very readable

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
```diff
@@ -45,51 +43,66 @@
...
    */
-  static PageStore create(AlluxioConfiguration conf) throws IOException {
-    PageStoreOptions options;
-    PageStoreType storeType = conf.getEnum(
-        PropertyKey.USER_CLIENT_CACHE_STORE_TYPE, PageStoreType.class);
-    switch (storeType) {
+  static PageStore create(PageStoreOptions options, @Nullable MetaStore metaStore,
+        @Nullable CacheEvictor evictor) throws IOException {
+    LOG.info("Create PageStore option={}", options.toString());
+    if (metaStore == null || evictor == null) {
```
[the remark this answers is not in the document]
apc999: this is not priority currently but only making the code more complicated.
I would defer the implementation

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
apc999: not sure I understand the comment

core/client/fs/src/main/java/alluxio/client/file/cache/PageStore.java
apc999: perhaps. but we should either use iterator or stream here because we will need to handle ~1 million pages restored, that could take ~GB memory if we use collection. due to this reason, I and @calvinjia  after discussion decided to stick to stream.

core/client/fs/src/main/java/alluxio/client/file/cache/store/RocksPageStore.java
apc999: fixed by adding closer to `onClose`

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -51,59 +45,24 @@
...
-        LOG.warn("Cannot recover from cached data: {}",
-            invalidPage ? "Invalid page file found" : "Cached data size exceeded configured value");
-        FileUtils.cleanDirectory(new File(mRoot));
-        mSize.set(0);
-        mBytes.set(0);
-      }
-    } catch (IOException e) {
-      throw new IOException(String.format("can't initialize page store at %s", mRoot), e);
-    }
+        String.format("%s/%d/(\\d+)/([^/]+)/(\\d+)", Pattern.quote(mRoot), mPageSize));
```
bf8086: It can be tricky for user to deal with trailing slash if we use the configured path without normalizing it.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -71,31 +76,70 @@
...
+    }
+    if (!restored) {
+      metaStore.reset();
+      evictor.reset();
+      pageStore = PageStore.create(options, true);
```
bf8086: Should the previous page store be closed before we create a new one?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
bf8086: The current `initialize()` function does not seem to clean up the store path of the current store type.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
apc999: reverted  and add a javadoc preventing future change

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: fixed

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: added

apc999:
@bf8086  PTAL

apc999:

apc999:

PR #10972, opened 2020-02-21
[SMALLFIX] Fix AFS.getFileBlockLocations when file is null

bradyoo:


core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -259,10 +259,11 @@ public long getDefaultBlockSize() {
   @Override
   public BlockLocation[] getFileBlockLocations(FileStatus file, long start, long len)
       throws IOException {
-    LOG.debug("getFileBlockLocations({}, {}, {})", file.getPath().getName(), start, len);
+    LOG.debug("getFileBlockLocations({}, {}, {})",
```
gpang: shouldn't this just be the same as old code, but moved to after the `null` check?

bradyoo: Neh. Because it seems like there is a debug message for when the code returns AND when the code gets called. (Note the LOG.debug further down this code)

bradyoo:
ping?

bradyoo:
@gpang can I merge this?

bradyoo:
ping on merge? @gpang

gpang:

apc999:

PR #11001, opened 2020-02-24
Refactor CommonUtils.waitFor to use value supplier

bradyoo:
We have many instances in our code where we use CommonUtils.waitFor to wait for a certain condition to be equal to something but the last state of the object is not clear when there is a Timeout. This code has a new API that could be opted into to get such a benefit.


core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -321,32 +323,53 @@ public static void waitFor(String description, Supplier<Boolean> condition)
...
+    T value;
     long start = System.currentTimeMillis();
     int interval = options.getInterval();
     int timeout = options.getTimeoutMs();
-    while ((t = operation.get()) == null) {
+    while (condition.apply(value = objectSupplier.get()) != true) {
       if (timeout != WaitForOptions.NEVER && System.currentTimeMillis() - start > timeout) {
-        throw new TimeoutException("Timed out waiting for " + description + " options: " + options);
+        throw new TimeoutException("Timed out waiting for " + description + " options: " + options
+            + " last value: " + ObjectUtils.toString(value));
```
bradyoo: This is the main benefit I want.

bradyoo:
ping? @gpang

gpang:

PR #11016, opened 2020-02-25
Fix "In Alluxio" calculation for DuCommand

ZacBlanco:
Previously, we were always using the "InMemoryPercentage" to calculate the
total amount in Alluxio. This is incorrect. We will now use the proper value

### DIFF
--- a/shell/src/test/java/alluxio/cli/fs/command/DuCommandTest.java
+++ b/shell/src/test/java/alluxio/cli/fs/command/DuCommandTest.java
@@ -1,0 +1,121 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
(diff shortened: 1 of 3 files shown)


apc999:
02:08:36.427 [WARNING] Missing header in: /usr/src/alluxio/shell/src/test/java/alluxio/cli/fs/command/DuCommandTest.java

shell/src/test/java/alluxio/cli/fs/command/DuCommandTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class DuCommandTest {
```
apc999: up to you but we have DuIntegrationTest if you want to consolidate

ZacBlanco: I am aware but I specifically made this test because running a full alluxio cluster is unnecessary in order to test this functionality. The time it would require to run this in an integration test is probably 10x higher than using a unit test - and we should work to reduce the execution times of our tests because they have been jumping higher recently.

ZacBlanco:

ZacBlanco:

ZacBlanco:

PR #11036, opened 2020-02-27
Add log server log env variable to template

madanadit:


ZacBlanco:
Why are we adding this? If I vaguely remember there were some deprecated environment variables that I removed in 2.0. I think this was one of them...

madanadit:
@ZacBlanco hmm, looks like the documentation is still pointing to these env variables are the recommended way to configure remote logging. are you familiar w/ the alternative?

ZacBlanco:
Let go back through my commits and see if this was actually one of the removed variables. If it is, then I'll let you know about the workaround

madanadit:
@ZacBlanco I found it: https://docs.alluxio.io/os/user/edge/en/reference/Properties-List.html#alluxio.logserver.hostname. Looks like we recommend using `ALLUXIO_JAVA_OPTS` now. I'll fix the docs

PR #11055, opened 2020-02-27
Build binaries for OS package managers

ZacBlanco:
This change adds some scripts that should allow us to build `deb` and `rpm` packages. The only requirements are `docker`, `bash`, and any valid alluxio tarball.

To build a deb package:

```console
$ ./dev/packaging/build.sh <alluxio_tarball>
```

To build an RPM

```console
$ ./dev/packaging/convert.sh <alluxio_deb_pkg>
```

Some minor TODOs off the top of my head that can make the life of users easier with the package:

- Need an official GPG key to sign packages
- Once hosting is sorted out, automatically add our repo to the system's `/etc/apt/sources.list.d/10_alluxio.list` so that once installed, updates are automatically applied. Same for the RPM - we can add a repo to `/etc/yum.repos.d/...`
- Create some `alluxio-*.service` files that can be used with SysVInit or systemd, so that the service can run as a part of the operating system
- Define `${ALLUXIO_HOME}` in the env by default


ZacBlanco:
@LuQQiu since you have some experience in building packages, could you take a look? Thanks :)

ZacBlanco:
Thanks for pointing this out. I thought it was more standard

Either way, I actually think it's possible to remove the usage of `realpath` all together. I will push an update soon

ZacBlanco:
jenkins, test this please

PR #11076, opened 2020-03-01
Log a cancelled message when sync is cancelled due to exception

yuzhu:
Previously, when a sync is canceled on the last directory to be synced, no messages are logged. Now it will be logged.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3396,6 +3396,10 @@ public final class DefaultFileSystemMaster extends CoreMaster
 
     // Update metadata for all the mount points
     for (String mountPoint : pathsToLoad) {
+      if (Thread.currentThread().isInterrupted()) {
+        LOG.warn("Thread syncing {} was interrupted before completion", inodePath.getUri());
+        return false;
+      }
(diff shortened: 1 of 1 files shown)


yuzhu:
@madanadit PTAL

ZacBlanco:
@yuzhu can you update the PR description with the explanation about how this changes resolves the issue?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3396,10 +3396,6 @@ private boolean syncMetadataInternal(RpcContext rpcContext, LockedInodePath inod
 
     // Update metadata for all the mount points
     for (String mountPoint : pathsToLoad) {
-      if (Thread.currentThread().isInterrupted()) {
```
madanadit: can we keep both if conditions? if we cancel before syncing the first path, we would sync the first path with this change.

yuzhu:
@madanadit comment addressed

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
yuzhu: done

yuzhu:

yuzhu:

PR #11082, opened 2020-03-02
Fix config check with default values

ZacBlanco:
This PR aims to fix the configuration check in the Alluxio UI when running locally and all values are not configured except for `alluxio.master.hostname` set to `localhost`

By default this should be passing, or else the user may think they have configured something incorrectly.

There are two main changes.

1. Fix a failing check from a PK with `ConsistencyCheckLevel.ENFORCE` that gets it's trailing spaces trimmed when the pk is sent over the wire. I just removed the extra trailing spaces as the usage didn't seem to require it.

2. Use the resolved value for the PK rather than the raw value. There are many PKs with "warnings" because the check on the master uses the resolved value rather than the raw value sent from the other masters/workers. I don't really see a good use for using the raw value directly, so all config values sent over the wire are resolved if they use something like `${pk.name}` on the remote process.


ZacBlanco:

ZacBlanco:

PR #11108, opened 2020-03-03
Improve how tests handle restarting workers

gpang:
Instead of relying on waiting for lost worker detection, force the workers to be lost.

### DIFF
--- a/minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java
+++ b/minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java
@@ -259,9 +259,12 @@ public abstract class AbstractLocalAlluxioCluster {
     mWorkerThreads.clear();
 
     // forget all the workers in the master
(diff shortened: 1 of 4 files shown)


minicluster/src/main/java/alluxio/master/AbstractLocalAlluxioCluster.java
```diff
@@ -255,6 +257,11 @@ public void stopWorkers() throws Exception {
       }
     }
     mWorkerThreads.clear();
+
+    // forget all the workers in the master
+    DefaultBlockMaster bm = (DefaultBlockMaster) getLocalAlluxioMaster().getMasterProcess()
+        .getMaster(BlockMaster.class);
+    bm.forgetAllWorkers();
```
ZacBlanco: Should we check if the master is started here/`getMasterProcess()` returns null?

I didn't check the method signature but I think I vaguely remember that it could be possible

gpang: it checks for `null` now

ZacBlanco:
jenkins, test this please

gpang:

PR #11129, opened 2020-03-05
Improve multi-tier block management

### LINKED ISSUE #9796: Eviction and Allocator on multi-tier storage not working
**Alluxio Version:**
2.0.1

**Describe the bug**
Multi-Tier storage 3x MEM(25GB/SSD 24GB /HDD 99GB) on 8 cores 32GB 10GiB network is not evicting data from MEM to the configured SSD.

`SpaceReserver failed to free 20132659200 bytes on tier MEM for high watermarks: Failed to find an eviction plan to free 20132659200 bytes space at location MEM `

It does look very much as this area of the code points into a looping issues.

```TieredBlockStore.java
/**
   * Tries to get an eviction plan to free a certain amount of space in the given location, and
   * carries out this plan with the best effort.
   *
   * @param sessionId the session id
   * @param availableBytes amount of space in bytes to free
   * @param location location of space
   * @param mode the eviction mode
   * @throws WorkerOutOfSpaceException if it is impossible to achieve the free requirement
   */
  private void freeSpaceInternal(long sessionId, long availableBytes, BlockStoreLocation location,
      Evictor.Mode mode) throws WorkerOutOfSpaceException, IOException {
    EvictionPlan plan;
    // NOTE:change the read lock to the write lock due to the endless-loop issue [ALLUXIO-3089]
    try (LockResource r = new LockResource(mMetadataWriteLock)) {
      plan = mEvictor.freeSpaceWithView(availableBytes, location, getUpdatedView(), mode);
      // Absent plan means failed to evict enough space.
      if (plan == null) {
        throw new WorkerOutOfSpaceException(
            ExceptionMessage.NO_EVICTION_PLAN_TO_FREE_SPACE, availableBytes, location.tierAlias());
      }
    }
```

**To Reproduce**
Standup Standalone cluster with 3 nodes - and write a distributed workload against Alluxio Cluster.

**Expected behavior**
Data is evicted from highest tier into the next configured tier

**Urgency**
Pretty much unusable for IO aware workloads and contradicts the purpose of a multi-tier storage system. 

**Additional context**
This is a fairly light workload just writing some files from multiple nodes.
Workload ran into storage limit on MEM:
image (7)

restart one of the 3 nodes: (eviction happens as configured)
image (6)

the individual node (now correctly rebalanced after restart) - Also the ORDER of the configured storage levels is UPSIDEDOWN (when you go to the individual worker).

image (9)

and the other two nodes ... not restarted no active workload running:
image (10)
image (11)

for simplicity you can also recreate the same issue by just configuring it without memory requirements as above and just use a two tier setup.
```
alluxio.user.file.readtype.default=CACHE_PROMOTE

alluxio.worker.allocator.class=alluxio.worker.block.allocator.MaxFreeAllocator
alluxio.worker.evictor.class=alluxio.worker.block.evictor.LRUEvictor
alluxio.worker.tieredstore.reserver.enabled=true

alluxio.worker.tieredstore.levels=2

alluxio.worker.tieredstore.level0.alias=SSD
alluxio.worker.tieredstore.level0.dirs.path=/ssd
alluxio.worker.tieredstore.level0.dirs.quota=5GB
alluxio.worker.tieredstore.level0.watermark.high.ratio=0.70
alluxio.worker.tieredstore.level0.watermark.low.ratio=0.5
alluxio.worker.tieredstore.level1.alias=HDD
alluxio.worker.tieredstore.level1.dirs.path=/hdd
alluxio.worker.tieredstore.level1.dirs.quota=100GB
alluxio.worker.tieredstore.level1.watermark.high.ratio=0.9
alluxio.worker.tieredstore.level1.watermark.low.ratio=0.75
```

### LINKED ISSUE #9804: Failed to load data into alluxio when data directory is pinned
**Alluxio Version:**

2.1.0-SNAPSHOT

**Describe the bug**

1. Create two layers in alluxio: MEM and SSD

```
# /opt/alluxio/bin/alluxio fsadmin report capacity
Capacity information for all workers:
    Total Capacity: 320.00GB
        Tier: MEM  Size: 120.00GB
        Tier: SSD  Size: 200.00GB
    Used Capacity: 118.05GB
        Tier: MEM  Size: 118.05GB
        Tier: SSD  Size: 0B
    Used Percentage: 36%
    Free Percentage: 64%

Worker Name      Last Heartbeat   Storage       Total            MEM           SSD
192.168.0.118    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          28.29GB (35%)    28.29GB       0B
192.168.0.195    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          29.92GB (37%)    29.92GB       0B
192.168.0.117    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          29.92GB (37%)    29.92GB       0B
192.168.0.194    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          29.92GB (37%)    29.92GB       0B
```

2. Pin the data directory

```
/opt/alluxio/bin/alluxio fs pin /training-data/images
```

3. Load the data into alluxio with distributedLoad

```
# /opt/alluxio/bin/alluxio fs -Dalluxio.user.ufs.block.read.location.policy=alluxio.client.block.policy.DeterministicHashPolicy  distributedLoad --replication 3 /training-data/images/
............................Failed to successfully complete the job.

```

4. 

From logging of the master

```
kubectl logs alluxio-master-0 -c alluxio-master
2019-08-28 03:25:58,460 WARN  SaslStreamServerDriver - Error received for channel: 1ce411a7-bef2-44bc-9b69-c05242ed1671. Error: io.grpc.StatusRuntimeException: CANCELLED: cancelled before receiving half close
```

From logging from the job master

```
2019-08-28 03:25:02,002 INFO  JobCoordinator - Starting job LoadConfig{FilePath=/training-data/images/train-00585-of-01024, Replication=1}
2019-08-28 03:25:02,003 WARN  JobCoordinator - No executor was selected.
2019-08-28 03:25:02,012 INFO  JobCoordinator - Starting job LoadConfig{FilePath=/training-data/images/train-00622-of-01024, Replication=1}
2019-08-28 03:25:22,052 INFO  JobCoordinator - Starting job LoadConfig{FilePath=/training-data/images/train-00622-of-01024, Replication=1}
2019-08-28 03:25:34,076 INFO  JobCoordinator - Starting job LoadConfig{FilePath=/training-data/images/train-00622-of-01024, Replication=1}
2019-08-28 03:25:46,099 INFO  JobCoordinator - Starting job LoadConfig{FilePath=/training-data/images/train-00622-of-01024, Replication=1}
```

5. In the worker, it is failed to allocate the space in the MEM. But it doesn't turn to SSD tieredstore

```
2019-08-28 07:54:13,044 WARN  SpaceReserver - SpaceReserver failed to free 9663676416 bytes on tier MEM for high watermarks: Failed to find an eviction plan to free 9663676416 bytes space at location MEM
2019-08-28 07:54:14,044 WARN  SpaceReserver - SpaceReserver failed to free 9663676416 bytes on tier MEM for high watermarks: Failed to find an eviction plan to free 9663676416 bytes space at location MEM
2019-08-28 07:54:15,044 WARN  SpaceReserver - SpaceReserver failed to free 9663676416 bytes on tier MEM for high watermarks: Failed to find an eviction plan to free 9663676416 bytes space at location MEM
2019-08-28 07:54:16,044 WARN  SpaceReserver - SpaceReserver failed to free 9663676416 bytes on tier MEM for high watermarks: Failed to find an eviction plan to free 9663676416 bytes space at location MEM
```

We can see the MEM layer are used up, but SSD is not used.

```
# /opt/alluxio/bin/alluxio fsadmin report capacity
Capacity information for all workers:
    Total Capacity: 320.00GB
        Tier: MEM  Size: 120.00GB
        Tier: SSD  Size: 200.00GB
    Used Capacity: 118.05GB
        Tier: MEM  Size: 118.05GB
        Tier: SSD  Size: 0B
    Used Percentage: 36%
    Free Percentage: 64%

Worker Name      Last Heartbeat   Storage       Total            MEM           SSD
192.168.0.118    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          28.29GB (35%)    28.29GB       0B
192.168.0.195    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          29.92GB (37%)    29.92GB       0B
192.168.0.117    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          29.92GB (37%)    29.92GB       0B
192.168.0.194    0                capacity      80.00GB          30.00GB       50.00GB
                                  used          29.92GB (37%)    29.92GB       0B
```

5. In the job worker, here is the logging.

```
2019-08-28 03:25:02,344 INFO  TaskExecutorManager - Task 0 for job 1566803410444 started
2019-08-28 03:25:21,205 WARN  GrpcBlockingStream - Received error io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Failed to request 1048576 bytes after 10,000ms to create blockId 10452205568 for stream (GrpcDataWriter{request=type: ALLUXIO_BLOCK
id: 10452205568
tier: 0
medium_type: ""
pin_on_create: true
, address=WorkerNetAddress{host=192.168.0.194, rpcPort=29999, dataPort=29999, webPort=30000, domainSocketPath=/opt/domain/47f4f2cf-18e9-4fb0-823f-afaa8d844f11, tieredIdentity=TieredIdentity(node=192.168.0.194, rack=null)}})
2019-08-28 03:25:21,298 INFO  TaskExecutorManager - Task 0 for job 1566803410444 failed: Failed to request 1048576 bytes after 10,000ms to create blockId 10452205568 (GrpcDataWriter{request=type: ALLUXIO_BLOCK
id: 10452205568
tier: 0
medium_type: ""
pin_on_create: true
, address=WorkerNetAddress{host=192.168.0.194, rpcPort=29999, dataPort=29999, webPort=30000, domainSocketPath=/opt/domain/47f4f2cf-18e9-4fb0-823f-afaa8d844f11, tieredIdentity=TieredIdentity(node=192.168.0.194, rack=null)}})
2019-08-28 03:25:21,298 WARN  TaskExecutor - Exception running task for job Load([LoadTask{blockId=10452205568}]) : Failed to request 1048576 bytes after 10,000ms to create blockId 10452205568 (GrpcDataWriter{request=type: ALLUXIO_BLOCK
id: 10452205568
tier: 0
medium_type: ""
pin_on_create: true
```

6. But after I upin the data directory, it seems working fine. And I'm able to see the SSD layer can be used when MEM layer is used up.

The expection is that when the Memory is used up, it should turn to SSD no mater the data directory is pinned.

### LINKED ISSUE #10150: Add ability to evict blocks of open files to lower storage tiers
Currently if you load a single file that is larger than your T0 cluster capacity (for example: 70GB, 10GB per worker, 7 workers), then the write will never end (eventually timing out) and the workers will loop with the same error in #9796.

```sh
SpaceReserver failed to free $(workerMEMCapacity*lowWaterMarkRatio) bytes on tier MEM for high watermarks: Failed to find an eviction plan to free $(workerMEMCapacity*lowWaterMarkRatio) bytes space at location MEM
```

This also happens if your blocks aren't distributed evenly across the cluster and one of your workers reach 100% MEM capacity.

A workaround is to split your large files into smaller parts and work with them individually (open `bigfile-part1`, do what you need to do, close `bigfile-part1`, open `bigfile-part2`, etc)

Have the ability to evict blocks that are a part of an open file from T0 to T1.
This would lead to inconsistent R/W speeds when specific blocks are being R/W and a lot of swapping between T0 and T1, but at least your entire Alluxio stack won't fall over.

Splitting `bigfile` into `bigfile-part1`, `bigfile-part2`, `bigfile-partN` is a possible workaround.

**Urgency**
Not very urgent. Would be a nice feature (or config flag that can be set)

### LINKED ISSUE #10318: Failed to distributed load data when the size of alluxio is smaller than size of data
**Alluxio Version:**
2.1.0-SNAPSHOT-ec12adc44298d5887697b517177d1b86963e7662

**Describe the bug**

1. Failed to distribute load the data when the size of alluxio cache is smaller than underlay filesystem(OSS in this case). The size of alluxio is 120GB, the size of data in OSS is 140GB.

```
#/opt/alluxio/bin/alluxio fs distributedLoad --thread 54 /images
/images/train-00852-of-01024 is loaded
/images/train-00991-of-01024 loading
/images/train-00922-of-01024 is loaded
/images/train-00072-of-01024 loading
/images/train-00622-of-01024 is loaded
/images/train-00556-of-01024 loading
java.lang.RuntimeException: Failed to successfully complete the job.'
```

2. Expected about 88GB(each worker should use 22GB) should be used, but only 51.06GB is used.

```
# alluxio fsadmin report capacity
Capacity information for all workers:
    Total Capacity: 120.00GB
        Tier: MEM  Size: 120.00GB
    Used Capacity: 51.06GB
        Tier: MEM  Size: 51.06GB
    Used Percentage: 42%
    Free Percentage: 58%

Worker Name      Last Heartbeat   Storage       MEM
192.168.0.117    0                capacity      30.00GB
                                  used          13.23GB (44%)
192.168.0.118    0                capacity      30.00GB
                                  used          12.59GB (41%)
192.168.0.119    0                capacity      30.00GB
                                  used          13.26GB (44%)
192.168.0.120    0                capacity      30.00GB
                                  used          11.98GB (39%)
```

**To Reproduce**

1. Make MEM and SDD into a single tired store, and there are 4 workers for it. There are 120 GB in Alluxio

```
# alluxio getConf |grep level0
alluxio.master.tieredstore.global.level0.alias=MEM
alluxio.worker.tieredstore.level0.alias=MEM
alluxio.worker.tieredstore.level0.dirs.mediumtype=MEM,SSD
alluxio.worker.tieredstore.level0.dirs.path=/dev/shm,/var/lib/docker/alluxio-ssd
alluxio.worker.tieredstore.level0.dirs.quota=10GB,20GB
alluxio.worker.tieredstore.level0.watermark.high.ratio=0.95
alluxio.worker.tieredstore.level0.watermark.low.ratio=0.7
alluxio.user.file.passive.cache.enabled: false
alluxio.user.ufs.block.read.location.policy: alluxio.client.block.policy.LocalFirstAvoidEvictionPolicy
alluxio.user.block.avoid.eviction.policy.reserved.size.bytes: 8GB
alluxio.worker.allocator.class: alluxio.worker.block.allocator.GreedyAllocator
```

2. Run distributed load， but it's failed.

```
#/opt/alluxio/bin/alluxio fs distributedLoad --thread 54 /images
java.lang.RuntimeException: Failed to successfully complete the job.'
```

3. All the logs are here
[diagnose_alluxio_1572233368.tar.gz](https://github.com/Alluxio/alluxio/files/3777079/diagnose_alluxio_1572233368.tar.gz)

**Expected behavior**

**Urgency**

**Additional context**

### LINKED ISSUE #10688: Performance Overhead using read type as CACHE_PROMOTE v.s NO_CACHE
**Alluxio Version:**
2.1.1

**Describe the bug**

Deploy alluxio in kubernetes with CACHE_PROMOTE and NO_CACHE, and I found the NO_CAHCE performance is much better than CACHE_PROMOTE. The file is about 130MB, and 1154 files in all. 

**To Reproduce**
1. Deploy Alluxio with helm chart, the values file uses NO_CACHE

```
image: alluxio/alluxio
imageTag: "2.1.1"
imagePullPolicy: Always

properties:
    alluxio.user.metrics.collection.enabled: false
    alluxio.user.file.passive.cache.enabled: false
    alluxio.user.ufs.block.read.location.policy: alluxio.client.block.policy.LocalFirstAvoidEvictionPolicy
    # It should be great than (120)*0.01 = 2GB
    alluxio.user.block.avoid.eviction.policy.reserved.size.bytes: 2GB
    alluxio.fuse.debug.enabled: "false"
    alluxio.user.file.writetype.default: MUST_CACHE
    alluxio.user.block.write.location.policy.class: alluxio.client.block.policy.LocalFirstAvoidEvictionPolicy
    alluxio.worker.allocator.class: alluxio.worker.block.allocator.GreedyAllocator
    alluxio.user.block.size.bytes.default: 32MB
    fs.oss.endpoint: oss-cn-shanghai-internal.aliyuncs.com
    # set OSS endpoint
    fs.oss.accessKeyId: xxx
    fs.oss.accessKeySecret: yyy
    alluxio.user.file.readtype.default: NO_CACHE
    alluxio.master.mount.table.root.ufs: oss://imagenet/
    #alluxio.user.file.master.client.threads: "128"
    #alluxio.user.block.master.client.threads: "128"
    # alluxio.underfs.oss.socket.timeout: 3000sec
    alluxio.user.network.reader.chunk.size.bytes: 32MB
    alluxio.worker.file.buffer.size: 320MB
    alluxio.security.stale.channel.purge.interval: 365d
    alluxio.user.metadata.cache.enabled: "false"
    alluxio.user.metadata.cache.max.size: "1000000"
    alluxio.user.direct.memory.io.enabled: "true"
    alluxio.fuse.cached.paths.max: "1000000"
    alluxio.job.worker.threadpool.size: "64"
    alluxio.worker.block.master.client.pool.size: 128


tieredstore:
  levels:
  - alias: MEM
    level: 0
    path: /dev/shm
    type: hostPath
    # quota: 120GB
    high: 0.99
    low: 0.7

fuse:
  image: alluxio/alluxio-fuse
  imageTag: "2.1.1"
  imagePullPolicy: Always
  args:
    - fuse
    - --fuse-opts=direct_io
```

2. Run the training jobs without any cache
3. Performance result is :

```
10000	images/sec: 897.3 +/- 2.1 (jitter = 37.6)	4.466
----------------------------------------------------------------
total images/sec: 7171.94
----------------------------------------------------------------
```

4. Deploy the alluxio without setting alluxio.user.file.readtype.default

```
image: alluxio/alluxio
imageTag: "2.1.1"
imagePullPolicy: Always

properties:
    alluxio.user.metrics.collection.enabled: false
    alluxio.user.file.passive.cache.enabled: false
    alluxio.user.ufs.block.read.location.policy: alluxio.client.block.policy.LocalFirstAvoidEvictionPolicy
    # It should be great than (120)*0.01 = 2GB
    alluxio.user.block.avoid.eviction.policy.reserved.size.bytes: 2GB
    alluxio.fuse.debug.enabled: "false"
    alluxio.user.file.writetype.default: MUST_CACHE
    alluxio.user.block.write.location.policy.class: alluxio.client.block.policy.LocalFirstAvoidEvictionPolicy
    alluxio.worker.allocator.class: alluxio.worker.block.allocator.GreedyAllocator
    alluxio.user.block.size.bytes.default: 32MB
    fs.oss.endpoint: oss-cn-shanghai-internal.aliyuncs.com
    # set OSS endpoint
    fs.oss.accessKeyId: xxx
    fs.oss.accessKeySecret: yyy
    alluxio.master.mount.table.root.ufs: oss://imagenet/
    #alluxio.user.file.master.client.threads: "128"
    #alluxio.user.block.master.client.threads: "128"
    # alluxio.underfs.oss.socket.timeout: 3000sec
    alluxio.user.network.reader.chunk.size.bytes: 32MB
    alluxio.worker.file.buffer.size: 320MB
    alluxio.security.stale.channel.purge.interval: 365d
    alluxio.user.metadata.cache.enabled: "false"
    alluxio.user.metadata.cache.max.size: "1000000"
    alluxio.user.direct.memory.io.enabled: "true"
    alluxio.fuse.cached.paths.max: "1000000"
    alluxio.job.worker.threadpool.size: "64"
    alluxio.worker.block.master.client.pool.size: 128


tieredstore:
  levels:
  - alias: MEM
    level: 0
    path: /dev/shm
    type: hostPath
    # quota: 120GB
    high: 0.99
    low: 0.7

fuse:
  image: alluxio/alluxio-fuse
  imageTag: "2.1.1"
  imagePullPolicy: Always
  args:
    - fuse
    - --fuse-opts=direct_io
```

5. Load all the data into allxuio
6. Run the training job
7. Get the performance result:

```
----------------------------------------------------------------
1000	images/sec: 620.6 +/- 7.8 (jitter = 264.4)	4.721
----------------------------------------------------------------
total images/sec: 4963.33
----------------------------------------------------------------
```

**Expected behavior**

**Urgency**

**Additional context**

ggezer:
Changes in this PR:
- Allow writes to utilize any available space on the cache, starting from the highest tier.
- When the cache is full, remove the worst block in order to open up space for the write.
- Move blocks between tiers based on the access pattern.

Fixes #9796, Fixes #9804, Fixes #10150, Fixes #10318, Fixes #10688


core/server/worker/src/test/java/alluxio/worker/block/management/TierSwapTaskTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Test
+  public void testOverlapElimination() throws Exception {
+    Random rnd = new Random();
+    StorageDir[] dirArray = new StorageDir[] {mTestDir1, mTestDir2, mTestDir3, mTestDir4};
+
+    // Start simulating random load on worker.
+    startSimulateLoad();
+
+    // Fill each directory, leaving free space on each as swap buffer.
```
ggezer: Fix this comment as swap space is internally managed.

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
```diff
@@ -0,0 +1,182 @@
...
+        final BlockManagementTask nextTask = getNextTask();
+        if (nextTask == null) {
+          LOG.info("No management task pending.");
+          Thread.sleep(mIdleSleepMs);
+          continue;
+        }
+
+        // Submit and wait for the task.
+        currentTask = nextTask;
+        mTaskExecutor.submit(() -> nextTask.run()).get();
```
calvinjia: Add TODO here to run multiple tasks in parallel?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -126,21 +121,23 @@ public TieredBlockStore() {
...
-    mEvictor = Evictor.Factory.create(initManagerView, mAllocator);
-    if (mEvictor instanceof BlockStoreEventListener) {
-      registerBlockStoreEventListener((BlockStoreEventListener) mEvictor);
-    }
-
-    mStorageTierAssoc = new WorkerStorageTierAssoc();
+    // Initialize and start coordinator.
+    mTaskCoordinator = new ManagementTaskCoordinator(this, mMetaManager,
+        new DefaultStoreLoadTracker(), () -> getUpdatedView());
+    mTaskCoordinator.start();
```
calvinjia: Is it okay to have this thread running by itself, should we keep track of it?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -526,42 +697,92 @@ private void abortBlockInternal(long sessionId, long blockId) throws BlockDoesNo
...
 
-      try (LockResource r = new LockResource(mMetadataWriteLock)) {
-        mMetaManager.commitTempBlockMeta(tempBlockMeta);
-      } catch (BlockAlreadyExistsException | BlockDoesNotExistException
-          | WorkerOutOfSpaceException e) {
-        throw Throwables.propagate(e); // we shall never reach here
-      }
+          if (dirView == null) {
+            LOG.error("Target tier: {} has no evictable space to store {} bytes for session: {}",
+                options.getLocation(), options.getSize(), sessionId);
```
calvinjia: Should we log this, I don't think its unexpected? Maybe we can increment some metric to keep track of how many failed allocations we had.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -526,42 +697,92 @@ private void abortBlockInternal(long sessionId, long blockId) throws BlockDoesNo
...
+        }
 
-      return loc;
-    } finally {
-      mLockManager.unlockBlock(lockId);
+        if (options.isEvictionAllowed()) {
+          // There is no space left on worker.
+          // Free more than requested by configured free-ahead size.
+          long freeAheadBytes =
+              ServerConfiguration.getBytes(PropertyKey.WORKER_TIERED_STORE_FREE_AHEAD_BYTES);
```
calvinjia: Is this concept necessary, it seems a bit too much of a heuristic?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -526,42 +697,92 @@ private void abortBlockInternal(long sessionId, long blockId) throws BlockDoesNo
...
+          freeSpace(sessionId, options.getSize(), options.getSize() + freeAheadBytes,
+              BlockStoreLocation.anyTier());
+
+          dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
+              BlockStoreLocation.anyTier(), allocatorView.refreshView());
```
calvinjia: Should we do a check if dirView is null here and do our failure handling (ie. log message now but I suggest to increment some metric)?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -526,42 +697,92 @@ private void abortBlockInternal(long sessionId, long blockId) throws BlockDoesNo
...
+        }
+      }
+    } catch (Exception e) {
+      LOG.error("Allocation failure. Options: {}. Error: {}", options, e);
```
calvinjia: return null here?

core/client/fs/src/main/java/alluxio/client/file/options/OutStreamOptions.java
```diff
@@ -144,6 +144,9 @@ private OutStreamOptions(ClientContext context, AlluxioConfiguration alluxioConf
     mReplicationMax = alluxioConf.getInt(PropertyKey.USER_FILE_REPLICATION_MAX);
     mReplicationMin = alluxioConf.getInt(PropertyKey.USER_FILE_REPLICATION_MIN);
     mMediumType = "";
+    if (alluxioConf.isSet(PropertyKey.USER_FILE_TARGET_MEDIA)) {
+      mMediumType = alluxioConf.get(PropertyKey.USER_FILE_TARGET_MEDIA);
+    }
```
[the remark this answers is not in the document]
ggezer: https://github.com/Alluxio/alluxio/issues/11154

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2272,9 +2272,12 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  /**
+   * @deprecated use block annotators instead
+   */
+  @Deprecated(message = "Use WORKER_BLOCK_ANNOTATOR_CLASS instead.")
```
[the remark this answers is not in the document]
ggezer: It's not exactly the same functionality. And keeping them separate simplifies emulation decision.

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
ggezer: At this level I don't intend to run things in parallel as each task here is I/O bound. I believe it's good to have priority based preemption here. Tasks themselves should spawn concurrent tasks though.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: This concept is a fail-safe against a corner case:
- Worker reached peak capacity
- Annotator is LRFU
- Concurrent readers/writers

In this case, each allocation could end up sorting the whole block list in order to find eviction candidate. This is due to LRFU being an online scheme. If CPU starts burning, we could evict more than necessary using this prop and avoid excessive sorting per allocation.

This should be mentioned in the tuning guides.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: Currently, as per previous trend, callers all throw/log for null response from this method. I think metrics would be fine. However I was planning an all over work on defining new metrics. It can be done at the end of this PR or as a separate PR.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: Closing this now as part of block-worker shutdown.

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
calvinjia: If we don't plan to run tasks in parallel, do we still need to submit to another threadpool?

ggezer: It's a single-thread-executor. 
It can be useful to keep a separate thread pool for the future in case we want to let some tasks run in parallel if not all.

core/server/worker/src/main/java/alluxio/worker/block/management/TierManagementTaskProvider.java
```diff
@@ -0,0 +1,133 @@
...
+            mLoadTracker);
+      default:
+        throw new IllegalArgumentException("Unknown task type.");
+    }
+  }
+
+  /**
+   * @return the next tier management task that is required
+   */
+  private TierManagementTaskType findNextTask() {
```
calvinjia: After this method, we only hint at what kind of task is necessary but not the details of the task. What is the computation cost here and could we do more computation to pass onto the execution of the task (ie. which blocks to swap)? Also since the swap will happen at a different time, are there any consistency issues?

core/server/worker/src/main/java/alluxio/worker/block/annotator/LRUAnnotator.java
```diff
@@ -0,0 +1,97 @@
...
+   * Creates a new LRU annotator.
+   */
+  public LRUAnnotator() {
+    mLRUClock = new AtomicLong(0);
+  }
+
+  @Override
+  public BlockSortedField updateSortedField(long blockId, LRUSortedField oldValue) {
+    long clockValue = 0;
+    if (oldValue != null) {
```
calvinjia: When will there be no `oldValue`, does it make sense to set the clock value to 0 (most recent)?

ggezer: There will be no `oldValue` when the block is seen the first time. Recency is by having greater values, so 0 means the least recent. 

Assumption here is that a block will first be committed, and have its value annotated as 0. Value will be bumped only when it's accessed after commit.

core/server/worker/src/main/java/alluxio/worker/block/management/TierManagementTaskProvider.java
ggezer: Consistency is loose based on `ConcurrentSkipListSet`'s iterator consistency. So decision given by this method, or the later swap-list that is acquired could no longer be valid. In that case, next run is supposed to fix that. However, with load tracking, inconsistency will not be vast and will be sorted out on next run.

As for the computations, this method's heaviest lift is the `BlockIterator:overlaps` method which is not directly translatable. I wouldn't worry about the wasted work here as overlap will be detected at first block comparison at a tier intersection.

core/server/worker/src/main/java/alluxio/worker/block/management/DefaultStoreLoadTracker.java
```diff
@@ -0,0 +1,138 @@
...
+ * Default {@link StoreLoadTracker} that reports load based on open reader/writer streams to the
+ * local block store.
+ *
+ * TODO(ggezer): TV2 - Add a safety net against close calls not being called.
+ */
+public class DefaultStoreLoadTracker implements StoreLoadTracker, BlockStreamListener {
+  /** Used to keep reference to stream readers/writers per location. */
+  private final ConcurrentHashMap<BlockStoreLocation, Set<Object>> mStreamsPerLocation;
+  /** Used for delayed removing of streams in order to emulate activity cool-down. */
+  private final ScheduledExecutorService mScheduler;
```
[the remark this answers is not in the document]
ggezer: ManagementTaskCoordinator also have a cool down. However, it's required to implement the cool-down outside of coordinator as each task also need to track load by itself.

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
ggezer: Actually, after working on concurrency, I think it's simpler to run tasks on coordinator thread. We can change If we want to allow multiple tier management tasks later.

core/server/worker/src/main/java/alluxio/worker/block/annotator/LRUAnnotator.java
calvinjia: Is this typical for LRU algorithms, or should the block have highest recency on commit?

core/server/worker/src/main/java/alluxio/worker/block/management/DefaultStoreLoadTracker.java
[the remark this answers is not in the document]
calvinjia: Could you elaborate on this a bit more? Does the `LoadTracker` need to understand the load per task (ie. so we can do storage directory level throttling) or just whether or not there has been load?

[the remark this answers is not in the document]
calvinjia: Could you elaborate on this a bit more? Does the LoadTracker need to understand the load per task (ie. so we can do storage directory level throttling) or just whether or not there has been load?

core/server/worker/src/main/java/alluxio/worker/block/annotator/LRUAnnotator.java
ggezer: This was the default in LRUEvictor.
I believe this could potentially be a configuration property.

core/server/worker/src/main/java/alluxio/worker/block/management/DefaultStoreLoadTracker.java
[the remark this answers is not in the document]
ggezer: `LoadTracker` needs to figure out if there was an external (user I/O) on a particular directory.
So it will be used by tasks to figure out if the directory that they're working on is under load, then back-off.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -106,8 +110,10 @@ protected void cancelRequest(BlockWriteRequestContext context) throws Exception
 
   @Override
   protected void cleanupRequest(BlockWriteRequestContext context) throws Exception {
-    WriteRequest request = context.getRequest();
-    mWorker.cleanupSession(request.getSessionId());
+    if (context.getBlockWriter() != null) {
+      context.getBlockWriter().close();
```
gpang: is this an unrelated bug fix?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -209,29 +206,23 @@ public BlockReader getBlockReader(long sessionId, long blockId, long lockId)
...
   }
 
   @Override
-  public TempBlockMeta createBlock(long sessionId, long blockId, BlockStoreLocation location,
-      long initialBlockSize) throws BlockAlreadyExistsException, WorkerOutOfSpaceException,
-      IOException {
-    LOG.debug("createBlock: sessionId={}, blockId={}, location={}, initialBlockSize={}", sessionId,
-        blockId, location, initialBlockSize);
-    RetryPolicy retryPolicy = new TimeoutRetry(FREE_SPACE_TIMEOUT_MS, EVICTION_INTERVAL_MS);
-    while (retryPolicy.attempt()) {
```
gpang: retry is no longer necessary?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -266,14 +257,40 @@ public void commitBlock(long sessionId, long blockId, boolean pinOnCreate)
       throws BlockAlreadyExistsException, InvalidWorkerStateException, BlockDoesNotExistException,
       IOException {
     LOG.debug("commitBlock: sessionId={}, blockId={}", sessionId, blockId);
```
gpang: should we add `pinOnCreate`?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -266,14 +257,40 @@ public void commitBlock(long sessionId, long blockId, boolean pinOnCreate)
...
+    } finally {
+      mLockManager.unlockBlock(lockId);
     }
   }
 
+  @Override
+  public long commitBlockLocked(long sessionId, long blockId, boolean pinOnCreate)
+      throws BlockAlreadyExistsException, InvalidWorkerStateException, BlockDoesNotExistException,
+      IOException {
+    LOG.debug("commitBlock: sessionId={}, blockId={}", sessionId, blockId);
```
gpang: `commitBlockLocked`? Should we add `pinOnCreate`?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -266,14 +257,40 @@ public void commitBlock(long sessionId, long blockId, boolean pinOnCreate)
       throws BlockAlreadyExistsException, InvalidWorkerStateException, BlockDoesNotExistException,
       IOException {
     LOG.debug("commitBlock: sessionId={}, blockId={}", sessionId, blockId);
-    BlockStoreLocation loc = commitBlockInternal(sessionId, blockId, pinOnCreate);
-    synchronized (mBlockStoreEventListeners) {
-      for (BlockStoreEventListener listener : mBlockStoreEventListeners) {
-        listener.onCommitBlock(sessionId, blockId, loc);
+    long lockId = mLockManager.lockBlock(sessionId, blockId, BlockLockType.WRITE);
```
gpang: can `commitBlock` reuse `commitBlockLocked`?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -291,51 +308,180 @@ public void requestSpace(long sessionId, long blockId, long additionalBytes)
...
+      StorageDirView allocationDir = allocateSpace(sessionId,
+          AllocateOptions.forRequestSpace(additionalBytes, tempBlockMeta.getBlockLocation()));
+      if (allocationDir == null) {
+        throw new WorkerOutOfSpaceException(String.format(
+            "Can't reserve more space for block: %d under session: %d.", blockId, sessionId));
+      }
+
+      if (!allocationDir.toBlockStoreLocation().equals(tempBlockMeta.getBlockLocation())) {
+        // If reached here, allocateSpace() failed to enforce 'forceLocation' flag.
+        throw new RuntimeException("Allocation error: location enforcement failed");
```
gpang: Is `RuntimeException` the correct exception to throw? Can we provide more information in the message?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -291,51 +308,180 @@ public void requestSpace(long sessionId, long blockId, long additionalBytes)
...
+
+  @Override
+  public void swapBlocks(long sessionId, long blockId1, long blockId2)
+      throws BlockDoesNotExistException, BlockAlreadyExistsException, InvalidWorkerStateException,
+      WorkerOutOfSpaceException, IOException {
+    BlockMeta blockMeta1;
+    BlockMeta blockMeta2;
+    boolean doFsSwap = false;
+
+    long lockId1 = mLockManager.lockBlock(sessionId, blockId1, BlockLockType.READ);
```
gpang: should there be a log debug message, like the other methods?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -291,51 +308,180 @@ public void requestSpace(long sessionId, long blockId, long additionalBytes)
...
+        Path dirPath2 = Paths.get(blockMeta2.getParentDir().getDirPath());
+        if (Objects.equals(dirPath1.getRoot(), dirPath2.getRoot())) {
+          doFsSwap = true;
+        }
+      }
+
+      // Allocate swap space from directories for swapping blocks.
+      if (!blockMeta1.getParentDir().allocateFromReserve(blockMeta2.getBlockSize())
+          || !blockMeta2.getParentDir().allocateFromReserve(blockMeta1.getBlockSize())) {
+        throw new WorkerOutOfSpaceException("No reserved space left for swapping.");
```
gpang: can we split this up and put the directory that ran out of space in the message?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -348,37 +494,61 @@ public void removeBlock(long sessionId, long blockId)
...
+          listener.onAccessBlock(sessionId, blockId);
+          listener.onAccessBlock(sessionId, blockId, blockMeta.getBlockLocation());
+        }
       }
     }
   }
 
   @Override
-  public void freeSpace(long sessionId, long availableBytes, BlockStoreLocation location)
+  public synchronized void freeSpace(long sessionId, long minAvailableBytes, long maxAvailableBytes,
```
gpang: why is this now `synchronized`?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -570,162 +791,107 @@ private BlockStoreLocation commitBlockInternal(long sessionId, long blockId, boo
...
+              listener.onRemoveBlockByClient(sessionId, blockMeta.getBlockId());
+              listener.onRemoveBlock(sessionId, blockMeta.getBlockId(),
+                  blockMeta.getBlockLocation());
             }
           }
+          minBytesToFree -= blockMeta.getBlockSize();
+          maxBytesToFree -= blockMeta.getBlockSize();
+          spaceFreed += blockMeta.getBlockSize();
+        } catch (BlockDoesNotExistException e) {
+          LOG.info("Failed to evict blockId {}, it could be already deleted", blockToDelete);
```
gpang: Should this be warn?

Also, it is always the case that the block doesn't exist right? The message sounds like the eviction failed, and one possible reason is that the block is already gone.

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDir.java
```diff
@@ -57,16 +57,23 @@
...
-  private StorageDir(StorageTier tier, int dirIndex, long capacityBytes, String dirPath,
-      String dirMedium) {
+  private StorageDir(StorageTier tier, int dirIndex, long capacityBytes, long reservedBytes,
+      String dirPath, String dirMedium) {
     mTier = Preconditions.checkNotNull(tier, "tier");
     mDirIndex = dirIndex;
     mCapacityBytes = capacityBytes;
-    mAvailableBytes = new AtomicLong(capacityBytes);
+    if (capacityBytes < reservedBytes) {
+      reservedBytes = capacityBytes / 20;
```
gpang: Do we need to make sure it is not too small either?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2284,21 +2287,126 @@ public String toString() {
...
+          .build();
+  public static final PropertyKey WORKER_MANAGEMENT_TASK_THREAD_COUNT =
+      new Builder(Name.WORKER_MANAGEMENT_TASK_THREAD_COUNT)
+          .setDefaultSupplier(() -> Runtime.getRuntime().availableProcessors(),
+              "Use {CPU core count} threads for all management tasks")
+          .setDescription("The number of threads for management task executor")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_MANAGEMENT_TIER_TASK_CONCURRENCY =
```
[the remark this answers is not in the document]
ggezer: This controls concurrency at the disk level. I try to limit the disks we're spinning using this.
Needs a better name for sure.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2284,21 +2287,126 @@ public String toString() {
...
+          .setDefaultValue("1GB")
+          .setDescription("The amount of space that is reserved from each storage directory "
+              + "for internal management tasks.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_MANAGEMENT_IDLE_SLEEP_TIME =
+      new Builder(Name.WORKER_MANAGEMENT_IDLE_SLEEP_TIME)
+          .setDefaultValue("10sec")
+          .setDescription("Management task coordinator will back-off for specified duration when"
```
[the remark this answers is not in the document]
ggezer: This is when there is no management required VS back-ing off to come back again.
Here because checking if management is required has its own cost. However I'm inclined to just reuse the cool-down for this purpose.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: No. There is nothing to wait on.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: Api is no more

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2284,21 +2287,126 @@ public String toString() {
...
+          .setDefaultValue(0.1)
+          .setDescription("Ratio of free space per-tier for moving blocks from below."
+              + " When under this value moving to that tier will be stopped.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_MANAGEMENT_SWAP_USING_FS =
+      new Builder(Name.WORKER_MANAGEMENT_SWAP_USING_FS)
+          .setDefaultValue(true)
+          .setDescription("Whether to use file-system moves to "
```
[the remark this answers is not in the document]
ggezer: This API is no more. Will remove the key.

core/server/worker/src/main/java/alluxio/worker/block/evictor/LRFUEvictor.java
```diff
@@ -70,9 +73,10 @@
    */
   public LRFUEvictor(BlockMetadataEvictorView view, Allocator allocator) {
     super(view, allocator);
-    mStepFactor = ServerConfiguration.getDouble(PropertyKey.WORKER_EVICTOR_LRFU_STEP_FACTOR);
+    mStepFactor =
+        ServerConfiguration.getDouble(PropertyKey.WORKER_BLOCK_ANNOTATOR_LRFU_STEP_FACTOR);
     mAttenuationFactor =
-        ServerConfiguration.getDouble(PropertyKey.WORKER_EVICTOR_LRFU_ATTENUATION_FACTOR);
+        ServerConfiguration.getDouble(PropertyKey.WORKER_BLOCK_ANNOTATOR_LRFU_ATTENUATION_FACTOR);
```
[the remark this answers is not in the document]
ggezer: @calvinjia also brought this point up.
My concern is to separate emulation mode. And it's not exactly an evictor anymore.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractWriteHandler.java
```diff
@@ -106,7 +106,14 @@ public void write(WriteRequest writeRequest) {
       try {
         if (mContext == null) {
           LOG.debug("Received write request {}.", writeRequest);
-          mContext = createRequestContext(writeRequest);
+          try {
```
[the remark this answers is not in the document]
ggezer: It's related to issues that are linked.

core/server/worker/src/main/java/alluxio/worker/block/BlockMetadataManager.java
```diff
@@ -129,6 +159,30 @@ public void commitTempBlockMeta(TempBlockMeta tempBlockMeta)
     dir.addBlockMeta(block);
   }
 
+  /**
+   * Swaps location of two blocks in metadata.
+   *
+   * @param blockMeta1 the first block meta
+   * @param blockMeta2 the second block meta
+   * @throws BlockDoesNotExistException
```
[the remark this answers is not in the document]
ggezer: this API is no more.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
ggezer: Yes. Again related to a linked issue.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: This API is no more.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: It now only removes blocks and shouldn't run concurrently as per allocation guarantees.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2284,21 +2287,126 @@ public String toString() {
...
           .setDefaultValue(0.25)
           .setDescription("A factor in [0, 1] to control the behavior of LRFU: smaller value "
               + "makes LRFU more similar to LFU; and larger value makes LRFU closer to LRU.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_MANAGEMENT_GLOBAL_LOAD_DETECTION_ENABLED =
+      new Builder(Name.WORKER_MANAGEMENT_GLOBAL_LOAD_DETECTION_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("Whether to postpone all management tasks under any user activity")
```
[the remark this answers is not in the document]
ggezer: added.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2284,21 +2287,126 @@ public String toString() {
...
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_MANAGEMENT_LOAD_DETECTION_COOL_DOWN_TIME =
+      new Builder(Name.WORKER_MANAGEMENT_LOAD_DETECTION_COOL_DOWN_TIME)
+          .setDefaultValue("10sec")
+          .setDescription("Management tasks will not run for this long after load detected.")
```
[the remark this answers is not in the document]
ggezer: added.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3004,6 +3121,10 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_FILE_TARGET_MEDIA =
+      new Builder(Name.USER_FILE_TARGET_MEDIA)
+          .setScope(Scope.CLIENT)
```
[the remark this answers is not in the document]
ggezer: no default.
added description.

core/common/src/main/java/alluxio/worker/block/io/BlockStreamTracker.java
```diff
@@ -0,0 +1,91 @@
...
+import alluxio.worker.block.BlockStoreLocation;
+
+import java.util.Set;
+
+/**
+ * Used to emit block reader/writer open/close events.
+ */
+public class BlockStreamTracker {
+  /** List of listeners for this tracker. */
+  private static Set<BlockStreamListener> sListeners = new ConcurrentHashSet<>();
```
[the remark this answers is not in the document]
ggezer: I didn't want to introduce another variable through already complex rpc layer.

core/server/common/src/main/java/alluxio/StorageTierAssoc.java
```diff
@@ -119,4 +121,17 @@ public int size() {
     }
     return ret;
   }
+
+  /**
+   * @return list of intersections between tier levels
+   */
+  public List<Pair<BlockStoreLocation, BlockStoreLocation>> intersectionList() {
```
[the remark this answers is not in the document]
ggezer: added to `StorageTierAssocTest`

core/server/worker/src/main/java/alluxio/worker/block/BlockMetadataEvictorView.java
```diff
@@ -111,7 +113,11 @@ public boolean isBlockLocked(long blockId) {
    * @return boolean, true if the block can be evicted
    */
   public boolean isBlockEvictable(long blockId) {
-    return (!isBlockPinned(blockId) && !isBlockLocked(blockId) && !isBlockMarked(blockId));
+    boolean pinned = isBlockPinned(blockId);
+    boolean locked = isBlockLocked(blockId);
+    boolean marked = isBlockMarked(blockId);
+    LOG.debug("Block: {}. Pinned: {}, Locked: {}, Marked: {}", blockId, pinned, locked, marked);
```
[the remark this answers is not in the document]
ggezer: That was beneficial to me for some investigations.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: No. The purpose of this API is to retain the lock after returning from this. They both use `commitBlockInternal` though

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: Right. It's more like `IllegalStateException`. Fixed.

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageDir.java
ggezer: I feel like that would require another prop to get a test pass.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: Yes.
It's printed out when we want to remove a block, based on annotations, and that block was not there. It can happen if block annotator is out of sync with the store.

ggezer:
jenkins, test this please

core/server/worker/src/main/java/alluxio/worker/block/evictor/BlockTransferInfo.java
```diff
@@ -64,7 +102,7 @@ public BlockStoreLocation getDstLocation() {
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
-        .add("blockId", mBlockId)
+        .add("blockId", mSrcBlockId)
```
gpang, line 105: should we have `srcBlockId` and `dstBlockId`?

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
```diff
@@ -0,0 +1,200 @@
...
+        if (mGlobalLoadDetectionEnabled
+            && mLoadTracker.loadDetected(BlockStoreLocation.anyTier())) {
+          LOG.debug("Load detected under global load detection mode.");
+          Thread.sleep(mLoadDetectionCoolDownMs);
+          continue;
+        }
+
+        final BlockManagementTask nextTask = getNextTask();
+        if (nextTask == null) {
+          LOG.debug("No management task pending.");
```
gpang, line 175: should we also mention "sleeping X ms"?

core/common/src/main/java/alluxio/conf/PropertyKey.java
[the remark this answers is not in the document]
ggezer: Cut this one.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -570,162 +689,129 @@ private BlockStoreLocation commitBlockInternal(long sessionId, long blockId, boo
...
-        throw new WorkerOutOfSpaceException(
-            ExceptionMessage.NO_EVICTION_PLAN_TO_FREE_SPACE, availableBytes, location.tierAlias());
+  private void freeSpaceInternal(long sessionId, long minContigiousBytes, long minAvailableBytes,
+      BlockStoreLocation location) throws WorkerOutOfSpaceException, IOException {
+    // TODO(ggezer): Too much memory pressure when pinned-inodes list is large.
+    BlockMetadataEvictorView evictorView = getUpdatedView();
+    LOG.debug(
+        "freeSpaceInternal - locAvailableBytes: {}, minContigiousBytes: {}, minAvailableBytes: {}",
+        evictorView.getAvailableBytes(location), minContigiousBytes, minAvailableBytes);
+    boolean contigiousSpaceFound = false;
```
calvinjia, line 747: `contiguousSpaceFound`

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -570,162 +689,129 @@ private BlockStoreLocation commitBlockInternal(long sessionId, long blockId, boo
...
+    boolean availableBytesFound = false;
```
calvinjia, line 748: Do we care about `availableBytesFound`? It seems like the only satisfactory condition is `contigiousSpaceFound`

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -570,162 +689,129 @@ private BlockStoreLocation commitBlockInternal(long sessionId, long blockId, boo
...
-    try (LockResource r = new LockResource(mMetadataWriteLock)) {
-      plan = mEvictor.freeSpaceWithView(availableBytes, location, getUpdatedView(), mode);
-      // Absent plan means failed to evict enough space.
-      if (plan == null) {
-        throw new WorkerOutOfSpaceException(
-            ExceptionMessage.NO_EVICTION_PLAN_TO_FREE_SPACE, availableBytes, location.tierAlias());
+  private void freeSpaceInternal(long sessionId, long minContigiousBytes, long minAvailableBytes,
+      BlockStoreLocation location) throws WorkerOutOfSpaceException, IOException {
+    // TODO(ggezer): Too much memory pressure when pinned-inodes list is large.
+    BlockMetadataEvictorView evictorView = getUpdatedView();
```
calvinjia, line 743: Does this need to be updated over the course of this method?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -570,162 +689,129 @@ private BlockStoreLocation commitBlockInternal(long sessionId, long blockId, boo
...
-          continue;
-        } catch (BlockAlreadyExistsException e) {
-          continue;
-        } catch (BlockDoesNotExistException e) {
-          LOG.info("Failed to demote blockId {}, it could be already deleted", blockId);
-          continue;
-        }
-        if (moveResult.getSuccess()) {
+          BlockMeta blockMeta = mMetaManager.getBlockMeta(blockToDelete);
+          removeBlockInternal(blockMeta);
```
calvinjia, line 789: We don't need any locks for this?

core/server/worker/src/main/java/alluxio/worker/block/management/BlockTransferExecutor.java
```diff
@@ -0,0 +1,260 @@
...
+      try {
+        if (mLoadTracker.loadDetected(transferInfo.getSrcLocation(),
+            transferInfo.getDstLocation())) {
+          LOG.warn("Skipping transfer-order: {} due to user activity.", transferInfo);
+          continue;
+        }
+
+        boolean useReservedSpace = transferInfo.isSwap();
+
+        synchronized (this) {
```
calvinjia, line 234: What is this lock intended to guard against? It looks like `BlockTransferExecutor` is never shared among threads?

core/server/worker/src/main/java/alluxio/worker/block/annotator/LRUAnnotator.java
calvinjia: @ggezer For the LRU evictor:
```
@Override
  public void onCommitBlock(long sessionId, long blockId, BlockStoreLocation location) {
    // Since the temp block has been committed, update Evictor about the new added blocks
    mLRUCache.put(blockId, UNUSED_MAP_VALUE);
  }
```
Doesn't this mean newly created blocks are added to the tail of the LRU cache (most recently used)?

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
```diff
@@ -0,0 +1,420 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * The default {@link BlockIterator} implementation that integrates with the
+ * {@link BlockMetadataManager}.
+ */
+public class DefaultBlockIterator implements BlockIterator {
```
calvinjia, line 41: Does this class need locking on top of the concurrent data structures? Which methods we expect to be called concurrently and what are the guarantees for `mPerDirOrderedSets`?

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
```diff
@@ -0,0 +1,420 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultBlockIterator.class);
+
+  /** Map of sorted block set collections per directory. */
+  private final HashMap<BlockStoreLocation, SortedBlockSet<BlockSortedField>> mPerDirOrderedSets;
```
calvinjia, line 45: It seems possible this will be modified concurrently during iteration?

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
```diff
@@ -0,0 +1,420 @@
...
+          blockUpdated(blockId, dirLocation);
+        }
+      }
+    }
+  }
+
+  /**
+   * Called by block-store event callbacks to update a block.
+   */
+  private void blockUpdated(long blockId, BlockStoreLocation location) {
```
calvinjia, line 102: For block updated and block removed, do we need to notify the block annotator?

core/server/worker/src/main/java/alluxio/worker/block/annotator/BlockAnnotator.java
```diff
@@ -0,0 +1,79 @@
...
+import alluxio.collections.Pair;
+import alluxio.conf.PropertyKey;
+import alluxio.conf.ServerConfiguration;
+import alluxio.util.CommonUtils;
+
+import javax.annotation.concurrent.ThreadSafe;
+import java.util.List;
+
+/**
+ * Interface for providers that annotates blocks for sorting.
```
calvinjia, line 24: Could you clarify the responsibilities of this class? What is the division of work between this class and `DefaultBlockIterator`?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -126,21 +116,23 @@ public TieredBlockStore() {
...
-        Collections.<Long>emptySet());
-    mEvictor = Evictor.Factory.create(initManagerView, mAllocator);
-    if (mEvictor instanceof BlockStoreEventListener) {
-      registerBlockStoreEventListener((BlockStoreEventListener) mEvictor);
-    }
-
-    mStorageTierAssoc = new WorkerStorageTierAssoc();
+    // Initialize and start coordinator.
+    mTaskCoordinator = new ManagementTaskCoordinator(this, mMetaManager,
+        new DefaultStoreLoadTracker(), () -> getUpdatedView());
```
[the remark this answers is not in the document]
ggezer: Yes, we still make use of them.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: No. Especially considering it's very lightweight now (only deletes) compared to before.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
ggezer: `freeSpaceInternal()` is only called by `allocateSpace()` which is required to be called with meta-write lock acquired.

We actually don't need this API in `BlockWorker` anymore.

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
ggezer: For updates, we need the annotator to give the block a new annotated value.
For removes, no need to invoke the annotator.

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
ggezer: Correct.

ggezer: I think you mean the `storageLost()` callback. Nice catch.
Updated to a concurrent-map.

ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
Failing due to unrelated flakes after I merged with master. 
Apparently master became flaky again.

ggezer:
jenkins, test this please

calvinjia:

PR #11139, opened 2020-03-06
Fix configuration sync with worker

bf8086:
We recently updated the configuration sync so all master properties will get pushed to worker unless worker specify something in config file specifically. This is okay for most cases but for property that has a server specific default value, the master default value will override worker default value, making it invalid in some cases.

For example, the default value for `alluxio.underfs.hdfs.configuration` is set to base on the configuration directory. If master and worker have different Alluxio directory for example in a containerized environment, the worker will sync the default path from master which does not exist in worker, causing worker to fail to talk to HDFS.

This change modifies the configuration sync so worker only sync the properties that are not default value on master. This way we preserve the server specific default values in the workers.

### DIFF
--- a/core/common/src/main/java/alluxio/ClientContext.java
+++ b/core/common/src/main/java/alluxio/ClientContext.java
@@ -132,7 +132,7 @@ public class ClientContext {
       return;
     }
     GetConfigurationPResponse response = ConfigurationUtils.loadConfiguration(address,
-        conf, !loadClusterConf, !loadPathConf);
+        conf, !loadClusterConf, !loadPathConf, false);
     if (loadClusterConf) {
       mClusterConf = ConfigurationUtils.getClusterConf(response, conf, Scope.CLIENT);
       mClusterConfHash = response.getClusterConfigHash();
--- a/core/common/src/main/java/alluxio/util/ConfigurationUtils.java
+++ b/core/common/src/main/java/alluxio/util/ConfigurationUtils.java
@@ -472,10 +472,12 @@ public final class ConfigurationUtils {
    * @param conf the existing configuration
    * @param ignoreClusterConf do not load cluster configuration related information
    * @param ignorePathConf do not load path configuration related information
+   * @param ignoreDefaultValue do not load configuration property with default value
    * @return the RPC response
    */
   public static GetConfigurationPResponse loadConfiguration(InetSocketAddress address,
-      AlluxioConfiguration conf, boolean ignoreClusterConf, boolean ignorePathConf)
+      AlluxioConfiguration conf, boolean ignoreClusterConf, boolean ignorePathConf,
+      boolean ignoreDefaultValue)
       throws AlluxioStatusException {
     GrpcChannel channel = null;
     try {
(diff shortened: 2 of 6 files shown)


bf8086:
@LuQQiu @apc999 PTAL.

apc999:
don't quite understand here: 

> For example, the default value for alluxio.underfs.hdfs.configuration is set to base on the configuration directory. If master and worker have different Alluxio directory for example in a containerized environment, the worker will sync the default path from master which does not exist in worker, causing worker to fail to talk to HDFS.

after this PR, what is the expected behavior to set `alluxio.underfs.hdfs.configuration` on workers? Do we still set it in `alluxio-site.properties` on workers or relie on system default value ?

bf8086:
> after this PR, what is the expected behavior to set `alluxio.underfs.hdfs.configuration` on workers? Do we still set it in `alluxio-site.properties` on workers or relie on system default value ?

After the PR, users can still set `alluxio.underfs.hdfs.configuration` on workers if they put `hdfs-site.xml` and `core-site.xml` in a specific location. Or if they just copy both xml files under Alluxio conf folder on workers, it will be automatically picked up by workers given the default value.

Here are some examples:

Before this PR:

| Value type  | Example 1 (property not set) | Example 2 (property set on master) | Example 3 (property set on both) |
| ------------- | ------------- |------------- |------------- |
| master site properties  | Not set  | /m/core-site.xml:/m/hdfs-site.xml | /m/core-site.xml:/m/hdfs-site.xml |
| worker site properties  | Not set  | Not set                 |/w/core-site.xml:/w/hdfs-site.xml |
| master evaluated configuration value  | master_alluxio_dir/conf/core-site.xml:master_alluxio_dir/conf/hdfs-site.xml |  /m/core-site.xml:/m/hdfs-site.xml | /m/core-site.xml:/m/hdfs-site.xml |
| worker evaluated configuration value  | master_alluxio_dir/conf/core-site.xml:master_alluxio_dir/conf/hdfs-site.xml (sync'ed from master)  | /m/core-site.xml:/m/hdfs-site.xml (sync'ed from master) | /w/core-site.xml:/w/hdfs-site.xml (worker overriden value) |

After this PR:

| Value type  | Example 1 (property not set) | Example 2 (property set on master) | Example 3 (property set on both) |
| ------------- | ------------- |------------- |------------- |
| master site properties  | Not set  | /m/core-site.xml:/m/hdfs-site.xml | /m/core-site.xml:/m/hdfs-site.xml |
| worker site properties  | Not set  | Not set                 |/w/core-site.xml:/w/hdfs-site.xml |
| master evaluated configuration value  | master_alluxio_dir/conf/core-site.xml:master_alluxio_dir/conf/hdfs-site.xml |  /m/core-site.xml:/m/hdfs-site.xml | /m/core-site.xml:/m/hdfs-site.xml |
| worker evaluated configuration value  | **worker_alluxio_dir/conf/core-site.xml:worker_alluxio_dir/conf/hdfs-site.xml (default value from worker)**  | /m/core-site.xml:/m/hdfs-site.xml (sync'ed from master) | /w/core-site.xml:/w/hdfs-site.xml (worker overriden value) |

So the only behavior changed is that worker no longer inherits from master default values. It only inherits value if the property is explicitly set on master.

PR #11169, opened 2020-03-15
Add docs check and apply fixes

Xenorith:


Xenorith:
jenkins, test this please

docs/cn/Overview.md
```diff
@@ -58,7 +58,7 @@ Alluxio 将三个关键领域的创新结合在一起，提供了一套独特的
 1. **智能多层级缓存**：Alluxio 集群能够充当底层存储系统中数据的读写缓存。可配置自动优化数据放置策略，以实现跨内存和磁盘（SSD/HDD）的性能和可靠性。缓存对用户是透明的，使用缓冲来保持与持久存储的一致性。有关详细信息，请参阅 [缓存功能文档]({{ '/cn/core-services/Caching.html' | relativize_url }})。
 1. **服务器端 API 翻译转换**：Alluxio支持工业界场景的API接口，例如HDFS API, S3 API, FUSE API, REST API。它能够透明地从标准客户端接口转换到任何存储接口。Alluxio 负责管理应用程序和文件或对象存储之间的通信，从而消除了对复杂系统进行配置和管理的需求。文件数据可以看起来像对象数据，反之亦然。
 
-要了解有关 Alluxio 内部的更多详细信息，请阅读 [Alluxio 架构和数据流]({{ '/cn/overview/Architecture.html' | relativize_url }})。
+要了解有关 Alluxio 内部的更多详细信息，请阅读 [Alluxio 架构和数据流]({{ '/en/overview/Architecture.html' | relativize_url }})。
```
ZacBlanco: cn docs should link to cn pages?

Xenorith: there is no cn page for this

docs/check.go
```diff
@@ -0,0 +1,294 @@
...
+// parseCategoryNames parses the given config file for the list of category names
+func parseCategoryNames(configPath string) (StringSet, error) {
+	f, err := os.Open(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("error opening file at %v: %v", configPath, err)
+	}
+	defer f.Close()
+
+	const categoryListLine, categoryPrefix = "categoryList:", "  - "
+	categoryNames := StringSet{}
```
ZacBlanco: I think we should make a struct and parse YAML instead of using this string manipulation

docs/cn/Overview.md
ZacBlanco: cn docs should link to cn pages?

docs/check.go
```diff
@@ -0,0 +1,294 @@
...
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("error scanning file: %v", err)
+	}
+	return categoryNames, nil
+}
+
+var (
+	// general format of a relative link, where the link will be computed by a jekyll function encapsulated in {{ }}
+	relativeLinkRe = regexp.MustCompile(`\[.+\]\({{.*}}\)`)
+	// path encapsulated in ' ' could have an optional search query "?q=queryStr" and/or an optional anchor reference "#anchor"
```
ZacBlanco: Is this only for links that link to other doc pages? 

For external links this doesn't capture all possible scenarios

docs/check.go
```diff
@@ -0,0 +1,294 @@
...
+	return nil
+}
+
+const groupKey = "group"
+
+var headerKeys = StringSet{
+	"layout":   {},
+	"title":    {},
+	"nickname": {},
+	groupKey:   {},
```
ZacBlanco: is this not supposed to be quoted?

docs/check.go
```diff
@@ -0,0 +1,294 @@
...
+func (ctx *checkContext) checkHeader(headers []string, mdFile string) {
+	for i := 0; i < len(headers); i++ {
+		l := headers[i]
+		if i == 0 || i == len(headers)-1 {
+			if l != "---" {
+				ctx.addError(mdFile, i, "header section should be surrounded by --- but was %v", l)
+			}
+			continue
+		}
+		split := strings.Split(l, ": ")
```
ZacBlanco: All content between `---` is yaml. We should use built-in deserializers if possible.

docs/check.go
Xenorith: parsing yaml -> vendoring code -> need to set GOPATH to execute
the generate tarball code needs to do the lattermost because it consists of several files, but i've been of the opinion of keeping go code in this repo as scripts. i want to avoid vendoring code unless we're completely on board with integrating all of the go code here into its own module as opposed to one-off scripts

docs/check.go
Xenorith: yes i do not want to validate any external links in a PR build

docs/check.go
Xenorith: it's a reference to the above constant

docs/check.go
Xenorith: same comment as above

docs/cn/Overview.md
Xenorith: there is no cn page for this, thus being caught by the check

ZacBlanco: we shouldn't link to the `en` side here. I recommend finding another page to link to

docs/check.go
ZacBlanco: I would prefer to remove the go code if possible. Since the docs use jekyll, it would probably be easier (assuming you know ruby) to write a verification program that can run as a [plugin](https://jekyllrb.com/docs/plugins/).

docs/cn/Overview.md
Xenorith: @LuQQiu can you suggest a similar page to link to or does one not exist? otherwise the alternative left is to delete the sentence

docs/check.go
Xenorith: never written ruby...

docs/cn/Overview.md
LuQQiu: Emm... I didn't find any page similar to architecture page... do you consider deleting this line?

ZacBlanco:
@Xenorith does the Alluxio-maven docker image even have go installed? I think you'll need to add that to the image as our builds will now depend on having `go` installed

Xenorith:
golang is already installed as part of the build image
https://github.com/Alluxio/alluxio/blob/master/dev/jenkins/Dockerfile#L23

and the new check is already part of this PR's verify

bf8086:

bf8086:

PR #11173, opened 2020-03-16
Change K8s Alluxio master to format journal in initContainer

jiacheliu3:
This moves the journal formatting logic from a standalone Job to an `initContainer` in the StatefulSet. This will resolve the conflicts against the journal volume between the master and job pods.


jiacheliu3:
I will rebase and resolve the helm chart version once the other K8s PRs are in.

integration/kubernetes/singleMaster-hdfsJournal/master/alluxio-master-statefulset.yaml.template
```diff
@@ -45,6 +45,7 @@ spec:
       nodeSelector:
       securityContext:
         fsGroup: 1000
+      initContainers:
```
jiacheliu3: This is because we moved the if-else to inside the `initContainers`. It's fine to have `initContainers` as empty.

jiacheliu3:
@madanadit Since the journal formatting job is removed. If the users are using `kubectl` to deploy Alluxio, they will need to modify the `StatefulSet` to either
```
          args:
            - master-only
            # remove this --no-format
            - --no-format
```
or add the `initContainer` themselves.

Is this an acceptable instruction to you?

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -60,8 +60,8 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: {{ .Values.fsGroup }}
-      {{- if and $needJournalVolume (not $isRoot) }}
       initContainers:
+      {{- if and $needJournalVolume (not $isRoot) }}
```
madanadit: is this `chown` still needed now that we use local pv instead of hostPath? i would like to get rid of `runAsUser: 0` as that might be unavailable in security constrained environments

jiacheliu3: Yea this should be removed.

jiacheliu3:
alluxio-bot, test this please

madanadit:

Issue #10253, opened 2019-10-24
Use local persistent volumes for embedded journal, single master local journal, off-heap metastore and tiered storage


madanadit:
**Is your feature request related to a problem? Please describe.**
Local persistent volumes is the preferred way for storage to persist across restarts of a pod (ref: https://kubernetes.io/docs/concepts/storage/volumes/#local). We should use persistent volumes for the following:
1. embedded journal volume for master pods
2. single master journal volume for master pod
3. off-heap metastore volume for master pods
4. tiered storage for worker pods

**Describe the solution you'd like**
Helm charts should be extended to support local persistent volumes w/ resource limits (e.g., size of journal / tiered storage). Documentation w/ and w/o helm should be added as part of this ask. Note: hostPath volumes cannot be provisioned with resource limits and require root user and are hence not desired for persistent local storage.

1. embedded journal volume for master pods: restarts of the master pod w/o persistent volumes will lead to loosing metadata. 
- The helm template `statefulset.yaml` should use persistentVolumeClaims instead of `emptyDir` volumes. 
- `journal-pvc.yaml` should create `masterCount` claim resources
- `journal-pv.yaml` should be changed to use local persistent volume instead of `hostPath` by default

2. single master journal volume for master pod
- As part of pv for embedded journal we may choose to rename the claim from `alluxio-pv-claim` to `alluxio-pv-claim-<i>` where `i=0` for a single master. 

3. off-heap metastore volume for master pods
- Switch from `emptyDir` volume to localPV with limits

4. tiered storage for worker pods
- Switch from unbounded `hostPath` volume to localPV with limits

**Describe alternatives you've considered**
Advanced end-user customizes the specs on their own

**Urgency**
It would be painful w/o extensive knowledge of Alluxio to roll out raw specifications for these features to be production ready. This deployment will not be possible w/ helm (w/o this feature).

**Additional context**
N/A

madanadit:
@cheyang FYI

jiacheliu3:
Ok

jiacheliu3:
1) and 2) are resolved by https://github.com/Alluxio/alluxio/pull/10686

jiacheliu3:
After some research, 4(worker tier storage) can be resolved by switching from using `emptyDir` like below
```
      volumes:
        - name: mem
          emptyDir:
            medium: "Memory"
            sizeLimit: 1G
```
to a PVC.
```
      volumes:
        - name: mem
          persistentVolumeClaim:
[...]

PR #11171, opened 2020-03-15
Support PVC for worker tiered store in K8s

jiacheliu3:
This resolves worker tiered storage using PVC, as point 4 in https://github.com/Alluxio/alluxio/issues/10253

The PVC will be satisfied by either `hostPath` or `local` under the hood to provide local storage to the worker.

PR #11175, opened 2020-03-16
Change Alluxio master in K8s to use PVC for RocksDB

jiacheliu3:
This resolves point 3 in https://github.com/Alluxio/alluxio/issues/10253

Alluxio master Pods now use PVC for RocksDB as metastore instead of `emptyDir`.

### DIFF
--- a/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
+++ b/docs/en/deploy/Running-Alluxio-On-Kubernetes.md
@@ -211,8 +211,8 @@ secrets:
 
 ***Example: Off-heap Metastore Management***
 
-The following configuration provisions an `emptyDir` volume with the specified configuration and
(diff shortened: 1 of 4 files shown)


jiacheliu3:
@madanadit Do you think it makes sense to take `emptyDir` off the options? Now we are switching to only supporting a PVC for Rocks. Do you think `emptyDir` should still exist as a valid option, maybe the user just wants a try?

madanadit:
@jiacheliu3  we should remove the option to prevent mis-configuration and confusion

jiacheliu3:
@madanadit Originally I was planning to update the docs in one go after the batch of PRs are in, but I guess putting the docs next to the code change makes more sense.

I will bump on the helm chart version after #11171, #11173 and this one are all in.

madanadit:

PR #11179, opened 2020-03-16
Enable local cache only when the process type is client

calvinjia:
Our master/workers also use clients, and they should not use the client cache. This should be fixed in a different way when we can better scope our property keys and separate the client-side client from the server-side client.


calvinjia:
jenkins, test this please.

calvinjia:
Failed tests are flakey and unrelated to this PR.

calvinjia:

calvinjia:

PR #11183, opened 2020-03-17
Improve metadata syncing behavior

ZacBlanco:
This PR introduces a few optimizations to reduce the time required when syncing metadata from a UFS.

There are main functions which require processing time and locking

- populateStatusCache (write lock)
- syncInodeMetadata (write lock)
- loadMetadataInternal (read lock, write when required)

This PR makes the following optimizations:

- Moves the populateStatusCache function outside of the write lock. This allows more concurrent reads against the filesystem tree while a sync is occurring.
- loadMetadataInternal now re-uses the statusCache that is generated for syncInodeMetadata.
- The processing for each status in loadMetadataInternal is also now done in chunks (based on tree depth) which is then be parallelized.

The main performance gains from this PR stem from the ability to re-use the cache and process UfsStatus objects in parallel within loadMetadataInternal.


ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2407,69 +2417,125 @@ private void loadMetadataInternal(RpcContext rpcContext, LockedInodePath inodePa
...
+    }
+
+    try (LockedInodePath descendant = inodePath.lockDescendant(
+        inodePath.getUri().joinUnsafe(childStatus.getName()), LockPattern.READ)) {
+      LoadMetadataContext loadMetadataContext = LoadMetadataContext
+          .mergeFrom(LoadMetadataPOptions.newBuilder()
+              .setLoadDescendantType(LoadDescendantPType.NONE).setCreateAncestors(false))
+          .setUfsStatus(childStatus);
+      try {
+        loadMetadataInternal(rpcContext, descendant, loadMetadataContext, null, false);
```
gpang: Why is parallelize `false` here? Can we leave a comment here?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2373,6 +2376,13 @@ public UfsInfo getUfsInfo(long mountId) {
    * @param rpcContext the rpc context
    * @param inodePath the path for which metadata should be loaded
    * @param context the load metadata context
+   * @param statusCache UFS statuses which has already been retrieved for the {@code inodePath}
+   *                    This can be set to null if no status cache is available
+   * @param shouldParallelize if set to true and we are loading metadata for child inodes, then it
+   *                          will load metadata for separate paths concurrently. This should only
+   *                          be set to true if the calling thread doesn't already hold a write
+   *                          lock. If a write lock is held, and this parameter is true, this will
```
gpang: `write lock is held`: Isn't there a way to see if a locked inode path has a write lock, and automatically be able to parallelize or not based on that result?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: These calls set the `LoadDescendantType` as `NONE`, so setting this to true or false doesn't really have an effect. We only parallelize when the load type is `ALL`.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: As far as I'm aware, calling `#getLockPattern` gets the lock pattern for the currenty inode path, but doesn't tell us if there is a path further up in the tree that is write locked.
i.e. if you write-locked `/` but then load metadata on `/mnt` which is read locked, the `lockedInodePath#getLockPattern` will return `READ`. Deadlock can occur if any inode or edge in the current path is `WRITE_*`

gpang: Yeah, I know. What if we added a method to LockedInodePath to see if there was any write lock along the path? That would help to automatically parallelize?

ZacBlanco: I think that would be a welcome addition. It would probably simplify the current implementation

ZacBlanco:
Closing this PR as there are currently too many issues with concurrent locking to add the concurrency portion into the PR. The rest of the work has been transferred here: #11193

PR #11191, opened 2020-03-18
Move fuse module back to default profile

bf8086:
Sometimes we still need to build Alluxio without specify a hadoop profile. Move fuse back to the default profile to ensure it is built without an explicit profile.

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -21,6 +21,10 @@
   <packaging>pom</packaging>
   <name>Alluxio Integration</name>
   <description>Parent POM for different integrations with Alluxio</description>
+  <modules>
+    <!-- exclude checker from hadoop-1 profile as yarn is not available in Hadoop 1.x -->
+    <module>fuse</module>
+  </modules>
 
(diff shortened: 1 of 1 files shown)


bf8086:
@apc999 PTAL.

integration/pom.xml
```diff
@@ -21,32 +21,27 @@
...
-      <id>hadoop-1</id>
-      <modules>
-        <!-- exclude checker from hadoop-1 profile as yarn is not available in Hadoop 1.x -->
-        <module>fuse</module>
-      </modules>
-    </profile>
     <profile>
       <id>hadoop-2</id>
       <modules>
         <module>checker</module>
```
apc999: unrelated to this:  maybe we can remove checker module --- not really see this used

bf8086:

integration/pom.xml
bf8086: agree, user just need to build it separately.

bf8086:

PR #11192, opened 2020-03-19
Fix categoryList parsing in docs check

Xenorith:
moved go code in dev/scripts alongside tarball go code and vendored gopkg.in/yaml.v3
users can rerun the doc check the same way as jenkins, execute `./dev/scripts/check-docs.sh`

i decided against adding a binary into the dockerfile because it would be clunky to run locally. i don't want to require users to install yq to revalidate the check or require them to run `go get` that would write files wherever their GOPATH happens to be.


docs/check.go
```diff
@@ -145,11 +145,16 @@ func parseCategoryNames(configPath string) (StringSet, error) {
...
+				// skip commented lines
 				continue
 			}
-			categoryNames.Add(strings.TrimPrefix(l, categoryPrefix))
+			if strings.HasPrefix(l, categoryPrefix) {
+				categoryNames.Add(strings.TrimPrefix(l, categoryPrefix))
+				continue
+			}
+			// stop parsing if line is neither comment or category
+			found = false
```
ZacBlanco: would it just be easier to add a yaml parsing command line tool in the CI container, and then use that to query for this?
ex:  https://kislyuk.github.io/yq/

Xenorith: pip -> python -> i have nothing more to say
we don't have python in the image, i would have given more consideration if it was already present and used for something else

Xenorith: but the next search result:
https://github.com/mikefarah/yq

ZacBlanco: I am ok adding to the CI image. It will help make the code here more readable

Xenorith: ok adding `python + pip + yq` or `wget yq && mv /usr/local/bin/`?

ZacBlanco: Isn't go installed? Can we not use `GO111MODULE=on go get github.com/mikefarah/yq/v3` from the link you gave above?

ZacBlanco: Either way, I'm good with adding the go-based one

Xenorith:
jenkins, test this please

ZacBlanco:

PR #11193, opened 2020-03-20
Improve sync performance and reduce write lock time

ZacBlanco:
This PR introduces a few optimizations to reduce the time required when syncing metadata from a UFS.

There are main functions which require processing time and locking

    populateStatusCache (write lock)
    syncInodeMetadata (write lock)
    loadMetadataInternal (read lock, write when required)

This PR makes the following 

- Moves the populateStatusCache function outside the write lock. This allows more concurrent reads against the filesystem tree while a sync is in progress
- loadMetadataInternal now re-uses the statusCache that is generated for syncInodeMetadata.

The main performance gains from this PR stem from the ability to re-use the cache


core/base/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * A class to hold a set of statuses which map from AlluxioURI to UfsStatus.
+ */
+public class UfsStatusCache {
+  private final Map<AlluxioURI, UfsStatus> mStatuses;
+  private final AlluxioURI mUfsUri;
```
gpang: Can you add a comment on what this is? Also, will this URI also be in the map above?

core/base/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,69 @@
...
+
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * A class to hold a set of statuses which map from AlluxioURI to UfsStatus.
+ */
+public class UfsStatusCache {
+  private final Map<AlluxioURI, UfsStatus> mStatuses;
```
gpang: are these uris in the alluxio namespace or in the ufs namespace?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2407,45 +2410,51 @@ private void loadMetadataInternal(RpcContext rpcContext, LockedInodePath inodePa
         loadDirectoryMetadata(rpcContext, inodePath, context);
 
         if (context.getOptions().getLoadDescendantType() != LoadDescendantPType.NONE) {
-          ListOptions listOptions = ListOptions.defaults();
-          if (context.getOptions().getLoadDescendantType() == LoadDescendantPType.ALL) {
-            listOptions.setRecursive(true);
+          UfsStatus[] children;
+          if (statusCache != null
+              && context.getOptions().getLoadDescendantType() == LoadDescendantPType.ALL) {
+            // Avoid re-caching. Makes the assumption the cache was created on the same ufs URI
```
gpang: Does this assume the status cache must have been created with ALL (or recursive)? Is that a valid assumption?

Also, what does "avoid re-caching" mean? I don't see where the caching is happening.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2628,6 +2637,7 @@ private void loadDirectoryMetadata(RpcContext rpcContext, LockedInodePath inodeP
       createDirectoryContext.setOperationTimeMs(lastModifiedTime);
     }
 
+    LOG.warn("lockFinalWriteEdge: {}", inodePath);
```
gpang: Remove this debugging logging?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3236,16 +3246,23 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
       if (changedFiles == null) {
         // full sync
         LockingScheme lockingScheme = new LockingScheme(path, LockPattern.READ, true);
+        // Populate the status cache before taking a write lock
+        UfsStatusCache statusCache;
+        try (LockedInodePath inodePath =
+            mInodeTree.lockInodePath(lockingScheme.getPath(), LockPattern.READ)) {
```
gpang: What would be the implications of not even doing any read locking and just calling `populateStatusCache`? I'm not sure what the read lock is providing.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3328,6 +3345,9 @@ private boolean syncMetadata(RpcContext rpcContext, LockedInodePath inodePath,
         try {
           UfsStatus[] children = ufs.listStatus(ufsUri.toString(), listOptions);
           if (children != null) {
+            // Since we know the number of elements, make sure we don't re-hash
+            // 0.75f is the default
+            statusCache = new HashMap<>(Math.round(children.length * 1.34f), 0.75f);
```
gpang: Why do we need to keep the .75 load factor? Can we use a larger load factor?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3561,7 +3581,7 @@ private SyncResult syncInodeMetadata(RpcContext rpcContext, LockedInodePath inod
       if (syncPlan.toUpdateMetaData()) {
         // UpdateMetadata is used when a file or a directory only had metadata change.
         // It works by calling SetAttributeInternal on the inodePath.
-        if (ufsFpParsed.isValid()) {
+        if (ufsFpParsed != null && ufsFpParsed.isValid()) {
```
gpang: When would `ufsFpParsed` be `null`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3608,25 +3633,29 @@ private SyncResult syncInodeMetadata(RpcContext rpcContext, LockedInodePath inod
...
+            if (type != DescendantType.ALL) {
+              type = DescendantType.NONE;
             }
-            SyncResult syncResult =
-                syncInodeMetadata(rpcContext, descendant, syncDescendantType, statusCache);
-            pathsToLoad.addAll(syncResult.getPathsToLoad());
+            return syncInodeMetadata(rpcContext, descendant, type, statusCache);
+          } catch (InvalidPathException | FileDoesNotExistException | IOException
+              | AccessControlException e) {
+            LOG.error("Failed to process inode path: {}", inodeEntry.getValue().toString(), e);
```
gpang: Is there more info we can add to this error message? For example, should this say "sync" instead of "process"?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: The read lock would prevent a mount from being removed before we resolve the mount path

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: We can, but it could increase the time to look up items in the map. I didn't want to change the performance characteristics of the status cache

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: I added this because intelliJ was complaining that `ufsFpParsed` could be null. Specifically, `Fingerprint.parse(ufsFingerprint);` a few lines above could return null

core/base/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: I will add a comment.

`mUfsUri` is used before calling `loadMetadataInternal` to see if the cache that was populated previously corresponds to the same path.

The mapping is alluxio path -> UfsStatus

core/base/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: Alluxio namespace

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: There are only two callers of `populateStatusCache`, both of which specify `ALL`.

I can add a parameter to the status cache which can identify the load type that the cache was created with. If it's not `ALL` and is not equal to the current LoadDescendantType, then I would probably just forgo using the cache. Most performance gain is only going to come when we have a huge tree after doing a recursive listing (called with `ALL`).

ZacBlanco: "re-caching" is an error. It should probably say "avoid retrieving statuses again"

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: updated

core/base/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: comments added

core/base/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: comments added

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2407,45 +2410,51 @@ private void loadMetadataInternal(RpcContext rpcContext, LockedInodePath inodePa
...
 
         if (context.getOptions().getLoadDescendantType() != LoadDescendantPType.NONE) {
-          ListOptions listOptions = ListOptions.defaults();
-          if (context.getOptions().getLoadDescendantType() == LoadDescendantPType.ALL) {
-            listOptions.setRecursive(true);
+          UfsStatus[] children;
+          if (statusCache != null
+              && context.getOptions().getLoadDescendantType() == LoadDescendantPType.ALL) {
+            // Avoid re-caching. Makes the assumption the cache was created on the same ufs URI
+            children = statusCache.values().toArray(new UfsStatus[0]);
```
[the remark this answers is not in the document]
ZacBlanco: Currently, I perform this check here: https://github.com/Alluxio/alluxio/blob/e85c520726318107e7c563f06fceec577daa61a6/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java#L3422-L3426

Though it might make more sense to put it within the function? What do you think?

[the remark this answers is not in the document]
yuzhu: yeah, let's put it in the function, so people like me won't miss it.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: From my understanding: higher load factor -> smaller array size -> more likely to have hash collisions -> longer lookup times

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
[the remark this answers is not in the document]
ZacBlanco: updated

ZacBlanco:
closing this in favor of the upcoming PR #11204

PR #11204, opened 2020-03-24
Improve metadata sync operations

ZacBlanco:
This PR improves the "syncMetadata" operation used throughout the DefaultFileSystemMaster.

This approach incrementally locks and syncs the paths in the tree instead of taking a write lock on the root of the tree. The "loadMetadata" step of syncing  is now contained within the "syncInodeMetadata" function.

It is implemented with a BFS. The previous implementation used DFS which could have been prone to stack overflows if a user created a deep enough file hierarchy.

Additionally, this implementation makes the minimum amount of RPC calls required to the UFS. UfsStatuses are re-used as much as possible, and discarded once they are no longer required.


ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
Leaving a note for myself that I'll need to also implement the changes from @madanadit in https://github.com/Alluxio/alluxio/pull/11257 so we don't have a regression :)

madanadit:
Thanks @ZacBlanco. I should add a test, probably mocking S3.

core/common/src/main/java/alluxio/resource/LockResource.java
```diff
@@ -35,24 +39,39 @@
...
+  public LockResource(Lock lock, boolean acquireLock, boolean useTryLock) {
     mLock = lock;
     if (acquireLock) {
-      mLock.lock();
+      if (useTryLock) {
+        while (!mLock.tryLock()) { // returns immediately
+          // The reason we don't use #tryLock(int, TimeUnit) here is because we found there is a bug
+          // somewhere in the internal accounting of the ReentrantRWLock that, even though all
+          // threads had released the lock, that a final thread would never be able to acquire it.
+          LockSupport.parkNanos(10000);
```
gpang, line 65: Should this be a parameter? Also, do we want to loop indefinitely, or throw some sort of exception when it was looping for too long? I'd imagine throwing an exception would prevent possible deadlocks or long delays, but failures would be less-user friendly. What do you think?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2132,6 +2132,30 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_METADATA_SYNC_CONCURRENCY_LEVEL =
+      new Builder(Name.MASTER_METADATA_SYNC_CONCURRENCY_LEVEL)
+          .setDefaultValue(6)
+          .setDescription("The maximum number of concurrent sync tasks running for a given sync "
```
gpang, line 2138: How does this relate to the other parameters, MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE and MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE? Also, it would be helpful to state what the potential tradeoffs are for a higher vs lower value (for the other parameters as well).

tests/src/test/java/alluxio/client/fs/LoadMetadataIntegrationTest.java
```diff
@@ -219,15 +219,9 @@ public void loadRecursive() throws Exception {
         fileWriter.close();
       }
     }
-    long startMs = CommonUtils.getCurrentMs();
     List<URIStatus> list = mFileSystem.listStatus(new AlluxioURI("/mnt"), options);
-    long durationMs = CommonUtils.getCurrentMs() - startMs;
     // 25 files, 25 level 2 dirs, 5 level 1 dirs, 1 file and 1 dir created in before
     Assert.assertEquals(25 * 2 + 5 + 2, list.size());
-
-    // Should load metadata once, in one recursive call
```
gpang, line 228: Is this test useful anymore, if it is not validating how the load is happening?

ZacBlanco: I think validating the listStatus size is better than validating the runtime? Runtime can vary for a lot of reasons during tests- especially in jenkins if a system is overloaded.

We might be able to come up with some better assertions, but I think the test is still validating that we load metadata since all of the file creations above are through the UFS, and then the listStatus call is executed through Alluxio, so we verify that at least all of the files expected are loaded in Alluxio under the `/mnt` path.

core/client/fs/src/test/java/alluxio/client/metrics/MetricsHeartbeatContextTest.java
```diff
@@ -41,8 +41,8 @@
 
   @Test
   public void testExecutorInitialized() {
-
     InstancedConfiguration conf = ConfigurationTestUtils.defaults();
+    conf.set(PropertyKey.MASTER_HOSTNAME, "localhost");
```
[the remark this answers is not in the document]
ZacBlanco: This test was failing for me locally without that change...I can try reverting it

core/common/src/main/java/alluxio/resource/LockResource.java
ZacBlanco: The current implementation before `tryLock` will also hang indefinitely, so I felt it was okay to not have any timeout for the `tryLock` code path.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: Good question

- `MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE`
  - This parameter determines exactly how many inodes can be processing sync in parallel.
  - The executor affected by this value is shared by _all_ RPC calls to the master. So if you have 64 total threads in this executor, then a syncMetadata operation from a `listStatus` will compete for threads (submit tasks to the same executor) as a `createDirectory` RPC.

Now imagine you have a user who does a `listStatus` RPC that is high up in the inode tree, say, `/dir1`; but then another user comes along who wants to call `getStatus` on `/dir2`. It is mutually exclusive from `/dir1`, and may trigger a metadata sync, but if we allow `listStatus` on `/dir1` to use all threads within this executor, concurrent sync operations could suffer in performance, even if `/dir2` does not contain a large tree structure.

- `MASTER_METADATA_SYNC_CONCURRENCY_LEVEL` restricts max #of threads a single RPC is allowed to use at any given time for processing an inode sync.

The reason that `MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE` is separate is because in order to make progress during sync, ufs prefetch threads must be available, and so if all threads in `MASTER_METADATA_SYNC_CONCURRENCY_LEVEL` are taken and none can be used for ufs fetching, then we could run into a deadlock scenario. There is no limit per sync operation on how many threads can be used for prefetch. Though the throughput of requests to the prefetch pool is limited to an extent by the `MASTER_METADATA_SYNC_CONCURRENCY_LEVEL`

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+   * All child statuses added via this method will be available via {@link #getStatus(AlluxioURI)}.
+   *
+   * @param path the directory inode path which contains the children
+   * @param children the children of the {@code path}
+   * @throws IllegalArgumentException when {@code path} already exists or if any child already
+   *                                  exists
+   */
+  public void addChildren(AlluxioURI path, Collection<UfsStatus> children) {
+    UfsStatus status = mStatuses.get(path);
+    // If this path doesn't yet exist, we can't keep track of the parent-child relationship
```
gpang, line 92: When is this parent-child relationship used?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+    // We can still add statuses to the cache regardless
+    if (status != null) {
```
gpang, line 94: Why can't we add this parent->children listing cache even if the parent status is not in the other map?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+   * Remove a status from the cache.
+   *
+   *  Any children added to this status will remain in the cache.
+   *
+   * @param path the path corresponding to the {@link UfsStatus} to remove
+   * @return the removed UfsStatus
+   */
+  public UfsStatus remove(AlluxioURI path) {
+    UfsStatus removed = mStatuses.remove(path);
+    if (removed == null) {
```
gpang, line 113: Is the contract that if the status DNE, then the children list DNE for that path? It is unclear what the relationship is between the 2 maps.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+   * @param children the children of the {@code path}
+   * @throws IllegalArgumentException when {@code path} already exists or if any child already
+   *                                  exists
+   */
+  public void addChildren(AlluxioURI path, Collection<UfsStatus> children) {
+    UfsStatus status = mStatuses.get(path);
+    // If this path doesn't yet exist, we can't keep track of the parent-child relationship
+    // We can still add statuses to the cache regardless
+    if (status != null) {
+      mChildren.computeIfAbsent(path, ufsStatus -> new ConcurrentHashSet<>()).addAll(children);
```
gpang, line 95: What if it was already cached? The children cache would be stale?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+    return mStatuses.get(path);
+  }
+
+  /**
+   * Fetches children of a given alluxio path, stores them in the cache, then returns them.
+   *
+   * Children can be returned in a few ways
+   * 1. Children already exist in the internal index. We simply return them
+   * 2. If children did not already exist in the index, then check if there was a scheduled
+   * prefetch job running for this path. If so, wait for the job to finish and return the result.
```
gpang, line 137: What happens if there is no scheduled prefetch job? Is that what (3) is?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+   */
+  @Nullable
+  public Collection<UfsStatus> fetchChildrenIfAbsent(AlluxioURI path, MountTable mountTable,
+      boolean useFallback)
+      throws IOException, InvalidPathException {
+    Collection<UfsStatus> children = getChildren(path);
+    if (children != null) {
+      return children;
+    }
+    Future<Collection<UfsStatus>> prefetchJob = mActivePrefetchJobs.remove(path);
```
gpang, line 157: Removing it means other callers which call this same method would think there are no prefetch jobs? Is that intended?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+    if (prefetchJob != null) {
+      try {
+        return prefetchJob.get();
+      } catch (InterruptedException | ExecutionException e) {
+        LOG.warn("Failed waiting to fetch children at {}", path);
```
gpang, line 162:
```suggestion
        LOG.warn("Failed waiting to prefetch children at {}", path);
```

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+        if (e instanceof  InterruptedException) {
```
gpang, line 163:
```suggestion
        if (e instanceof InterruptedException) {
```

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+   *
+   * @param path the path to prefetch
+   * @param mountTable the Alluxio mount table
+   */
+  public void prefetchChildren(AlluxioURI path, MountTable mountTable) {
+    if (mPrefetchExecutor == null) {
+      return;
+    }
+    try {
+      Future<Collection<UfsStatus>> fute =
```
gpang, line 255: hrmmm, is `fute` a real term? I am not familiar with it...

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+          mPrefetchExecutor.submit(() -> getChildrenIfAbsent(path, mountTable));
+      Future<Collection<UfsStatus>> prev = mActivePrefetchJobs.put(path, fute);
+      if (prev != null) {
+        prev.cancel(true);
```
gpang, line 259: Why cancel the old one? Could this cause an endless cycle of constantly cancelled tasks?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+      // process the sync result for the original path
+      try {
+        mRootPath.traverse();
+      } catch (InvalidPathException e) {
+        throw new RuntimeException(e);
+      }
+    } catch (AccessControlException | BlockInfoException | FileAlreadyCompletedException
+        | FileDoesNotExistException | InvalidFileSizeException | InvalidPathException
+        | IOException e) {
+      LOG.warn("FAILED TO SYNC METADATA: {}", e.getMessage(), e);
```
gpang, line 254: Does this have to be all caps?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+    } finally {
+      // regardless of the outcome, remove the UfsStatus for this path from the cache
+      mStatusCache.remove(mRootPath.getUri());
```
gpang, line 257: why remove the status?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+      // downgrade so that if operations are parallelized, the lock on the root doesn't restrict
+      // concurrent operations
+      mRootPath.downgradeToPattern(LockPattern.READ);
+    }
+
+    // Process any children after the root.
+    while (!mSyncMetadataQ.isEmpty() || !mSyncPathJobs.isEmpty()) {
+      if (Thread.currentThread().isInterrupted()) {
+        LOG.warn("Metadata syncing was interrupted before completion");
```
gpang, line 266: Please add the root path to message

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+        try {
+          // we synced the path successfully
+          if (job.get()) {
+            syncPathCount++;
+          }
+        } catch (InterruptedException | ExecutionException e) {
+          if (e instanceof  InterruptedException) {
+            Thread.currentThread().interrupt();
+          }
+          LOG.warn("metadata sync job was interrupted while waiting for completion");
```
gpang, line 291: Please add root path to message

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+          throw new IllegalStateException("Last node to be de-queued was not equal to the expected"
+              + "head of queue");
+        }
+        try {
+          // we synced the path successfully
+          if (job.get()) {
+            syncPathCount++;
+          }
+        } catch (InterruptedException | ExecutionException e) {
+          if (e instanceof  InterruptedException) {
```
gpang, line 288:
```suggestion
          if (e instanceof InterruptedException) {
```

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+        }
+        // remove the job because we know it is done.
+        if (mSyncPathJobs.poll() != job) {
+          throw new IllegalStateException("Last node to be de-queued was not equal to the expected"
+              + "head of queue");
+        }
+        try {
+          // we synced the path successfully
+          if (job.get()) {
+            syncPathCount++;
```
gpang, line 285: if some of the tasks failed for some reason, will the count accounting still work? If I want to load a directory with 10 children, but 1 failed, will it still stop after all 10 finished (fail or succeed)?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+      try {
+        oldestJob.get(); // block until the oldest job finished.
+      } catch (InterruptedException | ExecutionException e) {
+        if (e instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+        }
+        LOG.warn("Interrupted while waiting for metadata sync job to finish", e);
+      }
+    }
+    LOG.info("TRACING - Synced {} paths", syncPathCount);
```
gpang, line 325: debug logging?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+        Future<Boolean> job = mMetadataSyncService.submit(() -> processSyncPath(path));
+        mSyncPathJobs.offer(job);
+      }
+      // After submitting all jobs wait for the job at the head of the queue to finish.
+      Future<Boolean> oldestJob = mSyncPathJobs.peek();
+      if (oldestJob == null) { // There might not be any jobs, restart the loop.
+        continue;
+      }
+      try {
+        oldestJob.get(); // block until the oldest job finished.
```
gpang, line 317: Does it matter if the oldest job finishes first or last? How would that affect behavior?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+      } catch (InterruptedException | ExecutionException e) {
+        if (e instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+        }
+        LOG.warn("Interrupted while waiting for metadata sync job to finish", e);
```
gpang, line 322: add root path to message

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+      if (Thread.currentThread().isInterrupted()) {
+        LOG.warn("Thread syncing {} was interrupted before completion", inodePath.getUri());
+        return false;
+      }
+      syncInodeMetadata(inodePath);
+      return true;
+    } catch (AccessControlException | BlockInfoException | FileAlreadyCompletedException
+        | FileDoesNotExistException | InvalidFileSizeException | InvalidPathException
+        | IOException e) {
+      LOG.warn("FAILED TO SYNC METADATA: {}", e.getMessage(), e);
```
gpang, line 365: all caps?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: This will drop all references to the statuses allowing the JVM to GC it. If we don't remove it then for a huge namespace the statusCache can become large

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: Technically for more complicated directory structures it depends, (directory high up can fail, causing its children not to sync) but generally the accounting is only for paths which successfully synced. So if all paths fail, then this function returns false. If at least one path syncs, we return true.

We can keep track of the number of failed paths too and report them as debug logs?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: I just take the oldest job submitted, because as long as the threadpool uses a fair queueing policy for running jobs, it has the highest probability of being completed before other jobs. If the oldest job runs slower, it can theoretically prevent other tasks from being submitted.

In a follow-up I'd actually like to design a ThreadPoolExecutor that doesn't need this control logic and can simply just iterate over the queue of pending paths and submit them. My idea is that the executor pool can be shared in a limited manner, so that you can lease out a portion of the threads. ex:

```java
ExecutorService mainExecutor = Executors.newCachedThreadpool(64);

sharedExecutor = mainExecutor.lease(4); // executor service that uses the pool backing mainExecutor, that can have up to 4 tasks running concurrently.
sharedExecutor = mainExecutor.lease(32); // executor service that uses the pool backing mainExecutor, that can have up to 32 tasks running concurrently.
```

We can use a semaphore to block submitting tasks to the executor when they are already runing. Completely negates all the control logic of this thread. I haven't written the executor yet. This is all ideas yet. I don't plan on implementing it for this PR

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+   * The last component of the path in the {@link AlluxioURI} must match the result of
+   * {@link UfsStatus#getName()}. This method overrides any status currently cached for the same
+   * URI.
+   *
+   * @param path the Alluxio path to key on
+   * @param status the ufs status to store
+   * @throws IllegalArgumentException if the status name doesn't match the final URI path component
+   */
+  public void addStatus(AlluxioURI path, UfsStatus status) {
+    UfsStatus prev = mStatuses.putIfAbsent(path, status);
```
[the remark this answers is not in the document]
ZacBlanco: since it's a cache, I just ignore the previous value. We should never add the same status twice to the cache with the current code path, but if it does happen I think overwriting it is fine.

I will update this method to return the old status if there is one. No one in the current code will use the result though.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+    if (!path.getName().equals(status.getName())) {
```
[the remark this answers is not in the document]
ZacBlanco: ooh..good catch

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: [here](https://github.com/Alluxio/alluxio/blob/aea2599ba2565e052e3871653282b3f22a325203/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java#L510-L511) and [here](https://github.com/Alluxio/alluxio/blob/aea2599ba2565e052e3871653282b3f22a325203/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java#L602-L603)

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: ah, this was a limitation of my previous implementation. It is possible now.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: the child map is a 1:N mapping of alluxio directory to UFS children.

If we remove the parent status from the cache, then it's assumed that we won't need to access its parent-child relationship information either. The UfsStatus information for each individual child still resides in mStatuses after calling `remove`

ZacBlanco: Another solution is to make the parent-child relationship removal independent of the removing a single status, but then status removal will need 2 calls, and I think that is more prone to errors. We want to remove those references otherwise the JVM won't be able to GC the children if they are removed individually.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: `If no prefetch job` - isn't that what 3 says?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: In the current code path, we never have concurrent callers on the same path, but I will change the code so that it does work for concurrent callers.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated to `job`

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: How could it cause an endless cycle?

It only returns the previously submitted job for the given path. We have two options if there is already a prefetch job running on that path.

1. (current approach) Cancel the old one, submit the new one
2. Ignore the current request and rely on the future for the previously submitted future.

This method shouldn't really occur in the current code, but in the case it does happen I wanted to define the behavior. If the task is submitted later, I make the assumption that the user is submitting it because of some possible updated data in the UFS. The current prefetch job may already be executing or may be finished. I cancel it anyway in case it is running or about to run, and then submit the next one so whoever goes looking for the children later on will have a likely more up-to-date version of the UfsStatuses.

I don't see how this could cause an endless cycle?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+   * Whether the caller is {@link FileSystemMaster#getFileInfo(AlluxioURI, GetStatusContext)}.
+   * This is used for the {@link #mUfsSyncPathCache}.
+   */
+  private final boolean mIsGetFileInfo;
+
+  /** Whether to only read+create metadata from the UFS, or to update metadata as well. */
+  private final boolean mLoadOnly;
+
+  /** Queue used to keep track of paths that still need to be synced. */
+  private final ConcurrentLinkedQueue<AlluxioURI> mSyncMetadataQ;
```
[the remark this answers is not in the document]
ZacBlanco: yeah, i like that one better :D

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -0,0 +1,810 @@
...
+  private final MountTable mMountTable;
+
+  /** The lock manager used to try acquiring the persisting lock. */
+  private final InodeLockManager mInodeLockManager;
+
+  /** The FS master creating this object. */
+  private final DefaultFileSystemMaster mFsMaster;
+
+  /** Set this to true to force a sync regardless of the UfsPathCache. */
+  private final boolean mShouldSync;
```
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated to always overwrite

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,275 @@
...
+
+  private final ConcurrentHashMap<AlluxioURI, UfsStatus> mStatuses;
+  private final ConcurrentHashMap<AlluxioURI, Future<Collection<UfsStatus>>> mActivePrefetchJobs;
+  private final ConcurrentHashMap<AlluxioURI, Collection<UfsStatus>> mChildren;
+  private final ExecutorService mPrefetchExecutor;
+
+  /**
+   * Create a new instance of {@link UfsStatusCache}.
+   *
+   * @param prefetchExecutor the executor service used to prefetch statuses
```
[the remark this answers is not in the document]
ZacBlanco: updated

core/common/src/main/java/alluxio/collections/LockPool.java
```diff
@@ -77,8 +82,20 @@ public LockPool(Function<? super K, ? extends ReentrantReadWriteLock> defaultLoa
...
+
+  @Override
+  public void close() throws IOException {
+    mEvictorTask.cancel(true);
+    mEvictor.shutdownNow(); // immediately halt the evictor thread.
+    try {
+      mEvictor.awaitTermination(2, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new IOException("Failed to await evictor termination", e);
```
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+    try {
+      Future<Collection<UfsStatus>> job =
+          mPrefetchExecutor.submit(() -> getChildrenIfAbsent(path, mountTable));
+      Future<Collection<UfsStatus>> prev = mActivePrefetchJobs.put(path, job);
+      if (prev != null) {
+        prev.cancel(true);
+      }
+    } catch (RejectedExecutionException e) {
+      LOG.debug("Failed to submit prefetch job", e);
```
gpang, line 266: add path to message

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,279 @@
...
+    try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
+      UnderFileSystem ufs = ufsResource.get();
+      UfsStatus[] statuses = ufs.listStatus(ufsUri.toString());
+      if (statuses == null) {
+        return null;
+      }
+      children = Arrays.asList(statuses);
+      addChildren(path, children);
+    } catch (IllegalArgumentException | IOException e) {
+      LOG.debug("Failed to add status to cache", e);
```
gpang, line 226: add path to message

core/common/src/main/java/alluxio/util/LogUtils.java
```diff
@@ -127,7 +127,7 @@ public static void warnWithException(Logger logger, String message, Object ...ar
       logger.debug(message, args);
     } else {
       if (args.length > 0 && args[args.length - 1] instanceof Throwable) {
-        args[args.length - 1] = ((Throwable) args[args.length - 1]).getMessage();
+        args[args.length - 1] = (args[args.length - 1]).toString();
```
gpang, line 130: Won't this print out the entire stack trace of the exception?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -726,45 +744,53 @@ public FileInfo getFileInfo(AlluxioURI path, GetStatusContext context)
...
-      } catch (AccessControlException e) {
-        auditContext.setAllowed(false);
-        throw e;
-      }
-      // Possible ufs sync.
-      if (syncMetadata(rpcContext, inodePath, lockingScheme, DescendantType.ONE)) {
+        FileSystemMasterAuditContext auditContext =
+            createAuditContext("getFileInfo", path, null, null)) {
+
+      if (syncMetadata(rpcContext,
```
gpang, line 751: Is the general approach for RPCs to have a sync phase first, then the actual execution of the RPC? (Before it did the sync within the same contexts of the RPC execution).

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -815,73 +841,81 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
...
-      ensureFullPathAndUpdateCache(inodePath);
-      inode = inodePath.getInode();
-      auditContext.setSrcInode(inode);
-      DescendantType descendantTypeForListStatus =
-          (context.getOptions().getRecursive()) ? DescendantType.ALL : DescendantType.ONE;
-      listStatusInternal(context, rpcContext, inodePath, auditContext, descendantTypeForListStatus,
-          resultStream, 0);
-      auditContext.setSucceeded(true);
-      Metrics.FILE_INFOS_GOT.inc();
+        loadMetadataIfNotExist(rpcContext, inodePath, loadMetadataContext, false);
```
gpang, line 908: Can the new `syncMetadata` also take care of `loadMetadata`? I think the full sync is a superset of load metadata, so can't we just do the `sync`, which can optionally take care of loading metadata only if requested? I think load metadata only happens on getFileInfo and listStatus.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: > Well, if we keep calling prefetchChildren, it will never terminate
I don't understand what is "never terminating"?

calling this method only schedules the job and immediately returns. It may cancel a previous one if it exists. This method will never block.

we only ever call `prefetchChildren` once in the code, and it doesn't depend on whether or not there is a job in the cache for that path. For the code paths that use that job to retrieve the child statuses, they also don't have a strong dependency on the prefetch job existing or not. If the job they wait for gets cancelled (or doesn't exist) it will just make the call to the UFS serially.

core/common/src/main/java/alluxio/util/LogUtils.java
ZacBlanco: No, this just prints `<ExceptionClass>: <Exception message>`. Before we would lose the exception type.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: For All RPCs previously the sync phase typically came at the very beginning of the RPC after the initial permission check. This was the old way:

- create RPC, audit context, lock inode path
- check permissions
- perform sync
- perform RPC
- unlock

Because of the new locking requirements, there is a new paradigm of

- create RPC/audit context
- check if sync is required
  - check permissions
  - lock properly for sync
  - do sync
  - unlock from sync
- lock for RPC
- check permissions
- perform RPC
- unlock

The reason permission check is performed twice is because it requires having a locked inode path before doing the check.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: Unfortunately in my testing this breaks behavior and some assumptions about what metadata gets loaded during these RPCs. I did this because my goal around these improvements was to not change any assumptions about how we load metadata.

I don't remember exactly which tests broke, but basically loading is a strict load of metadata, whereas a sync could possibly delete or remove inodes, and so that broke a bunch of test assumptions. Breaking previous RPC behavior probably requires a larger discussion.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: 1. Yes
2. Yes. Do you see thread-safety issues?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: Yeah, I know load is different from sync. But maybe something like `load-only` could be an option to `sync`, so that when that option is used, it does not do the update/delete part of the sync?

This would only be for greater simplicity. I'm not sure if this would actually make it simpler, but I thought maybe it could.

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
gpang: In `fetchChildrenIfAbsent`, it is supposed to wait for the prefetch job. If the prefetch job is always getting canceled and re-created, I'd imagine the prefetch job would never finish, and that i where the "cycle" would happen.

But it sounds like you are saying `prefetchChildren` is only called once per path per instance of this class?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,282 @@
...
+    if (prefetchJob != null) {
+      try {
+        return prefetchJob.get();
+      } catch (InterruptedException | ExecutionException e) {
+        LogUtils.warnWithException(LOG, "Failed to get result for prefetch job on alluxio path {}",
+            path, e);
+        if (e instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+        }
+        throw new IOException(e);
```
gpang, line 167: If the prefetch job is canceled, this will throw an error, right? Does something external retry this?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3235,28 +3096,33 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
...
         // full sync
-        LockingScheme lockingScheme = new LockingScheme(path, LockPattern.READ, true);
-        try (LockedInodePath inodePath =
-            mInodeTree.lockInodePath(lockingScheme.getPath(), lockingScheme.getPattern())) {
-          syncMetadataInternal(rpcContext, inodePath, lockingScheme, DescendantType.ALL,
-              populateStatusCache(path, DescendantType.ALL));
+        long start = System.currentTimeMillis();
+
+        // Set sync interval to 0 to force a sync.
+        FileSystemMasterCommonPOptions options =
```
gpang, line 3102: Should we make the force sync option a static constant, and reuse it here and below?

gpang:
Jenkins, test this please

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: Yes, it should only ever happen once per path in the current code

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: There is currently no retry.

I think that we should only throw the `IOException` if it is an InterruptedException, because that's something signalling the thread to stop. Otherwise if it is an ExecutionException or CancelledException then we should just continue with the fallback operation?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: That's actually what it already does....go to the definition of loadMetadataIfNotExist :)

gpang:
Jenkins, test this please

core/common/src/main/java/alluxio/conf/PropertyKey.java
yuzhu: could we have some of these comments in the source code?  helps the next reader

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
yuzhu: in this case, are we ok with dangling cache entries? for a path /a/b/c, if /a is removed , /a/b -> {/a/b/c} is still in the cache, so is /a/b, do you think we should do cascading removal?

yuzhu: let's say /a is removed and added later, it will suddenly get many entries from the past

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: Yeah, I know `loadMetadata` call reuses sync. However, do you think it is possible for everything to be taken care of with the `syncMetadata()` at the beginning of this method? Basically, for `listStatus` and `fileInfo` they both have a concept of loading metadata, which is just the creating part of sync. I wonder if everything could be taken care of in the `syncMetadata()` call, so we don't even need the `loadMetadata()` method anymore.

(This would not be part of this PR)

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
gpang: @ZacBlanco Is there any relationship between the listings cache, and the status cache? Meaning, if the listing cache has the mapping `/parent/ -> [/parent/a, /parent/b, /parent/c]`, is the status cache guaranteed to have the status of `/parent/a`, `/parent/b`, and `/parent/c`?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -0,0 +1,282 @@
...
+
+/**
+ * This class is a cache from an Alluxio namespace URI ({@link AlluxioURI}, i.e. /path/to/inode) to
+ * UFS statuses.
+ *
+ * It also allows associating a path with child inodes, so that the statuses for a specific path can
+ * be searched for later.
+ */
+@ThreadSafe
+public class UfsStatusCache {
```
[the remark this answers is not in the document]
ZacBlanco: I am not sure how much it would help. Do you know what the benefits are?

Much of the implementation is specialized for the usage of `InodeSyncStream`

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: @yuzhu  We don't want to do a cascading removal because it would be removing statuses from the cache that are potentially used in a later part of the sync.

The implementation is designed for the inode sync stream so that the statuses can be re-used later without needing to do more UFS listings. Additionally, the way this is used is that the parent is **always** synced and removed _before_ a child status even begins to sync.

We remove only the information that was needed for the parent to sync, while leaving all information about children in-tact so they don't need to call to the UFS when they are synced. Then the children then add information about their children to the cache, while removing theirs once finished, etc...

@gpang When initially added, they are guaranteed to be in sync, i.e. if `/parent -> {/parent/a, /parent/b, /parent/c}` exists then `/parent/a, /parent/b, /parent/c` will all exist in the cache individually. However, it's possible they _could_ be removed before the parent listing.

In the current code, the parent is **always** synced and removed before the child even begins to sync because we don't add children to the processing queue until the parent has completed all of its syncing.

ZacBlanco: Also, I do see the issue with children still existing even if the parent doesn't.

There are not cases currently where that could happen as far as I'm aware in the current usage, but I meant to push an update for this when I changed the internal representation of the maps. Guess I missed it.

ZacBlanco:

PR #11218, opened 2020-03-30
Create new client pool map during context reinit

### LINKED ISSUE #11221: Leased block worker clients can be released into different pools
**Alluxio Version:**

2.2.0

**Describe the bug**

`BlockWorkerClient`s are acquired and released from the FileSystemContext class. However, the pools that the clients are acquired from may be changed in between leases due to updates to cluster or path configuration changes.

Block worker clients are acquired slightly different from the master clients because it uses a map and creates a block worker client pool for each worker that a client connects to. Currently the way that the pools are reset is that we simply clear the map instead of creating an entirely new one. The release method for block worker clients will then always utilize the same map, rather than saving the reference to the map that they were created with.

The map needs to be a completely new reference when the FileSystemContext is re-initialized.

**To Reproduce**

- A thread would acquire a block worker client from the pool map with key A.
- The context is re-initialized. All pools in the map are closed, the pool map is cleared
- Another thread acquires a block worker client with key A as well. This creates a new pool map entry with the same key A.
- The original thread tries to release the client it acquired before, and it releases back into the new blockworker client pool. The pool is not the same as before, and an exception is thrown.

**Expected behavior**

The clients should always get released back into their original pool. No Exceptions should be thrown.

**Urgency**

Medium. Should only occur when users change path or cluster config

ZacBlanco:
When a FileSystemContext is re-initialized, any acquired clients are expected to be released back into the same pool.

Before this change, it's possible for the following scenario to occur:
- A thread would acquire a block worker client from the pool map with key `A`.
- The context is re-initialized. All pools in the map are closed, the pool map is cleared
- Another thread acquires a block worker client with key `A` as well. This creates a new pool map entry with the same key `A`.
- The original thread tries to release the client it acquired before, and it releases back into the new blockworker client pool. The pool is not the same as before, and an exception is thrown.

Fixes #11221

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
@@ -92,8 +92,7 @@ public final class ConfigHashSync implements HeartbeatExecutor {
         mContext.getClientContext().getPathConfHash());
     if (isClusterConfUpdated || isPathConfUpdated) {
       try {
-        LOG.debug("Reinitializing FileSystemContext: Cluster conf updated: {}, path conf updated:"
-            + " {}", isClusterConfUpdated, isPathConfUpdated);
+
         mContext.reinit(isClusterConfUpdated, isPathConfUpdated);
         mException = null;
       } catch (UnavailableException e) {
--- a/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
@@ -92,7 +92,6 @@ public final class ConfigHashSync implements HeartbeatExecutor {
         mContext.getClientContext().getPathConfHash());
     if (isClusterConfUpdated || isPathConfUpdated) {
       try {
-
         mContext.reinit(isClusterConfUpdated, isPathConfUpdated);
         mException = null;
       } catch (UnavailableException e) {
--- a/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
@@ -92,6 +92,8 @@ public final class ConfigHashSync implements HeartbeatExecutor {
         mContext.getClientContext().getPathConfHash());
     if (isClusterConfUpdated || isPathConfUpdated) {
       try {
+        LOG.debug("Reinitializing FileSystemContext: Cluster conf updated: {}, path conf updated:"
+            + " {}", isClusterConfUpdated, isPathConfUpdated);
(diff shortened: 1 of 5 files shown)


ZacBlanco:
@ns1123 FYI

gpang:
@ZacBlanco Can you create and link a github issue for this?

ZacBlanco:
Added

ZacBlanco:
Here is a snippet of a very basic test which failed before this change:

```java
  @Test
  public void contextReinitAcquireBlockWorkerClientTest() throws Exception {
    try (CloseableResource<BlockWorkerClient> client =
        mFsContext.acquireBlockWorkerClient(sLocalAlluxioClusterResource.get().getWorkerAddress())) {
      mFsContext.reinit(true, true);
      mFsContext.acquireBlockWorkerClient(sLocalAlluxioClusterResource.get().getWorkerAddress())
          .close();
    }
```

core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
```diff
@@ -92,6 +92,7 @@ public synchronized void heartbeat() {
         mContext.getClientContext().getPathConfHash());
     if (isClusterConfUpdated || isPathConfUpdated) {
       try {
+
```
gpang: remove empty line

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -282,13 +283,13 @@ private synchronized void closeContext() throws IOException {
...
       mBlockMasterClientPool.close();
       mBlockMasterClientPool = null;
-      for (BlockWorkerClientPool pool : mBlockWorkerClientPool.values()) {
+      for (BlockWorkerClientPool pool : mBlockWorkerClientPoolMap.values()) {
         pool.close();
       }
       // Close worker group after block master clients in order to allow
       // clean termination for open streams.
       mWorkerGroup.shutdownGracefully(1L, 10L, TimeUnit.SECONDS);
-      mBlockWorkerClientPool.clear();
```
gpang: Why don't we want to clear anymore?

core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
ZacBlanco: done

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
ZacBlanco: accidentally removed. Good catch

ZacBlanco:

ZacBlanco:

PR #11226, opened 2020-04-01
Get service version out of retryRpc loop

### LINKED ISSUE #11228: Abstract Client infinite loop
**Alluxio Version:**
master

**Describe the bug**
In a condition where a channel creation succeeds but getRemoteServiceVersion() always fails;

`retryRpcInternal` will call `disconnect` after the RPC fails, then `connect` again on the next attempt.  `connect` will call `getRemoteServiceVersion()` again in the call chain because disconnect was called.

ggezer:
`getServiceVersion()` is a simple unauthenticated code, that is called after a successful connection.
Calling it with `retryRpc` utility opens a way for some allowed exceptions to fuel an infinite loop.

Fixes #11228

### DIFF
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -123,6 +123,7 @@ public abstract class AbstractClient implements Client {
   protected abstract ServiceType getRemoteServiceType();
 
   protected long getRemoteServiceVersion() throws AlluxioStatusException {
+    // Calling directly as this method is subject to an encompassing retry loop.
     return mVersionService
         .getServiceVersion(
             GetServiceVersionPRequest.newBuilder().setServiceType(getRemoteServiceType()).build())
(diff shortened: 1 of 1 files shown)


ggezer:
jenkins, test this please

core/common/src/main/java/alluxio/AbstractClient.java
```diff
@@ -123,11 +123,10 @@ public AbstractClient(ClientContext context, InetSocketAddress address,
   protected abstract ServiceType getRemoteServiceType();
 
   protected long getRemoteServiceVersion() throws AlluxioStatusException {
-    return retryRPC(() ->
```
gpang, line 126: Are you saying there is a retry of `getRemoteServiceVersion` outside of this, so we don't need to use `retryRPC`? I don't understand where the infinite loop is happening.

ggezer: When a channel creation succeeds but `getRemoteServiceVersion()` fails always.
In that case, `retryRpcInternal` will call disconnect after the rpc, then connect again. After disconnect, connect will call `getRemoteServiceVersion()` again.

ggezer: This will create connect -> getRemoteServiceVersion -> connect -> getRemoteServiceVersion loop. `connect` after `getRemoveServiceVersion` being the second attempts.

gpang: I see. Then, could you add a comment here saying we don't need to retry here, since the the entire connection/channel will be retried outside?

ggezer:

ggezer:

PR #11234, opened 2020-04-02
Update alluxio-mount.sh for FreeBSD

martenlindblad:
With the changes above, we can use alluxio with FreeBSD.

### DIFF
--- a/bin/alluxio-mount.sh
+++ b/bin/alluxio-mount.sh
@@ -88,7 +88,7 @@ function umount_ramfs_linux() {
 }
 
 function check_space_freebsd() {
-  local total_mem=$(($(sysctl -h hw.usermem | awk 'NR==1{print $2}' | sed 's/,//g')))
+  local total_mem=$(sysctl -n hw.usermem)
   if [[ ${total_mem} -lt ${MEM_SIZE} ]]; then
     echo "ERROR: Memory(${total_mem}) is less than requested ramdisk size(${MEM_SIZE}). Please
     reduce alluxio.worker.memory.size in alluxio-site.properties" >&2
--- a/bin/alluxio-mount.sh
+++ b/bin/alluxio-mount.sh
@@ -87,6 +87,65 @@ function umount_ramfs_linux() {
   fi
 }
 
+function check_space_freebsd() {
+  local total_mem=$(($(sysctl -h hw.usermem | awk 'NR==1{print $2}' | sed 's/,//g')))
+  if [[ ${total_mem} -lt ${MEM_SIZE} ]]; then
+    echo "ERROR: Memory(${total_mem}) is less than requested ramdisk size(${MEM_SIZE}). Please
+    reduce alluxio.worker.memory.size in alluxio-site.properties" >&2
+    exit 1
+  fi
+}
+
+function mount_ramfs_freebsd() {
+  TIER_PATH=${1}
+  echo "Formatting RamFS: ${TIER_PATH} (${MEM_SIZE})"
+  if [[ ${USE_SUDO} == true ]]; then
+    sudo mkdir -p ${TIER_PATH}
+  else
+    mkdir -p ${TIER_PATH}
+  fi
+  if [[ $? -ne 0 ]]; then
+    echo "ERROR: mkdir ${TIER_PATH} failed" >&2
+    exit 1
+  fi
+
+  if [[ ${USE_SUDO} == true ]]; then
+    sudo mount -t tmpfs -o size=${MEM_SIZE} tmpfs ${TIER_PATH}
+  else
+    mount -t tmpfs -o size=${MEM_SIZE} tmpfs ${TIER_PATH}
+  fi
+  if [[ $? -ne 0 ]]; then
+    echo "ERROR: mount RamFS ${TIER_PATH} failed" >&2
+    exit 1
+  fi
+
+  if [[ ${USE_SUDO} == true ]]; then
+    sudo chmod a+w ${TIER_PATH}
+  else
+    chmod a+w ${TIER_PATH}
+  fi
+  if [[ $? -ne 0 ]]; then
+    echo "ERROR: chmod RamFS ${TIER_PATH} failed" >&2
+    exit 1
+  fi
+}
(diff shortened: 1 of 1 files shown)


haoyuan:
Hey @martenlindblad. Thanks for the PR! Are you using Alluxio with BSD? Cheers!

cc @apc999

martenlindblad:
No problem! Yes we are using it with FreeBSD 12.1 without any serious issues as a layer between S3, Hadoop for Hive, Flink and some other stuff. All running FreeBSD!

ZacBlanco:
Hi @martenlindblad . These changes look good for the most part. Can you give a short summary on what the differences were in order to make the script compatible with FreeBSD?

bin/alluxio-mount.sh
```diff
@@ -87,6 +87,65 @@ function umount_ramfs_linux() {
...
+  else
+    mkdir -p ${TIER_PATH}
+  fi
+  if [[ $? -ne 0 ]]; then
+    echo "ERROR: mkdir ${TIER_PATH} failed" >&2
+    exit 1
+  fi
+
+  if [[ ${USE_SUDO} == true ]]; then
+    sudo mount -t tmpfs -o size=${MEM_SIZE} tmpfs ${TIER_PATH}
```
ZacBlanco, line 113: is `tmpfs` vs `ramfs` the only difference between this and the linux implementation?

ZacBlanco: Additionally, I came across this link in the freebsd handbook which suggests a different approach: https://www.freebsd.org/doc/handbook/disks-virtual.html

> To instead create a new memory-based memory disk with mdmfs, use this one command:
```
# mdmfs -s 5m md1 /mnt
```

Here is the man page on `mdfs`: www.freebsd.org/cgi/man.cgi?query=mdmfs&sektion=8&manpath=freebsd-release-ports

Have you considered this option?

martenlindblad: > Additionally, I came across this link in the freebsd handbook which suggests a different approach: https://www.freebsd.org/doc/handbook/disks-virtual.html
> 
> > To instead create a new memory-based memory disk with mdmfs, use this one command:
> 
> ```
> # mdmfs -s 5m md1 /mnt
> ```
> 
> Here is the man page on `mdfs`: [www.freebsd.org/cgi/man.cgi?query=mdmfs&sektion=8&manpath=freebsd-release-ports](http://www.freebsd.org/cgi/man.cgi?query=mdmfs&sektion=8&manpath=freebsd-release-ports)
> 
> Have you considered this option?

Yes I did but I wanted the code to be as similar to the linux version as possible. The mount -t tmpfs does basically the same thing based on the alluxio usage, see https://www.freebsd.org/cgi/man.cgi?query=mdmfs&sektion=8&manpath=freebsd-release-ports:

```
     When md-device is `auto', mdmfs uses tmpfs(5) if it is present in the
     kernel or can be loaded as	a module, otherwise it falls back to using
     md(4) auto-unit as	if `md'	had been specified.

     When md-device is `tmpfs',	mdmfs mounts a tmpfs(5)	filesystem, translat-
     ing the -s	size option, if	present, into a	`-o size=' mount option.  Any
     -o	options	on the command line are	passed through to the tmpfs(5) mount.
     Options specific to mdconfig(8) or	newfs(8) are ignored.
```

> is `tmpfs` vs `ramfs` the only difference between this and the linux implementation?

The other difference is that this implementation also uses sysctl instead of /proc/meminfo to avoid requiring mounting procfs.

martenlindblad:
> Hi @martenlindblad . These changes look good for the most part. Can you give a short summary on what the differences were in order to make the script compatible with FreeBSD?

Hi! Thanks for reviewing this. The only differences is that FreeBSD use different methods to get available memory and another filesystem type to mount a memory filesystem.

bin/alluxio-mount.sh
```diff
@@ -87,6 +87,65 @@ function umount_ramfs_linux() {
   fi
 }
 
+function check_space_freebsd() {
+  local total_mem=$(($(sysctl -h hw.usermem | awk 'NR==1{print $2}' | sed 's/,//g')))
```
ZacBlanco, line 91: Just trying to understand this command

> `sysctl -h hw.usermem`

prints out the system memory in human-readable format because of the `-h` flag

> `awk 'NR==1{print $2}`

This will print out the 2nd column from the first line, where the column is defined by a <space> char. I'm assuming the 2nd column is the value for the `hw.usermem`

> `sed 's/,//g'`

Removes an commas from the resulting output.

My only question with this set of commands is the `-h` flag. It will print in human readable format, but does that mean it will convert bytes to kb, mb, gb, etc dynamically based on the actual value?

In our linux version we use `/proc/meminfo` which always prints in kb, and then [multiply that value by 1024](https://github.com/Alluxio/alluxio/blob/a98d4f9bef7a12c2f89f1011101d4cbdbc72ecd8/bin/alluxio-mount.sh#L32). If we leave off the `-h` will it just give us the raw value in bytes? I think that is what we want.

(sorry I don't have a freebsd system available to test this myself)

bin/alluxio-mount.sh
ZacBlanco: Thanks for the clarification!

bin/alluxio-mount.sh
martenlindblad: You're right, this was really sloppy from me. The correct way of doing it is `sysctl -n hw.usermem`. That way we can omit the awk and sed commands. I will update the PR.

ZacBlanco:
@martenlindblad I just want to confirm that you were able to test the different paths (mount/SudoMount) on your FreeBSD system with the latest changes and that everything worked.

Once you confirm we can get this merged. Thanks!

martenlindblad:
> @martenlindblad I just want to confirm that you were able to test the different paths (mount/SudoMount) on your FreeBSD system with the latest changes and that everything worked.
> 
> Once you confirm we can get this merged. Thanks!

Verified it working on FreeBSD 12.1! Thanks

ZacBlanco:

PR #11235, opened 2020-04-02
Get service version out of retryRpc loop

### LINKED ISSUE #11228: Abstract Client infinite loop

ggezer:
`getServiceVersion()` is a simple unauthenticated code, that is called
after a successful connection.
Calling it with `retryRpc` utility opens a way for some allowed
exceptions to fuel an infinite loop.

Fixes #11228

pr-link: Alluxio/alluxio#11226
change-id: cid-3c38f3ed0d8f7f033baf565e927e265b24cd18d5

### DIFF
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -117,13 +117,11 @@ public abstract class AbstractClient implements Client {
   protected abstract ServiceType getRemoteServiceType();
 
   protected long getRemoteServiceVersion() throws AlluxioStatusException {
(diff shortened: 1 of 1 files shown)


ggezer:
@gpang PTAL.

ggezer:
jenkins, test this please

ggezer:

PR #11246, opened 2020-04-07
Move job configs back to the old location

bradyoo:
I previously moved classes like

alluxio.job.migrate.MigrateConfig to alluxio.job.plan.migrate.MigrateConfig

but that broke version compatibility when there is a version mismatch between client and server.


bradyoo:
@LuQQiu here is the fix for 2.2 job service client vs 2.1 job-service server.

LuQQiu:
I am a little worried that if we merge this change to master/2.2.3, then version 2.2.3 server doesn't work with 2.2.0 and 2.2.1 client. Thanks for having those changes ready.

bradyoo:
@LuQQiu: I reproduced the issue and confirmed that this does fix the issue between 2.1 <-> 2.2. I can make some additional changes to not break within 2.2.x by instead of doing a move, by doing a copy and supporting both (at the cost of code being a mess). Let me know if you think we should do it.

LuQQiu:
@bradyoo Thanks Bradley for verifying the issue. I don't think it's a big problem for breaking compatibility between 2.2 clients and 2.1 servers. The compatibility between minor versions may be more important. @ns1123 What's your opinion?

ns1123:
I agree. Client server version match is something users should be testing on their own to fix an issue before reporting.

PR #11250, opened 2020-04-08
Fix single master startup time optimization

ZacBlanco:


ZacBlanco:
@ggezer this looks like a regression that started at some point a while back. It's not a huge issue, so I don't think we need to backport the change anywhere. But I'm open to moving this to 2.1/2.2

ZacBlanco:

PR #11254, opened 2020-04-09
Fix Alluxio deploy with docker doc bug

JySongWithZhangCe:
Workers' capacity will not take effect unless add alluxio.worker.memory.size both in Masters' start command.


apc999:
@JySongWithZhangCe thanks for the PR!
@madanadit  can you verify if this fix is needed? I am curious why it necessary to also set worker memory size on master side?

docs/en/deploy/Running-Alluxio-On-Docker.md
```diff
@@ -159,7 +159,7 @@ $ docker run -d  --rm \
     -p 19998:19998 \
     --net=alluxio_network \
     --name=alluxio-master \
-    -e ALLUXIO_JAVA_OPTS="-Dalluxio.master.hostname=alluxio-master -Dalluxio.master.mount.table.root.ufs=/opt/alluxio/underFSStorage" \
+    -e ALLUXIO_JAVA_OPTS="-Dalluxio.master.hostname=alluxio-master -Dalluxio.master.mount.table.root.ufs=/opt/alluxio/underFSStorage -Dalluxio.worker.memory.size=1G" \
```
ZacBlanco, line 162: This is the master container. Why are we adding this to the master? Setting the worker memory size on the master is not required

JySongWithZhangCe: Actually it is necessary because Alluxio Workers'  ram capacity will be 2/3 of total system memory  although the conf of ram capacity  in Alluxio Workers' side is 1GB checked by "./bin/alluxio getConf"

madanadit: i see you're saying that in the following line in `alluxo-mount.sh` in the worker containter will RPC the master to get the worker conf?
```
bin/alluxio-mount.sh:  MEM_SIZE=$(${BIN}/alluxio getConf --unit B alluxio.worker.memory.size)
```

if this is the case, then we need to fix `alluxio-mount.sh` to look at local conf instead of adding worker properties to the master

JySongWithZhangCe: start master & worker without conf in master 
image

JySongWithZhangCe: worker's ram capacity is 2/3 to total ram
image
image

JySongWithZhangCe: worker's conf is right but doesn't worked
image

JySongWithZhangCe: redeploy with the conf in master
image

JySongWithZhangCe: working as except
image
image

JySongWithZhangCe: working as except
image
image

apc999:
@JySongWithZhangCe this looks to me more like a bug on docker integration. thanks for finding and reporting this issue. I would suggest to fix the code rather than patching the doc. @madanadit , @ZacBlanco  what do you think?

madanadit:
I agree, we should fix the code instead

JySongWithZhangCe:
> @JySongWithZhangCe this looks to me more like a bug on docker integration. thanks for finding and reporting this issue. I would suggest to fix the code rather than patching the doc. @madanadit , @ZacBlanco what do you think?

I agree

PR #11255, opened 2020-04-09
Keep state lock for duration of journal context

### LINKED ISSUE #11259: Incorrect state change lock may cause inconsistent backups
**Alluxio Version:**
1.8 and master

**Describe the bug**
Backups may be inconsistent, while RPCs are in-progress when the backup is started.

**To Reproduce**
Have many RPCs in progress, and start the backup.

**Expected behavior**
The backup is taken with the exclusive lock on the state of the system, without in-progress RPCs running.

**Urgency**
Critical for consistent backups.

**Additional context**
This is caused by the state change lock not being held for the duration of the RPC.

gpang:
The state change lock should be held for the duration of the RPC, so that when the pause lock is held, there are no outstanding RPCs. This change makes the state change lock to be held during the whole journal context, so that when backups pause the state, it is guaranteed there are no state changes or journal changes in progress.

Fixes #11259


gpang:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/AbstractMaster.java
```diff
@@ -133,12 +132,9 @@ protected ExecutorService getExecutorService() {
...
   public JournalContext createJournalContext() throws UnavailableException {
-    // All modifications to journaled state must happen inside of a journal context so that we can
-    // persist the state change. As a mechanism to allow for state pauses, we acquire the state
-    // change lock before entering any code paths that could modify journaled state.
-    try (LockResource l = new LockResource(mMasterContext.stateChangeLock())) {
-      return mJournal.createJournalContext();
-    }
+    // Use the state change lock for the journal context, since all modifications to journaled
+    // state must happen inside of a journal context.
+    return mJournal.createJournalContext(mMasterContext.stateChangeLock());
```
ggezer, line 137: Can't we leave the Journal interfaces the same but wrap context here with a forwarder context that holds the lock. 
So Journal implementations wouldn't need to manage the lock.

gpang: I added a new journal context for the state change lock

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -655,11 +655,13 @@ public long getNewContainerId() throws UnavailableException {
    * @return a {@link JournalEntry} representing the state of the container id generator
    */
   private JournalEntry getContainerIdJournalEntry() {
-    BlockContainerIdGeneratorEntry blockContainerIdGenerator =
-        BlockContainerIdGeneratorEntry.newBuilder().setNextContainerId(mJournaledNextContainerId)
-            .build();
-    return JournalEntry.newBuilder().setBlockContainerIdGenerator(blockContainerIdGenerator)
-        .build();
+    synchronized (mBlockContainerIdGenerator) {
```
ggezer, line 658: why is this synchronized now?

gpang: O, I think this is strictly more correct, since `mJournaledNextContainerId` is protected by `mBlockContainerIdGenerator`.

gpang:

gpang:

gpang:

PR #11256, opened 2020-04-10
Replace Web UI linter in build command with maven exec plugin

tieujason330:
Removing this linter for now due to issues with formatting during tarball generation.

### DIFF
--- a/webui/pom.xml
+++ b/webui/pom.xml
@@ -127,6 +127,21 @@
               </arguments>
             </configuration>
           </execution>
+          <execution>
+            <id>run-linter</id>
+            <phase>test</phase>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>node/node</executable>
+              <arguments>
+                <argument>node_modules/.bin/lerna</argument>
+                <argument>run</argument>
+                <argument>eslint</argument>
+              </arguments>
+            </configuration>
+          </execution>
         </executions>
       </plugin>
       <plugin>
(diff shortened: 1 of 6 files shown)


ZacBlanco:
will we have anything to enforce code style after this change?

tieujason330:
no, is it possible to create a custom skippable phase in the webui's pom.xml?

ZacBlanco:
Sure. It is definitely possible. Try using the maven exec plugin

```xml
<plugin>
  <artifactId>exec-maven-plugin</artifactId>
  <groupId>org.codehaus.mojo</groupId>
  <executions>
    <execution><!-- Run our version calculation script -->
      <id>Version Calculation</id>
      <phase>generate-sources</phase>
      <goals>
        <goal>exec</goal>
      </goals>
      <configuration>
        <executable>${basedir}/scripts/calculate-version.sh</executable>
      </configuration>
    </execution>
  </executions>
</plugin>
```

([borrowed from here](https://stackoverflow.com/questions/3491937/i-want-to-execute-shell-commands-from-mavens-pom-xml))

You can stick this inside a maven profile: http://maven.apache.org/guides/introduction/introduction-to-profiles.html

webui/pom.xml
```diff
@@ -127,6 +127,21 @@
               </arguments>
             </configuration>
           </execution>
+          <execution>
+            <id>run-linter</id>
+            <phase>test</phase>
```
ZacBlanco, line 132: should this be a part of the test phase? Do you think it's more apt to put it in `checkstyle` or `verify`?

 Will `eslint` fix the issues? or just complain about them?

tieujason330: from what i tried, i couldn't put it in `checkstyle` because i need all the extra steps of installing node/npm and running `npm install` first to get lerna ready. i'm not familiar with maven so let me know if if i got it all wrong.

i moved it to `verify` phase.

`eslint` will complain about the issues. dev will need to fix them with running `npm run format` or manually fix.

ZacBlanco: makes sense, thanks!

tieujason330:

ZacBlanco:

PR #11257, opened 2020-04-10
Fix inheritance for empty owner on createPath and sync

madanadit:
This PR includes 2 changes:
- For S3, when `inherit_acl=false` any inodes created during loadmetadata should inherit owner/group from the parent directory in alluxio instead of being empty. This was broken because `setOwner` was called after `inheritOwnerAndGroupIfEmpty` in `InodeTree::createPath`
- When syncing metadata with s3, an `updateMetadata` sync plan could lead to setting the owner and group to `_` in `DefaultFileSystemMaster::setAttributeSingleFile`. The fix is to not set the options in `SetAttributePOptions` to avoid setting to an empty string or `_`.

This can be reproduced when a directory is created in alluxio only, the same directory is then later created in the ufs and then synced. Steps to reproduce:
```
$ aws s3 ls s3://adit-bucket-west/
                           PRE /
$ afs mkdir /adit
Successfully created directory /adit
$ afs mount /adit/s3a s3a://adit-bucket-west/
Mounted s3a://adit-bucket-west/ at /adit/s3a
$ afs mkdir /adit/s3a/nested
Successfully created directory /adit/s3a/nested
$ aws s3 cp world s3://adit-bucket-west/nested/file
upload: ./world to s3://adit-bucket-west/nested/file
$ afs ls -Dalluxio.user.file.metadata.sync.interval=0 /adit/s3a/nested
-rwx------  _              _                            6       PERSISTED 04-09-2020 22:22:18:000   0% /adit/s3a/nested/file
$ afs ls /adit/s3a/
drwx------  _              _                            1       PERSISTED 04-09-2020 22:22:18:000  DIR /adit/s3a/nested
```

Fixes: https://github.com/Alluxio/alluxio/issues/11265, https://github.com/Alluxio/alluxio/issues/11266

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -3564,10 +3564,20 @@ public final class DefaultFileSystemMaster extends CoreMaster
         if (ufsFpParsed.isValid()) {
           short mode = Short.parseShort(ufsFpParsed.getTag(Tag.MODE));
           long opTimeMs = System.currentTimeMillis();
-          setAttributeSingleFile(rpcContext, inodePath, false, opTimeMs, SetAttributeContext
-              .mergeFrom(SetAttributePOptions.newBuilder().setOwner(ufsFpParsed.getTag(Tag.OWNER))
-                  .setGroup(ufsFpParsed.getTag(Tag.GROUP)).setMode(new Mode(mode).toProto()))
-              .setUfsFingerprint(ufsFingerprint));
+          SetAttributePOptions.Builder builder = SetAttributePOptions.newBuilder()
+              .setMode(new Mode(mode).toProto());
+          String owner = ufsFpParsed.getTag(Tag.OWNER);
+          if (!owner.equals(Fingerprint.UNDERSCORE)) {
+            // Only set owner if not empty
+            builder.setOwner(owner);
+          }
+          String group = ufsFpParsed.getTag(Tag.GROUP);
+          if (!group.equals(Fingerprint.UNDERSCORE)) {
+            // Only set group if not empty
+            builder.setGroup(group);
+          }
+          setAttributeSingleFile(rpcContext, inodePath, false, opTimeMs,
+              SetAttributeContext.mergeFrom(builder).setUfsFingerprint(ufsFingerprint));
         }
       }
       if (syncPlan.toDelete()) {
--- a/core/common/src/main/java/alluxio/underfs/Fingerprint.java
+++ b/core/common/src/main/java/alluxio/underfs/Fingerprint.java
@@ -45,7 +45,7 @@ public final class Fingerprint {
 
   private static final Pattern SANITIZE_REGEX = Pattern.compile("[" + KVDELIMTER
       + TAGDELIMTER + "]");
-  private static final String UNDERSCORE = "_";
+  public static final String UNDERSCORE = "_";
 
   private final Map<Tag, String> mValues;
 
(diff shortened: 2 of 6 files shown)


madanadit:
@ns1123 fyi

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3564,10 +3564,20 @@ private SyncResult syncInodeMetadata(RpcContext rpcContext, LockedInodePath inod
...
           short mode = Short.parseShort(ufsFpParsed.getTag(Tag.MODE));
           long opTimeMs = System.currentTimeMillis();
-          setAttributeSingleFile(rpcContext, inodePath, false, opTimeMs, SetAttributeContext
-              .mergeFrom(SetAttributePOptions.newBuilder().setOwner(ufsFpParsed.getTag(Tag.OWNER))
-                  .setGroup(ufsFpParsed.getTag(Tag.GROUP)).setMode(new Mode(mode).toProto()))
-              .setUfsFingerprint(ufsFingerprint));
+          SetAttributePOptions.Builder builder = SetAttributePOptions.newBuilder()
+              .setMode(new Mode(mode).toProto());
+          String owner = ufsFpParsed.getTag(Tag.OWNER);
+          if (!owner.equals(Fingerprint.UNDERSCORE)) {
```
yuzhu, line 3570: I am curious, under what conditions, would the ufs not have owner or group information?

madanadit: its's a corner case, took me a while to figure out. i'll update the description

madanadit: @yuzhu when inherit_acl is false all object ufs do not return owner and group. that is because the s3 account credential cannot be mapped to a user in those cases.

yuzhu: got it.

gpang:
@madanadit can you create an issue describing the bug and link it from this PR?

madanadit:
@gpang yeah, i'm working on the unit test already, will create bug(s)

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3564,10 +3564,20 @@ private SyncResult syncInodeMetadata(RpcContext rpcContext, LockedInodePath inod
...
+            // Only set owner if not empty
+            builder.setOwner(owner);
+          }
+          String group = ufsFpParsed.getTag(Tag.GROUP);
+          if (!group.equals(Fingerprint.UNDERSCORE)) {
+            // Only set group if not empty
+            builder.setGroup(group);
+          }
+          setAttributeSingleFile(rpcContext, inodePath, false, opTimeMs,
```
gpang, line 3579: Does `setAttributeSingleFile` handle the cases when owner or group are unset in the proto?

madanadit: yep, it does

madanadit:
@gpang I found another bug while writing the test case. In some cases, we could have set owner / group back to empty before. For example, depending on the code path if dir was created in alluxio (did not exist in the ufs yet), and later was created in  the ufs out of band, sync on the dir would try to persist and inherit owner/ group from the ufs again (which could be empty with s3).

madanadit:
@gpang PTAL

gpang:
@madanadit I think this looks good, but could you create issues and link them from this PR?

madanadit:

madanadit:

Issue #11265, opened 2020-04-10
S3 objects do not inherit of owner from Alluxio parent dir


madanadit:
**Alluxio Version:**
2.2

**Describe the bug**
For S3, when inherit_acl=false any inodes created during loadmetadata do not inherit owner/group from the parent directory in alluxio instead of being empty.

**To Reproduce**
- Set `alluxio.underfs.s3.inherit.acl=false`. S3 ufs returns empty owner / group in this case. 
- Create objects in s3 directly (e.g. using aws cli)
- Create a dir in alluxio `/mnt`
- Mount the s3 bucket as a nested mount at `/mnt/s3`
- Newly created objects under `/mnt/s3` do not inherit owner/group from the parent directory `/mnt`

**Expected behavior**
- Newly created objects under `/mnt/s3` should inherit owner/group from the parent directory `/mnt` if `alluxio.master.metastore.inode.inherit.owner.and.group=true`

**Urgency**
Urgent

**Additional context**
N/A

Issue #11266, opened 2020-04-10
Sync on S3 can set the owner to "" or "_"


madanadit:
**Alluxio Version:**
2.2

**Describe the bug**
S3 does not set owner/group for objects when `inherit_acl=false`. When syncing metadata with s3, an `updateMetadata` sync plan could lead to setting the owner and group to `_` in `DefaultFileSystemMaster::setAttributeSingleFile` or to "" (empty) when updating inode entry during sync

**To Reproduce**
This can be reproduced when a directory is created in alluxio only, the same directory is then later created in the ufs and then synced. 

Steps to reproduce:
```
$ aws s3 ls s3://adit-bucket-west/
                           PRE /
$ afs mkdir /adit
Successfully created directory /adit
$ afs mount /adit/s3a s3a://adit-bucket-west/
Mounted s3a://adit-bucket-west/ at /adit/s3a
$ afs mkdir /adit/s3a/nested
Successfully created directory /adit/s3a/nested
$ aws s3 cp world s3://adit-bucket-west/nested/file
upload: ./world to s3://adit-bucket-west/nested/file
$ afs ls -Dalluxio.user.file.metadata.sync.interval=0 /adit/s3a/nested
-rwx------  _              _                            6       PERSISTED 04-09-2020 22:22:18:000   0% /adit/s3a/nested/file
$ afs ls /adit/s3a/
drwx------  _              _                            1       PERSISTED 04-09-2020 22:22:18:000  DIR /adit/s3a/nested
```

**Expected behavior**
Owner and group inherited from parent dir previously should not be modified.

**Urgency**
Urgent

**Additional context**
N/A

ZacBlanco:
This has been closed by #11257

PR #11268, opened 2020-04-11
Port changes of [#11257] to branch-1.8

madanadit:
This PR includes 2 changes:

For S3, when inherit_acl=false any inodes created during loadmetadata should inherit owner/group from the parent directory in alluxio instead of being empty. This was broken because setOwner was called after inheritOwnerAndGroupIfEmpty in InodeTree::createPath
When syncing metadata with s3, an updateMetadata sync plan could lead to setting the owner and group to _ in DefaultFileSystemMaster::setAttributeSingleFile. The fix is to not set the options in SetAttributePOptions to avoid setting to an empty string or _.
This can be reproduced when a directory is created in alluxio only, the same directory is then later created in the ufs and then synced. Steps to reproduce:

$ aws s3 ls s3://adit-bucket-west/
                           PRE /
$ afs mkdir /adit
Successfully created directory /adit
$ afs mount /adit/s3a s3a://adit-bucket-west/
Mounted s3a://adit-bucket-west/ at /adit/s3a
$ afs mkdir /adit/s3a/nested
Successfully created directory /adit/s3a/nested
$ aws s3 cp world s3://adit-bucket-west/nested/file
upload: ./world to s3://adit-bucket-west/nested/file
$ afs ls -Dalluxio.user.file.metadata.sync.interval=0 /adit/s3a/nested
-rwx------  _              _                            6       PERSISTED 04-09-2020 22:22:18:000   0% /adit/s3a/nested/file
$ afs ls /adit/s3a/
drwx------  _              _                            1       PERSISTED 04-09-2020 22:22:18:000  DIR /adit/s3a/nested
Fixes: #11265, #11266

[This is a manual PR to cherry-pick committed PR #11257 into target branch branch-2.8]

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/Fingerprint.java
+++ b/core/common/src/main/java/alluxio/underfs/Fingerprint.java
@@ -38,7 +38,7 @@ public final class Fingerprint {
   private static final Tag[] CONTENT_TAGS = {Tag.TYPE, Tag.UFS, Tag.CONTENT_HASH};
 
   private static final Pattern SANITIZE_REGEX = Pattern.compile("[ :]");
-  private static final String UNDERSCORE = "_";
+  public static final String UNDERSCORE = "_";
 
   private final Map<Tag, String> mValues;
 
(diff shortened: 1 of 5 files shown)


madanadit:

PR #11279, opened 2020-04-12
Make config reinit executor static

madanadit:
Each FSContext instance creates a new single threaded executor pool for meta master config sync. For long running clients such as yarn RM, if the filesystem is not closed properly we would leak all non-static resources unless garbage collected. 

Fixes: https://github.com/Alluxio/alluxio/issues/11344

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerClientPool.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerClientPool.java
@@ -48,12 +48,14 @@ public final class BlockWorkerClientPool extends DynamicResourcePool<BlockWorker
    *
    * @param userState the parent userState
    * @param address address of the worker
+   * @param minCapacity the minimum capacity of the pool
    * @param maxCapacity the maximum capacity of the pool
    * @param alluxioConf Alluxio configuration
    */
-  public BlockWorkerClientPool(UserState userState, GrpcServerAddress address, int maxCapacity,
-      AlluxioConfiguration alluxioConf) {
-    super(Options.defaultOptions().setMaxCapacity(maxCapacity).setGcExecutor(GC_EXECUTOR));
+  public BlockWorkerClientPool(UserState userState, GrpcServerAddress address, int minCapacity,
+      int maxCapacity, AlluxioConfiguration alluxioConf) {
+    super(Options.defaultOptions().setMinCapacity(minCapacity).setMaxCapacity(maxCapacity)
+        .setGcExecutor(GC_EXECUTOR));
     mUserState = userState;
     mAddress = address;
     mConf = alluxioConf;
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
@@ -13,15 +13,17 @@ package alluxio.client.file;
 
 import alluxio.concurrent.CountingLatch;
 import alluxio.conf.PropertyKey;
-import alluxio.heartbeat.HeartbeatContext;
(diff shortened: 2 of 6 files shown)


core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
```diff
@@ -55,9 +57,7 @@
   public FileSystemContextReinitializer(FileSystemContext context) {
     mContext = context;
     mExecutor = new ConfigHashSync(context);
-    mExecutorService = Executors.newSingleThreadExecutor(ThreadFactoryUtils.build(
-        "config-hash-master-heartbeat-%d", true));
-    mExecutorService.submit(
+    REINIT_EXECUTOR.submit(
         new HeartbeatThread(HeartbeatContext.META_MASTER_CONFIG_HASH_SYNC, mContext.getId(),
```
ggezer, line 61: I suspect the first submission will hold the thread and rest will never have a chance to run.

madanadit: that makes sense, i'll change it to a scheduled fixed rate executor service and refactor the hearbeat executor

core/client/fs/src/main/java/alluxio/client/file/FileSystemContextReinitializer.java
```diff
@@ -55,14 +62,17 @@
...
     mExecutor = new ConfigHashSync(context);
-    mExecutorService = Executors.newSingleThreadExecutor(ThreadFactoryUtils.build(
-        "config-hash-master-heartbeat-%d", true));
-    mExecutorService.submit(
-        new HeartbeatThread(HeartbeatContext.META_MASTER_CONFIG_HASH_SYNC, mContext.getId(),
-            mExecutor, mContext.getClientContext().getClusterConf().getMs(
-                PropertyKey.USER_CONF_SYNC_INTERVAL),
-            mContext.getClientContext().getClusterConf(),
-            mContext.getClientContext().getUserState()));
+    REINIT_EXECUTOR.scheduleAtFixedRate(() -> {
```
ggezer, line 65: Can you store the Future<?> in order to cancel it during reinitializer shutdown?

ZacBlanco:
@madanadit is this PR still on the table?

madanadit:
@ZacBlanco yes

madanadit:

core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerClientPool.java
```diff
@@ -48,12 +48,14 @@
...
    * @param address address of the worker
+   * @param minCapacity the minimum capacity of the pool
    * @param maxCapacity the maximum capacity of the pool
    * @param alluxioConf Alluxio configuration
    */
-  public BlockWorkerClientPool(UserState userState, GrpcServerAddress address, int maxCapacity,
-      AlluxioConfiguration alluxioConf) {
-    super(Options.defaultOptions().setMaxCapacity(maxCapacity).setGcExecutor(GC_EXECUTOR));
+  public BlockWorkerClientPool(UserState userState, GrpcServerAddress address, int minCapacity,
+      int maxCapacity, AlluxioConfiguration alluxioConf) {
```
ggezer, line 56: Can't we just set minCapacity to 0 here rathat than defining a new prop?
I don't see a scenario where we'd set it anything else.

madanadit: @ggezer I updated the default to zero as like you said that is the common scenario and made it hidden. just to be consistent with other client thread pools, i didn't remove the property

madanadit:

madanadit:

PR #11280, opened 2020-04-13
Add flag to disable conf sync

madanadit:
Set `alluxio.user.conf.sync.enabled=false` to disable syncing of configuration with the master in a hearbeat thread. As a consequence, long running clients such as Presto, Hive or Yarn RM need to be restarted to pick up server side configuration updates.


ZacBlanco:
@madanadit is this PR still on the table?

madanadit:
@ZacBlanco nope, this is not needed

Issue #11271, opened 2020-04-11
Make Alluxio helm chart configurable for hostNetwork or not


cheyang:
**Alluxio Version:**
2.2.0

**Describe the bug**
The current implementation on Helm chart only supports non-hostnetwork environment: https://github.com/Alluxio/alluxio/commit/4d6126a5d9e03128c01ca73f2a8cad12558e981c

I suggest to make it configurable? the user can decide whether they want hostNetwork or non-hostNetwork. In fact, the network performance is much worse when using non hostNetwork. I

**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)

**Expected behavior**

**Urgency**
Describe the impact and urgency of the bug.

**Additional context**

jiacheliu3:
After https://github.com/Alluxio/alluxio/commit/4d6126a5d9e03128c01ca73f2a8cad12558e981c when ALLUXIO_WORKER_CONTAINTER_HOSTNAME is set on worker, the connections to worker will use this hostname even if the worker is using `hostNetwork`. I'll apply a small fix to remove ALLUXIO_WORKER_CONTAINTER_HOSTNAME when the workers are using `hostNetwork`.

jiacheliu3:
@cheyang please take a look if #11290 meets your need. Thanks!

cheyang:
@jiacheliu3 Thank you!

PR #11290, opened 2020-04-14
Add option to use hostNetwork in K8s worker pod

jiacheliu3:
This addresses https://github.com/Alluxio/alluxio/issues/11271

This change removes ALLUXIO_WORKER_CONTAINER_HOSTNAME when using `hostNetwork`. Under this setup the client side will talk to the worker via the node's network adapter, which is faster than resolving worker virtual IP.


jiacheliu3:
alluxio-bot, test this please

PR #11293, opened 2020-04-14
Improve performance of sasl authentication

ZacBlanco:
Before this PR, the ImpersonationAuthenticator class could effectively
bottleneck the amount of concurrent channels that a worker could handle. 

The `<init>` for the ImpersonationAuthenticator class required iterating over
all keys within an Alluxio configuration, not only doing key lookups, but also
performing regex matching. For some keys this could be especially expensive
(i.e, default supplier which calls Runtime.getRuntime().availableProcessors()).
On top of that, performing regex matching is also quite expensive for a few
hundred keys.

After taking a quick peek at the code it seems that it's entirely unnecessary to
instantiate the impersonation authenticator for every new handler that we
create. Instead, it can just be a field within the DefaultAuthenticationServer.
Then, it only ever needs to be instantiated once.

### DIFF
--- a/core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
+++ b/core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
@@ -62,6 +62,8 @@ public class DefaultAuthenticationServer
   /** Alluxio client configuration. */
   protected final AlluxioConfiguration mConfiguration;
 
+  private final ImpersonationAuthenticator mImpersonationAuthenticator;
+
   /**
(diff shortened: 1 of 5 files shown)


ZacBlanco:
jenkins, test this please

ZacBlanco:
alluxio-bot, test this please

ZacBlanco:
I swear this is one of the most unlucky PRs I've had in a while

ZacBlanco:
done

gpang:
Jenkins, test this please

ZacBlanco:

ZacBlanco:

ZacBlanco:

PR #11299, opened 2020-04-15
Improve performance of sasl authentication

ZacBlanco:
Before this PR, the ImpersonationAuthenticator class could effectively
bottleneck the amount of concurrent channels that a worker could handle.

The `<init>` for the ImpersonationAuthenticator class required iterating over
all keys within an Alluxio configuration, not only doing key lookups, but also
performing regex matching. For some keys this could be especially expensive
(i.e, default supplier which calls Runtime.getRuntime().availableProcessors()).
On top of that, performing regex matching is also quite expensive for afew
hundred keys.

After taking a quick peek at the code it seems that it's entirely unnecessary to
instantiate the impersonation authenticator for every new handler that we
create. Instead, it can just be a field within the DefaultAuthenticationServer.
Then, it only ever needs to be instantiated once.

pr-link: Alluxio/alluxio#11293
change-id: cid-96438acee91197e2e9a380494848db481c458660

### DIFF
--- a/core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
+++ b/core/common/src/main/java/alluxio/security/authentication/DefaultAuthenticationServer.java
@@ -64,6 +64,8 @@ public class DefaultAuthenticationServer
   /** Alluxio client configuration. */
   protected final AlluxioConfiguration mConfiguration;
 
+  private final ImpersonationAuthenticator mImpersonationAuthenticator;
+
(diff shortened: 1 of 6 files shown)


ZacBlanco:
The only code change for this cherry pick was a test fix, so I'm going to merge it. a2d72240d26dd39af9014f5990ba90cb7f09d419

ZacBlanco:

PR #11304, opened 2020-04-16
Optimize gRPC managed resource handling

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/11307


core/common/src/main/java/alluxio/grpc/GrpcConnection.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Objects;
+
+/**
+ * Used to gather gRPC level resources and indexes together.
+ */
+public class GrpcConnection implements AutoCloseable {
+
+  private GrpcConnectionKey mKey;
+  private ManagedChannel mManagedChannel;
+  private Channel mChannel;
```
calvinjia, line 30: These two are always the same?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
```diff
@@ -0,0 +1,378 @@
...
+ * Used to provide gRPC level connection management and pooling facilities.
+ *
+ * This class is used internally by {@link GrpcChannelBuilder} and {@link GrpcChannel}.
+ */
+@ThreadSafe
+public class GrpcConnectionPool {
+  private static final Logger LOG = LoggerFactory.getLogger(GrpcConnectionPool.class);
+
+  // Singleton instance.
+  private static GrpcConnectionPool sInstance;
```
calvinjia, line 50: Can this be static final and initialized in line?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
```diff
@@ -0,0 +1,378 @@
...
+            }
+          }
+          // Existing connection should shut-down.
+          if (shutdownExistingConnection) {
+            // TODO(ggezer): Implement GrpcConnectionListener for receiving notification.
+            existingRefCount = ref.getRefCount();
+            LOG.debug("Shutting down an existing unhealthy connection. "
+                + "ConnectionKey: {}. Ref-count: {}", key, existingRefCount);
+            // Shutdown the channel forcefully as it's already unhealthy.
+            shutdownManagedChannel(ref.get(), conf);
```
calvinjia, line 119: Instead of shutting down here, should we let other users of the channel eventually fail and only shutdown after the refcount is zero?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
```diff
@@ -0,0 +1,378 @@
...
+   *
+   * @param connectionKey the connection key
+   * @param conf the Alluxio configuration
+   */
+  public void releaseConnection(GrpcConnectionKey connectionKey, AlluxioConfiguration conf) {
+    mChannels.compute(connectionKey, (key, ref) -> {
+      Preconditions.checkNotNull(ref, "Cannot release nonexistent connection");
+      LOG.debug("Releasing connection for: {}. Ref-count: {}", key, ref.getRefCount());
+      // Shutdown managed channel.
+      if (ref.dereference() == 0) {
```
calvinjia, line 146: Is ref protected by the concurrency guarantees provided by the concurrenthashmap?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
```diff
@@ -0,0 +1,378 @@
...
+    } else {
+      channelBuilder = NettyChannelBuilder.forAddress(address);
+    }
+    // Apply default channel options for the multiplex group.
+    applyGroupDefaults(channelKey, channelBuilder, conf);
+    // Build netty managed channel.
+    return channelBuilder.build();
+  }
+
+  private void applyGroupDefaults(GrpcChannelKey key, NettyChannelBuilder channelBuilder,
```
calvinjia, line 197: Return the channel builder here?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
```diff
@@ -0,0 +1,378 @@
...
+      }
+    }
+    // Forceful shut down if still not terminated.
+    if (!managedChannel.isTerminated()) {
+      long timeoutMs = conf.getMs(PropertyKey.NETWORK_CONNECTION_SHUTDOWN_TIMEOUT);
+
+      managedChannel.shutdownNow();
+      try {
+        if (!managedChannel.awaitTermination(timeoutMs, TimeUnit.MILLISECONDS)) {
+          LOG.warn("Timed out forcefully shutting down connection: {}. ", managedChannel);
```
calvinjia, line 281: Is there any additional action we can take here?

core/common/src/main/java/alluxio/conf/RemovedKey.java
```diff
@@ -114,7 +114,7 @@ private static String replacedSince(String version, String newProperty) {
       put("alluxio.user.network.netty.reader.buffer.size.packets", removedSince(V2_0_0));
       put("alluxio.user.network.netty.reader.packet.size.bytes", removedSince(V2_0_0));
       put("alluxio.user.network.netty.timeout", replacedSince(V2_0_0,
-          PropertyKey.USER_NETWORK_DATA_TIMEOUT_MS.getName()));
+          PropertyKey.USER_STREAMING_DATA_TIMEOUT_MS.getName()));
```
madanadit, line 117: is this intended?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
ggezer: Correct.

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
```diff
@@ -106,7 +97,11 @@ public synchronized void shutdown() {
       mAuthDriver = null;
     }
     if (!mChannelReleased) {
-      GrpcManagedChannelPool.INSTANCE().releaseManagedChannel(mChannelKey, mShutdownTimeoutMs);
+      try {
+        mConnection.close();
+      } catch (Exception e) {
```
madanadit, line 102: why is this conversion necessary? maybe add a comment

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
ggezer: Not really.

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
ggezer: Though what would happen to the one that figured out the health issue?
- Assign a new group id and retry?
   - What if max connection is 1, so we keep retrying on the same channel?
- Block this one until other clients are failed released the connection?
  - This could never happen as they will be cached in pools and may not be used for a long time. 
    So this block could take at least as long as pool GC interval. (We can implement GrpcConnectionLister interface that is used to synchronously let them owners know)

core/common/src/main/java/alluxio/grpc/GrpcConnection.java
ggezer: `ManagedChannel` is the original netty side channel.
`Channel` is usually that channel wrapped with different interceptors. For example; each logical channel need to inject its own channel-id so they have their own interceptors. So we keep wrapping this channel until it has the final logical state we want.

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
ggezer: which conversion?

core/common/src/main/java/alluxio/conf/RemovedKey.java
ggezer: Should I make it `V2_2_3` ?

core/common/src/main/java/alluxio/grpc/GrpcChannelKey.java
```diff
@@ -11,66 +11,42 @@
...
-import java.util.Optional;
 import java.util.Random;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Used to identify a unique {@link GrpcChannel}.
  */
 public class GrpcChannelKey {
   private static final Random RANDOM = new Random();
```
madanadit, line 34: unused now

ggezer:
jenkins, test this please

core/common/src/main/java/alluxio/grpc/GrpcConnectionKey.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class GrpcConnectionKey {
+  private GrpcChannelKey mChannelKey;
+  private int mGroupIndex;
+
+  /**
+   * Creates a new gRPC connection key.
+   *
+   * @param channelKey the gRPC channel key
+   * @param groupIndex the group index
```
madanadit, line 27: nt: expand on the network group index usage

core/common/src/main/java/alluxio/conf/RemovedKey.java
madanadit: `V2_3_0`, this changes seems big enough to not go into a patch

core/common/src/main/java/alluxio/grpc/GrpcChannel.java
ggezer: It boils down to what do we want to do with failed release attempts. 
I believe It's good to have an exception with a stack-trace at least until stabilized.

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
calvinjia: Oh I see, we cannot replace the value in the map because the other users don't have a direct reference to the channel.

calvinjia:
@ggezer Thanks for the update, only concern I have is the property key changes. What kind of behavior will users using the old properties see after this patch?

ggezer:
@calvinjia 
If they defined connection-level properties, those would be left ineffective.
Additionally, their streaming connections would be bounded by default. This could mean perf-change for certain environments.

calvinjia:
Ideally if there is a logical translation of old properties to new properties, we should do that through alias (not sure if we support aliasing multiple properties to the same alias).
We should also deprecate the old properties and warn the user to use the new ones.
We want to avoid users seeing config change just because they upgraded a minor version.

core/common/src/main/java/alluxio/grpc/GrpcChannelBuilder.java
```diff
@@ -54,16 +46,10 @@ private GrpcChannelBuilder(GrpcServerAddress address, AlluxioConfiguration conf)
...
-    mShutdownTimeoutMs =
-        mConfiguration.getMs(PropertyKey.NETWORK_CONNECTION_SHUTDOWN_TIMEOUT);
-    mHealthCheckTimeoutMs =
-        mConfiguration.getMs(PropertyKey.NETWORK_CONNECTION_HEALTH_CHECK_TIMEOUT);
 
     // Set default overrides for the channel.
     mChannelKey = GrpcChannelKey.create(conf);
     mChannelKey.setServerAddress(address);
-    mChannelKey.setMaxInboundMessageSize(
-        (int) mConfiguration.getBytes(PropertyKey.USER_NETWORK_MAX_INBOUND_MESSAGE_SIZE));
```
ZacBlanco, line 66: Is there a reason we're not making this configurable anymore?

ggezer: It is configurable. Just not per channel but per connection. You can set different defaults to connections for streaming vs rpc.

ggezer:
@calvinjia @madanadit 
Updated. PTAL.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4873,8 +5052,24 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String USER_METRICS_HEARTBEAT_INTERVAL_MS =
         "alluxio.user.metrics.heartbeat.interval";
     public static final String USER_APP_ID = "alluxio.user.app.id";
-    public static final String USER_STREAMING_DATA_TIMEOUT_MS =
+    public static final String USER_NETWORK_DATA_TIMEOUT_MS =
```
calvinjia, line 5055: We can remove the ms suffix here.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4873,8 +5052,24 @@ private static String javadocLink(String fullyQualifiedClassname) {
...
         "alluxio.user.network.data.timeout";
+    public static final String USER_NETWORK_READER_BUFFER_SIZE_MESSAGES =
+        "alluxio.user.network.reader.buffer.size.messages";
+    public static final String USER_NETWORK_READER_CHUNK_SIZE_BYTES =
+        "alluxio.user.network.reader.chunk.size.bytes";
+    public static final String USER_NETWORK_WRITER_BUFFER_SIZE_MESSAGES =
+        "alluxio.user.network.writer.buffer.size.messages";
+    public static final String USER_NETWORK_WRITER_CHUNK_SIZE_BYTES =
+        "alluxio.user.network.writer.chunk.size.bytes";
+    public static final String USER_NETWORK_WRITER_CLOSE_TIMEOUT_MS =
```
calvinjia: We can remove the ms suffix in the variable name

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3463,62 +3468,230 @@ public String toString() {
...
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.CLIENT)
+          .build();
+  /**
+   * @deprecated use {@link #USER_STREAMING_WRITER_CLOSE_TIMEOUT_MS} instead
+   */
+  @Deprecated
+  public static final PropertyKey USER_NETWORK_WRITER_CLOSE_TIMEOUT_MS =
+      new Builder(Name.USER_NETWORK_WRITER_CLOSE_TIMEOUT_MS)
+          .setAlias("alluxio.user.network.writer.close.timeout.ms")
```
calvinjia, line 3552: Does this alias need to be propagated to the new property?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3463,62 +3468,230 @@ public String toString() {
...
           .build();
-  public static final PropertyKey USER_NETWORK_RPC_FLOWCONTROL_WINDOW =
-      new Builder(Name.USER_NETWORK_RPC_FLOWCONTROL_WINDOW)
+  /**
+   * @deprecated use {@link #USER_STREAMING_DATA_TIMEOUT} instead
+   */
+  @Deprecated
+  public static final PropertyKey USER_NETWORK_DATA_TIMEOUT_MS =
+      new Builder(Name.USER_NETWORK_DATA_TIMEOUT_MS)
+          .setAlias("alluxio.user.network.data.timeout.ms")
```
calvinjia, line 3491: Does this alias need to be propagated to the new property?

ggezer:

ggezer:

Issue #11309, opened 2020-04-17
AsyncBlockRemover stuck not freeing any blocks


jiacheliu3:
**Alluxio Version:**
2.1.1

**Describe the bug**
This issue is reported by @JySongWithZhangCe 

`alluxio fs free /` freed the root mount. One worker failed to free the space, as illustrated.

![image](https://user-images.githubusercontent.com/14806853/79546912-c7f73100-80c5-11ea-8c50-f9de1a793623.png)

In the worker.log there are a lot of below message
```
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 504054677506 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 502712500226 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 503652024322 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 504591548418 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 503584915458 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 505078087685 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 505010978821 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 504004345861 is being removed. Current queue size is 320.
2020-04-17 14:50:41,485 DEBUG AsyncBlockRemover - 502662168581 is being removed. Current queue size is 320.
```
AsycBlockRemover gets a lot of remove calls but remove is not making any progress. There are around 85K lines of above message, whereas the queue size stays 320.

Whenever a block is successfully removed by AsyncBlockRemover there should be one below line in the debug log. However compared to 85K above log msgs, there are like 2 blocks successfully removed.
```
Block XXX is removed in thread XXX.
```

I did a `jstack` on Alluxio worker and ALL remover threads are waiting on the block lock.
```
"block-removal-service-8" #1097 daemon prio=5 os_prio=0 tid=0x00007f8e0d5be800 nid=0x155c4d waiting on condition [0x00007f8a44ed1000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
[...]

PR #11310, opened 2020-04-17
Release resources properly within Worker ReadHandler

### LINKED ISSUE #11258: BlockWorker leaking locks after hitting pool limits
**Alluxio Version:**
2.2.+

**Describe the bug**
BlockWorker leaks locks when RPCs can't make progress due to the internal thread pool limit.
These leaked locks then hold reader executor threads alive.

**Expected behavior**
Internal resources of failed/timed-out RPCs must be properly disposed of.

**Urgency**
??

**Additional context**

```
2020-04-09 23:25:23,776 WARN  AbstractReadHandler - Exception occurred while processing read request block_id: 16777216000,........
alluxio.worker.grpc.GrpcExecutors$ImpersonateThreadPoolExecutor$$Lambda$107/1844138271@411da9ab **rejected from java.util.concurrent.ThreadPoolExecutor**@40399233[Running, pool size = 2048, active threads = 2048, queued tasks = 0, completed tasks = 63803]
```

Followed by stuck threads with below stack trace:

```
"BlockDataReaderExecutor-3263" #4125 daemon prio=5 os_prio=0 tid=0x00007f14840c6000 nid=0x2632 waiting on condition [0x00007f0cdbe42000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000006c1bda788> (a java.util.concurrent.Semaphore$NonfairSync)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireShared(AbstractQueuedSynchronizer.java:967)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(AbstractQueuedSynchronizer.java:1283)
	at java.util.concurrent.Semaphore.acquireUninterruptibly(Semaphore.java:494)
	at alluxio.worker.block.ClientRWLock$SessionLock.lock(ClientRWLock.java:92)
	at alluxio.worker.block.BlockLockManager.lockBlock(BlockLockManager.java:114)
	at alluxio.worker.block.TieredBlockStore.**lockBlock**(TieredBlockStore.java:149)
	at alluxio.worker.block.DefaultBlockWorker.moveBlock(DefaultBlockWorker.java:464)
	at alluxio.worker.grpc.BlockReadHandler$BlockDataReader.openBlock(BlockReadHandler.java:134)
	at alluxio.worker.grpc.BlockReadHandler$BlockDat0x00000006c1bda788aReader.getDataBuffer(BlockReadHandler.java:105)
	at alluxio.worker.grpc.BlockReadHandler$BlockDataReader.getDataBuffer(BlockReadHandler.java:70)
	at alluxio.worker.grpc.AbstractReadHandler$DataReader.runInternal(AbstractReadHandler.java:355)
	at alluxio.worker.grpc.AbstractReadHandler$DataReader.run(AbstractReadHandler.java:323)
	at alluxio.worker.grpc.BlockReadHandler$BlockDataReader.run(BlockReadHandler.java:70)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at alluxio.worker.grpc.GrpcExecutors$ImpersonateThreadPoolExecutor.lambda$execute$0(GrpcExecutors.java:82)
	at alluxio.worker.grpc.GrpcExecutors$ImpersonateThreadPoolExecutor$$Lambda$107/1844138271.run(Unknown Source)
```

ZacBlanco:
Before this change, it was possible for workers to get into a state where all reader threads were starved because locks were taken for a certain block, but never released, so every time a new reader thread would attempt to acquire a lock on that block, it would hang the thread forever. Eventually all reads would fail.

This PR attempts to fix the issue by making some more guarantees about the serialized task executor, as well as cleaning up resources and sending a proper response back to the client whenever an error occurs. The main cause of the starved readers is due to `RejectedExecutionException` being thrown when the threadpool corresponding to `alluxio.worker.network.block.reader.threads.max` was at full capacity.

First, we add a new threadpool to the GrpcExecutors class which backs the SerializedExecutors and is mainly responsible for sending the responses back to the client. We set the rejectedExecutionHandler to CallerRunsPolicy() so that if we do run out of threads, we fall back to the current thread running the task. Using this fallback policy is critical for the methods `replyError`, `replyEof`, and `replyCancel`, so the client does not get stuck waiting for a response. This now guarantees whenever these methods are called, we at least attempt to send the reply. Previously, submitting to the SerializingExecutor could fail, which means we never actually send the response to the client.

The 2nd major change is that when exceptions are thrown which should cause the stream to abort, we need to ensure that those messages are sent. The method handleStreamEndingException was added to ensure that when an exception is thrown, and we cannot continue the stream with the client, that we send a proper notification and then shut down the serializing executor and close any open blocks afterwards.

Fixes #11258


core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
```diff
@@ -0,0 +1,127 @@
...
+ * This class can execute a series of tasks on a separate thread in the order they are submitted.
+ * No guarantees can be made about the timing of how the tasks complete, or what the result of
+ * the tasks are.
+ *
+ * All tasks in the queue are guaranteed to complete before {@link #shutdown(long, TimeUnit)} is
+ * called so long as they complete within the provided timeout. At which time the task runner will
+ * stop accepting new tasks (all calls to {@link #addTask(Runnable)} will return false).
+ *
+ */
+public class SerializedTaskRunner implements Runnable {
```
jiacheliu3, line 21: nit: Do you think "sequential" is more expressive than "serialized"?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -126,14 +152,36 @@ public void onNext(alluxio.grpc.ReadRequest request) {
...
+    AlluxioStatusException statusExc =
+        new AlluxioStatusException(status);
+    if (mContext == null) {
+      mContext = createRequestContext(alluxio.grpc.ReadRequest.newBuilder().build());
+    }
+    setError(new Error(statusExc, true));
+    // After calling setError, this will run the data reader to complete the request and close
+    // any possibly opened blocks, then replies to the client with onError, and should eventually
+    // shut down the serializing executor.
+    createDataReader(mContext, mResponseObserver).run();
```
calvinjia, line 182: Previously we do all the work in a separate thread ensuring abstract read handler methods execute fairly quickly. Will executing the data reader on this thread cause significant slowdowns?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
```diff
@@ -82,27 +82,28 @@
...
-          mWorker.closeUfsBlock(context.getRequest().getSessionId(), context.getRequest().getId());
-          context.setBlockReader(null);
+      } finally {
+        if (!mWorker.unlockBlock(context.getRequest().getSessionId(),
+            context.getRequest().getId())) {
+          if (reader != null) {
+            mWorker.closeUfsBlock(context.getRequest().getSessionId(),
+                context.getRequest().getId());
+            context.setBlockReader(null);
+          }
```
calvinjia, line 98: Is there a functional change here, it seems like we lose the error message of failing to close the block reader?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -126,14 +152,36 @@ public void onNext(alluxio.grpc.ReadRequest request) {
...
     }
   }
 
+  private void handleStreamEndingException(StatusException e) {
+    handleStreamEndingException(e, e.getStatus());
+  }
+
+  // In this case, we hit an exception that does not allow us to proceed further
+  private void handleStreamEndingException(Exception e, Status status) {
+    // The executor is too busy to handle our requests. The stream should error.
```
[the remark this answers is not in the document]
ZacBlanco: you are right. Originally i had called this method `handleRejectedExecutionException` :)

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: After calling `setError`, the context status will have the `mError` flag as true. Once the error flag is set calling `createDataReader(...).run()` should be relatively fast.

There are a few places where we could see a slowdown.
- Entering the loop where we wait for `mLock`
  - https://github.com/Alluxio/alluxio/blob/557484bf6cf6923dc55ca7bfe63297dab02300c7/core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java#L358-L375
  - Because we call `setError` before running the DataReader, we immediately break from the while loop, as lock as we can access `mLock`.
- Afterwards we need to then perform any unlockBlock operations via the `completeRequest`. This should complete quickly.
  - https://github.com/Alluxio/alluxio/blob/557484bf6cf6923dc55ca7bfe63297dab02300c7/core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java#L431-L456
- Then finally will call `replyError`, which prevents any more operations from occurring. It will shutdown the serializing executor as well.
  - https://github.com/Alluxio/alluxio/blob/557484bf6cf6923dc55ca7bfe63297dab02300c7/core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java#L480-L491

So yes, I think this does have some potential to slow down the handler threads.

I haven't tested yet, but I don't think calling `setError` is enough to guarantee that we call `replyError`. The problem arises if there aren't enough executor threads to actually create the new data reader to send the final `mResponseObserve.onError` request. Running it here is a proactive measure against another possible `RejectedExecutionException`

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
ZacBlanco: good catch

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: Maybe instead we can make a comprise. As long as we submit the data reader to the executor successfully, we don't really care when it executes. Just that it does get run at some point after `setError`.

So in the handler here, we can do

```java
try {
  mDataReaderExecutor.submit(createDataReader(...));
} catch (RejectedExecutionException e) {
  createDataReader(...).run();
}
```

What do you think?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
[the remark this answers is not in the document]
ZacBlanco: fixed

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -126,14 +152,36 @@ public void onNext(alluxio.grpc.ReadRequest request) {
...
+    Long sessionId = mContext.getRequest() == null ? null : mContext.getRequest().getSessionId();
+    LogUtils.warnWithException(LOG, "handling RejectedExecutionException: sessionId: {}",
+        sessionId, e);
+    AlluxioStatusException statusExc =
```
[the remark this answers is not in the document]
ZacBlanco: fixed

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
ZacBlanco: updated

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
```diff
@@ -0,0 +1,138 @@
...
+
+  @Override
+  public void run() {
+    mRunningThread = Thread.currentThread();
+    while (mRunning) {
+      try {
+        Runnable task = mTasks.take();
+        try {
+          task.run();
+        } catch (RuntimeException e) {
```
witgo, line 71: Throwable

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
calvinjia: How about a separate threadpool with infinite size queue?

ZacBlanco: Right now we limit the number of clients that can read from a single worker by the max size of threadpool that we submit data readers to. By making that queue size unlimited, it makes it possible for the queue to grow excessively large under high load and the latency between submitting a data reader and actually beginning to process could end up being larger than client timeouts. Right now, when we submit a data reader, it's guaranteed to run immediately unless the thread pool is at max capacity.

I think it's better to fail fast when reaching a specified load on the worker, rather than letting  too many clients connect. With the new retries it will force the clients to try a new worker quicker than if we use an infinite-size queue.

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ZacBlanco: Updated.

Also, I apologize if this conflicts with your other PR. Correct me if I'm wrong, I don't think we're solving the same problems. We recently found the issue which I'm trying to solve in this PR. I guess the timing is just unfortunate.

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
```diff
@@ -0,0 +1,138 @@
...
+ * This class can execute a series of tasks on a separate thread in the order they are submitted.
+ * No guarantees can be made about the timing of how the tasks complete, or what the result of
+ * the tasks are.
+ *
+ * All tasks in the queue are guaranteed to complete before {@link #shutdown(long, TimeUnit)} is
+ * called so long as they complete within the provided timeout. At which time the task runner will
+ * stop accepting new tasks (all calls to {@link #addTask(Runnable)} will return false).
+ *
+ */
+public class SerializedTaskRunner implements Runnable {
```
ggezer, line 33: Why do we need this?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -126,11 +153,46 @@ public void onNext(alluxio.grpc.ReadRequest request) {
...
+      if (mContext == null) {
+        mContext = createRequestContext(alluxio.grpc.ReadRequest.newBuilder().build());
+      }
+      setError(new Error(statusExc, true));
+    }
+    // After calling setError, this will run the data reader to complete the request and close
+    // any possibly opened blocks, then replies to the client with onError, and should eventually
+    // shut down the serializing executor.
+    // If possible, we should free up this thread immediately by running the data reader in a
+    // separate thread because we want to avoid the possibility of blocking the `onNext` method
```
ggezer, line 189: We shouldn't care.
Thread here is allocated from gRPC's internal no-limit pool, just to serve onNext() and terminate afterward.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -337,7 +408,7 @@ private void runInternal() {
 
           if (chunk != null) {
             DataBuffer finalChunk = chunk;
-            mSerializingExecutor.execute(() -> {
+            mSerializingExecutor.addTask(() -> {
```
ggezer, line 411: Why not send it in-line?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -409,18 +482,16 @@ private void runInternal() {
...
-    protected abstract DataBuffer getDataBuffer(T context, StreamObserver<ReadResponse> response,
-        long offset, int len) throws Exception;
+    protected abstract DataBuffer getDataBuffer(T context, long offset, int len) throws Exception;
 
     /**
      * Writes an error read response to the channel and closes the channel after that.
      */
     private void replyError(Error error) {
-      mSerializingExecutor.execute(() -> {
+      mSerializingExecutor.addTask(() -> {
```
ggezer, line 494: Why not send it in-line?

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
```diff
@@ -48,6 +48,12 @@
           THREAD_STOP_MS, TimeUnit.MILLISECONDS, new SynchronousQueue<>(),
           ThreadFactoryUtils.build("BlockDataReaderExecutor-%d", true)));
 
+  public static final ExecutorService BLOCK_READER_SERIALIZED_RUNNER_EXECUTOR =
+      new ImpersonateThreadPoolExecutor(new ThreadPoolExecutor(THREADS_MIN,
+          ServerConfiguration.getInt(PropertyKey.WORKER_NETWORK_BLOCK_READER_THREADS_MAX),
```
ggezer, line 53: What does `WORKER_NETWORK_BLOCK_READER_THREADS_MAX` limit when used with this pool?

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
```diff
@@ -0,0 +1,138 @@
...
+     * Spawn a new task executor which can begin queueing new tasks.
+     *
+     * It is not guaranteed to immediately begin running tasks. It will depend on the underlying
+     * executor's ability to allocate a thread to this task runner.
+     *
+     * @return a new {@link SerializedTaskRunner}
+     */
+    public SerializedTaskRunner create() {
+      SerializedTaskRunner runner = new SerializedTaskRunner();
+      mService.execute(runner);
```
ggezer, line 135: Does this mean every new `SerializedTaskRunner` will hold a thread off this pool?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -237,9 +301,13 @@ protected abstract DataReader createDataReader(T context,
...
   public void onReady() {
     try (LockResource lr = new LockResource(mLock)) {
-      if (shouldRestartDataReader()) {
-        mDataReaderExecutor.submit(createDataReader(mContext, mResponseObserver));
-        mContext.setDataReaderActive(true);
+      if (shouldRestartDataReader() && !mFinished.get()) {
+        try {
+          mDataReaderExecutor.submit(createDataReader(mContext, mResponseObserver));
+          mContext.setDataReaderActive(true);
+        } catch (RejectedExecutionException e) {
```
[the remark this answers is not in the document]
ZacBlanco: There aren't any other that can be thrown from this code path.

Additionally, it's not good practice to catch `Exception`.

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ZacBlanco: Each data read handler needs to submit responses using `onNext`, `onError`, and `onCompleted`. In order for the client to understand the protocol properly, we can't have `onNext` messages intermingled with `onError` or `onCompleted`. That's why we used the `SerializingExecutor` before, in order to ensure tasks that were submitted completed one after another.

For one, the previous approach was bad, because it used an internal gRPC class which was not designed to be exposed to users of the library.

Two, because it previously used the same underlying executor as the data reader. In order to send messages back to the client in the proper order, we relied on always being able to submit a task to the serializing executor. If submitting one task failed, the client could time out get a failure because it never actually received the data (failed to submit task which replied with the data), or it could timeout even after the worker attempted to send `onError` or `onCompeted` because it coudl fail to submit the task that calls those methods. Basically when the thread pool filled up before, we would simply stop sending messages back to clients - even ones that had already connected.

This new class is designed so that each new creation of `AbstractReadHandler` corresponds with a lease of a `SerializedTaskRunner`. During the lifetime of the task runner it's guaranteed to be able to run the tasks that are submitted because it uses a dedicated thread. This ensures any client which begins to be handled by the worker can complete or terminate properly because we guarantee that at least one of `StreamObserver#onError` or `StreamObserver#onCompleted` are called.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: Sorry, I don't quite understand

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ZacBlanco: Yes, until the task runner is shut down

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
ZacBlanco: I used that value because it's the same one that corresponds to the `BlockDataReaderExecutor`. We should never have more task executors than we have threads from the `BlockDataReaderExecutor`

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: I was told previously that the "no limit" pool was merely for simple RPCs. 

For the stream handling aspect, the number of threads designated to serve the stream observers are coming from the workers netty threadpool, which is finite.

ggezer: I'd expect that only under `directExecutor()` mode which is not the case here.
Netty is supposed to do proto parsing and allocate a thread from the gRPC server executor for calling `StreamObserver<>` callbacks with deserialized prototypes.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ggezer: Again, why do we need to spawn a new thread?

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
ggezer: We will have more streams than we have reader-threads.
Not all streams are in a state to a read block, so total count should be allowable to be more than that. Otherwise, we're stalling completions/cancellations until readers are finished.

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ggezer: Is this the same with gRPC's `SerializingExecutor` ?

ZacBlanco: As far as I'm aware, no

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: This doesn't spawn a new thread, only sends a task to the thread responsible for sending messages back to the client

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: @bf8086 and @calvinjia I think have said something otherwise, or maybe I misunderstood. Maybe we should all have a short discussion?

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ggezer: Then maybe we should switch back to it for now?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -100,9 +103,10 @@
   AbstractReadHandler(ExecutorService executorService,
       StreamObserver<ReadResponse> responseObserver, AuthenticatedUserInfo userInfo) {
     mDataReaderExecutor = executorService;
-    mSerializingExecutor = new SerializingExecutor(executorService);
     mResponseObserver = responseObserver;
     mUserInfo = userInfo;
+    mSerializingExecutor =
```
witgo, line 108: Why do we need two thread pools?

witgo: Removed  `mDataReaderExecutor ` in #10540, it looks fine.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -126,11 +153,46 @@ public void onNext(alluxio.grpc.ReadRequest request) {
...
+   * @param exception the exception thrown
+   * @param status the type of {@link Status} exception which should be returned to the user
+   */
+  private void handleStreamEndingException(Exception exception, Status status) {
+    Long sessionId = mContext.getRequest() == null ? null : mContext.getRequest().getSessionId();
+    LogUtils.warnWithException(LOG, "fatal error occurred while handling read. sessionId: {}",
+        sessionId, exception);
+    AlluxioStatusException statusExc = new AlluxioStatusException(status);
+    try (LockResource lr = new LockResource(mLock)) {
+      if (mContext == null) {
```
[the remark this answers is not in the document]
ZacBlanco: We can, but we should still create one, because in the case that we want to respond with an error, and there are packets in flight that could trigger the handler's `onNext`, then it might still try to run the data reader even though we've already sent the error response.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: 1. The data reader thread shouldn't be blocked by sending responses back to the client. mSerializingExecutor could block the data reader from doing I/O. Ideally we minimize latency by using a dedicated threadpool for the data readers performing I/O.
2. The serializing executor should guarantee that everything submitted to it can be run (no RejectedExecutionExceptions). In order to guarantee this, we use a cached threadpool with a max capacity of `Integer.MAX_VALUE`. The pool in my testing does not grow above 10 threads or so, but if we use the same executor as the data reader, then we could fail to send responses to the client after erroring, reading data, or even completing the read entirely because submitting the tasks in the `reply*` methods could fail.

ZacBlanco: For context, I'm trying to snuff out issues when we hit the capacity of the `GrpcExecutors#BLOCK_READER_EXECUTOR`.

In my testing I set this to some artificially low value (32, 64, 128), and then generate a few thousand clients worth of load on a single worker. I want to make sure the system behaves well when we scale those up to 2048 threads and anywhere from 10k-50k clients, or at least fails gracefully in such situations.

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ZacBlanco: updates

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
ZacBlanco: updated

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
witgo: Cross-thread operations on large data blocks will cause CPU cache invalidation and decrease the throughput of worker processes.

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
```diff
@@ -48,6 +48,11 @@
           THREAD_STOP_MS, TimeUnit.MILLISECONDS, new SynchronousQueue<>(),
```
witgo, line 48: If it is changed to `new LinkedBlockingQueue<>()`,  the mDataReaderExecutor#submit will not throw a `RejectedExecutionException` exception .

ZacBlanco: I see, thanks for clarifying

core/common/src/main/java/alluxio/util/executor/SerializedTaskRunner.java
ZacBlanco: This has been removed

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: This code has been removed

ZacBlanco:
Done

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -439,8 +485,12 @@ private void replyEof() {
       mSerializingExecutor.execute(() -> {
         try {
           Preconditions.checkState(!mContext.isDoneUnsafe());
-          mContext.setDoneUnsafe(true);
-          mResponse.onCompleted();
+          if (!mContext.isDoneUnsafe()) {
+            mContext.setDoneUnsafe(true);
+            mResponse.onCompleted();
+          } else {
+            LOG.warn("Tried to replyEof when stream was already finished. context: {}", mContext);
```
bf8086, line 492: When will this happen? Will it be too chatty on the log?

ZacBlanco: From all of my testing, I never saw any of these logs appear. It's possible there's cases I haven't triggered yet. But I have about 500MB of worker logs from my testing and I don't see a single one where we tried to reply twice

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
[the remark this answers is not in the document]
bf8086: I see. That will never happen in current logic unless the client is hacked. I am fine with being extra cautious though.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: I can move this to debug logging? or remove it all together. In the 500M of logs I have, I only seen the `replyError` one

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -439,8 +485,12 @@ private void replyEof() {
       mSerializingExecutor.execute(() -> {
         try {
           Preconditions.checkState(!mContext.isDoneUnsafe());
```
bf8086, line 487: remove this line?

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -484,7 +484,6 @@ private void replyError(Error error) {
     private void replyEof() {
       mSerializingExecutor.execute(() -> {
         try {
-          Preconditions.checkState(!mContext.isDoneUnsafe());
```
ggezer, line 441: Why is this the case now?

ZacBlanco: we shouldn't need to throw an error if the context is already done (sent a stream-ending response back to the client). Rather than throw the exception, we just log it now in case that we do hit this and want to debug further. It is uncommon, but does happen from time to time.

ZacBlanco:
@witgo FYI we're going to get this PR in so that we can apply the changes for a 2.2 patch release. We'll review your PR soon. The changes in your PR will likely be slated for 2.3.

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
ZacBlanco: We can test that change in your PR. I am trying to limit the scope of this so that it is suitable for a 2.2.X release

Thanks for your for help!

ZacBlanco:

jiacheliu3:
@ZacBlanco Should this be cherry-picked?

ZacBlanco:
Thanks for checking. It should be eventually, but I'm waiting for some more testing first

ZacBlanco:

jiacheliu3:
@ZacBlanco  https://github.com/Alluxio/alluxio/issues/11309 is in Alluxio 2.1.1 so I think it should be cherry-picked to 2.1 too.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractReadHandler.java
```diff
@@ -355,19 +397,18 @@ private void runInternal() {
...
         } catch (Exception e) {
           LogUtils.warnWithException(LOG,
-              "Exception occurred while reading data for read request {}.", mContext.getRequest(),
+              "Exception occurred while reading data for read request {}. session {}",
+              mContext.getRequest(), mContext.getRequest().getSessionId(),
               e);
           setError(new Error(AlluxioStatusException.fromThrowable(e), true));
-          continue;
         }
+        continue;
```
ggezer, line 411: @ZacBlanco Is there a reason for moving this continue out? Is it possible for this to cause a tight-loop here?

ZacBlanco: While this looks like a typo I made during the PR. Looking at the logic I don't think it should have any effect.

The bracket `}` below the `continue` matches to the outer-loop `while (true) {`. The bracket above matches to the exception handler

Previously, the continue statement was within the exception handler which would have sent the program back to the beginning of the loop.

However, now that `continue` is the last statement in the loop which is essentially a no-op since it would have returned to the beginning anyway. I think even before it would have been a no-op because after exiting the exception handler the program flow would have sent it to the beginning of the loop anyway.

PR #11334, opened 2020-04-23
Prevent format calls in UnderFileSystemWithLogging

### LINKED ISSUE #11237: Optimize the UnderFileSystemWithLogging class

Most, if not all, UFS calls go through the `UnderFileSystemWithLogging` class. Each call executes the [`call`](https://github.com/Alluxio/alluxio/blob/8555b84420df79ff18700dafc64d49a85bb47ac4/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java#L946-L958) method using the internal [`UfsCallable<T>`](https://github.com/Alluxio/alluxio/blob/8555b84420df79ff18700dafc64d49a85bb47ac4/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java#L924-L937) class.

However, many of the methods use a [`String.format`](https://github.com/Alluxio/alluxio/blob/8555b84420df79ff18700dafc64d49a85bb47ac4/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java#L894) which then subsequently has part of the formatted string (the method name), [immediately extracted](https://github.com/Alluxio/alluxio/blob/8555b84420df79ff18700dafc64d49a85bb47ac4/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java#L948)

This is quite inefficient. We should pass the method names and path arguments as a separate field in the `UfsCallable<T>` and avoid doing a `String.format` unless necessary

A simple solution is to remove the `toString()` method and simply add a `getMethodName` and `getPathArgument`

```java
  public interface UfsCallable<T> {
    /**
     * Executes the call.
     *
     * @return the result of the call
     */
    T call() throws IOException;
    String getMethodName();
    String getPathArgument();
  }
```

**Urgency**
low. The ratio of performance impact depends on the UFS call. It is especially high for calls like `getAclPair` which returns null in many cases unless using HDFS.

ZacBlanco:
Previously, every single call to the UFS would perform a String.format and then String.split call to extract the method name. This is inefficient. Now, it will simply pull the method name from the callable object. After this, String.format is only called if the log statement is enabled.

Fixes #11237

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
+++ b/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
@@ -87,8 +87,8 @@ public class UnderFileSystemWithLogging implements UnderFileSystem {
       }
 
       @Override
-      public String toString() {
-        return "Cleanup";
+      public String methodName() {
+        return "cleanup";
       }
     });
   }
@@ -103,8 +103,8 @@ public class UnderFileSystemWithLogging implements UnderFileSystem {
       }
 
       @Override
-      public String toString() {
-        return "Close";
+      public String methodName() {
+        return "close";
       }
     });
   }
@@ -119,8 +119,13 @@ public class UnderFileSystemWithLogging implements UnderFileSystem {
       }
 
       @Override
+      public String methodName() {
+        return "ConnectFromMaster";
+      }
+
+      @Override
       public String toString() {
-        return String.format("ConnectFromMaster: hostname=%s", hostname);
+        return String.format("hostname=%s", hostname);
       }
     });
   }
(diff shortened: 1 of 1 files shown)


ZacBlanco:
From my testing, I think the previous overhead can be in the range of 5-10% just due to the formatting. I think `getQualifiedMetricName` is lower, but definitely still has some overhead. I think it can be addressed in another PR.

I found this while working on the metadata sync improvements. We spent tons of time on String.format because sync required a call to `getAclPair` which would just return `(null, null)` most of the time. Overhead I think was usually in the range of 5-10% due to string.format. I am unsure about `getQualifiedMetricName`

calvinjia:

ZacBlanco:

PR #11345, opened 2020-04-24
Improve UserState log

jiacheliu3:
There are multiple classes extending `UserState` with specific logic for `login()`. When an Exception is thrown, the exact class name will be helpful.


gpang:

PR #11354, opened 2020-04-26
Stop tracking connections in copycat client/server

### LINKED ISSUE #11355: Copycat transport continuously cycles connections causing memory leak
**Alluxio Version:**
2.3.0-SNAPSHOT

**Describe the bug**

When starting the master, if monitoring the number of threads and heap memory you should see spikes in the thread count as well as an associated increase in memory. After a while, if using a tool like VisualVM to force GC, eventually no more GC will occur, but the heap still grows.

This memory leak is due to the fact that the copycat client and server transport implementations (GrpcMessagingClient, GrpcMessagingServer) store an unbounded list of previous connections. Those lists are not purged until the client or server is closed. In the case of a single master deployment, that seems to only be on the startup and shutdown of the servers.

The memory leak is a product of many connections being opened and closed by the copycat client. After some debugging, it is clear that the GrpcMessagingConnections are constantly opened and closed due to the leader address in the copycat server not matching the address reported by the keepalive response (`localhost:19200` and `localhost/127.0.0.1:19200`). The difference is a resolved InetSocketAddress vs an unresolved one.

To fix the issue we need the serialization logic in the copycat protocol for KeepAliveResponse to respect whether or not the address is resolved. This is fixed in copycat 1.2.15

**To Reproduce**

- Compile on `2.3.0-SNAPSHOT` (as of the day this issue was created)
- Run
```console
$ ./bin/alluxio-start master
```
- Monitor heap+threads in VisualVM

**Expected behavior**

Number of threads should be closer to constant. Memory should not increase indefinitely.

**Urgency**
High. Master is unusable after a few hours.

**Additional context**

This behavior was most recently introduced due to changes in how we handle gRPC resources in #11335

ZacBlanco:
Many connections can be created by the transport layer, but it is rare for copycat to close the client. Tracking the created connections is unnecessary because copycat closes them before closing the clients anyways. This prevents us from holding on to the connection references for extended periods of time, which then causes behavior resembling a memory leak.

Additionally we bump the copycat version to 1.2.15 to avoid cycling connections. 1.2.15 introduces a changes to the KeepAliveResponse serde to respect unresolved addresses.

Fixes #11355

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/transport/GrpcMessagingServer.java
+++ b/core/server/common/src/main/java/alluxio/master/transport/GrpcMessagingServer.java
@@ -29,10 +29,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
@@ -56,9 +52,6 @@ public class GrpcMessagingServer implements Server {
   /** Listen future. */
   private CompletableFuture<Void> mListenFuture;
 
-  /** List of all connections created by this server. */
-  private final List<Connection> mConnections;
-
   /** Executor for building server listener. */
   private final ExecutorService mExecutor;
 
@@ -79,7 +72,6 @@ public class GrpcMessagingServer implements Server {
     mUserState = userState;
     mExecutor = executor;
     mProxy = proxy;
-    mConnections = Collections.synchronizedList(new LinkedList<>());
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


ZacBlanco:
Ah, a net removal of lines. Refreshing :)

core/server/common/src/main/java/alluxio/master/transport/GrpcMessagingServer.java
```diff
@@ -140,40 +127,16 @@ public GrpcMessagingServer(AlluxioConfiguration conf, UserState userState,
...
 
-  /**
-   * Used to keep track of all connections created by this server instance.
-   *
-   * @param serverConnection new client connection
-   */
-  private synchronized void addNewConnection(Connection serverConnection) {
-    mConnections.add(serverConnection);
+    return CompletableFuture.runAsync(() -> {
+      // Shut down gRPC server once all connections are closed.
```
ggezer, line 132: nit: update the comment

ZacBlanco: updated

ZacBlanco:
Jenkins, test this please

ZacBlanco:
alluxio-bot, test this please

ZacBlanco:

ZacBlanco:

PR #11356, opened 2020-04-26
Use an event-loop per network-group

ggezer:


ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
It looks like, an event-loop per connection yields higher throughput.

ggezer:
@calvinjia @ZacBlanco PTAL.

ZacBlanco:
Does this mean each JVM will now have a max of 2 event loops? 1 for streaming, 1 for RPC?

core/common/src/main/java/alluxio/grpc/GrpcConnectionPool.java
```diff
@@ -276,40 +276,41 @@ private void shutdownManagedChannel(ManagedChannel managedChannel, AlluxioConfig
...
       }
 
       // Create a new event-loop.
       ChannelType nettyChannelType =
           NettyUtils.getChannelType(PropertyKey.Template.USER_NETWORK_NETTY_CHANNEL
-              .format(channelKey.getNetworkGroup().getPropertyCode()), conf);
+              .format(key.getPropertyCode()), conf);
       int nettyWorkerThreadCount =
           conf.getInt(PropertyKey.Template.USER_NETWORK_NETTY_WORKER_THREADS
-              .format(channelKey.getNetworkGroup().getPropertyCode()));
```
ZacBlanco, line 294: Let's say we have 2 clients with different config for the netty worker threads. If one client is already active, the 2nd client's thread config will be ignored, correct?

ggezer: Yes. Connection level properties should be consistent.

ZacBlanco: That behavior seems a little sketchy to me

ggezer: How about we bring singleton back for a subset of props?

ZacBlanco: I don't really see a good way to do this unless we make the # of threads unconfigurable/unbounded. Otherwise if we expose the configuration to users it will be kind of confusing. Either way, I think that is fine to work on in a later PR.

ggezer: How about `CheckConsistency.ENFORCE` ?

ZacBlanco: Consistency level won't prevent an application in a client JVM writing `FileSystem.Factory.create(conf)` where the conf has two separate values for the netty threadpool

ggezer:

ggezer:

PR #11370, opened 2020-04-29
Improve Helm configmap template format

jiacheliu3:
This moves JavaOpts generation out of the multi-line properties, and puts them into the head section.

The reason for doing this is we cannot insert hashtag comments in a multi-line section like below. So if someone wants to generate comments in the helm code, that breaks the helm template.
```
ALLUXIO_MASTER_JAVA_OPTS: |-
  # This will propagate to the generated YAML, but not allowed in a multi-line property
  {{- /* This will not appear in the generated YAML */}}
  {{- $alluxioJavaOpts = printf "-Dalluxio.master.journal.type=%v" .Values.journal.type | append $alluxioJavaOpts }}
  ...omitted...
```


madanadit:

PR #11376, opened 2020-05-03
Improve PR build by inspecting PR diff

Xenorith:
we currently use this jenkins plugin to run our PR build checks: https://plugins.jenkins.io/ghprb/

the plugin defines an environment variable for the target branch that the PR is going to merge into. that means we can run `git --no-pager diff origin/master --name-only` to get a list of files that were changed in the PR. note this change also required a few tweaks to the jenkins job, most importantly the addition of pulling more refs so i can reference the ref marked as origin/master
since we have two checks, one for the maven build and one for docs, i defined two booleans that get set to true if a corresponding file is found in the diff. if this target branch variable is not available, it defaults to running all checks.
if there are more files or folders that we can incorporate into the diff check, we can waste less cycles running irrelevant checks

### DIFF
--- a/dev/jenkins/build.sh
+++ b/dev/jenkins/build.sh
@@ -35,6 +35,9 @@ else
     if [[ ${filepath} =~ ^docs/.* ]]; then
       # if any file starts with "docs/", run doc check
       RUN_DOC_CHECK="true"
+    elif [[ ${filepath} =~ ^integration/(dataproc|docker|emr|kubernetes|vagrant)/.* ]]; then
+      # do nothing for files in integration/ that don't contain java code
+      :
     else
       # if any other files are in the diff, run maven
       RUN_MAVEN="true"
--- a/dev/jenkins/build.sh
+++ b/dev/jenkins/build.sh
@@ -27,10 +27,7 @@ if [ -z "${TARGET_BRANCH}" ]; then
   RUN_MAVEN="true"
   RUN_DOC_CHECK="true"
 else
-  if [ -z "${TARGET_REMOTE}" ]; then
-    TARGET_REMOTE="origin"
-  fi
-  git --no-pager diff "${TARGET_REMOTE}/${TARGET_BRANCH}" --name-only > prFiles.diff
+  git --no-pager diff "refs/remotes/origin/${TARGET_BRANCH}" --name-only > prFiles.diff
   echo "PR diff is:"
   cat prFiles.diff
 
--- a/dev/jenkins/build.sh
+++ b/dev/jenkins/build.sh
@@ -15,34 +15,69 @@
 #
 set -e
 
-# Set things up so that the current user has a real name and can authenticate.
-myuid=$(id -u)
-mygid=$(id -g)
-echo "$myuid:x:$myuid:$mygid:anonymous uid:/home/jenkins:/bin/false" >> /etc/passwd
-
-export MAVEN_OPTS="-Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS"
-
-if [ -z ${ALLUXIO_BUILD_FORKCOUNT} ]
-then
-  ALLUXIO_BUILD_FORKCOUNT=4
-fi
-
 if [ -n "${ALLUXIO_GIT_CLEAN}" ]
 then
   git clean -fdx
 fi
-mvn -Duser.home=/home/jenkins -T 4C clean install -Pdeveloper -Dmaven.javadoc.skip -Dsurefire.forkCount=${ALLUXIO_BUILD_FORKCOUNT} $@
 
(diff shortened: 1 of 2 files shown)


Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

Xenorith:
jenkins, test this please

dev/jenkins/build.sh
```diff
@@ -13,36 +13,68 @@
...
+  echo "PR diff is:"
+  cat prFiles.diff
+
+  while IFS="" read -r filepath || [ -n "$filepath" ]; do
+    if [[ ${filepath} =~ ^docs/.* ]]; then
+      # if any file starts with "docs/", run doc check
+      RUN_DOC_CHECK="true"
+    else
+      # if any other files are in the diff, run maven
+      RUN_MAVEN="true"
```
ZacBlanco, line 40: what about the case where there is no diff, and we simply want to generate the jacoco report?

dev/jenkins/build.sh
```diff
@@ -13,36 +13,68 @@
...
+  then
+    # A separate step to run jacoco report, with all the generated coverage data. This requires the
+    # previous 'install' step to generate the jacoco exec data with the 'jacoco' profile.
+    #
+    # Must exclude some of the modules that fail to run verify again without a clean step. This is ok
+    # since these modules do not contain any source code to track for code coverage.
+    mvn -T 4C -Dfindbugs.skip -Dcheckstyle.skip -DskipTests -Dmaven.javadoc.skip -Dlicense.skip -PjacocoReport verify -pl '!webui,!shaded,!shaded/client,!shaded/hadoop'
+    # run sonar analysis
+    mvn $(echo "${ALLUXIO_SONAR_ARGS}") sonar:sonar
+  fi
```
ZacBlanco, line 71: similarly, if there's no files in the diff and we just want to generate sonar reports, how will we do that now?

dev/jenkins/build.sh
Xenorith: before checking for `TARGET_BRANCH`, i could add a
```
if [ -z "${ALLUXIO_SONAR_ARGS}"  ]; then
  RUN_MAVEN="true"
fi
```
== if the sonar args are provided, then force run maven

ZacBlanco: I think that's reasonable

ZacBlanco: @gpang might have an opinion on a better way though?

ZacBlanco:
@Xenorith another case I thought of is for the building on the OSMaster job. That just runs on a nightly basis without any changes either. We may need to add yet another flag as well to just denote whether we should skip the diff check all together and just run everything?

dev/jenkins/build.sh
```diff
@@ -13,36 +13,68 @@
...
-  #
-  # Must exclude some of the modules that fail to run verify again without a clean step. This is ok
-  # since these modules do not contain any source code to track for code coverage.
-  mvn -T 4C -Dfindbugs.skip -Dcheckstyle.skip -DskipTests -Dmaven.javadoc.skip -Dlicense.skip -PjacocoReport verify -pl '!webui,!shaded,!shaded/client,!shaded/hadoop'
-  # run sonar analysis
-  mvn $(echo "${ALLUXIO_SONAR_ARGS}") sonar:sonar
+RUN_MAVEN="false"
+RUN_DOC_CHECK="false"
+if [ -z "${TARGET_BRANCH}" ]; then
+  # if no target branch is specified, run all checks
```
Xenorith, line 26: @ZacBlanco i think this check addresses the scenario for the master builds. if this script isn't invoked by the PR build, this env var won't be set => defaults to running everything

ZacBlanco: got it, thanks!

dev/jenkins/build.sh
```diff
@@ -13,36 +13,68 @@
...
+  RUN_MAVEN="true"
+  RUN_DOC_CHECK="true"
+else
+  git --no-pager diff "refs/remotes/origin/${TARGET_BRANCH}" --name-only > prFiles.diff
+  echo "PR diff is:"
+  cat prFiles.diff
+
+  while IFS="" read -r filepath || [ -n "$filepath" ]; do
+    if [[ ${filepath} =~ ^docs/.* ]]; then
```
madanadit, line 35: @Xenorith the following under integration also do not require a maven build: `dataproc, docker, emr, kuberneres, vagrant`

ZacBlanco:

PR #11377, opened 2020-05-04
Implement Job Service based master stress tools

gpang:


gpang:
Jenkins, test this please

stress/common/src/main/java/alluxio/stress/graph/BarGraph.java
```diff
@@ -0,0 +1,94 @@
...
+   *
+   * @param title the title
+   * @param subTitle the lines in the sub title
+   * @param xTitle the title of the x-axis
+   */
+  public BarGraph(String title, List<String> subTitle, String xTitle) {
+    mTitle.add(title);
+    mSubTitle.addAll(subTitle);
+
+    mGraph.put("$schema", "https://vega.github.io/schema/vega-lite/v4.json");
```
ZacBlanco, line 62: hm. Should we be relying on external resources for this? I guess for prototyping it is fine, but in the case we want to test this somewhere without external internet connection, it may not be possible to use this.

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -0,0 +1,131 @@
...
+      command.add("runClass");
+      command.add(className);
+      command.addAll(Arrays.asList(args));
+      command.add(BaseParameters.IN_PROCESS_FLAG);
+      command.addAll(mBaseParameters.mJavaOpts);
+
+      LOG.info("running command: " + String.join(" ", command));
+      return ShellUtils.execCommand(command.toArray(new String[0]));
+    }
+  }
```
ZacBlanco, line 130: should we throw an exception if neither `mInProcess` or `mCluster` is true?

stress/common/src/main/java/alluxio/stress/master/MasterBenchParameters.java
```diff
@@ -0,0 +1,77 @@
...
+          + "termination condition.")
+  public int mStopCount = STOP_COUNT_INVALID;
+
+  @Parameter(names = {"--warmup"},
+      description = "The length of time to warmup before recording measurements. (1m, 10m, 60s, "
+          + "10000ms, etc.)")
+  public String mWarmup = "30s";
+
+  @Parameter(names = {"--fixed-count"},
+      description = "The number of paths in the fixed portion. Must be greater than 0.")
```
ZacBlanco, line 72: I don't quite understand what the "Fixed portion" is. Can you elaborate a bit more?

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -0,0 +1,382 @@
...
+    long durationMs = FormatUtils.parseTimeSize(mParameters.mDuration);
+    long warmupMs = FormatUtils.parseTimeSize(mParameters.mWarmup);
+    long startMs = mBaseParameters.mStartMs;
+    if (mBaseParameters.mStartMs == BaseParameters.UNDEFINED_START_MS) {
+      startMs = CommonUtils.getCurrentMs() + 1000;
+    }
+    long endMs = startMs + warmupMs + durationMs;
+    BenchContext context = new BenchContext(rateLimiter, startMs, endMs);
+    for (int i = 0; i < mParameters.mThreads; i++) {
+      service.submit(new BenchThread(context, mCachedFs[i % mCachedFs.length]));
```
ZacBlanco, line 149: you could use the executors service's `InvokeAll` method instead? It clean up some of this code:

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#invokeAll-java.util.Collection-long-java.util.concurrent.TimeUnit-

stress/shell/src/main/java/alluxio/stress/cli/report/GenerateReport.java
```diff
@@ -0,0 +1,134 @@
...
+    try (PrintWriter writer = new PrintWriter(
+        new FileWriter(Paths.get(outputDir.getAbsolutePath(), "index.html").toString()))) {
+
+      writer.println("<!DOCTYPE html>");
+      writer.println("<head>");
+      writer.println("<script src=\"https://cdn.jsdelivr.net/npm/vega@5\"></script>");
+      writer.println("<script src=\"https://cdn.jsdelivr.net/npm/vega-lite@4\"></script>");
+      writer.println("<script src=\"https://cdn.jsdelivr.net/npm/vega-embed@6\"></script>");
+      writer.println("</head>");
+      writer.println("<body>");
```
ZacBlanco, line 109: This seems a bit hacky...hmm. I wonder if there's a better way to write this out rather than putting HTML in java strings within the code.

Maybe we can use a template in the `resources` directory instead and format that?

ZacBlanco: Or use a real HTML parsing/object creation library instead?

stress/shell/src/main/java/alluxio/stress/cli/report/GenerateReport.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Generates a report from summaries.
+ */
+public class GenerateReport {
+  private static final Logger LOG = LoggerFactory.getLogger(GenerateReport.class);
+
+  @ParametersDelegate
+  private Parameters mParameters = new Parameters();
+
+  @com.beust.jcommander.Parameters(separators = "=")
```
ZacBlanco, line 43: what's the reason for using `=` here?

stress/shell/src/main/java/alluxio/stress/cli/report/GenerateReport.java
```diff
@@ -0,0 +1,134 @@
...
+
+    File outputDir = new File(mParameters.mOutputDir);
+    outputDir.mkdirs();
+
+    GraphGenerator graphGenerator = inputs.get(0).graphGenerator();
+
+    List<Graph> graphs = graphGenerator.generate(inputs);
+
+    try (PrintWriter writer = new PrintWriter(
+        new FileWriter(Paths.get(outputDir.getAbsolutePath(), "index.html").toString()))) {
```
ZacBlanco, line 101: I would like to see a more descriptive name rather than `index.html`

stress/shell/src/main/java/alluxio/stress/cli/report/GenerateReport.java
```diff
@@ -0,0 +1,134 @@
...
+
+      writer.println("<!DOCTYPE html>");
+      writer.println("<head>");
+      writer.println("<script src=\"https://cdn.jsdelivr.net/npm/vega@5\"></script>");
+      writer.println("<script src=\"https://cdn.jsdelivr.net/npm/vega-lite@4\"></script>");
+      writer.println("<script src=\"https://cdn.jsdelivr.net/npm/vega-embed@6\"></script>");
```
ZacBlanco, line 107: Similar comment here about relying on external resources. Not too sure how I feel about it.

While convenient, it also now depends on these resources always being available for the life of the code.

stress/shell/src/main/java/alluxio/stress/cli/suite/Suite.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Parameter(names = {CLUSTER_FLAG},
+        description = "If true, runs the benchmark via the job service cluster. Otherwise, runs "
+            + "locally.")
+    public boolean mCluster = false;
+
+    @Parameter(names = {"--java-opt"},
+        description = "The java options to add to the command line to for the task. This can be "
+            + "repeated. The options must be quoted and prefixed with a space, to avoid getting "
+            + "passed to the JVM. For example: --java-opt \" -Xmx4g\" --java-opt \" -Xms2g\"")
```
ZacBlanco, line 43: similar comment about having to _always_ add a space. I think we can do that ourselves?

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
gpang: done

stress/shell/src/main/java/alluxio/stress/cli/report/GenerateReport.java
gpang: Well, the resource is not tied to this code, but tied to the generated html file. Is there something else we can do?

stress/shell/src/main/java/alluxio/stress/cli/report/GenerateReport.java
gpang: I had considered a library, but those seemed more heavyweight for a simple file that I was generating. If I wanted more to the generated file, then maybe we need a library or template. I will look into it some...

stress/shell/src/main/java/alluxio/stress/cli/suite/Suite.java
ZacBlanco: If we are passing these on to another program, then we will get these values values as strings in the java program beforehand. What I'm suggesting is that if every property is supposed to be quoted and have a space before it, why don't we make it easier to just pass the property and do this in the code:

```java
mJavaOpts = mJavaOpts.stream()
  .map(property -> String.format("\" %s\"", property))
  .collect(Collectors.toList()) 
```

job/client/src/main/java/alluxio/client/job/JobGrpcClientUtils.java
```diff
@@ -106,6 +107,10 @@ private static JobInfo waitFor(final long jobId, AlluxioConfiguration alluxioCon
           }
           return true;
         }, WaitForOptions.defaults().setInterval(1000));
+      if (resultInfo != null) {
+        resultInfo = client.getJobStatusDetailed(resultInfo.getId());
```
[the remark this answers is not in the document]
bradyoo: I would prefer if we didn't always do this. There are some customers where attempting to get JobStatusDetailed currently crashes unless they increase some GRPC data size limit. If there is a caller that requires a detailed job status, can the caller call it instead?

gpang:
Jenkins, test this please

job/client/src/main/java/alluxio/client/job/JobGrpcClientUtils.java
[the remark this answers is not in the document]
gpang: I reverted this change.

gpang:
Jenkins, test this please

gpang:
alluxio-bot, test this please

bradyoo:
LGTM

gpang:

PR #11383, opened 2020-05-05
Make reserved space on directories only when required

ggezer:
This is to improve the compatibility of multi-tier management for configurations that don't need the functionality.

For when the reserved space is required by a later configuration change, the `swap-restore` task will make sure reserved space will be carved out if necessary.

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTier.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTier.java
@@ -63,7 +63,7 @@ public final class StorageTier {
     mTierOrdinal = new WorkerStorageTierAssoc().getOrdinal(tierAlias);
   }
 
-  private void initStorageTier()
+  private void initStorageTier(boolean isMultiTier)
       throws BlockAlreadyExistsException, IOException, WorkerOutOfSpaceException {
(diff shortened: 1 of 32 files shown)


ggezer:
@gpang  | @calvinjia PTAL.

core/server/worker/src/main/java/alluxio/worker/block/meta/StorageTier.java
```diff
@@ -99,7 +104,7 @@ private void initStorageTier()
       int mediumTypeindex = i >= dirMedium.length ? dirMedium.length - 1 : i;
       long capacity = FormatUtils.parseSpaceSize(dirQuotas[index]);
       try {
-        StorageDir dir = StorageDir.newStorageDir(this, i, capacity, reservedSpaceBytes,
+        StorageDir dir = StorageDir.newStorageDir(this, i, capacity, reservedBytes,
```
calvinjia, line 107: Should we reserve the total tier reserve bytes in each directory in the tier?

ggezer: This is indeed per-dir reserved bytes.
There is no total reserved bytes defined since we want to be able to make progress on each directory.

calvinjia:

PR #11399, opened 2020-05-08
Move validation tools from examples/ module into shell/

jiacheliu3:
This moves `runTests`, `runUfsTests` and `journalCrashTest` from `examples/` to `shell/` module.

Moving `journalCrashTest` too as it uses `CliUtils`. `CliUtils` is also renamed to `RunTestUtils` to be more expressive.


jiacheliu3:
jenkins, test this please

jiacheliu3:
alluxio-bot, test this please

shell/pom.xml
```diff
@@ -51,14 +51,18 @@
...
-   <dependency>
-     <groupId>org.apache.httpcomponents</groupId>
-     <artifactId>httpclient</artifactId>
-   </dependency>
-   <dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+    </dependency>
+    <dependency>
```
ZacBlanco, line 58: please fix the spacing on this dependency

shell/pom.xml
```diff
@@ -82,8 +86,8 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-        <groupId>org.apache.commons</groupId>
-        <artifactId>commons-compress</artifactId>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
```
ZacBlanco, line 90: please revert if there is no change here

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -111,42 +106,47 @@ private void runS3Operations() throws Exception {
...
    */
-  private void loadAndRunTests(Object operations, String testDir) throws Exception {
+  private int loadAndRunTests(Object operations, String testDir) throws Exception {
+    int failedTestCnt = 0;
     try {
       Class classToRun = operations.getClass();
-      Field[] fields = classToRun.getDeclaredFields();
-      for (Field field : fields) {
-        field.setAccessible(true);
-      }
```
ZacBlanco, line 129: why remove this?

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -111,42 +106,47 @@ private void runS3Operations() throws Exception {
...
         String testName = test.getName();
         if (testName.endsWith("Test")) {
           System.out.printf("Running test: %s...", testName);
+          boolean passed = false;
           try {
             test.invoke(operations);
-          } catch (InvocationTargetException e) {
+            passed = true;
+            cleanupUfs(testDir);
+          } catch (Exception e) {
```
ZacBlanco, line 133: we should avoid catch-all here. Is there a reason to use `Exception` over InvocationTargetException?

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -111,42 +106,47 @@ private void runS3Operations() throws Exception {
...
-            throw new IOException(e.getTargetException());
+            System.err.format("Test %s.%s aborted%n%s", test.getClass(), test.getName(), e);
+          } finally {
+            RunTestUtils.printPassInfo(passed);
+            if (!passed) {
+              failedTestCnt++;
+            }
           }
-          System.out.println("Test Passed!");
-          cleanupUfs(testDir);
```
ZacBlanco, line 144: should we not put `cleanupUfs` even on failure?

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -111,42 +106,47 @@ private void runS3Operations() throws Exception {
...
+            passed = true;
+            cleanupUfs(testDir);
+          } catch (Exception e) {
             if (mUfs.getUnderFSType().equals(S3_IDENTIFIER)) {
               logRelatedS3Operations(test);
             }
-            throw new IOException(e.getTargetException());
+            System.err.format("Test %s.%s aborted%n%s", test.getClass(), test.getName(), e);
+          } finally {
+            RunTestUtils.printPassInfo(passed);
```
ZacBlanco, line 139: does this print something different based on whether `passed` is true or false?

shell/pom.xml
jiacheliu3: I added the spaces just because these lines were incorrectly aligned. Please compare with the lines above and below.

shell/pom.xml
jiacheliu3: See comment above

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
jiacheliu3: Reverted. I made a mistake thinking these are all Methods. Tks for the catch.

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
jiacheliu3: Moved `cleanupUfs()` up into the finally block.

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
jiacheliu3: I renamed `CliUtils` to `RunTestUtils`, as `CliUtils` contains only util methods local to `runUfsTests`, instead of general to all other cli classes.

`RunTestUtils.printPassInfo()` corresponds to 
https://github.com/Alluxio/alluxio/blob/c4eba4706750ffd7ad00c486af9454b4342bf914/examples/src/main/java/alluxio/cli/CliUtils.java#L34

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
jiacheliu3: I think it makes sense to let all the tests complete, instead of letting one ufs operation err out and aborts the `runUfsTests`.

jiacheliu3:
@ZacBlanco do you mind merging it if this looks good to go?

bf8086:

PR #11400, opened 2020-05-08
Improve HdfsPositionedReadUnderFileInputStream with a heuristic

yuzhu:
The heuristic works as follows :
1. it starts with a pread, because a lot of times the client only does a single large read, and we want to benefit from pread in that case. 
2. if there are certain number of sequential reads in a row, we switch to buffered read mode. In this mode, preads are serviced via pread calls, and read calls are services by read calls. A sequential read is defined as a read that is within a movement limit of the previous read. This is to guard against workloads such as read, skip(2), read, skip(3) etc.
3. any non-sequential read, we change back to pread mode.

### DIFF
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
@@ -14,6 +14,7 @@ package alluxio.underfs.hdfs;
 import alluxio.underfs.SeekableUnderFileInputStream;
 import alluxio.util.io.BufferUtils;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.PositionedReadable;
@@ -31,10 +32,12 @@ public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputSt
   // TODO(david): make these parameters configurations and add diagnostic metrics.
   // After this many number of sequential reads (reads without large skips), it
   // will switch to sequential read mode.
-  private static final int SEQUENTIAL_READ_LIMIT = 3;
+  @VisibleForTesting
+  static final int SEQUENTIAL_READ_LIMIT = 3;
   // This describes the number of bytes that we can move forward in a stream without
   // switching to random read mode.
-  private static final int MOVEMENT_LIMIT = 512;
+  @VisibleForTesting
+  static final int MOVEMENT_LIMIT = 512;
 
   private long mPos;
   // The heuristic is that if there are certain number of sequential reads in a row,
@@ -87,9 +90,8 @@ public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputSt
   @Override
   public int read(byte[] buffer, int offset, int length) throws IOException {
     int bytesRead;
-    if (mSequentialReadCount >= SEQUENTIAL_READ_LIMIT) {
+    if (isSequentialReadMode() && mPos != ((Seekable) in).getPos()) {
       ((Seekable) in).seek(mPos);
-      mSequentialReadCount = 0;
     }
     if (mPos == ((Seekable) in).getPos()) {
       // same position, use buffered reads as default
@@ -104,12 +106,14 @@ public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputSt
     return bytesRead;
   }
 
+  private boolean isSequentialReadMode() {
+    return mSequentialReadCount >= SEQUENTIAL_READ_LIMIT;
+  }
+
   @Override
   public void seek(long position) throws IOException {
     if (position < mPos || position - mPos > MOVEMENT_LIMIT) {
       mSequentialReadCount = 0;
-    } else {
-      ((Seekable) in).seek(position);
     }
     mPos = position;
   }
@@ -121,8 +125,6 @@ public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputSt
     }
     if (n > MOVEMENT_LIMIT) {
       mSequentialReadCount = 0;
-    } else {
-      ((Seekable) in).seek(mPos + n);
     }
     mPos += n;
     return n;
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
@@ -16,10 +16,7 @@ import alluxio.util.io.BufferUtils;
 
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.fs.FSDataInputStream;
-import org.apache.hadoop.hdfs.client.HdfsDataInputStream;
-import org.slf4j.Logger;
 
-import java.io.BufferedInputStream;
 import java.io.IOException;
 
 /**
@@ -29,15 +26,19 @@ import java.io.IOException;
 public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputStream {
   private static final int SEQUENTIAL_READ_LIMIT = 3;
   private static final int MOVEMENT_LIMIT = 512;
+
   private long mPos;
+  // The heuristic is that if there are certain number of sequential reads in a row,
+  // we switch to buffered read mode. In this mode, preads are serviced via pread calls,
+  // and read calls are services by read calls. A sequential read is defined as a read that
+  // is within a movement limit of the previous read. This is to guard against workloads
+  // such as read, skip(2), read, skip(3) etc.
   private int mSequentialReadCount;
-  private long mLastRead;
 
   HdfsPositionedUnderFileInputStream(FSDataInputStream in, long pos) {
     super(in);
     mPos = pos;
     mSequentialReadCount = 0;
-    mLastRead = pos;
   }
 
   @Override
@@ -102,7 +103,6 @@ public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputSt
       ((FSDataInputStream) in).seek(position);
     }
     mPos = position;
-
   }
 
   @Override
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
@@ -22,11 +22,17 @@ import org.apache.hadoop.fs.Seekable;
 import java.io.IOException;
 
 /**
- * The input stream of HDFS as under filesystem. This input stream supports seeking but internally
- * uses the positionedRead {@link FSDataInputStream} API. This stream can be cached for reuse.
+ * The input stream of HDFS as under filesystem. This input stream has two mode of operations.
+ * Under sequential mode, it uses the read api and can take advantage of underlying stream's
+ * buffering. Under random read mode, it uses the positionedRead {@link FSDataInputStream} API.
+ * This stream can be cached for reuse.
  */
 public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputStream {
+  // After this many number of sequential reads (reads without large skips), it
+  // will switch to sequential read mode.
   private static final int SEQUENTIAL_READ_LIMIT = 3;
+  // This describes the number of bytes that we can move forward in a stream without
+  // switching to random read mode.
   private static final int MOVEMENT_LIMIT = 512;
 
   private long mPos;
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
@@ -90,32 +90,29 @@ public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputSt
     }
     if (bytesRead > 0) {
       mPos += bytesRead;
+      mSequentialReadCount++;
     }
-    mSequentialReadCount++;
     return bytesRead;
   }
 
   @Override
   public void seek(long position) throws IOException {
-    if (Math.abs(position - mPos) > MOVEMENT_LIMIT) {
+    if (position < mPos || position - mPos > MOVEMENT_LIMIT) {
       mSequentialReadCount = 0;
-    } else {
-      ((FSDataInputStream) in).seek(position);
     }
     mPos = position;
   }
 
   @Override
   public long skip(long n) throws IOException {
+    if (n <= 0) {
+      return 0;
+    }
     if (n > MOVEMENT_LIMIT) {
       mSequentialReadCount = 0;
-      mPos += n;
-      return n;
-    } else {
-      long skipped = ((FSDataInputStream) in).skip(n);
-      mPos += skipped;
-      return skipped;
     }
+    mPos += n;
+    return n;
   }
 
   @Override
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
@@ -28,6 +28,7 @@ import java.io.IOException;
  * This stream can be cached for reuse.
  */
 public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputStream {
+  // TODO(david): make these parameters configurations and add diagnostic metrics.
   // After this many number of sequential reads (reads without large skips), it
   // will switch to sequential read mode.
   private static final int SEQUENTIAL_READ_LIMIT = 3;
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
@@ -16,7 +16,10 @@ import alluxio.util.io.BufferUtils;
 
 import com.google.common.base.Preconditions;
 import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.hdfs.client.HdfsDataInputStream;
+import org.slf4j.Logger;
 
+import java.io.BufferedInputStream;
 import java.io.IOException;
 
 /**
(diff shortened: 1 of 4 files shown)


underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,24 +71,51 @@ public int read(byte[] b) throws IOException {
...
     if (bytesRead > 0) {
       mPos += bytesRead;
     }
+    mSequentialReadCount++;
     return bytesRead;
   }
 
   @Override
   public void seek(long position) throws IOException {
+    if (Math.abs(position - mPos) > MOVEMENT_LIMIT) {
```
calvinjia, line 100: If we seek backward should sequential count always be set to 0?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,24 +71,51 @@ public int read(byte[] b) throws IOException {
...
+    if (mPos == ((FSDataInputStream) in).getPos()) {
+      // same position, use buffered reads as default
+      bytesRead = in.read(buffer, offset, length);
+    } else {
+      bytesRead = ((FSDataInputStream) in).read(mPos, buffer, offset, length);
+    }
     if (bytesRead > 0) {
       mPos += bytesRead;
     }
+    mSequentialReadCount++;
```
calvinjia, line 94: Should this only be incremented if bytesRead > 0?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,24 +71,51 @@ public int read(byte[] b) throws IOException {
...
   @Override
   public long skip(long n) throws IOException {
-    mPos += n;
-    return n;
+    if (n > MOVEMENT_LIMIT) {
+      mSequentialReadCount = 0;
+      mPos += n;
+      return n;
+    } else {
+      long skipped = ((FSDataInputStream) in).skip(n);
```
calvinjia, line 115: Is it guaranteed that `in` is positioned correctly?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -24,11 +24,21 @@
  * uses the positionedRead {@link FSDataInputStream} API. This stream can be cached for reuse.
  */
 public class HdfsPositionedUnderFileInputStream extends SeekableUnderFileInputStream {
+  private static final int SEQUENTIAL_READ_LIMIT = 3;
+  private static final int MOVEMENT_LIMIT = 512;
```
calvinjia, line 28: Do these numbers require tuning or is the heuristic good enough?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: good catch, fixed

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: yeah i was undecided about this one, but probably rare to see someone jump forward and then jump back in the stream?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: done

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: We could further tune these, but experimentally journal replays are as fast as before, will verify remote spark tpcds behavior soon.

madanadit: nit: add a comment describing these as counts and not a bytes limit

madanadit: if they require tuning, i prefer and property. we can choose to make it hidden in case we don't want to add to the user's confusion.

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,22 +71,46 @@ public int read(byte[] b) throws IOException {
...
   }
 
   @Override
   public long skip(long n) throws IOException {
+    if (n <= 0) {
+      return 0;
+    }
+    if (n > MOVEMENT_LIMIT) {
+      mSequentialReadCount = 0;
+    }
```
calvinjia, line 113: We still need to skip in the stream if we are in the sequential read mode and n is less than the movement limit?

yuzhu: it is not necessary, because our mPos is the ground truth. As long as we maintain mPos, we can delay any seek or skip until when we actually read.

calvinjia: In that case, a read pattern like `read, read, read, skip(1), read, skip(1), read` will cause us to lose the sequential read mode at the last read. Is this intended?

yuzhu: yes, fixed. it became obvious once i started writing the test.

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: So I am actually leaning towards making these not tunable for now. A tunable parameter is only useful if there is some metric / log message that suggests these parameters need to be tuned. Right now we don't have any of that. These default values should work well out of the box. 

How about we add a TODO here to make another PR with the appropriate diagnostic metrics and make it tunable in that PR?

yuzhu:
@calvinjia @madanadit  PTAL,thanks

yuzhu:
alluxio-bot, test this please

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
```diff
@@ -63,4 +87,85 @@ public void prepareConfiguration() throws Exception {
     Assert.assertEquals(ufsConf.get(PropertyKey.UNDERFS_HDFS_IMPL), conf.get("fs.hdfs.impl"));
     Assert.assertTrue(conf.getBoolean("fs.hdfs.impl.disable.cache", false));
   }
+
+  class PreadSeekableStream extends FilterInputStream implements Seekable, PositionedReadable {
```
madanadit, line 91: nit: this doesn't need to be a nested class? separate file for readability

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
```diff
@@ -63,4 +87,85 @@ public void prepareConfiguration() throws Exception {
...
+    in.read();
+    in.read();
+    in.read();
+    in.read();
+    in.skip(2);
+    in.skip(2);
+    in.read();
+    in.skip(2);
+    in.read();
+    verify(spyStream, never()).read(anyInt(), any(byte[].class), anyInt(), anyInt());
```
madanadit, line 158: nit: comment why `pread` is never called

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
```diff
@@ -63,4 +87,85 @@ public void prepareConfiguration() throws Exception {
...
+    verify(spyStream, times(6)).read(any(byte[].class), anyInt(), anyInt());
+    in.skip(1000);
```
madanadit, line 160: nit: comment that `skip > MOVEMENT_LIMIT` causes pread. maybe even mark `MOVEMENT_LIMIT` `@VisibleForTesting` and re-use `2*MOVEMENT_LIMIT` to avoid changing the test if we tune the parameter.

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
```diff
@@ -63,4 +87,85 @@ public void prepareConfiguration() throws Exception {
...
+    in.read();
+    in.read();
+    in.read();
+    verify(spyStream, times(3)).read(anyInt(), any(byte[].class), anyInt(), anyInt());
+    verify(spyStream, times(6)).read(any(byte[].class), anyInt(), anyInt());
+    in.read();
```
madanadit, line 166: nit: comment that `read > SEQUENTIAL_READ_LIMIT` causes . similar to above consider exposing `SEQUENTIAL_READ_LIMIT` for testing and using a for loop here to avoid changing the test after possible tuning

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
```diff
@@ -553,19 +553,28 @@ public boolean mkdirs(String path, MkdirsOptions options) throws IOException {
...
   }
 
-  private boolean isReadLocal(FileSystem fs, Path filePath, OpenOptions options)
-      throws IOException {
+  private boolean isReadLocal(FileSystem fs, Path filePath, OpenOptions options) {
     String localHost = NetworkAddressUtils.getLocalHostName((int) mUfsConf
         .getMs(PropertyKey.NETWORK_HOST_RESOLUTION_TIMEOUT_MS));
     // Heuristic to determine whether to use positionedRead on hdfs ufs
     // If any block is not found on the same host, we use pread api
-    BlockLocation[] blockLocations = fs.getFileBlockLocations(filePath,
```
madanadit, line 562: nit: the comment above is mis-placed?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
```diff
@@ -553,19 +553,28 @@ public boolean mkdirs(String path, MkdirsOptions options) throws IOException {
...
+      blockLocations = fs.getFileBlockLocations(filePath,
+          options.getOffset(), options.getLength());
+      if (blockLocations == null) {
+        // no blocks exist
+        return true;
       }
+
+      for (BlockLocation loc : blockLocations) {
+        if (Arrays.stream(loc.getHosts()).noneMatch(localHost::equals)) {
+          // Some blocks are remote only, use pread api to HDFS
```
madanadit, line 572: nit: also this comment seems mis-placed as this method makes no decision about pread api

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
```diff
@@ -63,4 +87,85 @@ public void prepareConfiguration() throws Exception {
...
+      return ((FSDataInputStream) in).seekToNewSource(l);
+    }
+  }
+  /**
+   * Tests the dynamic switching between pread and read calls to underlying stream.
+   */
+  @Test
+  public void verifyPread() throws Exception {
+    File file = mTemporaryFolder.newFile("test.txt");
+    FileUtils.writeByteArrayToFile(file, CommonUtils.randomBytes(4096));
```
madanadit, line 142: nit: in this test can we also check if the data read back is what we expect, in addition to the API used to read?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,22 +80,50 @@ public int read(byte[] b) throws IOException {
...
+  /**
+   * If there has been a number of sequential reads in a row,
+   * we move to regular buffered reads.
+   */
   @Override
   public int read(byte[] buffer, int offset, int length) throws IOException {
-    int bytesRead = ((FSDataInputStream) in).read(mPos, buffer, offset, length);
+    int bytesRead;
+    if (mSequentialReadCount >= SEQUENTIAL_READ_LIMIT) {
+      ((Seekable) in).seek(mPos);
```
madanadit, line 91: is this `seek` necessary or guaranteed to be a no-op if say I have 10 continuous reads as we reset `mSequentialReadCount` and `seek` after every 3 reads?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,22 +80,50 @@ public int read(byte[] b) throws IOException {
...
     }
     return bytesRead;
   }
 
   @Override
   public void seek(long position) throws IOException {
+    if (position < mPos || position - mPos > MOVEMENT_LIMIT) {
+      mSequentialReadCount = 0;
+    } else {
+      ((Seekable) in).seek(position);
```
madanadit, line 112: what if
1. stream is in random read mode and `mPos=0`
2. `HdfsPositionedUnderFileInputeStream.seek(MOVEMENT_LIMIT-1)`
3. we end up calling seek on `FSDataInputStream` and that is undesired?

instead can we refactor this by creating an auxiliary method `isSequentialReadMode=mSequentialReadCount> SEQUENTIAL_READ_LIMIT`
```
   if (position < mPos || position - mPos > MOVEMENT_LIMIT) {
      mSequentialReadCount = 0;
    }

    if (isSequentialReadMode()) {
      ((Seekable) in).seek(position);
```

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
```diff
@@ -61,22 +80,50 @@ public int read(byte[] b) throws IOException {
...
 
   @Override
   public long skip(long n) throws IOException {
+    if (n <= 0) {
+      return 0;
+    }
+    if (n > MOVEMENT_LIMIT) {
+      mSequentialReadCount = 0;
+    } else {
+      ((Seekable) in).seek(mPos + n);
```
madanadit, line 125: nit: for readability i prefer a method called  `isSequentialReadMode` to determine whether to seek or not

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: I changed this part a bit so we don't do the noop seek. Basically we do the seek lazily if we are in sequential mode, and don't reset the counter until we seek or skip a large step.

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: see above, i now avoid this seek entirely by doing it lazily in the read.

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsPositionedUnderFileInputStream.java
yuzhu: done

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
yuzhu: done

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
yuzhu: done

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
```diff
@@ -63,4 +87,85 @@ public void prepareConfiguration() throws Exception {
     Assert.assertEquals(ufsConf.get(PropertyKey.UNDERFS_HDFS_IMPL), conf.get("fs.hdfs.impl"));
     Assert.assertTrue(conf.getBoolean("fs.hdfs.impl.disable.cache", false));
   }
+
+  class PreadSeekableStream extends FilterInputStream implements Seekable, PositionedReadable {
+
+    /**
```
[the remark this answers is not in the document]
yuzhu: done

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
yuzhu: done

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
yuzhu: done

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
yuzhu: done

underfs/hdfs/src/test/java/alluxio/underfs/hdfs/HdfsUnderFileSystemTest.java
yuzhu: done

yuzhu:
alluxio-bot, test this please

yuzhu:
alluxio-bot, test this please

yuzhu:

PR #11408, opened 2020-05-09
Add Alluxio AMI files

tieujason330:


tieujason330:
jenkins, test this please

tieujason330:
@Xenorith PTAL, thanks

integration/packer/alluxio-ami/packer.json
```diff
@@ -0,0 +1,168 @@
+{
+  "variables": {
+    "alluxio_download_url": "",
+    "alluxio_edition_flavor": "",
+    "alluxio_version": "",
```
Xenorith, line 5: set this to be the current SNAPSHOT version and have it update as part of `dev/scripts/update_versions.sh` so the version gets updated along with all the other versions when release is cut

integration/packer/alluxio-ami/packer.json
```diff
@@ -0,0 +1,168 @@
...
+        "most_recent": true,
+        "owners": [
+          "137112412989"
+        ]
+      },
+      "ssh_keypair_name": "{{user `ec2_keypair_name`}}",
+      "ssh_private_key_file": "{{user `ec2_keypair_pem_file`}}",
+      "ssh_username": "ec2-user",
+      "tags": {
+        "Origin": "",
```
Xenorith, line 54: do we need to introduce more variables to optionally populate the tag values? i wonder if it is possible to use a map as a variable value and just say "tags": {{user `tags_map`}}

integration/packer/alluxio-ami/src/alluxio.com/factory/command.go
```diff
@@ -0,0 +1,46 @@
...
+	"os/exec"
+	"strings"
+	"time"
+)
+
+func Exec(app string, args []string) (string, error) {
+	log.Println(fmt.Sprintf("%v %v", app, strings.Join(args, " ")))
+	cmd := exec.Command(app, args...)
+	stdout, err := cmd.Output()
+
```
Xenorith, line 26: extra newline

integration/packer/alluxio-ami/packer.json
tieujason330: updated

integration/packer/alluxio-ami/packer.json
tieujason330: packer doesn't allow user variables to be objects

integration/packer/alluxio-ami/src/alluxio.com/factory/command.go
tieujason330: removed

integration/packer/alluxio-ami/packer.json
Xenorith: looks like they started introducing the terraform HCL to packer in a newer release: https://www.packer.io/guides/hcl/variables/#maps

guess we can't do this generically yet, but for now we still need to be able to set this `Origin` tag value for cleanup purposes? or do you have another way of setting the tag?

tieujason330: we can have the origin tag be a user variable

PR #11413, opened 2020-05-10
Revert version of shaded hadoop client

apc999:


ZacBlanco:
updated in post-release commits

PR #11414, opened 2020-05-11
Prevent secondary UFS journal from modifying journal files

### LINKED ISSUE #11415: On failover, the old primary might modify journal files (for the UFS journal)
**Alluxio Version:**
2.2

**Describe the bug**
For UFS journal.
When failing over, the new secondary (was primary) can rename the journal file that the new primary is writing to. This can cause some journal entries to be lost.

**To Reproduce**
- during a flush command, the primary has to lose primacy
- the new primary writes new journal entries
- the old primary, while transitioning to secondary, will try to recover the log files, and in that process, will rename the last log file (which the new primary is already writing to)
- now, the new primary is writing entries to a file which was renamed to a different filename.

**Expected behavior**
The secondary should not modify journal files.

**Urgency**
Critical, for failovers.

gpang:
Fixes #11415

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
@@ -113,6 +113,7 @@ final class UfsJournalLogWriter implements JournalWriter {
   }
 
   public synchronized void write(JournalEntry entry) throws IOException, JournalClosedException {
+    checkIsWritable();
     try {
       maybeRecoverFromUfsFailures();
       maybeRotateLog();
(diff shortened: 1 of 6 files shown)


core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
```diff
@@ -314,10 +329,19 @@ private void completeLog(UfsJournalFile currentLog, long nextSequenceNumber) thr
...
         .encodeLogFileLocation(mJournal, currentLog.getStart(), nextSequenceNumber).toString();
 
+    try {
+      // Check again before the rename
+      checkIsWritable();
+    } catch (JournalClosedException e) {
+      // Do not throw error, just ignore if the journal is not writable
+      LOG.warn("Skipping completeLog() since journal is not writable. error: {}", e.getMessage());
+      return;
+    }
```
calvinjia, line 342: We could still lose primacy after the check, but that's fine because the rename will not pick up the wrong journal file? (ie. we check for primacy between updating `current` and attempting rename)

gpang: yeah, I was thinking about that. I don't know how else to prevent the gap, but this basically the smallest we can make this gap. I'm not sure what will happen if the failover happens after the check, but before the rename...

ggezer:
jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalLogWriter.java
calvinjia: I think it should be okay because it still is referencing the old journal file.

gpang:

gpang:

PR #11422, opened 2020-05-12
Add a HDFS throughput testing tool

jiacheliu3:
This change adds a UFS I/O speed test tool using the new Benchmark framework.

```
$ bin/alluxio runUfsIOTest --threads 8 --io-size 4G --path hdfs://<hdfs-node>:9000/alluxio/ --cluster-limit --cluster > io.json
$ bin/alluxio runClass alluxio.stress.cli.report.GenerateReport --input io.json  --output ./io.html
```

The workflow above produces the I/O speed test result with graph like below
Screen Shot 2020-05-19 at 11 38 59 PM

A collected stats (UfsIOBench output) example is like below:

```
{
  "points" : [ {
    "mMode" : "WRITE",
    "mDuration" : 15.535,
    "mDataSizeMB" : 256,
    "className" : "alluxio.stress.worker.IOTaskResult$Point"
  }, {
    "mMode" : "WRITE",
    "mDuration" : 15.582,
    "mDataSizeMB" : 256,
    "className" : "alluxio.stress.worker.IOTaskResult$Point"
  }, {
    "mMode" : "WRITE",
    "mDuration" : 15.59,
    "mDataSizeMB" : 256,
    "className" : "alluxio.stress.worker.IOTaskResult$Point"
  }, ... , {
    "mMode" : "READ",
    "mDuration" : 0.376,
    "mDataSizeMB" : 256,
    "className" : "alluxio.stress.worker.IOTaskResult$Point"
  } ],
  "errors" : [ ],
  "baseParameters" : {
    "mCluster" : false,
    "mId" : "HadoopSandbox-566-masters-1-1589901871455",
    "mJavaOpts" : [ ],
    "mDistributed" : true,
    "mStartMs" : 1589902562343,
    "mInProcess" : true
  },
  "readSpeedStat" : {
    "totalDuration" : "3.017s",
    "totalSize" : "2048MB",
    "maxSpeed" : "815.2866242038217MB/s",
    "minSpeed" : "632.0987654320987MB/s",
    "avgSpeed" : "678.8200198873053MB/s",
    "stdDev" : 151.33247300467477
  },
  "writeSpeedStat" : {
    "totalDuration" : "124.41s",
    "totalSize" : "2048MB",
    "maxSpeed" : "16.504416220746567MB/s",
    "minSpeed" : "16.42078255291854MB/s",
    "avgSpeed" : "16.46169922031991MB/s",
    "stdDev" : 0.0889953940200876
  },
  "parameters" : {
    "mThreads" : 8,
    "mDataSize" : 256,
    "mPath" : "hdfs://HadoopSandbox-566-masters-1:9000/alluxio/",
    "mWorkerNum" : 1
  },
  "className" : "alluxio.stress.worker.IOTaskSummary"
}
```


ZacBlanco:
Would it be better to use the UFS interface instead so that we can test throughput for more UFSes?

madanadit:
I agree w/ Zac. The IO task should preferably use the UFS interface.

stress/common/src/main/java/alluxio/stress/worker/IOTaskSummary.java
```diff
@@ -0,0 +1,360 @@
...
+      List<String> subTitle = new ArrayList<>(Splitter.fixedLength(100).splitToList(
+              summaries.get(0).mParameters.getDescription(fieldNames.getFirst())));
+      System.out.format("subtitle: %s%n", subTitle);
+
+      for (IOTaskSummary summary : summaries) {
+        String series = summary.mParameters.getDescription(fieldNames.getSecond());
+        subTitle.add(series);
+      }
+
+      BarGraph speedGraph = new BarGraph("Read/Write speed",
```
madanadit, line 330: are these any new system requirements for plotting the graphs? or do we include everything we need in the server tarball

job/server/src/test/java/alluxio/job/plan/io/IODefinitionTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void before() {
+        mMockFileSystem = PowerMockito.mock(FileSystem.class);
+        mMockBlockStore = PowerMockito.mock(AlluxioBlockStore.class);
+        mMockFsContext = PowerMockito.mock(FileSystemContext.class);
+        PowerMockito.mockStatic(AlluxioBlockStore.class);
+        PowerMockito.when(AlluxioBlockStore.create(any(FileSystemContext.class)))
+                .thenReturn(mMockBlockStore);
+        PowerMockito.when(mMockFsContext.getClientContext())
+                .thenReturn(ClientContext.create(ServerConfiguration.global()));
+        // TODO(jiacheng): need to hijack this conf?
```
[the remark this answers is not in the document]
jiacheliu3: Please ignore all unit test related stuff here. I will make the tests more meaningful once we are more settled on the functionality.

stress/common/src/main/java/alluxio/stress/worker/IOTaskSummary.java
jiacheliu3: There's no new system requirements for plotting. The HTML generation in `alluxio.stress.graph.LineGraph/BarGraph` wraps its own JS libraries.

job/server/src/main/java/alluxio/job/plan/io/IODefinition.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Override
+  public String runTask(IOConfig config, ArrayList<String> args, RunTaskContext runTaskContext)
+          throws Exception {
+    LOG.info("args={}", args);
+
+    List<String> command = new ArrayList<>(3 + config.getArgs().size());
+    command.add(ServerConfiguration.get(PropertyKey.HOME) + "/bin/alluxio");
+    command.add("runClass");
```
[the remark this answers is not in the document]
jiacheliu3: I think the report step should be separate as we can combine multiple results to one graph like
```
$ bin/alluxio runClass alluxio.stress.cli.report.GenerateReport --input io_4threads.json --input io_8threads.json --input io_16threads.json  --output-dir ./
```

The bar graphs will be next to each other providing a better illustration like below
WeChat2108560fa7e4b18b5114959841bf8eba

job/server/src/main/java/alluxio/job/plan/io/IODefinition.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    return result;
+  }
+
+  @Override
+  public String runTask(IOConfig config, ArrayList<String> args, RunTaskContext runTaskContext)
+          throws Exception {
+    LOG.info("args={}", args);
+
+    List<String> command = new ArrayList<>(3 + config.getArgs().size());
```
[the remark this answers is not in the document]
jiacheliu3: Updated this part to use the new `runUfsIOTest` class. Now it's different from `IODefinition`. (They do share a few lines but it's too specific to be a util method)

stress/common/src/main/java/alluxio/stress/worker/IOTaskSummary.java
jiacheliu3: The dependencies are wrapped in the `LineGraph` and `BarGraph` classes. They generate their own HTML referring to the JS libs.

bin/alluxio
```diff
@@ -32,7 +32,9 @@ function printUsage {
   echo -e "                \t NOTE: This command will stop the existing server and creates a new one!"
   echo -e "  runMesosTest  \t Test the Alluxio integration with Mesos. Try 'alluxio mesosTest -help' for more help."
   echo -e "                \t NOTE: This command requires Mesos to be running and will stop any Alluxio servers that are currently running."
-  echo -e "  runUfsTests    \t Test the integration between Alluxio and the target under filesystem. Try 'alluxio runUfsTests -help' for more help."
+  echo -e "  runUfsIOTest  \t Test the Alluxio throughput to the UFS path. Try 'alluxio runUfsIOTest -help' for more help."
+  echo -e "                \t NOTE: This command requrires a valid ufs path to write temporary files into."
+  echo -e "  runUfsTests   \t Test the integration between Alluxio and the target under filesystem. Try 'alluxio runUfsTests -help' for more help."
```
jiacheliu3, line 37: removed one extra space

stress/shell/src/main/java/alluxio/stress/cli/UfsIOBench.java
```diff
@@ -0,0 +1,262 @@
...
+        pool.shutdownNow();
+        pool.awaitTermination(30, TimeUnit.SECONDS);
+
+        // Aggregate the task results
+        return result;
+    }
+
+    @Override
+    public void prepare() throws Exception {
+        // TODO(jiacheng): any HDFS conf to add?
```
jiacheliu3, line 87: Any HDFS conf that will affect the IO throughput and I should add it here?

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -91,23 +99,16 @@ public String run(String[] args) throws Exception {
...
-      // remove the cluster flag
-      List<String> commandArgs =
-          Arrays.stream(args).filter((s) -> !BaseParameters.CLUSTER_FLAG.equals(s))
-              .filter((s) -> !s.isEmpty()).collect(Collectors.toList());
-
-      commandArgs.addAll(mBaseParameters.mJavaOpts);
-
-      long jobId = JobGrpcClientUtils
-          .run(new StressBenchConfig(className, commandArgs, 10000, mBaseParameters.mClusterLimit),
-              0, conf);
```
jiacheliu3, line 104: Moved into util method

jiacheliu3:
@gpang PTAL this adds an IO benchmark tool measuring the IO to UFS. Please let me know if I'm not using your benchmark tool in the way you designed. Thanks!

stress/common/src/main/java/alluxio/stress/worker/WorkerBenchParameters.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Parameters used in the UFS I/O throughput test.
+ * */
+public class WorkerBenchParameters extends Parameters {
+  @Parameter(names = {"--threads"}, description = "the number of threads to use")
+  public int mThreads = 16;
+
+  @Parameter(names = {"--io-size"},
+          description = "size of data to write or read in total, in MB")
+  public int mDataSize = 4096;
```
ZacBlanco, line 27: consider making this a String and using `FormatUtils` to parse a byte instead. Then users can specify 1k, 4k, 10M, 10g, etc instead. I think it is more user friendly.

job/server/src/main/java/alluxio/job/plan/stress/StressBenchDefinition.java
```diff
@@ -103,37 +93,13 @@ public String runTask(StressBenchConfig config, ArrayList<String> args,
     }
 
     command.addAll(args);
-
-    LOG.info("running command: " + String.join(" ", command));
```
gpang, line 107: was this intentional?

job/server/src/main/java/alluxio/job/plan/io/IODefinition.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Set;
+
+/**
+ * The definition for the UFS I/O throughput job, which generates concurrent streams to the UFS.
+ *
+ * {@link IOConfig} is the configuration class, each task takes a List<String> as a list of
+ * command-line arguments to the benchmark command, and each task returns the string output.
+ * The output will be serialized and deserialized using JSON, collected and merged.
+ */
+public class IODefinition implements PlanDefinition<IOConfig, ArrayList<String>, String> {
```
gpang, line 41: why can't this just reuse the stress definition? It looks similar?

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -126,8 +142,7 @@ public String run(String[] args) throws Exception {
       command.addAll(Arrays.asList(args));
       command.add(BaseParameters.IN_PROCESS_FLAG);
       command.addAll(mBaseParameters.mJavaOpts);
-
-      LOG.info("running command: " + String.join(" ", command));
+      LOG.debug("running command: " + String.join(" ", command));
```
gpang, line 145: Why is this debug now?

stress/shell/src/main/java/alluxio/stress/cli/UfsIOBench.java
```diff
@@ -0,0 +1,258 @@
...
+  @Override
+  public PlanConfig generateJobConfig(String[] args) {
+    // remove the cluster flag
+    List<String> commandArgs =
+            Arrays.stream(args).filter((s) -> !BaseParameters.CLUSTER_FLAG.equals(s))
+                    .filter((s) -> !s.isEmpty()).collect(Collectors.toList());
+
+    commandArgs.addAll(mBaseParameters.mJavaOpts);
+    String className = this.getClass().getCanonicalName();
+    return new IOConfig(className, commandArgs, mParameters);
```
gpang, line 64: this test couldn't re-use the existing definition? It someone defeats the purpose with only using a small portion of the existing stress benchmark?

job/server/src/main/java/alluxio/job/plan/io/IODefinition.java
jiacheliu3: Now it does look very similar since the `StressBenchDefinition#selectExecutors` selects k workers with the `mClusterLimit` parameter....

jiacheliu3: `IODefinition` and `IOConfig` are now removed. I think I misunderstood how to use `StressBenchDefinition` and `StressBenchConfig`...

stress/common/src/main/java/alluxio/stress/worker/WorkerBenchParameters.java
jiacheliu3: This is a cool idea. However I think using an integer in the unit of MB makes the most sense for this test. It doesn't really mean much running an I/O with KB and i think using int gets us away from annoying decimal sizes. I lean towards keeping it this way. What do you think?

stress/common/src/main/java/alluxio/stress/client/ClientIOTaskResult.java
```diff
@@ -132,22 +132,17 @@ private long computeLastEndMs() {
...
         if (it.hasNext()) {
           throw new IOException(
               "ClientIO is a single node test, so multiple task results cannot be aggregated.");
         }
-        if (!(taskResult instanceof ClientIOTaskResult)) {
-          throw new IOException(
-              "TaskResult is not of type ClientIOTaskResult. class: " + taskResult.getClass()
-                  .getName());
-        }
-        return (ClientIOTaskResult) taskResult;
```
jiacheliu3, line 150: I think generics can help remove some boilerplate code here

stress/common/src/main/java/alluxio/stress/worker/WorkerBenchParameters.java
gpang: I think Zac was saying to make this a string value, we can we specify like `4m`. In the code, we can convert it to an int with `FormatUtils.parseSpaceSize` like https://github.com/Alluxio/alluxio/blob/master/stress/shell/src/main/java/alluxio/stress/cli/client/StressClientIOBench.java#L239

stress/common/src/main/java/alluxio/stress/worker/WorkerBenchParameters.java
```diff
@@ -0,0 +1,37 @@
...
+package alluxio.stress.worker;
+
+import alluxio.stress.Parameters;
+
+import com.beust.jcommander.Parameter;
+
+/**
+ * Parameters used in the UFS I/O throughput test.
+ * */
+public class WorkerBenchParameters extends Parameters {
```
gpang, line 21: I think this should be renamed to something like `UfsIOParameters`? The ony test that uses this class is a test called `UfsIOBench`.

stress/shell/src/main/java/alluxio/stress/cli/UfsIOBench.java
```diff
@@ -0,0 +1,260 @@
...
+  @Override
+  public PlanConfig generateJobConfig(String[] args) {
+    // remove the cluster flag
+    List<String> commandArgs =
+            Arrays.stream(args).filter((s) -> !BaseParameters.CLUSTER_FLAG.equals(s))
+                    .filter((s) -> !s.isEmpty()).collect(Collectors.toList());
+
+    commandArgs.addAll(mBaseParameters.mJavaOpts);
+    String className = this.getClass().getCanonicalName();
+    return new StressBenchConfig(className, commandArgs, 0, mParameters.mWorkerNum);
```
gpang, line 64: Can you not reuse the clusterLimit parameter already available in base parameters? Also, the wait time should be non-zero, and probably be at least 5 or 10 seconds, so the distributed task can start on time.

I feel like this `generateJobConfig` method doesn't need to be overridden.

gpang:
```
16:13:28.662 [ERROR] Test for floating point equality in alluxio.stress.worker.IOTaskResult$Point.equals(Object) [alluxio.stress.worker.IOTaskResult$Point] At IOTaskResult.java:[line 195] FE_FLOATING_POINT_EQUALITY
16:13:28.662 [ERROR] alluxio.stress.worker.IOTaskSummary$StatDeserializer is Serializable; consider declaring a serialVersionUID [alluxio.stress.worker.IOTaskSummary$StatDeserializer] At IOTaskSummary.java:[lines 210-242] SE_NO_SERIALVERSIONID
16:13:28.663 [ERROR] alluxio.stress.worker.IOTaskSummary$StatSerializer is Serializable; consider declaring a serialVersionUID [alluxio.stress.worker.IOTaskSummary$StatSerializer] At IOTaskSummary.java:[lines 185-204] SE_NO_SERIALVERSIONID
```

stress/shell/src/main/java/alluxio/stress/cli/UfsIOBench.java
jiacheliu3: You are right. Keeping this override is an overlook. Removed.

Actually in my test I recorded the duration in each read/write operation so this start time is not used anywhere.

stress/common/src/main/java/alluxio/stress/worker/WorkerBenchParameters.java
jiacheliu3: Updated using formatted String now instead of int.

jiacheliu3:
alluxio-bot, test this please

stress/common/src/main/java/alluxio/stress/worker/IOTaskSummary.java
```diff
@@ -0,0 +1,371 @@
...
+
+    /**
+     * An empty constructor.
+     * */
+    public StatSerializer() {
+      super(SpeedStat.class);
+    }
+
+    @Override
+    public void serialize(SpeedStat value, JsonGenerator jgen, SerializerProvider provider)
```
gpang, line 198: Why do serializers and deserializers need to be used? Jackson can just ser/de the classes, with getters/setters, or with public fields, right? That is how all the other results and parameter classes work.

jiacheliu3: I provided customized ser/de because I need some better formatting on the field, so both the intermediate and final results are much more human readable.

Before formatting:
```
{"totalDuration":"2.0","totalSize":"415236096","maxSpeed":"200.0","minSpeed":"196.0","avgSpeed":"198.0","stdDev":2.8284271247461903}
```

After  formatting:
```
{"totalDuration":"2.0s","totalSize":"396.00MB","maxSpeed":"200.0MB/s","minSpeed":"196.0MB/s","avgSpeed":"198.0MB/s","stdDev":2.8284271247461903}
```

gpang: I see, so is the json intended to be the final human-readable output? Typically, these json files are fed into generate, to get the human-readable output.

madanadit: Well this json will be read by jason's new UI, which will render a graph. So it doesn't need to be human readable

jiacheliu3: Ah I see what you mean. Yes I intend to use this as the final output, as we are exploring the possibility to move the graph generation logic into out frontend. I think it also makes some sense if the user is reading the results before the GenerateReport step.

jiacheliu3: Also I think it's slightly more maintainable if we pass data with units to the frontend, as it's a mixture of "Bytes", "MB", and "MB/s".

gpang: I am a bit confused. If this json is the final human-readable output, then yes, these values can just be strings. But, if this json is supposed to be data (to be later graphed or displayed), then I think it would make sense for the value to be the scalar primitive. 

If units are a concern, other results have the units in the name of the field, like "durationMs", or "iompbs".

jiacheliu3: Having units in the field name sounds like the way to go. I'll update this.

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

madanadit:

PR #11424, opened 2020-05-12
Add bootstrap build scripts to tweak bootstrap files

tieujason330:
`build.sh` will have logic to make custom changes to bootstrap files. For starters, the alluxio download url can be replaced


tieujason330:
@Xenorith, PTAL thanks. The scripts are very basic for now, and will hopefully allow for easy customizations in the future

ZacBlanco:
if you want to avoid splits in logic for sed, you can try using the `perl`-based approach in `./dev/scripts/update-versions.sh`

integration/dataproc/build.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+# replace ALLUXIO_DOWNLOAD_URL in dataproc files
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  sed -i '' -e "s|^readonly ALLUXIO_DOWNLOAD_URL.*\$|readonly ALLUXIO_DOWNLOAD_URL=\"${ALLUXIO_DOWNLOAD_URL}\"|" ${DIR}/*
+else
+  sed -i -e "s|^readonly ALLUXIO_DOWNLOAD_URL.*\$|readonly ALLUXIO_DOWNLOAD_URL=\"${ALLUXIO_DOWNLOAD_URL}\"|" ${DIR}/*
```
Xenorith, line 26: if the file is updated in the future in a way that results in this sed operation doing zero replacements, is there a way to detect and trigger failure? trying to avoid the scenario where things get refactored and silently fail

tieujason330:
how about running `build.sh` creates a `.generated` directory with copies of the files (regardless of changes or not)? that way we can always run `build.sh` and use whatever is in `.generated`

Xenorith:
`.generated` sounds good to me

integration/dataproc/build.sh
Xenorith: ping on this comment

tieujason330: i'll diff the `.generated` files with the original to see if changes were made

Xenorith: maybe a grep for the matching string before the sed is easier? you could replace the url with the same url that was there and result in no diff

tieujason330: ok sounds good

integration/dataproc/build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+ALLUXIO_DOWNLOAD_URL=${1}
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+mkdir -p ${DIR}/.generated
+cp ${DIR}/alluxio.sh ${DIR}/.generated
+
+# replace ALLUXIO_DOWNLOAD_URL in dataproc init script (alluxio.sh)
+if [[ -n ${ALLUXIO_DOWNLOAD_URL} ]]; then
+    if [[ -n $(grep "readonly ALLUXIO_DOWNLOAD_URL=\"${ALLUXIO_DOWNLOAD_URL}\"" "${DIR}/alluxio.sh") ]]; then
```
Xenorith, line 24: this doesn't account for the "fail to find the string to replace" scenario
if i were refactoring the script and renamed ALLUXIO_DOWNLOAD_URL to ALLUXIO_DOWNLOAD_S3_URL, this script would not replace anything but pass.
```
if alluxio download url is set {
  if string to replace isn't found by grep -> fail
  replace string with perl cmd
}
```
it's fine to replace the same string; that's not much of a concern. i brought up the scenario as a counterexample to why using `diff` wouldn't be as dependable as using `grep`

integration/dataproc/build.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo "ERROR: ${ALLUXIO_DOWNLOAD_URL} is already set. No change was made to alluxio-emr.sh"
```
Xenorith, line 25: alluxio-emr.sh -> alluxio.sh
probably more convenient to make this a bash variable

tieujason330: it'll be difficult to use `build.sh` generically if i need to know the file names beforehand

Xenorith: i meant as a bash variable like `local FILENAME=alluxio.sh`, not an argument into the script

tieujason330:
can someone help me merge this?

ZacBlanco:

PR #11426, opened 2020-05-13
Fix default metadata sync executor sizes

ZacBlanco:
Previously, if a machine was too small the master can fail to start up.
The min pool size could have been greater than Runtime.getRuntime().availableProcessors() which would cause an IllegalStateException to be thrown

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2142,7 +2142,9 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE =
       new Builder(Name.MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE)
-          .setDefaultValue(32)
+          .setDefaultSupplier(() -> Runtime.getRuntime().availableProcessors(),
+              "The total number of threads which can concurrently execute metadata sync "
+                  + "operations.")
           .setDescription("The number of threads used to execute all metadata sync"
               + "operations")
           .setScope(Scope.MASTER)
@@ -2150,7 +2152,9 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE =
       new Builder(Name.MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE)
-          .setDefaultValue(32)
+          .setDefaultSupplier(() -> Math.max(1, Runtime.getRuntime().availableProcessors() / 2),
+              "The number of threads which can concurrently fetch metadata from UFSes during a "
+                  + "metadata sync operations")
           .setDescription("The number of threads used to fetch UFS objects for all metadata sync"
               + "operations")
           .setScope(Scope.MASTER)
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2142,15 +2142,19 @@ public String toString() {
...
+                  + "operations.")
           .setDescription("The number of threads used to execute all metadata sync"
               + "operations")
           .setScope(Scope.MASTER)
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
   public static final PropertyKey MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE =
       new Builder(Name.MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE)
-          .setDefaultValue(32)
+          .setDefaultSupplier(() -> Math.max(1, Runtime.getRuntime().availableProcessors() / 2),
```
ggezer, line 2155: This default is different than what you used before. Is it intentional?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -380,12 +380,14 @@
 
   final ThreadPoolExecutor mSyncPrefetchExecutor = new ThreadPoolExecutor(
       ServerConfiguration.getInt(PropertyKey.MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE),
```
ggezer, line 382: Can you change the value of `PropertyKey.Name.MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE` to meet the prop name. It currently ends with `parallelism`.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -438,6 +440,11 @@ public DefaultFileSystemMaster(BlockMaster blockMaster, CoreMasterContext master
     mSyncManager = new ActiveSyncManager(mMountTable, this);
     mTimeSeriesStore = new TimeSeriesStore();
     mAccessTimeUpdater = new AccessTimeUpdater(this, mInodeTree, masterContext.getJournalSystem());
+
+    // Sync executors should allow core threads to time out
```
gpang, line 444: What does this do? (and what is the behavior with and without this change?)

ZacBlanco: If no metadata syncs are occurring the executor will spin down threads to decrease resource utilization

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: This is intentional

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: The impact is that the first time a sync needs to occur there will be a slight overhead for that RPC to create the threads. The overhead is equivalent to the time to create the threads needed to process the sync.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: updated

ggezer:
jenkins, test this please

ZacBlanco:

PR #11428, opened 2020-05-14
Fix hashCode() for SortedBlockSet

ggezer:


ggezer:
This seems to be insufficient for fixing issues with LRFU. I'll open up another PR once fixed completely.

PR #11432, opened 2020-05-14
Stop worker heartbeat executor first on shutdown

### LINKED ISSUE #11421: Worker can hang on shutdown if master is not available
**Alluxio Version:**
2.3.0-SNAPSHOT

**Describe the bug**
A worker may hang on shutdown if the master isn't available.

**To Reproduce**
Bring up a local alluxio cluster (just master and worker). Stop the master and then stop the worker. The worker will hang for an undesireable amount of time during the shutdown sequence

**Expected behavior**
Quick worker shutdown.

**Urgency**

medium

**Additional context**

This seems to be due to an issue with the pin list sync not being able to connect to the master which prevents shutdown. I took a jstack of the worker while it was hanging. Here are the stacks of interest.

```
"SIGINT handler": awaiting notification on [0x000000076e4ee4e8], holding [0x00000006c0a00080]
	at java.lang.Object.wait(Native Method)
	at java.lang.Thread.join(Thread.java:1252)
	at java.lang.Thread.join(Thread.java:1326)
	at java.lang.ApplicationShutdownHooks.runHooks(ApplicationShutdownHooks.java:107)
	at java.lang.ApplicationShutdownHooks$1.run(ApplicationShutdownHooks.java:46)
	at java.lang.Shutdown.runHooks(Shutdown.java:123)
	at java.lang.Shutdown.sequence(Shutdown.java:167)
	at java.lang.Shutdown.exit(Shutdown.java:212)
	at java.lang.Terminator$1.handle(Terminator.java:52)
	at sun.misc.Signal$1.run(Signal.java:212)
```

The SIGINT handler runs the shutdown hooks and waits for them to execute. This triggers the following thread:

```
"alluxio-process-shutdown-hook": waiting to acquire [0x00000006c00174e0]
	at alluxio.AbstractClient.close(AbstractClient.java:310)
	at com.google.common.io.Closer.close(Closer.java:214)
	at alluxio.worker.block.DefaultBlockWorker.stop(DefaultBlockWorker.java:255)
	at alluxio.Registry.stop(Registry.java:148)
	at alluxio.worker.AlluxioWorkerProcess.stopWorkers(AlluxioWorkerProcess.java:287)
	at alluxio.worker.AlluxioWorkerProcess.stop(AlluxioWorkerProcess.java:275)
	at alluxio.ProcessUtils.lambda$stopProcessOnShutdown$0(ProcessUtils.java:96)
	at alluxio.ProcessUtils$$Lambda$84/1966670937.run(Unknown Source)
```

It is waiting to acquire `0x00000006c00174e0` which is held by:

```

"Worker Pin List Sync": sleeping, holding [0x00000006c00174e0]
	at java.lang.Thread.sleep(Native Method)
	at alluxio.time.ThreadSleeper.sleep(ThreadSleeper.java:26)
	at alluxio.retry.TimeBoundedRetry.attempt(TimeBoundedRetry.java:67)
	at alluxio.AbstractClient.connect(AbstractClient.java:209)
	at alluxio.AbstractClient.retryRPCInternal(AbstractClient.java:397)
	at alluxio.AbstractClient.retryRPC(AbstractClient.java:365)
	at alluxio.worker.file.FileSystemMasterClient.getPinList(FileSystemMasterClient.java:87)
	at alluxio.worker.block.PinListSync.heartbeat(PinListSync.java:54)
	at alluxio.heartbeat.HeartbeatThread.run(HeartbeatThread.java:119)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
```

This thread retries connecting to the master (I think 44 times?) before finally giving up and ceding the lock to the shutdown hook. I don't think this wait time is acceptable to users. Worker shutdown should be near-instantaneous for a fresh worker.

ZacBlanco:
Fixes #11421


core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -621,19 +621,33 @@ private Metrics() {} // prevent instantiation
...
       } catch (Exception e) {
         LOG.warn("Failed to check storage: {}", e.toString());
         LOG.debug("Exception: ", e);
       }
+      mThread.set(null);
     }
 
     @Override
     public void close() {
-      // Nothing to clean up
```
ggezer, line 636: Can you try swapping below two lines in `DefaultBlockWorker::stop()` instead of this logic?

```
// Stop heart-beat executors and clients.
mResourceCloser.close();
// Stop the base. (closes executors.)
super.stop();
```

ZacBlanco: Hmm so I tried this and it worked (the worker shut down in a timely manner)

But I don't quite understand why. `super.stop()` eventually calls `mExecutorService.shutdownNow()` which should interrupt the threads and stop them from executing.

But if the abstract client doesn't have the logic to respond to interrupts that I added here:
https://github.com/Alluxio/alluxio/pull/11432/files#diff-d7dc60da6f1005850b481cc5792912c5R213-R217

I don't understand why the threads are exiting now?

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -251,10 +251,10 @@ public void start(WorkerNetAddress address) throws IOException {
    */
   @Override
   public void stop() throws IOException {
-    // Stop heart-beat executors and clients.
-    mResourceCloser.close();
     // Stop the base. (closes executors.)
     super.stop();
+    // Stop heart-beat executors and clients.
```
gpang, line 256: It seems like this order is important. Do we know why this has to be the order, and can we add that into the comments here? It is not obvious with the code.

ZacBlanco: I'm still trying to wrap my head around _why_ this even works. Based on the code path the threads don't seem to be daemon threads, and the code path of the heartbeat threads don't indicate anywhere that they respond to thread interrupts.

I will add a comment once I understand why

ggezer: Do you have a consistent repro for this? 
I'd guess we'd still hit this issue without `AbstractClient` bailing out on interrupt.

ggezer: Looks like `BlockMasterSync`'s heartbeat could be stuck if close here is not reversed?

ZacBlanco: I have not been able to repro it after only making this change.

Without this PR, the consistent repro is

```console
$ ./bin/alluxio-start.sh master & ./bin/alluxio-start.sh worker & wait
$ ./bin/alluxio-stop.sh master
$ ./bin/alluxio-stop.sh worker
```

If you want to grab a jstack during that time you can run:

```console
$ ./bin/alluxio-start.sh master & ./bin/alluxio-start.sh worker & wait
$ ./bin/alluxio-stop.sh master && sleep 5 && (./bin/alluxio-stop.sh worker & sleep 1 && jstack -l `jps | grep AlluxioWorker | awk '{print $1}'`)
```

ggezer: This looks very aggressive, in restarting the master and that strengthens my hypothesis of `BlockMasterSync` being the culprit. After all, it's the only one that actually closes something in its `#close()`

ggezer: Looks like this is caused by sharing `DefaultBlockWorkerClient::mFileSystemMasterClient` with `PinListSync`.

`AbstractClient`'s `close` and `connect` methods are synchronized. While `PinListSync` is trying to connect (to a master that is down), `DefaultBlockWorker` is trying to close the very same client as part of `mResourceCloser.close()` call. So shutdown will be blocked until the `PinListSync` times out.

It seems, shutting down the base-executor first bails `PinListSync` out and later call to `mResourceCloser.close()` doesn't get stuck. This means we don't need that bail-out condition on `AbstractClient` as it seems to handle interruptions somewhere else.

That being said, I believe the proper fix is to host `FileSystemClientPool` in `DefaultBlockWorker` instead of a single `FileSystemMasterClient`. So we can isolate `FileSystemMasterClient` resources between the `DefaultBlockWorker` and the `PinListSync`.

ZacBlanco: @ggezer I implemented your suggested changes in the previous comment by creating an FS client pool instead. While at this point clients weren't locked on each other. It seemed that they now _all_ hang trying to connect to the master.

https://pastebin.com/Nc6yVr2q


The worker block sync still blocked shutdown:

ggezer: I see. Closing the pools is blocked.

> "alluxio-process-shutdown-hook" #292 prio=5 os_prio=0 tid=0x00007f6e2c002800 nid=0xdd9f waiting for monitor entry [0x00007f724f8f8000]
>    java.lang.Thread.State: BLOCKED (on object monitor)
>     at alluxio.AbstractClient.close(AbstractClient.java:310)
>     - waiting to lock <0x00000006c04cf6a8> (a alluxio.worker.block.BlockMasterClient)
>     at com.google.common.io.Closer.close(Closer.java:214)
>     at alluxio.worker.block.BlockMasterClientPool.close(BlockMasterClientPool.java:55)
>     at com.google.common.io.Closer.close(Closer.java:214)
>     at alluxio.worker.block.DefaultBlockWorker.stop(DefaultBlockWorker.java:261)
>     at alluxio.Registry.stop(Registry.java:148)

ggezer: Even though it's quite disruptive I'm OK to swap the lines at this point as I can't see a way around `HeartbeatExecutor`'s limitations.

ggezer: Ideally, we'd like to implement your initial commit as an abstract utility to all heartbeat executors at some point.

gpang: @ZacBlanco Do we know why this order works, and the other order didn't?

ZacBlanco: I believe this ordering works because `super.stop` will issue a `shutdownNow` on the executor which automatically sends an interrupt to all executor threads.

With this ordering, no interrupts are sent and we try to manually stop/close the heartbeat threads. Since no interrupt was performed, the heartbeat thread continues trying to connect until timing out.

ggezer: @gpang without first interrupting the threads, `mResourceCloser.close()` gets blocked because `AbstractClient.close()` is synchronized with `AbstractClient.connect()` and `connect` was trying to connect until it times out.

gpang: Can we add this info as a comment, since this ordering is intentional and critical.

ZacBlanco: updated

ZacBlanco:
jenkins, test this please

gpang:

PR #11436, opened 2020-05-15
Fix termination conditions for LRFU

ggezer:
LRFU can generate sorted-fields with equal value.
This PR fixes bugs related to this edge case.


ggezer:
@calvinjia Currently we can't know how the `AlignTask` has finished with. I'll work on making tasks to return completion/back-off statuses. That will make it possible to test this as well.

calvinjia:

PR #11439, opened 2020-05-15
Improve alluxio client local cache

apc999:
Introduce two changes:
(1) cleanup intermediate state when fail to add a new page
(2) make sure unchecked exception not esculated

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/NoExceptionCacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/NoExceptionCacheManager.java
@@ -1,0 +1,108 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.client.file.cache;
+
+import alluxio.metrics.MetricKey;
+import alluxio.metrics.MetricsSystem;
+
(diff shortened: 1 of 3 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/NoExceptionCacheManager.java
```diff
@@ -0,0 +1,96 @@
...
+import com.codahale.metrics.Counter;
+
+import java.nio.channels.ReadableByteChannel;
+
+import javax.annotation.Nullable;
+
+/**
+ * A wrapper class of CacheManager without throwing unchecked exceptions.
+ */
+public class NoExceptionCacheManager implements CacheManager {
```
calvinjia, line 26: Should this catch throwable, and should we log when we catch?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -76,8 +76,14 @@ public void put(PageId pageId, byte[] page) throws IOException {
...
       Files.createFile(p);
     }
-    try (FileOutputStream fos = new FileOutputStream(p.toFile(), false)) {
-      fos.write(page);
+    try {
+      // extra try to ensure output stream is closed
+      try (FileOutputStream fos = new FileOutputStream(p.toFile(), false)) {
+        fos.write(page);
+      }
+    } catch (IOException e) {
```
calvinjia, line 84: Should we catch throwable here?

apc999: changed to `Exception`. If `Error` happens, that's something really bad and we should escalate.

core/client/fs/src/main/java/alluxio/client/file/cache/NoExceptionCacheManager.java
apc999: add log. but IMO  `Error` should be escalated as these are not something we should recover.

calvinjia:

calvinjia:

PR #11442, opened 2020-05-16
Add more traces for worker management tasks

ggezer:


ggezer:
jenkins, test this please

core/common/src/main/java/alluxio/worker/block/BlockStoreLocation.java
```diff
@@ -150,25 +152,17 @@ public boolean belongsTo(BlockStoreLocation location) {
    */
   @Override
   public String toString() {
-    StringBuilder result = new StringBuilder();
-    if (mDirIndex == ANY_DIR) {
-      result.append("any dir");
-    } else {
-      result.append("dir ").append(mDirIndex);
+    MoreObjects.ToStringHelper strHelper = MoreObjects.toStringHelper(this);
```
calvinjia, line 155: Do we not want to show the `ANY` information anymore?

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
```diff
@@ -116,7 +116,8 @@ private void blockUpdated(long blockId, BlockStoreLocation location) {
     }
 
     if (LOG.isDebugEnabled()) {
-      LOG.debug("Block:{} updated at location:{}. Size:{}", blockId, location, sortedSet.size());
+      LOG.debug("Block:{} updated at {} with {} blocks.",
```
calvinjia, line 119: Can this fit on one line? (applies in general)

core/server/worker/src/main/java/alluxio/worker/block/management/tier/AlignTask.java
```diff
@@ -143,6 +163,8 @@ public void run() {
           blockLocPairListDst.get(i).getFirst()));
     }
 
+    LOG.debug("Generated {} swap transfers: \n ->{}", transferInfos.size(),
```
calvinjia, line 166: Do we need to wrap in `isDebugEnabled`? Also, based on our logging framework, will just using `debug` be sufficient to avoid the expensive portions for most of these?

ggezer: Yes, otherwise the parameters will be evaluated before making the decision to print.

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
ggezer: It usually leaves the last or two on the second line, that's why I moved all to the same line.

core/common/src/main/java/alluxio/worker/block/BlockStoreLocation.java
ggezer: It's implicit by not having it shown.
I couldn't find any justification to generate a human-readable string.
However this helps a great deal during investigations.

core/server/worker/src/main/java/alluxio/worker/block/annotator/DefaultBlockIterator.java
ggezer: This and one other fit indeed. However I did this mostly when the last or two parameter spilled to the second line.

core/server/worker/src/main/java/alluxio/worker/block/management/tier/AlignTask.java
ggezer: I also validated this on debugger.

core/common/src/main/java/alluxio/worker/block/BlockStoreLocation.java
calvinjia: Right, I would prefer explicit for easier debugging.

core/server/worker/src/main/java/alluxio/worker/block/management/tier/AlignTask.java
calvinjia: This message is not wrapped in `isDebugEnabled`?

ggezer: Oh. right.

core/common/src/main/java/alluxio/worker/block/BlockStoreLocation.java
ggezer: Made it explicit, but with the new string helper format.

calvinjia:
jenkins, test this please.

ggezer:
alluxio-bot, test this please

ggezer:

PR #11449, opened 2020-05-18
Improve incremental active sync

ZacBlanco:
This PR improves a number of items over the previous incremental sync implementation.

- The ActiveSyncer heartbeat thread no longer performs incremental syncs synchronously within the heartbeat. A sync is scheduled as a future on an executor during the heartbeat and the status is checked in the following heartbeat.

- The DFSInotifyStream returns EventBatches that the previous implementation assumed contains a significant number of events per batch (on the order of thousands). This is not the case. The EventBatch object contains only events pertaining to a specific transaction ID on the HDFS namenode. This equates to usually 1-2 events per EventBatch. The incremental sync loop now processes with the correct assumption. 

- The incremental sync loop no longer processes events in the same thread which polls from the DFSInotifyEventStream. Events are polled and then sent off to be processed in an executor. This increases the theoretical maximum throughput from 1 single thread the N threads configured.


ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

gpang:
@ZacBlanco could you add a brief PR description?

ZacBlanco:
Updated

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
```diff
@@ -61,11 +68,16 @@ public ActiveSyncer(FileSystemMaster fileSystemMaster, ActiveSyncManager syncMan
     mSyncManager = syncManager;
     mMountId = mountId;
     mMountTable = mountTable;
+    mSyncTasks = new LinkedBlockingQueue<>(32);
   }
 
   @Override
   public void heartbeat() {
     LOG.debug("start Active Syncer heartbeat");
```
gpang, line 76: NIT: can we add the mount id, and maybe the mount path in this message?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
```diff
@@ -61,11 +68,16 @@ public ActiveSyncer(FileSystemMaster fileSystemMaster, ActiveSyncManager syncMan
...
+    // Remove any previously completed sync tasks
+    while (mSyncTasks.peek() != null && !mSyncTasks.peek().isDone()) {
```
gpang, line 78: should this be `mSyncTasks.peek().isDone()`? Don't we want to remove the ones that are considered done?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
```diff
@@ -74,29 +86,47 @@ public void heartbeat() {
...
+            tasksPerSync.add(CompletableFuture.supplyAsync(() -> {
               processSyncPoint(ufsUri, syncInfo);
-              return null;
-            });
+              return syncInfo.getTxId();
+            }, mSyncManager.getExecutor()));
           }
-          mSyncManager.getExecutor().invokeAll(tasksPerSyncPoint);
           // Journal the latest processed txId
-          mFileSystemMaster.recordActiveSyncTxid(syncInfo.getTxId(), mMountId);
```
gpang, line 94: Was this wrong before, journaling too early?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
```diff
@@ -74,29 +86,47 @@ public void heartbeat() {
...
+          CompletableFuture<Void> syncTask =
+              CompletableFuture.allOf(tasksPerSync.toArray(new CompletableFuture<?>[0]))
+              .thenRunAsync(() -> mFileSystemMaster
+                      .recordActiveSyncTxid(syncInfo.getTxId(), mMountId),
+                  mSyncManager.getExecutor());
+          while (!mSyncTasks.offer(syncTask)) {
```
gpang, line 110: If the first task in the queue is slow for some reason, would that hold up all the rest of the processing, even if the rest of the tasks have completed?

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/activesync/SupportedHdfsActiveSyncProvider.java
```diff
@@ -258,38 +273,52 @@ public void stopSync(AlluxioURI ufsUri) {
...
-          count += batch.getEvents().length;
-          for (Event event : batch.getEvents()) {
-            processEvent(event, mUfsUriList, txId);
+        List<Callable<Integer>> process = new LinkedList<>();
+        for (int i = 0; i < mBatchSize; i++) {
+          EventBatch batch = eventStream.poll(mActiveUfsPollTimeoutMs, TimeUnit.MILLISECONDS);
+          if (batch == null) {
+            break;
           }
+          EventBatch finalBatch = batch;
```
gpang, line 290: Is this supposed to be `final`? Can the previous `EventBatch batch = ...` just be `final`?

bin/alluxio
```diff
@@ -149,7 +149,7 @@ function runJavaClass {
               CLASS_ARGS+=("${arg}")
       esac
   done
-  "${JAVA}" -cp ${CLASSPATH} ${ALLUXIO_USER_JAVA_OPTS} ${ALLUXIO_SHELL_JAVA_OPTS} ${CLASS} ${PARAMETER} "${CLASS_ARGS[@]}"
+  exec "${JAVA}" -cp ${CLASSPATH} ${ALLUXIO_USER_JAVA_OPTS} ${ALLUXIO_SHELL_JAVA_OPTS} ${CLASS} ${PARAMETER} "${CLASS_ARGS[@]}"
```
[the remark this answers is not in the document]
ZacBlanco: Oh I should make this a separate PR. I'll remove it.

I think technically it is a bug. Without using `exec` a `ctrl-c` doesn't actually kill the java process running if it's a CLI program. Rather it just kills the shell running the CLI command, so a haning CLI command i.e because of a large listStatus, or master unavailable doesn't actually get killed if the USER ctrl-Cs and it may continue to run in the background.

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
```diff
@@ -342,10 +348,6 @@ public void stopSyncAndJournal(RpcContext rpcContext, AlluxioURI syncPoint)
     }
     try (LockResource r = new LockResource(mLock)) {
       MountTable.Resolution resolution = mMountTable.resolve(syncPoint);
-      if (resolution == null) {
```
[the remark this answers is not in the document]
ZacBlanco: `resolve` is not nullable, so `resolution` is never null

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
ZacBlanco: sure

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
ZacBlanco: no, it worked properly before because the sync occurred synchronously within the heartbeat

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
ZacBlanco: its possible that it holds up removing tasks from the queue, though for this to happen is would require filling the incremental sync queue which right now holds up to 32 items. Given the standard heartbeat interval of 30 seconds that means this would need to take 16 minutes to finish the incremental sync.

I can try improving this by using a set to iterate over all of the elements and removing any that are finished, but would require iterating over all futures. I think for the size of the futures added to this queue that is a reasonable solution

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -384,12 +384,18 @@
...
 
-  final ThreadPoolExecutor mMetadataSyncExecutor = new ThreadPoolExecutor(
+  final ThreadPoolExecutor mSyncMetadataExecutor = new ThreadPoolExecutor(
       ServerConfiguration.getInt(PropertyKey.MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE),
       ServerConfiguration.getInt(PropertyKey.MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE),
       1, TimeUnit.MINUTES, new LinkedBlockingQueue<>(),
       ThreadFactoryUtils.build("alluxio-ufs-sync-%d", false));
 
+  final ThreadPoolExecutor mActiveSyncMetadataExecutor = new ThreadPoolExecutor(
+      ServerConfiguration.getInt(PropertyKey.MASTER_METADATA_SYNC_EXECUTOR_POOL_SIZE),
```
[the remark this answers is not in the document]
ZacBlanco: no

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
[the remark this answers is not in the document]
ZacBlanco: Also, if you inspect the code, we either throw an exception or always return a non-null object

bin/alluxio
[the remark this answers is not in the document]
ZacBlanco: reverted

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1973,7 +1973,8 @@ public String toString() {
           .build();
   public static final PropertyKey MASTER_UFS_ACTIVE_SYNC_THREAD_POOL_SIZE =
       new Builder(Name.MASTER_UFS_ACTIVE_SYNC_THREAD_POOL_SIZE)
-          .setDefaultValue("3")
+          .setDefaultSupplier(() -> Math.max(2, Runtime.getRuntime().availableProcessors() / 2),
+              "The number of threads used by the active sync provider process active sync events")
           .setDescription("Max number of threads used to perform active sync")
```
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
ZacBlanco: updated

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/activesync/SupportedHdfsActiveSyncProvider.java
ZacBlanco: updated

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -132,6 +132,37 @@ public UfsStatus getStatus(AlluxioURI path) {
...
+   * @param mountTable the Alluxio mount table
+   * @return The corresponding {@link UfsStatus} or {@code null} if there is none stored
+   */
+  @Nullable
+  public UfsStatus fetchStatusIfAbsent(AlluxioURI path, MountTable mountTable)
+      throws InvalidPathException {
+    UfsStatus status = mStatuses.get(path);
+    if (status != null) {
+      return status;
+    }
```
yuzhu, line 148: you might want to use `computeIfAbsent` or `putIfAbsent` to implement this?  otherwise k concurrent calls to this function with the same path might result in k ufs calls  Using one of those methods ensures only one call to the UFS is made.

ZacBlanco: Unfortunately I can't with the current implementation or else it will result in deadlock. This function calls `addStatus` which actually does the insert into the map. Calling `ConcurrentHashMap#put` while within a `computeIfAbsent` function will deadlock

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3125,9 +3139,13 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
...
                 FileSystemMasterCommonPOptions.newBuilder().setSyncIntervalMs(0).build();
-            try {
-              syncMetadata(rpcContext, changedFile, options, DescendantType.ONE, null, null, null);
-            } catch (InvalidPathException | AccessControlException e) {
+            LockingScheme scheme = createSyncLockingScheme(changedFile, options, false);
+            try (LockedInodePath inodePath = mInodeTree.lockInodePath(scheme)) {
+              InodeSyncStream sync = new InodeSyncStream(inodePath, mActiveSyncMetadataExecutor,
+                  this, mInodeTree, mInodeStore, mInodeLockManager, mMountTable, rpcContext,
+                  DescendantType.NONE, mUfsSyncPathCache, options, false, false, false);
+              sync.sync();
```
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3103,19 +3118,18 @@ public void activeSyncMetadata(AlluxioURI path, Collection<AlluxioURI> changedFi
...
-          syncMetadata(rpcContext, path, options, DescendantType.ALL, null, null, null);
-        } catch (AccessControlException e) {
-          // This shouldn never happen because the permission check function is passed as null.
-          LOG.error("Active sync full scan failed on {}", path, e);
+        LockingScheme scheme = createSyncLockingScheme(path, options, false);
+        try (LockedInodePath inodePath = mInodeTree.lockInodePath(scheme)) {
+          InodeSyncStream sync = new InodeSyncStream(inodePath, mActiveSyncMetadataExecutor, this,
+              mInodeTree, mInodeStore, mInodeLockManager, mMountTable, rpcContext,
+              DescendantType.ALL, mUfsSyncPathCache, options, false, false, false);
+          sync.sync();
```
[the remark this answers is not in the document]
ZacBlanco: updates

ZacBlanco:
alluxio-bot, test this please

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
yuzhu: instead of calling `addStatus` perhaps you could just return the right value, which will result in the values being inserted into the map by `computeIfAbsent` right?

ZacBlanco: I use `addStatus` because it performs a check to make sure the status has the correct naming scheme.

ZacBlanco: I guess I can refactor out the check and put it elsewhere. It just seems more error prone, because then statuses could still be inserted without that check performed

yuzhu: i see, sounds good to me

ZacBlanco:

PR #11451, opened 2020-05-19
Prevent secondary UFS journal from modifying journal files

gpang:


core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -281,17 +242,6 @@ public synchronized void awaitLosePrimacy() throws IOException {
...
-    mAsyncWriter = null;
-=======
-
-    // Close async writer first to flush pending entries.
-    mAsyncWriter.close();
-    mAsyncWriter = null;
->>>>>>> b358b1a6a3... Prevent secondary UFS journal from modifying journal files
     mWriter.close();
     mWriter = null;
     mAsyncWriter = null;
```
ggezer: Is `mAsyncWrite.close()` missing ?

gpang: It is missing because in 1.8, it never called `mAsyncWriter.close()`. In 1.8, `mAsyncWriter` doesn't even have a `close()` method.

gpang:

PR #11456, opened 2020-05-20
Clean up state on failed PUT

apc999:
When `LocalPageStore` failed to put a page due to reasons like disk full, the state was not cleaned in `LocalCacheManager`. This PR addes
- Logic to clean up the state on failed page put in meta store
- Logic to clean up a page on failed page get
- Tests

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
@@ -277,6 +277,14 @@ public class LocalCacheManager implements CacheManager {
       }
       if (enoughSpace) {
         boolean ret = addPage(pageId, page);
+        if (!ret) {
+          // something is wrong to add this page, let's remove it from meta store
+          try (LockResource r2 = new LockResource(mMetaLock.writeLock())) {
+            mMetaStore.removePage(pageId);
+          } catch (PageNotFoundException e) {
+            // best effort to remove this page from meta store and ignore the exception
+          }
+        }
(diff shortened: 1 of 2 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -343,6 +359,14 @@ public ReadableByteChannel get(PageId pageId, int pageOffset) {
...
       }
       ReadableByteChannel ret = getPage(pageId, pageOffset);
+      if (ret == null) {
+        // something is wrong to read this page, let's remove it from meta store
+        try (LockResource r2 = new LockResource(mMetaLock.writeLock())) {
+          mMetaStore.removePage(pageId);
+        } catch (PageNotFoundException e) {
+          // best effort to remove this page from meta store and ignore the exception
+        }
+      }
```
calvinjia, line 369: Consider putting this logic into a method?

jainxrohit: Yeah I would have thought of this an integral logic of getPage method.

apc999: I considered that too initially. but one reason I still leave this in `get()` is to make all locking related code in one method --- in here we are acquiring `mMetaLock`. It can be harder to track if somewhere else also acquire locks

apc999:
Jenkins, test this please

apc999:

apc999:

PR #11457, opened 2020-05-21
Fix dependencies

### LINKED ISSUE #9403: NoSuchMethodError when use filesystem api
**Alluxio Version:**
Alluxio-2.0

**Describe the bug**

when we use this filesytem api to access alluxio cluster
```
  val conf: InstancedConfiguration = new InstancedConfiguration(ConfigurationUtils.defaults())
  System.out.println(masterAddress.getHost() + ' ' + masterAddress.getPort())
  conf.set(PropertyKey.MASTER_HOSTNAME, masterAddress.getHost())
  conf.set(PropertyKey.MASTER_RPC_PORT, Integer.toString(masterAddress.getPort()))
  val fs = FileSystem.Factory.create(conf)
```
we will meet exception below:
```
Exception in thread "main" java.lang.NoSuchMethodError: org.apache.commons.codec.binary.Hex.encodeHexString([B)Ljava/lang/String;
	at alluxio.conf.Hash.compute(Hash.java:59)
	at alluxio.conf.Hash.get(Hash.java:73)
	at alluxio.conf.AlluxioProperties.hash(AlluxioProperties.java:271)
	at alluxio.conf.InstancedConfiguration.hash(InstancedConfiguration.java:421)
	at alluxio.ClientContext.<init>(ClientContext.java:99)
	at alluxio.ClientContext.create(ClientContext.java:61)
	at alluxio.client.file.FileSystemContext.create(FileSystemContext.java:166)
	at alluxio.client.file.FileSystemContext.create(FileSystemContext.java:155)
	at alluxio.client.file.FileSystem$Factory.create(FileSystem.java:117)
	at AlluxioUtil.<init>(AlluxioUtil.scala:22)
```
this is because in maven repo dependencies:

alluxio-core-client-fs:2.0.0 -->  alluxio-core-common:2.0.0 --> commons.httpclient:3.1 --> 
commons-codec:1.2

while in commons-codec:1.2,  Hex do not have  encodeHexString function.
in alluxio source code , we find that commons-codec's version is 1.10  in pom.xml

yuzhu:
fixes #9403 again
 commons-codec:1.2, which is included as part of the dependency chain, does not have the method that we want. need to explicitly include a newer version

### DIFF
--- a/core/client/fs/pom.xml
+++ b/core/client/fs/pom.xml
@@ -40,6 +40,10 @@
       <artifactId>protobuf-java</artifactId>
     </dependency>
     <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+    </dependency>
+    <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
@@ -49,11 +53,19 @@
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
       <artifactId>grpc-core</artifactId>
     </dependency>
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
     </dependency>
     <dependency>
       <groupId>org.rocksdb</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -310,6 +310,11 @@
       </dependency>
       <dependency>
         <groupId>io.grpc</groupId>
+        <artifactId>grpc-api</artifactId>
+        <version>${grpc.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
         <artifactId>grpc-core</artifactId>
         <version>${grpc.version}</version>
       </dependency>


yuzhu:
@apc999 PTAL, reported by the community to be a regression.

ZacBlanco:
I think we should also change the dependency versions so that we don't have two versions of a dependency which exist in the dependency tree, otherwise it will be possible for this to happen again

yuzhu:
@ZacBlanco the problem is we are already using the latest version of commons.httpclient , it just happens to have a dependency on an old version of commons-codec.

yuzhu:
@ZacBlanco @apc999 ping?

core/client/fs/pom.xml
```diff
@@ -39,6 +39,10 @@
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
     </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+    </dependency>
```
ZacBlanco, line 45: I understand the reasoning for adding this dependency, but why all of the others?

yuzhu: they are listed as used but not listed in pom, so I figured we should add them? probably ok to leave them out for now to ensure minimal change.

yuzhu: @ZacBlanco ping?

ZacBlanco: Can you update the description with more information? Otherwise LGTM

yuzhu:

PR #11466, opened 2020-05-22
Return stats from each block management task

ggezer:
This PR lays out the essential infrastructure for each management task to return stats in terms of how many block operations are executed/failed/backed-off.


ggezer:
jenkins, test this please

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
```diff
@@ -181,12 +181,19 @@ private void runManagement() {
         LOG.debug("Running task of type:{}", currentTask.getClass().getSimpleName());
         // Run the current task on coordinator thread.
         try {
-          currentTask.run();
+          BlockManagementTaskResult result = currentTask.run();
+          LOG.debug("Management task: {} finished with result: {}",
+              currentTask.getClass().getSimpleName(), result);
```
ggezer, line 186: @calvinjia 
What do you think about making this log INFO level?

At the worst this would print every 10 seconds when;
- BackoffStrategy is DIRECTORY
- Tiers don't align
- AlignTask can't make progress (due to failures or heavy load)

I prefer this over crafting an admin report because the latter can't provide information for a full range without persistence at worker/master. I think this would be sufficient until at least we stabilize worker management infra.

ggezer:
Failure seems unrelated: `validateCheckpointInClusterRestart`

core/server/worker/src/main/java/alluxio/worker/block/management/BlockManagementTaskResult.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Holds results of a {@link BlockManagementTask}.
+ *
+ * It contains sub-results for each {@link BlockOperationType} that is issued
+ * by the magement task.
+ */
+public class BlockManagementTaskResult {
+  /** Map of block-operation results. */
+  private Hashtable<BlockOperationType, BlockOperationResult> mBlockOpResults;
```
calvinjia, line 27: Use a map here?

core/server/worker/src/main/java/alluxio/worker/block/management/BlockManagementTaskResult.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Hashtable;
+import java.util.Map;
+
+/**
+ * Holds results of a {@link BlockManagementTask}.
+ *
+ * It contains sub-results for each {@link BlockOperationType} that is issued
+ * by the magement task.
+ */
+public class BlockManagementTaskResult {
```
calvinjia, line 25: Add `@NotThreadSafe`?

core/server/worker/src/main/java/alluxio/worker/block/management/ManagementTaskCoordinator.java
calvinjia: I think that's fine, we can also add a template to raise the log level of this class if it is too noisy.

ggezer: Cool. I don't think it will be noisy. I will watch for a while though.

ggezer:
jenkins, test this please

core/server/worker/src/main/java/alluxio/worker/block/management/BlockManagementTaskResult.java
```diff
@@ -0,0 +1,86 @@
...
+@NotThreadSafe
+public class BlockManagementTaskResult {
+  /** Map of block-operation results. */
+  private Map<BlockOperationType, BlockOperationResult> mBlockOpResults;
+
+  /**
+   * Creates a new empty task result.
+   */
+  public BlockManagementTaskResult() {
+    mBlockOpResults = new Hashtable<>();
```
calvinjia, line 35: Any reason we want to use `HashTable` here?

ggezer: no reason. just accidents. fixed.

calvinjia:
jenkins, test this please.

PR #11468, opened 2020-05-23
Support alluxio.tar.gz in alluxio-emr.sh emr bootstrap

LuQQiu:
Support users providing tarball with name `alluxio.tar.gz` for alluxio-emr.sh

### DIFF
--- a/integration/emr/alluxio-emr.sh
+++ b/integration/emr/alluxio-emr.sh
@@ -143,9 +143,12 @@ emr_install_alluxio() {
   download_file "${alluxio_tarball}"
 
   release=$(basename "${alluxio_tarball}")
-  local release_unzip
-  release_unzip="${release%%-*}" # trims everything after the first '-', ex. alluxio-foo-bar-whatever -> alluxio
-
+  local release_prefix
+  if [[ "${release}" == *-* ]]; then
+    release_unzip="${release%%-*}" # trims everything after the first '-', ex. alluxio-foo-bar-whatever -> alluxio
+  else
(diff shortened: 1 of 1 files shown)


integration/emr/alluxio-emr.sh
```diff
@@ -143,9 +143,12 @@ emr_install_alluxio() {
   download_file "${alluxio_tarball}"
 
   release=$(basename "${alluxio_tarball}")
-  local release_unzip
-  release_unzip="${release%%-*}" # trims everything after the first '-', ex. alluxio-foo-bar-whatever -> alluxio
-
+  local release_prefix
```
Xenorith, line 146: should this be release_unzip?

integration/emr/alluxio-emr.sh
```diff
@@ -143,9 +143,12 @@ emr_install_alluxio() {
...
+  if [[ "${release}" == *-* ]]; then
+    release_unzip="${release%%-*}" # trims everything after the first '-', ex. alluxio-foo-bar-whatever -> alluxio
```
Xenorith, line 148: both of these expressions seem very fragile to me :/
the original expression of converting alluxio-* -> alluxio seems unnecessarily difficult

is it not possible to trim the ".tar.gz" suffix from the tarball and use that as the directory name that the tarball unzips to?

LuQQiu: emmm... the problem is the unzipped folder from  <value>.tar.gz doesn't name <name>

LuQQiu:

PR #11473, opened 2020-05-25
Fix fuse restart failed

### LINKED ISSUE #11472: Fuse failed to restart when it's killed
**Alluxio Version:**
2.2.1

**Describe the bug**
I use helm to install the latest chart, and kill the process of alluxio-fuse. And I'm able to see the alluxio-fuse pod is not able to restart automatically.

The error is 

```
kubectl  describe po alluxio-fuse-cgwgj
 Warning  Failed            0s (x3 over 22s)   kubelet, cn-huhehaote.192.168.0.118  Error: failed to start container "alluxio-fuse": Error response from daemon: error while creating mount source path '/alluxio-fuse': mkdir /alluxio-fuse: file exists
```

The reason is that if the hostpath`/alluxio-fuse` is a bad mount path due to the process alluxio-fuse doesn't exit correctly. 

```
[root@iZhp3bku0ru8vuxq08lorvZ ~]# mkdir -p /alluxio-fuse
mkdir: cannot create directory ‘/alluxio-fuse’: File exists
[root@iZhp3bku0ru8vuxq08lorvZ ~]# mount |grep alluxio-fuse
alluxio-fuse on /alluxio-fuse type fuse.alluxio-fuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0)
[root@iZhp3bku0ru8vuxq08lorvZ ~]# cd /alluxio-fuse
-bash: cd: /alluxio-fuse: Transport endpoint is not connected
```

**To Reproduce**

**Expected behavior**

**Urgency**

**Additional context**

cheyang:
fix #11472

### DIFF
--- a/integration/docker/entrypoint.sh
+++ b/integration/docker/entrypoint.sh
@@ -15,6 +15,7 @@ set -e
 ALLUXIO_HOME="/opt/alluxio"
 NO_FORMAT='--no-format'
 FUSE_OPTS='--fuse-opts'
+MOUNT_POINT='/mnt/alluxio-fuse'
 
 # List of environment variables which go in alluxio-env.sh instead of
 # alluxio-site.properties
@@ -89,8 +90,10 @@ function mountAlluxioRootFSWithFuseOption {
   fi
 
   # Unmount first if cleanup failed and ignore error
-  ! integration/fuse/bin/alluxio-fuse unmount /alluxio-fuse
-  exec integration/fuse/bin/alluxio-fuse mount -n ${fuseOptions} /alluxio-fuse /
+  ! mkdir -p ${MOUNT_POINT}
+  ! umount ${MOUNT_POINT}
+  #! integration/fuse/bin/alluxio-fuse unmount ${MOUNT_POINT}
+  exec integration/fuse/bin/alluxio-fuse mount -n ${fuseOptions} ${MOUNT_POINT} /
 }
 
 # Sends a signal to each of the running background processes
--- a/integration/kubernetes/helm-chart/alluxio/Chart.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/Chart.yaml
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.5.9
+version: 0.6.0
 home: https://www.alluxio.io/
 maintainers:
 - name: Adit Madan
--- a/integration/kubernetes/helm-chart/alluxio/values.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/values.yaml
@@ -25,6 +25,9 @@ user: 1000
 group: 1000
 fsGroup: 1000
 
+# Mount path in the host
+mountPath: /mnt/alluxio-fuse
+
 # Site properties for all the components
 properties:
   # alluxio.user.metrics.collection.enabled: 'true'
--- a/integration/kubernetes/helm-chart/alluxio/values.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/values.yaml
@@ -25,8 +25,6 @@ user: 1000
 group: 1000
 fsGroup: 1000
 
-# Mount path in the host
-mountPath: /mnt/alluxio-fuse
 
 # Site properties for all the components
 properties:
@@ -256,6 +254,8 @@ fuse:
   args:
     - fuse
     - --fuse-opts=allow_other
+  # Mount path in the host
+  mountPath: /mnt/alluxio-fuse
   resources:
     requests:
       cpu: "0.5"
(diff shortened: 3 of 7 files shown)


integration/kubernetes/helm-chart/alluxio/Chart.yaml
```diff
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.5.9
+version: 0.6.0
```
apc999, line 15: not sure what's the convention here, is more common to have 0.5.10 or 0.6.0?
@cheyang @jiacheliu3

apc999:
@jiacheliu3 @madanadit  do you know if there is any documentation we should update here?

integration/docker/entrypoint.sh
```diff
@@ -89,8 +90,10 @@ function mountAlluxioRootFSWithFuseOption {
   fi
 
   # Unmount first if cleanup failed and ignore error
-  ! integration/fuse/bin/alluxio-fuse unmount /alluxio-fuse
-  exec integration/fuse/bin/alluxio-fuse mount -n ${fuseOptions} /alluxio-fuse /
+  ! mkdir -p ${MOUNT_POINT}
+  ! umount ${MOUNT_POINT}
+  #! integration/fuse/bin/alluxio-fuse unmount ${MOUNT_POINT}
+  exec integration/fuse/bin/alluxio-fuse mount -n ${fuseOptions} ${MOUNT_POINT} /
```
Binyang2014, line 96: I'm not sure if this change will fix this issue.
I encounter same error when alluxio-fuse daemon is kill by SIGKILL. In this situation, the `preStop` hook will not be executed.
And when alluxio-fuse daemon trying to restart, k8s failed to mount the `alluxio-fuse-mount`, and report the error: `Transport endpoint is not connected`. I suspect we don't have a change to execute this script, container will be failed in start phrase.

One workaround is to use sidecar to monitor the this mount point. If the mount point is broken, umount it in sidecar. But I think the more official way is using PV/PVC to deploy alluxio in k8s. And let k8s to manage the storage life-cycle.

cheyang: This sould be also used with `mountPath: /mnt` in `integration/kubernetes/helm-chart/alluxio/values.yaml` The pod is not mounted with the alluxio fuse mount path, instead its parent directory.

integration/kubernetes/helm-chart/alluxio/Chart.yaml
jiacheliu3: We don't really have a convention on the versioning here. I'm fine with bumping to 0.6.0 as this is relatively stable now.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -25,6 +25,10 @@ user: 1000
 group: 1000
 fsGroup: 1000
 
+# Mount path in the host
+#mountPath: /mnt/alluxio-fuse
+mountPath: /mnt
```
madanadit, line 30: please include this path under the `fuse` section
```
fuse:
  mountPath: /mnt
```

integration/docker/entrypoint.sh
```diff
@@ -89,8 +90,10 @@ function mountAlluxioRootFSWithFuseOption {
   fi
 
   # Unmount first if cleanup failed and ignore error
-  ! integration/fuse/bin/alluxio-fuse unmount /alluxio-fuse
-  exec integration/fuse/bin/alluxio-fuse mount -n ${fuseOptions} /alluxio-fuse /
+  ! mkdir -p ${MOUNT_POINT}
+  ! umount ${MOUNT_POINT}
+  #! integration/fuse/bin/alluxio-fuse unmount ${MOUNT_POINT}
```
madanadit, line 95: as far as i remember, this command kills the fuse daemon and is also called in `preStop`.  is this still needed when the `preStop` hook is not executed for an unclean shutdown?

cheyang: It's needed for the case `Transport endpoint is not connected` when the fuse daemon is killed for OOM issue.  We've tested it in customer scenario.

integration/kubernetes/helm-chart/alluxio/values.yaml
cheyang: Sure.

cheyang:
> Thanks @cheyang for the fix. Could you also update docs under this section with the new parameter: https://docs.alluxio.io/os/user/edge/en/deploy/Running-Alluxio-On-Kubernetes.html#posix-api

Done. Please take a look.

apc999:

apc999:

apc999:
@madanadit  do you have a good sense if this should be cherry-picked to 2.2 branch?

PR #11480, opened 2020-05-27
Add an HDFS validation tool

jiacheliu3:
This tool validates all aspects for an HDFS path. It serves as a pre-mount check for both root mount and nested mount.


jiacheliu3:
@LuQQiu @madanadit This tool is mostly completed and tested. There are still a couple of TODOs and I'm working on resolving them. Also I'm working on manually testing in real HDFS deployments and will be tweaking this while you are reviewing.

Please start with looking at the structure and workflow, how each test is invoked and wrapped and let me know if you have suggestions. On the individual tests, I will add markers on improvements that I can still add tomorrow.

Thanks!

core/common/src/main/java/alluxio/cli/JsonSerializable.java
```diff
@@ -9,14 +9,16 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.stress;
+package alluxio.cli;
```
jiacheliu3, line 12: Moved to a more generalized path

core/common/src/main/java/alluxio/cli/JsonSerializable.java
```diff
@@ -51,6 +53,12 @@ default String toJson() throws JsonProcessingException {
     return objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(this);
   }
 
+  static <T extends JsonSerializable> String listToJson(List<T> list) throws JsonProcessingException {
+    ObjectMapper objectMapper = new ObjectMapper();
+    objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+    return objectMapper.writeValueAsString(list);
+  }
```
jiacheliu3, line 60: Is there a better way for this kind of wrapper methods?

core/common/src/main/java/alluxio/security/authentication/ImpersonationAuthenticator.java
```diff
@@ -146,4 +148,12 @@ public void authenticate(String connectionUser, String impersonationUser)
...
         RuntimeConstants.ALLUXIO_SECURITY_DOCS_URL));
   }
+
+  public Map<String, Set<String>> getImpersonationUsers() {
+    return Collections.unmodifiableMap(mImpersonationUsers);
+  }
+
+  public Map<String, Set<String>> getmImpersonationGroups() {
+    return Collections.unmodifiableMap(mImpersonationGroups);
+  }
```
jiacheliu3, line 158: Setting these two to public and using this class to parse Alluxio impersonation properties

integration/tools/src/main/java/alluxio/cli/ValidateHms.java
```diff
@@ -154,7 +154,7 @@ private boolean checkHmsUri(String uriAddress) throws Exception {
       uri = new URI(uriAddress);
     } catch (Throwable t) {
       mResults.computeIfAbsent(State.FAILED, k -> new ArrayList<>()).add(
-          new TaskResult(State.FAILED, "HmsUrisSyntaxCheck", getErrorInfo(t),
+          new TaskResult(State.FAILED, "HmsUrisSyntaxCheck", ValidateUtils.getErrorInfo(t),
```
jiacheliu3, line 157: Moved to a util class

integration/tools/src/main/java/alluxio/cli/ValidateUtils.java
```diff
@@ -37,7 +41,7 @@ private ValidateUtils() {} // prevent instantiation
   /**
    * Represents the result of a given task.
    */
-  public static class TaskResult implements Serializable {
+  public static class TaskResult implements JsonSerializable {
```
jiacheliu3, line 44: Using JsonSerializable to do serialization.

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -52,14 +60,31 @@
...
+  /**
+   * Initiate the tests for a specific UFS path and UFS configs.
+   *
+   * @param path the UFS path
+   * @param conf the UFs configurations
+   * */
+  public UnderFileSystemContractTest(String path, InstancedConfiguration conf) {
+    mUfsPath = path;
+    mConf = conf;
+  }
```
jiacheliu3, line 82: Updated this test to be able to validate a nested mount path

shell/src/main/java/alluxio/cli/ValidateEnv.java
```diff
@@ -87,80 +93,108 @@
   private static final List<ValidationTask> MASTER_TASKS = new ArrayList<>();
   private static final List<ValidationTask> WORKER_TASKS = new ArrayList<>();
 
-  private static final AlluxioConfiguration CONF;
+  private final AlluxioConfiguration mConf;
+  private final String mPath;
 
-  static {
-    CONF = InstancedConfiguration.defaults();
```
jiacheliu3, line 93: Use a constructor now instead of static block, as the validation can be against a nested mount.

shell/src/main/java/alluxio/cli/ValidateHdfsMount.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  /**
+   * The entrance.
+   *
+   * @param args command line arguments
+   * */
+  public static void main(String[] args) throws Exception {
+    // TODO(jiacheng): use jccommand?
```
jiacheliu3, line 102: After some tests I find JCommand to be better. Will refactor this to use JCommand.

shell/src/main/java/alluxio/cli/validation/HadoopConfigurationFileParser.java
```diff
@@ -87,4 +87,32 @@ public HadoopConfigurationFileParser() {}
     }
     return ret;
   }
+
+  // TODO(jiacheng): keep only one
```
jiacheliu3, line 91: This will be resolved

shell/src/main/java/alluxio/cli/validation/HdfsConfParityValidationTask.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Compares HDFS configuration in Alluxio and in HDFS environment variables.
+ * */
+@ApplicableUfsType(ApplicableUfsType.Type.HDFS)
+public class HdfsConfParityValidationTask extends HdfsConfValidationTask {
```
jiacheliu3, line 17: This is the previous `alluxio.cli.validation.HdfsValidationTask` where the hdfs configs defined in Alluxio property are compared against the hadoop config defined in hadoop env variables.

shell/src/main/java/alluxio/cli/validation/HdfsImpersonationValidationTask.java
```diff
@@ -0,0 +1,264 @@
...
+      return new ValidateUtils.TaskResult(ValidateUtils.State.SKIPPED, getName(),
+              mMsg.toString(), mAdvice.toString());
+    }
+
+    ValidateUtils.TaskResult loadConfig = loadHdfsConfig();
+    if (loadConfig.getState() != ValidateUtils.State.OK) {
+      return loadConfig;
+    }
+
+    // TODO(jiacheng): do we want to check the current user even if
```
jiacheliu3, line 242: Do we want to assume the current running user needs impersonation? I don't think so.

shell/src/main/java/alluxio/cli/validation/HdfsImpersonationValidationTask.java
```diff
@@ -0,0 +1,264 @@
...
+    //  there's no impersonation setting?
+    switch (mMode) {
+      case USERS:
+        return validateImpersonationUsers();
+      case GROUPS:
+        return validateImpersonationGroups();
+      case HOSTS:
+        return validateImpersonationHosts();
```
jiacheliu3, line 250: One limitation of the current check is, we can only know which users need impersonation based on the Alluxio properties. If the user does not define `alluxio.security.impersonation.bob.users` field, there's no way to know if bob needs to be validated.

jiacheliu3: @LuQQiu One slightly better way is to define a list of users that we "think" should have impersonation, like "presto, yarn, spark", and give a warning if they are not configured. But then the drawback is, we will subsequently complain if those users are not configured in hadoop core-site.xml, which may be confusing to those who don't need these users.

shell/src/main/java/alluxio/cli/validation/HdfsVersionValidationTask.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Override
+  public String getName() {
+    return "ValidateHdfsVersion";
+  }
+
+  protected String parseVersion(String output) {
+    // An example output from "hadoop version" command:
+    //    Hadoop 2.7.2
```
jiacheliu3, line 37: Need to double check if the same format holds true in CDH.

shell/src/main/java/alluxio/cli/validation/NativeLibValidationTask.java
```diff
@@ -0,0 +1,58 @@
...
+
+import alluxio.cli.ValidateUtils;
+import alluxio.conf.AlluxioConfiguration;
+
+import java.io.File;
+import java.util.Map;
+import java.util.StringTokenizer;
+
+/**
+ * Validates the Java native libs defined in the system property.
```
jiacheliu3, line 11: Hadoop native lib may be defined here.

jiacheliu3: But then the drawback is for the default value for your JVM, there's a good chance you don't have them all in your env and you always see this warning.......

shell/src/main/java/alluxio/cli/validation/SecureHdfsValidationTask.java
```diff
@@ -27,7 +27,8 @@
 /**
  * Task for validating security configurations.
  */
-public final class SecureHdfsValidationTask extends HdfsValidationTask {
+@ApplicableUfsType(ApplicableUfsType.Type.HDFS)
+public final class SecureHdfsValidationTask extends AbstractValidationTask {
```
jiacheliu3, line 31: Partially updated classes like this are from `validateEnv`.

shell/src/main/java/alluxio/cli/validation/UfsDirectoryValidationTask.java
```diff
@@ -30,27 +31,41 @@
...
+  @Override
+  public ValidateUtils.TaskResult validate(Map<String, String> optionsMap) {
+    StringBuilder msg = new StringBuilder();
+    StringBuilder advice = new StringBuilder();
     try {
       UfsStatus[] listStatus = mUfs.listStatus(mPath);
       if (listStatus == null) {
-        System.err.format("Unable to list under file system path %s.%n", mPath);
-        return TaskResult.FAILED;
+        msg.append(String.format("Unable to list under file system path %s. ", mPath));
```
jiacheliu3, line 54: Now if the namenode / nameservice is not accessible there will be an exception here. I'm trying to add another test using `hadoopClient.isHealthy()` in hadoop client, which should be more verbose and specific.

shell/src/main/java/alluxio/cli/ValidateHdfsMount.java
```diff
@@ -0,0 +1,235 @@
...
+    ExecutorService executor = Executors.newFixedThreadPool(hosts.size());
+
+    // Invoke validateHdfsMount locally on each host
+    Map<String, CompletableFuture<CommandReturn>> resultFuture = new HashMap<>();
+    for (String host : hosts) {
+      LOG.info("validate hdfs mount on host {}", host);
+
+      // We make the assumption that the Alluxio WORK_DIR is the same
+      String workDir = conf.get(PropertyKey.WORK_DIR);
+      String alluxioBinPath = Paths.get(workDir, "bin/alluxio")
```
LuQQiu, line 166: @madanadit Is the ValidateHdfs required to run on all nodes? @jiacheliu3 I saw the local option and validate hdfs will run by default on local and remote nodes. The problem is this requires SSH and file permissions which is hard to fulfill in real env. I would suggest either remove the remote check (or at least make it optional) or expose validateHdfs rpcs in master/worker. Then we can ask for the server check through rpc channel.

shell/src/main/java/alluxio/cli/ValidateEnv.java
```diff
@@ -87,80 +93,108 @@
...
+  /**
+   * Initializes from the target UFS path and configurations.
+   *
+   * @param path the UFS path
+   * @param conf the UFS configurtions
+   * */
+  public ValidateEnv(String path, AlluxioConfiguration conf) {
```
LuQQiu, line 105: Why we choose to put the validateHdfs logic inside validateEnv?
personally i will prefer to have a separate class for the hdfs<>ValidationTask. and ValidateEnv can call the hdfs<>ValidationTasks if it's root ufs is hdfs. This is because many of the ValidateEnv tasks are not related to hdfs.

shell/src/main/java/alluxio/cli/ValidateHdfsMount.java
madanadit: In this release (2.3) we target single node (local) validation with no ssh requirement. I agree, remove the remote check.

jiacheliu3: Removed remote check

shell/pom.xml
```diff
@@ -63,6 +63,10 @@
       <groupId>com.beust</groupId>
       <artifactId>jcommander</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
jiacheliu3: `ValidateHdfsMount` uses apache commons cmdline to follow the same format as the `MountCommand`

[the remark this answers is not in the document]
jiacheliu3: And all the `ValidateEnv` family are using `org.apache.commons.cli.Option`. This is a side effect moving them here.

shell/src/main/java/alluxio/cli/ValidateHdfsMount.java
jiacheliu3: Not really, i'll remove this TODO. I found the existing cmdline works just fine. JCommand is not a better option anymore.

shell/pom.xml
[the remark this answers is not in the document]
jiacheliu3: Ooops, I didn't notice this is just moving the apache.commons from "internal depencies", which seems to be a wrong location for it.

shell/src/main/java/alluxio/cli/ValidateEnv.java
jiacheliu3: Actually I tried to move that into `ValidateHdfsMount` but turned out to be very hard. It's hard because the UfsContractTest is not providing a "functional" interface where you just feed it parameters and get results. It has much internal logic and using its embedded fields. Another reason for adding a method to UfsContractTest is, actually most of its tests are run. 

Left a TODO tho. I will have to refactor much of that class to achieve this.

shell/pom.xml
```diff
@@ -91,8 +95,14 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-compress</artifactId>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-examples</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
jiacheliu3: I found these extra pom are no longer necessary (I removed the things that originally needed them).

shell/src/main/java/alluxio/cli/validation/hdfs/HdfsProxyUserValidationTask.java
```diff
@@ -0,0 +1,106 @@
...
+      mMsg.append(String.format("Alluxio user %s can impersonate as any user/group in HDFS.%n",
+              userName));
+      return new ValidationUtils.TaskResult(ValidationUtils.State.OK, getName(),
+              mMsg.toString(), mAdvice.toString());
+    }
+
+    // There are proxyable users and groups for the Alluxio user in HDFS,
+    // but we cannot know if that is a full set.
+    // Leave a warning for the user to double check.
+    // TODO(jiacheng): can we do better check than this?
```
jiacheliu3, line 69: @madanadit PTAL at the impersonation check now, updated per our discussion today. Now the logic goes as follows:

```
// Skipped if authType is NOSASL
validateProxyUsers(String alluxioUser) {
// if neither hadoop.proxyuser.alluxioUser.users and hadoop.proxyuser.alluxioUser.groups are configured -> ERROR
// If either is *, OK
// Otherwise  -> WARN and let the user check. 
}
```

What do you think about the final WARN scenario? Maybe hadoop.proxyuser.alluxioUser.users or hadoop.proxyuser.alluxioUser.groups has some config but we don't really know if that's what we want. I left a TODO in case we want to add a little more to this part.

stress/common/src/main/java/alluxio/stress/Summary.java
```diff
@@ -11,6 +11,8 @@
 
 package alluxio.stress;
 
+import alluxio.util.JsonSerializable;
```
madanadit, line 14: is this unused?

shell/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -84,6 +109,63 @@ private void run() throws Exception {
...
+      // Check if the ufs path is valid
+      if (factory == null || !factory.supportsPath(mUfsPath)) {
+        msgStream.append(String.format("%s is not a valid path%n", mUfsPath));
+        adviceStream.append(String.format("Please validate if %s is a correct path\n", mUfsPath));
+        return new ValidateUtils.TaskResult(ValidateUtils.State.FAILED, TASK_NAME,
+                msgBuf.toString(), adviceBuf.toString());
+      }
+
+      // Set common properties
+      mConf.set(PropertyKey.UNDERFS_LISTING_LENGTH, "50");
```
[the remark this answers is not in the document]
LuQQiu: Those properties are inherited from original UfsContrctTests and are added to for testing purposes. Like listing less files , using small blocks and speed up tests. I think it will be better to keep these properties

madanadit:

PR #11488, opened 2020-05-29
[DOCFIX] Fix OBS doc

LuQQiu:


LuQQiu:
This PR is used to see if PR-BUILD function correctly

PR #11508, opened 2020-06-02
Use better method description for StressMaster report

bradyoo:


bradyoo:
@gpang

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -291,6 +293,15 @@ public synchronized void addAdditionalResult() throws IOException {
     public synchronized MasterBenchTaskResult getResult() {
       return mResult;
     }
+
+    private String getMethodDescription(String type, String methodName) {
+      if (type.contains("RPC")) {
+        final int classNameDivider = methodName.lastIndexOf(".");
```
gpang, line 299: Could this return -1, and will that cause a problem?

bradyoo: Yes it could return -1. No it wouldn't cause a problem because s.substring(0) returns s.

bradyoo: Conversely, returning the last index is also not a problem because 

s.substring(s.length) returns an empty string.

gpang:

PR #11513, opened 2020-06-03
Add readonly check while delete recursively

tinkerrrr:
When I use `rmr` to delete an alluxio path with a descendant which is a readonly mount point, alluxio will delete all the descendants no matter if it is a readonly mount point, this is extremely dangerous. So it's important to add readonly check while delete recursively.

**Use-case:**
```
./bin/alluxio fs mkdir /tmp
./bin/alluxio fs mount --readonly /tmp/mnt hdfs://example/tmp/mnt
```
UserA with all the permission did:
```
./bin/alluxio fs rmr /tmp/mnt
A write operation on /tmp/mnt under a readonly mount point hdfs://example/tmp/mnt is not allowed
```
UserA with all the permission did:
```
./bin/alluxio fs rmr /tmp
# Everything under hdfs://example/tmp/mnt gone
```
As a result, `hdfs://example/tmp/mnt` was unexpectedly deleted

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1549,7 +1549,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
               try {
                 mPermissionChecker.checkPermission(Mode.Bits.WRITE, childPath);
                 if (mMountTable.isMountPoint(childPath.getUri())) {
-                    mMountTable.checkUnderWritableMountPoint(childPath.getUri());
+                  mMountTable.checkUnderWritableMountPoint(childPath.getUri());
                 }
               } catch (AccessControlException e) {
                 failedChildren.add(e.getMessage());
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1548,6 +1548,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
             for (LockedInodePath childPath : descendants) {
               try {
                 mPermissionChecker.checkPermission(Mode.Bits.WRITE, childPath);
+                mMountTable.checkUnderWritableMountPoint(childPath.getUri());
               } catch (AccessControlException e) {
                 failedChildren.add(e.getMessage());
               }
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1548,7 +1548,9 @@ public final class DefaultFileSystemMaster extends CoreMaster
             for (LockedInodePath childPath : descendants) {
               try {
                 mPermissionChecker.checkPermission(Mode.Bits.WRITE, childPath);
-                mMountTable.checkIsWritableMountPoint(childPath.getUri());
+                if (mMountTable.isMountPoint(childPath.getUri())) {
+                    mMountTable.checkUnderWritableMountPoint(childPath.getUri());
+                }
               } catch (AccessControlException e) {
                 failedChildren.add(e.getMessage());
               }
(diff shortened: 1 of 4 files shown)


ZacBlanco:
@tinkerrrr thanks for reporting and fixing this. Could you sign the CLA using the email tied to your git commit?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1548,6 +1548,7 @@ public void delete(AlluxioURI path, DeleteContext context)
             for (LockedInodePath childPath : descendants) {
               try {
                 mPermissionChecker.checkPermission(Mode.Bits.WRITE, childPath);
+                mMountTable.checkUnderWritableMountPoint(childPath.getUri());
```
gpang, line 1551: How is this different from the check below? https://github.com/Alluxio/alluxio/pull/11513/files#diff-7f82122f923de058e3c4a20b43cd7e4eR1565

tinkerrrr: The check below won't check descendants of the path, it just check the path itself

tinkerrrr:
> @tinkerrrr thanks for reporting and fixing this. Could you sign the CLA using the email tied to your git commit?

I have already signed it

apc999:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: I see. Would checking only the descendant mount points work? I am concerned that checking every path may be unnecessarily expensive, especially for a large sub-tree of the namespace.

ZacBlanco: We already check the permissions of every file and directory when deleting recursively. I don't think this would make much of a difference

gpang: I thought it was unnecessary to check the same mount point multiple times. Also, `checkUnderWritableMountPoint()` calls `getMountPoint()`, and that needs to iterate over all the mount points to find the longest prefix.

tinkerrrr: > We already check the permissions of every file and directory when deleting recursively. I don't think this would make much of a difference

Well, I've updated the use-case in my first comment. As it described, the scenario is, a user with the right permission did this operation, the should-be-protected sub-directory was delete unexpectedly

tinkerrrr: > I thought it was unnecessary to check the same mount point multiple times. Also, `checkUnderWritableMountPoint()` calls `getMountPoint()`, and that needs to iterate over all the mount points to find the longest prefix.

Thanks for your advice! I think checking only the descendant mount points works

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1548,6 +1548,7 @@ public void delete(AlluxioURI path, DeleteContext context)
             for (LockedInodePath childPath : descendants) {
               try {
                 mPermissionChecker.checkPermission(Mode.Bits.WRITE, childPath);
+                mMountTable.checkIsWritableMountPoint(childPath.getUri());
```
gpang, line 1551: Instead of introducing a new method, I was thinking more like:

```
if (mMountTable.isMountPoint(childPath.getUri())) {
  mMountTable.checkUnderWritableMountPoint(childPath.getUri());
}
```

This would allow not creating a new method, and also, this performs the (potentially expensive) check for only the mount point paths.

What do you think?

tinkerrrr: I see, this way is much more elegant, thanks!

gpang:
Jenkins, test this please

gpang:

PR #11514, opened 2020-06-03
Limit printing too many blocks for MasterWorkerInfo toString()

jiacheliu3:
This makes `MasterWorkerInfo` only prints no more than 99 blocks instead of the whole list. The whole list can be incredibly long like thousands of blocks.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
+++ b/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
@@ -34,6 +34,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.annotation.concurrent.NotThreadSafe;
 
@@ -375,9 +376,13 @@ public final class MasterWorkerInfo {
 
   @Override
   public String toString() {
+    int blockSizeLimit = 100;
     return MoreObjects.toStringHelper(this).add("id", mId).add("workerAddress", mWorkerAddress)
         .add("capacityBytes", mCapacityBytes).add("usedBytes", mUsedBytes)
-        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs).add("blocks", mBlocks)
+        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs)
+        .add("blocks",
+              (mBlocks.size() < blockSizeLimit) ? mBlocks :
+                mBlocks.stream().limit(blockSizeLimit).collect(Collectors.toList()))
         .add("lostStorage", mLostStorage).toString();
   }
 
--- a/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
+++ b/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
@@ -380,6 +380,7 @@ public final class MasterWorkerInfo {
     return MoreObjects.toStringHelper(this).add("id", mId).add("workerAddress", mWorkerAddress)
         .add("capacityBytes", mCapacityBytes).add("usedBytes", mUsedBytes)
         .add("lastUpdatedTimeMs", mLastUpdatedTimeMs)
+        .add("blockCount", mBlocks.size())
         .add("blocks",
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -375,9 +376,13 @@ public boolean hasLostStorage() {
 
   @Override
   public String toString() {
+    int blockSizeLimit = 100;
     return MoreObjects.toStringHelper(this).add("id", mId).add("workerAddress", mWorkerAddress)
         .add("capacityBytes", mCapacityBytes).add("usedBytes", mUsedBytes)
-        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs).add("blocks", mBlocks)
+        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs)
+        .add("blocks",
+              (mBlocks.size() < blockSizeLimit) ? mBlocks :
```
ggezer, line 384: you shouldn't need this condition with `limit`.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -375,9 +376,13 @@ public boolean hasLostStorage() {
 
   @Override
   public String toString() {
+    int blockSizeLimit = 100;
```
ggezer, line 379: I feel like 100 is pretty low for this?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -375,9 +376,13 @@ public boolean hasLostStorage() {
...
     return MoreObjects.toStringHelper(this).add("id", mId).add("workerAddress", mWorkerAddress)
         .add("capacityBytes", mCapacityBytes).add("usedBytes", mUsedBytes)
-        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs).add("blocks", mBlocks)
+        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs)
+        .add("blocks",
+              (mBlocks.size() < blockSizeLimit) ? mBlocks :
+                mBlocks.stream().limit(blockSizeLimit).collect(Collectors.toList()))
         .add("lostStorage", mLostStorage).toString();
```
ggezer, line 386: now that we're truncating the list, can we add block-count as a separate field?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: I think adding this condition will reduce one copy if we don't need to truncate?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: This is a good question. I've no idea what is a good split and don't have statistics for the decision. I can only say a MasterWorkerInfo of a good 400 blocks can fill my 15-inch screen. 

The benefit of not truncating this list is we can search for a specific block id and know for sure if it exists on one worker. Once we start truncating there's no such nice certainty. So if we do truncate, intuitively I lean towards using a smaller limit, as we lose the certainty whatsoever.

ggezer: Can we do it like this, truncate in normal mode, and leave it as is if debug enabled.
@gpang WDYT about this in general?

jiacheliu3: Good idea!

ZacBlanco:
@ggezer @jiacheliu3 I would like to merge this. Can we finish the review?

ggezer:
@jiacheliu3 Can you open an issue linked to this PR. I don't think we should be cutting info but rather think this through later.

jiacheliu3:
@ggezer https://github.com/Alluxio/alluxio/issues/11729 logged for this topic

ZacBlanco:
ping!

jiacheliu3:
I'll close this PR as we need to further think it through (on the ticket).
(pong)

PR #11525, opened 2020-06-08
Improve robustness of LocalCacheManager

apc999:
This PR fixes a potential race condition in LocalCacheManager and improves its robustness.

Though designed to be threadsafe, `LocalCacheManager` returns a `ReadableByteChannel` to 
`LocalCacheFileInStream`, where I/O operations like reading a file is actually done at `read` or `positionedRead` at `LocalCacheFileInStream`. This leaves an opportunity to multiple threads racing on the same `file` outside threadsafe `LocalCacheManager`. E.g., a super slow reader thread may contend with a evictor thread who wants to remove the same page file per LRU policy, leaving the reader thread risk undefined behavior (up to operating system).

Invariants after this PR:

(1) operating a page files is always protected inside the page lock.
(2) given no operation on the cache `LocalCacheFileInStream` (moved to `LocalCacheManager` in this PR), if cache fails in certain unexpected way, exceptions will always be catched by the `NoExceptionCacheManager`.

This PR changes the API of `LocalCacheManager.read` from returning a channel to returning the number of bytes read but copying bytes inside, which is a bit different from what i previously discussed with @calvinjia . This will make the code much simpler.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
@@ -26,17 +26,19 @@ import com.codahale.metrics.Meter;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Suppliers;
 import com.google.common.io.Closer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
 
 import javax.annotation.concurrent.NotThreadSafe;
-import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.channels.ReadableByteChannel;
 
 /**
  * Implementation of {@link FileInStream} that reads from a local cache if possible.
  */
 @NotThreadSafe
 public class LocalCacheFileInStream extends FileInStream {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalCacheFileInStream.class);
 
   /** Page size in bytes. */
   protected final long mPageSize;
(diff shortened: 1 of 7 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -132,47 +134,40 @@ public int read(byte[] b, int off, int len) throws IOException {
...
     // for each page, check if it is available in the cache
-    while (bytesRead < lengthToRead) {
+    while (totalBytesRead < lengthToRead) {
       long currentPage = mPosition / mPageSize;
       int currentPageOffset = (int) (mPosition % mPageSize);
-      int bytesLeftInPage = (int) Math.min(mPageSize - currentPageOffset, lengthToRead - bytesRead);
+      int bytesLeftInPage =
+          (int) Math.min(mPageSize - currentPageOffset, lengthToRead - totalBytesRead);
       PageId pageId = new PageId(mStatus.getFileIdentifier(), currentPage);
-      try (ReadableByteChannel cachedData = mCacheManager.get(pageId, currentPageOffset)) {
```
apc999, line 143: move the IO operations inside `LocalCacheManager` under page locks to coordinate with other threads, also procted by `NoExceptionManager`

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -445,18 +445,33 @@ private boolean deletePage(PageId pageId, PageInfo pageInfo) {
...
+        Metrics.GET_ERRORS_FAILED_READ.inc();
+        throw new IOException(String.format(
+            "Failed to read page {}: supposed to read {} bytes, {} bytes actually read",
+            pageId, bytesToRead, buf.position() - offsetInBuffer));
+      }
     } catch (IOException | PageNotFoundException e) {
       LOG.error("Failed to get existing page {}: {}", pageId, e);
       Metrics.GET_ERRORS.inc();
-      return null;
+      return -1;
```
calvinjia, line 471: Should we have this logic here or leave it to `NoExceptionCacheManager`?

apc999: it needs to be here so we can clean up the state https://github.com/Alluxio/alluxio/pull/11525/files#diff-52b3926ca60dbb1ab7f30f23ada07efdR365

calvinjia: got it, thanks.

apc999:
Jenkins, test this please

apc999:
Jenkins, test this please

apc999:

jainxrohit:
@apc999  Can we merge this to branch-2.2 as well?

apc999:

Issue #11526, opened 2020-06-08
Add a root/nonroot mode switch for Helm chart


jiacheliu3:
**Is your feature request related to a problem? Please describe.**
This request is originally from @cheyang 

In current 2.3 we removed dependency on root privilege in Helm chart. Specifically we removed on `hostNetwork` and `hostPath`, and run Alluxio pods with non-root user. However, when a user has root privilege and wants to stick to `hostNetwork` and `hostPath`, there's no easy way to achieve it.

A user may want `hostNetwork` and `hostPath` for the maximized performance. 

**Describe the solution you'd like**
There should be a switch to easily switch between "root" and "nonroot" modes in the deployment. The check can be as easy as `{{- $isRoot := eq (int .Values.user) 0 }}`

Root mode:
1. Workers use `hostNetwork`
1. Masters use `hostPath` for journal and metastore, instead of PV.
1. Workers use `hostPath` for tiered storage and domain socket, instead of PV.

Nonroot mode:
1. Workers do not use `hostNetwork`, and set `ALLUXIO_WORKER_CONTAINER_HOSTNAME`.
1. Masters use `volumeClaimTemplates` to generate PVCs for journal and metastore.
1. Workers use PVC for tiered storage and domain socket

**Describe alternatives you've considered**

**Urgency**
Explain why the feature is important

**Additional context**

jiacheliu3:
@cheyang I don't have good sources for the performance difference in `hostNetwork` vs no `hostNetwork`, or `hostPath` vs `local`. Do you mind adding some references/statistics if you happen to have any? Thanks!

jiacheliu3:
@madanadit PTAL https://github.com/Alluxio/alluxio/pull/11527 is the PR for this feature. This PR provides everything except the worker tiered store, because the tiered stores are defined one by one manually.

If the feature request makes sense to you, I will proceed with all the necessary tests on my PR. Thanks!

cheyang:
@liujiacheng I suggest we shouldn't assume the user is root, then choose hostNetwork. And ClusterFirstWithHostNet should be used when hostnewtork  is true.

```
hostNetwork: {{ hasKey .Values.worker "hostNetwork" | ternary .Values.worker.hostNetwork $isRoot }}
[...]

PR #11527, opened 2020-06-08
Add a root/nonroot switch to helm chart

jiacheliu3:
This resolves https://github.com/Alluxio/alluxio/issues/11526

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
@@ -50,8 +50,8 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-master
     spec:
-      hostNetwork: {{ .Values.master.hostNetwork }}
-      dnsPolicy: {{ .Values.master.dnsPolicy }}
+      hostNetwork: {{ hasKey .Values.master "hostNetwork" | ternary .Values.master.hostNetwork $isRoot }}
+      dnsPolicy: {{ .Values.master.dnsPolicy | default ($isRoot | ternary "ClusterFirstWithHostNet" "ClusterFirst") }}
       nodeSelector:
       {{- if .Values.master.nodeSelector }}
 {{ toYaml .Values.master.nodeSelector | trim | indent 8  }}
@@ -194,8 +194,23 @@ spec:
             claimName: "{{ .name }}"
           {{- end }}
         {{- end }}
+  {{- if $isRoot }}
+  volumes:
+    {{- if $needJournalVolume }}
+    - name: alluxio-journal
+      hostPath:
+        path: {{ .Values.journal.hostPath }}
+        type: DirectoryOrCreate
+    {{- end }}
+    {{- if .Values.master.metastore }}
+    - name: alluxio-metastore
+      hostPath:
+        path: {{ .Values.master.metastore.hostPath }}
+        type: DirectoryOrCreate
+    {{- end }}
+  {{- else }}
   volumeClaimTemplates:
-  {{- if $needJournalVolume }}
+    {{- if $needJournalVolume }}
     - metadata:
         name: alluxio-journal
       spec:
@@ -206,7 +221,7 @@ spec:
           requests:
             storage: {{ .Values.journal.size }}
   {{- end }}
-  {{- if .Values.master.metastore }}
+    {{- if .Values.master.metastore }}
     - metadata:
         name: alluxio-metastore
       spec:
@@ -216,4 +231,5 @@ spec:
         resources:
           requests:
             storage: {{ .Values.master.metastore.size }}
-  {{- end }}
+    {{- end }}
+  {{- end}}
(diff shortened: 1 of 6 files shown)


integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -50,8 +50,8 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-master
     spec:
-      hostNetwork: {{ .Values.master.hostNetwork }}
-      dnsPolicy: {{ .Values.master.dnsPolicy }}
+      hostNetwork: {{ hasKey .Values.master "hostNetwork" | ternary .Values.master.hostNetwork $isRoot }}
```
jiacheliu3, line 53: Use the specified `.Values.master.hostNetwork` if specified, otherwise infer based on whether the user is root.

FYI for boolean values the `{{ .Values.master.hostNetwork | default $isRoot}}` grammar doesn't work, because a false value for `.Values.master.hostNetwork` will also go to the default branch.

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -194,8 +194,23 @@ spec:
             claimName: "{{ .name }}"
           {{- end }}
         {{- end }}
+  {{- if $isRoot }}
+  volumes:
+    {{- if $needJournalVolume }}
+    - name: alluxio-journal
+      hostPath:
```
jiacheliu3, line 201: I don't think we still want to consider `emptyDir`, do we?

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -73,15 +73,13 @@ master:
   # JVM options specific to the master container
   jvmOptions:
   nodeSelector: {}
-  hostNetwork: false
-  dnsPolicy: ClusterFirst
   # Example: use ROCKS DB instead of Heap
   # metastore:
   #   size: 1Gi
-  #   mountPath: /metastore
```
jiacheliu3, line 81: This field is obsoleted. Removing it is irrelevant to the change.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -73,15 +73,13 @@ master:
   # JVM options specific to the master container
   jvmOptions:
   nodeSelector: {}
-  hostNetwork: false
-  dnsPolicy: ClusterFirst
```
jiacheliu3, line 77: Removed these two by default and use inference. However they can still be set to override, especially when we want to use some dnsPolicy other than `ClusterFirst` and `ClusterFirstWithHostNet`

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
madanadit: no

integration/kubernetes/helm-chart/alluxio/values.yaml
madanadit: instead of removing this, i would comment it out and explain the order of precedence

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -119,6 +117,7 @@ journal:
   type: "UFS" # "UFS" or "EMBEDDED"
   ufsType: "local" # Ignored if type is "EMBEDDED". "local" or "HDFS"
   folder: "/journal" # Master journal folder
+  hostPath: "/tmp/journal"
```
madanadit, line 120: add comment defining when this property is used

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -229,6 +226,7 @@ shortCircuit:
     - ReadWriteOnce
   pvcSize: 1Mi
   pvcStorageClass: standard
+  hostPath: "/tmp/alluxio-domain"
```
madanadit, line 229: same here, comment

integration/kubernetes/helm-chart/alluxio/values.yaml
jiacheliu3: Good idea!

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -194,22 +194,8 @@ spec:
...
-            path: {{ printf "%v/%v" .Values.journal.hostPath (randAlphaNum 16) }}
-            type: DirectoryOrCreate
-    {{- end }}
-    {{- if .Values.master.metastore }}
-        - name: alluxio-metastore
-          hostPath:
-            path: {{ printf "%v/%v" .Values.master.metastore.hostPath (randAlphaNum 16) }}
-            type: DirectoryOrCreate
-    {{- end }}
-  {{- else }}
```
jiacheliu3: @madanadit I realized during testing that using hostPath is really hard when we deploy as StatefulSet, as each master should be using different paths. Current w/o internal logic in Alluxio (for example write to `<journal-path>/<master-id>/` when given the journal path) I don't think it will work. Currently in a non-k8s env, we don't really support deploying multiple masters on one machine right?

madanadit: yep, multiple masters on the same host is not a scenario we plan to support

madanadit:

PR #11529, opened 2020-06-08
Prevent TTL from being set on synced inodes

ZacBlanco:


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -645,6 +650,11 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
             LoadMetadataContext loadMetadataContext =
                 LoadMetadataContext.mergeFrom(LoadMetadataPOptions.newBuilder()
                     .setLoadDescendantType(LoadDescendantPType.NONE)
+                    // No Ttl on loaded files
+                    .setCommonOptions(FileSystemMasterCommonPOptions.newBuilder()
+                        .setTtl(-1)
+                        .setTtlAction(TtlAction.DELETE)
```
gpang, line 656: Do we have to set this explicitly? If so, why isn't it set in the other method above?

Also, can we have a constant for this "no ttl" common option. instead of building each time?

ZacBlanco: > Do we have to set this explicitly? If so, why isn't it set in the other method above?
"Other method above" is not clear to me.

ZacBlanco:

PR #11535, opened 2020-06-09
Implement interrupt based locking protocol over state-lock

ggezer:
This PR, in general, introduces a new locking framework over state-lock. Changes can be summarized as:

- Backups take the lock exclusive with an initial grace-cycle. Grace-cycle is defined as `tryLock-sleep` cycles until the lock is held. 3 graceful locking strategies have been implemented (TIMEOUT, FORCED)

- Immediate shared holders/waiters of state-lock may now be interrupted as part of interrupt-cycle task, which is activated before or after taking the lock.

- Shell/daily backups can now be configured for all grace modes via master config change. (Defaults for Shell: TIMEOUT, Daily: FORCED)

- A new maintenance duration( exclusive-only phase) is defined for masters, during which only exclusive locking will be allowed. This duration is effective after the first time masters are started, after reading the journals.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1317,6 +1317,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_BACKUP_STATE_LOCK_FORCED_DURATION =
+      new Builder(Name.MASTER_BACKUP_STATE_LOCK_FORCED_DURATION)
+          .setDefaultValue("15min")
+          .setDescription("Exclusive locking of the state-lock will timeout after "
+              + "this duration is spent on forced phase.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_INTERVAL =
       new Builder(Name.MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_INTERVAL)
           .setDefaultValue("30sec")
@@ -1357,12 +1365,12 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDefaultValue("TIMEOUT")
           .setDescription("Grace mode helps taking the state-lock exclusively for backup "
               + "with minimum disruption to existing RPCs. This low-impact locking phase "
-              + "is called grace-cycle. Three modes are supported: TIMEOUT/GUARANTEED/SKIP."
+              + "is called grace-cycle. Two modes are supported: TIMEOUT/FORCED."
               + "TIMEOUT: Means exclusive locking will timeout if it cannot acquire the lock"
               + "with grace-cycle. "
-              + "GUARANTEED: Means the state-lock will be taken directly if grace-cycle fails "
-              + "to acquire it. "
-              + "SKIP: Means the grace-cycle won't be tried.")
+              + "FORCED: Means the state-lock will be taken forcefully if grace-cycle fails "
+              + "to acquire it. Forced phase might trigger interrupting of existing RPCs if "
+              + "it is enabled.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
@@ -1391,15 +1399,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_DAILY_BACKUP_STATE_LOCK_GRACE_MODE =
       new Builder(Name.MASTER_DAILY_BACKUP_STATE_LOCK_GRACE_MODE)
-          .setDefaultValue("GUARANTEED")
+          .setDefaultValue("FORCED")
           .setDescription("Grace mode helps taking the state-lock exclusively for backup "
               + "with minimum disruption to existing RPCs. This low-impact locking phase "
-              + "is called grace-cycle. Three modes are supported: TIMEOUT/GUARANTEED/SKIP."
+              + "is called grace-cycle. Two modes are supported: TIMEOUT/FORCED."
               + "TIMEOUT: Means exclusive locking will timeout if it cannot acquire the lock"
               + "with grace-cycle. "
-              + "GUARANTEED: Means the state-lock will be taken directly if grace-cycle fails "
-              + "to acquire it. "
-              + "SKIP: Means the grace-cycle won't be tried.")
+              + "FORCED: Means the state-lock will be taken forcefully if grace-cycle fails "
+              + "to acquire it. Forced phase might trigger interrupting of existing RPCs if "
+              + "it is enabled.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
@@ -4898,6 +4906,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.backup.state.lock.exclusive.duration";
     public static final String MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_ENABLED =
         "alluxio.master.backup.state.lock.interrupt.cycle.enabled";
+    public static final String MASTER_BACKUP_STATE_LOCK_FORCED_DURATION =
+        "alluxio.master.backup.state.lock.forced.duration";
     public static final String MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_INTERVAL =
         "alluxio.master.backup.state.lock.interrupt.cycle.interval";
     public static final String MASTER_SHELL_BACKUP_STATE_LOCK_GRACE_MODE =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1305,7 +1305,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       new Builder(Name.MASTER_BACKUP_STATE_LOCK_EXCLUSIVE_DURATION)
           .setDefaultValue("0ms")
           .setDescription("Alluxio master will allow only exclusive locking of "
-              + "the state-lock for this duration.")
+              + "the state-lock for this duration. This duration starts after masters "
+              + "are started for the first time.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1301,6 +1301,30 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_BACKUP_STATE_LOCK_EXCLUSIVE_DURATION =
+      new Builder(Name.MASTER_BACKUP_STATE_LOCK_EXCLUSIVE_DURATION)
+          .setDefaultValue("0ms")
+          .setDescription("Alluxio master will allow only exclusive locking of "
+              + "the state-lock for this duration.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_ENABLED =
+      new Builder(Name.MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("This controls whether RPCs that are waiting/holding state-lock "
+              + "in shared-mode will be interrupted while state-lock is taken exclusively.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_BACKUP_STATE_LOCK_INTERRUPT_CYCLE_INTERVAL =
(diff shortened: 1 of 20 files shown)


tests/src/test/java/alluxio/client/cli/fsadmin/BackupCommandIntegrationTest.java
```diff
@@ -87,4 +101,54 @@ public void timeout() throws Exception {
...
+
+    FileSystemMaster fsMaster = masterProcess.getMaster(FileSystemMaster.class);
+    Whitebox.setInternalState(fsMaster, "mMasterContext", masterCtx);
+
+    FileSystem fs = mLocalAlluxioCluster.getClient();
+    FileOutStream fsOutStr = fs.createFile(new AlluxioURI("/test"));
+    fsOutStr.write("bogus".getBytes());
+    // Lock the state-change lock on the master.
+    LockResource lr =
+        stateLockManager.lockExclusive(new StateLockOptions(false, false, true, 0, 0, 0, 100));
```
ggezer, line 124: @gpang if you pass the last argument of StateLockOptions as `0` (which means no interrupt-waiters cycle), then the repro will fail as before.

ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -0,0 +1,197 @@
...
+   * Creates a new state-lock manager.
+   */
+  public StateLockManager() {
+    mSharedWaiters = new ConcurrentHashSet<>();
+    mSharedHolders = new ConcurrentHashSet<>();
+
+    mScheduler = Executors
+        .newSingleThreadScheduledExecutor(ThreadFactoryUtils.build("state-lock-manager-%d", true));
+
+    mSafeMode = ServerConfiguration.getBoolean(PropertyKey.MASTER_BACKUP_STATE_LOCK_SAFE_MODE);
```
gpang, line 72: does safemode never get reset? That means the master can ONLY take a backup and never service clients, even when it is primary?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -0,0 +1,188 @@
...
+      // Register as waiter.
+      mSharedWaiters.add(Thread.currentThread());
+      // Grab the lock interruptibly.
+      mStateLock.readLock().lockInterruptibly();
+    } finally {
+      // Remove itself from waiters.
+      mSharedWaiters.remove(Thread.currentThread());
+    }
+    // Register as holder.
+    mSharedHolders.add(Thread.currentThread());
```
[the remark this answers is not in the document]
gpang: hrmmm, this seems tricky and hard to guarantee correct behavior...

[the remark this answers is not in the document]
ggezer: what is "correct behaviour" ?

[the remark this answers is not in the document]
ggezer: Any inconsistency could lead to two faulty actions:
1- interrupt a thread when not intended
2- don't interrupt a thread when required

2 is covered by the interrupt-cycle. 
1 can be hardened by proposed locking over sets. Then it can only happen at the end of write-lock and that's acceptable.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: We can make it something that resets but need to coordinate this between masters.
Right now it's just a crude failsafe.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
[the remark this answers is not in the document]
ggezer: @gpang I merged waiters and holders sets as we can reliably keep them at sync without heavy locking over the sets. 

In the current state, I don't see any `unacceptable` inconsistency.

- **the interrupt could be missed for a thread:** this will be handled by the next tick.
- **the interrupted thread might already have released the lock:** this will be ignored as a victim of backup. (This can be fixed by keeping track of which threads the manager has interrupted and reset the flags based on that. I don't think that's necessary for the overall scenario though.)

WDYT?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -0,0 +1,180 @@
...
+      if (!lockOptions.lockAfterGraceCycle()) {
+        throw new TimeoutException(
+                ExceptionMessage.STATE_LOCK_TIMED_OUT.getMessage(lockOptions.getGraceCycleTimeoutMs()));
+      }
+      if (lockOptions.interruptCycleAfterGraceCycle()) {
+        // Schedule interrupt cycle before entering the lock because it might wait in the queue.
+        scheduleInterruptCycle(lockOptions.getInterruptWaitersInternalMs());
+      }
+      // Own the lock.
+      mStateLock.writeLock().lock();
```
gpang, line 141: This could block for a long time, and thus block other RPCs (including heartbeats) for a long time?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    // We have the lock, wrap it and return.
+    // Register an action for cancelling the interrupt cycle before releasing the lock.
+    return new LockResource(mStateLock.writeLock(), false, false, () -> {
+      // Before releasing the write-lock, deactivate interrupter if active.
+      if (mInterrupterFuture != null && !mInterrupterFuture.isDone()) {
```
gpang, line 148: Is there any danger that this `mInterrupterFuture` points to a different future than the one that was started?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1296,6 +1296,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_BACKUP_STATE_LOCK_SAFE_MODE =
+      new Builder(Name.MASTER_BACKUP_STATE_LOCK_SAFE_MODE)
+          .setDefaultValue(false)
+          .setDescription("Ignores client state-lock requests for backup to always take the lock.")
```
gpang, line 1302: I don't know how this should be used by users.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: Yes but interrupt cycle is activated before this point.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: It's not possible due to exclusive locking.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: The typical scenario is when you can't take backups (shell/daily) due to a never-ending RPC. Restarting Alluxio master with this will guarantee that backup can take the lock and continue.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: Need to support multiple exclusive lockers though. This PR should cover that. To do that properly, the interrupt cycle should be owned by the state-lock-manager.

core/common/src/main/java/alluxio/conf/PropertyKey.java
gpang: @ggezer 
Yes, but it will never service RPCs. That will make this master unusable, right? Why would a user want an unusable master?

ggezer: The point is not "wanting an unusable master", It's "wanting a backup".
@calvinjia @ns1123 WDYT about this mode?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: This is fixed now.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1352,6 +1372,20 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_DAILY_BACKUP_STATE_LOCK_INTERRUPT_HOLDERS_AFTER_GRACE =
+      new Builder(Name.MASTER_DAILY_BACKUP_STATE_LOCK_INTERRUPT_HOLDERS_AFTER_GRACE)
+          .setDefaultValue(true)
+          .setDescription("Whether to interrupt state-lock holders after grace period.")
```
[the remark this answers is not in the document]
ggezer: @gpang Please re-check all props as I did a major reorg.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ns1123: I'd rather instead of being a server side property, that his is an option on the `fsadmin backup` command. That way a backup can be forced by an admin in a dire situation.

ggezer: @ns1123 This PR already makes the changes for doing the best we can in taking the backup. This particular property is introducing a new maintenance mode under which backup will be "guaranteed".

ggezer:
jenkins, test this please

ggezer:
@gpang @calvinjia This PR is no longer a draft. PTAL.

ggezer:
@gpang @calvinjia Ping.

ggezer:
@gpang @calvinjia @ns1123 Made the changes discussed offline.
- 2 modes of grace (TIMEOUT, FORCED)
- Made the exclusive-only duration (guaranteed mode) start after journals are read. Now the admin doesn't need to account for when the journals could have been read. (It's still a one-time window, so it doesn't reset with each HA cycle. )

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -0,0 +1,247 @@
...
+          Thread.sleep(Math.min(lockOptions.getGraceCycleSleepMs(), remainingWaitMs));
+        }
+      }
+    }
+    if (lockAcquired) { // Lock was acquired within grace-cycle.
+      activateInterruptCycle();
+    } else { // Lock couldn't be acquired by grace-cycle.
+      if (graceMode == StateLockOptions.GraceMode.TIMEOUT) {
+        throw new TimeoutException(
+            ExceptionMessage.STATE_LOCK_TIMED_OUT.getMessage(lockOptions.getGraceCycleTimeoutMs()));
```
calvinjia, line 169: Will it be easy to search for log message, where will this be logged?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -0,0 +1,247 @@
...
+   *
+   * @return the lock resource
+   * @throws InterruptedException
+   */
+  public LockResource lockShared() throws InterruptedException {
+    // Do not allow taking shared lock during safe-mode.
+    if (System.currentTimeMillis() < mExclusiveOnlyDeadlineMs) {
+      String safeModeMsg = String.format(
+          "Master still in exclusive-only phase for the state-lock."
+              + "Please see documentation for %s.",
```
calvinjia, line 124: Would be helpful to state how much longer the deadline is

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: This will end up being caught at `RpcUtils.callAndReturn`, which prints an error-log for runtime exceptions.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: Added more traces https://github.com/Alluxio/alluxio/pull/11535/commits/b8724736bbe4463dadc8920db951ab867fb3cd4f

core/server/common/src/main/java/alluxio/master/StateLockOptions.java
```diff
@@ -0,0 +1,136 @@
...
+
+import java.util.concurrent.locks.Lock;
+
+/**
+ * Used to define state-lock options for taking it exclusively.
+ * A typical framework for obtaining the state-lock exclusively is:
+ *  1- Run a grace-cycle:
+ *      cycle of {@link Lock#tryLock()} - {@link Thread#sleep(long)} calls.
+ *  2- Interrupt state-lock holders if grace-cycle fails.
+ *  3- Take the lock using {@link Lock#lock()}
```
gpang, line 25: I think this is stale, since we are still using tryLock.

core/server/common/src/main/java/alluxio/master/StateLockOptions.java
```diff
@@ -0,0 +1,136 @@
...
+import alluxio.conf.ServerConfiguration;
+
+import java.util.concurrent.locks.Lock;
+
+/**
+ * Used to define state-lock options for taking it exclusively.
+ * A typical framework for obtaining the state-lock exclusively is:
+ *  1- Run a grace-cycle:
+ *      cycle of {@link Lock#tryLock()} - {@link Thread#sleep(long)} calls.
+ *  2- Interrupt state-lock holders if grace-cycle fails.
```
gpang, line 24: This is actually something like "continually interrupt state-lock holders and waiters", right?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1301,6 +1301,39 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_BACKUP_STATE_LOCK_EXCLUSIVE_DURATION =
+      new Builder(Name.MASTER_BACKUP_STATE_LOCK_EXCLUSIVE_DURATION)
+          .setDefaultValue("0ms")
+          .setDescription("Alluxio master will allow only exclusive locking of "
+              + "the state-lock for this duration. This duration starts after masters "
```
gpang, line 1308: I think we should be more explicit here. This will reject all RPCs which need to take the state-lock, which is most client RPCs.

Should we also mention that this can be useful to guarantee a window on restart to initiate a backup?

ggezer:

ggezer:

PR #11537, opened 2020-06-10
Add a cache for PropertyKey.isValid

apc999:
In certain applications like Presto, frequently initializing FileSystem instance may be costly in CPU overhead. Further analysis shows when Hadoop configuration is large, regexp used in Alluxio to check whether an given configuration property from Hadoop can be recognized by Alluxio property template contributed a lot to the CPU overhead.

This PR aims to remove this cost. My initial attempt was to introduce extra logic to avoid  `PropertyKey.isValid`. However, this approach has a fairly large footprint in multiple most basic and important classes in Alluxio codebase, and increases complexity. I am concerned with the potential risk to reason and maintain the code after a patch like this.

Later I decided to instead  add a cache inside `PropertyKey.isValid`, given the inquiries are likely to be repeated. This approach has much less footprint to the codebase. In this PR I also cleaned up `AbstractFileSystem.initialize` in terms of configuration init.

### DIFF
--- a/core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
+++ b/core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
@@ -473,15 +473,23 @@ public abstract class AbstractFileSystem extends org.apache.hadoop.fs.FileSystem
     mStatistics = statistics;
     mUri = URI.create(mAlluxioHeader);
 
+    // take the URI properties, hadoop configuration, and given Alluxio configuration and merge
+    // all three into a single object.
     Map<String, Object> uriConfProperties = getConfigurationFromUri(uri);
-
+    Map<String, Object> hadoopConfProperties =
+        HadoopConfigurationUtils.getConfigurationFromHadoop(conf);
+    LOG.info("Creating Alluxio configuration from Hadoop configuration {}, uri configuration {}",
+        hadoopConfProperties, uriConfProperties);
     AlluxioProperties alluxioProps =
         (alluxioConfiguration != null) ? alluxioConfiguration.copyProperties()
             : ConfigurationUtils.defaults();
-    LOG.info("Creating Alluxio configuration from Hadoop configuration {}, uri configuration {}",
-        conf, uriConfProperties);
-    AlluxioConfiguration alluxioConf = mergeConfigurations(uriConfProperties, conf, alluxioProps);
-    mAlluxioConf = alluxioConf;
+    // Merge relevant Hadoop configuration into Alluxio's configuration.
+    alluxioProps.merge(hadoopConfProperties, Source.RUNTIME);
+    // Merge relevant connection details in the URI with the highest priority
+    alluxioProps.merge(uriConfProperties, Source.RUNTIME);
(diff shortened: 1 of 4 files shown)


core/client/hdfs/src/main/java/alluxio/hadoop/AbstractFileSystem.java
```diff
@@ -473,15 +473,23 @@ public synchronized void initialize(URI uri, org.apache.hadoop.conf.Configuratio
...
+        hadoopConfProperties, uriConfProperties);
     AlluxioProperties alluxioProps =
         (alluxioConfiguration != null) ? alluxioConfiguration.copyProperties()
             : ConfigurationUtils.defaults();
-    LOG.info("Creating Alluxio configuration from Hadoop configuration {}, uri configuration {}",
-        conf, uriConfProperties);
-    AlluxioConfiguration alluxioConf = mergeConfigurations(uriConfProperties, conf, alluxioProps);
-    mAlluxioConf = alluxioConf;
+    // Merge relevant Hadoop configuration into Alluxio's configuration.
+    alluxioProps.merge(hadoopConfProperties, Source.RUNTIME);
```
apc999, line 487: let's put all configuration merge in the same place, rather than distributed in multiple methods

core/client/hdfs/src/main/java/alluxio/hadoop/HadoopConfigurationUtils.java
```diff
@@ -30,35 +24,25 @@
...
-   * Merges Hadoop {@link org.apache.hadoop.conf.Configuration} with Alluxio properties.
+   * Extracts relevant configuration from Hadoop {@link org.apache.hadoop.conf.Configuration}.
    *
-   * @param hadoopConf the {@link org.apache.hadoop.conf.Configuration} to merge
-   * @param alluxioProps the Alluxio properties to merge
-   * @return a configuration with properties all merged
+   * @param hadoopConf the {@link org.apache.hadoop.conf.Configuration} to extract
+   * @return all relevant properties in a Map instance
    */
-  public static InstancedConfiguration mergeHadoopConfiguration(
```
jainxrohit, line 44: Presto was using this method 
https://github.com/prestodb/presto/blob/472538a02722c5409649533f9a602b5b76d9dd2d/presto-cache/src/main/java/com/facebook/presto/cache/alluxio/AlluxioCachingFileSystem.java#L98

jainxrohit: If we want to make these changes, we will need to fix this method
https://github.com/prestodb/presto/blob/472538a02722c5409649533f9a602b5b76d9dd2d/presto-cache/src/main/java/com/facebook/presto/cache/alluxio/AlluxioCachingFileSystem.java#L87

apc999: ok, i will keep this method

apc999:
Jenkins, test this please

apc999:

apc999:

jainxrohit:
> alluxio-bot, cherry-pick this to branch-2.2 please

I love these bots.

PR #11541, opened 2020-06-10
Simplify lock list downgrades

ZacBlanco:
Before this change lock downgrades did NOT work if the end
of the lock list had more than one write locked inode. This
hasn't burned us until now because the new syncing changes
assumed all inodes in the locklist would be downgraded when
calling LockedInodePath#downgradeToPattern(LockPattern).
This was not the case, which ended up causing a deadlock
when the lock list contained many write-locked entries

One common case where this can occur is when a user is
loading a deeply nested file or directory via some getStatus or
listStatus call which syncs metadata and in-turn, will create
the paths as it discovers them. In doing so, the lock list is
initially populated with write locks. We require them to all
be downgraded to read locks for the metadata sync to be
performed properly.

The new syncing design also now only requires downgrades
to read locks which simplifies a lot of logic which existed
to support downgrades from WRITE_EDGE to
WRITE_INODE. These changes removes a good chunk of
that complicated code in favor of a simpler approach
which simply performs the lock downgrade for all inodes
in the list.

A new integration test case was added for this regression.

### DIFF
--- a/core/common/src/main/java/alluxio/collections/LockPool.java
+++ b/core/common/src/main/java/alluxio/collections/LockPool.java
@@ -14,6 +14,7 @@ package alluxio.collections;
 import alluxio.Constants;
 import alluxio.concurrent.LockMode;
 import alluxio.resource.LockResource;
+import alluxio.resource.RWLockResource;
 import alluxio.resource.RefCountLockResource;
 import alluxio.util.ThreadFactoryUtils;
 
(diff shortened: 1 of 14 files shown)


ZacBlanco:
jenkins, test this please

gpang:
Jenkins, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
alluxio-bot, test this please

ZacBlanco:
alluxio-bot, test this please

ZacBlanco:

PR #11545, opened 2020-06-10
Update emr script

yuzhu:
allow empty ufs uri (means local hdfs mount)
-c flag which installs client jar and conf but do not start alluxio

The reason for these changes: 
1. so that we can launch an emr cluster where the hdfs of that cluster is mounted as root mount
2. so that we can load jars on to a cluster that does not need alluxio running. In this case, a hive metastore containing alluxio uris.

### DIFF
--- a/integration/emr/alluxio-emr.sh
+++ b/integration/emr/alluxio-emr.sh
@@ -171,15 +171,15 @@ main() {
   print_help() {
     local -r USAGE=$(cat <<USAGE_END
 
-Usage: alluxio-emr.sh [-u <root-ufs-uri>]
+Usage: alluxio-emr.sh <root-ufs-uri>
                       [-b <backup_uri>]
+                      [-c]
                       [-d <alluxio-download-uri>]
                       [-f <file_uri>]
                       [-i <journal_backup_uri>]
                       [-n <storage percentage>]
                       [-p <delimited_properties>]
                       [-s <property_delimiter>]
-                      [-c]
 
 alluxio-emr.sh is a script which can be used to bootstrap an AWS EMR cluster
 with Alluxio. It can download and install Alluxio as well as add properties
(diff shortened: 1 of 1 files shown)


yuzhu:
@Xenorith PTAL

integration/emr/alluxio-emr.sh
```diff
@@ -170,13 +171,15 @@ main() {
...
     local -r USAGE=$(cat <<USAGE_END
 
-Usage: alluxio-emr.sh <root-ufs-uri> [-b <backup_uri>]
-                             [ -d <alluxio-download-uri>]
-                             [-f <file_uri>]
-                             [-i <journal_backup_uri>]
-                             [-n <storage percentage>]
-                             [-p <delimited_properties>]
-                             [-s <property_delimiter>]
+Usage: alluxio-emr.sh [-u <root-ufs-uri>]
```
Xenorith, line 174: alphabetical ordering of flags (but i may suggest using other characters in a bit)

integration/emr/alluxio-emr.sh
```diff
@@ -244,17 +249,16 @@ USAGE_END
...
-    print_help 1
-  fi
-
-  local root_ufs_uri="${1}"
-  shift
-  while getopts "b:d:f:i:n:p:s:" option; do
-    OPTARG=$(echo -e "${OPTARG}" | tr -d '[:space:]')
+  local root_ufs_uri=""
+  local client_only="false"
+  while getopts ":u:b:d:f:i:n:p:s:c" option; do
```
Xenorith, line 254: the getopts string should be `"b:cd:f:i:n:p:s:u:"`
`:` means expect an argument value afterwards, so `c` doesn't get one

integration/emr/alluxio-emr.sh
madanadit, line 174: @yuzhu why introduce a new flag? we are breaking backwards compatibility here

yuzhu: the alternative is to use empty string to represent "use local hdfs as root ufs" 
I am fine with that too.

Xenorith: is that possible? would bash read this correctly? `alluxio-emr.sh "" -a someArg -b someArg ...`

yuzhu:
@madanadit @Xenorith  fixed back compat and alphabetical order

yuzhu:
alluxio-bots, merge this please

PR #11557, opened 2020-06-12
Fix inconsistent persistence status after setOwner failed

bf8086:
When Alluxio mounts an HDFS UFS as a non superuser, any file persist action would fail because it does not have permission to set owner on HDFS. However, user would see the file is already persisted on HDFS but yet Alluxio still show the file as TO_BE_PERSISTED. Further attempt to persist the file would end up timeout with misleading error message "Failed to rename x.0x000001724B700242.tmp to x".

This problem occurs because the `setOwner` operation is done after the `rename` operation on UFS, so when `setOwner` fails it leaves the persisted file in UFS but keeps the Alluxio persistence state as "TO_BE_PERSISTED". Further persist attempts would then fails due to file already exist and thus keeps on retrying and failing. 

This PR moves the setOwner call to be before rename happens, so when it fails the temporary files got removed correctly and the error in master log will state it is a permission issue instead of a renaming issue.


bf8086:
@gpang @apc999 PTAL.

bf8086:

Xenorith:

PR #11572, opened 2020-06-16
Support active sync for give list in EMR bootstrap

LuQQiu:
User can pass in a list of relative paths of root UFS to enable active sync on

### DIFF
--- a/integration/emr/alluxio-emr.sh
+++ b/integration/emr/alluxio-emr.sh
@@ -306,7 +306,7 @@ set_custom_alluxio_properties() {
   if [[ "${delimited_properties}" ]]; then
     # inject user defined properties from args
     echo "Setting user defined properties in ${ALLUXIO_SITE_PROPERTIES}"
-    IFS="${property_delimiter}" read -ra conf <<< "${delimited_properties}"
+    IFS="${delimiter}" read -ra conf <<< "${delimited_properties}"
     for property in "${conf[@]}"; do
       local key=${property%%"="*}
       local value=${property#*"="}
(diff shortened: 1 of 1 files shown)


LuQQiu:
@madanadit PTAL, thanks

integration/emr/alluxio-emr.sh
```diff
@@ -461,6 +462,12 @@ If a different Alluxio version is desired, see the -d option.
                     will be downloaded, and upon Alluxio startup, the Alluxio
                     master will read and restore the backup.
 
+  -l                A string containing a delimited list of relative paths under
```
madanadit, line 465: nit: 
- "list of relative paths" can be misleading. I prefer simply "list of paths"
- "UFS metadata will be periodically syncing to Alluxio namespace." to "UFS metadata will be periodically synced with the Alluxio namespace."

LuQQiu: list of paths could also hard to understand, it's the relative path of the given root ufs.  List of alluxio paths after mounting the given root UFS?

madanadit: i don''t get what you're trying to say here. "list of Alluxio paths" is good. where does relative come from? relative from root is an absolute path; these paths don't have to be a mount point, it can be any alluxio path. i'm fine with whatever you choose

LuQQiu: ok got it. What i mean is if root ufs is `hdfs:///a/` which is mounted to alluxio root, the list to sync is `hdfs:///a/b`, then they should provide `/b` path. `Alluxio path` is definitely more clear. Thanks

LuQQiu:

PR #11573, opened 2020-06-16
Reduce number of clients in DistributedLoadCommand

bradyoo:
Creating many clients in DistributedLoadCommand creates an issue in certain environments. Reduce the number of clients and also remove the parallelism around it to prevent race conditions when using the same client.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
@@ -67,15 +67,12 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
   private class JobAttempt {
     private final LoadConfig mJobConfig;
     private final RetryPolicy mRetryPolicy;
-    private final JobMasterClient mClient;
 
     private Long mJobId;
 
-    private JobAttempt(LoadConfig jobConfig, RetryPolicy retryPolicy, ClientContext clientContext) {
+    private JobAttempt(LoadConfig jobConfig, RetryPolicy retryPolicy) {
       mJobConfig = jobConfig;
       mRetryPolicy = retryPolicy;
-      mClient = JobMasterClient.Factory.create(
-          JobMasterClientContext.newBuilder(clientContext).build());
     }
 
     private boolean run() {
@@ -134,15 +131,11 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
       }
       return Status.RUNNING;
     }
-
-    private void close() throws IOException {
-      // propagate failing to close up to prevent potential memory leak
-      mClient.close();
-    }
   }
 
   private List<JobAttempt> mSubmittedJobAttempts;
   private int mActiveJobs;
+  private JobMasterClient mClient;
 
   /**
    * Constructs a new instance to load a file or directory in Alluxio space.
@@ -152,6 +145,7 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
   public DistributedLoadCommand(FileSystemContext fsContext) {
     super(fsContext);
     mSubmittedJobAttempts = Lists.newArrayList();
+    mClient = null;
   }
 
   @Override
@@ -186,8 +180,14 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
    * @param replication The replication of file to load into Alluxio memory
    */
   private JobAttempt newJob(AlluxioURI filePath, int replication) {
+    final ClientContext clientContext = ClientContext.create(mFsContext.getPathConf(filePath));
+    if (mClient == null) {
+      mClient = JobMasterClient.Factory.create(
+          JobMasterClientContext.newBuilder(clientContext).build());
+    }
+
     JobAttempt jobAttempt = new JobAttempt(new LoadConfig(filePath.getPath(), replication),
-        new CountingRetry(3), ClientContext.create(mFsContext.getPathConf(filePath)));
+        new CountingRetry(3));
 
     jobAttempt.run();
 
@@ -200,7 +200,7 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
   private void waitJob() {
     AtomicBoolean removed = new AtomicBoolean(false);
     while (true) {
-      mSubmittedJobAttempts = mSubmittedJobAttempts.parallelStream().filter((jobAttempt) -> {
+      mSubmittedJobAttempts = mSubmittedJobAttempts.stream().filter((jobAttempt) -> {
         Status check = jobAttempt.check();
         switch (check) {
           case CREATED:
@@ -209,11 +209,6 @@ public final class DistributedLoadCommand extends AbstractFileSystemCommand {
           case CANCELED:
           case COMPLETED:
             removed.set(true);
-            try {
-              jobAttempt.close();
-            } catch (IOException e) {
-              LOG.warn("Unable to close job master client", e);
-            }
             return false;
           case FAILED:
             removed.set(true);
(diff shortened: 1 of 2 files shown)


bradyoo:
@gpang

bradyoo:

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -200,7 +200,7 @@ private JobAttempt newJob(AlluxioURI filePath, int replication) {
   private void waitJob() {
     AtomicBoolean removed = new AtomicBoolean(false);
     while (true) {
-      mSubmittedJobAttempts = mSubmittedJobAttempts.parallelStream().filter((jobAttempt) -> {
+      mSubmittedJobAttempts = mSubmittedJobAttempts.stream().filter((jobAttempt) -> {
```
ggezer, line 203: Does this method spin over checking jobs?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -186,8 +180,14 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
    * @param replication The replication of file to load into Alluxio memory
    */
   private JobAttempt newJob(AlluxioURI filePath, int replication) {
+    final ClientContext clientContext = ClientContext.create(mFsContext.getPathConf(filePath));
```
ggezer, line 183: Do you need the path-conf here?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -67,15 +67,12 @@
...
-    private final JobMasterClient mClient;
 
     private Long mJobId;
 
-    private JobAttempt(LoadConfig jobConfig, RetryPolicy retryPolicy, ClientContext clientContext) {
+    private JobAttempt(LoadConfig jobConfig, RetryPolicy retryPolicy) {
       mJobConfig = jobConfig;
       mRetryPolicy = retryPolicy;
-      mClient = JobMasterClient.Factory.create(
-          JobMasterClientContext.newBuilder(clientContext).build());
```
ggezer, line 78: Why can't we just use this one always?

bradyoo: What does that mean?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
bradyoo: I left it currently as the first context of the first file it hits though it may not be necessary.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
bradyoo: This method spins over checking jobs until it removed one because one is complete/cancelled/failed.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
ggezer: If job_worker is the one that eventually uses the path-conf, then yeah.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -209,11 +209,6 @@ private void waitJob() {
           case CANCELED:
           case COMPLETED:
             removed.set(true);
-            try {
-              jobAttempt.close();
-            } catch (IOException e) {
-              LOG.warn("Unable to close job master client", e);
-            }
```
ggezer, line 216: Was this a bug?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
ggezer: Can you make it parallel but put a magic sleep (100ms) at the end.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
ggezer: nevermind, I was thinking this was being done above job-attempt.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
ggezer: if we can know if there is no pending task left after this, then we can even increase the sleep (to 500ms) but avoid when there are pending jobs.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
ggezer: I see.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
bradyoo: No that's not good. The whole point is that this mSubmittedJobAttempts is a running list of 1000 jobs. Once it removes a single one from this list, it escapes this loop to add more to fill 1000.

bradyoo:
reflected feedback

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
ggezer: Can you cache the result of `mClient.getJobStatus(mJobId)` in the job attempt whenever it's Completed/Failed?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -145,7 +147,11 @@ private Status check() {
   public DistributedLoadCommand(FileSystemContext fsContext) {
     super(fsContext);
     mSubmittedJobAttempts = Lists.newArrayList();
-    mClient = null;
+
+    InstancedConfiguration conf = new InstancedConfiguration(ConfigurationUtils.defaults());
+    final ClientContext clientContext = ClientContext.create(conf);
+    mClient = JobMasterClient.Factory.create(
```
ggezer, line 153: Can you override the Command#close() and close this client there?

bradyoo: oops. You are right. I forgot. Done now.

bradyoo:
Done with feedback.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
gpang: This whole structure seems strange? Could we not just have a threadpool/executor service that just runs tasks, one for each path? The generator just goes through the tree and adds tasks. It is strange that there is this spin loop that has to check all jobs?

bradyoo:

bradyoo:
@gpang it could be threadsafe but this is the one that was verified for an issue that we can't really reproduce.

bradyoo:
Furthermore, the parallel bit wasn't even released yet so this isn't really a regression.

bradyoo:
It seems like the path configuration is necessary to pass the DistributedLoadCommandTest. @gpang do you know why?

So I made it such that we use the path configuration of the path that is used in the command.

gpang:
@bradyoo I'm not familiar with that test or how it is related to path configs

bradyoo:
Realized that client context derived from file system context is good enough.

bradyoo:
Ready for review again.

bradyoo:
@gpang Can you merge and cherry-pick for me?

gpang:

gpang:

PR #11574, opened 2020-06-17
Handle escaping for single quotes in EMR script args

Xenorith:


Xenorith:
yeah i barely know what it does either, even with the stackoverflow link. but staring at the bash script execution it seems to be doing the thing i want, which is to preserve all the quotes

ZacBlanco:

PR #11575, opened 2020-06-17
Clean up files in parallel to avoid large recursive delete

gpang:


gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -130,6 +135,76 @@ public void prepare() throws Exception {
...
+          callables.add(() -> {
+            while (!Thread.currentThread().isInterrupted()) {
+              long counter = globalCounter.getAndIncrement();
+              if (counter >= limit) {
+                globalCounter.getAndDecrement();
+                return null;
+              }
+              Path deletePath;
+              if (counter < fixedSize) {
+                deletePath = new Path(fixedBase, Long.toString(counter));
```
bradyoo, line 179: Is this correct for this to happen for every subdir?

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -130,6 +135,76 @@ public void prepare() throws Exception {
...
+              } else {
+                deletePath = new Path(subDir.getPath(), Long.toString(counter));
+              }
+              if (fs.delete(deletePath, true)) {
```
bradyoo, line 183: This design definitely misses some files because the 1 minute could end and the code could be in between globalCounter.getAndIncrement and delete. Is that okay?

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -98,7 +100,10 @@ public void prepare() throws Exception {
 
       if (mParameters.mOperation == Operation.CreateFile
           || mParameters.mOperation == Operation.CreateDir) {
-        prepareFs.delete(basePath, true);
+        long start = CommonUtils.getCurrentMs();
+        deletePaths(prepareFs, basePath);
```
bradyoo, line 104: Should this later be moved to some cleanup step? (As opposed to prepare step)

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -130,6 +135,76 @@ public void prepare() throws Exception {
...
+                deletePath = new Path(subDir.getPath(), Long.toString(counter));
+              }
+              if (fs.delete(deletePath, true)) {
+                success.getAndIncrement();
+              }
+            }
+            return null;
+          });
+        }
+        service.invokeAll(callables, 1, TimeUnit.MINUTES);
```
bradyoo, line 190: Despite what the doc seems to say, invokeAll doesn't necessarily guarantee that all the futures are done. So I think here, we should be checking if all of the futures are done. Log if some are not done and shutdown/reconfigure the service on every loop.

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
```diff
@@ -130,6 +135,76 @@ public void prepare() throws Exception {
...
+        LOG.info("Removed {} files", success.get());
+      }
+    }
+
+    service.shutdownNow();
+    service.awaitTermination(10, TimeUnit.SECONDS);
+
+    // Cleanup the rest recursively, which should be empty or much smaller than the full tree.
+    LOG.info("Deleting base directory: {}", basePath);
+    fs.delete(basePath, true);
```
bradyoo, line 205: I guess this covers the case of skips.

stress/shell/src/main/java/alluxio/stress/cli/StressMasterBench.java
gpang: This is ok. If this times out, it cancels all remaining tasks. That is fine, because the next round will just create more tasks, and any remaining tasks will be taken care of the final recursive delete.

bradyoo:
LGTM

gpang:

PR #11579, opened 2020-06-17
Fix deadlock issue on metadata sync

ZacBlanco:
This deadlock occurs when concurrent clients submit requests to
paths where one is the parent of another. It is most likely to
occur one path is a direct parent of another. There needs to be
less threads in the executor than there are paths to sync.
When the deadlock occurs it is because the child path holds a read
lock on a path the parent may be attempting to sync. This prevents
the parent path from continuing to sync. The child path though may
keep queuing sync jobs which can prevent the child from completing
its sync and thus causing deadlock because the parent can never
finish.

In order to fix this there were two major issues to be solved:

1. Unlock the root path so the ancestry of one sync does not block
another. After internal discussion, we decided that because we
don't hold the lock for any children for the duration of the sync
that it is unnecessary to hold the parent lock for the entire
duration as well.

2. Because the inode sync stream has to lock and unlock the root
path while syncing, the function loadMetadataIfNotExists needed
to be modified so that the path is unlocked before loading the
metadata. The major issue is that checking whether we need to
load metadata on a path is expensive because it requires a lock.
In order to maintain high RPC throughput for GetFileInfo and
ListStatus (the users of loadMetadataIfNotExist) this PR
introduces a two-step scheme for locking on those RPCs. The
general approach is to lock and then check if metadata needs to
be loaded, in which case we unlock and go back to the beginning
of the RPCs main logic with an extra flag set to load metdata.
In the case where the path already exists and does not need to be
loaded, then the RPC continues as normal.

There will be a slight overhead in occasions where we are loading
metadata for paths that don't yet exist, but that is not the
common case, so I am slightly less concerned about the performance
there. There will be no more locks taken there than there would be
in a normal metadata sync, so the runtimes should be comparable
anyway.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -740,7 +740,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
     try (RpcContext rpcContext = createRpcContext()) {
       /*
       In order to prevent locking twice on RPCs where metadata does _not_ need to be loaded, we use
-      a two-phase scheme as an optimization to prevent the extra lock. loadMetadataIfNotExists
+      a two-step scheme as an optimization to prevent the extra lock. loadMetadataIfNotExists
       requires a lock on the tree to determine if the path should be loaded before executing. To
       prevent the extra lock, we execute the RPC as normal and use a conditional check in the
       main body of the function to determine whether control flow should be shifted out of the
@@ -812,17 +812,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
                   .setTtl(context.getOptions().getCommonOptions().getTtl())
                   .setTtlAction(context.getOptions().getCommonOptions().getTtlAction())));
       /*
-      In order to prevent locking twice on RPCs where metadata does _not_ need to be loaded, we use
-      a two-phase scheme as an optimization to prevent the extra lock. loadMetadataIfNotExists
-      requires a lock on the tree to determine if the path should be loaded before executing. To
-      prevent the extra lock, we execute the RPC as normal and use a conditional check in the
-      main body of the function to determine whether control flow should be shifted out of the
-      RPC logic and back to the loadMetadataIfNotExists function.
-
-      If loadMetadataIfNotExists runs, then the next pass into the main logic body should
-      continue as normal. This may present a slight decrease in performance for newly-loaded
-      metadata, but it is better than affecting the most common case where metadata is not being
-      loaded.
+      See the comments in #getFileIdInternal for an explanation on why the loop here is required.
        */
       boolean run = true;
       boolean loadMetadata = false;
@@ -863,8 +853,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
           ret = fileInfo;
         }
       }
-      return Preconditions.checkNotNull(ret,
-          "fileInfo returned should not be null. This is a bug.");
+      return ret;
     }
   }
 
@@ -950,17 +939,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
         context.getOptions().setLoadMetadataType(LoadMetadataPType.NEVER);
       }
       /*
-      In order to prevent locking twice on RPCs where metadata does _not_ need to be loaded, we use
-      a two-phase scheme as an optimization to prevent the extra lock. loadMetadataIfNotExists
-      requires a lock on the tree to determine if the path should be loaded before executing. To
-      prevent the extra lock, we execute the RPC as normal and use a conditional check in the
-      main body of the function to determine whether control flow should be shifted out of the
-      RPC logic and back to the loadMetadataIfNotExists function.
-
-      If loadMetadataIfNotExists runs, then the next pass into the main logic body should
-      continue as normal. This may present a slight decrease in performance for newly-loaded
-      metadata, but it is better than affecting the most common case where metadata is not being
-      loaded.
+      See the comments in #getFileIdInternal for an explanation on why the loop here is required.
        */
       DescendantType loadDescendantType;
       if (context.getOptions().getLoadMetadataType() == LoadMetadataPType.NEVER) {
@@ -1534,7 +1513,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
   }
 
   @Override
-  public Map<String, MountPointInfo> getMountInfo() {
+  public Map<String, MountPointInfo> getMountPointInfoSummary() {
     SortedMap<String, MountPointInfo> mountPoints = new TreeMap<>();
     for (Map.Entry<String, MountInfo> mountPoint : mMountTable.getMountTable().entrySet()) {
       mountPoints.put(mountPoint.getKey(), getDisplayMountPointInfo(mountPoint.getValue()));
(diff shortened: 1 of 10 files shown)


ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
```diff
@@ -234,7 +234,7 @@ long getNewBlockIdForFile(AlluxioURI path) throws FileDoesNotExistException, Inv
   /**
    * @return a copy of the current mount table
    */
-  Map<String, MountPointInfo>  getMountTable();
+  Map<String, MountPointInfo>  getMountInfo();
```
gpang, line 237: Why is this renamed? mount info sounds like it is the info for a single mount, but mount table sounds like the entire mapping?

core/server/master/src/test/java/alluxio/master/file/FileSystemMasterTest.java
```diff
@@ -2672,6 +2673,46 @@ public void propagatePersisted() throws Exception {
             GetStatusContext.defaults()).getPersistenceState());
   }
 
+//  @Test
```
gpang, line 2676: What about this test?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -257,21 +293,18 @@ public boolean sync() {
...
-        | InvalidPathException | IOException e) {
+        | IOException e) {
       LogUtils.warnWithException(LOG, "Failed to sync metadata on root path {}",
           toString(), e);
     } finally {
       // regardless of the outcome, remove the UfsStatus for this path from the cache
-      mStatusCache.remove(mRootPath.getUri());
-      // downgrade so that if operations are parallelized, the lock on the root doesn't restrict
-      // concurrent operations
-      mRootPath.downgradeToRead();
```
gpang, line 274: is this no longer necessary because everything is unlocked?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -245,9 +269,21 @@ public boolean sync() {
     if (LOG.isDebugEnabled()) {
       start = System.currentTimeMillis();
     }
-
-    try {
-      syncInodeMetadata(mRootPath);
+    try (LockedInodePath path = mInodeTree.lockInodePath(mRootScheme)) {
```
gpang, line 272: This will relock and recheck permissions? Can we add that to the comments of this method or class?

core/server/master/src/main/java/alluxio/master/file/FileSystemMaster.java
ZacBlanco: I renamed it because I added another method which actually returns the `MountTable` object. I felt it was more apt to call that `getMountTable`. This was used by a few other places and I couldn't think of a good name. Since it wasn't the actual `MountTable` object I just settled on `getMountInfo`. I'm open to other ideas

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: yes

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: > This will relock and recheck permissions

The permission check and locking were all moved into this block. The `syncMetadata` function in DFSM no longer does the lock and permission check

core/server/master/src/test/java/alluxio/master/file/FileSystemMasterTest.java
ZacBlanco: I was trying to write a test which reproduced the issue consistently, but this one only reproduced it every 1/25 times or so...I can try adding more to make sure that it does reproduce the issue and that this change fixes it, but it might take me a bit longer - or we can just go without this test.

What do you think?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -737,18 +737,48 @@ public long getFileId(AlluxioURI path) throws AccessControlException, Unavailabl
...
+      If loadMetadataIfNotExists runs, then the next pass into the main logic body should
+      continue as normal. This may present a slight decrease in performance for newly-loaded
+      metadata, but it is better than affecting the most common case where metadata is not being
+      loaded.
+       */
+      LoadMetadataContext lmCtx = LoadMetadataContext.mergeFrom(
+          LoadMetadataPOptions.newBuilder().setCreateAncestors(true));
+      boolean run = true;
+      boolean loadMetadata = false;
+      while (run) {
```
gpang, line 758: Is there any chance this could run for more than 2 iterations?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2518,12 +2592,32 @@ public UfsInfo getUfsInfo(long mountId) {
...
-  private void loadMetadataIfNotExist(RpcContext rpcContext, LockedInodePath inodePath,
-      LoadMetadataContext context, boolean isGetFileInfo) {
-    Preconditions.checkState(inodePath.getLockPattern() == LockPattern.READ);
+  private void loadMetadataIfNotExist(RpcContext rpcContext, AlluxioURI path,
+      LoadMetadataContext context, boolean isGetFileInfo)
+      throws InvalidPathException, AccessControlException {
+    LockingScheme scheme = new LockingScheme(path, LockPattern.READ, false);
+    boolean lm;
+    try (LockedInodePath inodePath = mInodeTree.lockInodePath(scheme)) {
+      lm = shouldLoadMetadataIfNotExists(inodePath, context);
```
gpang, line 2603: the rpcs use `shouldLoadMetadataIfNotExists` to check to load metadata, but this method also calls `shouldLoadMetadataIfNotExists` again. Is that intentional, and why do we need to check again?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -769,49 +799,72 @@ public FileInfo getFileInfo(AlluxioURI path, GetStatusContext context)
...
-        if (context.getOptions().getUpdateTimestamps() && context.getOptions().hasAccessMode() && (
-            accessMode.imply(Mode.Bits.READ) || accessMode.imply(Mode.Bits.WRITE))) {
-          mAccessTimeUpdater
-              .updateAccessTime(rpcContext.getJournalContext(), inodePath.getInode(), opTimeMs);
-        }
-        auditContext.setSrcInode(inodePath.getInode()).setSucceeded(true);
-        return fileInfo;
       }
+      return Preconditions.checkNotNull(ret,
+          "fileInfo returned should not be null. This is a bug.");
```
gpang, line 867: Is there a better message we can use, to help us debug or fix?

core/server/master/src/test/java/alluxio/master/file/FileSystemMasterTest.java
ZacBlanco: even with a longer path is failed to reproduce. To writing the actual test that fails (deadlocks) in the earlier code and not in the current code might take me a while to write.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -769,49 +799,72 @@ public FileInfo getFileInfo(AlluxioURI path, GetStatusContext context)
...
-        } catch (AccessControlException e) {
-          auditContext.setAllowed(false);
-          throw e;
+      LoadMetadataContext lmCtx = LoadMetadataContext.mergeFrom(
+          LoadMetadataPOptions.newBuilder().setCreateAncestors(true).setCommonOptions(
+              FileSystemMasterCommonPOptions.newBuilder()
+                  .setTtl(context.getOptions().getCommonOptions().getTtl())
+                  .setTtlAction(context.getOptions().getCommonOptions().getTtlAction())));
+      /*
+      In order to prevent locking twice on RPCs where metadata does _not_ need to be loaded, we use
```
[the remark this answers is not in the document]
ZacBlanco: Yeah, I can make it so there's only one comment. I just wanted to be extra clear in where I was using it.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: Do you see a case where it could? 

I stepped through the logic myself to iron out as many errors as possible when writing it and when doing the tests.

For each of the implementations there are two flags 
- `run`
- `loadMetadata`


In the main body of the loop `run` is always immediately set `false` after entering the loop and is only set again within a conditional that contains `if (!loadMetadata)`

Within the conditional of `if (!loadMetadata)` we set `run` to `true`, but also set the `loadMetadata` flag to true, which means that we shouldn't enter the same conditional the 2nd time around to be able to set `run` to true again - preventing more than 2 iterations.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: It was originally intentional because I was trying to preserve the old behavior, and it's possible that in between the original check and running this function the tree changed, so this can prevent us from loading the metadata unnecessarily

But I am thinking more about it now; the cost of the lock to perform this check vs the probability of the path being loaded since the last check outside this function is likely low, and so the cost of performing this lock vs the amount of CPU time it _could_ save is likely going to do more harm than good. I'll remove it.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: Oh so, I guess IntelliJ became smart again - after looking at this code the IDE is reporting to me that "ret is always non-null" at this point. I guess I can remove this Precondition check.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ZacBlanco: If you have a better idea of how you would implement this logic I'm all ears. I'm sure there are ways it can be improved. This was the simplest thing I came up with.

gpang: Yeah, after looking at it, it should be fine. The important part is that we have to make sure run is set to false in the beginning, and setting `run=true` can only happen when `loadMetadata == false`.

ZacBlanco:
jenkins, test this please

ZacBlanco:

PR #11580, opened 2020-06-17
Add more traces for state-lock management

ggezer:


core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -123,6 +124,9 @@ public void mastersStartedCallback() {
    * @throws InterruptedException
    */
   public LockResource lockShared() throws InterruptedException {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Thread-%d entered lockShared().", Thread.currentThread().getId());
```
gpang, line 128: do you also want thread name, or just the id? Also, I think for logging you need "{}", not "%d". "%d" I think is for `String.format()`.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -153,11 +157,14 @@ public LockResource lockShared() throws InterruptedException {
    */
   public LockResource lockExclusive(StateLockOptions lockOptions)
       throws TimeoutException, InterruptedException {
+    LOG.debug("Thread-%d entered lockExclusive().", Thread.currentThread().getId());
```
gpang, line 160: why is the other one guarded by `LOG.isDebugEnabled()`?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -153,11 +157,14 @@ public LockResource lockShared() throws InterruptedException {
...
     // Run the grace cycle.
     StateLockOptions.GraceMode graceMode = lockOptions.getGraceMode();
     boolean lockAcquired = false;
     long deadlineMs = System.currentTimeMillis() + lockOptions.getGraceCycleTimeoutMs();
     while (System.currentTimeMillis() < deadlineMs) {
+      LOG.debug("Thread-%d entered grace-cycle of try-sleep: %ms-%ms",
```
gpang, line 166: I don't think this syntax is correct. It doesn't have enough arguments, and is "%ms" correct?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -169,6 +176,7 @@ public LockResource lockExclusive(StateLockOptions lockOptions)
       }
     }
     if (lockAcquired) { // Lock was acquired within grace-cycle.
+      LOG.debug("Thread-%d acquired the lock within grace-cycle.", Thread.currentThread().getId());
```
gpang, line 179: Could this be useful for INFO level?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -169,6 +176,7 @@ public LockResource lockExclusive(StateLockOptions lockOptions)
...
       activateInterruptCycle();
     } else { // Lock couldn't be acquired by grace-cycle.
       if (graceMode == StateLockOptions.GraceMode.TIMEOUT) {
```
gpang, line 182: Could it be useful to log something here at INFO, since we were not able to get the lock?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -247,6 +260,9 @@ private void deactivateInterruptCycle() {
    * Scheduled routine that interrupts waiters/holders of shared lock.
    */
   private void waiterInterruptRoutine() {
+    LOG.debug("Interrupt-cycle interrupting %d waiters/holders: {%s}",
+        Thread.currentThread().getId(), mSharedWaitersAndHolders.size(), mSharedWaitersAndHolders
+            .stream().map((th) -> Long.toString(th.getId())).collect(Collectors.joining(",")));
```
gpang, line 265: this may not be accurate, since the real interrupt happens in the loop below.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: Right. I now made many of them as INFO.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ZacBlanco, line 265: I think that we might want to wrap this in a `if (Log.isDebugEnabled())` because grabbing iterating, and joining into a String can be pretty expensive?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -178,6 +186,9 @@ public LockResource lockExclusive(StateLockOptions lockOptions)
       // Activate the interrupt cycle before entering the lock because it might wait in the queue.
       activateInterruptCycle();
       // Force the lock.
+      LOG.debug("Thread-%d forcing the lock with %d waiters/holders: {%s}",
+          Thread.currentThread().getId(), mSharedWaitersAndHolders.size(), mSharedWaitersAndHolders
+              .stream().map((th) -> Long.toString(th.getId())).collect(Collectors.joining(",")));
```
ZacBlanco, line 191: I think that we might want to wrap this in a if (Log.isDebugEnabled()) because grabbing iterating, and joining into a String can be pretty expensive?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: Bumped it down for consistent logging.

ggezer: I made it INFO now. We don't want to miss this trace when it happens.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: I made it INFO now. This is OK to be seen imo.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ZacBlanco: How often do we expect this method to run?

ggezer: every 30secs by default only if it is activated under certain conditions.

core/common/src/main/java/alluxio/util/ThreadUtils.java
```diff
@@ -77,5 +77,20 @@ public static void shutdownAndAwaitTermination(ExecutorService pool, long timeou
...
+  public static String getCurrentThreadIdentifier() {
+    return getThreadIdentifier(Thread.currentThread());
+  }
+
+  /**
+   * @param thread the thread
+   * @return a string representation of the given thread
+   */
+  public static String getThreadIdentifier(Thread thread) {
+    return String.format("{}({})", thread.getId(), thread.getName());
```
gpang, line 92: I thought format used "%s" and "%d"?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ZacBlanco: ok, I think it's fine then. I just wanted to make sure it doesn't need to get computed frequently.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -153,11 +158,15 @@ public LockResource lockShared() throws InterruptedException {
    */
   public LockResource lockExclusive(StateLockOptions lockOptions)
       throws TimeoutException, InterruptedException {
+    LOG.debug("Thread-{} entered lockExclusive().", ThreadUtils.getCurrentThreadIdentifier());
     // Run the grace cycle.
     StateLockOptions.GraceMode graceMode = lockOptions.getGraceMode();
     boolean lockAcquired = false;
     long deadlineMs = System.currentTimeMillis() + lockOptions.getGraceCycleTimeoutMs();
     while (System.currentTimeMillis() < deadlineMs) {
+      LOG.info("Thread-{} entered grace-cycle of try-sleep: {}ms-{}ms for the total of {}ms",
```
gpang, line 167: This will print every 10 minutes for 12 hours? Is that too noisy?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -169,6 +178,8 @@ public LockResource lockExclusive(StateLockOptions lockOptions)
       }
     }
     if (lockAcquired) { // Lock was acquired within grace-cycle.
+      LOG.info("Thread-{} acquired the lock within grace-cycle.",
+          ThreadUtils.getCurrentThreadIdentifier());
       activateInterruptCycle();
     } else { // Lock couldn't be acquired by grace-cycle.
       if (graceMode == StateLockOptions.GraceMode.TIMEOUT) {
```
gpang, line 185: Hrmmm, is this right? I thought TIMEOUT mode would not force the lock by interrupting threads?

ggezer: no we have to interrupt during the lock. not just when forcing the lock.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: made it one time.

ggezer:
alluxio-bot, test this please

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: If the lock is acquired by the grace-cycle, we still need to run the interrupt-cycle for maintaining the heart-beats.

TIMEOUT and FORCED make sense when the lock was not grabbed by the grace-cycle. In this context FORCED would mean to activate the interrupt-cycle before taking the lock.

ggezer:

ggezer:

ggezer:

PR #11583, opened 2020-06-17
Prevent adding conf on local start if it exists

ZacBlanco:
Previously, if debug options were enabled on the JVM via java options like ALLUXIO_MASTER_JAVA_OPTS, then alluxio getConf would not return the expected result causing `alluxio.master.hostname` to be added to `alluxio-site.properties` multiple times. This change allows debug options to be enabled on the JVM without filling up the alluxio-site.properties with a duplicated line.

There is one catch: If someone has `alluxio.master.hostname` set in `alluxio-env.sh`, then this would still be added exactly once to the site properties, but the system property would still take


ZacBlanco:
jenkins, test this please

ZacBlanco:
alluxio-bot, test this please

ZacBlanco:

PR #11587, opened 2020-06-17
Use exec in runJavaClass

ZacBlanco:
Using `exec` on the `runJavaClass` function is a better practice to ensure that signals like `ctrl-c` and `ctrl-z` get sent to the actual java process running and not the bash shell which launched the process.

### DIFF
--- a/bin/alluxio
+++ b/bin/alluxio
@@ -154,7 +154,7 @@ function runJavaClass {
               CLASS_ARGS+=("${arg}")
       esac
   done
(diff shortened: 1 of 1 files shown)


ZacBlanco:
jenkins, test this please

apc999:
is there any side effect? e.g., in cases of error is thrown in JVM? or any implication in docker environment?

ZacBlanco:
This is a strict improvement in the way that we use `runJavaClass`. `exec` is basically replaces the process executing the command with the new process. Otherwise you have a bash shell wrapping the java process which is undesirable in the way we use `runJavaClass` - signals get sent to the bash process which don't always make it to the java process properly. Using exec makes sure signals sent via `ctrl-c` are made to the java process directly rather than through a shell process.

Usage of the `runJavaClass` function is always the last step of any part of the script that calls the function, so it's okay to add `exec` as it won't affect the way we use the function.

ZacBlanco:
ping

PR #11588, opened 2020-06-17
Implement interrupt based locking protocol over state-lock

ggezer:
This PR, in general, introduces a new locking framework over state-lock.
Changes can be summarized as:

- Backups take the lock exclusive with an initial grace-cycle.
Grace-cycle is defined as `tryLock-sleep` cycles until the lock is held.
3 graceful locking strategies have been implemented (TIMEOUT, FORCED)

- Immediate shared holders/waiters of state-lock may now be interrupted
as part of interrupt-cycle task, which is activated before or after
taking the lock.

- Shell/daily backups can now be configured for all grace modes via
master config change. (Defaults for Shell: TIMEOUT, Daily: FORCED)

- A new maintenance duration( exclusive-only phase) is defined for
masters, during which only exclusive locking will be allowed. This
duration is effective after the first time masters are started, after
reading the journals.

pr-link: Alluxio/alluxio#11535
change-id: cid-b0ca6b03b8b9f563ad537ebd14991ce4fe1cacf5

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1283,6 +1283,41 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.MASTER)
           .build();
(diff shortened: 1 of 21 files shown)


ggezer:
alluxio-bot, test this please

ggezer:

PR #11592, opened 2020-06-18
Add a state-lock call-tracker for detecting interrupt-cycles

ggezer:
This change introduces a new FSM call-tracker that is attached to select recursive APIs (`list`/`setAttributes`/`setAcl`) in order for them to respect a pending backup and bail.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/StateLockManager.java
+++ b/core/server/common/src/main/java/alluxio/master/StateLockManager.java
@@ -32,6 +32,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -65,7 +66,7 @@ public class StateLockManager {
   /** Used to synchronize execution/termination of interrupt-cycle. */
   private Lock mInterruptCycleLock = new ReentrantLock(true);
   /** How many active exclusive locking attempts. */
-  private volatile long mInterruptCycleRefCount = 0;
+  private AtomicInteger mInterruptCycleRefCount = new AtomicInteger(0);
   /** The future for the active interrupt cycle. */
   private ScheduledFuture<?> mInterrupterFuture;
 
@@ -223,7 +224,7 @@ public class StateLockManager {
    * @return the call tracker
    */
   public CallTracker getInterruptCycleTracker() {
-    return () -> mInterruptCycleRefCount > 0;
+    return () -> mInterruptCycleRefCount.get() > 0;
   }
 
   /**
@@ -240,7 +241,7 @@ public class StateLockManager {
     }
     try (LockResource lr = new LockResource(mInterruptCycleLock)) {
       // Don't reschedule if it was before.
-      if (mInterruptCycleRefCount++ > 0) {
+      if (mInterruptCycleRefCount.incrementAndGet() > 0) {
         return;
       }
       // Setup the cycle.
@@ -261,9 +262,9 @@ public class StateLockManager {
       return;
     }
     try (LockResource lr = new LockResource(mInterruptCycleLock)) {
-      Preconditions.checkArgument(mInterruptCycleRefCount > 0);
+      Preconditions.checkArgument(mInterruptCycleRefCount.get() > 0);
       // Don't do anything if there are exclusive lockers.
-      if (--mInterruptCycleRefCount > 0) {
+      if (mInterruptCycleRefCount.decrementAndGet() > 0) {
         return;
       }
       // Cancel the cycle.
--- a/core/server/common/src/main/java/alluxio/master/StateLockManager.java
+++ b/core/server/common/src/main/java/alluxio/master/StateLockManager.java
@@ -218,6 +218,15 @@ public class StateLockManager {
   }
 
   /**
+   * Used to generate a call-tracker that knows when an interrupt cycle is active.
+   *
+   * @return the call tracker
+   */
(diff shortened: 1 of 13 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -2807,6 +2809,7 @@ public void setAcl(AlluxioURI path, SetAclAction action, List<AclEntry> entries,
       SetAclContext context)
       throws FileDoesNotExistException, AccessControlException, InvalidPathException, IOException {
     Metrics.SET_ACL_OPS.inc();
+    addFileSystemMasterCallTrackers(context);
```
gpang, line 2812: Can we do this for recursive delete?

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -23,41 +29,50 @@
...
-   * @return underlying proto message instance
+   * Adds a new call-tracker to this context.
+   *
+   * @param callTracker the call tracker
    */
-  public T getOptions() {
-    return mOptionsBuilder;
+  public void addCallTracker(CallTracker callTracker) {
+    synchronized (mCallTrackers) {
+      mCallTrackers.add(Preconditions.checkNotNull(callTracker));
```
gpang, line 58: We should specify the parameter name:

```suggestion
      mCallTrackers.add(Preconditions.checkNotNull(callTracker, "callTracker"));
```

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -23,41 +29,50 @@
...
   }
 
   /**
    * @return {@code true} if the call is cancelled by the client
    */
   public boolean isCancelled() {
-    return mCallTracker.isCancelled();
+    synchronized (mCallTrackers) {
+      if (mCallTrackers.isEmpty()) {
+        throw new IllegalStateException("No tracker registered.");
```
gpang, line 68: Why does this throw an Exception, and not just return `false`?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -217,6 +217,15 @@ public LockResource lockExclusive(StateLockOptions lockOptions)
     });
   }
 
+  /**
+   * Used to generate a call-tracker that knows when an interrupt cycle is active.
+   *
+   * @return the call tracker
+   */
+  public CallTracker getInterruptCycleTracker() {
+    return () -> mInterruptCycleRefCount > 0;
```
gpang, line 226: Does this create a new CallTracker each time? Is there a way to have a single instance of the call tracker?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3535,6 +3539,17 @@ public void stopSync(AlluxioURI syncPoint)
...
 
+  /**
+   * Used to augment given context with fsm bound call-trackers
+   * TODO(ggezer): Call this for each call and filter by a new method annotation (@Tracked).
+   *
+   * @param context the operation context
+   */
+  private void addFileSystemMasterCallTrackers(OperationContext context) {
+    // Add state-lock interrupt-cycle call-tracker to given context.
+    context.addCallTracker(mMasterContext.getStateLockManager().getInterruptCycleTracker());
```
gpang, line 3550: What is preventing us from just adding the tracker to every single context? I thought it is up to the individual rpc to use it or not?

ggezer: The reasoning was to track tracked-APIs by explicitly working on them.
Currently, if an API decides to use the call-tracker, it must have been explicitly attached otherwise an exception is raised.

I already have a todo in this method to make it general.

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
ggezer: explained on another comment.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ggezer: Since deleteInternal() is used by many, I believe we can't go with fail-if-not-tracked approach anymore. So I might make this method called by all APIs as well.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
ggezer: singleton now.

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
ggezer: n/a

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
ggezer: done.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4261,6 +4271,11 @@ private void removeBlocks(List<Long> blocks) throws IOException {
     throw new IOException("Failed to remove deleted blocks from block master", lastThrown);
   }
 
+  @Override
+  public CallTracker composeCallTracker(CallTracker transportTracker) {
```
gpang, line 4275: This is allowed to be `@Nullable`, right? We should handle that case.

ggezer: removed nullable.

ggezer:
alluxio-bot, test this please

ggezer:

ggezer:

PR #11596, opened 2020-06-18
Fix fetching partition for Glue UDB

HelloHorizon:
Glue `getPartions` fetches a random number of partition from glue service without setting the `MaxResults` and `NextToken`. Add the `withMaxResults` and `withNextToken` to fix the partition fetching for Glue UDB.

### DIFF
--- a/table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
+++ b/table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
@@ -42,6 +42,7 @@ import com.amazonaws.services.glue.model.EntityNotFoundException;
 import com.amazonaws.services.glue.model.GetDatabaseRequest;
 import com.amazonaws.services.glue.model.GetDatabaseResult;
 import com.amazonaws.services.glue.model.GetPartitionsRequest;
+import com.amazonaws.services.glue.model.GetPartitionsResult;
 import com.amazonaws.services.glue.model.GetTableRequest;
 import com.amazonaws.services.glue.model.GetTablesRequest;
 import com.amazonaws.services.glue.model.GetTablesResult;
@@ -377,15 +378,26 @@ public class GlueDatabase implements UnderDatabase {
 
   private List<Partition> batchGetPartitions(AWSGlueAsync glueClient, String tableName)
       throws IOException {
+    // TODO(shouwei): make getPartition multi-thread to accelerate the large table fetching
     List<Partition> partitions = new ArrayList<>();
+    String nextToken = null;
     try {
-      GetPartitionsRequest getPartitionsRequest =
-          new GetPartitionsRequest()
-              .withCatalogId(mGlueConfiguration.get(Property.CATALOG_ID))
-              .withDatabaseName(mGlueDbName)
-              .withTableName(tableName);
-      if (glueClient.getPartitions(getPartitionsRequest).getPartitions() != null) {
-        partitions = glueClient.getPartitions(getPartitionsRequest).getPartitions();
+      do {
+        GetPartitionsRequest getPartitionsRequest =
+            new GetPartitionsRequest()
+                .withCatalogId(mGlueConfiguration.get(Property.CATALOG_ID))
+                .withDatabaseName(mGlueDbName)
+                .withTableName(tableName)
+                .withMaxResults(mGlueConfiguration.getInt(Property.MAX_GLUE_FETCH_PARTITIONS))
+                .withNextToken(nextToken);
+        GetPartitionsResult getPartitionsResult = glueClient.getPartitions(getPartitionsRequest);
+        getPartitionsResult.getPartitions().forEach(partition -> partitions.add(partition));
+        nextToken = getPartitionsResult.getNextToken();
+        LOG.debug("Glue table {}.{} added {} partitions.",
+            mGlueDbName, tableName, partitions.size());
+      } while (nextToken != null);
+
+      if (partitions != null) {
         LOG.info("Glue table {}.{} has {} partitions.",
             mGlueDbName, tableName, partitions.size());
         if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
```diff
@@ -377,15 +378,26 @@ public UdbTable getTable(String tableName) throws IOException {
...
+      do {
+        GetPartitionsRequest getPartitionsRequest =
+            new GetPartitionsRequest()
+                .withCatalogId(mGlueConfiguration.get(Property.CATALOG_ID))
+                .withDatabaseName(mGlueDbName)
+                .withTableName(tableName)
+                .withMaxResults(mGlueConfiguration.getInt(Property.MAX_GLUE_FETCH_PARTITIONS))
+                .withNextToken(nextToken);
+        GetPartitionsResult getPartitionsResult = glueClient.getPartitions(getPartitionsRequest);
+        getPartitionsResult.getPartitions().forEach(partition -> partitions.add(partition));
```
gpang, line 394: can this be something like `partitions.addAll(getPartitionsResult.getPartitions()`?

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
```diff
@@ -377,15 +378,26 @@ public UdbTable getTable(String tableName) throws IOException {
...
+        nextToken = getPartitionsResult.getNextToken();
+        LOG.debug("Glue table {}.{} added {} partitions.",
+            mGlueDbName, tableName, partitions.size());
```
gpang, line 397: should this be `getPartitionsResult.getPartitions().size()` instead of `partitions.size()`?

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
```diff
@@ -142,11 +142,17 @@ public static void unregister(Property Property) {
           .setDescription("The maximum number of connection to glue metastore.")
           .build();
 
+  public static final Property MAX_GLUE_FETCH_PARTITIONS =
+      new Builder(Name.MAX_GLUE_FETCH_PARTITIONS)
+          .setDefaultValue("256")
```
gpang, line 147: Is this large enough? Do you know if we can make this larger while still allowing it to return the correct amount?

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
```diff
@@ -197,6 +203,7 @@ public String getDefaultValue() {
   public static final class Name {
     // AWS Glue related properties
     public static final String MAX_GLUE_CONNECTION = "max.connection";
+    public static final String MAX_GLUE_FETCH_PARTITIONS = "max.partitions";
```
gpang, line 206: NIT: maybe `partitions.fetch.max`

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
HelloHorizon: It allows us to track the progress of `getPartitions`. Is it better to show the total added partitions for the current table?

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
HelloHorizon: This value means the maximum number of partitions to return in a single response, from 1 to 1000, 256 should large enough for most cases. I will update the explanation of this property.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
HelloHorizon: changed.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
HelloHorizon: changed.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
gpang: If we set it to 1000, is it slow? I think setting it larger would be better.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
gpang: I think both can be useful. How about something like "Glue table {}.{} adding {} batch partitions with {} total partitions."

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
HelloHorizon: I don't know the performance difference with an accurate number, but it should has little difference.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
HelloHorizon: yeah, it makes sense, I can add both of them.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
HelloHorizon: I changed the default value to 512, which should more than enough for most cases.

HelloHorizon: I changed it to 512, it still only fetch 460 partitions every time, it would not reach the limitation.

HelloHorizon:
jenkins, test this please

gpang:

PR #11597, opened 2020-06-19
fix bug in meta cache in MetadataCachingBaseFileSystem

zqxjjj:
Description:
meta data may be deleted in client cache although it is very hot. 
Now MetadataCache holds mStatus which stores the current file/dir status and mDirStatuses which stores the children status.
```
  private class CachedItem {
    private final URIStatus mStatus;
    private final List<URIStatus> mDirStatuses;
```
1. For example, There is a path '/A/B/C'.
2. We run the listStatus request on the directory '/A/B'. After receiving result from alluxio master, it caches 'A/B' into the cache map which stores the children status for the path '/A/B' as mDirStatuses and its children '/A/B/C' into the cach map which stores the file/dir status as mStatus. 
3. Then we run the listStatus request on the directory '/A'. After receiving result from alluxio master, it caches '/A' into the cache map which stores the children status for the path '/A' as mDirStatuses  and its children '/A/B' into the cach map which stores the file/dir status as mStatus. It replaces the previous content with a new CachedItem for '/A/B'.
```
public void put(AlluxioURI dir, List<URIStatus> statuses) {
    mCache.put(dir.getPath(), new CachedItem(statuses));
    for (URIStatus status : statuses) {
      mCache.put(status.getPath(), new CachedItem(status));
    }
```
4. If you rerrun the listStatus request on the directory '/A/B', you could not get the result. Because although the path '/A/B' is in the cache, it holds the content with mStatus. And the CachedItem with mDirStatuses has been replaced in the step 3.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/MetadataCache.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/MetadataCache.java
@@ -19,6 +19,7 @@ import com.google.common.cache.CacheBuilder;
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.Callable;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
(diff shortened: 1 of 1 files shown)


zqxjjj:
Duplicate of # 11598

PR #11600, opened 2020-06-19
Implement utility for dedicated tracing of well-known events

ggezer:


core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
```diff
@@ -0,0 +1,78 @@
+package alluxio.client.file;
```
gpang, line 1: Why is this in "client"? Shouldn't this be in server?

Also, it probably also shouldn't be in "file" either, since these are really server level events, not file system events.

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
```diff
@@ -0,0 +1,78 @@
...
+ * Defines and Manages well-known Alluxio events.
+ *
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ * !!! - Keep increasing the event-id for new events. !!!
+ * !!! - Don't reassign event ids.                    !!!
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ */
+public enum AlluxioEvent {
+  // Alluxio master events.
+  Master_Process_Starting(1, EventType.INFO, "Master process starting."),
```
gpang, line 16: I feel like these should either be camelcase, or all caps with underscores.

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
```diff
@@ -0,0 +1,78 @@
...
+        LOG.info(eventString);
+        break;
+      case ERROR:
+        LOG.error(eventString);
+        break;
+      case WARNING:
+        LOG.warn(eventString);
+        break;
+      case DIAGNOSTIC:
+        LOG.debug(eventString);
```
gpang, line 64: shouldn't the event log always be logging? Or is the idea that this log always enables debug?

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
```diff
@@ -0,0 +1,78 @@
...
+  Master_Process_Stopping(2, EventType.INFO, "Master process stopping."),
+  // Alluxio worker events.
+  Worker_Process_Starting(3, EventType.INFO, "Worker process started."),
+  Worker_Process_Stopping(4, EventType.INFO, "Worker process stopping.");
+
+  // logger.
+  private static final Logger LOG = LoggerFactory.getLogger(AlluxioEvent.class);
+
+  /** The unique event id. */
+  private final int mId;
```
gpang, line 26: What is the id for? Wouldn't the name of the enum be enough?

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Fires the event.
+   *
+   * @param args arguments to event message string
+   */
+  public void fire(Object... args) {
+    String msgString = String.format(mMsgFormat, args);
+    String eventString = String.format("Id: %d, Name: %s, Message:%s", mId, this.name(), msgString);
```
gpang, line 52: Isn't the name of the enum enough of an id rather than the id int?

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
ggezer: I tried all caps and it looked intrusive. 
I will try camel casing.

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
ggezer: Numeric IDs are easier to communicate imo. So that's why I took the trouble to define it.

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
ggezer: Since we are using log4j as backend, it seemed good to leverage logging levels. If confusing we can call `log` for  info/warn/error. However `debug` should be left for noisy diagnostic events.

core/client/fs/src/main/java/alluxio/client/file/AlluxioEvent.java
ggezer: camel casing is better.

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,92 @@
...
+    mMsgFormat = msgFormat;
+  }
+
+  /**
+   * Fires the event.
+   *
+   * @param args arguments to event message string
+   */
+  public void fire(Object... args) {
+    String msgString = String.format(mMsgFormat, args);
```
gpang, line 65: I feel like instead of treating the event message as a format, we should just use the input as additional info/text for the event.

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,92 @@
...
+        LOG.info(eventString);
+        break;
+      case ERROR:
+        LOG.error(eventString);
+        break;
+      case WARNING:
+        LOG.warn(eventString);
+        break;
+      case DIAGNOSTIC:
+        LOG.debug(eventString);
```
gpang, line 78: Isn't the event log going to a separate file?

In any case, events should always be logged, so I don't think any event should be at debug level. The goal for events is to see every event that occurs. Events logging should just not be noisy or extraneous.

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,92 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Defines and Manages well-known Alluxio events.
+ *
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ * !!! - Keep increasing the event-id for new events. !!!
+ * !!! - Don't reassign event ids.                    !!!
```
gpang, line 22: We need a unit test which inspects all ids programmatically to validate there are no duplicates.

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,92 @@
...
+package alluxio;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Defines and Manages well-known Alluxio events.
+ *
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ * !!! - Keep increasing the event-id for new events. !!!
```
gpang, line 21: I don't think this is as important, since it may make sense to partition the id space for different types of events. Like if we do event logging on the worker, maybe those are in a different segment of the id space.

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,92 @@
...
+ * !!! - Don't reassign event ids.                    !!!
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ */
+public enum AlluxioEvent {
+  // Alluxio master events.
+  MasterProcessStarting(1000, EventType.INFO, "Master process starting."),
+  JournalSystemStarted(1001, EventType.INFO, "Journal system started. %s"),
+  MasterIsPrimary(1002, EventType.INFO, "Master process is now the primary."),
```
gpang, line 29: We may need to be more precise about these events. For example, I'd imagine the sequence of events could be the following. I don't know how these events map to this sequence:
process started -> journal started -> primacy gained -> serving started -> ... -> primacy lost -> serving stopped -> process stopped

Also, something like "MasterIsPrimary" seems more like a state instead of an event. This could be "primacy gained" if that is what this means.

We may also want to be precise of past tense and present tense. For example, "process starting" is different from "process started". I feel like past tense might be more clear what it means. Like, "journal catchup started" -> "journal catchup finished".

ggezer: Incorporated these to new event names. PTAL.

ggezer:
alluxio-bot, test this please

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,94 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ * Defines and Manages well-known Alluxio events.
+ */
+public enum AlluxioEvent {
+  // [1000,2000) Alluxio process events.
```
gpang, line 24: does this include job processes too?

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,94 @@
...
+  MasterProcessCreated(1000, "Alluxio master process created."),
+  JournalSystemStarted(1001, "Journal system started."),
```
gpang, line 26: Will there be an example of when the message is different from the name of the enum? It seems like every event so far has a message which just mirrors the enum name.

core/server/common/src/main/java/alluxio/master/AbstractPrimarySelector.java
```diff
@@ -50,6 +51,8 @@
   private State mState = State.SECONDARY;
 
   protected final void setState(State state) {
+    AlluxioEvent.MasterIsTransitioning
+        .fire(String.format("Old-State: %s, New-State: %s", mState, state));
```
gpang, line 55: fire already joins multiple arguments, so should this be split up into multiple strings?

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Fires the event.
+   *
+   * @param additionalInfo additional information to include in the event
+   */
+  public void fire(Object... additionalInfo) {
+    String eventString = String.format("Id: %d, Name: %s, Message:%s, AdditionalInfo: %s",
+        mId, this.name(), mMessage,
+        Arrays.stream(additionalInfo).map(Object::toString).collect(Collectors.joining(",")));
```
gpang, line 91: I think there should be a space after the comma?

core/server/master/src/main/java/alluxio/master/backup/BackupLeaderRole.java
```diff
@@ -183,6 +185,7 @@ public BackupStatus backup(BackupPRequest request, StateLockOptions stateLockOpt
       // Store backup id to query later for async requests.
       backupId = mBackupTracker.getCurrentStatus().getBackupId();
     }
+    AlluxioEvent.BackupStarted.fire(backupId);
```
gpang, line 188: should this have a label for the id, like `String.format("backupId: %d", backupId)`

core/server/master/src/main/java/alluxio/master/backup/BackupTracker.java
```diff
@@ -221,8 +222,10 @@ private void signalIfFinished() {
 
     if (mBackupStatus.isCompleted()) {
       mCompletion.set(null);
+      AlluxioEvent.BackupFinished.fire(mBackupStatus.getBackupId());
```
gpang, line 225: we should label the int values in the output

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,94 @@
...
+    return mMessage;
+  }
+
+  /**
+   * Fires the event.
+   *
+   * @param additionalInfo additional information to include in the event
+   */
+  public void fire(Object... additionalInfo) {
+    String eventString = String.format("Id: %d, Name: %s, Message:%s, AdditionalInfo: %s",
```
gpang, line 89: How do you feel about an output like "EventName(EventId): Message. AdditionalInfo"? (We may or may not need message) Being concise could be more readable.

core/common/src/main/java/alluxio/AlluxioEvent.java
ggezer: i didn't intend to.

core/common/src/main/java/alluxio/AlluxioEvent.java
ggezer: yeah seems redundant till now.

ggezer:
jenkins, test this please

ggezer:
@gpang Added log4J configuration and reallocated event ids.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -666,4 +670,10 @@ public void format() throws IOException {
   public PrimarySelector getPrimarySelector() {
     return mPrimarySelector;
   }
+
+  @Override
+  public String toString() {
+    return String.format("RaftJournalSystem, in quorum: {%s}", mConf.getClusterAddresses().stream()
```
gpang, line 676: Should we just use `MoreObjects.toStringHelper()` for this?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -212,4 +216,9 @@ public void checkpoint() throws IOException {
       journal.checkpoint();
     }
   }
+
+  @Override
+  public String toString() {
+    return String.format("UfsJournalSystem, based at %s", mBase.toString());
```
gpang, line 222: Should we just use `MoreObjects.toStringHelper()` for this?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -805,9 +807,20 @@ private void freeSpaceInternal(long sessionId, long minContiguousBytes, long min
...
       LOG.error(
-          "Failed to free space. Min contiguous requested: {}, Min available requested: {}, "
+          "Failed to free space in session: {}. "
+              + "Min contiguous requested: {}, Min available requested: {}, "
               + "Blocks iterated: {}, Blocks removed: {}, " + "Space freed: {}",
-          minContiguousBytes, minAvailableBytes, blocksIterated, blocksRemoved, spaceFreed);
+          sessionId, minContiguousBytes, minAvailableBytes,
+          blocksIterated, blocksRemoved, spaceFreed);
+
+      AlluxioEvent.BlockStoreEvictionFailed.fire(
```
gpang, line 816: Is it possible that this is a noisy event? Would this be considered an event?

conf/log4j.properties
```diff
@@ -181,5 +181,16 @@ log4j.appender.FUSE_LOGGER.MaxBackupIndex=10
 log4j.appender.FUSE_LOGGER.layout=org.apache.log4j.PatternLayout
 log4j.appender.FUSE_LOGGER.layout.ConversionPattern=%d{ISO8601} %-5p %c{1} - %m%n
 
+# Appender for Alluxio events
+log4j.appender.EVENT_LOGGER=org.apache.log4j.RollingFileAppender
+log4j.appender.EVENT_LOGGER.File=${alluxio.logs.dir}/events.log
```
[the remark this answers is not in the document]
ggezer: This is mostly to aid production cluster investigations.
Even when not, I think seeing logs in a single file is not a problem.

ggezer:
jenkins, test this please

conf/log4j.properties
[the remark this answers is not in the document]
gpang: Is it possible/correct for multiple processes to write to the same log file?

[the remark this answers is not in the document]
ggezer: I removed the worker events as I already have the exact same information in worker logs.
We can later figure out the plumbing for log4j If we want to emit events from other processes.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -666,4 +671,12 @@ public void format() throws IOException {
   public PrimarySelector getPrimarySelector() {
     return mPrimarySelector;
   }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("Quorum-Members", mConf.getClusterAddresses().stream()
```
gpang, line 678: many other labels don't have dashes. Should this be "QuorumMembers"?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -212,4 +217,11 @@ public void checkpoint() throws IOException {
       journal.checkpoint();
     }
   }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("Ufs-Uri", mBase)
```
gpang, line 224: should this be "UfsUri"?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
```diff
@@ -78,6 +80,7 @@ public ActiveSyncer(FileSystemMaster fileSystemMaster, ActiveSyncManager syncMan
   @Override
   public void heartbeat() {
     LOG.debug("start sync heartbeat for {} with mount id {}", mMountUri, mMountId);
+    AlluxioEvent.ActiveSyncStarted.fire();
```
gpang, line 83: I feel like these active sync events will be noisy. Do we need them now?

core/server/common/src/main/java/alluxio/master/AbstractPrimarySelector.java
```diff
@@ -50,6 +52,8 @@
   private State mState = State.SECONDARY;
 
   protected final void setState(State state) {
+    AlluxioEvent.MasterIsTransitioning.fire(
+        new Pair<>("Old-State", mState), new Pair<>("New-State", state));
```
gpang, line 56: should this be "OldState" and "NewState"?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
ggezer: why theyb would be noisy?

ggezer:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
gpang: Isn't it syncing on a heartbeat? We would see this event on every tick, which could be frequent and noisy. I don't think we want an event associated with a heartbeat tick.

core/common/src/main/java/alluxio/AlluxioEvent.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public enum AlluxioEvent {
+  // Alluxio events.
+  // Next id:1019.
+  MasterProcessCreated(1000),
+  JournalSystemStarted(1001),
+  JournalSystemStopped(1002),
+  JournalSystemGainedPrimacy(1003),
+  JournalSystemLostPrimacy(1004),
+  MasterRpcServerStarted(1005),
```
gpang, line 35: Are these related to `MasterProcessCreated` event? Should they be grouped together?

ggezer: grouped how?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncer.java
ggezer: removed events.

PR #11619, opened 2020-06-22
Wait for alluxio master to come up before startSync

LuQQiu:
StartSync may fail if master doesn't come up quick

### DIFF
--- a/integration/emr/alluxio-emr.sh
+++ b/integration/emr/alluxio-emr.sh
@@ -685,6 +685,10 @@ IN
       fi
 
       if [[ "${sync_list}" ]]; then
+        until ${ALLUXIO_HOME}/bin/alluxio fsadmin report
+        do
+          sleep 5
+        done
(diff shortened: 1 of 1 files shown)


integration/emr/alluxio-emr.sh
```diff
@@ -685,6 +685,10 @@ IN
       fi
 
       if [[ "${sync_list}" ]]; then
+        until ${ALLUXIO_HOME}/bin/alluxio fsadmin report
```
Xenorith, line 688: what do you think about always waiting after starting so any future additional alluxio commands will be safe to run? so i propose to move this immediately after register_backup_on_shutdown block. the bootstrapping shouldn't take any longer since workers still have to start and this is on master

LuQQiu: Sounds good, it may be clearer to do that

LuQQiu:

PR #11621, opened 2020-06-22
Implement listLocatedStatus hdfs api

gpang:
The default implementation of `listLocatedStatus` calls `listStatus`, then for each path, calls `getFileStatus`, and `getBlockLocations`. However, the Alluxio status includes the block locations, so this PR eliminates the 2 additional RPCs per path in the listing.


gpang:
Jenkins, test this please

PR #11630, opened 2020-06-23
Add a time out during raft unit tests

ggezer:


ZacBlanco:
Could we also use the `@Timeout` annotation on each individual test?

ggezer:

ZacBlanco:
FYI, I also created #11637 

so while this will prevent PR builds from failing, We should eventually revisit and run these tests without a timeout to find the actual bug.

PR #11641, opened 2020-06-24
Fix the socket timeout config processing in hms tool

LuQQiu:
SocketTimeoutConfig can be provided as String or integer object. 
Process both types and use default socket timeout if the object is not String or integer


LuQQiu:
@madanadit because the object map is error prone. Even if i decide to go for one type, we may use it wrong. But yeah, at least in the two places, it should be consistent.

PR #11650, opened 2020-06-26
Refactor stream reader/writer tracking

ggezer:
Extends `BlockReader` and `BlockWriter` interfaces from a new `BlockClient` interface for simplifying the block tracking interfaces. It also adds more traces to be able to track open stream reader/writer counts.


core/server/worker/src/main/java/alluxio/worker/block/management/DefaultStoreLoadTracker.java
```diff
@@ -96,6 +101,8 @@ public void writerClosed(BlockWriter writer, BlockStoreLocation location) {
    * Used to activate stream reader/writer for load tracking.
    */
   private void streamOpened(Object stream, BlockStoreLocation location) {
+    LOG.debug("Stream opened. Type: {}, Id: {}, Location: {}",
+        stream.getClass().getSimpleName(), System.identityHashCode(stream), location);
```
ZacBlanco, line 105: can we encapsulate this into some kind of object so that we can keep the open/close messages consistent?

core/server/worker/src/main/java/alluxio/worker/block/io/BlockStreamTracker.java
```diff
@@ -21,14 +21,14 @@
  */
 public class BlockStreamTracker {
   /** List of listeners for this tracker. */
-  private static Set<BlockStreamListener> sListeners = new ConcurrentHashSet<>();
+  private static Set<BlockClientListener> sListeners = new ConcurrentHashSet<>();
```
ZacBlanco, line 24: Do the `BlockClientListener` implementations implement `hashCode` and `equals`?

ggezer: It doesn't and there is only 1 registrar.
Changed it to a synchronized list anyways.

ZacBlanco: won't adding/removing be slow then? Under heavy load we'd have to iterate over at minimum, the current number of concurrent block requests. With current defaults that's 3k (2k reads, 1k writes)

gpang:
@ggezer Could you provide a description for the PR?

ggezer:
jenkins, test this please

core/server/worker/src/main/java/alluxio/worker/block/io/BlockStreamTracker.java
ggezer: > there is only 1 registrar.

ggezer: and current defaults are not sane.

ggezer: in other words, if this causes an i/o bottleneck then it's a good bottleneck.

ggezer: Changed it to cow list anyway.

ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
jenkins, test this please

ggezer:
@ZacBlanco ping!

core/server/worker/src/main/java/alluxio/worker/block/RemoteBlockReader.java
```diff
@@ -29,7 +30,7 @@
 /**
  * Reads a block from a remote worker node. This should only be used for reading entire blocks.
  */
-public class RemoteBlockReader implements BlockReader {
+public class RemoteBlockReader extends AbstractBlockClient implements BlockReader {
```
ZacBlanco, line 33: I think the code would be cleaner if `BlockReader` and `BlockWriter` were an abstract classes which extended `AbstractBlockClient` so that is wouldn't require both an `extends` and `implements` here. Then we could just have the default constructor for the Reader/writer set its type to `READER/WRITER` and you wouldn't need `super(READER)/super(WRITER)` everywhere since it would be taken care of by the class def extending `BlockReader` or `BlockWriter`

core/common/src/main/java/alluxio/worker/block/io/BlockClient.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package alluxio.worker.block.io;
+
+import java.io.Closeable;
+
+/**
+ * Base interface for block reader/writers.
+ */
+public interface BlockClient extends Closeable {
```
ZacBlanco, line 19: I would prefer using `AutoCloseable` unless you think there is a reason we shouldn't? It will make these all compatible with the try-with-resources pattern

ggezer: `Closeable` is also an `AutoCloseable`. So it should be fine to use it in try-with-resources right?

core/server/worker/src/main/java/alluxio/worker/block/RemoteBlockReader.java
ggezer: I didn't want to change the existing class names.
However, I think it's OK to make them abstract without stating in their class name...

core/common/src/main/java/alluxio/worker/block/io/BlockClient.java
```diff
@@ -11,10 +11,57 @@
...
  */
-public interface BlockClient extends Closeable {
+public abstract class BlockClient implements Closeable {
+  private static final Logger LOG = LoggerFactory.getLogger(BlockClient.class);
+
+  /** Used to keep unique client ids. */
+  private static AtomicInteger sNextClientId = new AtomicInteger(0);
+
+  /** Internal client id. */
+  private int mClientId;
```
ZacBlanco, line 32: this should be final?

core/common/src/main/java/alluxio/worker/block/io/BlockClient.java
```diff
@@ -11,10 +11,57 @@
...
+  /** the client type. */
+  private Type mClientType;
```
ZacBlanco, line 34: This should be final?

core/common/src/main/java/alluxio/worker/block/io/BlockClient.java
```diff
@@ -11,10 +11,57 @@
...
 /**
- * Base interface for block reader/writers.
+ * Provides basic tracking and representation for block reader/writer clients.
  */
-public interface BlockClient extends Closeable {
+public abstract class BlockClient implements Closeable {
+  private static final Logger LOG = LoggerFactory.getLogger(BlockClient.class);
+
+  /** Used to keep unique client ids. */
+  private static AtomicInteger sNextClientId = new AtomicInteger(0);
```
ZacBlanco, line 29: I think this should also be `final` since it's the reference to the `AtomicInteger` object

ggezer:

ggezer:

PR #11658, opened 2020-06-26
[DOCFIX] Add example Ozone configuration files

### LINKED ISSUE #11657: Add example Ozone configuration to documentation
**Page**

docs/en/ufs/Ozone.md

**Summary**

Add the Ozone-site.xml and core-site.xml

maobaolong:
Fixes #11657 

Add an example configuration file to help avoid making mistakes

### DIFF
--- a/docs/en/ufs/Ozone.md
+++ b/docs/en/ufs/Ozone.md
@@ -84,8 +84,8 @@ $ ./bin/alluxio fs mount \
   /ozone o3fs://<OZONE_BUCKET>.<OZONE_VOLUME>/
 ```
 
-Possible core-site.xml and ozone-site.xml
-- core-site.xml
+Possible `core-site.xml` and `ozone-site.xml`
+- `core-site.xml`
 
 ```xml
 <configuration>
@@ -97,25 +97,13 @@ Possible core-site.xml and ozone-site.xml
     <name>fs.AbstractFileSystem.o3fs.impl</name>
     <value>org.apache.hadoop.fs.ozone.BasicOzFs</value>
   </property>
-  <property>
-    <name>fs.defaultFS</name>
-    <value>o3fs://mybucket.myvol</value>
-  </property>
 </configuration>
 ```
 
-- ozone-site.xml
+- `ozone-site.xml`
 
 ```xml
 <configuration>
-  <property>
-    <name>ozone.replication</name>
-    <value>1</value>
-  </property>
-  <property>
-    <name>ozone.metadata.dirs</name>
-    <value>metadata.dir</value>
-  </property>
   <property>
     <name>ozone.scm.names</name>
     <value>localhost</value>
--- a/docs/en/ufs/Ozone.md
+++ b/docs/en/ufs/Ozone.md
@@ -84,6 +84,65 @@ $ ./bin/alluxio fs mount \
   /ozone o3fs://<OZONE_BUCKET>.<OZONE_VOLUME>/
 ```
 
+Possible core-site.xml and ozone-site.xml
+- core-site.xml
+
+```xml
+<configuration>
+  <property>
+    <name>fs.o3fs.impl</name>
+    <value>org.apache.hadoop.fs.ozone.BasicOzoneFileSystem</value>
+  </property>
+  <property>
+    <name>fs.AbstractFileSystem.o3fs.impl</name>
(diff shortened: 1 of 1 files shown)


maobaolong:
@ZacBlanco Please take a look, thanks. Without some necessary configure item, it cannot works.

docs/en/ufs/Ozone.md
```diff
@@ -84,6 +84,65 @@ $ ./bin/alluxio fs mount \
   /ozone o3fs://<OZONE_BUCKET>.<OZONE_VOLUME>/
 ```
 
+Possible core-site.xml and ozone-site.xml
+- core-site.xml
```
ZacBlanco, line 88: can you put all the references to `core-site.xml` and `ozone-site.xml` inside of backticks ("`")? IMO it makes it easier to read and stand out more.

docs/en/ufs/Ozone.md
```diff
@@ -84,6 +84,65 @@ $ ./bin/alluxio fs mount \
...
+    <value>org.apache.hadoop.fs.ozone.BasicOzoneFileSystem</value>
+  </property>
+  <property>
+    <name>fs.AbstractFileSystem.o3fs.impl</name>
+    <value>org.apache.hadoop.fs.ozone.BasicOzFs</value>
+  </property>
+  <property>
+    <name>fs.defaultFS</name>
+    <value>o3fs://mybucket.myvol</value>
+  </property>
```
ZacBlanco, line 103: I would prefer that for these configs we use the bare-minimum amount of properties. In this case, I think `fs.defaultFs` is unnecessary - the other two are ok though

docs/en/ufs/Ozone.md
```diff
@@ -84,6 +84,65 @@ $ ./bin/alluxio fs mount \
...
+    <value>60s</value>
+  </property>
+  <property>
+    <name>hdds.ratis.raft.client.async.outstanding-requests.max</name>
+    <value>32</value>
+  </property>
+  <property>
+    <name>hdds.ratis.raft.client.rpc.watch.request.timeout</name>
+    <value>180s</value>
+  </property>
```
ZacBlanco, line 142: Again, I'd prefer to have only the necessary properties in these examples - Are any of these actually necessary? I don't know ozone that well, but I assume that the threshold, timeouts, and max requests are unnecessary. Based on my limited knowledge of ozone and my intuition, the only one here that looks like it might be required is `ozone.scm.names`. In that case I'd only like to have just that property here in the example.

If there are any others you think are definitely required, we may keep them in too. I would just like these trimmed as much as possible.

docs/en/ufs/Ozone.md
maobaolong: Done

docs/en/ufs/Ozone.md
maobaolong: Done.

docs/en/ufs/Ozone.md
maobaolong: Sorry for forgot to explain why these configs are necessary. Because there is a bug in Ozone, some configs without specific key would be set to 0, and the 0 cannot work. This bug related the fellowing configs
```xml
<property>
    <name>scm.container.client.max.size</name>
    <value>256</value>
  </property>
  <property>
    <name>scm.container.client.idle.threshold</name>
    <value>10s</value>
  </property>
  <property>
    <name>hdds.ratis.raft.client.rpc.request.timeout</name>
    <value>60s</value>
  </property>
  <property>
    <name>hdds.ratis.raft.client.async.outstanding-requests.max</name>
    <value>32</value>
  </property>
  <property>
    <name>hdds.ratis.raft.client.rpc.watch.request.timeout</name>
    <value>180s</value>
  </property>
```

maobaolong:
@ZacBlanco Thank you for your review, i remove some unnecessary configs, left the necessary configs. PTAL.

docs/en/ufs/Ozone.md
ZacBlanco: Thank you - that makes sense. We may leave them in then

ZacBlanco:

ZacBlanco:

maobaolong:
@ZacBlanco Thank you for your review.

Issue #11672, opened 2020-06-26
Allow a cancelled RPC to interrupt metadata syncing


ZacBlanco:
**Is your feature request related to a problem? Please describe.**

If a user triggers a sync (i.e. via a listStatus) on a large directory and the sync takes an extended period of time, then the user might be inclined to ctrl-C to cancel the RPC and issue it again.

The problem with this now is that the metadata sync will continue to run on the master until completion. Only after completion will the handler eventually run into an error when returning the RPC response.

Imagine syncing a large directory that creates > 1M files. This will consume a large swath of CPU for some period of time. Depending on the hardware configuration, this can be a _long_ time. If the user cancels the RPC, then it should interrupt the metadata sync which should allow more operations

**Describe the solution you'd like**

The solution itself is straightforward. Each RPC is associated with an `OperationContext<T>` - the context has a method `isCancelled` which will return whether the RPC issued by the user has been cancelled for one or more reasons (user, network, etc..). It uses the underlying gRPC `StreamObserver#isCancelled` method.

Within a metadata sync we should be able to provide an `OperationContext` and periodically check the `isCancelled` on the operation to see if we should continue syncing.

**Describe alternatives you've considered**

None at the moment. The proposed solutions seems like a simple and straightforward implementation with few, if any, downsides. The only argument I could see against this is that someone wants a "fire-and-forget" implementation for syncing where they can issue an `ls`, cancel the command, and continue using their terminal while the sync runs. In that case, the user would just never know when the sync is finished - but they would at least know that it was started.

**Urgency**

Prevent over-utilization on the master that doesn't benefit any client operation.

**Additional context**
N/A

ZacBlanco:
One other issue I'd like to bring up is that in the event of a cancelled RPC, then we can leave a half-synced directory. In the implementation we probably want to think about what we should do in such an event.

If the sync interval is set to, let's say, 1h. Then someone comes along that does an ls which triggers the sync and cancels it, then even if only half or 10% of the directory was synced, another sync won't be attempted for another hour. 


We should probably remove at least the root path from the UFS sync path cache in this situation so an RPC following a cancel can still trigger a sync without having to wait for the metadata interval. @ggezer What do you think?

ggezer:
If we do reset the sync status for the cancelled sync session, then the root need not be an exception and it's more consistent to reset for all base paths. 

[...]

PR #11674, opened 2020-06-26
Refactor FSM RPC cancellation support

ggezer:
This PR enables FSM internal call-trackers by default for all RPCs.
It also improves the overall RPC cancellation coverage by including sync sessions.

Fixes https://github.com/Alluxio/alluxio/issues/11672


core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
```diff
@@ -148,8 +148,8 @@ public void createDirectory(CreateDirectoryPRequest request,
     CreateDirectoryPOptions options = request.getOptions();
     RpcUtils.call(LOG, () -> {
       AlluxioURI pathUri = getAlluxioURI(request.getPath());
-      mFileSystemMaster.createDirectory(pathUri,
-          CreateDirectoryContext.create(options.toBuilder()));
+      mFileSystemMaster.createDirectory(pathUri, (CreateDirectoryContext) CreateDirectoryContext
+          .create(options.toBuilder()).withTracker(new GrpcCallTracker(responseObserver)));
```
ZacBlanco, line 152: if `withTracker` returns the generic type of the `OperationContext` (`C` in your implementation), then is it necessary to include the cast to `CreateDirectoryContext`? It just doesn't seem clean.


edit - I looked into it and figured out the issue. Poor use of generics. Change the class definition of `CreatePathContext` to remove the need to cast

old:

```java
public abstract class CreatePathContext<T extends GeneratedMessageV3.Builder<?>, K>
    extends OperationContext<T, CreatePathContext> {
```

new:
```java
public abstract class CreatePathContext<T extends GeneratedMessageV3.Builder<?>,
    K extends OperationContext> extends OperationContext<T, K> {
```

core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
```diff
@@ -159,8 +159,10 @@ public void createFile(CreateFilePRequest request,
       StreamObserver<CreateFilePResponse> responseObserver) {
     RpcUtils.call(LOG, () -> {
       AlluxioURI pathUri = getAlluxioURI(request.getPath());
-      return CreateFilePResponse.newBuilder().setFileInfo(GrpcUtils.toProto(mFileSystemMaster
-          .createFile(pathUri, CreateFileContext.create(request.getOptions().toBuilder()))))
+      return CreateFilePResponse.newBuilder()
+          .setFileInfo(GrpcUtils.toProto(mFileSystemMaster.createFile(pathUri,
+              (CreateFileContext) CreateFileContext.create(request.getOptions().toBuilder())
```
ZacBlanco, line 164: See other comment.

Change to the `CreatePathContext` class def removed the need for the extra cast

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -11,54 +11,89 @@
...
+    }
+
+    List<CallTracker> cancelledTrackers = new LinkedList<>();
+    for (CallTracker tracker : mCallTrackers) {
+      if (tracker.isCancelled()) {
+        cancelledTrackers.add(tracker);
+      }
+    }
+
+    return cancelledTrackers;
```
ZacBlanco, line 97: The list of call trackers should always be small, correct? 1 most of the time?

Are there any cases when we could have more than that?

core/server/master/src/test/java/alluxio/master/file/RpcContextTest.java
```diff
@@ -34,11 +35,12 @@
...
   private BlockDeletionContext mMockBDC = mock(BlockDeletionContext.class);
   private JournalContext mMockJC = mock(JournalContext.class);
+  private OperationContext mMockOC = mock(OperationContext.class);
   private RpcContext mRpcContext;
 
   @Before
   public void before() {
-    mRpcContext = new RpcContext(mMockBDC, mMockJC);
+    mRpcContext = new RpcContext(mMockBDC, mMockJC, mMockOC);
   }
```
ZacBlanco, line 44: Would like to see some test coverage on throwing for cancelled trackers.

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
ggezer: Every context has a state-lock tracker. Some has an additional transport level tracker (GrpcCallTracker).
Even for those, it's very unlikely to have more than 1 canceller. However, it's possible so I give out the list here. (We might also wanna add more trackers in the future...)

core/server/master/src/test/java/alluxio/master/file/RpcContextTest.java
ggezer: I can't see a clear way to test cancellation while the rpc is in progress, aka the loop break logic. I'll try adding some validation for the rest.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -632,6 +638,7 @@ private void loadMetadataForPath(LockedInodePath inodePath)
   private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context)
       throws AccessControlException, BlockInfoException, FileAlreadyCompletedException,
       FileDoesNotExistException, InvalidFileSizeException, InvalidPathException, IOException {
+    mRpcContext.throwIfCancelled();
```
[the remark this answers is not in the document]
ggezer: Does this have no potential to take a long time?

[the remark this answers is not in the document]
ZacBlanco: `loadMetadata` only ever executes for a single inode. I think in relative terms it is not a long amount of time.

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -11,54 +11,89 @@
...
    *
-   * @param optionsBuilder Internal proto message builder instance
-   * @param callTracker client call tracker, or {@code null} if no tracking is desired
+   * @param context the source context
+   * @return the updated instance
    */
-  public OperationContext(T optionsBuilder, CallTracker callTracker) {
-    mOptionsBuilder = optionsBuilder;
-    mCallTracker = callTracker;
+  public C withTracker(OperationContext context) {
```
gpang, line 58: I think it is strange to have an API called `withTracker` but the input parameter is not a tracker but a context? Why can't we just use the other implementation only?

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
```diff
@@ -11,54 +11,89 @@
...
 import javax.annotation.concurrent.NotThreadSafe;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 
 /**
  * Used as a base class for wrapping context around proto messages.
  *
  * @param <T> Proto message type
+ * @param <C> extended type
```
gpang, line 25: Is the only reason for this, to have a `withTracker()` call? It seems inconvenient to add the same class as a template class parameter. Instead, can't we just have a `setTracker()`, that updates the instance, and we just pass in the instance to the rpc call?

core/server/master/src/test/java/alluxio/master/file/RpcContextTest.java
ZacBlanco: I think what I would do is use Mockito's `Spy` class and wrap in conjunction with some locks to allow for the control of entering/exiting methods. See the `UfsStatusCacheTest` for an example. I'll paste some code from one of the tests inline here so you can see what I mean.


```java
createUfsDirs("dir0/dir1");
    ExecutorService executor =
        new ThreadPoolExecutor(1, 1, 1, TimeUnit.MINUTES, new SynchronousQueue<>());
    mCache = new UfsStatusCache(executor);
    mCache = Mockito.spy(mCache);
    Lock l = new ReentrantLock();
    l.lock();
    doAnswer((invocation) -> {
      try {
        l.lock();
        return invocation.callRealMethod();
      } finally {
        l.unlock();
      }
    }).when(mCache).getChildrenIfAbsent(any(AlluxioURI.class), any(MountTable.class));
    assertNotNull(mCache.prefetchChildren(new AlluxioURI("/dir0"), mMountTable));
    assertNull(mCache.prefetchChildren(new AlluxioURI("/dir0"), mMountTable)); // rejected
    assertNull(mCache.prefetchChildren(new AlluxioURI("/dir0"), mMountTable)); // rejected
    assertNull(mCache.prefetchChildren(new AlluxioURI("/dir0"), mMountTable)); // rejected
    assertNull(mCache.prefetchChildren(new AlluxioURI("/dir0"), mMountTable)); // rejected
    l.unlock();
    Collection<UfsStatus> statuses =
        mCache.fetchChildrenIfAbsent(new AlluxioURI("/dir0"), mMountTable, false);
    assertEquals(1, statuses.size());
    statuses.forEach(s -> assertEquals("dir1", s.getName()));
```

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
ggezer: I'm just trying to make service-handler clean by making the API pretty to use.

core/server/master/src/main/java/alluxio/master/file/contexts/OperationContext.java
gpang: I still have this question.

ggezer: turned out it was a dead method. removed.

core/server/master/src/test/java/alluxio/master/file/RpcContextTest.java
ggezer: I went for a simpler UT and think going beyond that will require an integration test with mocking which is not feasible.

core/server/master/src/test/java/alluxio/master/file/RpcContextTest.java
```diff
@@ -105,6 +108,46 @@ public void journalContextThrows() throws Throwable {
...
+    });
+    // Create RPC context.
+    RpcContext rpcCtx = new RpcContext(mMockBDC, mMockJC, opCtx);
+    // Verify the RPC is cancelled due to tracker that's always cancelled.
+    assertTrue(rpcCtx.isCancelled());
+    try {
+      // Verify cancellation throws.
+      rpcCtx.throwIfCancelled();
+      fail("Call should have been cancelled.");
+    } catch (RuntimeException e) {
```
ZacBlanco, line 146: This is a good case for the `ExpectedException` rule since there aren't any assertions afterwards ;)

ggezer:
alluxio-bot, test this please

ggezer:

ggezer:

PR #11679, opened 2020-06-28
Add proxy support for glue udb

HelloHorizon:


table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
```diff
@@ -173,6 +173,42 @@ public static void unregister(Property Property) {
           .setDescription("The secret key to access the aws glue.")
           .build();
 
+  public static final Property AWS_ENABLE_PROXY =
+      new Builder(Name.AWS_ENABLE_PROXY)
+          .setDefaultValue("")
+          .setDescription("Enable proxy setting for glue client.")
+          .build();
+
+  public static final Property AWS_PROXY_PROTOCOL =
```
yuzhu, line 182: indentation>?

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
```diff
@@ -173,6 +173,42 @@ public static void unregister(Property Property) {
           .setDescription("The secret key to access the aws glue.")
           .build();
 
+  public static final Property AWS_ENABLE_PROXY =
+      new Builder(Name.AWS_ENABLE_PROXY)
```
yuzhu, line 177: is there a boolean Property type?

yuzhu: shouldn't the default be false then?

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
HelloHorizon: fixed.

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
HelloHorizon: changed.

HelloHorizon:
@yuzhu I tested it with an AWS tiny proxy server. Since it has dependencies on proxy server and glue, we may want to add the tests in Golluxio other than Alluxio repo.

HelloHorizon:
jenkins, test this please

HelloHorizon:
jenkins, test this please

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/Property.java
```diff
@@ -208,5 +244,11 @@ public String getDefaultValue() {
     public static final String CATALOG_ID = "aws.catalog.id";
     public static final String AWS_GLUE_ACCESS_KEY = "aws.accesskey";
     public static final String AWS_GLUE_SECRET_KEY = "aws.secretkey";
+    public static final String AWS_ENABLE_PROXY = "aws.enable.proxy";
```
gpang, line 247: Do we need this? Instead, can we assume that if `aws.proxy.host` is set, then proxy is enabled? We would be able to do this if the proxy is enabled IFF `aws.proxy.host` is set.

Otherwise, if we need to keep this, can this be renamed to `aws.proxy.enable`? It would feel more consistent with the other proxy related properties.

Lastly, is this an aws type of configuration, or a glue specific configuration? If it is glue specific, I feel like we don't even need the `aws.` prefix, since this is for the glue UDB?

HelloHorizon: You are right, I changed to use `aws.proxy.host` as trigger. These configurations are used to building the AWS client configurations, it may better to maintain the `aws.` prefix?

HelloHorizon:
jenkins, test this please

yuzhu:
jenkins, test this please

yuzhu:

PR #11704, opened 2020-07-01
Optimize UFS access on missing paths

ZacBlanco:
The main contribution of this PR is to remove a 2nd UFS call against the UFS when performing a listStatus/getStatus against a path which doesn't exist in Alluxio or the UFS.  Previously, `fetchStatusIfAbsent` would return null, and then call `loadMetadata` with a null status which then would make an `exists` call against the UFS. Now we will instead throw a FileNotFoundException which is the same behavior as specified in the UFS API.

### DIFF
--- a/core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
+++ b/core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
@@ -137,11 +138,11 @@ public class UfsStatusCache {
    *
    * @param path the path the retrieve
    * @param mountTable the Alluxio mount table
-   * @return The corresponding {@link UfsStatus} or {@code null} if there is none stored
+   * @return The corresponding {@link UfsStatus}
+   * @throws java.io.FileNotFoundException if the status can't be retrieved from the UFS
    */
-  @Nullable
   public UfsStatus fetchStatusIfAbsent(AlluxioURI path, MountTable mountTable)
-      throws InvalidPathException {
+      throws InvalidPathException, FileNotFoundException {
     UfsStatus status = mStatuses.get(path);
     if (status != null) {
       return status;
@@ -152,7 +153,7 @@ public class UfsStatusCache {
       UnderFileSystem ufs = ufsResource.get();
       UfsStatus ufsStatus = ufs.getStatus(ufsUri.toString());
       if (ufsStatus == null) {
-        return null;
+        throw new FileNotFoundException("fetched status is null for: " + ufsUri);
       }
       ufsStatus.setName(path.getName());
       addStatus(path, ufsStatus);
@@ -160,7 +161,7 @@ public class UfsStatusCache {
     } catch (IllegalArgumentException | IOException e) {
       LogUtils.warnWithException(LOG, "Failed to fetch status for {}", path, e);
     }
-    return null;
+    throw new FileNotFoundException("Failed to fetch status of " + ufsUri);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -152,15 +153,15 @@ public UfsStatus fetchStatusIfAbsent(AlluxioURI path, MountTable mountTable)
...
+        throw new FileNotFoundException("fetched status is null for: " + ufsUri);
       }
       ufsStatus.setName(path.getName());
       addStatus(path, ufsStatus);
       return ufsStatus;
     } catch (IllegalArgumentException | IOException e) {
       LogUtils.warnWithException(LOG, "Failed to fetch status for {}", path, e);
     }
-    return null;
+    throw new FileNotFoundException("Failed to fetch status of " + ufsUri);
```
ggezer, line 164: is this retried on caller site?

core/server/master/src/test/java/alluxio/underfs/UfsStatusCacheTest.java
```diff
@@ -371,18 +372,26 @@ public void testFetchExistingSingleStatus() throws Exception {
   @Test
   public void testFetchSingleStatusNonExistingPath() throws Exception {
     spyUfs();
-    UfsStatus fetched = mCache.fetchStatusIfAbsent(new AlluxioURI("/testFile"), mMountTable);
+    try {
+      UfsStatus fetched = mCache.fetchStatusIfAbsent(new AlluxioURI("/testFile"), mMountTable);
+      fail("Should have thrown FileNotFoundException");
```
ggezer, line 377: Can you replace these with`ExpectedException` rule?

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
ZacBlanco: no

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -152,15 +153,15 @@ public UfsStatus fetchStatusIfAbsent(AlluxioURI path, MountTable mountTable)
       UnderFileSystem ufs = ufsResource.get();
       UfsStatus ufsStatus = ufs.getStatus(ufsUri.toString());
       if (ufsStatus == null) {
-        return null;
+        throw new FileNotFoundException("fetched status is null for: " + ufsUri);
```
calvinjia, line 156: Will this exception be thrown to the client? If so we should make it more explanatory (ie. what does `null` status mean?)

ZacBlanco: No, this is not returned to the client

core/server/master/src/test/java/alluxio/underfs/UfsStatusCacheTest.java
ZacBlanco: Using the ExpectedException rule I can't use `Mockito.verify` afterwards

gpang:
@ZacBlanco This still has the commits of https://github.com/Alluxio/alluxio/pull/11720 ?

ZacBlanco:
Sorry, forgot to push

ZacBlanco:
alluxio-bot, test this please

ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
calvinjia: Just want to double check, I don't see `FileNotFoundException` being caught by the callers? ie. `processSyncPath`

ZacBlanco: `FileNotFoundException` extends from `IOException` which is in the method definition of the caller `loadMetadataForPath` as a part of it's `throws` declarations.

`IOException` is then caught by the try-catch block on line 422 in `processSyncPath` or if the root path is missing, then it is caught in the handler on line 300

calvinjia:

calvinjia:

PR #11709, opened 2020-07-03
Ignore worker response when closing gRPC data reader

ggezer:
It has been observed many times that a worker might get busy/hung after a read is served and closing gRPC stream times out.
After the content is read and the underlying stream is closed, there is no need to wait for worker to acknowledge.


ggezer:
Seems like it's tricky to ensure buffer release.

PR #11715, opened 2020-07-05
Fix potential issue caused by unhandled exception of jnifuse

iluoeli:


gpang:
Jenkins, test this please

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -132,27 +132,49 @@ public void umount() {
   }
 
   public int openCallback(String path, ByteBuffer buf) {
-    FuseFileInfo fi = new FuseFileInfo(buf);
-    return open(path, fi);
+    try {
+      return open(path, FuseFileInfo.wrap(buf));
+    } catch (Exception e) {
+      e.printStackTrace();
```
apc999, line 138: `e.printStackTrace();` prints to stdout, not appropriate for production code.
Use `LOG.error("Failed to open {}", path, e)` here so the stack trace will still be printed.
with claiming  `private static final Logger LOG = LoggerFactory.getLogger(AbstractFuseFileSystem.class);` in the beginning.

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -132,27 +132,49 @@ public void umount() {
...
+    }
   }
 
   public int readCallback(String path, ByteBuffer buf, long size, long offset, ByteBuffer fibuf) {
-    FuseFileInfo fi = new FuseFileInfo(fibuf);
-    return read(path, buf, size, offset, fi);
+    try {
+      return read(path, buf, size, offset, FuseFileInfo.wrap(fibuf));
+    } catch (Exception e) {
+     e.printStackTrace();
```
apc999, line 147: `LOG.error("Failed to read {}, size {}, offset {}", path, size, offset, e)`

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -132,27 +132,49 @@ public void umount() {
...
+    }
   }
 
   public int getattrCallback(String path, ByteBuffer buf) {
-    FileStat stat = new FileStat(buf);
-    return getattr(path, stat);
+    try {
+      return getattr(path, FileStat.wrap(buf));
+    } catch (Exception e) {
+      e.printStackTrace();
```
apc999, line 156: `LOG.error("Failed to getattr {}", path, e)`

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -132,27 +132,49 @@ public void umount() {
...
+      return -ErrorCodes.EIO();
+    }
   }
 
   public int readdirCallback(String path, long bufaddr, FuseFillDir filter, long offset,
       ByteBuffer fi) {
-    return readdir(path, bufaddr, filter, offset, new FuseFileInfo(fi));
+    try {
+      return readdir(path, bufaddr, filter, offset, new FuseFileInfo(fi));
```
apc999, line 164: `LOG.error("Failed to readdir {}, offset {}", path, offset, e)`

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -132,27 +132,49 @@ public void umount() {
...
+      return -ErrorCodes.EIO();
+    }
   }
 
   public int unlinkCallback(String path) {
-    return unlink(path);
+    try {
+      return unlink(path);
+    } catch (Exception e) {
+      e.printStackTrace();
```
apc999, line 175: `LOG.error("Failed to unlink {}: ", path, e)`

apc999: please also add similar try-catch to other callbacks, including flushCallback, releaseCallback and etc even we didn't implement them yet.

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -132,70 +137,147 @@ public void umount() {
   }
 
   public int openCallback(String path, ByteBuffer buf) {
-    FuseFileInfo fi = new FuseFileInfo(buf);
-    return open(path, fi);
+    try {
+      return open(path, FuseFileInfo.wrap(buf));
+    } catch (Exception e) {
+      LOG.error("Failed to open {}: ", path, e);
```
gpang, line 143: Something to be careful of is that if you want to log the exception with `LOG.error(..., e)`, it doesn't work with the `{}` substitution ability. So, that means all of the log lines with exception as the last parameter should look something like:
```
LOG.error(String.format("... %s", path), e)
```

This way, the log message has all the parameters, and the exception and stack is printed.

gpang:
Jenkins, test this please

integration/fuse/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
apc999: @gpang this is actually unnecessary to use `String.format`. https://stackoverflow.com/a/51781806/3161750
As long as you pass a Throwable as the last arg, the full log of exception will be printed as an undocumented feature.
I've been leveraging this feature for a long time

apc999:
> Jenkins, test this please

this is a dev branch, test will not pass @gpang

apc999:

PR #11720, opened 2020-07-06
Add minor RPC optimizations

ZacBlanco:
Based on my testing on a local machine these sites accounted for anywhere between 1% and 3% of the CPU time when processing the RPC.

- When calling getObjectListing on the ObjectUnderFileSystem, we should store the value of the breadcrumbs enabled as a boolean rather than always parsing from the config
- When calling `lockShared()` in the state lock manager, prevent a String.format call by wrapping with a conditional statement.
- When creating the DefaultBlockDeletionContext, initialize the array list with a size of 0 to prevent unnecessarily allocating small arrays over and over again with the default ArrayList constructor, when most times we won't even add to this ArrayList.


ZacBlanco:

PR #11724, opened 2020-07-07
[DOCFIX] Improve logLevel command explanation

jiacheliu3:


docs/en/operation/Basic-Logging.md
```diff
@@ -110,6 +110,24 @@ And the following command returns the log level of the class `alluxio.heartbeat.
 $ ./bin/alluxio logLevel --logName=alluxio.heartbeat.HeartbeatContext --target=workers
 ```
 
+You can also update the log level at a package level.
+For example, you can update the log level of all classes in `alluxio.heartbeat` package with the following command:
+```console
+$ ./bin/alluxio logLevel --logName=alluxio.heartbeat --target=workers --level=DEBUG
+```
+This works because log4j loggers will inherit the log level from their ancestors.
+In this case `alluxio.heartbeat.HeartbeatContext` inherits the log level if it is set on `alluxio.heartbeat`.
```
ZacBlanco, line 119: I feel like we can use a better example than `alluxio.heartbeat`. Maybe `alluxio.underfs`? Its something more users might understand/know about.

Just a suggestion

jiacheliu3: Good idea! Done.

ZacBlanco:

Issue #8687, opened 2019-04-04
Single-Tier Storage configuration is not working 


perrywang:
**Alluxio Version:**
1.8.1
**Describe the bug**
following https://www.alluxio.org/docs/1.8/en/advanced/Alluxio-Storage-Management.html#single-tier-storage
config below in alluxio-site.properties
alluxio.worker.memory.size=16GB
alluxio.worker.tieredstore.level0.dirs.path=/mnt/ramdisk,/home/hdd
alluxio.worker.tieredstore.level0.dirs.quota=16GB,200GB
but the worker create below ramfs mount point
**/mnt/ramdisk,/home/hdd**
the comma is as part of directory name of **ramdisk,**

**To Reproduce**
config as above and start worker

**Expected behavior**
working consistent with doc
https://www.alluxio.org/docs/1.8/en/advanced/Alluxio-Storage-Management.html#single-tier-storage

**Urgency**
intermediate
**Additional context**
N/A

aaudiber:
@perrywang thanks for reporting this. I tried it out locally and I can reproduce the problem.

There is a check in the start script which reads `level0.dirs.path` and checks that the path exists or the worker will try to mount it. To work around the issue, you can mount the ramdisk beforehand, then start the worker with `bin/alluxio-start.sh worker NoMount`. For this to work, comment out this check in `alluxio-start.sh`:
https://github.com/Alluxio/alluxio/blob/branch-1.8/bin/alluxio-start.sh#L109

perrywang:
> @perrywang thanks for reporting this. I tried it out locally and I can reproduce the problem.
> 
> There is a check in the start script which reads `level0.dirs.path` and checks that the path exists or the worker will try to mount it. To work around the issue, you can mount the ramdisk beforehand, then start the worker with `bin/alluxio-start.sh worker NoMount`. For this to work, comment out this check in `alluxio-start.sh`:
> https://github.com/Alluxio/alluxio/blob/branch-1.8/bin/alluxio-start.sh#L109

I can manually mount ramfs but workaround is not working because NoMount will ignore /home/hdd

aaudiber:
@perrywang I tried it out locally with this configuration:

```
alluxio.master.hostname=localhost
alluxio.worker.tieredstore.level0.dirs.path=/Volumes/ramdisk,/Users/andrew/hdd
alluxio.worker.tieredstore.level0.dirs.quota=2GB,10GB
```

[...]

PR #11743, opened 2020-07-10
Add tests for GreedyAllocator

apc999:
Address https://github.com/Alluxio/alluxio/issues/8687 when loading data from UFS using `GreedyAllocator`
Before this change, when the first storage dir has only space smaller than the block size but larger than 8mb (default request buffer size), GreedyAllocator will always pick this storage dir and fail, even there are next storage dir with sufficient space.
The reason to have incremental growth when writing a block is only an optimization for writing a new block when we don't know the block size before hand. Whereas here for blocks in UFS, we always know the block size so it is safe to just request space as large as the block size.

this is reopen of https://github.com/Alluxio/alluxio/pull/10463/


tests/src/test/java/alluxio/server/tieredstore/TieredStoreIntegrationTest.java
```diff
@@ -236,4 +242,38 @@ public void promoteBlock() throws Exception {
...
+  /**
+   * With setting of "StorageDir1, StorageDir2", loading a block from UFS that is bigger than
+   * the available space of StorageDir1 should land in StorageDir2 using GreedyAllocator.
+   * https://github.com/Alluxio/alluxio/issues/8687
+   */
+  @Test
+  public void greedyAllocator() throws Exception {
+    mLocalAlluxioClusterResource.stop();
+    int fileLen = 8 * Constants.MB;
+    mLocalAlluxioClusterResource
```
gpang, line 255: Can't we use the `Config` annotation to update configuration for a particular test? Like: https://github.com/Alluxio/alluxio/blob/master/tests/src/test/java/alluxio/client/cli/fs/command/PersistCommandTest.java#L362

Then we don't have to stop and start the cluster, which is not cheap.

core/server/worker/src/main/java/alluxio/worker/block/UnderFileSystemBlockReader.java
```diff
@@ -119,7 +117,7 @@ public static UnderFileSystemBlockReader create(UnderFileSystemBlockMeta blockMe
   private UnderFileSystemBlockReader(UnderFileSystemBlockMeta blockMeta, boolean positionShort,
       BlockStore localBlockStore, UfsManager ufsManager, UfsInputStreamManager ufsInstreamManager)
       throws IOException {
-    mInitialBlockSize = ServerConfiguration.getBytes(PropertyKey.WORKER_FILE_BUFFER_SIZE);
+    mInitialBlockSize = blockMeta.getBlockSize();
```
gpang, line 120: Are there any implications to this change?

tests/src/test/java/alluxio/server/tieredstore/TieredStoreIntegrationTest.java
apc999: that's where I started with `config` annotation. Then I realized it requires compile-time **String constant**, in here a lot values I'm setting are derived (e.g.,  tmp dir and even the tiered property keys are derived from template)

core/server/worker/src/main/java/alluxio/worker/block/UnderFileSystemBlockReader.java
apc999: I don't think there is any downside here. @calvinjia  ?

apc999:
@gpang I updated the PR description. PTAL

core/server/worker/src/main/java/alluxio/worker/block/UnderFileSystemBlockReader.java
ggezer: Doesn't`read()` method should never call `requestSpace()` after this change here?

Otherwise, each read call would allocate more space on top of the initial full block size. However, requested spaces wouldn't leak because the block-store would be updated with the physical file size as part of the commit.

apc999: @ggezer  you are correct. `requestspace()` should not be called any more to load this block from UFS.

ggezer: This condition in `read` looks broken, hence always calling `requestSpace`.

>>> if (mBlockWriter != null && mBlockWriter.getPosition() < mInStreamPos) {
  mLocalBlockStore.requestSpace(mBlockMeta.getSessionId(), mBlockMeta.getBlockId(),
            mInStreamPos - mBlockWriter.getPosition());

apc999:
Resume this PR after https://github.com/Alluxio/alluxio/pull/12182 is merged
Now this PR is only adding tests to ensure GreedyAllocator can work

yuzhu (approved):
LGTM

apc999:

PR #11746, opened 2020-07-10
Simplify UfsStatus handling in loadMetadata

ZacBlanco:
This reverts a change from #11704 while also removing redundant UFS calls. The `ufs.exists` in `loadMetadata` is no longer necessary because every call to `loadMetadata` goes through call site place which will always set the `UfsStatus` in the `LoadMetadataContext`. The attempt to retrieve the `UfsStatus` from the `UfsStatusCache` will always pull the status from the internal map, or reach out to the UFS. If the status couldn't be retrieved on the call to `fetchStatusIfAbsent` then it does not exist in the UFS.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -625,6 +625,9 @@ public class InodeSyncStream {
       FileAlreadyCompletedException, InvalidFileSizeException, BlockInfoException {
 
     UfsStatus status = mStatusCache.fetchStatusIfAbsent(inodePath.getUri(), mMountTable);
+    if (status == null) {
+      return;
+    }
     LoadMetadataContext ctx = LoadMetadataContext.mergeFrom(
         LoadMetadataPOptions.newBuilder()
             .setCommonOptions(NO_TTL_OPTION)
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -635,8 +635,8 @@ public class InodeSyncStream {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ZacBlanco:
jenkins, test this please

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -639,10 +639,8 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
       FileDoesNotExistException, InvalidFileSizeException, InvalidPathException, IOException {
     AlluxioURI path = inodePath.getUri();
     MountTable.Resolution resolution = mMountTable.resolve(path);
-    AlluxioURI ufsUri = resolution.getUri();
-    try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
-      UnderFileSystem ufs = ufsResource.get();
-      if (context.getUfsStatus() == null && !ufs.exists(ufsUri.toString())) {
+    try {
+      if (context.getUfsStatus() == null) {
```
gpang, line 643: This means for load metadata, this `context.getUfsStatus()` must always be prepopulated with the UFS status? Something earlier in the call stack will be populating everything? What is the difference between this `context.getUfsStatus()` and the status cache?

This also implies this method itself will never call UFS?

Can we add this to the method javadoc?

ZacBlanco: `context.getUfsStatus` is can be thought of as similar to an RPC options object. The `context` object is basically a POJO that just holds information. No fancy processing. Just getters and setters.

When calling this method it is now expected that the ufs status is already populated. And yes, this method shouldn't call the UFS except in the case where the child statuses may need to be retrieved (see the `fetchChildren` call lower in the code)

gpang:

gpang:

PR #11750, opened 2020-07-10
Fix Object UFS edge case again

bradyoo:
Adding / to the keys was not a good idea.

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
+++ b/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
@@ -1118,14 +1118,25 @@ public abstract class ObjectUnderFileSystem extends BaseUnderFileSystem {
    */
   @VisibleForTesting
   public String stripPrefixIfPresent(String path) {
-    final String normalizedPath = PathUtils.normalizePath(path, PATH_SEPARATOR);
     String stripedKey = CommonUtils.stripPrefixIfPresent(
-        normalizedPath,
+        path,
         PathUtils.normalizePath(mRootKeySupplier.get(), PATH_SEPARATOR));
-    if (!stripedKey.equals(normalizedPath)) {
+    if (!stripedKey.equals(path)) {
       return stripedKey;
     }
-    return CommonUtils.stripPrefixIfPresent(normalizedPath, PATH_SEPARATOR);
+
+    // See if adding a PATH_SEPARATOR makes a difference in stripping prefix
+    final String normalizedPath = PathUtils.normalizePath(path, PATH_SEPARATOR);
+    if (!path.equals(normalizedPath)) {
+      stripedKey = CommonUtils.stripPrefixIfPresent(
+          normalizedPath,
+          PathUtils.normalizePath(mRootKeySupplier.get(), PATH_SEPARATOR));
+      if (!stripedKey.equals(normalizedPath)) {
+        return stripedKey;
+      }
+    }
+
+    return CommonUtils.stripPrefixIfPresent(path, PATH_SEPARATOR);
   }
 
   /**
--- a/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
+++ b/core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
@@ -1118,22 +1118,13 @@ public abstract class ObjectUnderFileSystem extends BaseUnderFileSystem {
    */
   @VisibleForTesting
   public String stripPrefixIfPresent(String path) {
+    if (mRootKeySupplier.get().equals(path)) {
+      return "";
+    }
     String stripedKey = CommonUtils.stripPrefixIfPresent(
-        path,
-        PathUtils.normalizePath(mRootKeySupplier.get(), PATH_SEPARATOR));
+        path, PathUtils.normalizePath(mRootKeySupplier.get(), PATH_SEPARATOR));
     if (!stripedKey.equals(path)) {
       return stripedKey;
-    }
-
-    // See if adding a PATH_SEPARATOR makes a difference in stripping prefix
-    final String normalizedPath = PathUtils.normalizePath(path, PATH_SEPARATOR);
-    if (!path.equals(normalizedPath)) {
-      stripedKey = CommonUtils.stripPrefixIfPresent(
-          normalizedPath,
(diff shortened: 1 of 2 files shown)


bradyoo:
@gpang

underfs/s3a/src/test/java/alluxio/underfs/s3a/S3AUnderFileSystemTest.java
```diff
@@ -244,8 +244,8 @@ public void getOperationMode() throws Exception {
   public void stripPrefixIfPresent() throws Exception {
     Assert.assertEquals("", mS3UnderFileSystem.stripPrefixIfPresent("s3a://" + BUCKET_NAME));
     Assert.assertEquals("", mS3UnderFileSystem.stripPrefixIfPresent("s3a://" + BUCKET_NAME + "/"));
-    Assert.assertEquals("test/", mS3UnderFileSystem.stripPrefixIfPresent("test"));
-    Assert.assertEquals("test/",
-        mS3UnderFileSystem.stripPrefixIfPresent("s3a://" + BUCKET_NAME + "/test/"));
```
gpang, line 249: can we keep this test case? `"s3a://" + BUCKET_NAME + "/test/"`

bradyoo: No because I don't think /test/ is actually a valid input to this function. It was becuase I thought it was a valid input that I made this mistake.

bradyoo:

underfs/s3a/src/test/java/alluxio/underfs/s3a/S3AUnderFileSystemTest.java
gpang: We can't pass in directory paths to this method? I thought any s3 UFS path can be passed to this method?

gpang: I think directories are also allowed, since a few lines above, we have `"s3a://" + BUCKET_NAME + "/"` which is a directory.

bradyoo:
ah right. Sorry reflected feedback. Fixed. Also, confirmed that this fixed some of the autobots issues.

underfs/s3a/src/test/java/alluxio/underfs/s3a/S3AUnderFileSystemTest.java
```diff
@@ -244,7 +244,7 @@ public void getOperationMode() throws Exception {
   public void stripPrefixIfPresent() throws Exception {
     Assert.assertEquals("", mS3UnderFileSystem.stripPrefixIfPresent("s3a://" + BUCKET_NAME));
     Assert.assertEquals("", mS3UnderFileSystem.stripPrefixIfPresent("s3a://" + BUCKET_NAME + "/"));
-    Assert.assertEquals("test/", mS3UnderFileSystem.stripPrefixIfPresent("test"));
+    Assert.assertEquals("test", mS3UnderFileSystem.stripPrefixIfPresent("test"));
```
gpang, line 247: Can we do the full matrix of test cases?

```
test
test/
/test
/test/
```

are these also possible?
```
""
"/"
```

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
```diff
@@ -1118,14 +1118,25 @@ protected boolean parentExists(String path) throws IOException {
...
       return stripedKey;
     }
-    return CommonUtils.stripPrefixIfPresent(normalizedPath, PATH_SEPARATOR);
+
+    // See if adding a PATH_SEPARATOR makes a difference in stripping prefix
+    final String normalizedPath = PathUtils.normalizePath(path, PATH_SEPARATOR);
+    if (!path.equals(normalizedPath)) {
+      stripedKey = CommonUtils.stripPrefixIfPresent(
+          normalizedPath,
+          PathUtils.normalizePath(mRootKeySupplier.get(), PATH_SEPARATOR));
```
gpang, line 1133: Can we save this normalized root at the top, so both branches can re-use it?

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
```diff
@@ -1118,14 +1118,25 @@ protected boolean parentExists(String path) throws IOException {
...
-        normalizedPath,
+        path,
         PathUtils.normalizePath(mRootKeySupplier.get(), PATH_SEPARATOR));
-    if (!stripedKey.equals(normalizedPath)) {
+    if (!stripedKey.equals(path)) {
       return stripedKey;
     }
-    return CommonUtils.stripPrefixIfPresent(normalizedPath, PATH_SEPARATOR);
+
+    // See if adding a PATH_SEPARATOR makes a difference in stripping prefix
```
gpang, line 1128: What is an example where this will make a difference? `PathUtils.normalizePath` just adds a separator at the end, so does this only apply to the single string `s3://bucket`?

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
bf8086, line 1133: (nit) reuse the result from line 1123?

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
bradyoo: That is correct.

bradyoo:
Added test cases and simplified code based on gene's comment.

core/common/src/main/java/alluxio/underfs/ObjectUnderFileSystem.java
```diff
@@ -1118,14 +1118,16 @@ protected boolean parentExists(String path) throws IOException {
    */
   @VisibleForTesting
   public String stripPrefixIfPresent(String path) {
-    final String normalizedPath = PathUtils.normalizePath(path, PATH_SEPARATOR);
+    if (mRootKeySupplier.get().equals(path)) {
```
bradyoo, line 1121: mRootKeySupplier.get() is a memoized function so don't be concerned about callng twice.

bradyoo:

gpang:

gpang:

gpang:

PR #11783, opened 2020-07-20
Add more options to collectInfo command

jiacheliu3:
This adds 3 features to `collectInfo`

1. With `--exclude-logs <filename-prefixes>`, `--include-logs <filename-prefixes>` and `--additional-logs` the user is able to specify log files to take or not take. By default only Alluxio-relevant files under `${ALLUXIO_HOME}/logs` are included and unrecognized files are ignored.

For example, `alluxio collectInfo --additional-logs alluxio_gc --exclude-logs master collectLog <tmp-dir>` will include all `alluxio_gc*` files (supposedly for GC logs) and exclude all `master*` files like `master.log.*` and `master.out`.

`alluxio collectInfo --include-logs master collectLog <tmp-dir>` will, however, take only the `master*` files like `master.log` and `master.out`.

2. With `--start-time <datetime>` and `--end-time <datetime>` the user can specify a certain time window to cover. This serves the need where only events in one window is of interest. The files that end before the `start time` or starts after the `end time` will be ignored. The `end time` of a file is the last modification time. The `start time` of a file is inferred by reading the first set of rows.

3. `alluxio-site.properties` will not be collected as it can contain unmasked credentials. This is replaced with doing a local `alluxio getConf`.

### DIFF
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -839,7 +839,7 @@ public final class CommonUtils {
     File[] files = dir.listFiles();
     // File#listFiles can return null when the path is invalid
     if (files == null) {
-      return Collections.EMPTY_LIST;
+      return Collections.emptyList();
     }
     List<File> result = new ArrayList<>(files.length);
     for (File f : files) {
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -830,21 +830,21 @@ public final class CommonUtils {
   }
 
   /**
-   * Recursively lists a dir and all its subdirs and return all the files.
+   * Recursively lists a local dir and all its subdirs and return all the files.
    *
    * @param dir the directory
    * @return a list of all the files
    * */
-  public static List<File> recursiveListDir(File dir) {
+  public static List<File> recursiveListLocalDir(File dir) {
     File[] files = dir.listFiles();
     // File#listFiles can return null when the path is invalid
     if (files == null) {
-      return new ArrayList<>();
+      return Collections.EMPTY_LIST;
     }
     List<File> result = new ArrayList<>(files.length);
     for (File f : files) {
       if (f.isDirectory()) {
-        result.addAll(recursiveListDir(f));
+        result.addAll(recursiveListLocalDir(f));
         continue;
       }
       result.add(f);
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -837,6 +837,10 @@ public final class CommonUtils {
    * */
   public static List<File> recursiveListDir(File dir) {
     File[] files = dir.listFiles();
+    // File#listFiles can return null when the path is invalid
+    if (files == null) {
+      return new ArrayList<>();
+    }
     List<File> result = new ArrayList<>(files.length);
     for (File f : files) {
       if (f.isDirectory()) {
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -829,6 +829,12 @@ public final class CommonUtils {
     }
   }
 
+  /**
+   * Recursively lists a dir and all its subdirs and return all the files.
+   *
+   * @param dir the directory
+   * @return a list of all the files
+   * */
   public static List<File> recursiveListDir(File dir) {
     File[] files = dir.listFiles();
     List<File> result = new ArrayList<>(files.length);
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -35,6 +35,7 @@ import org.apache.commons.lang.ObjectUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.reflect.Constructor;
@@ -828,16 +829,18 @@ public final class CommonUtils {
     }
   }
 
-  public static final String[] VALID_DATE_FORMATS = new String[]{
-          "yyyy-mm-dd HH:mm:ss,SSS", // "2020-03-19 11:57:58,883"
-          "yyyy-MM-dd'T'HH:mm:ss.SSSXX", // 2020-06-23T23:52:23.970+0800
-          "yyyy/MM/dd HH:mm:ss",
-          "dd MMM yyyy HH:mm:ss",
-          "MM-dd-yyyy HH:mm:ss",
-          "yyyy/MM/dd",
-          "dd MMM yyyy",
-          "MM-dd-yyyy"
-  };
+  public static List<File> recursiveListDir(File dir) {
+    File[] files = dir.listFiles();
+    List<File> result = new ArrayList<>(files.length);
+    for (File f : files) {
+      if (f.isDirectory()) {
+        result.addAll(recursiveListDir(f));
+        continue;
+      }
+      result.add(f);
+    }
+    return result;
+  }
 
   private CommonUtils() {} // prevent instantiation
 }
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -828,5 +828,16 @@ public final class CommonUtils {
     }
   }
 
+  public static final String[] VALID_DATE_FORMATS = new String[]{
+          "yyyy-mm-dd HH:mm:ss,SSS", // "2020-03-19 11:57:58,883"
+          "yyyy-MM-dd'T'HH:mm:ss.SSSXX", // 2020-06-23T23:52:23.970+0800
+          "yyyy/MM/dd HH:mm:ss",
+          "dd MMM yyyy HH:mm:ss",
+          "MM-dd-yyyy HH:mm:ss",
+          "yyyy/MM/dd",
+          "dd MMM yyyy",
+          "MM-dd-yyyy"
+  };
+
   private CommonUtils() {} // prevent instantiation
 }
--- a/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
+++ b/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
@@ -80,8 +80,6 @@ public class CollectInfo extends AbstractShell {
           + "collectJvmInfo:     collects jstack from the JVMs.\n"
           + "collectLog:         collects the log files under ${ALLUXIO_HOME}/logs/.\n"
           + "collectMetrics:     collects Alluxio system metrics.\n\n"
-          + "<filename-prefixes> filename prefixes, separated by comma\n"
-          + "<datetime>          a datetime string like 2020-06-27T11:58:53\n"
           + "<outputPath>        the directory you want the collected tarball to be in\n\n"
           + "WARNING: This command MAY bundle credentials. To understand the risks refer "
           + "to the docs here.\nhttps://docs.alluxio.io/os/user/edge/en/operation/"
--- a/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
+++ b/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
@@ -64,7 +64,9 @@ import java.util.stream.Collectors;
 public class CollectInfo extends AbstractShell {
   private static final Logger LOG = LoggerFactory.getLogger(CollectInfo.class);
   private static final String USAGE =
-      "USAGE: collectInfo [--max-threads <threadNum>] [--local] [--help] COMMAND <outputPath>\n\n"
+      "USAGE: collectInfo [--max-threads <threadNum>] [--local] [--help] "
+          + "[--exclude-logs <filename-prefixes>] [--include-logs <filename-prefixes>] "
+          + "[--start-time <datetime>] [--end-time <datetime>] COMMAND <outputPath>\n\n"
           + "collectInfo runs a set of sub-commands which collect information "
           + "about your Alluxio cluster.\nIn the end of the run, "
           + "the collected information will be written to files and bundled into one tarball.\n"
@@ -78,6 +80,8 @@ public class CollectInfo extends AbstractShell {
           + "collectJvmInfo:     collects jstack from the JVMs.\n"
           + "collectLog:         collects the log files under ${ALLUXIO_HOME}/logs/.\n"
           + "collectMetrics:     collects Alluxio system metrics.\n\n"
+          + "<filename-prefixes> filename prefixes, separated by comma\n"
+          + "<datetime>          a datetime string like 2020-06-27T11:58:53\n"
           + "<outputPath>        the directory you want the collected tarball to be in\n\n"
           + "WARNING: This command MAY bundle credentials. To understand the risks refer "
           + "to the docs here.\nhttps://docs.alluxio.io/os/user/edge/en/operation/"
@@ -90,9 +94,10 @@ public class CollectInfo extends AbstractShell {
   // CMD_ALIAS map.
   private static final Set<String> UNSTABLE_ALIAS = ImmutableSet.of();
 
+  private static final String TARBALL_NAME = "alluxio-info.tar.gz";
+  private ExecutorService mExecutor;
+
   private static final String MAX_THREAD_OPTION_NAME = "max-threads";
-  private static final String LOCAL_OPTION_NAME = "local";
-  private static final String HELP_OPTION_NAME = "help";
   private static final Option THREAD_NUM_OPTION =
           Option.builder().required(false).longOpt(MAX_THREAD_OPTION_NAME).hasArg(true)
                   .desc("the number of threads this command uses\n"
@@ -100,46 +105,45 @@ public class CollectInfo extends AbstractShell {
                           + "Use a smaller number to constrain the network IO when "
                           + "transmitting tarballs.")
                   .build();
+  private static final String LOCAL_OPTION_NAME = "local";
   private static final Option LOCAL_OPTION =
           Option.builder().required(false).longOpt(LOCAL_OPTION_NAME).hasArg(false)
                   .desc("specifies this command should only collect information "
                           + "about the localhost")
                   .build();
+  private static final String HELP_OPTION_NAME = "help";
   private static final Option HELP_OPTION =
           Option.builder().required(false).longOpt(HELP_OPTION_NAME).hasArg(false)
                   .desc("shows the help message").build();
+  // Build the options for collectInfo, then aggregate local options for each sub-command
   private static final Options OPTIONS = loadOptions(new Options()
           .addOption(THREAD_NUM_OPTION)
           .addOption(LOCAL_OPTION)
           .addOption(HELP_OPTION));
 
+  // Load the options defined in each sub-command class
   private static Options loadOptions(Options options) {
     Reflections reflections = new Reflections(Command.class.getPackage().getName());
     for (Class<? extends Command> cls : reflections.getSubTypesOf(Command.class)) {
       try {
         for (Field f : cls.getDeclaredFields()) {
           if (f.getName().equals("OPTIONS")) {
-            System.out.format("Class %s found OPTIONS %n", cls.getName());
             Options clsOptions = ((Options) f.get(null));
             if (clsOptions == null) {
               continue;
             }
             for (Option o : clsOptions.getOptions()) {
-              System.out.format("Found option %s from class %s%n", o, cls.getName());
               options.addOption(o);
             }
           }
         }
       } catch (IllegalAccessException e) {
-        e.printStackTrace();
+        LOG.warn("Failed to load OPTIONS from class {}: {}",
+                cls.getCanonicalName(), e.getMessage());
       }
     }
     return options;
   }
-
-  private static final String TARBALL_NAME = "alluxio-info.tar.gz";
-
-  private ExecutorService mExecutor;
 
   /**
    * Creates a new instance of {@link CollectInfo}.
@@ -195,7 +199,6 @@ public class CollectInfo extends AbstractShell {
     CommandLine cmd;
     try {
       cmd = parser.parse(OPTIONS, argv, true /* stopAtNonOption */);
-      System.out.format("Parsed options %s%n", Arrays.toString(cmd.getOptions()));
     } catch (ParseException e) {
       return;
     }
@@ -316,7 +319,7 @@ public class CollectInfo extends AbstractShell {
         String fromPath = Paths.get(targetDir, CollectInfo.TARBALL_NAME)
                 .toAbsolutePath().toString();
         String toPath = tarballFromHost.getAbsolutePath();
-        LOG.debug("Copying %s:%s to %s", host, fromPath, toPath);
+        LOG.debug("Copying {}:{} to {}", host, fromPath, toPath);
 
         try {
           CommandReturn cr =
@@ -324,7 +327,7 @@ public class CollectInfo extends AbstractShell {
           return cr;
         } catch (IOException e) {
           // An unexpected error occurred that caused this IOException
-          LOG.error("Execution failed %s", e);
+          LOG.error("Execution failed on {}", e);
           return new CommandReturn(1, e.toString());
         }
       }, mExecutor);
@@ -405,7 +408,7 @@ public class CollectInfo extends AbstractShell {
       // Case 2. Execute a single command
       try {
         int childRet = executeAndAddFile(args, cmdLine, filesToCollect);
-        if (ret == 0 && childRet != 0) {
+        if (childRet != 0) {
           ret = childRet;
         }
       } catch (AlluxioException e) {
@@ -533,11 +536,10 @@ public class CollectInfo extends AbstractShell {
   @Override
   protected Map<String, Command> loadCommands() {
     // Give each command the configuration
-    Map<String, Command> commands = CommandUtils.loadCommands(
+    return CommandUtils.loadCommands(
             CollectInfo.class.getPackage().getName(),
             new Class[] {FileSystemContext.class},
             new Object[] {FileSystemContext.create(mConfiguration)});
-    return commands;
   }
 
   @Override
--- a/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
+++ b/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
@@ -239,6 +239,22 @@ public class CollectInfo extends AbstractShell {
     System.exit(ret);
   }
 
+  // Convert the command line back to a list of arguments
+  private List<String> cmdLineToArgs(CommandLine cmd) {
+    List<String> args = new ArrayList<>();
+    for (Option opt : cmd.getOptions()) {
+      if (opt.equals(LOCAL_OPTION) || opt.equals(THREAD_NUM_OPTION)) {
+        continue;
+      }
+      args.add("--" + opt.getLongOpt());
+      if (opt.hasArg()) {
+        args.add(opt.getValue());
+      }
+    }
+    args.addAll(cmd.getArgList());
+    return args;
+  }
+
   /**
    * Finds all nodes in the cluster.
    * Then invokes collectInfo with --local option on each of them locally.
@@ -279,7 +295,9 @@ public class CollectInfo extends AbstractShell {
 
         String[] collectInfoArgs =
                 (String[]) ArrayUtils.addAll(
-                        new String[]{alluxioBinPath, "collectInfo", "--local"}, args);
+                        new String[]{alluxioBinPath, "collectInfo", "--local"},
+                        cmdLineToArgs(cmdLine).toArray(new String[0]));
+        System.out.format("Invoking command %s%n", Arrays.toString(collectInfoArgs));
         try {
           CommandReturn cr = ShellUtils.sshExecCommandWithOutput(host, collectInfoArgs);
           return cr;
--- a/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
+++ b/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
@@ -21,7 +21,6 @@ import alluxio.conf.PropertyKey;
 import alluxio.conf.Source;
 import alluxio.exception.AlluxioException;
 import alluxio.shell.CommandReturn;
-import alluxio.util.CommonUtils;
 import alluxio.util.ConfigurationUtils;
 import alluxio.util.ShellUtils;
 import alluxio.util.io.FileUtils;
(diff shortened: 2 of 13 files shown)


shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
```diff
@@ -404,7 +466,7 @@ private int executeAndAddFile(String[] argv, List<File> filesToCollect) throws I
       printHelp(String.format("%s is an unknown command.%n", subCommand));
       return 1;
     }
-    int ret = run(argv);
+    int ret = cmd.run(cmdLine);
```
jiacheliu3, line 469: The parsed options will propagate to subcommands.

shell/src/main/java/alluxio/cli/bundler/command/AbstractCollectInfoCommand.java
```diff
@@ -56,12 +56,15 @@ public void validateArgs(CommandLine cl) throws InvalidArgumentException {
    * */
   public String getWorkingDirectory(CommandLine cl) {
     String[] args = cl.getArgs();
-    String baseDirPath = args[0];
+    String baseDirPath = args[1];
```
jiacheliu3, line 59: The index changed cuz the invocation method changed.

```
-    int ret = run(argv);
+    int ret = cmd.run(cmdLine);
```

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -53,15 +156,149 @@ public boolean hasSubCommand() {
...
+
+    List<File> allFiles = CommonUtils.recursiveListDir(mLogDir);
+    for (File f : allFiles) {
+      String relativePath = getRelativePathToLogDir(f);
+      try {
+        if (!shouldCopy(f, relativePath, checkTimeStamp)) {
+          continue;
+        }
+        File targetFile = new File(mWorkingDirPath, relativePath);
+        FileUtils.copyFile(f, targetFile, true);
```
jiacheliu3, line 205: This will create parent directories if not existent.

shell/src/test/java/alluxio/cli/bundler/InfoCollectorTestUtils.java
```diff
@@ -37,12 +44,37 @@ public static File createTemporaryDirectory() {
...
+    return newDir;
+  }
+
   public static void create() {
     Files.createTempDir();
   }
+
+  public static void verifyAllFiles(File targetDir, Set<String> expectedFiles) throws IOException {
+    Set<String> copiedFiles = getAllFileNamesRelative(targetDir, targetDir);
+    assertEquals(expectedFiles, copiedFiles);
```
jiacheliu3, line 64: This invokes `Set#equals` which compares the size and checks existence of all elements.

shell/src/test/java/alluxio/cli/bundler/command/CollectLogCommandTest.java
```diff
@@ -19,56 +22,500 @@
...
+    }
+
+    // Some logs are created after this time, should be ignored
+    File masterLog = new File(mTestDir, "master.log");
+    String log = "2020-07-08 18:53:45,129 INFO  CopycatServer - Server started successfully!\n"
+            + "2020-07-08 18:53:59,129 INFO  RaftJournalSystem - Started Raft Journal System..\n"
+            + "2020-07-09 00:01:59,135 INFO  DefaultMetaMaster - Standby master with address...\n"
+            + "2020-07-09 00:03:59,135 INFO  AlluxioMasterProcess - All masters started\n"
+            + "2020-07-09 01:12:59,138 INFO  AbstractPrimarySelector - Primary selector..\n"
+            + "2020-07-09 01:53:59,139 INFO  AbstractMaster - TableMaster: Stopped secondary..";
```
jiacheliu3, line 209: It might be controversial whether we use real content or abstractions here. I would argue real logs capture more.

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
```diff
@@ -61,7 +64,9 @@
 public class CollectInfo extends AbstractShell {
   private static final Logger LOG = LoggerFactory.getLogger(CollectInfo.class);
   private static final String USAGE =
-      "USAGE: collectInfo [--max-threads <threadNum>] [--local] [--help] COMMAND <outputPath>\n\n"
+      "USAGE: collectInfo [--max-threads <threadNum>] [--local] [--help] "
+          + "[--exclude-logs <filename-prefixes>] [--include-logs <filename-prefixes>] "
```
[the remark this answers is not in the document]
jiacheliu3: The help messages for options are printed at https://github.com/Alluxio/alluxio/blob/c1ea6f3c14812930f014f830d44a79b94574c23e/shell/src/main/java/alluxio/cli/bundler/CollectInfo.java#L154 Basically the description for each option is shown.

For a message that is more verbose, like working examples and how they can be organized to serve use cases, I guess that better suits to be in the document (linked in the help message for CollectInfo). What do you think?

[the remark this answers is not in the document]
gpang: ahhh, ok. Now, I see the sub-command help. Then, why does this top level usage help mention start-time and include-logs, etc, when it is only a sub command option?

shell/src/main/java/alluxio/cli/bundler/command/AbstractCollectInfoCommand.java
```diff
@@ -56,12 +56,15 @@ public void validateArgs(CommandLine cl) throws InvalidArgumentException {
...
     String workingDirPath =  Paths.get(baseDirPath, this.getCommandName()).toString();
     LOG.debug("Command %s works in %s", this.getCommandName(), workingDirPath);
     // mkdirs checks existence of the path
     File workingDir = new File(workingDirPath);
-    workingDir.mkdirs();
+    if (!workingDir.exists()) {
+      System.out.format("Working path %s does not exist. mkdir first%n", workingDirPath);
```
gpang, line 65: This message makes it sound like the user should run mkdir. Instead, this could say:

"Creating working directory: %s %n"

So it sounds like the command is creating the directory.

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
```diff
@@ -30,6 +36,10 @@
   public static final String COMMAND_NAME = "collectConfig";
   private static final Logger LOG = LoggerFactory.getLogger(CollectConfigCommand.class);
 
+  private static final Set<String> EXCLUDED_FILES = Stream.of(
+          Constants.SITE_PROPERTIES
```
gpang, line 40: is this to avoid copying credentials?

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
```diff
@@ -30,6 +36,10 @@
...
+  ).collect(Collectors.toSet());
```
gpang, line 41: Would `ImmutableSet.of(...)` work?

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
```diff
@@ -52,10 +62,21 @@ public boolean hasSubCommand() {
...
+    String confDirPath = mFsContext.getClusterConf().get(PropertyKey.CONF_DIR);
 
-    // TODO(jiacheng): phase 2 copy intelligently, check security risks
-    FileUtils.copyDirectory(new File(confDir), new File(mWorkingDirPath), true);
+    File confDir = new File(confDirPath);
+    List<File> allFiles = CommonUtils.recursiveListDir(confDir);
+    for (File f : allFiles) {
+      String relativePath = confDir.toURI().relativize(f.toURI()).getPath();
+      // Ignore file prefixes to exclude
+      for (String prefix : EXCLUDED_FILES) {
```
gpang, line 72: If these are supposed to prefixes, the variable name should be something like `EXCLUDED_FILE_PREFIXES` or something.

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
```diff
@@ -52,10 +62,21 @@ public boolean hasSubCommand() {
...
+        if (relativePath.startsWith(prefix)) {
```
gpang, line 73: Is `relativePath` just the filename part?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -71,4 +308,34 @@ public String getUsage() {
...
+   * Returns null if unable to identify a datetime.
+   *
+   * @param s a log entry
+   * @return identified datetime
+   * */
+  @Nullable
+  public static LocalDateTime parseDateTime(String s) {
+    for (Map.Entry<String, Integer> entry : FORMAT_TO_LEN.entrySet()) {
+      String f = entry.getKey();
+      int len = entry.getValue();
```
gpang, line 323: Is `String.length()` slow?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,121 @@
...
+      "user"
+  ).collect(Collectors.toSet());
+  // We tolerate the beginning of a log file to contain some rows that are not timestamped.
+  // 30 is chosen because a YARN application log can have >20 rows in the beginning for
+  // general information about a job.
+  // The timestamped log entries start after this general information block.
+  private static final int TRY_PARSE_LOG_ROWS = 30;
+
+  // Preserves the order of iteration, we try the longer pattern before the shorter one
+  private static final Map<String, Integer> FORMAT_TO_LEN = new LinkedHashMap<String, Integer>(){
```
gpang, line 75: Why can't this just be a list? Why do we need to cache the length?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,121 @@
...
+      put("yyyy-MM-dd HH:mm:ss", 19);
+      put("yyyy-MM-dd HH:mm", 16);
+      put("yy/MM/dd HH:mm:ss", 17);
+      put("yy/MM/dd HH:mm", 14);
+      put("yyyy-MM-dd'T'HH:mm:ss.SSSXX", 28);
+      put("yyyy-MM-dd'T'HH:mm:ss", 19);
+      put("yyyy-MM-dd'T'HH:mm", 16);
+    }
+  };
+  private static final Map<String, DateTimeFormatter> STRING_TO_FORMATTER = initFormatters();
```
gpang, line 87: why can't we just have an ordered list of formatters? Why do we need the string representation?

shell/src/main/java/alluxio/cli/bundler/CollectInfo.java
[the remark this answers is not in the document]
jiacheliu3: I extracted these option arguments here because they are shared by two options each. Seems it has more confusion that benefit, I will move it back to under the option.

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
jiacheliu3: Yes this avoids copying unmasked credentials from alluxio-site.properties. We now do a local "alluxio getConf" which prints the credentials masked, and collect the output.

However if the credentials are in alluxio-env.sh or elsewhere (given via ALLUXIO_JAVA_OPTS for example), I don't have a way to skip/mask them.

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
jiacheliu3: Your are right, when checking the prefix it should be checking against the filename. The `relativePath` is the path relative to the conf dir, when the config files can be in nested directories. I think we still need the `relativePath` when copying, as we want to preserve the directory structure.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: The thing is, the real length of the datetime string like "2020-07-25T12:05:12.123+0800" is not the same as length of the pattern "yyyy-MM-dd'T'HH:mm:ss.SSSXX". Note that the single quotes around `T` should not be counted and `XX` expands to "+0800".

Yea I should add examples to these formats.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: We need the length because the `DateTimeFormatter` CANNOT parse anything other than the datetime. For example a DateTimeFormatter of format "yyyy-mm-dd" cannot parse "2020-07-15 INFO i did something".  I have to cut the target string before parsing. 

One example of existing wheels is `org.apache.commons.lang.time.DateUtils.parse(String s)`. It builds its own state machine but still cannot avoid the need to index and cut the string `s`.

All in all, for each format I try, i need a DateTimeFormatter and a length of how much string I need. I've moved these two into one `LinkedHashMap` now, instead of using two separated maps.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: Yes the string representation is not necessary. It's removed and I only keep one map now instead of two.

jiacheliu3:
```
Tests in error: 
alluxio.server.ft.journal.ufs.UfsJournalIntegrationTest.file(alluxio.server.ft.journal.ufs.UfsJournalIntegrationTest)
  Run 1: UfsJournalIntegrationTest.file:328->fileTestUtil:332->createFsMasterFromJournal:603 » Runtime
  Run 2: UfsJournalIntegrationTest.file:328->fileTestUtil:332->createFsMasterFromJournal:603 » Runtime
```

jiacheliu3:
alluxio-bot, test this please

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -828,5 +829,28 @@ public static boolean isAddressReachable(String hostname, int port) {
...
+   * Recursively lists a local dir and all its subdirs and return all the files.
+   *
+   * @param dir the directory
+   * @return a list of all the files
+   * */
+  public static List<File> recursiveListLocalDir(File dir) {
+    File[] files = dir.listFiles();
+    // File#listFiles can return null when the path is invalid
+    if (files == null) {
+      return Collections.EMPTY_LIST;
```
gpang, line 842: I think `Collections.emptyList()` will work better with type safety?

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
```diff
@@ -52,10 +60,22 @@ public boolean hasSubCommand() {
...
-    FileUtils.copyDirectory(new File(confDir), new File(mWorkingDirPath), true);
+    File confDir = new File(confDirPath);
+    List<File> allFiles = CommonUtils.recursiveListLocalDir(confDir);
+    for (File f : allFiles) {
+      String filename = f.getName();
+      String relativePath = confDir.toURI().relativize(f.toURI()).getPath();
+      // Ignore file prefixes to exclude
+      for (String prefix : EXCLUDED_FILE_PREFIXES) {
+        if (filename.startsWith(prefix)) {
+          continue;
```
gpang, line 73: Doesn't this `continue` just move to the next entry in `EXCLUDED_FILE_PREFIXES`? Don't we want to go to the next entry in `allFiles` instead?

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
```diff
@@ -52,10 +60,22 @@ public boolean hasSubCommand() {
...
+        }
+        File targetFile = new File(mWorkingDirPath, relativePath);
```
gpang, line 75: I think this block needs to be outside the iteration over all `EXCLUDED_FILE_PREFIXES`.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,122 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Command to collect Alluxio logs.
  * */
 public class CollectLogCommand  extends AbstractCollectInfoCommand {
   public static final String COMMAND_NAME = "collectLog";
   private static final Logger LOG = LoggerFactory.getLogger(CollectLogCommand.class);
+  public static final Set<String> FILE_NAMES = Stream.of(
```
gpang, line 51: These are prefixes, in order to handle `master.log.1`, right? Should we include "prefix" in the name of the variable?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,122 @@
...
+      "proxy.out",
+      "task.log",
+      "task.out",
+      "user"
+  ).collect(Collectors.toSet());
+  // We tolerate the beginning of a log file to contain some rows that are not timestamped.
+  // 30 is chosen because a YARN application log can have >20 rows in the beginning for
+  // general information about a job.
+  // The timestamped log entries start after this general information block.
+  private static final int TRY_PARSE_LOG_ROWS = 30;
```
gpang, line 71: I wonder if we need to make this larger, in order to ignore any long stack trace at the beginning of the file or something?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,122 @@
...
+
+  // Preserves the order of iteration, we try the longer pattern before the shorter one
```
gpang, line 73: comment on what the second field is (what that length represents)

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,122 @@
...
+  private Set<String> mIncludedPrefix;
+  private Set<String> mExcludedPrefix;
+  private LocalDateTime mStartTime;
+  private LocalDateTime mEndTime;
+
+  public static final String INCLUDE_OPTION_NAME = "include-logs";
+  private static final Option INCLUDE_OPTION =
+          Option.builder().required(false).argName("filename-prefixes")
+                  .longOpt(INCLUDE_OPTION_NAME).hasArg(true)
+                  .desc("extra log file name prefixes to include in ${ALLUXIO_HOME}/logs. "
```
gpang, line 108: what is the order of operation between include and exclude? For example, if `master.log` is in both include and exclude, what happens?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,122 @@
...
+          Option.builder().required(false).argName("filename-prefixes")
+                  .longOpt(EXCLUDE_OPTION_NAME).hasArg(true)
+                  .desc("extra log file name prefixes to exclude in ${ALLUXIO_HOME}/logs. "
+                          + "The files that start with the prefix will be excluded.\n"
+                          + "<filename-prefixes> filename prefixes, separated by comma").build();
+  private static final String START_OPTION_NAME = "start-time";
+  private static final Option START_OPTION =
+          Option.builder().required(false).argName("datetime")
+                  .longOpt(START_OPTION_NAME).hasArg(true)
+                  .desc("logs that do not contain entries after this time will be ignored\n"
```
gpang, line 122: Does start time require end time?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,122 @@
...
+                          + "<datetime> a datetime string like 2020-06-27T11:58:53").build();
+  private static final String END_OPTION_NAME = "end-time";
+  private static final Option END_OPTION =
+          Option.builder().required(false).argName("datetime")
+                  .longOpt(END_OPTION_NAME).hasArg(true)
+                  .desc("logs that do not contain entries before this time will be ignored\n"
```
gpang, line 128: Does end time require start time?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -53,15 +157,144 @@ public boolean hasSubCommand() {
...
 
     // TODO(jiacheng): phase 2 Copy intelligently find security risks
-    // TODO(jiacheng): phase 2 components option
-    FileUtils.copyDirectory(new File(logDir), new File(mWorkingDirPath), true);
+    mIncludedPrefix = new HashSet<>(FILE_NAMES);
+    // Define include list and exclude list
+    if (cl.hasOption(INCLUDE_OPTION_NAME)) {
+      Set<String> toInclude = parseFileNames(cl.getOptionValue(INCLUDE_OPTION_NAME));
+      System.out.format("Include the following filename prefixes: %s%n", toInclude);
+      mIncludedPrefix.addAll(toInclude);
```
gpang, line 167: `mIncludedPrefix` starts off with the entire list. Does that mean it can never be reduced? For example, if I use the option `--include master.log` this will still retrieve all logs because `mIncludedPrefix = new HashSet<>(FILE_NAMES)`?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -53,15 +157,144 @@ public boolean hasSubCommand() {
...
+    for (String s : mIncludedPrefix) {
+      if (fileName.startsWith(s)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private boolean fileTimeStampIsWanted(File f) throws IOException {
+    long timestamp = f.lastModified();
```
gpang, line 249: We should guard this with `if (mEndTime != null)` and in the block, return false if it fails.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -53,15 +157,144 @@ public boolean hasSubCommand() {
...
+    LocalDateTime fileEndTime =
+            LocalDateTime.ofInstant(Instant.ofEpochMilli(timestamp), ZoneId.systemDefault());
+
+    // Infer file start time by parsing the first bunch of rows
```
gpang, line 253: We should guard this with `if (mStartTime != null)` and in the block, return false if it fails.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -53,15 +157,144 @@ public boolean hasSubCommand() {
...
+        }
+        r++;
+      }
+    }
+    return null;
+  }
+
+  private Set<String> parseFileNames(String input) {
+    Set<String> names = new HashSet<>();
+    names.addAll(Stream.of(input.split(",")).map(String::trim).collect(Collectors.toList()));
```
gpang, line 294: will this skip empty strings?

shell/src/main/java/alluxio/cli/bundler/command/CollectConfigCommand.java
jiacheliu3: Yea I messed up with two nested loops. I've changed this to use stream and only one loop. PTAL.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: That's a good question. In fact, I believe log files all start with a complete log entry (meaning the line starts with a datetime if log4j configures normally). I've never seen a log file that starts with half of the log traces. 

Besides if the long broken stack traces are possible, they can have nested exceptions (A ... caused by B... caused by C...) can be like 100 lines. I can probably change it to 100 lines. But there can be 100 log files rotating for each log group (master.log.1 ~ 99, worker.log.1 ~ 99 etc), I think it might get up to non-trivial much file parsing.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: No it doesn't. When not specified, we keep everything that contains or starts after this start time with no end time. Added to the description.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: Vice versa to the start time.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: I think since we don't require start time to have end time, we want it to default to true unless the file is filtered out with the specified time interval.

Note that the special case is we are not able to infer the start time of the file, we lean towards keeping it.

Updated the method description, hoping it makes the logic clearer.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: same

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: Yes

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: Great question. This `--include-logs` was designed for extra logs that we don't know, like "alluxio_gc.log" which some people use for gc logs. These logs should be collected additional to all the rest of logs.

The use case you are thinking is also valid, where people just want a subset, like "master.log" and "master.out". This was not in the scope.

I think I will change the current design to:
`--include-logs <file-names>` to replace the included list. Eg. `--include-logs "master.log, master.out"` copies only master logs and master outs.
`--additional-logs <file-names>` adds sth extra to the include list. Eg. `--additional-logs "alluxio_gc.log"` takes the specified gc logs tgt with all the rest of logs.

jiacheliu3: This is added.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: The definition of this `--include-logs` option was not very clear. Now it's changed to the below 3 options:

`--include-logs`: replace the filename prefixes with what is defined here. Does not work tgt with the other two options (err out if given together).

`--exclude-logs`: minus from the default filenames. This is checked before `--additional-logs`. If "master.log" is in both `exclude` and `additional`, it will be excluded. This caveat is now in the option description.

`--additional-logs`: add to the default filenames. Checked after the exclusion.

jiacheliu3:
```
05:14:35.240 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project alluxio-core-base: ExecutionException The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```

jiacheliu3:
alluxio-bot, test this please

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
gpang: O ok. If the start of the log always starts with a complete log entry, then why do we need this parsing limit?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
```diff
@@ -14,21 +14,148 @@
...
+  private Set<String> mIncludedPrefix;
+  private Set<String> mExcludedPrefix;
+  private LocalDateTime mStartTime;
+  private LocalDateTime mEndTime;
+
+  public static final String INCLUDE_OPTION_NAME = "include-logs";
+  public static final String EXCLUDE_OPTION_NAME = "exclude-logs";
+  public static final String ADDITIONAL_OPTION_NAME = "additional-logs";
+  private static final Option INCLUDE_OPTION =
+          Option.builder().required(false).argName("filename-prefixes")
```
gpang, line 116: I think maybe we can achieve everything with only the `--include` flag, but that flag simply specifies the files to get. We can set the default for this flag as the list of `FILE_NAMES_PREFIXES`.

However, if a user wants to exclude or include files, then they would have to add a `--include` with the full list they want.

If that is too unfriendly, then can't we just assume we download everything under `logs/`, and only have an `--exclude` flag, so any additional logs will automatically get retrieved, and the exclude is a simple way to not download some.

I just thought having 3 flags was too much. Ideally we only need one. What do you think?

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: Ah gotcha. That makes sense to me. Updated

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: I guess keeping only the `--include` will make it too verbose. For example I want everything plus my gc logs, then I will have to type `--include-logs master,worker,job-master,job-worker,user,task,alluxio_gc`. For less experienced users we may need to teach them how to make sure all components are included, to avoid another iteration. This makes `--additional` useful. For similar argument I would suggest keeping `--exclude` too.

On your other option, where we keep everything unless explicitly excluded. I've seen cases where some users put many irrelevant logs under Alluxio log dir and `collectInfo` is run by a general L1 team that just follow handbooks without looking into the Alluxio log dir. Then hundreds of blue talon and ZK logs are copied (10M * 100 for each node) and the collectInfo tarball size went to the ceiling. I hope to avoid this kind of pitfall.

gpang: Ahhh, I see. I didn't realize there would be so many extra logs like zk or others.

Why do we need 3 flags? Can't we just have the internal list of what we know, and just 2 flags `--exclude` and `--additional`?

gpang:
@jiacheliu3 Thanks! Just one more remaining comment.

shell/src/main/java/alluxio/cli/bundler/command/CollectLogCommand.java
jiacheliu3: Logically we can remove the `--include` flag but that will make it a bit harder to just keep logs for one or two components. We have six different groups of log/out files `master,worker,job-master,job-worker,user,task` and I don't expect users to be very clear about what to exclude.

It gets trickier if we want finer granularity, like "get the audit logs", which translates to `--include-logs master_audit` or `--exclude master.log, master.out, job-master, worker, job-worker, task, user`.

gpang: Ok, for now, this should be fine.

gpang:

jiacheliu3:

PR #11797, opened 2020-07-21
[DOCFIX] Update Code-Convention Page with Javadoc guide

apc999:


docs/en/contributor/Code-Conventions.md
```diff
@@ -75,6 +75,57 @@ warnings are introduced:
...
+ * @param param_n description of nth parameter
+ * @return description of return argument (if applicable)
+ * @throws exception_1 description of 1st exception case
+ * ...
+ * @throws exception_n description of nth exception case
+ */
+```
+- An exception to the above rule is that `@throws` doesn’t need to be provided for `@Test` methods,
+or for generic exceptions like IOException when there is nothing interesting to document.
+- Do not put unchecked exceptions like `RuntimeException` in javadoc unless it is critical for this methd.
```
gpang, line 110: We may want to mention something like this (https://docs.alluxio.io/os/user/stable/en/contributor/Code-Conventions.html#selectively-javadoc-exceptions) here as well.

gpang:
From here (https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#descriptions) there are also guidelines for writing the description. Some we can make explicit here are:

- first sentence should be a concise summary of the class or method.
- the description should generally be implementation-independent.

We might also want descriptions (not the first sentence) to include any significant performance implications.

docs/en/contributor/Code-Conventions.md
apc999: done

apc999:
@gpang  updated , Ptal

gpang:

apc999:

PR #11799, opened 2020-07-21
Upgrade parquet libraries version from 1.10.0 to 1.11.0

bradyoo:
Parquet libraries started having Java 11 CI from 1.11.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,7 +146,7 @@
     <powermock.version>1.6.2</powermock.version>
     <prometheus.version>0.8.0</prometheus.version>
     <guava.version>27.0.1-jre</guava.version>
-    <parquet.version>1.10.0</parquet.version>
+    <parquet.version>1.11.0</parquet.version>
     <protobuf.version>3.11.0</protobuf.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -146,7 +146,7 @@
     <powermock.version>1.6.2</powermock.version>
     <prometheus.version>0.8.0</prometheus.version>
     <guava.version>27.0.1-jre</guava.version>
-    <parquet.version>1.10.0</parquet.version>
+    <parquet.version>1.11.0</parquet.version>
```
ZacBlanco, line 149: Is parquet a transitive dependency of any other libraries in our codebase? i.e. any of the hadoop clients/classes

bradyoo: I took a look here https://mvnrepository.com/artifact/org.apache.parquet/parquet-avro/usages?p=1 and overall seems fine.

ZacBlanco: Can you run `mvn dependency:graph` and look for the parquet usages?

ZacBlanco: I took a look and it looks like the only artifact that may conflict is the `parquet-hadoop-bundle` (v1.8.1) but that dependency is already on a different version anyway. So I think it's ok

ZacBlanco:

PR #11801, opened 2020-07-22
Use longs for file offsets in stress client io

gpang:
Using longs supports larger file sizes.

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/client/StressClientIOBench.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/client/StressClientIOBench.java
@@ -285,8 +285,8 @@ public class StressClientIOBench extends Benchmark<ClientIOTaskResult> {
     private final byte[] mBuffer;
     private final ByteBuffer mByteBuffer;
     private final int mThreadId;
-    private final int mFileSize;
-    private final int mMaxOffset;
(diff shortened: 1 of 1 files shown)


stress/shell/src/main/java/alluxio/stress/cli/client/StressClientIOBench.java
```diff
@@ -377,7 +377,7 @@ private int applyOperation() throws IOException {
           mInStream = mFs.open(mFilePath);
         }
         if (mParameters.mReadRandom) {
-          mCurrentOffset = mRandom.nextInt(mMaxOffset);
+          mCurrentOffset = Math.abs(mRandom.nextLong()) % mMaxOffset;
```
ZacBlanco, line 380: interesting. Is there no `Random#nextLong(long)` API?

gpang: Yeah, there isn't. However, I updated this to use the long stream.

ZacBlanco:
I think the PR is fine as it is, but do you think it would be worth adding a seed parameter so that we can make reproducible tests? or run equivalent workloads?

gpang:
Sure, that can come in a future PR.

gpang:

PR #11808, opened 2020-07-22
Upgrade commons-lang3 from 3.4 to 3.11

bradyoo:
hive-metastore has a dependency to commons-lang3 as well which will also get upgraded from 3.4 to 3.11.


ZacBlanco:
> The only dependency I saw for this was from hive-metastore

I don't quite understand what you mean here. Do you mean that `hive-metastore` uses this dependency?

bradyoo:
Yeah hive-metastore uses this dependency.

ZacBlanco:

PR #11813, opened 2020-07-23
[SMALLFIX] Do not trigger async caching on posRead if source is local

calvinjia:
Small fix to make posRead behavior mimic normal read behavior. If the source is local, we have already cached the data. Unfortunately our testing framework is not fine grained enough to track the RPC calls (see AlluxioFileInStreamTest#L752). 
In general, our strategy of sending async cache requests has been aggressive because we believed it is cheap for the worker to filter out duplicate requests. If this is not the case, we can probably further decrease the number of async cache requests we make.


calvinjia:

apc999:

PR #11814, opened 2020-07-23
Enable the JVM pause monitor by default

ZacBlanco:
In addition to turning it on by default;
- Many new tests were added. Code coverage is 82/89 (92%)
- Minor tweaks and improvements to the class

### DIFF
--- a/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
+++ b/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
@@ -54,7 +54,7 @@ public class JvmPauseMonitor {
   /** Times extra sleep time exceed INFO. */
   private final AtomicLong mInfoTimeExceeded = new AtomicLong();
   /** Total extra sleep time. */
-  private AtomicLong mTotalExtraTimeMs = new AtomicLong();
+  private final AtomicLong mTotalExtraTimeMs = new AtomicLong();
 
   private Thread mJvmMonitorThread;
 
--- a/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
+++ b/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
@@ -68,8 +68,6 @@ public class JvmPauseMonitor {
     Preconditions.checkArgument(gcSleepIntervalMs > 0, "gc sleep interval must be > 0");
     Preconditions.checkArgument(warnThresholdMs > 0, "warn threshold must be > 0");
     Preconditions.checkArgument(infoThresholdMs > 0, "info threshold must be > 0");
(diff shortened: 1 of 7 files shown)


ZacBlanco:
jenkins, test this please

ZacBlanco:
jenkins, test this please

gpang:
Jenkins, test this please

ZacBlanco:
jenkins, test this please

core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
```diff
@@ -48,11 +50,11 @@
...
 
   /** Times extra sleep time exceed WARN. */
-  private long mWarnTimeExceeded = 0;
+  private final AtomicLong mWarnTimeExceeded = new AtomicLong();
   /** Times extra sleep time exceed INFO. */
-  private long mInfoTimeExceeded = 0;
+  private final AtomicLong mInfoTimeExceeded = new AtomicLong();
   /** Total extra sleep time. */
-  private long mTotalExtraTimeMs = 0;
+  private AtomicLong mTotalExtraTimeMs = new AtomicLong();
```
gpang, line 57: Why is this also not `final`?

ZacBlanco: fixed

core/server/master/src/test/java/alluxio/master/AlluxioMasterProcessTest.java
```diff
@@ -156,7 +158,7 @@ private void waitForServing(ServiceType service) throws TimeoutException, Interr
       } catch (IOException e) {
         throw new RuntimeException(e);
       }
-    });
+    }, WaitForOptions.defaults().setTimeoutMs(5 * Constants.MINUTE_MS));
```
gpang, line 161: what was this change for?

ZacBlanco: Previously, this test would hang for multiple hours if the master/worker failed to come up. Now, it will time out.

gpang: But, is this unrelated to jvm pause monitor tests? If so, I think this should be its own PR/commit.

ZacBlanco: This was kind-of related. A bug in the monitor usage caused this test to fail (hang) for 2+ hours. It was very hard to track down with jenkins because I couldn't tell which tests are running when jenkins aborts the job. It would have been easier to find had this test contained a timeout.

gpang:

PR #11833, opened 2020-07-27
Update guava from 27.0.1-jre to 29.0-jre

bradyoo:
Other dependencies that got it's guava updated as a result:

hive-metastore: 27.0.1-jre -> 29.0-jre


yuzhu:

PR #11834, opened 2020-07-27
Upgrade hadoop version from 2.7.3 to 3.3.0

bradyoo:
All dependencies to it appear to be internal

Didn't touch underfs/hdfs/pom.xml's hadoop version (changing this to 3.3.0 causes breakage)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
     <grpc.version>1.28.1</grpc.version>
     <netty.version>4.1.45.Final</netty.version>
     <rocksdb.version>5.15.10</rocksdb.version>
-    <hadoop.version>2.7.3</hadoop.version>
+    <hadoop.version>3.3.0</hadoop.version>
     <jacoco.version>0.8.5</jacoco.version>
     <java.version>1.8</java.version>
     <jersey.version>2.29.1</jersey.version>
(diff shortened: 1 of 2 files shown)


yuzhu:
@calvinjia @apc999 PTAL at the hdfs client dependency change. Does this require any changes in generateTarball script?

apc999:
yes we need to update the generate-tarball script
please also check underfs-hdfs module and I believe there are some macros checking Hadoop versions and switch code there.

taking a step back, @bradyoo  @calvinjia  what's the motivation of bumping hadoop version to 3.3 ?

yuzhu:
java 11 support

shaded/hadoop/pom.xml
```diff
@@ -27,7 +27,7 @@
     <!-- The build path need to be defined here as well as in the parent pom so that mvn can -->
     <!-- run properly from sub-project directories -->
     <build.path>${project.parent.parent.basedir}/build</build.path>
-    <ufs.hadoop.version>2.7.3</ufs.hadoop.version>
+    <ufs.hadoop.version>3.3.0</ufs.hadoop.version>
```
calvinjia, line 30: Do we need to update the default profile to ufs-hadoop-3?

pom.xml
```diff
@@ -132,7 +132,7 @@
     <grpc.version>1.28.1</grpc.version>
     <netty.version>4.1.45.Final</netty.version>
     <rocksdb.version>5.15.10</rocksdb.version>
-    <hadoop.version>2.7.3</hadoop.version>
+    <hadoop.version>3.3.0</hadoop.version>
```
calvinjia, line 135: Does this cause any issues with the Hadoop compatible API? Is the FileSystem interface backward compatible?

yuzhu:
included in #11851 . Closing this for now

PR #11858, opened 2020-07-29
Shade out mozilla/public-suffix-list.txt

bradyoo:


bradyoo:
@LuQQiu httpclient's mozilla/public-suffix-list.txt is causing duplicate resource issues when rebundling, I don't think this is the right way of shading out. Can you help?

yuzhu:

PR #11867, opened 2020-07-31
Force target to be java 8 even if we are using jdk 11 to compile

yuzhu:
We need our final artifact to be compatible with java 8 even if we are running jdk 11 compiler.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1304,7 +1304,7 @@
         <jdk>11</jdk>
       </activation>
       <properties>
-        <java.version>8</java.version>
+        <java.version>1.8</java.version>
       </properties>
       <dependencyManagement>
         <dependencies>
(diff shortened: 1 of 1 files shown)


yuzhu:

pom.xml
```diff
@@ -1304,7 +1304,7 @@
         <jdk>11</jdk>
       </activation>
       <properties>
-        <java.version>11</java.version>
+        <java.version>8</java.version>
```
gpang, line 1307: Does this set the target to 8? Is it supposed to 8, or 1.8, or does it not matter?

yuzhu: oops, fixed

yuzhu:
@gpang PTAL

gpang:
@yuzhu I think this is good. @ZacBlanco can you verify, since you initially added this block?

yuzhu:

PR #11878, opened 2020-08-03
Add an option to docker and build script to allow mvn to run all tests

yuzhu:


dev/jenkins/build.sh
```diff
@@ -20,6 +20,12 @@ then
   git clean -fdx
 fi
 
+local mvn_args=""
+if [ -n "${ALLUXIO_MVN_RUNTOEND}" ]
+then
+  mvn_args+=" -fn --fail-at-end"
```
Xenorith, line 26: no need leading space in string

yuzhu: it is not necessary, but this makes it easier to copy/paste later?

Xenorith: true that, and there's no harm in having the extra space

yuzhu:

PR #11894, opened 2020-08-04
Add java 11 support for cleaning direct buffers

yuzhu:


yuzhu:

yuzhu:

yuzhu:

yuzhu:

yuzhu:
verified to work on java 11

yuzhu:
@apc999 I think you originally wrote this? PTAL

core/common/src/main/java/alluxio/util/io/BufferUtils.java
```diff
@@ -51,6 +53,53 @@ public static int byteToInt(byte b) {
    * this direct buffer should be discarded. This is unsafe operation and currently a work-around to
    * avoid huge memory occupation caused by memory map.
    *
+   * @param buffer bytebuffer
+   */
+  public static synchronized void cleanDirectBuffer(ByteBuffer buffer) {
+    Preconditions.checkNotNull(buffer, "buffer is null");
+    Preconditions.checkArgument(buffer.isDirect(), "buffer isn't a DirectByteBuffer");
+    int javaVersion = Integer.parseInt(System.getProperty("java.version").split("\\D+")[0]);
```
ZacBlanco, line 61: I would suggest making this a static constant as this is going to be constant for the life of the JVM.

Also, I've seen in the JVM that this can report `1.8` for java 8, and `11` for java 11. I'm not sure if that would make a difference

yuzhu: made it a utils method backed by a static constant.

yuzhu: also modifieid it to handle 1.8 or 9-11

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -852,5 +861,18 @@ public static boolean isAddressReachable(String hostname, int port) {
     return result;
   }
 
+  private static int initMajorVersion() {
```
gpang, line 864: Please add a comment on what this will return. I'm assuming `8` for java 1.8, `11` for java 11 ?

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -852,5 +861,18 @@ public static boolean isAddressReachable(String hostname, int port) {
...
+    String version = System.getProperty("java.version");
+    if (version.startsWith("1.")) {
+      version = version.substring(2, 3);
+    } else {
+      int dot = version.indexOf(".");
+      if (dot != -1) {
```
gpang, line 870: Will there always be a dot?

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -852,5 +861,18 @@ public static boolean isAddressReachable(String hostname, int port) {
...
+        version = version.substring(0, dot);
+      }
+    }
+    return Integer.parseInt(version);
```
gpang, line 874: Will this always work?

core/common/src/main/java/alluxio/util/io/BufferUtils.java
```diff
@@ -51,6 +55,53 @@ public static int byteToInt(byte b) {
    * this direct buffer should be discarded. This is unsafe operation and currently a work-around to
    * avoid huge memory occupation caused by memory map.
    *
+   * @param buffer bytebuffer
+   */
+  public static synchronized void cleanDirectBuffer(ByteBuffer buffer) {
```
gpang, line 60: What was the reason for making everything `synchronized`? Is it for initializing the method variable, or is it also for the actual cleaning part?

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -118,6 +120,13 @@ public static long getCurrentMs() {
     return System.currentTimeMillis();
   }
 
+  /**
+   * @return the current jvm major version, such as 1.8 or 11
```
gpang, line 124: Is this comment true? Will it be `1.8` or will it be `8`?

core/common/src/main/java/alluxio/util/CommonUtils.java
yuzhu: yes, the version string looks like `1.8.1` or '11.0.2`
https://www.oracle.com/java/technologies/javase/versioning-naming.html

core/common/src/main/java/alluxio/util/CommonUtils.java
yuzhu: I don't see why it wouldn't?

core/common/src/main/java/alluxio/util/io/BufferUtils.java
yuzhu: it was synchronized before. I think the reason is that we initialize static variable within the method, so it has to be synchronized. cleaning part does not require synchronization.

core/common/src/main/java/alluxio/util/CommonUtils.java
yuzhu: fixed comment

yuzhu:
@gpang addressed comments

core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -852,5 +861,22 @@ public static boolean isAddressReachable(String hostname, int port) {
     return result;
   }
 
+  /**
+   * @return the major version of the current JVM, 8 for 1.8, 11 for java 11
+   * see https://www.oracle.com/java/technologies/javase/versioning-naming.html for reference
+   */
+  private static int initMajorVersion() {
```
ZacBlanco, line 868: can you add one or more unit test(s) for this method

core/common/src/main/java/alluxio/util/io/BufferUtils.java
```diff
@@ -51,6 +55,53 @@ public static int byteToInt(byte b) {
...
+      theUnsafeField.setAccessible(true);
+      Object theUnsafe = theUnsafeField.get(null);
+      sByteBufferCleanerMethod.invoke(theUnsafe, buffer);
+    } catch (Exception e) {
+      LOG.warn("Failed to unmap direct ByteBuffer: {}, error message: {}",
+          buffer.getClass().getName(), e.getMessage());
+    }
+  }
+
+  /**
```
ZacBlanco, line 104: do we have a test for this?

core/common/src/main/java/alluxio/util/CommonUtils.java
yuzhu: emmm not sure how we can test this? like how do we know the ground truth, from within a jvm?

ZacBlanco: test expected JVM inputs (`1.8.X-XXX`, `11.X.X.`, etc?) and make sure this function returns the correct output

core/common/src/main/java/alluxio/util/io/BufferUtils.java
yuzhu: we have a test for the method. java 8 version or java 11 version will be tested depending on the jvm

yuzhu:

PR #11902, opened 2020-08-04
Fix yarn version check in generate-tarball

Xenorith:
version check logic was wrong (ex. if hadoop version = 3.0, comparison of "is greater than 2.4) returned was false) . added correct comparison via helper function

facepalm


ZacBlanco:
Can you add a description to the PR?

ZacBlanco:

PR #11912, opened 2020-08-05
Fix jets3t class not found issue

LuQQiu:
alluxio runUfsTests --path gs://
has 
Caused by: org.jets3t.service.ServiceException: Request Error: Invalid class name: org.jets3t.service.utils.RestUtils$ConnManagerFactory issue.

Using classloader solve the issue


yuzhu:
@LuQQiu can you add a test for this?

LuQQiu:
@yuzhu ufs contract test targets for it. Because it's related to actual code running, unit test doesn't fulfill the requirement.

yuzhu:

PR #11916, opened 2020-08-05
Format only the local master to avoid wiping the whole cluster

ns1123:
This prevents the start command from formatting an entire cluster if the journal directory doesn't exist. Original behavior was built to ensure that the journal folder existed in EMBEDDED mode. However, if a user accidentally removes the directory (e.g. during an upgrade), then we want to ensure that not all nodes are formatted.


ZacBlanco:
good catch

ZacBlanco:
jenkins, test this please

ZacBlanco:

PR #11924, opened 2020-08-06
Fix LoginModuleTest in Java 11

LuQQiu:
JDK 11 doesn't support logout twice


yuzhu:
@LuQQiu do we assume logout can be called multiple times in the code?  we might have that assumption in our code.

yuzhu:

core/common/src/test/java/alluxio/security/login/LoginModuleTest.java
```diff
@@ -62,10 +62,6 @@ public void simpleLogin() throws Exception {
     // logout and verify the user is removed
     loginContext.logout();
     assertTrue(subject.getPrincipals(User.class).isEmpty());
-
-    // logout twice should be no-op.
-    loginContext.logout();
-    assertTrue(subject.getPrincipals(User.class).isEmpty());
```
ZacBlanco, line 68: If this test assumed that logout twice is a no-op, is there any place in the code that uses it which makes the same assumption?

yuzhu: it turns out we don't call logout ever.....

LuQQiu:
Looks like we only call logout in tests

core/common/src/test/java/alluxio/security/login/LoginModuleTest.java
LuQQiu: no ... we don't even use logout

yuzhu: @bf8086  could you take a look at this and confirm that we don't assume logout is idempotent anywhere in our code?

bf8086: It looks like we do it on EE for relogins. It is also likely this is indirectly called from JDK or Hadoop. Maybe we can disable it only for Java 11 to ensure it doesn't regress on Java 8?

yuzhu: @bf8086 is there a test we can write to ensure our relogins on EE will work under java 11? do we need special handling for java 11? if so, there is a function available for it here. https://github.com/Alluxio/alluxio/blob/5ce8371f9a8505f19fa62b1d0edaf285e23f74de/core/common/src/main/java/alluxio/util/CommonUtils.java#L127

bf8086: We do have a relogin test in Autobots. Not sure what is the impact from Java 11. I don't think we should specialize in our production code. If it doesn't work in Java 11 due to the idempotent issue then we should make it not assume that regardless of the java version.

I do still prefer we specialize the test here for java 11 to make sure it doesn't break the promise for other callers(i.e. in case Java 8 itself calls logout internally multiple times).

yuzhu: @LuQQiu do you mind updating the PR to only test calling twice on Java 8?

yuzhu:

PR #11928, opened 2020-08-07
Fix classloader related test failures

yuzhu:


yuzhu:

yuzhu:

yuzhu:

yuzhu:
@bradyoo this reduces the number of errors under java 11 to 8, PTAL

yuzhu:

core/common/src/main/java/alluxio/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,45 @@
...
+public class ClassLoaderUtils {
+  /**
+   * Add a path to be loaded.
+   * @param path full path
+   */
+  public static void addURL(String path) throws Exception {
+    ClassLoader classLoader = ClassLoader.getSystemClassLoader();
+    try {
+      Method method = classLoader.getClass().getDeclaredMethod("addURL", URL.class);
+      method.setAccessible(true);
```
bradyoo, line 34: This is only used in tests right? if so, I would put this in the test and use WhiteBox.

bradyoo: Actually, why does WhiteBox.invokeMethod fail?

core/common/src/main/java/alluxio/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,45 @@
...
+      method.invoke(classLoader, new File(path).toURI().toURL());
+    } catch (NoSuchMethodException e) {
+      Method method = classLoader.getClass().getDeclaredMethod(
+          "appendToClassPathForInstrumentation", String.class);
```
bradyoo, line 38: Oh I see.. I might just call WhiteBox.invokeMethod twice

core/common/src/main/java/alluxio/util/ClassLoaderUtils.java
yuzhu: if our goal is to eventually remove the powermock dependency,  why would we use Whitebox? 
I will move this util class to somewhere in test package.

yuzhu:
alluxio-bot, test this please

bradyoo:
LGTM

yuzhu:

PR #11929, opened 2020-08-07
Fix Wasb UFS to respect block size

apc999:
without this fix, WASB ufs always set block size to 512MB (default value returned by wasb UFS)
the block size for WASB is artificial since it is object store, but always having a constant block size so large could affect Alluxio performance


apc999:

apc999:

PR #11934, opened 2020-08-07
Ignore RelativePath tests in Java 11

bradyoo:
Setting user.dir to test relative path no longer works as of Java 11 (See https://bugs.openjdk.java.net/browse/JDK-8202127)
There is not much value to this test because this ultimately just tests that new File(Relative Path) is a working behavior.

### DIFF
--- a/tests/src/test/java/alluxio/client/cli/fs/command/CopyFromLocalCommandIntegrationTest.java
+++ b/tests/src/test/java/alluxio/client/cli/fs/command/CopyFromLocalCommandIntegrationTest.java
@@ -32,6 +32,7 @@ import alluxio.util.io.BufferUtils;
 
 import org.apache.commons.io.FileUtils;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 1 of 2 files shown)


bradyoo:
@yuzhu

yuzhu:
@bradyoo Perhaps only ignore the test when it is running under java 11?  by using a runtime check? also could you fix the other two related tests as well? Thanks!

bradyoo:
Done

tests/src/test/java/alluxio/client/cli/fs/command/CopyFromLocalCommandIntegrationTest.java
```diff
@@ -306,6 +306,12 @@ public void copyFromLocalWildcard() throws IOException {
 
   @Test
   public void copyFromLocalRelativePath() throws Exception {
+    String version = System.getProperty("java.version");
```
yuzhu, line 309: could we do  `CommonUtil::getJavaVersion() >= 11` instead? so that we can avoid changing this when we go to java 14, whenever that is.

bradyoo: I would like to remove or replace this test when we get there.

yuzhu:
alluxio-bot, test this please

yuzhu:

PR #11940, opened 2020-08-08
Migrate state machine to ratis

bf8086:
Initial changes for hooking up embedded journal state machine with ratis API.  Implemented leader election, journal application and taking snapshot on followers.

### DIFF
--- a/core/server/common/pom.xml
+++ b/core/server/common/pom.xml
@@ -43,6 +43,14 @@
       <artifactId>copycat-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.ratis</groupId>
+      <artifactId>ratis-server</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.ratis</groupId>
+      <artifactId>ratis-grpc</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
     </dependency>
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -63,11 +63,11 @@ import javax.annotation.concurrent.ThreadSafe;
  * A state machine for managing all of Alluxio's journaled state. Entries applied to this state
  * machine will be forwarded to the appropriate internal master.
  *
- * The state machine starts by resetting all state, then applying the entries offered by copycat.
+ * The state machine starts by resetting all state, then applying the entries offered by Ratis.
  * When the master becomes primary, it should wait until the state machine is up to date and no
  * other primary master is serving, then call {@link #upgrade}. Once the state machine is upgraded,
- * it will ignore all entries appended by copycat because those entries are applied to primary
- * master state before being written to copycat.
+ * it will ignore all entries appended by Ratis because those entries are applied to primary
+ * master state before being written to Rati.
  *
  */
 @ThreadSafe
@@ -232,10 +232,7 @@ public class JournalStateMachine extends BaseStateMachine {
   }
 
   /**
-     * Applies a journal entry commit to the state machine.
-     *
-     * This method is automatically discovered by the Copycat framework.
-     *
+   * Applies a journal entry commit to the state machine.
    * @param commit the commit
    */
   public void applyJournalEntryCommand(TransactionContext commit) {
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -19,6 +19,7 @@ import alluxio.master.journal.JournalUtils;
 import alluxio.master.journal.Journaled;
 import alluxio.master.journal.checkpoint.CheckpointInputStream;
 import alluxio.proto.journal.Journal.JournalEntry;
+import alluxio.util.LogUtils;
 import alluxio.util.StreamUtils;
 
 import com.google.common.base.Preconditions;
@@ -47,6 +48,7 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.Collection;
@@ -134,27 +136,25 @@ public class JournalStateMachine extends BaseStateMachine {
     unpause();
   }
 
-  private long loadSnapshot(SingleFileSnapshotInfo snapshot) throws IOException {
+  private void loadSnapshot(SingleFileSnapshotInfo snapshot) throws IOException {
     if (snapshot == null) {
       LOG.info("No snapshot to load");
-      return RaftLog.INVALID_LOG_INDEX;
     }
     LOG.info("Loading Snapshot {}", snapshot);
     final File snapshotFile = snapshot.getFile().getPath().toFile();
     if (!snapshotFile.exists()) {
       LOG.error("The snapshot {} does not exist", snapshot);
-      return RaftLog.INVALID_LOG_INDEX;
+      throw new FileNotFoundException(
+          String.format("The snapshot file %s does not exist", snapshotFile.getPath()));
     }
-
-    try (DataInputStream inputStream = new DataInputStream(new FileInputStream(snapshotFile))) {
+    try {
       resetState();
       setLastAppliedTermIndex(snapshot.getTermIndex());
-      install(inputStream);
+      install(snapshotFile);
     } catch (Exception e) {
       LOG.error("Failed to load snapshot {}", snapshot, e);
       throw e;
     }
-    return snapshot.getTermIndex().getIndex();
   }
 
   @Override
@@ -212,7 +212,7 @@ public class JournalStateMachine extends BaseStateMachine {
         suspend();
       }
     } catch (IOException e) {
-      LOG.warn("State machine pause failed", e);
+      LOG.error("State machine pause failed", e);
       throw new IllegalStateException(e);
     }
     getLifeCycle().transition(LifeCycle.State.PAUSED);
@@ -247,7 +247,7 @@ public class JournalStateMachine extends BaseStateMachine {
           commit.getStateMachineLogEntry().getLogData().asReadOnlyByteBuffer());
     } catch (Exception e) {
       ProcessUtils.fatalError(LOG, e,
-          "Encountered invalid journal entry in 'commit': %s.", commit);
+          "Encountered invalid journal entry in commit: %s.", commit);
       System.exit(-1);
       throw new IllegalStateException(e); // We should never reach here.
     }
@@ -345,7 +345,7 @@ public class JournalStateMachine extends BaseStateMachine {
     try {
       tempFile = createTempSnapshotFile();
     } catch (IOException e) {
-      LOG.warn("Failed to create temp snapshot file", e);
+      LogUtils.warnWithException(LOG, "Failed to create temp snapshot file", e);
       return RaftLog.INVALID_LOG_INDEX;
     }
     LOG.info("Taking a snapshot to file {}", tempFile);
@@ -355,7 +355,7 @@ public class JournalStateMachine extends BaseStateMachine {
       JournalUtils.writeToCheckpoint(outputStream, getStateMachines());
     } catch (Exception e) {
       tempFile.delete();
-      LOG.warn("Failed to take snapshot: {}", snapshotId, e);
+      LogUtils.warnWithException(LOG, "Failed to take snapshot: {}", snapshotId, e);
       return RaftLog.INVALID_LOG_INDEX;
     }
     try {
@@ -372,21 +372,21 @@ public class JournalStateMachine extends BaseStateMachine {
           System.currentTimeMillis() - mLastSnapshotStartTime);
     } catch (Exception e) {
       tempFile.delete();
-      LOG.warn("Failed to take snapshot: {}", snapshotId, e);
+      LogUtils.warnWithException(LOG, "Failed to take snapshot: {}", snapshotId, e);
       return RaftLog.INVALID_LOG_INDEX;
     }
     try {
       mStorage.loadLatestSnapshot();
     } catch (Exception e) {
       snapshotFile.delete();
-      LOG.warn("Failed to refresh latest snapshot: {}", snapshotId, e);
+      LogUtils.warnWithException(LOG, "Failed to refresh latest snapshot: {}", snapshotId, e);
       return RaftLog.INVALID_LOG_INDEX;
     }
     mSnapshotting = false;
     return last.getIndex();
   }
 
-  private void install(DataInputStream stream) {
+  private void install(File snapshotFile) {
     if (mClosed) {
       return;
     }
@@ -396,7 +396,7 @@ public class JournalStateMachine extends BaseStateMachine {
     }
 
     long snapshotId = 0L;
-    try {
+    try (DataInputStream stream =  new DataInputStream(new FileInputStream(snapshotFile))) {
       snapshotId = stream.readLong();
       JournalUtils.restoreFromCheckpoint(new CheckpointInputStream(stream), getStateMachines());
     } catch (Exception e) {
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -86,12 +86,14 @@ public class JournalStateMachine extends BaseStateMachine {
   private volatile long mLastPrimaryStartSequenceNumber = 0;
   private volatile long mNextSequenceNumberToRead = 0;
   private volatile boolean mSnapshotting = false;
+  private volatile boolean mIsLeader = false;
+
   // The start time of the most recent snapshot
   private volatile long mLastSnapshotStartTime = 0;
   /** Used to control applying to masters. */
   private BufferedJournalApplier mJournalApplier;
   private final SimpleStateMachineStorage mStorage = new SimpleStateMachineStorage();
-  private Object mRaftGroupId;
+  private RaftGroupId mRaftGroupId;
 
   /**
    * @param journals     master journals; these journals are still owned by the caller, not by the
@@ -157,7 +159,7 @@ public class JournalStateMachine extends BaseStateMachine {
 
   @Override
   public long takeSnapshot() {
-    if (mJournalSystem.isLeader()) {
+    if (mIsLeader) {
       // TODO(feng): if secondary master has a more recent snapshot, install it
       return RaftLog.INVALID_LOG_INDEX;
     } else {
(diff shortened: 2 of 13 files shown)


bf8086:
@ggezer FYI.

bf8086:
Jenkins, test this please.

bf8086:
@ggezer @gpang PTAL.

core/server/common/pom.xml
```diff
@@ -42,6 +42,14 @@
       <groupId>io.atomix.copycat.alluxio</groupId>
       <artifactId>copycat-client</artifactId>
```
gpang, line 43: When can we get rid of these copycat dependencies?

tests/src/test/java/alluxio/server/ft/journal/TriggeredCheckpointTest.java
```diff
@@ -121,7 +121,7 @@ private void validateCheckpointInClusterRestart(MultiProcessCluster cluster)
       throws Exception {
     cluster.stopMasters();
     cluster.startMasters();
-    cluster.waitForAllNodesRegistered(20 * Constants.SECOND_MS);
+    cluster.waitForAllNodesRegistered(30 * Constants.SECOND_MS);
```
gpang, line 124: why does it take so long to register nodes?

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
```diff
@@ -242,9 +242,10 @@ public void gainPrimacyAfterSuspend() throws Exception {
     Assert.assertEquals(0, countingMaster.getApplyCount());
     // Gain primacy in follower journal and validate it catches up.
     mFollowerJournalSystem.gainPrimacy();
-    CommonUtils.waitFor("full state acquired after resume",
-        () -> countingMaster.getApplyCount() == entryCount, mWaitOptions);
-
+    CommonUtils.waitFor(
+        "full state acquired after resume", () -> mFollowerJournalSystem.getCurrentSequenceNumbers()
+            .values().stream().distinct().collect(Collectors.toList()).get(0) == entryCount - 1,
```
gpang, line 247: old calculation was wrong?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+      super.initialize(server, groupId, raftStorage);
+      mRaftGroupId = groupId;
+      mStorage.init(raftStorage);
+      loadSnapshot(mStorage.getLatestSnapshot());
+    });
+  }
+
+  @Override
+  public void reinitialize() throws IOException {
+    LOG.info("Reinitializing state machine.");
```
gpang, line 133: Is there any identifier we can add to this message?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+    getLifeCycle().startAndTransition(() -> {
+      super.initialize(server, groupId, raftStorage);
+      mRaftGroupId = groupId;
+      mStorage.init(raftStorage);
+      loadSnapshot(mStorage.getLatestSnapshot());
+    });
+  }
+
+  @Override
+  public void reinitialize() throws IOException {
```
gpang, line 132: I'm curious. Why does `reinitialize()` and `initialize()` look completely different? I would have imagined that reinitialize would be quite similar to initialize, or would share some code?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+    LOG.info("Reinitializing state machine.");
+    mStorage.loadLatestSnapshot();
+    loadSnapshot(mStorage.getLatestSnapshot());
+    unpause();
+  }
+
+  private void loadSnapshot(SingleFileSnapshotInfo snapshot) throws IOException {
+    if (snapshot == null) {
+      LOG.info("No snapshot to load");
```
gpang, line 141: Should be ERROR. Also, when does this happen? I don't know what this message is supposed to tell the user.

Why is this not throwing an exception instead?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+      return;
+    }
+    LOG.info("Loading Snapshot {}", snapshot);
+    final File snapshotFile = snapshot.getFile().getPath().toFile();
+    if (!snapshotFile.exists()) {
+      LOG.error("The snapshot {} does not exist", snapshot);
```
gpang, line 147: Do we need to log and throw the same exception?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+      throw new FileNotFoundException(
+          String.format("The snapshot file %s does not exist", snapshotFile.getPath()));
+    }
+    try {
+      resetState();
+      setLastAppliedTermIndex(snapshot.getTermIndex());
+      install(snapshotFile);
+    } catch (Exception e) {
+      LOG.error("Failed to load snapshot {}", snapshot, e);
```
gpang, line 156: I think if you use the logging with `{}`, you cannot use the last argument as the exception. You have to use `String.format()` to pass in a single string, and then the exception.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+      throw e;
```
gpang, line 157: Should we throw a new exception with the message like above? Also, do we need to log and throw an exception?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,164 @@
...
+  @Override
+  public void pause() {
+    getLifeCycle().transition(LifeCycle.State.PAUSING);
+    try {
+      if (!mJournalApplier.isSuspended()) {
+        suspend();
+      }
+    } catch (IOException e) {
+      LOG.error("State machine pause failed", e);
+      throw new IllegalStateException(e);
```
gpang, line 217: Do we need to throw and log? Can we have this message include the message above?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -171,31 +317,77 @@ private void applySingleEntry(JournalEntry entry) {
...
+      return RaftLog.INVALID_LOG_INDEX;
+    }
+    LOG.info("Taking a snapshot to file {}", tempFile);
+    final File snapshotFile = mStorage.getSnapshotFile(last.getTerm(), last.getIndex());
+    try (DataOutputStream outputStream = new DataOutputStream(new FileOutputStream(tempFile))) {
+      outputStream.writeLong(snapshotId);
+      JournalUtils.writeToCheckpoint(outputStream, getStateMachines());
+    } catch (Exception e) {
+      tempFile.delete();
+      LogUtils.warnWithException(LOG, "Failed to take snapshot: {}", snapshotId, e);
```
gpang, line 359: Can you include the file name?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -171,31 +317,77 @@ private void applySingleEntry(JournalEntry entry) {
...
+        LOG.warn("Failed to rename snapshot from {} to {}", tempFile, snapshotFile);
+        return RaftLog.INVALID_LOG_INDEX;
+      }
+      LOG.info("Completed snapshot up to SN {} in {}ms", snapshotId,
+          System.currentTimeMillis() - mLastSnapshotStartTime);
     } catch (Exception e) {
-      ProcessUtils.fatalError(LOG, e, "Failed to take snapshot: %s", snapshotId);
-      throw new RuntimeException(e);
+      tempFile.delete();
+      LogUtils.warnWithException(LOG, "Failed to take snapshot: {}", snapshotId, e);
```
gpang, line 376: Include filename? Also, this message is identical to the previous section. Can we distinguish them?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -314,10 +506,11 @@ public boolean isSnapshotting() {
...
   }
 
-  /**
-   * Closes the journal state machine, causing all further modification requests to be ignored.
-   */
-  public void close() {
-    mClosed = true;
+  @Override
+  public void notifyLeaderChanged(RaftGroupMemberId groupMemberId, RaftPeerId raftPeerId) {
+    if (mRaftGroupId == groupMemberId.getGroupId()) {
```
gpang, line 511: What about if this is not true? What would that happen, and do we want to log some sort of warning when that happens?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftPrimarySelector.java
```diff
@@ -29,33 +23,13 @@
  */
 @ThreadSafe
 public class RaftPrimarySelector extends AbstractPrimarySelector {
```
gpang, line 25: Can you update some of the comments in this file that still say copycat?

core/server/master/src/main/java/alluxio/master/journal/tool/RaftJournalDumper.java
```diff
@@ -81,74 +75,69 @@ void dumpJournal() throws Throwable {
...
+
+      try (DataInputStream inputStream = new DataInputStream(new FileInputStream(snapshotFile))) {
+        LOG.debug("Reading snapshot-Id: {}", inputStream.readLong());
+        try (CheckpointInputStream checkpointStream = new CheckpointInputStream(inputStream)) {
+          readCheckpoint(checkpointStream, Paths.get(checkpointPath));
+        } catch (Exception e) {
+          LOG.error("Failed to read snapshot from journal.", e);
+        }
+      } catch (Exception e) {
+        LOG.error("Failed to load snapshot {}", snapshotFile, e);
```
gpang, line 138: If you want to use the exception as the last argument, you will have to use `String.format()` to print parameters.

core/server/common/pom.xml
bf8086: Once we removes all copycat usages in transport and backup restore functions then we can remove it.

tests/src/test/java/alluxio/server/ft/journal/TriggeredCheckpointTest.java
bf8086: I am not exactly sure. There might be some parameters that we can tune to make it faster.

core/server/master/src/main/java/alluxio/master/journal/tool/RaftJournalDumper.java
bf8086: IIRC in slf4j we can add one throwable in addition to the string parameters to print full exception stack: http://www.slf4j.org/faq.html#paramException

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
bf8086: Gaining primacy in this test does not ensure all journal entries are applied. Sometimes a few entries are processed after the journal system is promoted and thus get ignored. The sequence number reflects entries that are processed which will include the entries processed before and after the follower becomes primary.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: initialize() is invoked when the server is first started. reinitialize() is invoked when a follower got a new snapshot from leader and need to reload. 

There isn't much difference except maybe two things:
- the `initialize()` does a little more to initialize the group id and raft storage.
- `mStorage.loadLatestSnapshot()` was implicitly called in `initialize()` as part of `mStorage.init(raftStorage);`

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: A `null` snapshot is given if there isn't a snapshot in the journal folder. This can happen when the master is started for the first time or before a snapshot is taken. It is an expected case therefore is not logged as an error.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Can you explain a little bit what you want to achieve here? Are you looking to create the exception, put it in the log and then throw it?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: AFAIK there is only one state machine per master. Is there a specific identifier you are looking for?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: For slf4j we can pass an exception in addition to all string parameters to include the call stack in the log.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftPrimarySelector.java
bf8086: Updated.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Yeah this should always be true given we only have one raft group. I will add a warning.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Updated.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Added.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Removed the logging message.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: I initially added the log for debugging. Removed.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Removed the logging message and included it in a new exception.

bf8086:
@gpang Thanks for the review. I updated the PR to address the comments. PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
gpang: Ok. Is there a local path where the files will be initialized?

core/server/master/src/main/java/alluxio/master/journal/tool/RaftJournalDumper.java
gpang: hrmmm, that is strange. I remember our logging messages would ignore the throwable, if we were using the string parameters...

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -333,7 +375,7 @@ public synchronized void gainPrimacy() {
 
   @Override
   public synchronized void losePrimacy() {
-    if (!mServer.isRunning()) {
+    if (mServer.getLifeCycleState() != LifeCycle.State.RUNNING) {
       // Avoid duplicate shut down copycat server
```
gpang, line 379: replace copycat in the comment

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: The local path where the journal logs and snapshots are stored is the journal directory set through the raft server configuration.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -124,19 +139,18 @@
  */
 @ThreadSafe
 public final class RaftJournalSystem extends AbstractJournalSystem {
+  public static final UUID RAFT_GROUP_ID = UUID.fromString("02511d47-d67c-49a3-9011-abb3109a44c1");
```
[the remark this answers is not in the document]
bf8086: The raft group id associate the single state machine we have for master. We need it to be a constant for masters to join the right cluster when they start.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -159,7 +173,7 @@
   /**
    * Copycat server.
```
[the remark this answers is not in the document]
bf8086: Updated.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Updated.

bf8086:
@gpang @LuQQiu Thanks for the review. PTAL the updated PR.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -70,35 +88,158 @@
...
+  }
+
+  @Override
+  public void initialize(RaftServer server, RaftGroupId groupId,
+      RaftStorage raftStorage) throws IOException {
+    getLifeCycle().startAndTransition(() -> {
+      super.initialize(server, groupId, raftStorage);
+      mRaftGroupId = groupId;
+      mStorage.init(raftStorage);
+      loadSnapshot(mStorage.getLatestSnapshot());
```
LuQQiu, line 127: I saw the examples of loading snapshots in initialize and reinitialize, but not sure why.
Are there guaranteed that all journal entries are included in the latest snapshot? will some journal entries after latest snapshot lost?

bf8086: The snapshot does not have to include all entries. Ratis will catch up on entries after the snapshot before the server starts serving.

bf8086:

PR #11953, opened 2020-08-09
Use a compatibility layer to handle older hive metastore

yuzhu:
Version 2.2.0 of hive metastore client worked with a wide range of server versions, but we have to upgrade it for java 11 support. Hence we use this compatibility library which wraps the HMS client and replaces certain calls to be compatible with older versions.

https://github.com/HotelsDotCom/hcommon-hive-metastore

### DIFF
--- a/table/server/underdb/hive/pom.xml
+++ b/table/server/underdb/hive/pom.xml
@@ -26,8 +26,8 @@
     <!-- The following paths need to be defined here as well as in the parent pom so that mvn can -->
     <!-- run properly from sub-project directories -->
     <build.path>${project.parent.parent.parent.parent.basedir}/build</build.path>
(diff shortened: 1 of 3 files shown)


table/server/underdb/hive/pom.xml
```diff
@@ -37,6 +37,11 @@
       <artifactId>guava</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>com.hotels</groupId>
+      <artifactId>hcommon-hive-metastore</artifactId>
+      <version>1.4.2</version>
+    </dependency>
```
ZacBlanco, line 44: I think this dependency is of questionable quality...

yuzhu: what is our standard for including something as a dependency? @gpang

yuzhu:
closing this in favor of #11962

PR #11956, opened 2020-08-10
Remove guice dependency

yuzhu:


yuzhu:

shaded/hadoop/pom.xml
```diff
@@ -142,9 +142,6 @@
                 <relocation>
                   <pattern>com.google</pattern>
                   <shadedPattern>${shading.prefix}.com.google</shadedPattern>
-                  <excludes>
```
gpang, line 145: Does this mean that all of these packages will be shaded now? Why were they not shaded before?

yuzhu: actually guice has not been included in hdfs package for a long time, so it did not have an effect. 
not sure why it was shaded before, @LuQQiu do you know?

yuzhu: discussed offline, not sure why it was excluded before. but i think we should shade it.

yuzhu:
@gpang PTAL, updated comments on shading exclusion

yuzhu:

PR #11962, opened 2020-08-11
Create a compatibility layer to handle older hive metastores

yuzhu:
Version 2.2.0 of hive metastore client worked with a wide range of server versions, but we have to upgrade it for java 11 support. Hence we add a compatibility library which proxies a number of calls to achieve compatibility with older versions.

### DIFF
--- a/table/server/underdb/hive/pom.xml
+++ b/table/server/underdb/hive/pom.xml
@@ -26,8 +26,8 @@
     <!-- The following paths need to be defined here as well as in the parent pom so that mvn can -->
     <!-- run properly from sub-project directories -->
     <build.path>${project.parent.parent.parent.parent.basedir}/build</build.path>
-    <!-- 2.2.0 client works with a wide range of HMS, from 1.0 to 3.1 -->
-    <hive-metastore.version>2.2.0</hive-metastore.version>
+    <!-- 2.3.7 client needs a compatibility layer to work with lower versions -->
+    <hive-metastore.version>2.3.7</hive-metastore.version>
   </properties>
 
   <dependencies>
--- a/table/server/underdb/hive/pom.xml
+++ b/table/server/underdb/hive/pom.xml
@@ -26,7 +26,7 @@
     <!-- The following paths need to be defined here as well as in the parent pom so that mvn can -->
     <!-- run properly from sub-project directories -->
     <build.path>${project.parent.parent.parent.parent.basedir}/build</build.path>
-    <!-- 2.3.7 client needs a compatibility layer to work with lower versions -->
+    <!-- 2.3.7 client needs a compatibility layer to work with versions <= 2.2.x -->
     <hive-metastore.version>2.3.7</hive-metastore.version>
   </properties>
 
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
@@ -42,7 +42,8 @@ class CompatibleMetastoreClient implements InvocationHandler {
         if (mCompat != null
             && delegateException.getCause().getClass()
             .isAssignableFrom(TApplicationException.class)) {
-          LOG.debug("Attempting to call hive metastore with compatibility client");
+          LOG.debug("Attempting to call hive metastore with compatibility class {}",
+              mCompat.getClass().getName());
           return invokeCompatibility(method, args);
         }
       } catch (InvocationTargetException compatibilityException) {
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
@@ -1,3 +1,14 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
 package alluxio.table.under.hive.util;
 
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
@@ -20,13 +20,15 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
+import javax.annotation.Nullable;
+
 class CompatibleMetastoreClient implements InvocationHandler {
   private static final Logger LOG = LoggerFactory.getLogger(CompatibleMetastoreClient.class);
 
   private final IMetaStoreClient mDelegate;
   private final HMSShim mCompat;
 
-  CompatibleMetastoreClient(IMetaStoreClient delegate, HMSShim compatibility) {
+  CompatibleMetastoreClient(IMetaStoreClient delegate, @Nullable HMSShim compatibility) {
     mDelegate = delegate;
     mCompat = compatibility;
   }
@@ -37,11 +39,10 @@ class CompatibleMetastoreClient implements InvocationHandler {
       return method.invoke(mDelegate, args);
     } catch (InvocationTargetException delegateException) {
       try {
-        LOG.info("Couldn't invoke method {}", method.toGenericString());
         if (mCompat != null
             && delegateException.getCause().getClass()
             .isAssignableFrom(TApplicationException.class)) {
-          LOG.info("Attempting to invoke with {}", mCompat.getClass().getName());
+          LOG.debug("Attempting to call hive metastore with {}", mCompat.getClass().getName());
           return invokeCompatibility(method, args);
         }
       } catch (InvocationTargetException compatibilityException) {
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
@@ -42,7 +42,7 @@ class CompatibleMetastoreClient implements InvocationHandler {
         if (mCompat != null
             && delegateException.getCause().getClass()
             .isAssignableFrom(TApplicationException.class)) {
-          LOG.debug("Attempting to call hive metastore with {}", mCompat.getClass().getName());
+          LOG.debug("Attempting to call hive metastore with compatibility client");
           return invokeCompatibility(method, args);
         }
       } catch (InvocationTargetException compatibilityException) {
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
@@ -1,0 +1,68 @@
+package alluxio.table.under.hive.util;
+
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.thrift.TApplicationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+class CompatibleMetastoreClient implements InvocationHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(CompatibleMetastoreClient.class);
+
+  private final IMetaStoreClient mDelegate;
+  private final HMSShim mCompat;
+
+  CompatibleMetastoreClient(IMetaStoreClient delegate, HMSShim compatibility) {
+    mDelegate = delegate;
+    mCompat = compatibility;
+  }
+
+  @Override
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    try {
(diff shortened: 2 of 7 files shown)


table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
```diff
@@ -0,0 +1,79 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+class CompatibleMetastoreClient implements InvocationHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(CompatibleMetastoreClient.class);
+
+  private final IMetaStoreClient mDelegate;
+  private final HMSShim mCompat;
+
+  CompatibleMetastoreClient(IMetaStoreClient delegate, HMSShim compatibility) {
```
gpang, line 29: The second parameter should be `@Nullable`

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
```diff
@@ -0,0 +1,79 @@
...
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    try {
+      return method.invoke(mDelegate, args);
+    } catch (InvocationTargetException delegateException) {
+      try {
+        LOG.info("Couldn't invoke method {}", method.toGenericString());
+        if (mCompat != null
+            && delegateException.getCause().getClass()
+            .isAssignableFrom(TApplicationException.class)) {
+          LOG.info("Attempting to invoke with {}", mCompat.getClass().getName());
```
gpang, line 44: These info logs could be spammy, since it would log for each invocation of those calls. Should we just cache this info and proactively use the compat method since that decision won't change over time. Then, we can log it when cache the method decision.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSClientFactory.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+    return newInstance(delegate, compatibility);
+  }
+
+  static IMetaStoreClient newInstance(IMetaStoreClient delegate, HMSShim compatibility) {
+    ClassLoader classLoader = IMetaStoreClient.class.getClassLoader();
+    Class<?>[] interfaces = new Class<?>[] { IMetaStoreClient.class };
+    CompatibleMetastoreClient handler = new CompatibleMetastoreClient(delegate,
+        compatibility);
+    return (IMetaStoreClient) Proxy.newProxyInstance(classLoader, interfaces, handler);
```
gpang, line 38: Everything will go through Proxy and reflection now? That will have an overhead for any HMS interaction?

yuzhu: I am not worried about the performance overhead at all. these calls are going to be dominated by network cost. 
according to http://ordinaryjava.blogspot.com/2008/08/benchmarking-cost-of-dynamic-proxies.html, dynamic proxies are about 1.63x of a normal method call, which is almost instant.

table/server/underdb/hive/pom.xml
```diff
@@ -26,8 +26,8 @@
     <!-- The following paths need to be defined here as well as in the parent pom so that mvn can -->
     <!-- run properly from sub-project directories -->
     <build.path>${project.parent.parent.parent.parent.basedir}/build</build.path>
-    <!-- 2.2.0 client works with a wide range of HMS, from 1.0 to 3.1 -->
-    <hive-metastore.version>2.2.0</hive-metastore.version>
+    <!-- 2.3.7 client needs a compatibility layer to work with lower versions -->
```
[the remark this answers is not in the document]
yuzhu: done

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSClientFactory.java
```diff
@@ -0,0 +1,40 @@
...
+    HMSShim compatibility = null;
+    try {
+      compatibility = new HMSShim(delegate);
+    } catch (Throwable t) {
+      LOG.warn("Unable to initialize hive metastore compatibility client", t);
+    }
+    return newInstance(delegate, compatibility);
+  }
+
+  static IMetaStoreClient newInstance(IMetaStoreClient delegate, HMSShim compatibility) {
```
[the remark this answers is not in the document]
yuzhu: done

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSClientFactory.java
```diff
@@ -0,0 +1,40 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(HMSClientFactory.class);
+
+  static IMetaStoreClient newInstance(IMetaStoreClient delegate) {
+    HMSShim compatibility = null;
+    try {
+      compatibility = new HMSShim(delegate);
+    } catch (Throwable t) {
+      LOG.warn("Unable to initialize hive metastore compatibility client", t);
+    }
+    return newInstance(delegate, compatibility);
```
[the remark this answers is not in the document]
yuzhu: it is ok, because we check for null in the call.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,171 @@
...
+public class HMSShim implements HiveCompatibility {
+  private final TServiceClient mClient;
+
+  /**
+   * Constructor for HMSShim.
+   *
+   * @param client another client as delegate
+   */
+  public HMSShim(IMetaStoreClient client) {
+    while (Proxy.isProxyClass(client.getClass())) {
```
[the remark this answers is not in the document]
yuzhu: added a comment

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,171 @@
...
+        result = (T) field.get(object);
+      } else {
+        field.setAccessible(true);
+        result = (T) field.get(object);
+        field.setAccessible(false);
+      }
+      return result;
+    } catch (SecurityException | NoSuchFieldException
+        | IllegalArgumentException | IllegalAccessException e) {
+      throw new RuntimeException("Unable to hack client", e);
```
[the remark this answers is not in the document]
yuzhu: done

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,171 @@
...
+   */
+  public HMSShim(IMetaStoreClient client) {
+    while (Proxy.isProxyClass(client.getClass())) {
+      InvocationHandler handler = Proxy.getInvocationHandler(client);
+      if (handler.getClass().isAssignableFrom(RetryingMetaStoreClient.class)) {
+        client = getField(handler, "base");
+        continue;
+      }
+      // Other handlers can be added here
+      throw new RuntimeException("Unknown InvocationHandler " + handler.getClass());
```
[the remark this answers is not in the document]
yuzhu: done

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
yuzhu: done

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
yuzhu: reduced log level

yuzhu:
@gpang PTAL. Tested this on metastore 2.2 and 2.3, with java 11 as well.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,172 @@
...
+    while (Proxy.isProxyClass(client.getClass())) {
+      InvocationHandler handler = Proxy.getInvocationHandler(client);
+      if (handler.getClass().isAssignableFrom(RetryingMetaStoreClient.class)) {
+        client = getField(handler, "base");
+        continue;
+      }
+      // Other handlers can be added here
+      throw new RuntimeException("Unknown proxy handler for IMetaStoreClient");
+    }
+    mClient = getField(client, "client");
```
gpang, line 53: Is this safe? We are grabbing a field by name, of the hdfs client?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,172 @@
...
+        result = (T) field.get(object);
+      } else {
+        field.setAccessible(true);
+        result = (T) field.get(object);
+        field.setAccessible(false);
+      }
+      return result;
+    } catch (SecurityException | NoSuchFieldException
+        | IllegalArgumentException | IllegalAccessException e) {
+      throw new RuntimeException("Unable to access the client through reflection", e);
```
gpang, line 70: add the `fieldName` to the message

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,172 @@
...
+
+  private static Table deepCopy(Table table) {
+    return table.deepCopy();
+  }
+
+  private void sendBase(String methodName, TBase<?, ?> args) throws TException {
+    try {
+      Method sendBase = TServiceClient.class.getDeclaredMethod(
+          "sendBase", String.class, TBase.class);
+      sendBase.setAccessible(true);
```
gpang, line 82: Why do we have to set and then unset each time?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  private static Table deepCopy(Table table) {
+    return table.deepCopy();
+  }
+
+  private void sendBase(String methodName, TBase<?, ?> args) throws TException {
+    try {
+      Method sendBase = TServiceClient.class.getDeclaredMethod(
```
gpang, line 80: Should this be saved/cached?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    try {
+      return method.invoke(mDelegate, args);
+    } catch (InvocationTargetException delegateException) {
+      try {
+        if (mCompat != null
+            && delegateException.getCause().getClass()
+            .isAssignableFrom(TApplicationException.class)) {
+          LOG.debug("Attempting to call hive metastore with {}", mCompat.getClass().getName());
```
gpang, line 45: Does the name of the method already imply the compatibility shim? If not, can we be explicit in the message that we are using the compatibility shim?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
yuzhu: If this constructor fails, it will be caught at the higher level and we would behave as if there is no compatibility layer. 
This reflection allows us to access the  thrift client and issue hive 2.2 compatible requests.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
yuzhu: you are right, we don't need to unset

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
yuzhu: done

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
yuzhu: FYI: other places where reflection is used, there is the risk of failed reflection as well. In those cases, we catch throwable and fall back to the default behavior of the base client.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HMSShim.java
yuzhu: i don't think there is a lot of value in caching these objects, mainly because the cost of reflection is going to be minimal compared to the network cost of these thrift calls.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
yuzhu: done

yuzhu:
@gpang PTAL, comments addressed

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/CompatibleMetastoreClient.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    try {
+      return method.invoke(mDelegate, args);
+    } catch (InvocationTargetException delegateException) {
+      try {
+        if (mCompat != null
+            && delegateException.getCause().getClass()
+            .isAssignableFrom(TApplicationException.class)) {
+          LOG.debug("Attempting to call hive metastore with compatibility client");
```
gpang, line 45: Thanks for adding the compatibility part, but I think the method name is also useful. Could you add back the method name?

yuzhu: done

yuzhu:

PR #11966, opened 2020-08-11
Remove JvmPauseMonitorTest#interruptOnStop test

ZacBlanco:
Mockito was recently upgraded which includes a new feature that allows us to mock final methods and classes. I rewrote the test to be simpler using one of Mockito's Spy classes.

### DIFF
--- a/core/common/src/test/java/alluxio/util/JvmPauseMonitorTest.java
+++ b/core/common/src/test/java/alluxio/util/JvmPauseMonitorTest.java
@@ -13,9 +13,10 @@ package alluxio.util;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-import static org.mockito.Matchers.any;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doAnswer;
 
 import alluxio.TestLoggerRule;
@@ -26,9 +27,10 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.mockito.Mockito;
 import org.mockito.stubbing.Answer;
+import org.powermock.reflect.Whitebox;
 
-import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CyclicBarrier;
+import java.util.function.Supplier;
 
 public final class JvmPauseMonitorTest {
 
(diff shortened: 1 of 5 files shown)


core/common/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
ZacBlanco, line 1: In reference to why this is needed: https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods

core/common/src/test/java/alluxio/util/JvmPauseMonitorTest.java
```diff
@@ -49,28 +49,16 @@ public void pauseMonitorStartStopTest() {
...
-        fail("Failed to await on barrier");
-      }
-    });
-    t1.start();
+    JvmPauseMonitor mon = Mockito.spy(new JvmPauseMonitor(1000, 900000, 90000));
     mon.start();
-    barrier.await();
+    Thread pmThread = Whitebox.getInternalState(mon, "mJvmMonitorThread");
+    Thread spied = Mockito.spy(pmThread);
+    doThrow(InterruptedException.class).when(spied).join();
```
ZacBlanco, line 56: Thread#join is final

core/common/src/test/java/alluxio/util/JvmPauseMonitorTest.java
```diff
@@ -14,9 +14,9 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-import static org.mockito.Matchers.any;
+import static org.mockito.ArgumentMatchers.any;
```
ZacBlanco, line 17: org.mockito.Matchers is now deprecated, so I moved to the newer class

core/common/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
yuzhu: does this affect other tests? i saw several other tests failed. Could you take a look?

ZacBlanco: Took a look. Latest commits should fix. I make use of the new Mockito#mockStatic methods in 3.4.x and above

ZacBlanco: ok..well.. more issues arose. Latest commit fixes the failures of the last build

yuzhu:
@ZacBlanco so the final decision is to remove it?

ZacBlanco:
Yeah. I couldn't find a reliable way to test this without using Mockito's inline mock make, but that breaks all of our powermock tests.

ZacBlanco:

PR #11972, opened 2020-08-12
Improve exception messages for FileAlreadyExistsException

gpang:
FileAlreadyExistsException message is not always obvious what it means, so adjusted the messages to be more specific to the situation.


gpang:
Jenkins, test this please

core/client/hdfs/src/test/java/alluxio/hadoop/AbstractFileSystemTest.java
```diff
@@ -585,6 +587,46 @@ public void getBlockLocationsNoUfsLocationsWithFallback() throws Exception {
...
+    Path path = new Path("/file");
+    alluxio.client.file.FileSystem alluxioFs =
+        mock(alluxio.client.file.FileSystem.class);
+    when(alluxioFs.exists(new AlluxioURI(HadoopUtils.getPathWithoutScheme(path))))
+        .thenReturn(true);
+    when(alluxioFs.createFile(eq(new AlluxioURI(HadoopUtils.getPathWithoutScheme(path))), any()))
+        .thenThrow(new FileAlreadyExistsException(path.toString()));
+    FileSystem alluxioHadoopFs = new FileSystem(alluxioFs);
+
+    try {
```
apc999, line 620: Shall we use junit `expectedMessage`?
```
expectedEx.expectMessage("message to expect");
```
see
https://stackoverflow.com/questions/2469911/how-do-i-assert-my-exception-message-with-junit-test-annotation

core/client/hdfs/src/test/java/alluxio/hadoop/AbstractFileSystemTest.java
```diff
@@ -585,6 +587,46 @@ public void getBlockLocationsNoUfsLocationsWithFallback() throws Exception {
...
+  @Test
+  public void createWithoutOverwrite() throws Exception {
+    Path path = new Path("/file");
+    alluxio.client.file.FileSystem alluxioFs =
+        mock(alluxio.client.file.FileSystem.class);
+    when(alluxioFs.exists(new AlluxioURI(HadoopUtils.getPathWithoutScheme(path))))
+        .thenReturn(true);
+    when(alluxioFs.createFile(eq(new AlluxioURI(HadoopUtils.getPathWithoutScheme(path))), any()))
+        .thenThrow(new FileAlreadyExistsException(path.toString()));
+    FileSystem alluxioHadoopFs = new FileSystem(alluxioFs);
```
apc999, line 618: I will make this `try-with-resource` to make sure it is closed

apc999: same comments apply to other parts

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -675,16 +676,22 @@ public InodeDirectory getRoot() {
...
-        return new ArrayList<>();
+        return Collections.emptyList();
       } else {
-        throw new FileAlreadyExistsException(path);
+        String pathType = "file";
+        if (context instanceof CreateDirectoryContext) {
+          pathType = "directory";
+        }
+        String errorMessage =
+            String.format("Not allowed to create existing %s: %s", pathType, path);
```
apc999, line 693: not clear to me. do we mean:
Not allowed to create a new file with existing path?

core/client/hdfs/src/test/java/alluxio/hadoop/AbstractFileSystemTest.java
gpang: Done

core/client/hdfs/src/test/java/alluxio/hadoop/AbstractFileSystemTest.java
gpang: In general, I have been favoring the manual catching, mainly because the `ExpectedException` only works with 1 exception per test. It is common that tests might expect multiple exceptions in a test, so I tend to manually catch the Exception. In these specific tests, we only deal with 1 Exception, so `ExpectedException` would have worked, but I usually avoid using it because of that limitation.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
gpang: I updated it. Does the message make more sense?

gpang:
Jenkins, test this please

gpang:

gpang:

apc999:

PR #11974, opened 2020-08-12
Update docker image jdk11 to include jdk8 as well

yuzhu:


dev/jenkins/Dockerfile-jdk11
```diff
@@ -13,6 +13,79 @@
 
 FROM maven:3.6.3-jdk-11
 
+ENV JAVA_HOME /usr/local/openjdk-8
```
Xenorith, line 16: add comment with link to where you got this from 
probably also good idea to comment that this overrides the default java version to 8 and the reason why we need 8 in the 11 image

yuzhu:

PR #11980, opened 2020-08-13
Remove unnecessary log4j dependency

### LINKED ISSUE #11979: Multiple SLF4J bindings on Classpath
**Alluxio Version:**
What version of Alluxio are you using?
2.4.0-SNAPSHOT

**Describe the bug**
The project dependencies declare conflicting SLF4J implementations

**To Reproduce**
Compile and run any test

**Expected behavior**
There should only be one declaration of the SLF4J impl

**Urgency**
middle to low.

With the current project structure, the SLF4J logger will simply pick one of the bindings (either log4j 1.X, or the newer 2.x that is declared). We might still be using the old and outdated dependencies inadvertently

**Additional context**

I am constantly seeing this message in stdout/stderr

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/jenkins/.m2/repository/org/slf4j/slf4j-log4j12/1.7.25/slf4j-log4j12-1.7.25.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/jenkins/.m2/repository/org/apache/logging/log4j/log4j-slf4j-impl/2.13.3/log4j-slf4j-impl-2.13.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
```

ZacBlanco:
Fixes #11979 by excluding slf4j-log4j12

For more info, see http://www.slf4j.org/codes.html#multiple_bindings

### DIFF
--- a/core/common/pom.xml
+++ b/core/common/pom.xml
@@ -99,10 +99,6 @@
       <groupId>org.reflections</groupId>
       <artifactId>reflections</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-slf4j-impl</artifactId>
-    </dependency>
 
     <!-- Internal dependencies -->
     <dependency>
--- a/core/common/pom.xml
+++ b/core/common/pom.xml
@@ -99,6 +99,10 @@
       <groupId>org.reflections</groupId>
       <artifactId>reflections</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-slf4j-impl</artifactId>
+    </dependency>
 
     <!-- Internal dependencies -->
     <dependency>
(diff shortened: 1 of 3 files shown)


gpang:
```
/usr/src/alluxio/core/common/src/main/java/alluxio/util/LogUtils.java:26:8: error: Unused import - java.lang.reflect.Field.
```

ZacBlanco:
So this PR actually has a few issues that we need to tackle in order to entirely remove the Log4j12 classes.

We have some code that depends on the Log4j12 implementation (LogUtils.setLevel, log server filter). Once fully removed the `LogUtils.setLevel` ceases to function properly, and the LogServer filter implementation won't compile.

Essentially our upgrade to Log4j2 was in vain because we didn't remove any of the lingering log4j 1.2.x dependencies.

I've filed the issue for this, but it will probably take a fair amount of work investigating the new 2.x APIs

PR #11982, opened 2020-08-13
Revert rocksdb version

LuQQiu:
rockdsdb version upgrading to  6.11.4 cause the create file and delete file performance downgrade a lot when the alluxio system has a lot of files. Revert it back to previous version until we are clear about the downgradation reason


yuzhu:

PR #11984, opened 2020-08-14
Trim java opts when applying them in the command line

yuzhu:
This affects running benchmarks because job master and command line processes spaces differently.
It can cause benchmark jobs to fail.

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
@@ -87,12 +87,13 @@ public abstract class Benchmark<T extends TaskResult> {
    * @return the JobConfig
    * */
   public PlanConfig generateJobConfig(String[] args) {
-    // remove the cluster flag
-    List<String> commandArgs =
-            Arrays.stream(args).filter((s) -> !BaseParameters.CLUSTER_FLAG.equals(s))
-                    .filter((s) -> !s.isEmpty()).collect(Collectors.toList());
+    // remove the cluster flag and java opts
+    List<String> commandArgs = Arrays.stream(args).filter((s) ->
+        !BaseParameters.CLUSTER_FLAG.equals(s) && !s.isEmpty())
+        .collect(Collectors.toList());
 
-    commandArgs.addAll(mBaseParameters.mJavaOpts);
+    commandArgs.addAll(mBaseParameters.mJavaOpts.stream().map(String::trim)
+        .collect(Collectors.toList()));
     String className = this.getClass().getCanonicalName();
     return new StressBenchConfig(className, commandArgs, 10000, mBaseParameters.mClusterLimit);
   }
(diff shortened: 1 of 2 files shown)


gpang:
Is there an example of what was broken? I don't understand the change.

yuzhu:
`TestXraySuite/TestMasterMaxThroughput` fails using default values..

yuzhu:
If you have not seen this behavior before, it might be due to the recent upgrade of JCommander dependency..

gpang:
@yuzhu Ahhh, I see. Yeah, this was not a problem before. Could you verify that parameter like `' -Xmx8g'` does propagate all the way to the process spawned by the job worker? I think this can be verified on a local instance.

yuzhu:

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -87,10 +87,12 @@ protected static void mainInternal(String[] args, Benchmark benchmark) {
    * @return the JobConfig
    * */
   public PlanConfig generateJobConfig(String[] args) {
-    // remove the cluster flag
+    // remove the cluster flag and java opts
```
gpang, line 90: The java opts are intentionally passed to the spawned tasks, so that the tasks can also be started with a larger mem size

yuzhu:
jenkins, test this please

yuzhu:
jenkins, test this please

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
yuzhu: when are the java opts used?

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -87,12 +87,15 @@ protected static void mainInternal(String[] args, Benchmark benchmark) {
...
     List<String> commandArgs =
-            Arrays.stream(args).filter((s) -> !BaseParameters.CLUSTER_FLAG.equals(s))
-                    .filter((s) -> !s.isEmpty()).collect(Collectors.toList());
+            Arrays.stream(args).filter((s) -> !BaseParameters.JAVA_OPT_FLAG.equals(s)
+                && !BaseParameters.CLUSTER_FLAG.equals(s)
+                && !s.isEmpty()
+                && !mBaseParameters.mJavaOpts.contains(s)).collect(Collectors.toList());
 
-    commandArgs.addAll(mBaseParameters.mJavaOpts);
```
yuzhu, line 95: the JavaOpts is applied to the job launch process here. are you saying there is another level of launching jobs within the job?

gpang: basically, all job tasks on the workers also spawn processes, to make it look exactly like it would have if it was run locally (not in cluster mode). https://github.com/Alluxio/alluxio/blob/master/job/server/src/main/java/alluxio/job/plan/stress/StressBenchDefinition.java#L114

I just want to make sure if we set java opts for max mem, that it will affect local mode, as well as cluster mode, all in the same way.

yuzhu:
@gpang
I added back passing down the java opts. The important part was to trim the java opts which are meant to be read by the jvm.

yuzhu:
discussed with @gpang and verified it works for local and cluster mode

yuzhu:

yuzhu:

PR #12013, opened 2020-08-19
Fix local cache locking

apc999:
Fix two issues:

(1) Previously MetaStore and Evictor are both used as the state to make eviction decisions, but they are not put in the same critical section, and not coordinated to be consistent in tricky cases of failed page store deletes. This PR makes evictor a part of MetaStore so their content will always be consistent
(2) Eviction is a 2PC (phase 1 in metastore and phase 2 in pagestore), this PR fixes the undo logic when this 2PC failed in the middle.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
@@ -13,21 +13,36 @@ package alluxio.client.file.cache;
 
 import alluxio.exception.PageNotFoundException;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
+
+import javax.annotation.Nullable;
 
 /**
  * The default implementation of a metadata store for pages stored in cache. This implementation
  * is not thread safe and requires synchronizations on external callers.
  */
 public class DefaultMetaStore implements MetaStore {
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultMetaStore.class);
   /** A map from PageId to page info. */
   private final Map<PageId, PageInfo> mPageMap = new HashMap<>();
   /** The number of logical bytes used. */
   private final AtomicLong mBytes = new AtomicLong(0);
   /** The number of pages stored. */
   private final AtomicLong mPages = new AtomicLong(0);
+  /** The evictor. */
+  private final CacheEvictor mEvictor;
+
+  /**
+   * @param evictor cache evictor
+   */
+  public DefaultMetaStore(CacheEvictor evictor) {
+    mEvictor = evictor;
+  }
 
   @Override
   public boolean hasPage(PageId pageId) {
@@ -39,6 +54,7 @@ public class DefaultMetaStore implements MetaStore {
     mPageMap.put(pageId, pageInfo);
     mBytes.addAndGet(pageInfo.getPageSize());
     mPages.incrementAndGet();
+    mEvictor.updateOnPut(pageId);
   }
 
   @Override
@@ -46,6 +62,7 @@ public class DefaultMetaStore implements MetaStore {
     if (!mPageMap.containsKey(pageId)) {
       throw new PageNotFoundException(String.format("Page %s could not be found", pageId));
     }
+    mEvictor.updateOnGet(pageId);
     return mPageMap.get(pageId);
   }
 
@@ -57,6 +74,7 @@ public class DefaultMetaStore implements MetaStore {
     PageInfo pageInfo = mPageMap.remove(pageId);
     mBytes.addAndGet(-pageInfo.getPageSize());
     mPages.decrementAndGet();
+    mEvictor.updateOnDelete(pageId);
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -300,25 +297,25 @@ private boolean putInternal(PageId pageId, byte[] page) {
       }
     }
 
-    Pair<ReadWriteLock, ReadWriteLock> pageLockPair = getPageLockPair(pageId, victim);
+    Pair<ReadWriteLock, ReadWriteLock> pageLockPair =
+        getPageLockPair(pageId, victimPageInfo.getPageId());
     try (LockResource r1 = new LockResource(pageLockPair.getFirst().writeLock());
-        LockResource r2 = new LockResource(pageLockPair.getSecond().writeLock())) {
+         LockResource r2 = new LockResource(pageLockPair.getSecond().writeLock())) {
```
calvinjia, line 303: Spacing

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -300,25 +297,25 @@ private boolean putInternal(PageId pageId, byte[] page) {
...
       try (LockResource r3 = new LockResource(mMetaLock.writeLock())) {
         if (mMetaStore.hasPage(pageId)) {
           LOG.debug("{} is already inserted by a racing thread", pageId);
           // TODO(binfan): we should return more informative result in the future
           return true;
         }
-        if (!mMetaStore.hasPage(victim)) {
+        if (!mMetaStore.hasPage(victimPageInfo.getPageId())) {
```
calvinjia, line 310: Should we change the metric below to be more specific?

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
```diff
@@ -74,5 +84,16 @@ public void reset() {
...
+  }
+
+  @Override
+  @Nullable
+  public PageInfo evict() {
+    PageId victim = mEvictor.evict();
+    if (victim == null) {
+      return null;
+    }
+    return mPageMap.get(victim);
```
calvinjia, line 97: If this is `null`, do we need to do something to update the evictor state?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -327,37 +324,43 @@ private boolean putInternal(PageId pageId, byte[] page) {
           mMetaStore.addPage(pageId, new PageInfo(pageId, page.length));
         }
       }
-      if (deletePage(victim, victimPageInfo)) {
+      // Regardless enoughSpace, evicting victim to make progress
```
calvinjia, line 327: Not sure if this comment is accurate, we already removed the victim from the metastore on line 316, so I think we need to go through with the ondisk delete?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -327,37 +324,43 @@ private boolean putInternal(PageId pageId, byte[] page) {
...
+      if (deletePage(victimPageInfo.getPageId())) {
         Metrics.BYTES_EVICTED_CACHE.mark(victimPageInfo.getPageSize());
         Metrics.PAGES_EVICTED_CACHE.mark();
       } else {
-        LOG.debug("Failed to evict page: {}", victim);
+        LOG.debug("Failed to evict page: {}", victimPageInfo.getPageId());
         Metrics.PUT_EVICTION_ERRORS.inc();
+        // something is wrong to add this page, let's remove it from meta store
```
calvinjia, line 334: Shouldn't this page have already been removed from the metastore?

calvinjia, line 334: Change this comment to `failed to evict page, remove new page from metastore as there will not be enough space`?

core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java
apc999: good catch. In theory, this should never happen if evictor is implemented correctly.
but no one can guarantee that. corresponding error log and cleanup added

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: count it as `PUT_RACING_EVICTION_ERRORS`

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -279,87 +278,86 @@ private boolean putInternal(PageId pageId, byte[] page) {
         }
       }
       if (enoughSpace) {
-        boolean ok = addPage(pageId, page);
-        if (ok) {
+        try {
+          mPageStore.put(pageId, page);
```
calvinjia, line 282: Do we need to return after succeeding?

apc999: oops, good catch!

apc999:

calvinjia:

PR #11955, opened 2020-08-10
Update grpc and netty to resolve CVEs

patrickofriel-wk:
Updates for netty 4.1.48.Final and grpc 1.29.0 as outlined in #11923

grpc 1.31.0 had some changes that broke the build so I stopped at 1.29.0 since it had the needed Netty updates to resolve the CVE

PR #12014, opened 2020-08-19
Downgrade grpc to 1.28.1 and upgrade netty to 4.1.51.Final

yuzhu:
GRPC upgrade  beyond 1.29 (included) introduces significant perf regression in alluxio metadata operations. 
We downgrade it to 1.28.1 but upgrade the problematic netty package to fix CVE. 

This addresses perf regression introduced in #11955 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,8 +129,8 @@
     <build.path>build</build.path>
     <copycat.version>1.2.15</copycat.version>
     <glusterfs-hadoop.version>2.3.13</glusterfs-hadoop.version>
-    <grpc.version>1.31.0</grpc.version>
-    <netty.version>4.1.48.Final</netty.version>
+    <grpc.version>1.28.1</grpc.version>
+    <netty.version>4.1.51.Final</netty.version>
     <rocksdb.version>5.15.10</rocksdb.version>
     <hadoop.version>3.3.0</hadoop.version>
     <jacoco.version>0.8.5</jacoco.version>


yuzhu:
@patrickofriel-wk  FYI

yuzhu:
@ggezer PTAL , @gpang @calvinjia FYI

patrickofriel-wk:
Appreciate the heads up!

pom.xml
```diff
@@ -129,8 +129,8 @@
     <build.path>build</build.path>
     <copycat.version>1.2.15</copycat.version>
     <glusterfs-hadoop.version>2.3.13</glusterfs-hadoop.version>
-    <grpc.version>1.31.0</grpc.version>
-    <netty.version>4.1.48.Final</netty.version>
+    <grpc.version>1.28.1</grpc.version>
+    <netty.version>4.1.51.Final</netty.version>
```
ZacBlanco, line 133: Do these versions match the gRPC versions?

ZacBlanco: https://github.com/grpc/grpc-java/blob/v1.28.1/repositories.bzl#L27
If we use gRPC 1.28.1 we need to use netty 4.1.45.Final

https://github.com/grpc/grpc-java/blob/v1.28.1/repositories.bzl#L27-L38

yuzhu: nope. is that going to be a problem?  and this was done on purpose to avoid the GRPC regression but to address netty CVE.

yuzhu:
jenkins, test this please

pom.xml
ZacBlanco: because we have some parts of our codebase that touch netty and gRPC relies heavily on netty, it can lead to runtime errors if the versions that we declare mismatch from gRPC

ggezer: it's not mandatory. We should be fine as soon as tests pass.

yuzhu:

PR #12020, opened 2020-08-22
Add Kubernetes operator for Alluxio

cheyang:


apc999:
@cheyang  this PR is very large with 2000+ files. Are there any files auto-generated so we could remove from this PR?

cheyang:
> @cheyang this PR is very large with 2000+ files. Are there any files auto-generated so we could remove from this PR?

The vendor directory can be generated during build time, which is always through `go get`. But the concern is that `go get` depends on git commit, and if there is `hard reset`, the build may be failed.

Xenorith:
what do you mean by "`go get` depends on git commit"? if the go dependencies are vendored correctly, with each dependency's source repository and corresponding commit hash in a go.mod and go.sum file, there should be no issues pulling in the dependencies before building. the vendor directory can be ignored so that the downloaded vendor files won't show up as untracked files in git.
does this address your concern or is there another issue i am not understanding?

apc999:
Thanks for the contribution!

apc999:

PR #12025, opened 2020-08-25
Add synchronized to the builder

yuzhu:


yuzhu:
This might have caused the NPE during the sync process seen by the community.

yuzhu:
alluxio-bot, test this please

yuzhu:
@HelloHorizon PTAL

gpang:
@yuzhu 
Why does this cause an NPE? I don't think the `synchronized` is necessary, because in that part of the method, there is only 1 thread accessing that builder.

yuzhu:
@gpang this is not the cause of the NPE. closing it now.

PR #12027, opened 2020-08-25
Add <parent> declaration to webui pom

ZacBlanco:
License check hasn't been properly working for a while on this. Adding the parent declaration will fix this. It will also allow us to run `mvn license:check` from the root of the repo. Before, `license:check` would fail on the webui because it would attempt to use a different plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -912,6 +912,11 @@
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-deploy-plugin</artifactId>
+          <version>3.0.0-M1</version>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -910,6 +910,11 @@
           <artifactId>maven-dependency-plugin</artifactId>
           <version>3.0.1</version>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-deploy-plugin</artifactId>
+          <version>3.0.0-M1</version>
+        </plugin>
```
ZacBlanco, line 917: I've added this because there are two plugins which define a `<skip>true</skip>` for this plugin, but the plugin wasn't defined in `pluginManagement` of the root POM, so maven threw a warning. Adding a specific version to `pluginManagement` makes this warning go away

ZacBlanco:

PR #12028, opened 2020-08-25
Remove test that hasn't ran for 2 years

bradyoo:


bradyoo:
@yuzhu

yuzhu:
why remove it now? maybe we can resurrect it when we find the time?

bradyoo:
the problem is the way the tests are getting split up when using github-actions doesn't follow slow-tests and since slow-tests has only one entry that hasn't been touched for 2 years, I have decided to remove it.

bradyoo:
There is no way for this test to have any reasonable runtime. This test was slow since the test was instated and never ran ever.

yuzhu:

PR #12030, opened 2020-08-26
Increase master serving timeout for slower machines

bradyoo:
When running tests on less faster machines, the timeout for master serving timeout becomes an issue.


bradyoo:
@gpang

gpang:

PR #12033, opened 2020-08-26
Reduce the number of operations on BackupRestoreTest

bradyoo:
Too many threads are being created for this operation to be done well on less beefy machines.


bradyoo:
I don't think this one is good. Closing.

PR #12046, opened 2020-08-28
Remove synchronized in ConfigHashSync

bradyoo:
If the ConfigHashSync's Heartbeat thread is running, the client can't be closed. Unfortunately, the cancel on heartbeat thread that gets called can be unreliable at times and as a result, there is a possibility of a deadlock.

One Thread:
(waiting on ConfigHashSync synchronized)
at alluxio.client.file.ConfigHashSync.close(ConfigHashSync.java:112)
at alluxio.client.file.FileSystemContextReinitializer.close(FileSystemContextReinitializer.java:189)
(has FileSystemContext synchronized lock)
at alluxio.client.file.FileSystemContext.close(FileSystemContext.java:262)

HeartBeatThread:
(waiting on FileSystemContext's synchronized lock)
at alluxio.client.file.FileSystemContext.getMasterAddress(FileSystemContext.java:417)
at alluxio.client.file.FileSystemContext.reinit(FileSystemContext.java:348)
(has ConfigHashSync synchronized lock)
at alluxio.client.file.ConfigHashSync.heartbeat(ConfigHashSync.java:95)
    

Resolve this by removing synchronized on ConfigHashSync since all of its instance variables are threadsafe.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
@@ -72,7 +72,7 @@ public final class ConfigHashSync implements HeartbeatExecutor {
   }
 
   @Override
-  public void heartbeat() {
+  public synchronized void heartbeat() {
     if (!mContext.getClientContext().getClusterConf().clusterDefaultsLoaded()) {
       // Wait until the initial cluster defaults are loaded.
       return;
--- a/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
@@ -72,7 +72,7 @@ public final class ConfigHashSync implements HeartbeatExecutor {
   }
 
   @Override
-  public synchronized void heartbeat() {
+  public void heartbeat() {
     if (!mContext.getClientContext().getClusterConf().clusterDefaultsLoaded()) {
       // Wait until the initial cluster defaults are loaded.
       return;
@@ -108,7 +108,7 @@ public final class ConfigHashSync implements HeartbeatExecutor {
   }
 
   @Override
-  public synchronized void close() {
+  public void close() {
     mClient.close();
   }
 }


bradyoo:
@yuzhu

bradyoo:

core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
```diff
@@ -72,7 +72,7 @@ public void resetMetaMasterConfigClient(MasterClientContext context) {
   }
 
   @Override
-  public synchronized void heartbeat() {
+  public void heartbeat() {
```
gpang, line 75: I'm not entirely sure we are allowed to remove this `synchronized`. However, I know we can remove the one with `close()`, since it just calls another method which is already `synchronized`. If we just remove the `synchronized` on the `ConfigHashSync#close()`, that should remove the deadlock cycle, right?

I'm just concerned that `synchronized` might be required for `heartbeat()`.

yuzhu, line 75: `mClient` is used in this method.. so what if another thread closes it before an operation on `mClient`?

yuzhu:
@gpang looks like we need a `synchronized` on the review!

core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
bradyoo: We have one thread that should be running heartbeat so there or not, it shouldn't really make a difference. I'll put it back.

bradyoo: mClient is also threadsafe by having its own synchronized methods.

bradyoo:
feedback addressed.

core/client/fs/src/main/java/alluxio/client/file/ConfigHashSync.java
yuzhu: mClient being synchronized does not help right? 
`mClient.close()` runs first, and then `mClient.someRPCCall()` or `mClient.connect()`  would fail and throw a `FailedPreconditionException` which is not handled gracefully..

gpang:
Jenkins, test this please

gpang:
Jenkins, test this please

bradyoo:
alluxio-bot, test this please

bradyoo:
@gpang can you merge it?

gpang:

PR #12052, opened 2020-08-31
Fixed - Downloading or uploading artifacts over insecure protocol/ HT…

chetkhatri:
Fixed - Downloading or uploading artifacts over insecure protocol/ HTTP >> HTTPS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
       <id>haoyuan</id>
       <name>Haoyuan Li</name>
       <email>haoyuan@alluxio.com</email>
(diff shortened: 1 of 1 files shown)


Xenorith:
this change will not work as this artifacts url is backed by a s3 bucket via static website hosting. 

> Amazon S3 doesn't support HTTPS access for website endpoints.
taken from https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html

is there a particular problem you are facing that necessitates this change?

PR #12056, opened 2020-09-01
Remove the catalyst dependencies in backup transport

LuQQiu:


LuQQiu:
06:05:08.418 [ERROR] alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterCatchup
06:05:08.418 [ERROR]   Run 1: RaftJournalTest.gainPrimacyAfterCatchup:286 » Timeout Timed out waiting for fu...
06:05:08.419 [ERROR]   Run 2: RaftJournalTest.gainPrimacyAfterCatchup:286 » Timeout Timed out waiting for fu...

LuQQiu:
Jenkins, test this please

LuQQiu:
17:39:17.802 [ERROR] alluxio.server.ft.journal.raft.EmbeddedJournalIntegrationTest.restartStress  Time elapsed: 120.198 s  <<< ERROR!
java.util.concurrent.TimeoutException: Timed out waiting for 11 successes options: WaitForOptions{interval=20, timeout=120000} last value: false
	at alluxio.server.ft.journal.raft.EmbeddedJournalIntegrationTest.restartStress(EmbeddedJournalIntegrationTest.java:337)

LuQQiu:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/transport/Listeners.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.function.Consumer;
+
+/**
+ * A class to hold all listeners belong to given type T.
+ *
+ * @param <T> the listener type
+ */
+public class Listeners<T> implements Iterable<Listener<T>> {
```
bf8086, line 29: Is there any Java built-in or guava alternative of this listener we can use?

core/server/common/src/main/java/alluxio/master/transport/GrpcMessagingContext.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+/**
+ * The context for Grpc messaging single thread.
+ * This context uses a {@link ScheduledExecutorService} to schedule events on the context thread.
+ */
+public class GrpcMessagingContext {
```
bf8086, line 37: Is this class needed? Will there be any problem running everything without this context?

LuQQiu: This class mainly store two things, one executor, and one serializer. 
In all the places, the current code makes sure that we read the response and write request to the same serializer.
all the requests are handled by the same executor, all the listeners acceptances are handled by the same executor.
It's more for the correctness and reduces the potential race conditions.

core/server/common/src/main/java/alluxio/master/transport/Listeners.java
LuQQiu: The current implementation bounds the listeners with a shared thread context. So that all the listeners accept method will be executed by the same executor.

core/server/common/src/main/java/alluxio/master/transport/GrpcMessagingContext.java
LuQQiu: I took a deeper look into the code, the backup leader may create more than one bi-di stream (in case need to reconnect?) with the client, but all the streams shared the same context (same executor and serializer).

bf8086: I see. We can leave it as it for now. 

For serializer we will probably replace it with protobuf built-in serialization. The gRPC also provide some guarantee in message order, so this class may eventually become unnecessary.

LuQQiu:

PR #12059, opened 2020-09-01
Remove assert in JVM Pause Monitor stop

bradyoo:
This assert came up a lot when running integration tests in smaller machine.

### DIFF
--- a/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
+++ b/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
@@ -88,8 +88,9 @@ public class JvmPauseMonitor {
    * Stops jvm monitor.
    */
   public void stop() {
-    Preconditions.checkState(mJvmMonitorThread != null,
-        "JVM monitor thread does not start");
+    if (mJvmMonitorThread == null) {
+      return;
+    }
     mJvmMonitorThread.interrupt();
     try {
       mJvmMonitorThread.join();


core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
```diff
@@ -88,8 +88,9 @@ public void start() {
    * Stops jvm monitor.
    */
   public void stop() {
-    Preconditions.checkState(mJvmMonitorThread != null,
-        "JVM monitor thread does not start");
+    if (mJvmMonitorThread == null) {
```
gpang, line 91: I think this should be fine, but maybe we should also add `synchronized` on both `start` and `stop` to keep the starting/stopping thread-safe.

How do tests call start and stop in different threads?

bradyoo: I honestly am not sure if its really the result of calling it in different threads or if it's a result of some bad code. But my main thought around this was that we really don't need assert statements that block you from doing better cleanup when we are tearing things down.

bradyoo:
ping @gpang

bradyoo:
ping @gpang

core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
gpang: If start and stop are not synchronized, couldn't there be a situation where the clean up is not performed at all? It would be the same situation when the assert failed.

gpang (commented):
@bradyoo

core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
bradyoo: Well, that's one way this could fail. But if you look at this alternatively, if you crash here, all you are saying is that I'm going to stop this process without doing rest of the cleanup. When the whole point of stopping this is to stop the process.

gpang (approved):
LGTM

gpang:

PR #12066, opened 2020-09-02
Change Whitespace to show Github Actions running

bradyoo:


bradyoo:
this is how it looks like on PR @gpang @yuzhu note that it will be double this for Java 11. (Except for Checkstyle, Findbugs, Doc Check)

PR #8517, opened 2019-03-05
Fix the way we shut down the checkpoint thread

aaudiber:
This PR makes two changes:
1. During checkpoint thread shutdown, only interrupt the journal checkpoint thread if it is in the process of creating a checkpoint. Interrupting is unnecessary outside of checkpointing, and may cause the thread to exit early instead of replaying all logs and waiting the quiet period.
2. When checkpointing throws an exception, inspect the causal chain to determine whether the root cause was interruption. If it wasn't, log an error. Either way, continue instead of crashing the checkpoint thread.

Previously we would catch `InterruptedException` thrown during `writeToCheckpoint()`, but this isn't enough because interrupted exceptions can get converted to `InterruptedIOException` or wrapped in a `RuntimeException` like `KyroException` (or both, like below).

> ERROR UfsJournalCheckpointThread - FileSystemMaster: Failed to run journal checkpoint thread, crashing.
> com.esotericsoftware.kryo.KryoException: java.io.InterruptedIOException: Interrupted while waiting for data to be acknowledged by pipeline
> Caused by: java.io.InterruptedIOException: Interrupted while waiting for data to be acknowledged by pipeline

Caveats:
- The checkpoint thread is not properly interruptible while shutting down and taking a checkpoint. Instead of exiting promptly, it will run until all logs are replayed and the quiet period elapses.
- If an implementation of `writeToCheckpoint` throws an exception because of interruption, but doesn't include an `InterruptedException` or `InterruptedIOException` in its causal chain, we will log an error

PR #12070, opened 2020-09-03
Propagate RuntimeExceptions in UfsJournalCheckpointThread

bradyoo:
Partial revert of https://github.com/Alluxio/alluxio/pull/8517/files#diff-5e678cb4bd385e5896b4bd970c8ed5eaR261

Occasionally writing to checkpoint fails because of some bad state. When this happens, it is very likely that repeating the reading of journals back from the beginning won't help while crashing an restarting the process may. This is an attempt at that.


bradyoo:
Ping? @gpang

bradyoo:
@bf8086

PR #12072, opened 2020-09-03
Merge JNIFuse support to master branch

apc999:


integration/fuse/src/main/java/alluxio/fuse/StackFS.java
```diff
@@ -0,0 +1,154 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Stack FS.
+ */
+public class StackFS extends AbstractFuseFileSystem {
```
iluoeli, line 32: is StackFS still used ? I think it's only used for debuging.

integration/docker/Dockerfile.fuse
```diff
@@ -29,17 +29,19 @@ RUN cd /opt && \
 # Remove the UFS libraries from the container, as the fuse daemon doesn't need them
 RUN rm -rf /opt/alluxio/lib/*
 
-FROM ubuntu:16.04
+FROM cheyang/fuse2:ubuntu1604-customize
```
iluoeli, line 32: The docker image is customized

PR #12075, opened 2020-09-05
Add s3 connection TTL config property

cde-jq:


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -988,6 +988,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_S3_CONNECT_TTL =
+          new Builder(Name.UNDERFS_S3_CONNECT_TTL)
+                  .setDefaultValue(-1)
+                  .setDescription("The expiration time of S3 connections in ms.")
```
gpang, line 994: Could you add to description saying that `-1` means the connection will never expire? Also, could you add some text about the implications of a shorter and longer ttl value?

gpang (changes requested):
@cde-jq Thanks for the feature! I left a question.

core/common/src/main/java/alluxio/conf/PropertyKey.java
cde-jq: Sorry, I cann't describe the implications of a shorter and longer ttl value. I've added a description of the default value.

We found broken pipe excepion in Alluxo Worker process  in high pressure. when use THOURH write type.   We tested with a private object storage which use nginx.  The nginx config keepalive_timeout to 180s.  Set S3 SDK connectTTL should solve the broken pipe problem.

gpang (approved):
LGTM

gpang:

PR #12079, opened 2020-09-08
Implement serializer for messaging transport

LuQQiu:


LuQQiu:
Close and keep catalyst dependency for now.
In the future, change to GRPC proto and direct rpc with detailed messages

PR #12081, opened 2020-09-08
Fix active sync manager error on secondary master

bf8086:
There has been errors when an active sync enabled secondary master get promoted to primary. It is caused by active sync manager trying to access UFS when master is in secondary mode. UFS client are not initialized for mount points on secondary master so error occurs when active sync manager accesses it during snapshotting and server stopping process.

This fix updated active sync manager to avoid unnecessary access to UFS client on a secondary master.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
+++ b/core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
@@ -23,6 +23,7 @@ import alluxio.heartbeat.HeartbeatThread;
 import alluxio.master.file.FileSystemMaster;
 import alluxio.master.file.RpcContext;
 import alluxio.master.file.meta.MountTable;
+import alluxio.master.file.meta.options.MountInfo;
 import alluxio.master.journal.checkpoint.CheckpointName;
 import alluxio.master.journal.JournalContext;
 import alluxio.master.journal.Journaled;
@@ -98,6 +99,7 @@ public class ActiveSyncManager implements Journaled {
   private final FileSystemMaster mFileSystemMaster;
   // a local executor service used to launch polling threads
   private final ThreadPoolExecutor mExecutorService;
+  private boolean mStarted;
 
   /**
    * Constructs a Active Sync Manager.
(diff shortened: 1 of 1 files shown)


bf8086:
@yuzhu @gpang PTAL.

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
```diff
@@ -164,6 +166,7 @@ public boolean isUnderSyncPoint(AlluxioURI path) {
    * Start the polling threads.
    */
   public void start() throws IOException {
+    mStarted = true;
```
gpang, line 169: Where is this set to `false`?

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
```diff
@@ -428,8 +431,9 @@ public boolean hasNext() {
         while (mountId == -1) {
           try {
             syncPointPath = mEntry.getPath();
-            MountTable.Resolution resolution = mMountTable.resolve(mEntry);
-            mountId = resolution.getMountId();
+            String mountPoint = mMountTable.getMountPoint(mEntry);
```
gpang, line 434: What is the difference from before? How does the new code avoid UFS access?

gpang (commented):
@bf8086 Thanks! I left a few questions

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
bf8086: Added in `stop()`.

core/server/master/src/main/java/alluxio/master/file/activesync/ActiveSyncManager.java
bf8086: `MountTable::resolve()` will actually creates a UFS client to query the UFS path while the new code only grabs the `MountInfo`. This function only need the mount id so resolution is not required.

yuzhu (approved):
thanks for fixing this! LGTM after test passes

gpang (approved):
LGTM

bf8086:
Alluxio-bot, merge this please.

bf8086:

PR #12087, opened 2020-09-09
Fix slow and contented local cache restart

apc999:
- remove restore logic to a separate background thread meanwhile open the cache manager in a "readonly" mode to other query threads. in this way, early queries will suffer from low hit ratio but hopefully they will catch up.
- handle capacity exceeding gracefully: instead of failing the restore and delete all pages on disk, just remove the excessive pages

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
@@ -27,11 +27,15 @@ public interface CacheManager extends AutoCloseable {
     /**
      * this cache is read only.
      */
-    READ_ONLY(0),
+    NOT_IN_USE(0),
+    /**
+     * this cache is read only.
+     */
+    READ_ONLY(1),
     /**
      * this cache can both read and write.
      */
-    READ_WRITE(1);
+    READ_WRITE(2);
 
     private final int mValue;
 
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
@@ -18,7 +18,35 @@ import java.io.IOException;
 /**
  * Interface for managing cached pages.
  */
-public interface CacheManager extends AutoCloseable  {
+public interface CacheManager extends AutoCloseable {
+
+  /**
+   * State of a cache.
+   */
+  enum State {
+    /**
+     * this cache is read only
+     */
+    READ_ONLY(0),
+    /**
+     * this cache can both read and write
+     */
+    READ_WRITE(1);
+
+    private final int mValue;
+
+    State(int value) {
+      mValue = value;
+    }
+
+    /**
+     * @return the value of the state
+     */
+    public int getValue() {
+      return mValue;
+    }
+  }
+
   /**
    * @param conf the Alluxio configuration
    * @return an instance of {@link CacheManager}
@@ -71,4 +99,9 @@ public interface CacheManager extends AutoCloseable  {
    * @return true if the page is successfully deleted, false otherwise
    */
   boolean delete(PageId pageId);
+
+  /**
+   * @return state of this cache.
+   */
+  State state();
 }
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
@@ -25,11 +25,11 @@ public interface CacheManager extends AutoCloseable {
    */
   enum State {
     /**
-     * this cache is read only
+     * this cache is read only.
      */
     READ_ONLY(0),
     /**
-     * this cache can both read and write
+     * this cache can both read and write.
      */
     READ_WRITE(1);
 
@@ -101,7 +101,7 @@ public interface CacheManager extends AutoCloseable {
   boolean delete(PageId pageId);
 
   /**
-   * @return state of this cache.
+   * @return state of this cache
    */
   State state();
 }
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
@@ -25,7 +25,7 @@ public interface CacheManager extends AutoCloseable {
    */
   enum State {
     /**
-     * this cache is read only.
+     * this cache is not in use.
      */
     NOT_IN_USE(0),
     /**
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
@@ -11,6 +11,7 @@
 
 package alluxio.client.file.cache;
 
+import static alluxio.client.file.cache.CacheManager.State.NOT_IN_USE;
 import static alluxio.client.file.cache.CacheManager.State.READ_ONLY;
 import static alluxio.client.file.cache.CacheManager.State.READ_WRITE;
 
@@ -74,7 +75,6 @@ public class LocalCacheManager implements CacheManager {
   private static final Logger LOG = LoggerFactory.getLogger(LocalCacheManager.class);
 
   private static final int LOCK_SIZE = 1024;
-
   private final long mPageSize;
   private final long mCacheSize;
   private final boolean mAsyncWrite;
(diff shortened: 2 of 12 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -107,10 +107,13 @@ private static boolean restore(
...
-        metaStore.addPage(pageInfo.getPageId(), pageInfo);
+        PageId pageId = pageInfo.getPageId();
+        metaStore.addPage(pageId, pageInfo);
         if (metaStore.bytes() > pageStore.getCacheSize()) {
-          LOG.error("Loaded pages exceed cache capacity ({} bytes)", pageStore.getCacheSize());
-          return false;
+          metaStore.removePage(pageId);
+          pageStore.delete(pageId);
+          LOG.warn("Abandon {} in restore: cache size is {} bytes and {} bytes used",
+              pageInfo, pageStore.getCacheSize(), metaStore.bytes());
```
calvinjia, line 116: Will this part be performant enough if there are a large number of extra pages? Is there any way to parallelize or otherwise delete the rest of the stream at once after we detect the metastore is out of space?

calvinjia (changes requested):
@apc999 Thanks for the quick fix, had one comment.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: removed the warning but show the aggregated result in the end.
there is no easy way to bulk-delete the files after metastore is filled

apc999:
Jenkins, test this please

calvinjia (approved):

apc999:
@calvinjia  PTAL

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -424,9 +408,87 @@ public boolean delete(PageId pageId) {
...
+      Iterator<PageInfo> iterator = stream.iterator();
+      while (iterator.hasNext()) {
+        PageInfo pageInfo = iterator.next();
+        if (pageInfo == null) {
+          LOG.error("Invalid page info");
+          return false;
+        }
+        PageId pageId = pageInfo.getPageId();
+        ReadWriteLock pageLock = getPageLock(pageId);
+        try (LockResource r = new LockResource(pageLock.readLock())) {
```
calvinjia, line 457: Should this be a write lock because we might delete the page if there is not enough space?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -938,6 +950,12 @@ public MetricKey build() {
           .setMetricType(MetricType.COUNTER)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey CLIENT_CACHE_STATE =
+      new Builder(Name.CLIENT_CACHE_STATE)
+          .setDescription("State of the cache: 0 (NOT_READY), 1 (READ_ONLY) and 2 (READ_WRITE)")
```
calvinjia, line 955: Is this description accurate? It looks like we only have two states?

calvinjia (changes requested):
Left a few comments

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -424,9 +408,87 @@ public boolean delete(PageId pageId) {
...
+  @Override
+  public State state() {
+    return mState.get();
+  }
+
+  /**
+   * Restores a page store a the configured location, updating meta store.
+   */
+  private void restoreOrInit(PageStoreOptions options) {
+    Preconditions.checkState(mState.get() == READ_ONLY);
```
calvinjia, line 420: One concern here is if multiple threads can go into `restoreOrInit`.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -69,99 +74,67 @@
...
-        try {
-          pageStore.close();
-        } catch (Exception e) {
-          LOG.error("Failed to close PageStore", e);
-        }
-      }
-      metaStore.reset();
-      pageStore = PageStore.create(options, true);
+  @VisibleForTesting
+  static LocalCacheManager create(AlluxioConfiguration conf, MetaStore metaStore,
```
calvinjia, line 116: Are we guaranteed `create` is only called once? Otherwise we could queue up multiple `restoreOrInit`

core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
```diff
@@ -18,7 +18,35 @@
...
+public interface CacheManager extends AutoCloseable {
+
+  /**
+   * State of a cache.
+   */
+  enum State {
+    /**
+     * this cache is read only.
+     */
+    READ_ONLY(0),
```
calvinjia, line 30: If `restoreOrInit` hasn't completed, is it safe for `get` calls? Will there be some objects which could be null?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: ensured by https://github.com/Alluxio/alluxio/pull/12087/files#diff-f604b8c2e5cc2efb92656285c2a61cbfR50

apc999: also made a few methods synchronized to further ensure this

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: synchronized

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: good catch. corrected

apc999:
@calvinjia  PTAL

core/client/fs/src/main/java/alluxio/client/file/cache/CacheManager.java
apc999: protected by locks

apc999:
@calvinjia  I introduced `NOT_IN_USE` state in case `restoreOrInit` also failed. PTAL

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -938,6 +956,12 @@ public MetricKey build() {
           .setMetricType(MetricType.COUNTER)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey CLIENT_CACHE_STATE =
+      new Builder(Name.CLIENT_CACHE_STATE)
+          .setDescription("State of the cache: 0 (NOT_IN_USE), 1 (READ_ONLY) and 2 (READ_WRITE)")
```
calvinjia, line 961: Should this be -1, 0, and 1?

calvinjia (changes requested):
Had a minor comment, thanks for the updates

core/common/src/main/java/alluxio/metrics/MetricKey.java
apc999: I feel 0,1,2 is more friendly for monitoring? my would go with 0, 1, 2

calvinjia: Yeah I agree, but is that how the value works in the code, doesn't the Metric.STATE counter initialize at 0?

apc999: ah you are right. Let me update the state transistion.

apc999:
@calvinjia  PTAL

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -371,6 +358,11 @@ public int get(PageId pageId, int pageOffset, int bytesToRead, byte[] buffer,
         "buffer does not have enough space: bufferLength=%s offsetInBuffer=%s bytesToRead=%s",
         buffer.length, offsetInBuffer, bytesToRead);
     LOG.debug("get({},pageOffset={}) enters", pageId, pageOffset);
+    if (mState.get() == NOT_IN_USE) {
+      Metrics.GET_NOT_READY_ERRORS.inc();
+      Metrics.GET_ERRORS.inc();
+      return 0;
```
calvinjia, line 364: Should this be `return -1` not sure if this counts as error or page not found?

calvinjia (changes requested):
Looks good overall, had a minor comment

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: updated

calvinjia (approved):

apc999:
Jenkins, test this please

apc999:

PR #12094, opened 2020-09-10
Prevent UI from installing dependencies twice

ZacBlanco:
The maven build actually executes the generate-sources phase twice
because the maven-sources-plugin calls the generate-sources phase during
its attach-sources target. This causes the UI module to install node/npm
AND run npm install twice during the build. Moving these plugins from
generate-sources to process-sources (the next step in the maven
lifecycle) prevents the extra calls and saves time during the build.

### DIFF
--- a/webui/pom.xml
+++ b/webui/pom.xml
@@ -40,7 +40,7 @@
             <goals>
               <goal>install-node-and-npm</goal>
             </goals>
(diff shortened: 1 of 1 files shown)


tieujason330 (approved):

ZacBlanco:

PR #12099, opened 2020-09-10
Enable new master to join an existing quorum

bf8086:
This change implements the quorum join process for new master.

Ratis provides an API for update quorum member list. To implement an automatic join comparable with what we did with Copycat, the new master will send a join request to the quorum leader after it initialized the journal. The leader will  atomically convert the request to a raft configuration change and apply it to the quorum. 

This operation is a noop for a master that is already in the quorum.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -14,6 +14,7 @@ package alluxio.master.journal.raft;
 import alluxio.ProcessUtils;
 import alluxio.conf.PropertyKey;
 import alluxio.conf.ServerConfiguration;
+import alluxio.grpc.AddQuorumServerRequest;
 import alluxio.grpc.JournalQueryRequest;
 import alluxio.master.journal.CatchupFuture;
 import alluxio.master.journal.JournalUtils;
@@ -57,6 +58,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
@@ -184,6 +186,17 @@ public class JournalStateMachine extends BaseStateMachine {
       if (reply != null) {
         future.complete(reply);
         return future;
+      }
+      if (queryRequest.hasAddQuorumServerRequest()) {
+        AddQuorumServerRequest addRequest = queryRequest.getAddQuorumServerRequest();
+        return CompletableFuture.supplyAsync(() -> {
+          try {
+            mJournalSystem.addQuorumServer(addRequest.getServerAddress());
+          } catch (IOException e) {
+            throw new CompletionException(e);
+          }
+          return Message.EMPTY;
+        });
       }
     } catch (Exception e) {
       LOG.error("failed processing request {}", request, e);
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -182,11 +182,14 @@ public class JournalStateMachine extends BaseStateMachine {
       JournalQueryRequest queryRequest = JournalQueryRequest.parseFrom(
           request.getContent().asReadOnlyByteBuffer());
       LOG.debug("Received query request: {}", queryRequest);
+      // give snapshot manager a chance to handle snapshot related requests
       Message reply = mSnapshotManager.handleRequest(queryRequest);
       if (reply != null) {
         future.complete(reply);
         return future;
       }
+      // Snapshot manager returned null indicating the request is not handled. Check and handle
+      // other type of requests.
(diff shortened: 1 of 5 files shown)


bf8086:
@gpang @LuQQiu PTAL.

bf8086:
jenkins, test this please

bf8086:
jenkins, test this please

bf8086:
Jenkins, test this please

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -562,6 +566,38 @@ public synchronized void startInternal() throws InterruptedException, IOExceptio
...
+            JournalQueryRequest
+                .newBuilder()
+                .setAddQuorumServerRequest(request)
+                .build().toByteArray()
+        ))).whenComplete((reply, t) -> {
+          if (t != null) {
+            LOG.error("Exception occurred while joining quorum", t);
+          }
+          if (reply != null && reply.getException() != null) {
+            LOG.error("Received an error while joining quorum", reply.getException());
```
LuQQiu, line 593: If joinQuorum failed, the master with raft journal system will still run?? Should we shut it down or retry joining?

LuQQiu (commented):
Thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -185,6 +187,17 @@ public StateMachineStorage getStateMachineStorage() {
         future.complete(reply);
         return future;
       }
+      if (queryRequest.hasAddQuorumServerRequest()) {
```
gpang, line 190: This journal request can have various types of requests, like snapshot (above), and now add quorum? There should only be 1 set at a time?

Also, should this section be like:
```
if (queryRequest.hasGetSnapshotInfoRequest()) {
...
else if (queryRequest.hasGetSnapshotRequest()) {
...
else if (queryRequest.hasAddQuorumServerRequest()) {
...
}
```

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -562,6 +566,38 @@ public synchronized void startInternal() throws InterruptedException, IOExceptio
...
+            .setRpcPort(localAddress.getPort()))
+        .build();
+    RaftClient client = createClient();
+    client.sendReadOnlyAsync(Message.valueOf(
+        UnsafeByteOperations.unsafeWrap(
+            JournalQueryRequest
+                .newBuilder()
+                .setAddQuorumServerRequest(request)
+                .build().toByteArray()
+        ))).whenComplete((reply, t) -> {
```
gpang, line 588: Does `whenComplete` run even on failure?

gpang (changes requested):
@bf8086 I left some minor comments

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Yes, it always runs at the end of the action. When the action fails a throwable is given in as the second parameter.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Yes. The join process is a best effort call with a retry. When a new quorum is forming a new leader might not be available when master starts. We should not shutdown the master for that. If the master later is rejected by the quorum because the unsuccessful join, the server will be shutdown.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Yes there should be only one set for a request.

Both `GetSnapshotInfoRequest` and `GetSnapshotRequest` are handled by `SnapshotManager`. Only when a valid reply is not given does this function continue to check other types of request.

bf8086:
@LuQQiu @gpang Thanks! I replied to the comments. PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
gpang: I see. Can we just add a small comment, either here or above, saying something like handleRequest() handles GetSnapshotInfoRequest and GetSnapshotRequest, or returns `null`, so that this part handles the other requests (AddQuorumServerRequest).

gpang (changes requested):
@bf8086 Just a minor comment

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: Added some comments.

bf8086:
Thanks @gpang! I updated the PR. PTAL.

gpang (approved):
LGTM

LuQQiu (approved):

bf8086:

PR #12100, opened 2020-09-11
Avoid taking snapshot during journal suspension

bf8086:
When journal is suspended, snapshot should not be taken because some entries included up to the snapshot index might not have been applied by the suspended journal applier.

### DIFF
--- a/core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
+++ b/core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
@@ -139,6 +139,50 @@ public class RaftJournalTest {
         () -> countingMaster.getApplyCount() == entryCount + 1, mWaitOptions);
   }
 
+  @Test
+  public void suspendSnapshotRestart() throws Exception {
+    // Create a counting master implementation that counts how many journal entries it processed.
+    CountingDummyFileSystemMaster countingMaster = new CountingDummyFileSystemMaster();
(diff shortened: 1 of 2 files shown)


bf8086:
@gpang @LuQQiu PTAL.

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
```diff
@@ -139,6 +139,50 @@ public void suspendCatchupResume() throws Exception {
...
+              .setInodeLastModificationTime(
+                  File.InodeLastModificationTimeEntry.newBuilder().setId(entryCount).build())
+              .build());
+    }
+
+    // Ask the follower to do a snapshot.
+    mFollowerJournalSystem.checkpoint();
+
+    // Restart the follower.
+    mFollowerJournalSystem.stop();
```
LuQQiu, line 175: why added restart follower here

LuQQiu (commented):
Thanks, left a minor comment

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
bf8086: Restart causes the follower snapshot to be replayed so its content can be verified.

gpang (approved):
LGTM

LuQQiu (approved):
LGTM

bf8086:

PR #12104, opened 2020-09-12
Support dynamic user to run alluxio

cheyang:
The issue is that alluxio docker image only supports two users: alluxio and root, but in the customer env, they have more users and the data of each user are isolated. So we need to make alluxio docker image can support run container by specifying during running time. 

This feature by default is disabled.

### DIFF
--- a/integration/docker/Dockerfile
+++ b/integration/docker/Dockerfile
@@ -18,6 +18,7 @@
 FROM alpine:3.10.2 AS extractor
 # Note that downloads for *-SNAPSHOT tarballs are not available
 ARG ALLUXIO_TARBALL=http://downloads.alluxio.io/downloads/files/2.4.0-SNAPSHOT/alluxio-2.4.0-SNAPSHOT-bin.tar.gz
+# (Alert):It's not recommended to set this Argument to true, unless you know exactly what you are doing 
 ARG ENABLE_DYANMIC_USER=false
 
 ADD ${ALLUXIO_TARBALL} /opt/
@@ -66,6 +67,7 @@ RUN if [ ${ALLUXIO_USERNAME} != "root" ] \
 COPY --chown=${ALLUXIO_USERNAME}:${ALLUXIO_GROUP} --from=extractor /opt/ /opt/
 COPY --chown=${ALLUXIO_USERNAME}:${ALLUXIO_GROUP} conf /opt/alluxio/conf/
 COPY --chown=${ALLUXIO_USERNAME}:${ALLUXIO_GROUP} entrypoint.sh /
+
 
 RUN if [ ${ENABLE_DYANMIC_USER} = "true" ] ; then \
        chmod -R 777 /journal; \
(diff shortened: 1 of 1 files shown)


apc999 (approved):

integration/docker/Dockerfile
```diff
@@ -26,11 +27,16 @@ RUN cd /opt && \
     (if ls | grep -q ".tar.gz"; then tar -xzf *.tar.gz && rm *.tar.gz; fi) && \
     ln -s alluxio-* alluxio
 
+RUN if [ ${ENABLE_DYANMIC_USER} = "true" ] ; then \
```
jiacheliu3, line 30: @apc999 I'm concerned about the security implications. If we really have to give a 777 permission, I think at least there should be warnings in the printout and code comments saying that's high security risk.

jiacheliu3 (commented):
The code part LGTM except the security risk concern. Please treat this as approval once everybody is okay with how the permission works.

integration/docker/Dockerfile
cheyang: By default it's disabled, but if the user want to build it for the user which is not in alluxio docker image. They have flexibility to enable it.

integration/docker/Dockerfile
```diff
@@ -61,6 +67,10 @@ COPY --chown=${ALLUXIO_USERNAME}:${ALLUXIO_GROUP} --from=extractor /opt/ /opt/
 COPY --chown=${ALLUXIO_USERNAME}:${ALLUXIO_GROUP} conf /opt/alluxio/conf/
 COPY --chown=${ALLUXIO_USERNAME}:${ALLUXIO_GROUP} entrypoint.sh /
 
+RUN if [ ${ENABLE_DYANMIC_USER} = "true" ] ; then \
```
madanadit, line 70: The journal is only accessed by Alluxio servers. Is there a case for running server containers not as root or Alluxio? Or is the requirement only for Alluxio client containers which can be a larger set of users?

integration/docker/Dockerfile
jiacheliu3: Yes, and I think in that case at least warnings should be given to the user in some printout (like `echo`?) and in the comment in the Dockerfile.

integration/docker/Dockerfile
cheyang: The user case is that:

every user in data scientist team has own NFS directory, and this directory is only accessible by his account. root  or alluxio account doesn't work. so alluxio container has to be run as the same user account.

cheyang: And I also wonder what's the real risk to make alluxio binary can be run by other users. And is there any simple sample for me to understand better?

integration/docker/Dockerfile
cheyang: Added

madanadit (approved):

apc999 (approved):

apc999:

apc999:
alluxio-bot, cherry-pick this to branch-2.3-fuse please

Issue #12088, opened 2020-09-09
异步元数据加载


cheyang:
**Is your feature request related to a problem? Please describe.**
我想加载ufs的元数据，目前的方法是通过"alluxio fs -Dalluxio.user.file.metadata.sync.interval=0 ls -R /"完成，但是这里会有两个问题：

1.同步操作，加载时间过长
2.alluxio fs 客户端会出现OOM的情况

**Describe the solution you'd like**

1.我希望能够异步触发alluxio加载metadata，甚至此时可以支持并发加载元数据
2.如果元数据还在加载中，我能查询到对应查询任务的状态，是完成了还是运行中
3.如果元数据加载完毕，我可以通过allxuio fs du -sh 查询到视图结果无需重新计算和拉取

**Describe alternatives you've considered**

**Urgency**
Urgent， we really rely on this feature.

**Additional context**

maobaolong:
```bash
$ alluxio fs startSync /syncedDir
$ alluxio fs stopSync /syncedDir
$ alluxio fs getSyncPathList
$ alluxio fs checkConsistency -r /
```

Does these above command and satisfy your request?

cheyang:
> ```shell
> $ alluxio fs startSync /syncedDir
> $ alluxio fs stopSync /syncedDir
> $ alluxio fs getSyncPathList
> $ alluxio fs checkConsistency -r /
> ```
> 
> Does these above command and satisfy your request?

If there are too many files， the alluxio client is OOM

apc999:
fixed by #12421

PR #12105, opened 2020-09-12
Add loadMetadata command

apc999:
We used to have a command `fs loadMetadata`
This PR restores loadMetadata command, but as a client-side optimization without storing all returned `ls` results, preventing OOM for massive amount of small files.

This optimization is experimental, going to branch-2.3-fuse only for now

Fix https://github.com/Alluxio/alluxio/issues/12088


apc999:

PR #12111, opened 2020-09-16
Add Filters to Job Failure History

bradyoo:


bradyoo:
@yuzhu

bradyoo:
@yuzhu

job/server/src/main/java/alluxio/master/job/plan/PlanTracker.java
```diff
@@ -276,7 +277,7 @@ private synchronized boolean removeFinished() {
   /**
    * @return list of failed plans
    */
-  public List<PlanInfo> failed() {
-    return Lists.newArrayList(mFailed);
+  public Stream<PlanInfo> failed() {
```
yuzhu, line 280: kind of unusual to pass stream back as return value, is the idea here to save a potential copy?

yuzhu (commented):
@bradyoo just a question on a function return value..

job/server/src/main/java/alluxio/master/job/plan/PlanTracker.java
bradyoo: I don't want to send the list back because that's basically making a copy. I don't want to send the underlying data structure back because that makes it accessible. So stream was both the most straightforward and most accessible thing to send back.

yuzhu:

PR #12117, opened 2020-09-21
Use local raft api for journal writer

bf8086:
This change may improve flushing performance by removing additional network layer from journal flushing.


bf8086:
Resubmitted as #12174

PR #12133, opened 2020-09-23
Update sync status after processing the entire sync

### LINKED ISSUE #12106: UfS sync causing duplication of data
**Alluxio Version:**
alluxio-2.3.0/ alluxio-2.4.0-Snapshot

**Describe the bug**
I am facing an issue with alluxio ufs sync.  I have some tables whose data get overwrite every hour in S3 but Alluxio not purging old files so my data is getting duplicated while I am query through Alluxio catalog in Presto.

As per doc:
Metadata sync keeps a fingerprint of each UFS file so that Alluxio can update the file if it changes. The fingerprint includes information such as file size and last modified time. If a file is modified in the UFS, Alluxio will detect this from the fingerprint, free the existing data for that file, and reload the metadata for the updated file. If a file is added or deleted in the UFS, Alluxio will update the metadata in its namespace as well.

**Data in Alluxio showing files of 2 hrs.** 

./bin/alluxio fs ls /catalog/dataplatform/tables/task/hive/
-rwx------  systemmanager  systemmanager                0       PERSISTED 09-08-2020 19:08:15:000 100% /catalog/dataplatform/tables/task/hive/_SUCCESS
-rwx------  systemmanager  systemmanager        129016033       PERSISTED 09-09-2020 08:07:55:000 100% /catalog/dataplatform/tables/task/hive/part-00000-e4b562fe-3d03-41ef-8b43-727e26c8aa1a-c000.snappy.parquet
-rwx------  systemmanager  systemmanager        128775563       PERSISTED 09-09-2020 09:08:26:000   0% /catalog/dataplatform/tables/task/hive/part-00000-e6ec7eac-f87a-498d-8429-e209a922b26e-c000.snappy.parquet
-rwx------  systemmanager  systemmanager        117141346       PERSISTED 09-09-2020 08:07:54:000 100% /catalog/dataplatform/tables/task/hive/part-00001-e4b562fe-3d03-41ef-8b43-727e26c8aa1a-c000.snappy.parquet
-rwx------  systemmanager  systemmanager        115965603       PERSISTED 09-09-2020 09:08:26:000   0% /catalog/dataplatform/tables/task/hive/part-00001-e6ec7eac-f87a-498d-8429-e209a922b26e-c000.snappy.parquet
-rwx------  systemmanager  systemmanager         91051189       PERSISTED 09-09-2020 08:07:53:000 100% /catalog/dataplatform/tables/task/hive/part-00002-e4b562fe-3d03-41ef-8b43-727e26c8aa1a-c000.snappy.parquet
-rwx------  systemmanager  systemmanager         86995170       PERSISTED 09-09-2020 09:08:23:000   0% /catalog/dataplatform/tables/task/hive/part-00002-e6ec7eac-f87a-498d-8429-e209a922b26e-c000.snappy.parquet
-rwx------  systemmanager  systemmanager         44151789       PERSISTED 09-09-2020 08:07:49:000 100% /catalog/dataplatform/tables/task/hive/part-00003-e4b562fe-3d03-41ef-8b43-727e26c8aa1a-c000.snappy.parquet
-rwx------  systemmanager  systemmanager         49253364       PERSISTED 09-09-2020 09:08:21:000   0% /catalog/dataplatform/tables/task/hive/part-00003-e6ec7eac-f87a-498d-8429-e209a922b26e-c000.snappy.parquet

**Data in S3:** 
image

**To Reproduce**
1.  add a file to the test s3 UFS path
     s3://.../test-a
2.  force the sync (expected to show just test-a)
    ./bin/alluxio fs -Dalluxio.user.file.metadata.sync.interval=0 ls /dir/
3.  remove the existing file and add a different file
     remove s3://.../test-a 
     add s3://.../test-b
4.  Use a longer sync interval (still expected to show just test-a)
     ./bin/alluxio fs -Dalluxio.user.file.metadata.sync.interval=5m ls /dir/
5.  wait over 5 minutes
6. Use a longer sync interval to see what it returns (expected to show just test-b but it will show both test-a and test-b)
     ./bin/alluxio fs -Dalluxio.user.file.metadata.sync.interval=5m ls /dir/

**Expected behavior**
When the sync is happening it is able to detect new files that got added in S3 but it not able to detect deleted files. So it detects delete also and those files should be deleted from Alluxio. 

**Urgency**
It's a very critical issue as we can not use alluxio with this issue as all the data is incorrect because of duplication. 

**Additional context**

gpang:
Marking paths as sync during the sync is inaccurate, because the syncs for descendants are run in a separate thread pool. Therefore, only mark the root path as synced after everything is finished. This also reduces the number of entries in the sync cache, instead of marking each path in the tree.

Fixes #12106

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -387,9 +387,15 @@ public class InodeSyncStream {
           syncPathCount + failedSyncPathCount, syncPathCount, failedSyncPathCount, end - start,
           mRootScheme);
     }
+    boolean success = syncPathCount > 0;
+    if (success) {
+      // update the sync path cache for the root of the sync
+      // TODO(gpang): Do we need special handling for failures and thread interrupts?
+      mUfsSyncPathCache.notifySyncedPath(mRootScheme.getPath().getPath(), mDescendantType);
+    }
     mStatusCache.cancelAllPrefetch();
     mSyncPathJobs.forEach(f -> f.cancel(true));
-    return syncPathCount > 0;
+    return success;
   }
 
   /**
@@ -598,7 +604,6 @@ public class InodeSyncStream {
     if (loadMetadata) {
       loadMetadataForPath(inodePath);
     }
-    mUfsSyncPathCache.notifySyncedPath(inodePath.getUri().getPath(), DescendantType.ONE);
 
     if (syncChildren) {
       // Iterate over Alluxio children and process persisted children.
--- a/tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
+++ b/tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
@@ -711,6 +711,38 @@ public class UfsSyncIntegrationTest extends BaseIntegrationTest {
     Assert.assertEquals(4, paths.size());
   }
 
+  @Test
+  public void deleteUfsFileGetStatus() throws Exception {
+    new File(ufsPath("/delete")).mkdirs();
+    writeUfsFile(ufsPath("/delete/file"), 10);
+
+    List<URIStatus> paths = mFileSystem.listStatus(new AlluxioURI(alluxioPath("/delete")),
+        ListStatusPOptions.newBuilder().setRecursive(false).setCommonOptions(PSYNC_ALWAYS).build());
+
+    Assert.assertEquals(1, paths.size());
+    Assert.assertEquals("file", paths.get(0).getName());
+
+    // delete the file and wait a bit
+    new File(ufsPath("/delete/file")).delete();
+    CommonUtils.sleepMs(2000);
+
+    // getStatus (not listStatus) on the root, with a shorter interval than the sleep.
+    // This will sync that directory. The sync interval has to be long enough for the internal
+    // syncing process to finish within that time.
+    mFileSystem.getStatus(new AlluxioURI(alluxioPath("/delete")), GetStatusPOptions.newBuilder()
+        .setCommonOptions(
+            FileSystemMasterCommonPOptions.newBuilder().setSyncIntervalMs(1000).build()).build());
+
+    // verify that the file is deleted, without syncing
+    try {
+      mFileSystem.getStatus(new AlluxioURI(alluxioPath("/delete/file")),
+          GetStatusPOptions.newBuilder().setCommonOptions(PSYNC_NEVER).build());
+      Assert.fail("the ufs deleted file is not expected to exist after sync via getStatus");
+    } catch (FileDoesNotExistException e) {
+      // expected
+    }
+  }
+
   private String ufsPath(String path) {
     return mLocalUfsPath + path;
   }


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -387,9 +387,15 @@ public boolean sync() throws AccessControlException, InvalidPathException {
           syncPathCount + failedSyncPathCount, syncPathCount, failedSyncPathCount, end - start,
           mRootScheme);
     }
+    boolean success = syncPathCount > 0;
+    if (success) {
+      // update the sync path cache for the root of the sync
+      // TODO(gpang): Do we need special handling for failures and thread interrupts?
+      mUfsSyncPathCache.notifySyncedPath(mRootScheme.getPath().getPath(), mDescendantType);
```
ZacBlanco, line 394: So now, if two RPCs on the same path come in at the same time, we'll end up doing a lot more work, right?

gpang: Yes, that is true. Is there a better way to mark a directory as "synced" only after all the descendants are synced?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -387,9 +387,15 @@ public boolean sync() throws AccessControlException, InvalidPathException {
           syncPathCount + failedSyncPathCount, syncPathCount, failedSyncPathCount, end - start,
           mRootScheme);
     }
+    boolean success = syncPathCount > 0;
```
yuzhu, line 390: slightly concerned about failure and cancellations, because we would mark the directory as synced when perhaps only one file in the directory is synced. It would also prevent it from everything in the directory getting synced for a while..

gpang: Yeah, but I think that is true even with the old code, since once the directory inode itself is synced, it is marked as synced, even before any of its children are even processed.

yuzhu: can we declare the sync successful if the failedCount == 0 and we are not canceled (nothing left in the queue?)
Would that be a better condition for synced?

gpang: Currently the `failedCount` even includes paths which did not require sync, so `failedCount` is not an accurate name.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: I think this is fine. A trade-off between memory overhead (keeping track of all inodes in the sync path cache) vs doing more work in unfortunate scenarios

ZacBlanco (approved):

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: is it possible to split that out? consider a directory containing only a directory that is already synced. Would the `syncPathCount` be 0 and the parent directory would never be considered synced?

apc999:
Gene, can you link the issue you are fixing in case there is one

yuzhu (approved):

gpang:

Issue #12136, opened 2020-09-23
fuse pod hang forever after open&read many files without close the files


chaowangnk1:
**Alluxio Version:**
latest build on master branch

**Describe the bug**
With alluxio deployed on k8s, fuse pod hang forever after open&read many files without close the files. The file operations are from separated ML training job run pods, on alluxio mounted path.
After checking stack traces of java process in fuse pod, I observed thread hang at:
```
"Thread-5" #31 daemon prio=5 os_prio=0 tid=0x00007f29d0001000 nid=0xab9c waiting on condition [0x00007f2a09a99000]
   java.lang.Thread.State: TIMED_WAITING (parking)
at sun.misc.Unsafe.park(Native Method)
- parking to wait for  <0x000000072be23138> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2163)
at alluxio.resource.DynamicResourcePool.acquire(DynamicResourcePool.java:352)
at alluxio.resource.DynamicResourcePool.acquire(DynamicResourcePool.java:295)
at alluxio.client.file.FileSystemContext.acquireBlockWorkerClientInternal(FileSystemContext.java:541)
at alluxio.client.file.FileSystemContext.acquireBlockWorkerClient(FileSystemContext.java:521)
at alluxio.client.block.stream.GrpcDataReader.<init>(GrpcDataReader.java:83)
at alluxio.client.block.stream.GrpcDataReader.<init>(GrpcDataReader.java:49)
at alluxio.client.block.stream.GrpcDataReader$Factory.create(GrpcDataReader.java:203)
at alluxio.client.block.stream.BlockInStream.readChunk(BlockInStream.java:396)
at alluxio.client.block.stream.BlockInStream.readInternal(BlockInStream.java:274)
at alluxio.client.block.stream.BlockInStream.read(BlockInStream.java:264)
at alluxio.client.file.AlluxioFileInStream.read(AlluxioFileInStream.java:188)
at alluxio.fuse.AlluxioJniFuseFileSystem.readInternal(AlluxioJniFuseFileSystem.java:336)
at alluxio.fuse.AlluxioJniFuseFileSystem.lambda$read$4(AlluxioJniFuseFileSystem.java:305)
[...]

PR #12139, opened 2020-09-23
Fix deadlock of racing worker client acquisition and file completion

chaowangnk1:
Fix bug: fuse pod hang forever if open&read many files without close  the files

Root cause:
      `mBlockInStream` in `AlluxioFileInStream` won't be released even reading to end of the block.
      In this case, `BlockWorkerClient` will be held in `BlockInStream`. Given many files opened, `BlockWorkerClient` in the pool will be exhausted and the following operations will be blocked.
Fix:
      Release `mBlockInStream` after reading to end of the block.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
@@ -147,6 +147,10 @@ public class AlluxioFileInStream extends FileInStream {
         if (result != -1) {
           mPosition++;
         }
+        if (mBlockInStream.remaining() == 0) {
+          closeBlockInStream(mBlockInStream);
+          mBlockInStream = null;
+        }
         return result;
       } catch (IOException e) {
         lastException = e;
@@ -191,6 +195,10 @@ public class AlluxioFileInStream extends FileInStream {
         }
         retry = mRetryPolicySupplier.get();
         lastException = null;
+        if (mBlockInStream.remaining() == 0) {
+          closeBlockInStream(mBlockInStream);
+          mBlockInStream = null;
+        }
       } catch (IOException e) {
         lastException = e;
         if (mBlockInStream != null) {
--- a/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
@@ -279,6 +279,10 @@ public class AlluxioFileInStream extends FileInStream {
         if (mCachedPositionedReadStream.getSource() != BlockInStream.BlockInStreamSource.LOCAL) {
           triggerAsyncCaching(mCachedPositionedReadStream);
         }
+        if (bytesRead == mBlockSize - offset) {
+          mCachedPositionedReadStream.close();
+          mCachedPositionedReadStream = null;
+        }
       } catch (IOException e) {
         lastException = e;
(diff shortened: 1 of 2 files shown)


chaowangnk1:
This is to fix issue: https://github.com/Alluxio/alluxio/issues/12136

apc999:
@calvinjia  @madanadit, this fix from  @chaowangnk1 addresses a particular issue in ML workloads:
- in FUSE integration, we don't close the `FileInStream` (and thus will keep `blockWorkerClient` resource)  until FUSE API `release` is called.
- For ML workloads, a large number of concurrent reads may happen at the same time. Even they finish reads, before `release` is called, `FileInStream` is not closed and `blockWorkerClient` resource is kept.

I think @chaowangnk1 's fix definitely makes sense for ML workloads. @calvinjia @madanadit  do you see any possible downsides for other workloads by closing `BlockInStream` a bit earlier (on reading to the block boundary, rather than exceeding block boundary)? I am for this fix but want to get your opinions.

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
```diff
@@ -147,6 +147,10 @@ public int read() throws IOException {
         if (result != -1) {
           mPosition++;
         }
+        if (mBlockInStream.remaining() == 0) {
+          closeBlockInStream(mBlockInStream);
+          mBlockInStream = null;
```
calvinjia, line 152: The setting of `mBlockInStream = null` is handled in `closeBlockInStream`.

calvinjia (changes requested):
@chaowangnk1 Thanks for the improvement! Could you also apply this logic to `mCachedPositionedReadStream` and add a unit test?

apc999:
@calvinjia  are we changing a bit of the way alluxio client is running? my understanding is previously we close blockInStream when moving outside of a current block, now if remaining == 0 we will also close this blockInstream. right? Just to make sure if you good with this possible change in terms of performance and resource

calvinjia:
@apc999 the new behavior should be fine. The read patterns which would be impacted are those which read to the end of a block and then seek to an earlier part of the block.

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
chaowangnk1: removed this line.

calvinjia (changes requested):
@chaowangnk1 Thanks for the update. Do you mind making a new UT instead of updating an existing one? The test can be the same, I would prefer having some repetitive tests (these all execute very quickly) and have it be clear what each test is asserting.

chaowangnk1:
@calvinjia I added UT for the code change and applied the logic to mCachedPositionedReadStream. I also noticed that a new DataReaderFactory and DataReader is provisioned in BlockInStream.positionedRead on demand. I'm wondering if we could apply the same logic to regular read()?

calvinjia:
@chaowangnk1 we try to avoid creating the data reader on each read because there is a large overhead.

apc999 (approved):

calvinjia (approved):

calvinjia:

calvinjia:
@chaowangnk1 Thank you for optimizing the client!

apc999:

apc999:
alluxio-bot, cherry-pick this to branch-2.3-fuse please

apc999:

PR #12144, opened 2020-09-24
Downgrade swagger-maven-plugin from 3.1.7 back to 3.1.1

bradyoo:
This broke job and worker REST API docs

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1018,7 +1018,7 @@
         <plugin>
           <groupId>com.github.kongchen</groupId>
           <artifactId>swagger-maven-plugin</artifactId>
-          <version>3.1.7</version>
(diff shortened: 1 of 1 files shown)


ZacBlanco (approved):
Have you investigated why they don't work? And can you provide a description of what is broken? Also, this is just for the workers - and not for the masters, right?

bradyoo:
I don't really know why they don't work but the rest doc pages look like this. https://docs.alluxio.io/os/restdoc/edge/proxy/index.html

The proxy and job ones are broken so I doubt it has much to do with workers vs masters.

bradyoo:
Also, can you merge it for me?

ZacBlanco:

PR #12147, opened 2020-09-26
Add GitHub actions

bradyoo:


bradyoo:
jenkins, test this please.

bradyoo:

bradyoo:
@gpang Want to make this official?

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,58 @@
...
+name: Checkstyle, Findbugs, Doc Check, etc.
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  pull_request:
+    branches: [ github_actions, master ]
+
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
Xenorith, line 12: where are these secret strings set?

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,58 @@
...
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      # Runs a single command using the runners shell
+      - name: Run a one-line script
+        run: |
+          mkdir -p ~/.m2
+          ALLUXIO_DOCKER_NO_TTY=true ALLUXIO_DOCKER_GIT_CLEAN=true ALLUXIO_DOCKER_MVN_RUNTOEND=true \
```
Xenorith, line 46: nit: split each env var into its own line for readability and same throughout similar files

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: artifact
+          path: |
+            **/target/artifacts/*
+            **/target/logs/*
+            log*
```
Xenorith, line 58: newline at eof and many other new files

.github/workflows/integrationtests.yml
```diff
@@ -0,0 +1,81 @@
+# This is a basic workflow to help you get started with Actions
```
Xenorith, line 1: rename to file integration_tests.yml or whatever is allowed as long as there's some kind of separation between words?

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,58 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Checkstyle, Findbugs, Doc Check, etc.
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  pull_request:
+    branches: [ github_actions, master ]
```
Xenorith, line 9: are wildcards allowed for branches? we will need these to run on release branches (ex. `branch-2.x`) as well

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,69 @@
+# This is a basic workflow to help you get started with Actions
```
Xenorith, line 1: ditto for file name here

dev/github/run_tests.sh
```diff
@@ -0,0 +1,100 @@
...
+  git clean -fdx
+fi
+
+mvn_args=""
+if [ -n "${ALLUXIO_MVN_RUNTOEND}" ]
+then
+  mvn_args+=" -fn -DfailIfNoTests=false --fail-at-end"
+fi
+
+RUN_MAVEN="false"
```
Xenorith, line 29: this is a copy of the bash code in `run_checks.sh`; can they be consolidated?

Xenorith (changes requested):
can you leave a README in `dev/github/` explaining the new layout of files and describe how github actions invokes the build? it's not very straightforward even as someone who knows how this currently works with jenkins

dev/github/run_tests.sh
bradyoo: I don't intend to consolidate them. It's very very annoying to consolidate shell code in a reasonable way

bradyoo: though I can probably delete a lot of things here so gonna do that first.

Xenorith (commented):
it seems to me that you also removed the logic for "run only the necessary checks/tests based on the PR diff". can each build do its own diff and exits quickly if the diff is irrelevant to what the build is testing for?

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,56 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Checkstyle, Findbugs, Doc Check, etc.
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  pull_request:
+    branches: [ github_actions, master, branch-2.* ]
```
Xenorith, line 9: i was a bit lax in my previous comment. we need to run PR builds for the release branches but only starting from 2.4. i presume none of this would work for 2.3 or earlier, unless you think the full set of changes to make this work can be backported

bradyoo:
@Xenorith diffing was really never really a thing.

.github/workflows/checkstyle.yml
bradyoo: Since the yml files don't exist in the branches themselves, it won't actually run them. For example, note that this code is in the github_actions branch in alluxio/alluxio but it's not actually running in master branch for others right now!

Xenorith: good point. so why is this line here at all?

bradyoo:

bradyoo:
@gpang I got through the review from @Xenorith, would you like to review?

.github/workflows/checkstyle.yml
Xenorith: ping

Xenorith (changes requested):
and readme on how github actions interacts with these files?

dev/github/run_tests.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Set things up so that the current user has a real name and can authenticate.
+myuid=$(id -u)
+mygid=$(id -g)
+echo "$myuid:x:$myuid:$mygid:anonymous uid:/home/jenkins:/bin/false" >> /etc/passwd
+
+export MAVEN_OPTS="-Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS"
+
+mvn -Duser.home=/home/jenkins test -Pdeveloper -Dmaven.main.skip -Dskip.protoc=true  -Dmaven.javadoc.skip -Dlicense.skip=true \
+-Dcheckstyle.skip=true -Dfindbugs.skip=true -Dsurefire.forkCount=2 ${mvn_args} $@
```
gpang, line 34: why are checkstyle and findbugs disabled in the PR builds?

gpang (commented):
@bradyoo Thanks, I left a small question.

dev/github/run_tests.sh
bradyoo: It is not it is under run_checks.sh checkstyle.yml in .github/ goes through a separate codepath that does builds, checkstyles, etc. note that you can see that category separately in the checks below.

bradyoo: Checkstyle, Findbugs, Doc Check, etc. / build (pull_request)

.github/workflows/checkstyle.yml
bradyoo: because the line is still required for it to run. Even with this code being checked in, without this line of branch-2.* it will not run on it.

dev/github/run_tests.sh
gpang: Ahhh, I see. Thanks!

dev/github/helper.sh
```diff
@@ -0,0 +1,39 @@
...
+fi
+
+export MAVEN_OPTS="-Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss.SSS"
+
+# Always use java 8 to compile the source code
+JAVA_HOME_BACKUP=${JAVA_HOME}
+PATH_BACKUP=${PATH}
+JAVA_HOME=/usr/local/openjdk-8
+PATH=$JAVA_HOME/bin:$PATH
+mvn -Duser.home=/home/jenkins -T 4C clean install -Pdeveloper -Dfindbugs.skip -Dcheckstyle.skip -DskipTests -Dmaven.javadoc.skip \
```
gpang, line 38: This is for the checkstyle and findbugs task, right? This is also skipping checkstyle, findbugs, etc. Is this correct? I am a bit confused.

This is an output for this PR's checkstyle check. https://pipelines.actions.githubusercontent.com/110HvfICxw4TXSOnsd72jk32naeMsNgZF0xPcAxH5PwqdiBHUN/_apis/pipelines/1/runs/106/signedlogcontent/3?urlExpires=2020-09-30T17%3A18%3A37.2939398Z&urlSigningMethod=HMACV1&urlSignature=y%2B13RULZz8azm5yA9KZIWAnIzZTiBsg0JVWduMCxaDQ%3D 

However, in that output, I see
```
2020-09-30T08:20:56.6182722Z + mvn -Duser.home=/home/jenkins -T 4C clean install -Pdeveloper -Dfindbugs.skip -Dcheckstyle.skip -DskipTests -Dmaven.javadoc.skip -Dlicense.skip -Dsurefire.forkCount=2 -fn -DfailIfNoTests=false --fail-at-end
```

This looks like all the checks are skipped?

gpang (commented):
@bradyoo I had additional questions about the checkstyle checks

dev/github/helper.sh
bradyoo: Oh wow you are right. I accidentally dropped it while doing the refactor during the review process. I ended up undoing the refactor and moved helper.sh into the respective callers and modified the findbug/checkstyle skip parts.

bradyoo:
alluxio-bot, test this please

dev/github/helper.sh
bradyoo: Fixed now.

gpang: cool, thanks! The mvn command looks correct now.

bradyoo:
Tests passed.

gpang (approved):
LGTM

Thanks for this!

Xenorith (approved):

bradyoo:
Feel free to merge. @gpang

Xenorith:
actually i think we need to coordinate when this merges along with other associated changes, so let's hold off for a bit @gpang

bradyoo:
Added retries into the github_action itself so that users don't have to do much retries.

Xenorith:

PR #12149, opened 2020-09-27
Add worker master periodical connections timeout

LuQQiu:
In a cluster with multiple masters using embedded journal, when network partitioned the original leading master, leadership changes to a new leading master. The workers hang in GRPC connections with the previous leading master and don't register with the new leading master.

This PR adds the timeout to periodical connections (operations) between workers and leading masters to prevent workers from hanging forever.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2659,7 +2659,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey WORKER_MASTER_PERIODICAL_RPC_TIMEOUT =
       new Builder(Name.WORKER_MASTER_PERIODICAL_RPC_TIMEOUT)
-          .setDefaultValue("1min")
+          .setDefaultValue("5min")
           .setDescription("Timeout for periodical RPC between workers "
               + "and the leading master. This property is added to prevent workers "
               + "from hanging in periodical RPCs with previous leading master "
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2651,6 +2651,16 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
       .setScope(Scope.WORKER)
       .build();
+  public static final PropertyKey WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT =
+      new Builder(Name.WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT)
+          .setDefaultValue("1min")
+          .setDescription("Timeout for periodical connections between workers "
+              + "and the leading master. Periodical connections between workers "
+              + "and previous leading master may hang which prevents workers "
+              + "from registering with the new leading master "
+              + "during flaky network situations.")
+          .setScope(Scope.WORKER)
+          .build();
   public static final PropertyKey WORKER_MASTER_CONNECT_RETRY_TIMEOUT =
       new Builder(Name.WORKER_MASTER_CONNECT_RETRY_TIMEOUT)
           .setDescription("Retry period before workers give up on connecting to master and exit.")
@@ -5149,6 +5159,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     public static final String WORKER_FREE_SPACE_TIMEOUT = "alluxio.worker.free.space.timeout";
     public static final String WORKER_HOSTNAME = "alluxio.worker.hostname";
     public static final String WORKER_KEYTAB_FILE = "alluxio.worker.keytab.file";
+    public static final String WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT =
+        "alluxio.worker.master.periodical.connect.timeout";
     public static final String WORKER_MASTER_CONNECT_RETRY_TIMEOUT =
         "alluxio.worker.master.connect.retry.timeout";
     public static final String WORKER_MEMORY_SIZE = "alluxio.worker.memory.size";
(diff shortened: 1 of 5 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2651,6 +2651,16 @@ public String toString() {
       .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
       .setScope(Scope.WORKER)
       .build();
+  public static final PropertyKey WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT =
```
bf8086, line 2654: It can be somehow confusing to have both `WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT` and `WORKER_MASTER_CONNECT_RETRY_TIMEOUT` in the configuration. Can we have a more distinguishable name for this one? How about `WORKER_MASTER_PERIODICAL_RPC_TIMEOUT`?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2651,6 +2651,16 @@ public String toString() {
...
+      new Builder(Name.WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT)
```
bf8086, line 2655: (nit) Please sort the property name by alphabetical order.

core/common/src/main/java/alluxio/util/network/NetworkAddressUtils.java
```diff
@@ -670,7 +671,9 @@ public static void pingService(InetSocketAddress address, alluxio.grpc.ServiceTy
         .build();
     try {
       ServiceVersionClientServiceGrpc.ServiceVersionClientServiceBlockingStub versionClient =
-          ServiceVersionClientServiceGrpc.newBlockingStub(channel);
+          ServiceVersionClientServiceGrpc.newBlockingStub(channel)
+              .withDeadlineAfter(conf.getMs(PropertyKey.WORKER_MASTER_PERIODICAL_CONNECT_TIMEOUT),
```
bf8086, line 675: `pingService` is used in a bunch of places including client and job services. We probably want to use a different configuration property. Can we maybe reuse the `USER_MASTER_POLLING_TIMEOUT` or something more generic?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2651,6 +2651,16 @@ public String toString() {
...
+          .setDefaultValue("5min")
+          .setDescription("Timeout for periodical connections between workers "
+              + "and the leading master. Periodical connections between workers "
+              + "and previous leading master may hang which prevents workers "
+              + "from registering with the new leading master "
+              + "during flaky network situations.")
```
bf8086, line 2661: Can we explain in description how the value of the property can affect the delay before a worker attempt to register with a new primary master?

bf8086 (changes requested):
Thanks! I added some comments.

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Sounds good

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2657,6 +2657,18 @@ public String toString() {
           .setDefaultValue("1hour")
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_MASTER_PERIODICAL_RPC_TIMEOUT =
+      new Builder(Name.WORKER_MASTER_PERIODICAL_RPC_TIMEOUT)
+          .setDefaultValue("5min")
```
gpang, line 2662: Does this control the connection phase, or the entire RPC duration as well?

Also, is 5m too long? This means if there is a failover, workers will take 5 minutes to re-connect?

How did failovers work before this parameter?

gpang (commented):
@LuQQiu I left a few questions

bf8086 (approved):
LGTM.

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: This only controls the RPC duration of selected periodical worker connections. Control the entire RPC duration.
The worker hangs problem only exists if the previous leading master is network partitioned (not responding, instead of rejection). 
I give a long timeout in the assumption that this case rarely happens and try not to affect our normal RPCs. A frequently called RPC is getPinList(). If the cluster has a large amount of files/dirs get pinned, i worried that if the timeout is short, the getPinList() may keep getting timeout unexpectedly.

LuQQiu: Before this parameter, if the previous leading master is network partitioned (not responding), our workers will hang in periodical RPCs and fail to response to any requests or register with the new leading master. After the previous leading master come back, workers will receive exception message from previous leading master and continue functioning (register with new leading master)

When network condition is good, the rpcs should direct fail when connecting to previous leading master and no hanging should happen.

LuQQiu: @bf8086 What's the timeout that you think is reasonable. I saw we getPinList every second, probably 30sec or 1min timeout is good for this case?

gpang: @LuQQiu 
I see. Let's say we have a 1m timeout, and the worker heartbeat takes longer than 1m, even though nothing has happened to the master (no failover, no network flakiness). What will happen to the worker? Will it just try to reconnect again?

If the worker heartbeat always takes more than 1m, does that mean that worker will never be able to successfully perform a heartbeat?

LuQQiu: Yeah for both questions, it will create a new client and retry operations. If the operations consistent taking longer than 1m, then heartbeat will never succeed. That's why i prefer to give the timeout a larger value.
This problem exists in previous versions but at least never be caught by us. It could be the network partition happens rarely and can be recovered in a short period of time. This change mainly protects the RPC from hanging forever.

gpang: @LuQQiu I see. Then, maybe we do want to make the timeout longer.

From an earlier comment, you said typically, when a master does a failover, the worker will immediately know, because the old master would actively reject the connection? Is this correct? That means this timeout will not affect normal failovers, but only unreachable old masters? If this is the case, then this is more reason to make the timeout long.

Thanks for the explanations.

LuQQiu: Yeah, normally the worker will immediately know it cannot connect to the old leading master and connect to the new leading master. This timeout will not affect normal failover and only be useful when old leading master is unreachable.

gpang (approved):
LGTM (after tests pass)

bf8086:

PR #12150, opened 2020-09-27
Make it more restrictive when a snapshot can be taken

bf8086:
Sometimes a snapshot is ordered when the journal log is not consistent with the state machine. This change makes it more restrictive that snapshots can only be taken when a follower is in running state.


bf8086:
@gpang @LuQQiu PTAL.

LuQQiu (approved):

gpang (approved):
LGTM

bf8086:

PR #12151, opened 2020-09-27
Make Ratis client retry on catchup failures

bf8086:
Sometimes a Ratis client decides to timeout and close connection when master tries to gain primacy. This change makes the master to notify Ratis client to reconnect when an exception is throw, so we don't just keep retrying on a closed client.


bf8086:
@gpang @LuQQiu PTAL.

LuQQiu (approved):

gpang (approved):
LGTM

bf8086:

PR #12152, opened 2020-09-28
Make master try harder when joininig a quorum

bf8086:
Sometimes a master attempts to join a quorum during leader election. This change extends the retry period so the retry can extends pass an election. Also lower the log severity given its best effort nature.


bf8086:
@gpang @LuQQiu PTAL.

gpang (approved):
LGTM

bf8086:

PR #12166, opened 2020-09-30
Fix deadlock when raft journal throws exception

bf8086:
In an unlucky situation when raft server throw an exception while losing primacy, the master will attempt to shutdown but stuck in a deadlock of `RaftJournalSystem::losePrimacy -> System::exit -> ProcessUtils::stopProcessOnShutdown -> RaftJournalSystem::stop`. It can be fixed with throwing directly within `losePrimacy`.


bf8086:
@gpang @LuQQiu PTAL.

LuQQiu (approved):

bf8086:

PR #12135, opened 2020-09-23
Fix cp with wildcard and special characters

LuQQiu:
Previously if we try to use wildcard to copy files in path with special characters like '.', '+', '^', "$', '*', the cp will fail.
This test fix the escape function to escape special characters used in regrex pattern matching.

PR #12172, opened 2020-10-01
Port changes of [#12135] to branch-2.2

LuQQiu:
Previously if we try to use wildcard to copy files in path with special characters like '.', '+', '^', "$', '*', the cp will fail.
This test fix the escape function to escape special characters used in regrex pattern matching.

[This is an auto-generated PR to cherry-pick committed PR #12135 into target branch branch-2.2]


ns1123:
jenkins, retest this please

LuQQiu:

PR #12174, opened 2020-10-05
Implement a local first raft client for improved journal performance

bf8086:
Current the raft client for writing journal entries will always checking for leadership and send the request through network interface, introducing significant overhead. This change implements a raft client which attempts sending messages directly to local server first, and falls back to the default client if the local server is no longer a leader.

The client also has built-in retry function for raft client which is closed due to timeout.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
@@ -1,0 +1,128 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.master.journal.raft;
+
+import alluxio.util.LogUtils;
+
+import org.apache.ratis.client.RaftClient;
+import org.apache.ratis.protocol.AlreadyClosedException;
+import org.apache.ratis.protocol.ClientId;
+import org.apache.ratis.protocol.Message;
+import org.apache.ratis.protocol.NotLeaderException;
+import org.apache.ratis.protocol.RaftClientReply;
+import org.apache.ratis.protocol.RaftClientRequest;
+import org.apache.ratis.server.RaftServer;
+import org.apache.ratis.util.TimeDuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+
+/**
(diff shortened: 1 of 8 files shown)


bf8086:
@gpang @LuQQiu PTAL.

bradyoo:
Hey, I noticed that this code is not rebased/merged with master and since alluxio PR check is disabled, this is not running any tests. Can you do a rebase/merge so that tests are running?

core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
```diff
@@ -0,0 +1,128 @@
...
+    if (mClient == null) {
+      return sendLocalRequest(message, timeout);
+    } else {
+      return sendRemoteRequest(message);
+    }
+  }
+
+  private CompletableFuture<RaftClientReply> sendLocalRequest(Message message,
+      TimeDuration timeout) throws IOException {
+    return mServer.submitClientRequestAsync(
```
LuQQiu, line 76: How long and how frequently will the client retry? is it configurable?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -338,6 +338,7 @@ private RaftClient createClient() {
         .build();
     return RaftClient.newBuilder()
         .setRaftGroup(mRaftGroup)
+        .setClientId(mClientId)
```
LuQQiu, line 341: Why this is needed?

LuQQiu (commented):
Thanks feng! Overall looks like, left some questions

core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
bf8086: This class will rebuild a client every time a client throws an `AlreadyClosedException`. The actual retry logic for sending a message is implemented by the caller and is not changed in this PR.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: It's for easier tracking so requests from same node will have the same client id.

LuQQiu (approved):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -525,14 +528,20 @@ private void catchUp(JournalStateMachine stateMachine, RaftClient client)
...
+            TimeDuration.valueOf(5, TimeUnit.SECONDS));
+        RaftClientReply reply = future.get(5, TimeUnit.SECONDS);
+        ex = reply.getException();
+      } catch (TimeoutException | ExecutionException | IOException e) {
+        ex = e;
+      }
+      if (ex != null) {
+        LOG.info("Exception submitting term start entry: {}", ex.toString());
+        // avoid excessive retries when server is not ready
+        Thread.sleep(100);
```
gpang, line 544: Should this be configurable? I can't tell if 100ms is too short or too long. This means we will retry 10x per second forever? That seems quite frequent.

gpang (commented):
@bf8086 just a minor comment.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Yeah it is still significantly better than retrying immediately and blowing up the logs in no time. I added a configuration property for 1s.

bf8086:
@gpang Thanks, I updated the PR. PTAL.

gpang (approved):
LGTM

bf8086:

PR #12179, opened 2020-10-05
Avoid reloading old snapshots

bf8086:
Ratis leader sometimes send a request to a follower to install a snapshot that is older than the follower's log. This change implements a check to avoid reloading the state machine after such snapshots are downloaded to prevent some accounting issue.


LuQQiu (approved):

gpang (approved):
LGTM

bf8086:

PR #12200, opened 2020-10-08
Remove java stack trace log to fix list miss file status perf

HelloHorizon:
Logging java stack trace hurt the performance of listing a large number of miss files, remove java stack trace to fix the performance issue.

### DIFF
--- a/core/server/common/src/main/java/alluxio/RpcUtils.java
+++ b/core/server/common/src/main/java/alluxio/RpcUtils.java
@@ -119,7 +119,7 @@ public final class RpcUtils {
         MetricsSystem.counter(getQualifiedFailureMetricName(methodName)).inc();
         if (!logger.isDebugEnabled()) {
           logger.warn("Exit (Error): {}: {}, Error={}", methodName,
-              String.format(description, args), e);
+              String.format(description, args), e.getMessage());
         }
       }
       throw AlluxioStatusException.fromAlluxioException(e).toGrpcStatusException();


core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -119,7 +119,7 @@ private RpcUtils() {} // prevent instantiation
         MetricsSystem.counter(getQualifiedFailureMetricName(methodName)).inc();
         if (!logger.isDebugEnabled()) {
           logger.warn("Exit (Error): {}: {}, Error={}", methodName,
-              String.format(description, args), e);
+              String.format(description, args), e.getMessage());
```
LuQQiu, line 122: @gpang Looks like they change from printing only the message to print all the stack trace in logger.warn.
Can you help confirm that @HelloHorizon? 
Maybe we need to change other warn loggings in the whole system as well.

HelloHorizon: @LuQQiu They print the stack trace in 2.x as follows:
```
warn(CharSequence message, Throwable t)
         Logs a CharSequence at the WARN level including the stack trace of the Throwable t passed as parameter.`
And for 1.x as follows:
```
warn(Object message, Throwable t)
          Log a message with the WARN level including the stack trace of the Throwable t passed as parameter.
```
Both of them throw the stack trace?

gpang: In the logs themselves, do the stacks shows up for both versions?

gpang:
@HelloHorizon Can you verify if the amount of logging is different or the same between the 2 versions? So, we want to understand if we are logging more, or if we are logging the same amount but slower?

The concern with not printing the stack is when there really is a failure, the stack is very helpful in determining what the issue was. With only the message, it may be very difficult to understand what happened.

core/server/common/src/main/java/alluxio/RpcUtils.java
HelloHorizon: The stacks are not showing up in the previous version but for the latest version.

HelloHorizon:
@gpang Two versions have a close amount of loggings **before removing java stack trace - (Old 23,309,217 VS new 13,504,558 **new one reach the limitation of number of master logs), after removing java stack trace - (Old 23,309,233 VS New 23,309,184)** between the 2 versions.

And before we remove the java stack trace, the total size of logging is **before removing java stack trace - (Old 865.4 VS New 1.05GB)**. After we remove the stack trace, the total size of logging is **after removing java stack trace - (Old 865.5MB VS New 751.6MB)**.

So we can estimate that the new version of Alluxio produces log files with 2GB space, which has 1.5 more logs in space compares to the old version Alluxio, in the list miss file status test. So I think the performance regression is because we logging the same amount of logs but with more space and more content.

gpang (approved):
LGTM

core/server/common/src/main/java/alluxio/RpcUtils.java
rastogiasr: @gpang - based on the comment from @HelloHorizon , this should be good to go - please confirm.

LuQQiu:

LuQQiu:

PR #12214, opened 2020-10-09
Fix backup coordination and error handling

bf8086:
This change fixed a few issues related to backup:

- Make backup interrupt correctly when raft journal needs to reload from a snapshot. Currently if backup occurs right before a journal reload it can buffer up some entries and then corrupt journal by resuming after the reload is completed. This change added an interrupt callback to make sure backup is correctly canceled and journal is resumed before a reload happens.

- Block backup when journal is being reloaded. When a raft journal is being reloaded, the backup should not happen because it will capture inconsistent state while the master metadata is re-applied from snapshot.

- Cancel backup sub-tasks properly. Currently when a backup is cancelled due to error or interruption, the sub-tasks do not explicitly respond to interrupt signals, leaving them running in background while opening backup service for more requests. This causes increased memory consumption when multiple backup tasks are running concurrently.

- Incomplete backups. There is currently no way to tell if a backup file is incomplete before restoring it to the cluster. This change added a marker file after the backup is done to indicate the completeness.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -295,7 +295,7 @@ public class JournalStateMachine extends BaseStateMachine {
    * Unpause the StateMachine. This should be done after uploading new state to the StateMachine.
    */
   public synchronized void unpause() {
-    LOG.info("Unpausiing raft state machine.");
+    LOG.info("Unpausing raft state machine.");
     getLifeCycle().startAndTransition(() -> {
       if (mJournalApplier.isSuspended()) {
         LOG.warn("Journal should not be suspended while state machine is paused.");
@@ -515,12 +515,8 @@ public class JournalStateMachine extends BaseStateMachine {
   public synchronized void resume() throws IOException {
     LOG.info("Resuming raft state machine");
     mInterruptCallback = null;
-    if (mJournalApplier.isSuspended()) {
-      mJournalApplier.resume();
-      LOG.info("Raft state machine resumed");
-    } else {
-      LOG.warn("Raft state machine is already resumed");
-    }
+    mJournalApplier.resume();
+    LOG.info("Raft state machine resumed");
   }
 
   /**
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -144,7 +144,7 @@ public class JournalStateMachine extends BaseStateMachine {
     unpause();
   }
 
-  private void loadSnapshot(SingleFileSnapshotInfo snapshot) throws IOException {
+  private synchronized void loadSnapshot(SingleFileSnapshotInfo snapshot) throws IOException {
     if (snapshot == null) {
       LOG.info("No snapshot to load");
       return;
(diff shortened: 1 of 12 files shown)


bf8086:
@gpang @LuQQiu PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -98,6 +99,7 @@
   private volatile long mNextSequenceNumberToRead = 0;
   private volatile boolean mSnapshotting = false;
   private volatile boolean mIsLeader = false;
+  private volatile Runnable mInterruptCallback;
```
gpang, line 102: I am confused as to what this is supposed to be. It is called "interrupt callback", but it is set in `suspend` and used in `pause`. I don't understand the flow of interaction. Could you explain what this is?

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
```diff
@@ -186,6 +186,8 @@ protected AlluxioURI takeBackup(BackupPRequest request, AtomicLong entryCounter)
           // Create the backup from master state.
           mBackupManager.backup(ufsStream, entryCounter);
         }
+        // Add a marker file indicating the file is completed.
+        ufs.create(backupFilePath + ".complete").close();
```
gpang, line 190: Does anything look for this file?

gpang (changes requested):
@bf8086 I left a few comments

core/server/master/src/main/java/alluxio/master/backup/AbstractBackupRole.java
bf8086: No it is not checked programatically.
Ideally we should check this when doing a restore, but that can break backward compatibility with backups taken by older versions.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
bf8086: The callback is set in `suspend()` by backup worker before taking a backup, and invoked in `pause()` by the state machine before reloading the states from a snapshot.

The interrupt callback is used to inform the caller of the `suspend()` method that the suspension is interrupted and the metadata states are about to be changed. This is needed because the follower cannot wait for the long running backup process to complete and resume the state machine. It has to reload immediately to avoid further downtime. 

The overall flow is:

1. Backup worker suspends state machine before backup, passing in the callback.
2. Backup worker writes journal entries to UFS
3. Raft state machine downloads a new snapshot from leader.
4. Raft state machine transitions to PAUSE state and invokes the callback.
5. Backup worker handles the callback and interrupts the backup tasks.
6. Raft state machine starts reloading the states.
7. Raft state machine finished the reload and transitions back to RUNNING state.

bf8086:
Thanks @gpang! I updated the PR and replied to your comments. PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
gpang: I see. Can you add some of this in the comments?

Also, is there a scenario where the new snapshots always interrupt backups, thus never taking a backup?

bf8086: I added a few comments to explain the workflow.

The interruption only happens after a follower installs a snapshot from master due to falling behind the earliest journal entry available. This only occur when a new master joins, or a master has not been in contact with the quorum due to downtime or outage. It will not occur under common situation.

gpang: ok

bf8086:
@gpang I updated the PR to address your comment. PTAL.

gpang (approved):
LGTM (after tests pass)

bf8086:

bf8086:

bf8086:

PR #12269, opened 2020-10-13
Fix backup worker crashing master on interruption

bf8086:
When a backup is interrupted, there is a chance that the resume callback cannot be canceled and thus happens after the server is already resumed by the journal reload process. While it makes sense for backup to crash the master if resume fails to complete and master cannot accept new journal entries, it is completely fine for master to continue if backup tries to resume an already resumed journal. This change updates the journal to not throw if someone attempts to resume after the journal is already resumed.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -537,8 +537,12 @@ public class JournalStateMachine extends BaseStateMachine {
   public synchronized void resume() throws IOException {
     LOG.info("Resuming raft state machine");
     mInterruptCallback = null;
-    mJournalApplier.resume();
-    LOG.info("Raft state machine resumed");
+    if (mJournalApplier.isSuspended()) {
+      mJournalApplier.resume();
+      LOG.info("Raft state machine resumed");
+    } else {
+      LOG.info("Raft state machine is already resumed");
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


bf8086:
@gpang @LuQQiu PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -537,8 +537,12 @@ public synchronized void suspend(Runnable interruptCallback) throws IOException
   public synchronized void resume() throws IOException {
     LOG.info("Resuming raft state machine");
     mInterruptCallback = null;
-    mJournalApplier.resume();
-    LOG.info("Raft state machine resumed");
+    if (mJournalApplier.isSuspended()) {
+      mJournalApplier.resume();
```
LuQQiu, line 541: @bf8086 Can you provide a timeline for the double resume case?

bf8086: Basically the backup starts and suspends the journal, then the journal resumes, interrupts the backup and reloads from the snapshot, and then the backup attempts to resume the journal again and hit this issue.

LuQQiu: Got it, thanks!

LuQQiu (approved):
LGTM, thanks!

LuQQiu:
alluxio-bot, test this please

LuQQiu:
No need to rerun tests since the integration tests passed for twice and only failed for the third run

bf8086:

bf8086:

Issue #12338, opened 2020-10-17
Maven fails to build after ozone ufs upgrade to 1.0.0


Xenorith:
**Alluxio Version:**
2.4.0

**Describe the bug**
After merging the PR to bump ozone ufs version (https://github.com/Alluxio/alluxio/pull/12226), the following maven command hangs:
`mvn -T 4C clean install -Dmaven.javadoc.skip=true -DskipTests -Dlicense.skip=true -Dcheckstyle.skip=true -Dfindbugs.skip=true -P'\''!ufs-hadoop-1.2,!ufs-hadoop-2.2,!ufs-hadoop-2.6'\''`

When I run jstack on the process, the one running thread approximately has the following stacktrace:
```
"BuilderThread 17" #60 prio=5 os_prio=0 tid=0x00007fcea0106800 nid=0xbaa runnable [0x00007fcdb9c2a000]
   java.lang.Thread.State: RUNNABLE
        at org.jdom2.Element.isAncestor(Element.java:1052)
        at org.jdom2.ContentList.checkPreConditions(ContentList.java:222)
        at org.jdom2.ContentList.add(ContentList.java:244)
        at org.jdom2.Element.addContent(Element.java:950)
        at org.apache.maven.plugins.shade.pom.MavenJDOMWriter.insertAtPreferredLocation(MavenJDOMWriter.java:292)
        at org.apache.maven.plugins.shade.pom.MavenJDOMWriter.iterateExclusion(MavenJDOMWriter.java:488)
        at org.apache.maven.plugins.shade.pom.MavenJDOMWriter.updateDependency(MavenJDOMWriter.java:1335)
        at org.apache.maven.plugins.shade.pom.MavenJDOMWriter.iterateDependency(MavenJDOMWriter.java:386)
        at org.apache.maven.plugins.shade.pom.MavenJDOMWriter.updateModel(MavenJDOMWriter.java:1623)
        at org.apache.maven.plugins.shade.pom.MavenJDOMWriter.write(MavenJDOMWriter.java:2156)
        at org.apache.maven.plugins.shade.pom.PomWriter.write(PomWriter.java:75)
        at org.apache.maven.plugins.shade.mojo.ShadeMojo.rewriteDependencyReducedPomIfWeHaveReduction(ShadeMojo.java:1049)
        at org.apache.maven.plugins.shade.mojo.ShadeMojo.createDependencyReducedPom(ShadeMojo.java:978)
[...]

PR #12367, opened 2020-10-22
Add a releaseProfile to generate dependency reduced pom

maobaolong:
#12338

This is a workaround for MSHADE-148, which leads to an infinite loop when building Alluxio.

There are two scenarios.
- Releasing Alluxio
Maven build with `-Prelease`. 
  - If using newer than `3.3.x` version of maven, you also have to use sequential build, Like `mvn clean install -Prelease xxxx`.
  - If using lower version maven before `3.3.x`, you can use parallel build, e.g. `mvn clean install -T 4 -Prelease xxxx`
- Not Releasing Alluxio
If you are not releasing Alluxio, you can build Alluxio like before, please don't add the `-Prelease` into you build command line. e.g. `mvn clean install -T 4 xxxx`

This patch adds a -Prelease. If present, it will set createDependencyReducedPom true. The consequences are:

If you are releasing Alluxio with this profile, you are fine as long as before.
If you are releasing Alluxio without this profile, the `alluxio-shaded-xxx-yyy.pom` will be the same as the pom.xml under the module directory.
If you are not releasing Alluxio but you are using this profile, you may run into #12338
If you are not releasing Alluxio and you did not include this profile, you are fine.
This is all documented in pom.xml and tested locally with maven 3.6.3.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1489,7 +1489,7 @@
       </build>
     </profile>
     <profile>
-      <id>release-profile</id>
+      <id>release</id>
       <properties>
         <!--
             The maven shade plugin has a bug where enabling the `createDependencyReducedPom`
--- a/pom.xml
+++ b/pom.xml
@@ -163,6 +163,7 @@
     <update.check.enabled>true</update.check.enabled>
     <update.check.host>https://diagnostics.alluxio.io</update.check.host>
     <findbugs.skip>false</findbugs.skip>
+    <create.dependency.reduced.pom>false</create.dependency.reduced.pom>
   </properties>
 
   <modules>
@@ -1486,6 +1487,20 @@
           </plugin>
         </plugins>
       </build>
+    </profile>
+    <profile>
+      <id>release-profile</id>
+      <properties>
+        <!--
+            The maven shade plugin has a bug where enabling the `createDependencyReducedPom`
+            property causes maven to go into an infinite loop (MSHADE-148). This is only an
+            issue for the Alluxio build if the maven version is 3.3.x or newer.
+            However, since disabling this property has the side effect of not resolving
+            variables in the released pom files, we need to enable this during
+            releases.
+        -->
+        <create.dependency.reduced.pom>true</create.dependency.reduced.pom>
+      </properties>
     </profile>
   </profiles>
 </project>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1487,5 +1488,19 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>release-profile</id>
```
apc999, line 1492: shall we add it here too @Xenorith  `-Prelease-profile`
https://github.com/Alluxio/alluxio/blob/384f944fa6c6986adeca8e598b62feebd7872276/dev/scripts/src/alluxio.org/build-distribution/cmd/generate-tarball.go#L99

Xenorith: from my understanding but correct me if i'm wrong, `-Prelease-profile` is only necessary when building artifacts to be published to maven. in all other cases, we would not want this profile to be enabled

apc999 (commented):
I would just call this `release` instead of `relase-profile`.
otherwise LGTM

@calvinjia  do you want to review this pr too?

pom.xml
Xenorith: @LuQQiu can you confirm that we do not want to append this profile argument to `generate-tarball.go` because this doesn't get executed when publishing to maven?

LuQQiu: Not quite sure if we need it when generating the tarball. But publish to maven doesn't have relationships with generate-tarball.go

maobaolong: @LuQQiu Thanks for your reply, as publish to maven doesn't have relationships with generate-tarball.go, so I think we don't need to add `-Prelease` here.

maobaolong:
@apc999 @LuQQiu @Xenorith  Thanks for your review and suggestion, I pushed a new commit to rename the release-profile to release, PTAL.

Xenorith (approved):
thanks for your efforts to resolve this issue. i verified that building maven without the release profile works. will wait for others to take a look and approve before merging

maobaolong:
@Xenorith Thanks a lot.

apc999 (approved):

apc999:

maobaolong:
Thank @apc999 for merging this.

apc999:

PR #12383, opened 2020-10-23
Fix memory leak in inode tree iterator

bf8086:
Currently when a user iterate through the inode tree journal entries, a few threads are created to iterate through different directories. The threads will be closed when the iteration completes, but if an iteration is not fully completed, then the threads are leaked and never get properly shutdown.

This change modifies the `JournalEntryIterable` interface to return a closeable iterator of the journal entries, so the user of the iterator can explicitly shutdown the threads by calling the `close()` method or using the `try-with-resource` pattern.

The majority of the logic are in `InodeTreeBufferedIterator` class and `CloseableIterator` class.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/InodeTreeBufferedIterator.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/InodeTreeBufferedIterator.java
@@ -15,11 +15,13 @@ import alluxio.conf.PropertyKey;
 import alluxio.conf.ServerConfiguration;
 import alluxio.master.metastore.InodeStore;
 import alluxio.proto.journal.Journal;
+import alluxio.resource.CloseableIterator;
 import alluxio.util.ThreadFactoryUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.Closeable;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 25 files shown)


bf8086:
@gpang PTAL.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreeBufferedIterator.java
```diff
@@ -302,4 +319,11 @@ public boolean hasNext() {
   public void remove() {
     throw new UnsupportedOperationException("remove is not supported in inode tree iterator");
   }
+
+  @Override
+  public void close() {
+    LOG.debug("Closing {} inode tree iterators", mActiveCrawlers.size());
+    mCoordinatorExecutor.shutdownNow();
```
gpang, line 326: is this the correct order to shut them down?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreeBufferedIterator.java
```diff
@@ -44,12 +46,13 @@
  * makes applying those entries later more efficient by guaranteeing that a parent of an inode is
  * iterated before it.
  */
-public class InodeTreeBufferedIterator implements Iterator<Journal.JournalEntry> {
+public class InodeTreeBufferedIterator implements Iterator<Journal.JournalEntry>, Closeable {
```
gpang, line 49: Is this the only iterator that implements close() right now?

gpang (commented):
@bf8086 minor questions

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreeBufferedIterator.java
bf8086: I don't think the order matters much. Both shutdown calls are asynchronous so the order of close does not make a big difference.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreeBufferedIterator.java
bf8086: Yes.

gpang (approved):
LGTM

bf8086:

bf8086:

PR #12394, opened 2020-10-26
Improve implementation of DistributedCp and other distributed jobs

bradyoo:
Few Changes:

1. DistributedMv is now DistributedCp + Delete (the ASYNC_THROUGH writeType safe guard for persisted file is now removed)
2. DistributedCp now pre-creates all of the folders from the client
3. DistributedCp now submits the request one file at a time.
4. Removed CheckMigrateValid because once you submit files one at a time, checkMigrateValid calls getStatus on a missing file each time which is very expensive.
5. The behavior of file overwriting another file used to be broken and is still broken in this implementation


bradyoo:
Brought back DistributedMv to be DistributedCp + Delete Source.

shell/src/main/java/alluxio/cli/fs/command/DistributedMvCommand.java
```diff
@@ -50,31 +44,22 @@ public String getCommandName() {
 
   @Override
   public void validateArgs(CommandLine cl) throws InvalidArgumentException {
-    CommandUtils.checkNumOfArgsEquals(this, cl, 2);
+    mCpCommand.validateArgs(cl);
   }
 
   @Override
   public int run(CommandLine cl) throws AlluxioException, IOException {
+    mCpCommand.run(cl);
```
gpang, line 52: If this can throw an exception, should we try to catch it, to let the user know that the copy phase did not complete, so it might be in an inconsistent state?

If it is in an inconsistent state, what happens if we run distributedMv again?

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
```diff
@@ -0,0 +1,81 @@
...
+      waitJob();
+    }
+  }
+
+  /**
+   * Waits for at least one job to complete.
+   */
+  protected void waitJob() {
+    AtomicBoolean removed = new AtomicBoolean(false);
+    while (true) {
```
gpang, line 58: Is it intentional that this while loop is very tight? Can we add a sleep in here to avoid too much unnecessary cpu?

shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
```diff
@@ -59,23 +71,85 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
+    }
+  }
+
+  private void addJob(String srcPath, String dstPath) {
+    if (mSubmittedJobAttempts.size() >= mActiveJobs) {
+      // Wait one job to complete.
+      waitJob();
+    }
+    mSubmittedJobAttempts.add(newJob(srcPath, dstPath));
+    System.out.println("Copying " + srcPath + " to " + dstPath);
```
gpang, line 150: Did we always print out every file? This could be a huge output, if we are copying a large tree?

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
```diff
@@ -0,0 +1,81 @@
...
+  protected void drain() {
+    while (!mSubmittedJobAttempts.isEmpty()) {
+      waitJob();
+    }
+  }
+
+  /**
+   * Waits for at least one job to complete.
+   */
+  protected void waitJob() {
```
gpang, line 56: This is not threadsafe so we should add `synchronized` for this method.

Also, maybe rename to `waitForCompleteJob`?

shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
```diff
@@ -59,23 +71,85 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
+
+    return jobAttempt;
+  }
+
+  private void distributedCp(AlluxioURI srcPath, AlluxioURI dstPath)
+      throws IOException, AlluxioException {
+    if (mFileSystem.getStatus(srcPath).isFolder()) {
+      createFolders(srcPath, dstPath);
+    }
+    copy(srcPath, dstPath);
```
gpang, line 104: `createFolders` does a listing, and `copy` also does a listing. Can we avoid doing a double recursive listing of the source?

gpang (changes requested):
@bradyoo Thanks! I left a few comments.

shell/src/main/java/alluxio/cli/fs/command/DistributedMvCommand.java
bradyoo: distributedMv fails because overwrite is broken. But at the very least, you can delete the destination directory and then rerun distributedMv

shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
bradyoo: We didn't but we did for distributedLoad and that has been popular so choosing that instead.

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
bradyoo: This is a command. There is nothing threadsafe about anything in any of these series of classes.

shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
bradyoo: I would prefer not because:

1. copy is expensive anyways so listStatus is a small cost here.
2. More intuitive log messages.

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
```diff
@@ -0,0 +1,81 @@
...
+  protected int mActiveJobs;
+  protected final JobMasterClient mClient;
+
+  protected AbstractDistributedJobCommand(FileSystemContext fsContext) {
+    super(fsContext);
+    mSubmittedJobAttempts = Lists.newArrayList();
+    final ClientContext clientContext = mFsContext.getClientContext();
+    mClient = JobMasterClient.Factory.create(
+        JobMasterClientContext.newBuilder(clientContext).build());
+    mActiveJobs = DEFAULT_ACTIVE_JOBS;
```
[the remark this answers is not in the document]
bradyoo: This number is large enough and it hasn't actually mattered in distributedLoad case.

bradyoo:
Addressed Feedback

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
```diff
@@ -0,0 +1,83 @@
...
+            removed.set(true);
+            return false;
+          default:
+            throw new IllegalStateException(String.format("Unexpected Status: %s", check));
+        }
+      }).collect(Collectors.toList());
+      if (removed.get()) {
+        return;
+      }
+      CommonUtils.sleepMs(5);
```
gpang, line 80: Is there a specific reason we need to loop so quickly? For example, if we loop every second, does that slow anything down unnecessarily?

shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
gpang: Could you explain why it is more intuitive with 2 listings, and why 1 listing cannot have good messages?

One concern is that if a listing is a large tree, it would have to lock that large tree twice. Ideally, we would only have to lock that subtree once.

gpang (commented):
@bradyoo Just a few more questions.

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
bradyoo: Second will be quite detrimental to the performance of this when copying lots of small files that were initially held back by a backlog of other jobs.

bradyoo:
@gpang 
1. The best way to achieve 1 listing is to store the 1st listing in some way. This is a lot of memory usage.
2. For validation purposes, I would like to go through the folder structure first before doing any "real" copying by submitting jobs.

shell/src/main/java/alluxio/cli/fs/command/AbstractDistributedJobCommand.java
gpang: ok, sounds good

gpang (approved):
LGTM

bradyoo:

Issue #12126, opened 2020-09-22
loadMetadata can't return immediately after all the metadata is loaded


TrafalgarZZZ:
**Alluxio Version:**
What version of Alluxio are you using?
https://github.com/Alluxio/alluxio/tree/branch-2.3-fuse

**Describe the bug**
I mount an Aliyun OSS including about 3.8 million small files on Alluxio, after then I run the `alluxio fs loadMetadata -R /` to load metadata of all the files in my OSS. 
What confuses me is that before the `alluxio fs loadMetadata -R /` return, I run the followiing command:
```
bash-4.4# alluxio fsadmin report metrics | grep Master.TotalPaths
Master.TotalPaths  (Type: GAUGE, Value: 3,804,850)
```
And the value "3804850" is exactly the number of all my files in my OSS. So it seems to me all the metadata has already been loaded, however, the `alluxio fs loadMetadata -R /` command still won't return until about 10min later.

**To Reproduce**
Steps to reproduce the behavior (as minimally and precisely as possible)
Try to load metadata of large amount of files and run the `report metrics` to check `Master.TotalPaths` 

**Expected behavior**
The loadMetadata command should return as soon as the job is done

**Urgency**
Describe the impact and urgency of the bug.

**Additional context**

apc999:
this may be related to client-side calculating results streamed from master
We may need bring back some previous change from 
https://github.com/Alluxio/alluxio/pull/3272/files
https://github.com/Alluxio/alluxio/pull/8546/files

apc999:
this should be fixed by #12421

PR #12421, opened 2020-10-29
Add load metadata command

JySongWithZhangCe:
First step for #12088 and #12126


JySongWithZhangCe:
alluxio-bot, test this please

core/client/fs/src/test/java/alluxio/client/file/cache/LocalCacheFileInStreamTest.java
```diff
@@ -497,6 +498,12 @@ public void iterateStatus(AlluxioURI path, ListStatusPOptions options,
       throw new UnsupportedOperationException();
     }
 
+    @Override
+    public void loadMetadata(AlluxioURI path, LoadMetadataPOptions options)
+        throws FileDoesNotExistException, IOException, AlluxioException {
+      throw new UnsupportedOperationException();
```
jiacheliu3, line 504: Unsupported?

core/client/fs/src/main/java/alluxio/client/file/DelegatingFileSystem.java
```diff
@@ -120,11 +121,17 @@ public URIStatus getStatus(AlluxioURI path, GetStatusPOptions options)
 
   @Override
   public void iterateStatus(AlluxioURI path, ListStatusPOptions options,
-                            Consumer<? super URIStatus> action)
```
jiacheliu3, line 123: Fix the whitespace here

core/transport/src/main/proto/grpc/file_system_master.proto
```diff
@@ -193,10 +193,16 @@ message ListStatusPRequest {
 
 message LoadMetadataPOptions {
   /** whether to load metadata recursively */
-  optional bool recursive = 1;
-  optional bool createAncestors = 2;
-  optional fscommon.LoadDescendantPType loadDescendantType = 3;
-  optional FileSystemMasterCommonPOptions commonOptions = 4;
+  //  optional bool recursive = 1;
+  optional bool createAncestors = 1;
+  optional fscommon.LoadDescendantPType loadDescendantType = 2;
```
jiacheliu3, line 198: Use `LoadDescendantType` to replace `recursive`?

shell/src/main/java/alluxio/cli/fs/command/LoadMetadataCommand.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+
+import java.io.IOException;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Load Metadata for the path.
```
jiacheliu3, line 32: Can you explain more pls?

shell/src/test/java/alluxio/cli/fs/command/LoadMetadataCommandTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+public class LoadMetadataCommandTest {
+
+  /**
+   * Metadata not in Alluxio, LoadMetadata should
+   * load the metadata into Alluxio inode.
+   */
+  @Test
+  public void noneInAlluxio() {
```
jiacheliu3, line 23: Add the tests so we know what you want the behavior to be?

jiacheliu3 (changes requested):
Thanks @JySongWithZhangCe for the contribution!

Left a few comments, let's further discuss when you resolved those. Thanks!

core/client/fs/src/main/java/alluxio/client/file/FileSystemMasterClient.java
```diff
@@ -169,6 +170,15 @@ void iterateStatus(AlluxioURI path, ListStatusPOptions options,
   List<URIStatus> listStatus(AlluxioURI path, ListStatusPOptions options)
       throws AlluxioStatusException;
 
+  /**
+   * Loads the metadata of a file from the under file system.
```
apc999, line 174: nit: a file -> a path

core/client/fs/src/test/java/alluxio/client/file/cache/LocalCacheFileInStreamTest.java
apc999: yeah this is a mock

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,23 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
     return fileInfos;
   }
 
+  @Override
+  public void loadMetaData(AlluxioURI path, LoadMetadataContext context)
+      throws AccessControlException, FileDoesNotExistException, InvalidPathException, IOException {
+    try (RpcContext rpcContext = createRpcContext(context)) {
+      DescendantType syncDescendantType =
```
apc999, line 1073: can we invoke `loadMetadataIfNotExist` here? It looks to me some duplicated code with that method

core/transport/src/main/proto/grpc/file_system_master.proto
```diff
@@ -193,10 +193,16 @@ message ListStatusPRequest {
 
 message LoadMetadataPOptions {
   /** whether to load metadata recursively */
-  optional bool recursive = 1;
-  optional bool createAncestors = 2;
-  optional fscommon.LoadDescendantPType loadDescendantType = 3;
-  optional FileSystemMasterCommonPOptions commonOptions = 4;
+  //  optional bool recursive = 1;
```
apc999, line 196: instead of replacing old field which will break backward compatibility,
let's add new fields after 4, starting from 5

shell/src/main/java/alluxio/cli/fs/command/LoadMetadataCommand.java
```diff
@@ -0,0 +1,111 @@
...
+    AlluxioURI path = new AlluxioURI(args[0]);
+    runWildCardCmd(path, cl);
+
+    return 0;
+  }
+
+  private void loadMetadata(AlluxioURI path, boolean recursive) throws IOException {
+    LoadMetadataPOptions options;
+    try {
+      if (recursive) {
```
apc999, line 81: Too verbose. how about
```
LoadMetadataPOptions.newBuilder().
    .setLoadDescendantType(recursive  ? LoadDescendantPType.ALL : LoadDescendantPType.ONE)
    .build()
```

shell/src/test/java/alluxio/cli/fs/command/LoadMetadataCommandTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.cli.fs.command;
+
+import org.junit.Test;
+
+public class LoadMetadataCommandTest {
```
apc999, line 16: Let's make an integration test instead?
see `alluxio.client.cli.fs.command.LoadCommandIntegrationTest`

core/transport/src/main/proto/grpc/file_system_master.proto
```diff
@@ -524,6 +530,7 @@ service FileSystemMasterClientService {
    */
   rpc ListStatus(ListStatusPRequest) returns (stream ListStatusPResponse);
 
+  rpc ListStatusCli(ListStatusCliPRequest) returns (ListStatusCliPResponse);
```
apc999, line 533: for this PR, let's just focus on `loadMetadata`, and introduce only

```
rpc LoadMetadata(LoadMetadataPRequest) returns (LoadMetadataPResponse);
```

Let's not aim for multiple purposes in a single PR

apc999 (changes requested):
Left some comments. Thansk!

core/transport/src/main/proto/grpc/file_system_master.proto
JySongWithZhangCe: I agree

shell/src/main/java/alluxio/cli/fs/command/LoadMetadataCommand.java
JySongWithZhangCe: Sure

shell/src/test/java/alluxio/cli/fs/command/LoadMetadataCommandTest.java
JySongWithZhangCe: Ok, I'm working on it.

core/client/fs/src/main/java/alluxio/client/file/FileSystemMasterClient.java
JySongWithZhangCe: Sure

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,23 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
...
+              GrpcUtils.fromProto(context.getOptions().getLoadDescendantType());
+      FileSystemMasterCommonPOptions commonOptions =
+              context.getOptions().getCommonOptions();
+      InodeSyncStream sync = new InodeSyncStream(new LockingScheme(path, LockPattern.READ, true),
+              this, rpcContext, syncDescendantType, commonOptions, false, true, false);
+
```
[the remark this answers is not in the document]
JySongWithZhangCe: Sure

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,23 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
     return fileInfos;
   }
 
+  @Override
+  public void loadMetaData(AlluxioURI path, LoadMetadataContext context)
+      throws AccessControlException, FileDoesNotExistException, InvalidPathException, IOException {
+    try (RpcContext rpcContext = createRpcContext(context)) {
+      DescendantType syncDescendantType =
+              GrpcUtils.fromProto(context.getOptions().getLoadDescendantType());
```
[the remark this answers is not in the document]
JySongWithZhangCe: Ok

shell/src/main/java/alluxio/cli/fs/command/LoadMetadataCommand.java
JySongWithZhangCe: Good idea!

shell/src/test/java/alluxio/cli/fs/command/LoadMetadataCommandTest.java
JySongWithZhangCe: I will try

core/transport/src/main/proto/grpc/file_system_master.proto
JySongWithZhangCe: You ever suggested RPCs which associated with ListStatus should be merged into a common RPC, and this design is not complicated. So shaw I apply this design in another PR？

core/client/fs/src/main/java/alluxio/client/file/DelegatingFileSystem.java
JySongWithZhangCe: Sure

tests/src/test/java/alluxio/client/cli/fs/command/LoadMetadataCommandIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+public final class LoadMetadataCommandIntegrationTest extends AbstractFileSystemShellTest {
+  @Test
+  public void loadMetadataDir() throws IOException, AlluxioException {
+    String dirPath = "/testRoot/layer1/layer2/layer3/";
+    String filePathA = dirPath + "testFileA";
+    String filePathB = dirPath + "testFileB";
+    FileSystemTestUtils
+        .createByteFile(sFileSystem, filePathA, WritePType.CACHE_THROUGH, 10);
+    FileSystemTestUtils
+        .createByteFile(sFileSystem, filePathB, WritePType.THROUGH, 30);
```
jiacheliu3, line 42: Why creating two files with different WriteType? If you want to validation all write types, maybe add files for each of them?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,31 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
     return fileInfos;
   }
 
+  /**
+   * {@link DefaultFileSystemMaster} has two private methods associated with sync metadata:
+   * 1. {@link DefaultFileSystemMaster#loadMetadataIfNotExist} sync metadata in load-only mode,
```
jiacheliu3, line 1071: what is load-only mode?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,31 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
...
+   * 2. {@link DefaultFileSystemMaster#syncMetadata} sync metadata not in force mode.
```
jiacheliu3, line 1072: what is "not in force mode"? what is "force mode"? Can you explain more clearly?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,31 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
...
+   * Due to loadMetaData need work on both non-load-only mode and force mode, we need custom a
+   * InodeSyncStream object which foreSync param is true and loadOnly param is false.
+   * Further more, synchronize metadata may modify the structure of the inode tree, the lock
+   * pattern of the InodeSyncStream object should be LockPattern.WRITE_EDGE.
```
jiacheliu3, line 1076: I don't really understand the comment. When you are using words like "foreSync param", "loadOnly param", could you explain them first, so the readers can understand what you are referring to?

jiacheliu3 (changes requested):
Thanks @JySongWithZhangCe !
The code looks mostly good to me. Could you improve the comment to make them more friendly to readers with little context? Thanks!

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
JySongWithZhangCe: load-only mode won't update metadata which is not consistency with UFS, this mode will load metadata which not in alluxio inode tree but exists in UFS.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
JySongWithZhangCe: That is to say, sync may not execute

core/client/fs/src/main/java/alluxio/client/file/RetryHandlingFileSystemMasterClient.java
```diff
@@ -260,6 +262,19 @@ public void iterateStatus(final AlluxioURI path, final ListStatusPOptions option
...
+  @Override
+  public void loadMetadata(AlluxioURI path, LoadMetadataPOptions options)
+      throws AlluxioStatusException {
+    retryRPC(
+        () -> mClient.loadMetadata(
+              LoadMetadataPRequest.newBuilder()
+                  .setPath(path.getPath())
+                  .setLoadMetadataPOptions(options)
+                  .build()),
+        RPC_LOG, "loadMetadata", "path=%s,options=%s", path, options
```
apc999, line 274:
```suggestion
        RPC_LOG, "LoadMetadata", "path=%s,options=%s", path, options
```

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1066,6 +1066,31 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
...
+   */
+  @Override
+  public void loadMetaData(AlluxioURI path, LoadMetadataContext context)
+      throws AccessControlException, FileDoesNotExistException, InvalidPathException, IOException {
+    try (RpcContext rpcContext = createRpcContext(context)) {
+      DescendantType syncDescendantType =
+          GrpcUtils.fromProto(context.getOptions().getLoadDescendantType());
+      FileSystemMasterCommonPOptions commonOptions = context.getOptions().getCommonOptions();
+      InodeSyncStream sync =
+          new InodeSyncStream(new LockingScheme(path, LockPattern.WRITE_EDGE, true), this,
```
apc999, line 1086: @gpang  to sync metadata, do we need to start with `LockPattern.WRITE_EDGE`? can we start with `READ`?
Also, can you check if the parameters for `InodeSyncStream` here is correct?

core/server/master/src/main/java/alluxio/master/file/FileSystemMasterClientServiceHandler.java
```diff
@@ -237,6 +240,19 @@ public void listStatus(ListStatusPRequest request,
...
+  public void loadMetadata(LoadMetadataPRequest request,
+      StreamObserver<LoadMetadataPResponse> responseObserver) {
+    RpcUtils.call(LOG, () -> {
+      AlluxioURI pathUri = getAlluxioURI(request.getPath());
+      if (request.hasLoadMetadataPOptions()) {
+        mFileSystemMaster.loadMetaData(pathUri,
+            LoadMetadataContext.create(request.getLoadMetadataPOptions().toBuilder()));
+      }
+      return LoadMetadataPResponse.newBuilder().build();
+    }, "loadMetadata", "request=%s", responseObserver, request);
```
apc999, line 253:
```suggestion
    }, "LoadMetadata", "request=%s", responseObserver, request);
```

core/transport/src/main/proto/grpc/file_system_master.proto
```diff
@@ -524,6 +529,7 @@ service FileSystemMasterClientService {
    */
   rpc ListStatus(ListStatusPRequest) returns (stream ListStatusPResponse);
 
+  rpc LoadMetadata(LoadMetadataPRequest) returns (LoadMetadataPResponse);
```
apc999, line 532: Can you comment this rpc and also leave the space between this definition and the next RPC definition

tests/src/test/java/alluxio/client/cli/fs/command/LoadMetadataCommandIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.io.IOException;
+
+/**
+ * Tests for loadMetadata command.
+ */
+public final class LoadMetadataCommandIntegrationTest extends AbstractFileSystemShellTest {
+  @Test
+  public void loadMetadataDir() throws IOException, AlluxioException {
+    String dirPath = "/testRoot/layer1/layer2/layer3/";
+    String filePathA = dirPath + "testFileA";
```
apc999, line 37:
```suggestion
    String filePathA = PathUtils.concatPath(dirPath, "testFileA");
```

tests/src/test/java/alluxio/client/cli/fs/command/LoadMetadataCommandIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+    // Use LoadMetadataPType.NEVER to avoid loading metadata during get file status.
+    GetStatusPOptions getStatusPOptions =
+        GetStatusPOptions.newBuilder().setLoadMetadataType(LoadMetadataPType.NEVER).build();
+    // Check testFileA's metadata.
+    URIStatus statusAfterA = sFileSystem.getStatus(uriA, getStatusPOptions);
+    assertEquals(statusBeforeA.getFileInfo().getName(), statusAfterA.getFileInfo().getName());
+    assertEquals(statusBeforeA.getFileInfo().getLength(), statusAfterA.getFileInfo().getLength());
+    // Check testFileB's metadata.
+    URIStatus statusAfterB = sFileSystem.getStatus(uriB, getStatusPOptions);
+    assertEquals(statusBeforeB.getFileInfo().getName(), statusAfterB.getFileInfo().getName());
```
apc999, line 63: Possibly to `assertEquals(statusBefore, statusAfterB)`?

apc999 (changes requested):
left some comments, thanks!

JySongWithZhangCe:
Thx Bin and Jiacheng, help me get better!

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: I think it can be READ too. However, shouldn't this behave the same as `listStatus` until it is time to return results? I feel this command should be just the prefix of the code for `listStatus`. Alternatively, should there just be an option to `listStatus` which ignores sending results, or something like "forceLoadOnly"? There would be a lot of shared code between load metadata and list status.

gpang (commented):
@apc999

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
apc999: thanks. @JySongWithZhangCe  based on Gene's feedback, let's try to emulate what `listStatus` is doing and reuse code as much as possible. 

Meanwhile, @gpang , I agree with you that it can be a new option in `ListStatus` rpc. However, given the code we are adding a minimal, I view it is cleaner to have a separate RPC for its own purpose rather than complicating listStatus which is already pretty complicated. Let me know if you feel it strongly and we can discuss

apc999 (changes requested):
Comments left w.r.t. [creating a new InodeSyncStream](https://github.com/Alluxio/alluxio/pull/12421/files#r516451881)

core/client/fs/src/main/java/alluxio/util/FileSystemOptions.java
```diff
@@ -180,9 +180,9 @@ public static ListStatusPOptions listStatusDefaults(AlluxioConfiguration conf) {
   public static LoadMetadataPOptions loadMetadataDefaults(AlluxioConfiguration conf) {
     return LoadMetadataPOptions.newBuilder()
         .setCommonOptions(commonDefaults(conf))
-        .setRecursive(false)
         .setCreateAncestors(false)
         .setLoadDescendantType(LoadDescendantPType.NONE)
+        .setRecursive(false)
```
jiacheliu3, line 185: nit: restore the order

core/client/fs/src/test/java/alluxio/util/FileSystemOptionsTest.java
```diff
@@ -39,8 +39,8 @@ public void before() throws Exception {
   public void loadMetadataOptionsDefaults() {
     LoadMetadataPOptions options = FileSystemOptions.loadMetadataDefaults(mConf);
     assertNotNull(options);
-    assertFalse(options.getRecursive());
     assertFalse(options.getCreateAncestors());
+    assertFalse(options.getRecursive());
```
jiacheliu3, line 43: nit: restore the order

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1040,6 +1040,10 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
           }
           ensureFullPathAndUpdateCache(inodePath);
 
+          if (context.getOptions().getLoadMetadataOnly()) {
+            continue;
```
jiacheliu3, line 1044: @apc999 Do you think we should update the auditContext here?
```
auditContext.setSucceeded(true);
```

Similarly, update the metrics too?

jiacheliu3 (changes requested):
Thanks! Mostly looks good to me now, except one question. Also please resolve Bin's previous comments.

jiacheliu3 (approved):
LGTM. Thanks for the contribution!

core/transport/src/main/proto/grpc/file_system_master.proto
```diff
@@ -184,6 +184,8 @@ message ListStatusPOptions {
   optional LoadMetadataPType loadMetadataType = 2;
   optional FileSystemMasterCommonPOptions commonOptions = 3;
   optional bool recursive = 4;
+  // Only loads metadata about a path in the UFS to Alluxio. No data will be transferred.
+  optional bool loadMetadataOnly = 5;
```
gpang, line 188: How does this option behave if the other option `LoadMetadataPType loadMetadataType` is set to `NEVER`? It looks like this option is really about sending results back or not, not about loading metadata? Is this a correct understanding?

If this option only controls returning results to the client or not, then we should rename the option to be more specific about returning results, instead of mentioning loading metadata.

gpang (commented):
@JySongWithZhangCe Thanks! I left a question about one of the options.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1040,6 +1040,11 @@ public void listStatus(AlluxioURI path, ListStatusContext context,
           }
           ensureFullPathAndUpdateCache(inodePath);
 
+          if (context.getOptions().getLoadMetadataOnly()) {
```
apc999, line 1043: @gpang  is this the correct location to continue if we don't need results returned?

core/transport/src/main/proto/grpc/file_system_master.proto
apc999: the understanding is correct.
perhaps renamed this flag to something like `resultsRequired`? @gpang

apc999:
@gpang  can you take a look at the two questions I put? thanks

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -289,6 +289,18 @@ public void iterateStatus(AlluxioURI path, final ListStatusPOptions options,
     });
   }
 
+  @Override
+  public void loadMetadata(AlluxioURI path, final ListStatusPOptions options)
+      throws FileDoesNotExistException, IOException, AlluxioException {
+    checkUri(path);
+    rpc(client -> {
+      ListStatusPOptions mergedOptions = FileSystemOptions.listStatusDefaults(
+          mFsContext.getPathConf(path)).toBuilder().mergeFrom(options).build();
```
apc999, line 298: set the added flag explicitly here?

apc999: @JySongWithZhangCe 

Better to add `.setLoadMetadataOnly(true)` explicitly here?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
apc999: @gpang  ping

core/transport/src/main/proto/grpc/file_system_master.proto
apc999: the flag added to `ListStatusPOption`

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
apc999: Let's still keep the param as `ListStatusPOptions options`, to be consistent with other methods in this file

JySongWithZhangCe: Ok

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
gpang: I think this is fine, but I think I would suggest a slightly different approach. The part after this code gathers all the results, and updates metrics. If we return here, we will not get some of the metrics accounting. Instead, it could look something like this:

```java
ensureFullPathAndUpdateCache(inodePath);

auditContext.setSrcInode(inodePath.getInode());
if (!context.getOptions().getLoadMetadataOnly()) {  // this line is new
  DescendantType descendantTypeForListStatus =
    (context.getOptions().getRecursive()) ? DescendantType.ALL : DescendantType.ONE;
  listStatusInternal(context, rpcContext, inodePath, auditContext,
    descendantTypeForListStatus, resultStream, 0);
}    // this line is new
auditContext.setSucceeded(true);
Metrics.FILE_INFOS_GOT.inc();
if (!ufsAccessed) {
  MountTable.Resolution resolution = mMountTable.resolve(inodePath.getUri());
  Metrics.getUfsCounter(mMountTable.getMountInfo(resolution.getMountId())
      .getUfsUri().toString(),
    Metrics.UFSOps.LIST_STATUS).inc();
}
```

Basically, what this is doing is the method is unchanged, but the option only guards the call to `listStatusInternal()` which is the method that fills in all the results. If we don't want results, we only skip calling that method.

What do you think?

gpang (changes requested):
@JySongWithZhangCe @apc999 I responded to the question.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
apc999: @JySongWithZhangCe 
we should add `.setLoadMetadataOnly(true)` explicitly here, right?
otherwise, how to ensure callers of this method has this set?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
apc999: SGTM, @JySongWithZhangCe  Can you address @gpang 's comments?

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
JySongWithZhangCe: I agree

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
JySongWithZhangCe: SGTM, Thx @gpang and @apc999 !

JySongWithZhangCe:
alluxio-bot,test this please

JySongWithZhangCe:
alluxio-bot, test this please

JySongWithZhangCe:
alluxio-bot, test this please

apc999 (approved):
thanks for the improvement!

apc999:

JySongWithZhangCe:
Thinks !!! @apc999 @jiacheliu3 @gpang

yuzhu:
@JySongWithZhangCe I noticed that this change introduces a flag `ResultsRequired` in the listStatusOptions, and it defaults to false.  As a result, older versions of alluxio client can no longer listStatus from the newer version of alluxio master.  Do you think it makes sense to change it to 'ResultsNotRequired' flag so the backwards compatibility is maintained? Basically we need it to default to return results, otherwise older alluxio clients will always get empty listings.   FYI @apc999 @gpang @jiacheliu3 This causes catalog service client to fail to get results for queries because those clients are older.

JySongWithZhangCe:
@yuzhu I agree with you. @apc999 @jiacheliu3 What do you think?

jiacheliu3:
@yuzhu This makes sense to me.

apc999:
@JySongWithZhangCe agree.

apc999:
Instead of changing it to`ResultsNotRequired`, how about `SummaryOnly`?

JySongWithZhangCe:
`SummaryOnly` looks like better, I'll make a pr later

yuzhu:
@JySongWithZhangCe thanks!

apc999:
https://github.com/Alluxio/alluxio/pull/12577/ @yuzhu

PR #12440, opened 2020-10-30
Fix flakey temp dir creation

yuzhu:


core/common/src/test/java/alluxio/AlluxioTestDirectory.java
```diff
@@ -70,17 +72,15 @@ public static File createTemporaryDirectory(File parent, String prefix) {
...
-        throw new RuntimeException(
-            "Failed to create testing directory " + tmpDir.getAbsolutePath());
+    try {
+      final Path tmpDir = Files.createTempDirectory("alluxio-tests");
+      if (Files.exists(tmpDir)) {
+        cleanUpOldFiles(tmpDir.toFile());
       }
+      return tmpDir.toFile();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
gpang, line 82: Can you add a similar messages as before?

gpang (changes requested):
@yuzhu Thanks! Could you explain what was the issue with the old behavior? It is not obvious from the PR description.

Issue #12372, opened 2020-10-23
Alluxio ListStatus Always synchronize the metadata of files in the next level directory


cammette:
**Alluxio Version:**
alluxio 2.3.0

**Describe the bug**
ths alluxio path  /nfs is nfs underfs，and the /nfs has the following directory structure:
```
/nfs/1990
/nfs/1991
...
/nfs/2020
```
There are 5000 files in every /nfs/{year} subdirectory，and list the alluxio path /nfs is very slow
```
time alluxio fs ls  -Dalluxio.user.file.metadata.sync.interval=0 /nfs
Getting directory status of 13 files or sub-directories may take a while.
           5000       PERSISTED 10-23-2020 14:12:35:000  DIR /nfs/1990
           5000       PERSISTED 10-23-2020 14:13:48:000  DIR /nfs/1991
           5000       PERSISTED 10-23-2020 14:43:05:000  DIR /nfs/1992
           5000       PERSISTED 10-23-2020 14:46:31:000  DIR /nfs/1993
           5000       PERSISTED 10-23-2020 15:08:56:000  DIR /nfs/1994
           5000       PERSISTED 10-23-2020 16:30:17:000  DIR /nfs/1995
           5165       PERSISTED 10-23-2020 17:26:54:000  DIR /nfs/1996
           5000       PERSISTED 10-23-2020 17:27:32:000  DIR /nfs/1997
           5000       PERSISTED 10-18-2020 12:11:00:000  DIR /nfs/2016
           5000       PERSISTED 10-18-2020 12:07:59:000  DIR /nfs/2017
           5000       PERSISTED 10-18-2020 12:07:01:000  DIR /nfs/2018
           5000       PERSISTED 10-18-2020 12:05:27:000  DIR /nfs/2019
           1001       PERSISTED 10-18-2020 12:03:51:000  DIR /nfs/2020
real	2m49.228s
user	0m2.127s
sys	0m2.089s
```
I don't need to browse the metadata information of subdirectories in advance, just check whether there is a new directory under / nfs，Any suggestions？

maobaolong:
@cammette  Thanks for report this issue.

I can see each line of the subdir of /nfs contains a size, most of them are 5000, for example `/nfs/1990` contains 5000 children, without loading children of `/nfs/1990`, Alluxio cannot get the childCount, do you have any other idea?

apc999:
[...]

PR #12441, opened 2020-10-31
Prevent unnecessary inode syncing

apc999:
Fix https://github.com/Alluxio/alluxio/issues/12372

Before this PR, running 
```
bin/alluxio fs ls -Dalluxio.user.file.metadata.sync.interval=0 /path
```
will trigger syncing inodes:
(1) `/path`
(2) `/path/{children}`
(3) `/path/{children}/{children}` (added to pending inode queue when `InodeSyncStream` traversing (2))

Reading (3) is unnecessary w.r.t. the command we are interested. This PR check and prevent setting flag `syncChildren` unnecessarily.

In addition, this patch also saves one unnecessary RPC on ls command.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -67,7 +67,6 @@ import com.google.common.base.MoreObjects;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nullable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Collection;
@@ -82,6 +81,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.function.Function;
+
+import javax.annotation.Nullable;
 
 /**
  * This class is responsible for maintaining the logic which surrounds syncing metadata between
@@ -569,9 +570,16 @@ public class InodeSyncStream {
       }
     }
 
-    syncChildren = syncChildren
-        && inode.isDirectory()
-        && mDescendantType != DescendantType.NONE;
+    // Only sync children when
+    // (1) DescendantType.ALL or (2) syncing root of this stream && DescendantType.ONE
+    if (mDescendantType == DescendantType.ONE) {
+      syncChildren =
+          syncChildren && inode.isDirectory() && mRootScheme.getPath().equals(inodePath.getUri());
+    } else if (mDescendantType == DescendantType.ALL) {
+      syncChildren = syncChildren && inode.isDirectory();
+    } else {
+      syncChildren = false;
+    }
 
     Map<String, Inode> inodeChildren = new HashMap<>();
     if (syncChildren) {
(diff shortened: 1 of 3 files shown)


maobaolong:
image
@apc999 Thanks for this PR, after a test, I think this PR can reduce the next next level metadata sync, but the result here is not correct about the children count.

apc999:
@maobaolong  I view stale child count should be fine because we are listing  dir `/abc/12345` here, rather than the content of `/abc/12345/a`. In terms of content of this `/abc/12345/` dir, it is still correct.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -569,9 +570,16 @@ private void syncExistingInodeMetadata(LockedInodePath inodePath)
       }
     }
 
-    syncChildren = syncChildren
-        && inode.isDirectory()
-        && mDescendantType != DescendantType.NONE;
+    // Only sync children when
+    // (1) DescendantType.ALL or (2) syncing root of this stream && DescendantType.ONE
```
gpang, line 574: Can we encode this desired behavior in a simple unittest?

gpang (changes requested):
@apc999 just a minor comment

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
apc999: Gene, there is no unit test for this class `InodeSyncStream` though I think this class should be covered by UT definitely. Do you know if there is any other integration tests covering this part? I am very unfamiliar with this part of source code and it would be a lot to investigate to understand this class and write a UT

gpang: I think we can use `UfsSyncIntegrationTest` for testing sync behavior.

apc999: added. PTAL

tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
```diff
@@ -168,6 +184,21 @@ public void listDirSync() throws Exception {
...
 
+  // https://github.com/Alluxio/alluxio/issues/12372
+  @Test
+  public void listDirSyncOnlyTouchingChildren() throws Exception {
+    String dir1 = PathUtils.concatPath(EXISTING_DIR, "dir_should_sync");
+    String dir2 = PathUtils.concatPath(dir1, "dir_should_not_sync");
+    new File(ufsPath(dir1)).mkdirs();
+    new File(ufsPath(dir2)).mkdirs();
+    ListStatusPOptions optionsAlways = ListStatusPOptions.newBuilder()
+        .setLoadMetadataType(LoadMetadataPType.NEVER).setCommonOptions(PSYNC_ALWAYS).build();
```
gpang, line 195: Does this work because recursive is `false` by default? We want it false, right?

gpang (approved):
@apc999 Just a minor question, but looks good

LGTM (after tests pass)

tests/src/test/java/alluxio/client/fs/UfsSyncIntegrationTest.java
apc999: correct

apc999:

LuQQiu:

yuzhu:

PR #12442, opened 2020-10-31
Add etag header to s3 api proxy

### LINKED ISSUE #12212: ETag headers in S3 API Proxy

Trying to read files from the [Alluxio S3 API Proxy](https://docs.alluxio.io/os/user/stable/en/api/S3-API.html) with the Python library [s3fs](https://github.com/dask/s3fs/).

When trying to read a file, the s3fs library code throws an exception with the following message:

    KeyError: 'ETag'

I think this exception occurs because the Alluxio S3 API Proxy does not send `ETag` headers and the s3fs library expects to receive them.

It looks like the lack of `ETag` headers is a known issue based on comments at [line 397](https://github.com/Alluxio/alluxio/blob/71554a48a840ea894db58b20560f30214f663b0b/core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java#L397) and [line 478](https://github.com/Alluxio/alluxio/blob/71554a48a840ea894db58b20560f30214f663b0b/core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java#L478) of `S3RestServiceHandler.java`.

Ideally, the Alluxio S3 API Proxy would send an `ETag` header for `Get Object` and `Get Object Info (HEAD)` requests, and the value of that header would be an MD5 hash of the file's contents. This would match the behavior of Amazon S3 in most cases.

I was able to work around the problem by modifying the s3fs library to ignore the missing `ETag` header.

Another alternative would be to modify the Alluxio S3 API Proxy to send an `ETag` header that is not an MD5 hash but acts as an equivalent signal when the file changes. For example, it could be a hash of the file's modification time. There is precedent for this, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag):

> The method by which `ETag` values are generated is not specified.
> Often, a hash of the content, a hash of the last modification timestamp,
> or just a revision number is used.

**Urgency**

Low priority since a workaround is possible.

**Additional context**

Since Amazon S3 always sends `ETag` headers, there may be other S3 client libraries with a similar issue.

maobaolong:
Fix #12212

### DIFF
--- a/core/server/proxy/src/main/java/alluxio/proxy/s3/S3Constants.java
+++ b/core/server/proxy/src/main/java/alluxio/proxy/s3/S3Constants.java
@@ -19,6 +19,7 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public final class S3Constants {
   public static final String S3_CONTENT_LENGTH_HEADER = "Content-Length";
+  public static final String S3_ETAG_HEADER = "ETAG";
   public static final String S3_DATE_FORMAT_REGEXP = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
   public static final int S3_DEFAULT_MAX_KEYS = 1000;
   public static final String S3_EMPTY_ETAG = "";
--- a/core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
+++ b/core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
@@ -397,7 +397,7 @@ public final class S3RestServiceHandler {
           // TODO(cc): Consider how to respond with the object's ETag.
           return Response.ok()
               .lastModified(new Date(status.getLastModificationTimeMs()))
-              .header(S3Constants.S3_ETAG_HEADER, status.getLastModificationTimeMs())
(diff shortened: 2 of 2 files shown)


stuartsierra:
This Pull Request is sufficient to fix the specific issue with the Python s3fs library in #12212. However, I would recommend checking documentation about the `ETag` HTTP Header.

For example, [ETag - HTTP - MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) specifies that the value of `ETag` should be wrapped in double-quotation (`"`) marks, like this:

    ETag: "abcdef12345"

Wrapping the value in double-quotation marks should be sufficient for S3 clients that follow HTTP specifications.

However, some S3 clients **may** expect the ETag header to match the [behavior of Amazon S3 ETags](https://teppen.io/2018/06/23/aws_s3_etags/), which is more complex. I do not think it is necessary to fully emulate the behavior of S3 in this case, but it may be desirable in the future.

maobaolong:
@stuartsierra Thanks for your suggestion,  I've add double-quotation to the etag header, for the `matching the behavior of Amazon S3 ETags`, yeah, its true, let desirable in the future.

apc999 (approved):
Thanks for the fix!

apc999:

LuQQiu:

PR #12443, opened 2020-11-01
Change Default TTL Action from DELETE to FREE

### LINKED ISSUE #12316: can alluxio change default ttlaction from delete to free?
**Alluxio Version:
2.3

it is very dangerous for ttlaction default to be delete. ufs file may be deleted

cheyang:
fixes #12316


jja725:
close since duplicate #16823

PR #12445, opened 2020-11-02
Add support for speed check against ufs with hidden credentials

yuzhu:
Speed check can not be performed on a UFS right now because ufs credentials are masked when they are returned from the master.
This works around that issue by calling into the file master to launch a speed test which will run as a job.

### DIFF
--- a/core/server/master/pom.xml
+++ b/core/server/master/pom.xml
@@ -123,6 +123,12 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-stress-shell</artifactId>
+      <version>2.5.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/core/server/master/pom.xml
+++ b/core/server/master/pom.xml
@@ -97,6 +97,11 @@
       <artifactId>alluxio-job-client</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-stress-shell</artifactId>
+      <version>alluxio-stress-shell</version>
+    </dependency>
 
     <!-- External test dependencies -->
     <dependency>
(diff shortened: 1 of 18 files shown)


job/server/pom.xml
```diff
@@ -87,7 +87,7 @@
     </dependency>
     <dependency>
       <groupId>org.alluxio</groupId>
-      <artifactId>alluxio-stress-common</artifactId>
+      <artifactId>alluxio-stress-shell</artifactId>
```
madanadit, line 90: why is this needed?

core/server/master/pom.xml
```diff
@@ -97,6 +97,11 @@
       <artifactId>alluxio-job-client</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-stress-shell</artifactId>
```
madanadit, line 102: @yuzhu the flow as of now is the new master RPC depend on the shell module to execute commands which end up making RPCs to the job master process?

madanadit (commented):
Thanks @yuzhu. Like I mentioned before, the pom dependency feels weird but looks good otherwise. Requested review from Gene and Jiacheng to see if there is a better way.

core/server/master/pom.xml
yuzhu: yes

jiacheliu3: Yea currently this is the only way.

jiacheliu3 (approved):
LGTM except the questions on job server pom from Adit. Thanks for the improvement!

job/server/pom.xml
yuzhu: reverted

madanadit (approved):

gpang (commented):
@yuzhu I don't like that we have to add an RPC endpoint to FSM to just run this specific test. It feels too specific and too much like a hack. Instead, can we figure out how we can pass mount configuration to a stress job?

For example, is there something we can do in `StressBenchDefinition` to get and pass along mount config?

yuzhu:
@gpang I think the central issue here is that `UFSIOBench` directly operates on the `UnderFileSystem` class, and that's why it needs the credentials. I suppose we can read and write through the master, but that will surely involve other components of the alluxio master such as metadata ops and is not a pure UFS IO benchmark.

gpang:
I think JobWorkers do have access to the mount points, via JobServerContext and the UfsManager (https://github.com/Alluxio/alluxio/blob/d8193189a3de15cc01cfb315cb9c981b08b13e3e/job/server/src/main/java/alluxio/job/JobServerContext.java#L31)

We just need to figure out how that information can be propagated to the task itself.

yuzhu:
@gpang The task can get an instance of the `UfsManager`, but then it needs to do a shell exec to launch `UfsIOBench`, so this still wouldn't work. We need to somehow allow `UfsManager` to return credentials to make this work. so we can pass it to the shell execute.

yuzhu:
See #12480 for a version without POM changes and going through the FileSystemMaster

PR #12449, opened 2020-11-02
Log slow remote reads and writes

gpang:
For worker remote reads and writes, if a particular read or write takes longer than expected, we log (via the sampling logger) a message to give a sense of the delays.


core/server/worker/src/main/java/alluxio/worker/grpc/AbstractWriteHandler.java
```diff
@@ -59,6 +61,8 @@
 @NotThreadSafe
 abstract class AbstractWriteHandler<T extends WriteRequestContext<?>> {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractWriteHandler.class);
+  private static final Logger SLOW_WRITE_LOG = new SamplingLogger(LOG, 5 * Constants.MINUTE_MS);
+  private static final long SLOW_WRITE_MS = 10 * Constants.SECOND_MS;
```
calvinjia, line 65: Is there any value in making this configurable?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
```diff
@@ -58,6 +60,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(BlockReadHandler.class);
   private static final long UFS_BLOCK_OPEN_TIMEOUT_MS =
       ServerConfiguration.getMs(PropertyKey.WORKER_UFS_BLOCK_OPEN_TIMEOUT_MS);
+  private static final Logger SLOW_BUFFER_LOG = new SamplingLogger(LOG, 5 * Constants.MINUTE_MS);
```
calvinjia, line 63: For the sampling logger, is it possible to retain a count of how many were skipped due to sampling?

calvinjia (changes requested):

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
gpang: It would be possible, but I think that functionality should go into a separate PR.

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractWriteHandler.java
gpang: Yeah, I was torn about this. My concern was that there could be too many logging thresholds to configure, which could be confusing.

On the other hand, the hardcoded threshold may not be correct or reasonable.

I guess we could combine the read and write threshold into something like "worker.remote.read.slow.io.threshold"? That could handle both the read and write case?

What do you think?

gpang (commented):
@calvinjia responded

core/server/worker/src/main/java/alluxio/worker/grpc/AbstractWriteHandler.java
gpang: I added a config parameter.

calvinjia: Yeah, I think one threshold param would be sufficient.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
calvinjia: Makes sense

calvinjia (approved):

gpang:

gpang:

PR #12452, opened 2020-11-03
Improve hdfs validation task error messages

ZacBlanco:
Improvements now provide more context for a few cases
- insufficient permissions to access
- HA configured but not used (nameservices)
- more info about checks which require additional configuration
- advice about XML parsing

### DIFF
--- a/integration/tools/pom.xml
+++ b/integration/tools/pom.xml
@@ -42,6 +42,12 @@
       <version>${project.version}</version>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-shaded-hadoop</artifactId>
+      <version>${hadoop.version}</version>
+    </dependency>
   </dependencies>
 
 </project>
--- a/integration/tools/validation/src/main/java/alluxio/cli/UfsDirectoryValidationTask.java
+++ b/integration/tools/validation/src/main/java/alluxio/cli/UfsDirectoryValidationTask.java
@@ -22,8 +22,8 @@ import java.util.Map;
  */
 @ApplicableUfsType(ApplicableUfsType.Type.ALL)
 public final class UfsDirectoryValidationTask extends AbstractValidationTask {
-  private final UnderFileSystem mUfs;
   private final String mPath;
+  private final AlluxioConfiguration mConf;
 
   /**
    * Creates a new instance of {@link UfsDirectoryValidationTask}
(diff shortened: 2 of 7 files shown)


jiacheliu3 (approved):

yuzhu:
@ZacBlanco looks good, how were these tested? specifically the new validation test, does it pass locally? what about on a cluster?

ZacBlanco:
I didn't test any of these in a cluster. Specifically, I just generated the scenarios with the particular issues. I don't think they require a cluster to test.

- Insufficient permissions to access: I simply changed the permissions to `root:root` with permissions 700 and attempted to read
- HA configured but not used: I followed HDFS documentation for configuring HA with nameservices. I added the `dfs.ha.nameservices` parameter, and then passed an incorrect URI which doesn't contain one of the nameservices as the host field in the HDFS URI. e.g. `dfs.ha.nameservices=ns1` but the mount URI is `hdfs://hdfs-namenode-1:9000/`. This check fails if the URI host doesn't match any of the nameservices configured. I also tested this with `dfs.ha.nameservices` configured with multiple values: `ns1,ns2`.
- more info about checks requiring additional configuration. This doesn't need much testing. Behavior is the same, its just that the messages are updated with some more info.
- advice about XML parsing: this change just modifies the message when we receive an XML parsing exception

by the way, @jiacheliu3 is there a reason we don't just use the Configuration class from `org.apache.hadoop` to parse the configuration? I realize we currently don't have a dependency on it, but was avoiding the dependency the reason for parsing the XML ourselves?

jiacheliu3:
> * advice about XML parsing: this change just modifies the message when we receive an XML parsing exception

On XML parsing, I was not intentionally avoiding the dependency. When I was originally working on these tasks, they already have the XML parsing logic. So I didn't refactor by changing it to using hadoop's `Configuration` utility. In short I don't think anything stops us from using hadoop's `Configuration`.

yuzhu:
In that case, could we use hadoop's `Configuration` to parse? @ZacBlanco

ZacBlanco:
That's what I would prefer. I can try making the change here, but what we have seems to work - so I'm not super inclined to try to change it.

I'll do a POC on my local and if it seems easy to make the swap I'll do it, otherwise I'll response here for why I can't or don't want to move forward using the hadoop object.

ZacBlanco:
I made the update to the parser. The changes were pretty straightforward

ZacBlanco:
@yuzhu any more comments?

yuzhu (approved):
LGTM,  thanks

ZacBlanco:

yuzhu:

PR #12455, opened 2020-11-03
Change Default Forkcount for jenkins build to 1

bradyoo:


bradyoo:
This is just me naively following the advice of https://stackoverflow.com/questions/23260057/the-forked-vm-terminated-without-saying-properly-goodbye-vm-crash-or-system-exi

Though this may be a bad idea because it will certainly have an effect on these unit tests running right now. So checking how they run

dev/jenkins/build.sh
```diff
@@ -65,7 +65,7 @@ export MAVEN_OPTS="-Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simple
 
 if [ -z "${ALLUXIO_BUILD_FORKCOUNT}" ]
 then
-  ALLUXIO_BUILD_FORKCOUNT=4
+  ALLUXIO_BUILD_FORKCOUNT=1
```
Xenorith, line 68: do you need to set this env var to 4 in the github action yamls so they run with the same settings as before this change? i'm not sure how many builds run through this codepath without setting this env var; would need some time to comb through them all

Xenorith (approved):
i looked for other callers to this and didn't find any that would be affected. github actions also look good?

bradyoo:
github actions don't even hit this code.

bradyoo:

LuQQiu:

PR #12466, opened 2020-11-06
Prevent the grpc reader from waiting for a server response

gpang:
When a grpc reader is closing, there is no message that it is expecting from the server, however, it is currently waiting for some confirmation from the server. This can sometimes lead to an exception in the close() phase of the reader, which is unnecessary, since the client is completely finished with reading.

In this change, we avoid waiting long for the server when the reader is closing. If something does occur during the wait, we ignore it. This avoids any unnecessary timeouts or errors during close(), after the client is finished reading successfully.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
@@ -67,7 +67,8 @@ public class GrpcBlockingStream<ReqT, ResT> {
   private Throwable mError;
   /** This condition is met if mError != null or client is ready to send data. */
   private final Condition mReadyOrFailed = mLock.newCondition();
-  private boolean mClosedFromRemote = false;
+  /** This is set by the grpc threads, and checked/read by the client. */
+  private volatile boolean mClosedFromRemote = false;
 
   /**
    * @param rpcFunc the gRPC bi-directional stream stub function
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
@@ -27,8 +27,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
@@ -51,9 +51,9 @@ public class GrpcBlockingStream<ReqT, ResT> {
   /** Buffer that stores responses to be consumed by {@link GrpcBlockingStream#receive(long)}. */
   private final BlockingQueue<Object> mResponses;
   private final String mDescription;
-  private boolean mCompleted = false;
-  private boolean mClosed = false;
-  private boolean mCanceled = false;
+  private volatile boolean mCompleted = false;
+  private volatile boolean mClosed = false;
+  private volatile boolean mCanceled = false;
 
   /**
    * Uses to guarantee the operation ordering.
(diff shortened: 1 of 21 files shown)


core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
```diff
@@ -77,7 +78,9 @@
   public GrpcBlockingStream(Function<StreamObserver<ResT>, StreamObserver<ReqT>> rpcFunc,
       int bufferSize, String description) {
     LOG.debug("Opening stream ({})", description);
-    mResponses = new ArrayBlockingQueue<>(bufferSize);
+    // Use an unlimited queue to avoid blocking the network threads. Depend on custom flow
+    // control to limit the size of buffer.
+    mResponses = new LinkedBlockingQueue<>();
```
bf8086, line 83: Is the default buffer size (16) not large enough? With 2MB message size we are looking at 32MB buffer per reader, this can grow quickly if multi-thread reading is used.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
```diff
@@ -168,7 +169,19 @@ public void close() throws IOException {
...
+
+      // When a reader is closed, there is technically nothing the client requires from the server.
+      // However, the server does need to cleanup resources for a client close(), including closing
+      // or canceling any temp blocks. Therefore, we should wait for some amount of time for the
+      // server to finish cleanup, but it should not be very long (since the client is finished
+      // with the read). Also, if there is any error when waiting for the complete, it should be
+      // ignored since again, the client is completely finished with the read.
+      try {
+        // Wait a short time for the server to finish the close, and then let the client continue.
+        mStream.waitForComplete(5 * Constants.SECOND_MS);
```
bf8086, line 181: Some applications may depend on the closure to be done on the server side so they can perform other operations on the file safely. Having the reader skip the wait may cause subsequent operations to fail.

bf8086 (changes requested):
Thanks, I left a couple comments.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
gpang: I originally saw this grow to around 20-25 before.

I changed it back, but is it a bad thing to block the grpc thread? Because that will sometimes happen if the size is 16.

gpang (commented):
@bf8086

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcBlockingStream.java
bf8086: I think when a grpc thread is blocked it will potentially result in new thread being created for handling other concurrent RPC calls, so there is a potential overhead of one blocking thread per stream.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
gpang: If another client tries to read this block, there shouldn't be an issue, since they would only be grabbing a read lock, which should be allowed. I don't think another client can write lock the block, since Alluxio blocks are write-once-read-many. Can you think of any other problem behavior?

Ideally, I would have just removed this `waitForComplete` entirely, but there are some tests which require some async caching tasks to complete (even though async caching in general is best-effort). Because of these tests, waiting some short amount of time is beneficial for the tests.

gpang (commented):
@bf8086

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
bf8086: One possible issue is when the same client which just closes a stream tries to do another write operation such as rename or delete. If the stream close function does not wait until server release the lock, the subsequent write operation may fail when attempting to acquire a lock. This breaks the file stream contract that many applications assumes.

gpang:
@bf8086 I updated this PR with a client parameter for the reader close wait.

bf8086 (approved):
LGTM.

gpang:

gpang:

PR #12467, opened 2020-11-06
Log Some Job Failure Stacktraces with sampling logger

bradyoo:


bradyoo:
@gpang ping?

gpang (commented):
@bradyoo I have a few high-level questions:
- Do we want it logged on the server, or inserted into the error message, or both?
- If we only want it logged on the server, we could just use the sampling logger, right?

bradyoo:
@gpang Changed it to sampling log. It's a lot simpler.

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
```diff
@@ -201,6 +203,7 @@ public synchronized void notifyTaskFailure(long jobId, long taskId, Throwable t)
     }
     finishTask(id);
     LOG.info("Task {} for job {} failed: {}", taskId, jobId, errorMessage);
+    SAMPLING_LOGGER.info(Throwables.getStackTraceAsString(t));
```
gpang, line 206: I think this message should have taskId and jobId, so we know what the stack is for?

Something like:
```java
SAMPLING_LOGGER.info("Stack trace for taskId: {} jobId: {} : {}", taskId, jobId, Throwables.getStackTraceAsString(t));```

gpang (changes requested):
@bradyoo

job/server/src/main/java/alluxio/worker/job/task/TaskExecutorManager.java
bradyoo: Thanks. Done.

gpang (approved):
LGTM

gpang:

PR #12468, opened 2020-11-06
Decrease default job retention time

bradyoo:
Note that job retention time only kicks in when the JobMaster is at capacity in terms of number of jobs stored.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -4468,7 +4468,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       new Builder(Name.JOB_MASTER_FINISHED_JOB_RETENTION_TIME)
           .setDescription("The length of time the Alluxio Job Master should save information about "
               + "completed jobs before they are discarded.")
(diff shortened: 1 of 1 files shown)


gpang:
@bradyoo What is the benefit of this? For reducing mem usage?

If I recall, I think this retention was for the clients to poll to check if the job was completed or not. If the job status is removed before the client can check, the client will assume it failed. This may not matter most of the time, but on failover, there could be a long period of time for the secondary to start checking with the job master again.

bradyoo:
@gpang If there is a master failover. Job Master will throw out all of the existing jobs. For the case of Master checking in the JobMaster for completion of persist, it shouldn't really matter too much because they could assume the job has failed if the job doesn't exist. (I believe it already does that.)

There is no memory implication here. The design of job service is that if the capacity is full, it will reject any attempt at creating a new job. So what currently happens is that when a customer tries to DistributedLoad a bunch of small files and they complete really quickly, they eventually run into a limit because there are too many finished jobs that hasn't hit the 5 minute mark of retention.

bradyoo:
@gpang ping?

gpang (approved):
LGTM

gpang:

PR #12474, opened 2020-11-09
Add unit test for SharedGrpcDataReader

LuQQiu:
Add unit test for SharedGrpcDataReader.
The SharedGrpcDataReader should be thread-safe and supports multi-threads read from a single Alluxio client (not only bound to the fuse case)

### DIFF
--- a/core/client/fs/src/test/java/alluxio/client/block/stream/SharedGrpcDataReaderTest.java
+++ b/core/client/fs/src/test/java/alluxio/client/block/stream/SharedGrpcDataReaderTest.java
@@ -1,0 +1,421 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.client.block.stream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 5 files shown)


core/client/fs/src/test/java/alluxio/client/block/stream/SharedGrpcDataReaderTest.java
```diff
@@ -0,0 +1,421 @@
...
+    when(mContext.getClientContext())
+        .thenReturn(ClientContext.create(ConfigurationTestUtils.defaults()));
+    when(mContext.getClusterConf()).thenReturn(ConfigurationTestUtils.defaults());
+    mAddress = mock(WorkerNetAddress.class);
+    mClient = mock(BlockWorkerClient.class);
+    mRequestObserver = mock(ClientCallStreamObserver.class);
+    when(mContext.acquireBlockWorkerClient(mAddress)).thenReturn(
+        new NoopClosableResource<>(mClient));
+    when(mClient.readBlock(any(StreamObserver.class))).thenReturn(mRequestObserver);
+    when(mRequestObserver.isReady()).thenReturn(true);
```
apc999, line 86: in general for unit test, I would not rely on mockito because the UT code becomes hard to maintain. For example, if we change the internal implementation of context or grpc streaming (which are totally internal and transparent to the class we are testing), it requires a lot of change here.

Here is how I would write this unit test. We can create a Mock class extending `BufferCachingGrpcDataReader`, and pass this mock reader to a `SharedGrpcDataReader` instance.  In this mock reader, we can return certain data, e.g., predefined for a block. And check the results on `SharedGrpcDataReader`.

This design pattern to inject a mock argument is more maintainable. Because you can focus on the logic of `SharedGrpcDataReader` that you want to test, without worrying about future change of `BufferCachingGrpcDataReader`. Let me know if I make myself clear. I am happy to write some example code too if that helps.

apc999 (changes requested):
A general comment is to write and use mock class, rather than replying on reflection (mockito) for the tests. In this way, the Unit test can be more maintainable.

LuQQiu:
This PR is replaced by https://github.com/Alluxio/alluxio/pull/12847

PR #12486, opened 2020-11-11
Use lower tiers in workers based on tier usage

jiacheliu3:
This is a probabilistic approach on when we decide to assign new blocks to lower tiers.

Before we only assign new blocks to lower tiers when tier 0 cannot hold the initial size of a block (by default 1MB or 2MB). This leads to the issue that most blocks are assigned to tier 0 and expand there, refusing to go to lower tiers and freeing space on tier 0. Consequently we have poor utilization in lower tiers.

In this change, we define heuristics that we reject allocation to certain storage dirs if they are close to full. The closer to full it is, the lower the chance that a block is allocated to it (and grows there subsequently). If an allocation is rejected, we move on to trying lower tiers and other storage dirs until we find a place.


core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -705,6 +705,10 @@ private TempBlockMeta createBlockMetaInternal(long sessionId, long blockId, bool
       }
 
       // Allocate space.
+      LOG.warn("createBlockMetaInternal allocating space with options {}", options);
+      LOG.warn("Allocate more than necessary to force allocation go to lower tiers.");
```
jiacheliu3, line 709: These are debug loggings for myself, as I'm testing the PR in various ways.

core/server/worker/src/main/java/alluxio/worker/block/allocator/RoundRobinAllocator.java
```diff
@@ -79,31 +86,56 @@ private StorageDirView allocateBlock(long sessionId, long blockSize,
...
+                    candidateDir.toBlockStoreLocation());
+          }
+        }
+        // Either the dirViewIndex is -1, or we rejected the allocation to a dirViewIndex > 0
+        offset = dirViewIndex;
       }
     } else if (location.equals(BlockStoreLocation.anyDirInTierWithMedium(location.mediumType()))) {
-      String medium = location.mediumType();
-      int tierIndex = 0; // always starting from the first tier
-      for (int i = 0; i < mMetadataView.getTierViews().size(); i++) {
```
jiacheliu3, line 98: The previous code was wrong -- the quest is `anyDirInTierWithMedium` but here it is checking every tier. Maybe I should fix this in a separate PR.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticReviewer.java
```diff
@@ -0,0 +1,56 @@
...
+      }
+      // 2 points:
+      // Axis X: space usage (commitment)
+      // Axis Y: Probability of using this tier
+      // (1.0 - CUTOFF, 1.0)
+      // (1.0, 0)
+      double k = -1.0 / CUTOFF; // What if CUTOFF == 0?
+      double b = 1.0 / CUTOFF;
+      double y = k * usage + b;
+      LOG.warn("Space usage in tier {} is {}. Probability of staying is {}.", dirView.getParentTierView().getTierViewAlias(), usage, y);
```
jiacheliu3, line 49: Please ignore this kind of loggings. They are for my own testing for the moment. They will be removed/formatted at the end of this work.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/Reviewer.java
```diff
@@ -0,0 +1,36 @@
...
+  boolean reviewAllocation(StorageDirView dirView);
+
+  /**
+   * Factory for {@link Reviewer}.
+   */
+  class Factory {
+
+    private Factory() {} // prevent instantiation
+
+    // TODO(jiacheng): Is it possible we need a list of reviewers in the future, each checking one criteria?
```
jiacheliu3, line 24: @gpang Do you think it's possible that we may want a pipeline of Reviewers, instead of having only one?

core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticReviewer.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Reviews a block allocation, and probably reject a bad block allocation.
+ * The more full a StorageDir, the higher the chance that this reviewer decides
+ * not to use it for a new block.
+ * */
+public class ProbabilisticReviewer implements Reviewer {
+    private static final Logger LOG = LoggerFactory.getLogger(ProbabilisticReviewer.class);
+
+    // TODO(jiacheng): Make this a property
+    private static final double CUTOFF = 0.1;
```
jiacheliu3, line 20: 0.1 is probably too big, like when we have 1TB for a storage dir, that is too far from a block size. 
64MB / 1TB = 0.00006. I don't think we want to assign this small decimal to a property.
If a `long` in bytes make more sense, maybe we can reuse `PropertyKey.WORKER_TIERED_STORE_FREE_AHEAD_BYTES`?

`WORKER_TIERED_STORE_FREE_AHEAD_BYTES` is a property we added in 2.3 together with the new CACHE_PROMOTE. I don't think many people use that. I also don't think the property is very useful at the moment. It only controls how much extra we free, on each `TieredBlockStore.freeSpace`.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticReviewer.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public boolean reviewAllocation(StorageDirView dirView) {
+      long availableBytes = dirView.getAvailableBytes();
+      long capacityBytes = dirView.getCapacityBytes();
+      long blockSize = ServerConfiguration.global().getBytes(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT);
+      long cutoffSize = (long) Math.floor(CUTOFF * capacityBytes);
+      // Rules:
+      // 1. If more than CUTOFF left, we use this tier. Prob=100%
+      // 2. If less that block size left, we ignore this tier. Prob=0%
+      // 3. If in the middle, the probability is linear to the space left, the less space the lower.
```
jiacheliu3, line 33: @gpang Does this heuristic of probability make sense to you? Do you think instead of a linear function, something quadratic or logistic make more sense to make the probability drop quicker?

core/server/worker/src/main/java/alluxio/worker/block/reviewer/Reviewer.java
gpang: Hrmmm, I'm not sure. I don't know what a Reviewer is yet.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticReviewer.java
gpang: I think it makes sense, but we could just make that factor a parameter. Something like 
```
((cutoff - free) / cutoff) ^ alpha
```

if `alpha` is 1, it is linear, and the user can adjust the curve by adjusting `alpha`. Maybe there are nicer curves out there, but this is advanced and may not be necessary.

gpang (commented):
@jiacheliu3

jiacheliu3:
I'm doing the formatting and cleanup on this PR. This will be in a ready-for-review state today.

core/server/worker/src/main/java/alluxio/worker/block/allocator/RoundRobinAllocator.java
jiacheliu3: This is addressed separately in https://github.com/Alluxio/alluxio/pull/12509

core/server/worker/src/main/java/alluxio/worker/block/evictor/AbstractEvictor.java
```diff
@@ -217,8 +218,7 @@ public EvictionPlan freeSpaceWithView(long bytesToBeAvailable, BlockStoreLocatio
   protected void onRemoveBlockFromIterator(long blockId) {}
 
   /**
-   * Updates the block store location if the evictor wants to free space in a specific location. For
-   * example, {@link PartialLRUEvictor} always evicts blocks from a dir with max free space.
+   * Updates the block store location if the evictor wants to free space in a specific location.
```
jiacheliu3, line 221: Removed reference to an obsolete class

core/server/worker/src/test/java/alluxio/worker/block/allocator/AllocatorTestBase.java
```diff
@@ -157,4 +162,48 @@ protected void assertTempBlockMeta(Allocator allocator, BlockStoreLocation locat
   protected BlockMetadataEvictorView getMetadataEvictorView() {
     return new BlockMetadataEvictorView(mManager, new HashSet<Long>(), new HashSet<Long>());
   }
+
+  /**
```
jiacheliu3, line 166: These tests are shared by all `Allocator` subclasses

core/server/worker/src/test/java/alluxio/worker/block/reviewer/ProbabilisticBufferReviewerTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @After
+  public void reset() {
+    mReviewer = null;
+    ServerConfiguration.reset();
+  }
+
+  @Test
+  public void testProbabilityFunction() throws Exception {
+    // Empty - 100%
```
jiacheliu3, line 57: See how the probability function works here

core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticReviewer.java
jiacheliu3: Done. Please see if the new properties make sense. Thanks!

core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticReviewer.java
jiacheliu3: Agreed there are nicer curves but may not be necessary. After all the difference would be just like 1 block. I think we can just start with a linear function. I will try out higher rank functions like quadratic to the space left in the near future, but not in this PR.

core/server/worker/src/main/java/alluxio/worker/block/allocator/RoundRobinAllocator.java
```diff
@@ -70,43 +77,43 @@ public StorageDirView allocateBlockWithView(long sessionId, long blockSize,
...
-      String medium = location.mediumType();
-      int tierIndex = 0; // always starting from the first tier
+                location.mediumType()))) {
       for (int i = 0; i < mMetadataView.getTierViews().size(); i++) {
-        StorageTierView tierView = mMetadataView.getTierViews().get(tierIndex);
-        int dirViewIndex = getNextAvailDirInTier(tierView, blockSize, medium);
+        StorageTierView tierView = mMetadataView.getTierViews().get(i);
+        // The review logic is handled in getNextAvailDirInTier
+        int dirViewIndex = getNextAvailDirInTier(tierView, blockSize,
+                location.mediumType(), skipReview);
```
jiacheliu3, line 108: I refactored this a bit to remove redundant local variables. I believe the logic are kept the same but please do take an extra look for me. Thanks!

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2881,6 +2881,52 @@ public String toString() {
...
+              + "in the Allocator. Each time a block allocation decision is made by "
+              + "the Allocator, the Reviewer will review the decision and rejects it,"
+              + "if the allocation does not meet certain criteria of the Reviewer."
+              + "The Reviewer prevents the worker to make a bad block allocation decision."
+              + "Valid options include:"
+              + "`alluxio.worker.block.reviewer.ProbabilisticBufferReviewer`.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_REVIEWER_ENABLED =
```
jiacheliu3, line 2923: I had to add this property because many integration tests on tier storage depend on the deterministic behavior of the `Allocator`. Another option is to use a `NoopReviewer` that never rejects an allocation.

calvinjia: I'm in favor of the pass through reviewer

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -5279,6 +5325,12 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.worker.block.master.client.pool.size";
     public static final String WORKER_PRINCIPAL = "alluxio.worker.principal";
     public static final String WORKER_RAMDISK_SIZE = "alluxio.worker.ramdisk.size";
+    public static final String WORKER_REVIEWER_BUFFER_HARDLIMIT_BYTES =
+            "alluxio.worker.reviewer.buffer.hardlimit.bytes";
```
calvinjia, line 5329: Is the namespace for this property correct? Since it seems to only affect the probabilistic reviewer.

Probably not something we need to address in this PR - if users are plugging in their own reviewers, how will they add configuration options?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -633,18 +633,20 @@ private StorageDirView allocateSpace(long sessionId, AllocateOptions options) {
...
-
+          options.getLocation(), allocatorView, false);
       if (dirView != null) {
         return dirView;
       }
 
       if (options.isForceLocation()) {
         if (options.isEvictionAllowed()) {
+          LOG.debug("Free space for block expansion: freeing {} on {}. ",
+                  options.getSize(), options.getLocation());
```
calvinjia, line 644: Will this be extremely verbose?

calvinjia (changes requested):
A high level comment - it seems like the block reviewer is tightly coupled with the Allocator. Is there a way to decouple them such that allocators do not need to have code to handle the reviewer?

tests/src/test/java/alluxio/server/tieredstore/SpecificTierWriteIntegrationTest.java
```diff
@@ -81,6 +81,7 @@
               LOW_WATERMARK)
           .setProperty(PropertyKey.USER_BLOCK_WRITE_LOCATION_POLICY,
               LocalFirstPolicy.class.getTypeName())
+          .setProperty(PropertyKey.WORKER_REVIEWER_ENABLED, String.valueOf(false))
```
gpang, line 84: maybe for integration tests, we should just set the setting as disabled (or noop reviewer). I think that is in `ConfigurationTestUtils`

gpang (changes requested):
@jiacheliu3 Thanks! I left a few comments.

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: By the `buffer` i mean the `ProbabilisticBufferReviewer`, but I see it can be confusing due to the multiple meanings of `buffer`. On second thought I'm changing it to `alluxio.worker.reviewer.probabilistic.hardlimit.bytes` to make the reference to `ProbabilisticBufferReviewer` clearer.

For plugging in own Reviewer implementation, that would be `alluxio.worker.reviewer.<foo>.<property key>`. Does that make more sense to you?

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
```diff
@@ -53,13 +54,22 @@ public static Allocator create(BlockMetadataView view) {
...
+   * The parameter skipReview specifies whether the review should be skipped.
+   * The review should be skipped when we want the behavior of the allocator to be deterministic
+   * and do not want it to be affected by the reviewer.
+   * E.g. We just freed up some space in Alluxio and want the allocation to use the freed space.
+   *
    * @param sessionId the id of session to apply for the block allocation
    * @param blockSize the size of block in bytes
    * @param location the location in block store
    * @param view of the block metadata
+   * @param skipReview whether the review should be skipped
```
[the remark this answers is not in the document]
jiacheliu3: Per our discussion, I added a TODO.

Under the current implementation of the `Allocator` class, the Allocator loops the `StorageDir` in an order it defines and makes allocation decisions on the way. When the `Reviewer` rejects an allocation, by design it should go back to the looping and try the next dir. This logic cannot be very easily extracted and hence decouple Allocator from Reviewer.

For the refactor (TODO), it makes more sense if the Allocator gives a list of candidate `StorageDir` in the order that it defines, then another coordinator (or maybe the `Reviewer` itself) looks at the candidates and makes the call.

tests/src/test/java/alluxio/server/tieredstore/SpecificTierWriteIntegrationTest.java
jiacheliu3: Added. For the tests i modified, they are not using `ConfigurationTestUtils.defaults()`, but this will surely help the integration tests which do.

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: Added `AcceptingReviewer`

core/common/src/test/java/alluxio/ConfigurationTestUtils.java
```diff
@@ -75,6 +75,10 @@ public static InstancedConfiguration defaults() {
       conf.put(PropertyKey.Template.WORKER_TIERED_STORE_LEVEL_DIRS_PATH.format(level),
           Joiner.on(',').join(newPaths));
     }
+
+    // Sets up the block allocation and review policy
+    conf.put(PropertyKey.WORKER_REVIEWER_CLASS, "alluxio.worker.block.reviewer.AcceptingReviewer");
```
jiacheliu3, line 80: This will help integration tests that use `defaults()` to avoid the non-deterministic block allocation reviewing

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Ah okay, I didn't make the connection, this is much clearer. My question about plugins is more around the fact that they would need to add the property key to our `PropertyKey.java` file which requires recompiling.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
jiacheliu3: I think this should be okay. At least for my case, when we turn on debug log for `TieredBlockStore` there are something going wrong with tiered allocation, and space freeing is definitely a very critical part that we want to look into. Plus we don't free space on every buffer flush (every `freeSpace()` cleans up at least one block), so I think it's okay.

calvinjia (approved):
I think the TODO for refactoring is quite important, and this should be an experimental feature for now so we don't need to respect the reviewer API currently provided (if someone were to implement their own).

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: I see, that's a very good question. I did a random search for our policy configuration like `Allocator`, `Evictor` and `BlockLocationPolicy`. The implementations seem to be using hard coded parameters so I don't find an existing great way of plugging in customized properties w/o recompiling. If you are not aware of an existing solution either, I think we'll need to define how that's done.

jiacheliu3:
> I think the TODO for refactoring is quite important, and this should be an experimental feature for now so we don't need to respect the reviewer API currently provided (if someone were to implement their own).

Added a note the API is currently experimental

LuQQiu:

LuQQiu:

PR #12501, opened 2020-11-13
Add UFS version validation task

ZacBlanco:
This new task helps expose some more information about available UFS versions to users. The cases for configuration and corresponding outputs are defined below.

- UFS version not configured: skip validation
- UFS version configured, but doesn't exist in list of valid versions: error with info and list of available versions
- UFS version configured, but list of versions empty: this indicates the UFS doesn't support versioning (i.e. non-HDFS), and directs the user to change configuration
- UFS version configured and exists in available version list: success

### DIFF
--- a/integration/tools/validation/src/main/java/alluxio/cli/UfsVersionValidationTask.java
+++ b/integration/tools/validation/src/main/java/alluxio/cli/UfsVersionValidationTask.java
@@ -1,3 +1,14 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
 package alluxio.cli;
 
 import alluxio.conf.AlluxioConfiguration;
--- a/integration/tools/validation/src/main/java/alluxio/cli/UfsVersionValidationTask.java
+++ b/integration/tools/validation/src/main/java/alluxio/cli/UfsVersionValidationTask.java
@@ -1,0 +1,74 @@
+package alluxio.cli;
+
+import alluxio.conf.AlluxioConfiguration;
+import alluxio.conf.PropertyKey;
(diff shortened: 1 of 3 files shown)


yuzhu:
@ZacBlanco license header?

integration/tools/validation/src/main/java/alluxio/cli/UfsVersionValidationTask.java
```diff
@@ -0,0 +1,74 @@
...
+    List<String> availableVersions =
+        UnderFileSystemFactoryRegistry.getSupportedVersions(mUfsPath, ufsConf);
+    ValidationTaskResult result = new ValidationTaskResult();
+    result.setName(getName());
+    result.setDesc("Validates that the configured UFS version exists as a library on the "
+        + "system.");
+
+    if (!mConf.isSetByUser(PropertyKey.UNDERFS_VERSION)) {
+      result.setState(ValidationUtils.State.SKIPPED);
+      result.setOutput("The UFS version was not configured by the user.");
```
jiacheliu3, line 46: Do you think it's possible there's a case where the user should have configured the `UNDERFS_VERSION` but did not?

jiacheliu3 (changes requested):
@ZacBlanco Thanks! I left a question.

integration/tools/validation/src/main/java/alluxio/cli/UfsVersionValidationTask.java
ZacBlanco: Users should only configure UFS version in the case that UFSes support a version. In our current implementation, Alluxio doesn't require a version to be specified, even if there is more than one version supported. It just picks a default if it is not configured. So, it is not necessary for a ever configure a version for a UFS to be chosen.

However, there is a a chance that the default version used is _not_ compatible with the particular UFS. This may induce runtime errors upon connection or data transfer. In those cases we defer to the other UFS checks to flesh our those errors. This check is designed just to make sure that the configured version corresponds to a valid Alluxio UFS library. There's no good way to detect is the UFS is _actually_ compatible at runtime. And even if it's not, any errors that occur likely won't say anything along the line of "hey, you're using the wrong client version". It will be much more difficult to decipher.

tl;dr yes, it is possible but determining the case in which the UFS version is wrong is a non-trivial problem

jiacheliu3: I see, that's fair to say. (Should tldr be on the top?)

jiacheliu3 (approved):

yuzhu (approved):
LGTM

yuzhu:

yuzhu:

PR #12510, opened 2020-11-16
Support using IP as connect host

### LINKED ISSUE #12420: Alluxio client accesses workers via IP address as an alternative when hostname cannot be resolved
Alluxio client can only get access to an alluxio worker via the according hostname of the worker. However when the mapping relation of hostname and IP address is not specified in the /etc/hosts of the worker machine or a DNS server, an **Unknownhost Exception** will be thrown. The problem occurs when I run a presto cluster on alluxio with the hosts mapping relation not specified in the hosts file. The alluxio client embedded in presto worker can not resolve the hostname returned from the alluxio master, then the query submitted to presto failed.

Let alluxio client can access workers via IP address as an alternative when the hostname cannot be resolved.

To solve the problem, I asked the Administrator to modify the /etc/hosts provisionally, and it works well. But it  is not a perfect way since once new workers are added to the cluster, such operation is required. And deploying a DNS server is hard to implement as well. 

**Urgency**
High

**Additional context**
Such feature can be switched on by a configuration property.

liupan664021:
Fix #12420 When hostname & bind.host not specified, local hostname is chosen as the connect host of the service by default, which may cause an UnknownHostException when the **LOCAL** service hostname cannot be resolved by a **REMOTE** client. This problem can be avoided by specifying alluxio.<service_name>.hostname or alluxio.<service_name>.bind.host using an IP address. But it is really troublesome when deploying lots of nodes with different IP addresses. 

Add alluxio.service.use.ip to support using ip as the connect host when enable.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -439,6 +439,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.ALL)
           .build();
+  public static final PropertyKey SERVICE_USE_IP =
+      new Builder(Name.SERVICE_USE_IP)
+          .setDefaultValue("false")
+          .setDescription("If true, when alluxio.<service_name>.hostname and "
+              + "alluxio.<service_name>.bind.host of a service not specified, "
+              + "use IP as the connect host of the service.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.ALL)
+          .build();
   public static final PropertyKey TEST_MODE =
       new Builder(Name.TEST_MODE)
           .setDefaultValue(false)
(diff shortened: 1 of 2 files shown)


maobaolong (approved):
Thanks for this contribution, LGTM.

maobaolong:
Hi @apc999 @gpang Would you please take a look at this PR? Thanks.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4770,6 +4779,7 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String NETWORK_HOST_RESOLUTION_TIMEOUT_MS =
         "alluxio.network.host.resolution.timeout";
     public static final String SITE_CONF_DIR = "alluxio.site.conf.dir";
+    public static final String SERVICE_USE_IP = "alluxio.service.use.ip";
```
apc999, line 4782: name suggestion: 
`alluxio.network.ip.address.used`

apc999 (approved):
One comment on property naming left. otherwise LGTM. 

Thanks for improving the network address resolution!

core/common/src/main/java/alluxio/conf/PropertyKey.java
liupan664021: Thanks for the name suggestion!

liupan664021:
Hi @apc999 The suggested property name is updated. Can you take a look? Thanks!

apc999:

Issue #12444, opened 2020-11-02
 AsyncRemover thread stuck


iAlaska:
**Alluxio Version:**
We use both 2.2.0 and 2.3.0

**Describe the bug**
Below is  our use scenario：
1.Mount some hdfs directory (contains time based partitions) on alluxio
2.Maintain a time window that every day, Load newest partition into alluxio and Free oldest partition out of alluxio

After some Maintain cycle，all of AsyncRemover thread stuck at acquiring ClientRWLock
```
"block-removal-service-9" #2184 daemon prio=5 os_prio=0 tid=0x00007fe336e9d000 nid=0x340399 waiting on condition [0x00007fdae7cfb000]
   java.lang.Thread.State: WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x000000065d50c5b8> (a java.util.concurrent.Semaphore$NonfairSync)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireShared(AbstractQueuedSynchronizer.java:967)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(AbstractQueuedSynchronizer.java:1283)
        at java.util.concurrent.Semaphore.acquireUninterruptibly(Semaphore.java:494)
        at alluxio.worker.block.ClientRWLock$SessionLock.lock(ClientRWLock.java:92)
        at alluxio.worker.block.BlockLockManager.lockBlock(BlockLockManager.java:114)
        at alluxio.worker.block.TieredBlockStore.removeBlock(TieredBlockStore.java:378)
        at alluxio.worker.block.TieredBlockStore.removeBlock(TieredBlockStore.java:371)
        at alluxio.worker.block.DefaultBlockWorker.removeBlock(DefaultBlockWorker.java:477)
        at alluxio.worker.block.AsyncBlockRemover$BlockRemover.run(AsyncBlockRemover.java:106)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
 
[...]

PR #12519, opened 2020-11-17
Add block lock cleaner

LuQQiu:
One of the potential fix/improve of #12444 and #11309 

Before Alluxio 1.5, Alluxio has a `session cleaner`, each session(client) will do session heartbeat to show session active. When a session timeout, all locks and temporary blocks holding by this session will be cleanup. After changes in #5377, session cleaner is not working anymore. #12444 and #11309  show a potential lock leak happening. When block lock leaks, we currently have no way to clean up those locks which may block our async block removers and space allocators.

This PR makes the following changes
1. Remove the `session id` concept in `BlockLockManager`. Session id concept is still useful for cleaning the temporary blocks when writing blocks, but is useless and can be removed in other places.
2. Add a `lastAccessTime` info in the block lock record. If a block lock hasn't been accessed (through `blockLockHeartbeat` or `validateLock`) for a period of time (5 min by default in my pr), it will be cleaned up when another write lock block request of the same block is coming in. In other words, when we want to write lock a block, we first clean up all the outdated locks of this block.
3. Add `blockLockHeartbeat` in the `BlockDataReader.openBlock()`. So that when a client is reading a block chunk data, it will first update the `lastAccessTime` of that lock. Making sure that we don't delete in-use locks in unexpected long block remote read process.

The potential risk of this fix:
1.  Mistakenly delete in-use locks. `ShortCircutBlockReadHandler` doesn't do `blockLockHeartbeat`. If a `ShortCircutBlockReadHandler` doesn't finish within the timeout (5min by default), it may error out because the block already been removed. If other master operations like `moveBlock, commitBlock, removeBlock` take longer than the timeout, it may also error out.

Things left for future PRs:
1. Remove the session id from places that not need it.
2. How to cleanup temporary blocks of lost sessions

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2444,6 +2444,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_BLOCK_LOCK_TIMEOUT =
+      new Builder(Name.WORKER_BLOCK_LOCK_TIMEOUT)
+          .setDefaultValue("3min")
+          .setDescription("The timeout value of block locks lifetime. If a block lock "
+              + "is holding for more than the given timeout without updates, "
+              + "the lock will be invalidated to not block other block write operations.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
   public static final PropertyKey WORKER_CONTAINER_HOSTNAME =
       new Builder(Name.WORKER_CONTAINER_HOSTNAME)
           .setDescription("The container hostname if worker is running in a container.")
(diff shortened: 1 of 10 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2444,6 +2444,15 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_BLOCK_LOCK_TIMEOUT =
+      new Builder(Name.WORKER_BLOCK_LOCK_TIMEOUT)
+          .setDefaultValue("3min")
```
gpang, line 2449: Does this mean all block reads or writes must complete within this time? In high load situations, I have seen remote reads of a block take almost 30 minutes to finish reading. In this scenario, what would happen?

gpang (commented):
@LuQQiu I left a question

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: I add a lockHeartbeat (basically updateLockAccessTime) in the BlockReadHandler. Every read from remote with update the lock access time. But yeah, for the `ShortCurcuitReadHandler`, I don't have ways to track the locks, so it may be better to set this value to be bigger. If `ShortCircuitBlockReadHandler` takes too long, they may be error out as blocks are removed..

LuQQiu:
@apc999 @gpang I updated the PR and its description, PTAL. Test changes will be updated after we reach a consensus on this lock cleaner approach.Thanks

apc999:
@LuQQiu  is this PR still needed?

LuQQiu:
This PR is not needed

PR #12566, opened 2020-11-21
[DOCFIX] Add Allocator & Reviewer sections to Caching.md

jiacheliu3:
This change:
1. Adds Allocator as a section. This is an open interface configuring the allocation strategy, but is not documented.
1. Adds Reviewer as a section. This is a new interface added in 2.4.1.
1. Fixes multiple typos and updates a few doc link formats.

### DIFF
--- a/docs/en/core-services/Caching.md
+++ b/docs/en/core-services/Caching.md
@@ -144,7 +144,7 @@ The user can also specify the tier that the data will be written to via
 
 If the data is already in Alluxio, the client will simply read the block from where it is already stored.
 If Alluxio is configured with multiple tiers, the block may not be necessarily read from the top tier,
-since it could have been moved to a lower tier transparently. 
+since it may have been loaded into a lower tier, or moved to a lower tier transparently. 
 This logic applies to both `ReadType.CACHE` and `ReadType.CACHE_PROMOTE`.
 
 The difference is, reading data with `ReadType.CACHE_PROMOTE` will attempt to first transfer the block to the
@@ -152,6 +152,7 @@ top tier before it is read from the worker. This can also be used as a data mana
 explicitly moving hot data to higher tiers.
 
 For `ReadType.CACHE`, Alluxio will cache the block into the highest tier that has the available space.
+So you will read the cache block with disk speed if the block is currently on a disk(SSD/HDD).
 
 > Note: In 2.3, Alluxio default ReadType is changed from CACHE_PROMOTE to CACHE.
 
--- a/docs/en/core-services/Caching.md
+++ b/docs/en/core-services/Caching.md
@@ -308,7 +308,7 @@ This allows writing data bigger than Alluxio storage capacity. However, it also 
 To enforce the assumption that tiers are configured from fastest to slowest, Alluxio now moves blocks around tiers based on block annotation policies.
 
 Below configuration is honoured by each individual tier management task:
-- `alluxio.worker.management.task.thread.count`: How many threads to use for management tasks. (Default:`CPU core count`)
+- `alluxio.worker.management.task.thread.count`: How many threads to use for management tasks. (Default: `CPU core count`)
 - `alluxio.worker.management.block.transfer.concurrency.limit`: How many block transfers can execute concurrently. (Default:`CPU core count`/ 2)
 
 #### Block Aligning (Dynamic Block Placement)
--- a/docs/en/core-services/Caching.md
+++ b/docs/en/core-services/Caching.md
@@ -154,7 +154,11 @@ explicitly moving hot data to higher tiers.
 For `ReadType.CACHE`, Alluxio will cache the block into the highest tier that has the available space.
 So you will read the cache block with disk speed if the block is currently on a disk(SSD/HDD).
 
-> Note: In 2.3, Alluxio default ReadType is changed from CACHE_PROMOTE to CACHE.
+> Note: In 2.3, Alluxio default ReadType was changed from `CACHE_PROMOTE` to `CACHE`.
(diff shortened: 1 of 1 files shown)


docs/en/core-services/Caching.md
```diff
@@ -144,12 +144,18 @@ The user can also specify the tier that the data will be written to via
...
 
-Reading data with the `ReadType.CACHE_PROMOTE` will attempt to first transfer the block to the
+The difference is, reading data with `ReadType.CACHE_PROMOTE` will attempt to first transfer the block to the
 top tier before it is read from the worker. This can also be used as a data management strategy by
 explicitly moving hot data to higher tiers.
 
+For `ReadType.CACHE`, Alluxio will cache the block into the highest tier that has the available space.
+So you will read the cache block with disk speed if the block is currently on a disk(SSD/HDD).
+
+> Note: In 2.3, Alluxio default ReadType is changed from CACHE_PROMOTE to CACHE.
```
jiacheliu3, line 157: @gpang  If `CACHE_PROMOTE` moves the block up to MEM tier so only the first read on the existing block is disk space and subsequent ones are supposed to be memory speed, does that imply `CACHE_PROMOTE` will be faster in some scenarios? I don't find where we talk about that.

docs/en/core-services/Caching.md
```diff
@@ -205,6 +211,46 @@ A typical configuration will have three tiers for Memory, SSD, and HDD.
...
+new blocks into this directory probabilistically. The probability is determined by `alluxio.worker.reviewer.probabilistic.hardlimit.bytes`
+and `alluxio.worker.reviewer.probabilistic.softlimit.bytes`. When the available space in the directory is under
+`alluxio.worker.reviewer.probabilistic.hardlimit.bytes`, which is `64MB` by default, new blocks will be rejected.
+When the available space in the directory is above `alluxio.worker.reviewer.probabilistic.softlimit.bytes`, which is
+`256MB` by default, new blocks will NOT be rejected. When the available space is between these two values, the probability
+of accepting a new block goes down linearly to as the availability goes down. We choose to reject new blocks early before
+a directory is filled up, because the existing blocks in the directory can expand in size as we read new data in the block.
+Leaving a buffer in each directory can reduce the chance of eviction. **This is the default behavior.**
+
+- **AcceptingReviewer**: This reviewer accepts every block allocation. So the behavior will be exactly the same
```
jiacheliu3, line 251: @gpang Do we want to mention the tier allocation issue if they use this AcceptingReviewer?

docs/en/core-services/Caching.md
```diff
@@ -237,7 +283,7 @@ available options are:
 #### Evictor Emulation
 The old eviction policies are now removed and Alluxio provided implementations are replaced with appropriate annotation policies.
 Configuring old Alluxio evictors will cause worker startup failure with `java.lang.ClassNotFoundException`.
-Also, the old watermark based configuration is invalidated. So the following previous configuration options are ineffective:
+Also, **the old watermark based configuration is invalidated**. So the following previous configuration options are ineffective:
```
jiacheliu3, line 286: This kind of change is quite important actually. I don't think we have a notice to those who are relying on this watermark feature.

docs/en/core-services/Caching.md
```diff
@@ -333,32 +379,32 @@ $ ./bin/alluxio fs free ${PATH_TO_UNUSED_DATA}
 
 This will remove the data at the given path from Alluxio storage. The data is still accessible if
 it is persisted to a UFS. For more information refer to the
-[command line interface documentation]({{ '/en/operation/User-CLI.html' | relativize_url }}#free)
+[command line interface documentation]({{ '/en/operation/User-CLI.html#free' | relativize_url }})
```
jiacheliu3, line 382: I think the hashtag part should go together with the URL, as how we do it in most other docs. This change is tested.

jiacheliu3:
@LuQQiu FYI

docs/en/core-services/Caching.md
```diff
@@ -144,12 +144,18 @@ The user can also specify the tier that the data will be written to via
 
 If the data is already in Alluxio, the client will simply read the block from where it is already stored.
 If Alluxio is configured with multiple tiers, the block may not be necessarily read from the top tier,
-since it could have been moved to a lower tier transparently.
+since it may have been loaded into a lower tier, or moved to a lower tier transparently.
```
jiacheliu3, line 147: @gpang Do we want to mention how the blocks could have been asynchronously loaded? Very interestingly we only mention the async caching fact in https://docs.alluxio.io/os/user/stable/en/operation/Performance-Tuning.html#async-block-caching I don't think the doc path for that feature is intuitive, but I also don't think this section is a good place for that. What do you think?

docs/en/core-services/Caching.md
gpang: Can you explain why it was changed to CACHE? It was because the async alignment tasks will readjust blocks according to the annotation policy, so we don't require the synchronous move to top tier, since that can cause unnecessary delays.

docs/en/core-services/Caching.md
gpang: No, I don't think that is necessary.

gpang (commented):
@jiacheliu3 Thanks! I left a few minor comments.

docs/en/core-services/Caching.md
jiacheliu3: I just added more explanation. Does it make more sense now?

gpang (approved):
LGTM

gpang:

gpang:

PR #12577, opened 2020-11-25
Change ResultRequired to SummaryOnly

apc999:


yuzhu (approved):

gpang:
@apc999 Is this misleading? I think we just send back an empty list, not a summary of the listing. I would assume for a summary we would send back maybe the # of paths, etc, but the client will just get back an empty list.

yuzhu:
does `loadMetadataOnly` work better? Seems like that is the only purpose of this flag right?

apc999:

JySongWithZhangCe:
Thx!

PR #12593, opened 2020-11-30
Increase Active Jobs Size for DistributedJobCommands

bradyoo:
Increase default active jobs size because this increases performance for smaller file size distributedCp/Load cases without hurting the large files. (The only increased cost is increased memory usage but the memory usage here isn't large.)


gpang (approved):
LGTM

gpang:

PR #12618, opened 2020-12-09
Fix creating UFS file without a parent directory

bf8086:
We have been using default `CreateOptions` when creating a UFS file. This can fail in some configuration when a file is written to an Alluxio directory not backed by a UFS one. This change adds the option when UFS file writer creates the UFS file so the directory is created automatically.

https://github.com/Alluxio/alluxio/issues/12621


bradyoo:
I don't think I'm the right person to review this @gpang can you take a look?

bf8086:
@gpang @yuzhu PTAL.

yuzhu (commented):
Have a high level comment. Should we create the parent? `cp file /nonexistent/file` fails, why should `createfile /nonexistent/file` succeed?

bf8086:
> Have a high level comment. Should we create the parent? `cp file /nonexistent/file` fails, why should `createfile /nonexistent/file` succeed?

This option is only added for worker code path to persist a file to UFS. The actual parent path check is done on the master at the metadata level.

yuzhu (approved):
Got it.  LGTM.

gpang:
@bf8086 Can you add a github issue, and then link it in this PR description?

bf8086:

bf8086:

PR #12625, opened 2020-12-11
Decouple Allocator and Reviewer

jiacheliu3:
This change adds an `AllocationCoordinator` to decouple `Allocator` and `Reviewer`.

Goal:
1. An `Allocator` instance should not hold an instance of `Reviewer`.
1. Ideally the `Allocator` should be singleton, as `RoundRobinAllocator` is stateful. The state is the pointer to where we allocate the next block.


core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -127,9 +129,12 @@ public TieredBlockStore() {
...
     BlockMetadataEvictorView initManagerView = new BlockMetadataEvictorView(mMetaManager,
         Collections.<Long>emptySet(), Collections.<Long>emptySet());
-    mAllocator = Allocator.Factory.create(initManagerView);
-    if (mAllocator instanceof BlockStoreEventListener) {
-      registerBlockStoreEventListener((BlockStoreEventListener) mAllocator);
+
+
+//    mAllocator = Allocator.Factory.create(initManagerView);
+    mAllocationCoordinator = AllocationCoordinator.getInstance(initManagerView);
+    if (mAllocationCoordinator.getAllocator() instanceof BlockStoreEventListener) {
```
jiacheliu3, line 136: Currently none of the `Allocator` implementations implements `BlockStoreEventListener` but for the sake of backward compatibility I'm keeping this.

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
```diff
@@ -73,5 +76,5 @@ public static Allocator create(BlockMetadataView view) {
    *         otherwise
    */
   StorageDirView allocateBlockWithView(long sessionId, long blockSize, BlockStoreLocation location,
-      BlockMetadataView view, boolean skipReview);
+                                       BlockMetadataView view, Predicate<StorageDirView> reviewFunc);
```
jiacheliu3, line 79: `skipReview` is now replaced by a passed-in reviewer lambda.

core/server/worker/src/main/java/alluxio/worker/block/allocator/GreedyAllocator.java
```diff
@@ -33,7 +35,6 @@
   private static final Logger LOG = LoggerFactory.getLogger(GreedyAllocator.class);
 
   private BlockMetadataView mMetadataView;
-  private Reviewer mReviewer;
```
jiacheliu3, line 36: The `Reviewer` instance is decoupled and removed from the `Allocator`

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -127,9 +128,10 @@ public TieredBlockStore() {
 
     BlockMetadataEvictorView initManagerView = new BlockMetadataEvictorView(mMetaManager,
         Collections.<Long>emptySet(), Collections.<Long>emptySet());
-    mAllocator = Allocator.Factory.create(initManagerView);
-    if (mAllocator instanceof BlockStoreEventListener) {
-      registerBlockStoreEventListener((BlockStoreEventListener) mAllocator);
+
+    mAllocationCoordinator = AllocationCoordinator.getInstance(initManagerView);
```
jiacheliu3, line 132: `AllocationCoordinator.getInstance()`, formerly `Allocator.Factory.create(initManagerView)` is called twice in non-test code. This can cause issues where the Allocator is stateful like `RoundRobinAllocator`. Honestly I don't think anyone is using or should use `RoundRobinAllocator`, but anyways.

The 2 references are here and https://github.com/Alluxio/alluxio/blob/355bca2291caef1370858684316b6a188d4f105e/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java#L119, see my change in `BlockMetadataManager`.

I think using a singleton can avoid having multiple stateful Allocator instances maintaining separate states. This is currently not a very critical issue yet, because currently we are creating 2 Allocator instances, both from empty block view in the same `TieredBlockStore` constructor. But I think this change may help in the future. What do you think @gpang ?

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
```diff
@@ -68,10 +69,10 @@ public static Allocator create(BlockMetadataView view) {
    * @param blockSize the size of block in bytes
    * @param location the location in block store
    * @param view of the block metadata
-   * @param skipReview whether the review should be skipped
+   * @param reviewFunc the {@link Reviewer} logic that rejects bad block allocations
```
jiacheliu3, line 72: By using a `reviewFunc` lambda, `skipReview` can be replaced. Whether to skip can be determined by `AllocationCoordinator` on one level above.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+      }
+    }
+    return sInstance;
+  }
+
+  /**
+   * This is currently only used by tests, where the singleton state needs to be reconfigured.
+   * */
+  public static synchronized void destroyInstance() {
```
jiacheliu3, line 33: This is only used by tests where we want to update the reviewer logic or refresh the stateful Allocator.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
```diff
@@ -0,0 +1,52 @@
...
+    sInstance = null;
+  }
+
+  public StorageDirView allocateBlockWithView(long sessionId, long blockSize, BlockStoreLocation location,
+                                       BlockMetadataView view, boolean skipReview) {
+    if (skipReview) {
+      return mAllocator.allocateBlockWithView(sessionId, blockSize, location, view, v -> true);
+    }
+    return mAllocator.allocateBlockWithView(sessionId, blockSize, location, view, mReviewer::acceptAllocation);
```
jiacheliu3, line 38: Although we can do this, I still DO NOT find a way to completely remove `AcceptingReviewer`. The reason is some integration tests:
```
// This is the levels
LocalAlluxioCluster
  WorkerProcess
    TieredBlockStore
      TieredBlockStore.allocateSpace
        TieredBlockStore.mAllocationCoordinator.allocateBlockWithView with true or false, depending on the logic in allocateSpace
```

In the integration tests we are creating real worker processes, there's no way to force the allocator to skip all reviews, except having an `AcceptingReviewer`. With that said, `AcceptingReviewer` class is ONLY NEEDED by the integration tests.

I also tried to move `AcceptingReviewer` into the `alluxio-test` module, but that DOES NOT work, because the worker module does NOT put the `alluxio-test` module on the classpath so can NOT find the `AcceptingReviewer` if that is in the test module. 

Continue on that thread, we are not really enabling people to extend our Allocator interface. Imagine I'm a user, the ONLY way to extend Allocator is to put my implementation in the `alluxio-worker` module (or the ones it depends on, essentially alluxio core modules), and recompile Alluxio. It's not really possible to plug my class in as a jar or something, despite the fact that we have `PropertyKey,ALLOCATOR_CLASS` which looks very extendable. @gpang I think it makes sense to change to service loader, does that make sense to you?

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
jiacheliu3: Actually I removed the `synchronized` cuz I don't think it's necessary at the moment.

jiacheliu3: The `TieredBlockStore` is not threadsafe itself.

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
jiacheliu3: Although the Allocator is decoupled from specific Reviewer implementation (not holding the instance anymore), admittedly this is not 100% decoupled. This is due to 2 reasons:

1. Java DOES NOT have "generator" like python `yield` grammar. I have to generate the full list of possible allocation proposals and pass it back to the `AllocationCoordinator` to review. There's no "Allocator saying I have this proposal, Reviewer rejects so Allocator you think of one more" interactive allocation decision.

1. Even if we implement passing the list of proposals to the `AllocationCoordinator` and select one, it's hard to update the stateful `Allocator` after `AllocationCoordinator` makes a decision. `RoundRobinAllocator` is stateful and needs to know the decision in order to make the next one. `RoundRobinAllocator` has been the source of evil in a few different locations :(

jiacheliu3: @gpang With above being said, I currently don't have a better idea than the implementation im putting here. See if this is acceptable to you.

jiacheliu3: Well, if we organize the StorageDir in a LinkedList we can probably have a "reject and gimme next proposal" implementation in the `AllocationCoordinator`.

jiacheliu3: If this implementation is not clean (not decoupling enough) for you, there are other options:

1. There are not many StorageDir. It hurts very little to just order them and pass the list back from the `Allocator`. It's just a bit hard to update the Allocator with the allocation decision, but that is also feasible.
1. If we order `StorageDir` in the Allocator as a LinkedList, we can do a "reject and Allocator gimme your next proposal" within the `AllocationCoordinator`.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
```diff
@@ -0,0 +1,51 @@
...
+  public Allocator mAllocator;
+
+  private static AllocationCoordinator sInstance = null;
+
+  private AllocationCoordinator(BlockMetadataView view) {
+    mAllocator = Allocator.Factory.create(view);
+    mReviewer = Reviewer.Factory.create();
+  }
+
+  public static AllocationCoordinator getInstance(BlockMetadataView view) {
```
gpang, line 22: Why do we need a singleton for the allocation coordinator?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
gpang: It is not clear to me why the allocation coordinator needs to be a singleton? If the only times we create it is in the constructor of `TieredBlockStore` then it seems trivial to make it not a singleton and just use a single instance?

I also don't understand how a stateful allocator affects the singleton/instance decision, since the allocator state is living in the allocatorCoordinator, which is either a single instance or a singleton.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
gpang: I see. I don't expect the reviewer to be extended anytime soon, so we can just keep it like this for now.

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * This is currently only used by tests, where the singleton state needs to be reconfigured.
+   * */
+  public static void destroyInstance() {
+    sInstance = null;
+  }
+
+  public StorageDirView allocateBlockWithView(long sessionId, long blockSize, BlockStoreLocation location,
+                                       BlockMetadataView view, boolean skipReview) {
```
gpang, line 37: Is this `view` different from the view passed in for `getInstance()`?

core/server/worker/src/main/java/alluxio/worker/block/reviewer/AllocationCoordinator.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public static AllocationCoordinator getInstance(BlockMetadataView view) {
+    if (sInstance == null) {
+      sInstance = new AllocationCoordinator(view);
+    }
+    return sInstance;
+  }
+
+  /**
+   * This is currently only used by tests, where the singleton state needs to be reconfigured.
```
gpang, line 30: I feel like making this not a singleton would remove the need for this method too?

core/server/worker/src/main/java/alluxio/worker/block/allocator/Allocator.java
gpang: I see, so the main issue is that the allocator needs to update internal state, depending on the decision by the reviewer.

However, would changing the api to basically return an iterator be too big of a change? basically it could look something like:

```
it = allocator.getNewIterator();

while (it.hasNext()) {
  if (reviewer.accept(it.next()) {
    // This storage dir is accepted!
  }
}
// NO storage dir possible
```

In this model, all the internal state is managed within the iterator implementation, for each allocator.

Could something like this work?

gpang (commented):
@jiacheliu3 Thanks for this! I basically have one high-level comment which is would we be able to convert our allocators to be an iterator-style api, where the state of iteration is managed within the iterator, and the coordinator can just iterate over them and accept/reject them?

I'm just not sure how much effort it will be do to that, and if all allocators can fit within the iterator model.

jiacheliu3:
> @jiacheliu3 Thanks for this! I basically have one high-level comment which is would we be able to convert our allocators to be an iterator-style api, where the state of iteration is managed within the iterator, and the coordinator can just iterate over them and accept/reject them?
> 
> I'm just not sure how much effort it will be do to that, and if all allocators can fit within the iterator model.

@gpang  I tried implementing an iterator flavor, but found that's not trivial.

The biggest challenge roots in the fact that we have both stateful and stateless Allocators:
1. [Stateless] GreedyAllocator: This implementation is meant to be an example, not for any "real" use.
2. [Stateful] RoundRobinAllocator: Although this is not meant to be an example, I don't think many people are using it.
3. [Stateless] MaxFreeAllocator: This is the default. Since most people are not specifying the allocator, I believe most people are on this now.

The Iterator flavor should fit in easily with the stateful one, but maybe hard for the stateless ones. Let's take your example:
```
it = allocator.getNewIterator();

while (it.hasNext()) {
  if (reviewer.accept(it.next()) {
    // This storage dir is accepted!
  }
}
// NO storage dir possible
```

For the stateful Allocator, it maintains the "current" in its own state. The allocator itself can serve as an iterator. But for the stateless ones, we will have to pass in the current state. Let's only discuss how to make a "stateless" Allocator an iterator. Because however we achieve that, it'll be only easier to do the same for the stateful one. There are multiple ways to make an iterator out of a stateless Allocator:
**1. [Naive] Every time we allocate space, order the `StorageDir` that meet the criteria (space, medium, tier), and return the iterator of the list...** 
```
List<StorageDirView> list = new ArrayList<>();
// loop all StorageDir and put the good ones into the list
return list.iterator();
```

**2. Imagine the `StorageDirView` itself is chained.**
```
class StorageDirView implements Iterator<StorageDirView> {
  // omitted..
  StorageDirView mNext; // This is deterministic. We know how many StorageDir there are on startup.
  boolean hasNext() {return mNext == null;}

  // On generation, we chain the StorageDirView by tier and index
}

class GreedyAllocator {
  StorageDirView mDummyHead;
  mDummyHead.next = the first StorageDirView in Alluxio
  Iterator<StorageDirView> getNewIterator(...) {
    return mDummyHead;
  }
}
```
The challenge is, we can have an iterator for the `StorageDirView` sure, but now it's hard to do filter based on the allocation criteria (space, tier, medium). We need to be able to start the chain from a tier/medium and end it too.

**3. This PR**
We face the challenge from stateful vs stateless Allocator only when we want to iterate outside. If we pass in the reviewer as a Predicate into the Allocator, we get around that particular challenge... And that is my current solution in the PR.

So conclusion on the 3 options I tried/thought over:
1. Waste CPU/Heap but the logic is simpler.
2. Much redesign. Also it's controversial(in which interface) where to put the filter on the StorageDirView chain.
3. Not elegant enough, not breaking the cohesion.

jiacheliu3:
> > @jiacheliu3 Thanks for this! I basically have one high-level comment which is would we be able to convert our allocators to be an iterator-style api, where the state of iteration is managed within the iterator, and the coordinator can just iterate over them and accept/reject them?
> > I'm just not sure how much effort it will be do to that, and if all allocators can fit within the iterator model.
> 
> @gpang I tried implementing an iterator flavor, but found that's not trivial.
> 
> The biggest challenge roots in the fact that we have both stateful and stateless Allocators:
> 
> 1. [Stateless] GreedyAllocator: This implementation is meant to be an example, not for any "real" use.
> 2. [Stateful] RoundRobinAllocator: Although this is not meant to be an example, I don't think many people are using it.
> 3. [Stateless] MaxFreeAllocator: This is the default. Since most people are not specifying the allocator, I believe most people are on this now.
> 
> The Iterator flavor should fit in easily with the stateful one, but maybe hard for the stateless ones. Let's take your example:
> 
> ```
> it = allocator.getNewIterator();
> 
> while (it.hasNext()) {
>   if (reviewer.accept(it.next()) {
>     // This storage dir is accepted!
>   }
> }
> // NO storage dir possible
> ```
> 
> For the stateful Allocator, it maintains the "current" in its own state. The allocator itself can serve as an iterator. But for the stateless ones, we will have to pass in the current state. Let's only discuss how to make a "stateless" Allocator an iterator. Because however we achieve that, it'll be only easier to do the same for the stateful one. There are multiple ways to make an iterator out of a stateless Allocator:
> **1. [Naive] Every time we allocate space, order the `StorageDir` that meet the criteria (space, medium, tier), and return the iterator of the list...**
> 
> ```
> List<StorageDirView> list = new ArrayList<>();
> // loop all StorageDir and put the good ones into the list
> return list.iterator();
> ```
> 
> **2. Imagine the `StorageDirView` itself is chained.**
> 
> ```
> class StorageDirView implements Iterator<StorageDirView> {
>   // omitted..
>   StorageDirView mNext; // This is deterministic. We know how many StorageDir there are on startup.
>   boolean hasNext() {return mNext == null;}
> 
>   // On generation, we chain the StorageDirView by tier and index
> }
> 
> class GreedyAllocator {
>   StorageDirView mDummyHead;
>   mDummyHead.next = the first StorageDirView in Alluxio
>   Iterator<StorageDirView> getNewIterator(...) {
>     return mDummyHead;
>   }
> }
> ```
> 
> The challenge is, we can have an iterator for the `StorageDirView` sure, but now it's hard to do filter based on the allocation criteria (space, tier, medium). We need to be able to start the chain from a tier/medium and end it too.
> 
> **3. This PR**
> We face the challenge from stateful vs stateless Allocator only when we want to iterate outside. If we pass in the reviewer as a Predicate into the Allocator, we get around that particular challenge... And that is my current solution in the PR.
> 
> So conclusion on the 3 options I tried/thought over:
> 
> 1. Waste CPU/Heap but the logic is simpler.
> 2. Much redesign. Also it's controversial(in which interface) where to put the filter on the StorageDirView chain.
> 3. Not elegant enough, not breaking the cohesion.

@calvinjia do you mind taking a look at this discussion? Thanks!

calvinjia (commented):
@jiacheliu3 Could you summarize the outstanding comments to be addressed (if there are any)?

Also, just so I am up to speed on the motivation for this PR -
We want to decouple the Allocator and Reviewer, therefore we created AllocationCoordinator which acts as a bundle of Allocator and Reviewer.

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
```diff
@@ -58,12 +56,11 @@ public StorageDirView allocateBlockWithView(long sessionId, long blockSize,
    * @param sessionId the id of session to apply for the block allocation
    * @param blockSize the size of block in bytes
    * @param location the location in block store
-   * @param skipReview whether the review should be skipped
    * @return a {@link StorageDirView} in which to create the temp block meta if success,
    *         null otherwise
    */
   private StorageDirView allocateBlock(long sessionId, long blockSize,
```
ggezer, line 62: Does reviewer heuristic help MaxFreeAllocator?

ggezer (commented):
It looks quite challenging to decouple reviewer logic from allocator. Also, even though I agree to this heuristic, I'm not sure if that's applicable for all allocation schemes.

How about if we provide these soft/hard limits from `StorageDir` level and then make use of them in appropriate allocators? We already have a similar logic in `StorageDirView` for representing internal reserved space. (More like a hack currently ).

jiacheliu3:
Sorry for the late reply. Yes as @calvinjia 's understanding, the intention is to decouple the `Allocator` interface with `Reviewer`. Now the `Reviewer` instance is in the `Allocator` instance, eg. https://github.com/Alluxio/alluxio/blob/5970957e176fccc93e3fd5cffe15ee59b8ba178d/core/server/worker/src/main/java/alluxio/worker/block/allocator/RoundRobinAllocator.java#L57 The intention of this PR is to remove this coupling by having a class that overlooks the allocate->review pipeline. 

The main challenge of that is the fact that we have both stateless(`MaxFreeAllocator` which simply gives you the most free dir) and stateful(`RoundRobinAllocator`). It's hard to use the same design and at the same time change as little of the interface as possible. For the stateful one the main challenge is how the state should be maintained when the decision can be rejected and then you retry. For the stateless one, it's hard to reject and ask for a different decision.

That said, taking a step back I think it's non-trivial to decouple these two interfaces. This refactor  doesn't really do the decouple. Therefore I'll close this PR.

However after a discussion with @ggezer , here are the thoughts:
1. From the design perspective, `Allocator` and `Reviewer` should still be decoupled.
2. We will need to examine over each Allocator+Reviewer combination to see if that makes sense.
3. Based on (2) we can come up with the overseer design(currently `AllocationCoordinator` in this PR but that can be a better name). That design should make sense for the combinations we examine and therefore the implementation should be able to last longer.

jiacheliu3:
Closing this PR. I think it makes more sense to convert the discussion into this https://github.com/Alluxio/alluxio/issues/13289 to track. Please move the discussion there.

PR #12640, opened 2020-12-15
Reduce cycle times

ljk-1993:
When tier is null, there is no need to do the following loop calculation.


apc999:
@ljk-1993  thanks for contributing! can you check if you are checking to the correct branch? Your change currently has ~2000 lines of change which is huge

PR #12637, opened 2020-12-14
Revert "Avoid loads ufs jars multiple times during journal formatting"

apc999:
This reverts commit 22b67274fd032ce46532c3e21a46e973c7ccc307.

Release tests are failing on format

@maobaolong

PR #12366, opened 2020-10-22
Avoid loads ufs jars multiple times during journal formatting

maobaolong:
Fix #12113 

I do the following test, and can only search one line loading core jars output, it is my expected.

```bash
alluxio formatMaster
```

I also start the master to check master can start successful, check alluxio master log only contains one line loading core jars output, it is my expected.

Issue #12113, opened 2020-09-18
Alluxio loads ufs jars multiple times during journal formatting


ChernikovP:
**Alluxio Version:**
I'm working of `branch-2.3`

**Describe the bug**
Formatting journal results in `Native Library /tmp/mapr-root-libMapRClient.unknown.so already loaded in another classloader` error. For full log, please, see https://gist.github.com/ChernikovP/687d9bdcbec2b0f06fe318489ddb45ce.

What's happening is that for every single service: `BlockMaster`, `FileSystemMaster`, `MetaMaster`, `MetricsMaster` and `TableMaster` execution of https://github.com/Alluxio/alluxio/blob/680b2fd4ff13d3349572518f15e1f9f5b2fffa93/core/server/common/src/main/java/alluxio/cli/Format.java#L120 results in loading all possible ufs factories, filtering by path prefix the necessary one and trying to process journal entry

**To Reproduce**
- Add `mapr` repository to `pom.xml`:

```
<repository>
  <id>mapr-releases</id>
  <url>https://repository.mapr.com/maven/</url>
  <snapshots>
    <enabled>false</enabled>
  </snapshots>
  <releases>
    <enabled>true</enabled>
  </releases>
</repository>
```

- add follwoing module to `ufsModules` in `dev/scripts/src/alluxio.org/build-distribution/cmd/common.go`:

```
{"hadoop-mapr", true, "-pl underfs/hdfs -Pufs-hadoop-2 -Dufs.hadoop.version=2.7.0-mapr-1808 -PhdfsActiveSync"}
```

- build binaries by running `./dev/scripts/generate-tarballs single -ufs-modules=ufs-hadoop-mapr`

- conf file:

```
# Underfs setup
alluxio.underfs.hdfs.configuration=/opt/mapr/hadoop/hadoop-2.7.0/etc/hadoop/core-site.xml:/opt/mapr/hadoop/hadoop-2.7.0/etc/hadoop/hdfs-site.xml
alluxio.underfs.hdfs.prefixes=maprfs:/
alluxio.underfs.hdfs.impl=com.mapr.fs.MapRFileSystem
# Master setup
alluxio.master.backup.directory=maprfs:///storage/scratch/alluxio/backup
alluxio.master.journal.folder=maprfs:///storage/scratch/alluxio/journal
alluxio.master.journal.type=UFS
alluxio.master.mount.table.root.ufs=maprfs:///
```

- run `./bin/alluxio format` on Alluxio master

[...]

PR #12642, opened 2020-12-15
Avoid loads ufs jars multiple times during journal formatting

maobaolong:
This PR contains the changes in #12366 to solve #12113
and fix the alluxio format errors when running with HDFS journal mentioned in #12637

### DIFF
--- a/core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
+++ b/core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
@@ -28,6 +28,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.net.URI;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -101,7 +103,7 @@ public abstract class AbstractUfsManager implements UfsManager {
       new ConcurrentHashMap<>();
 
   private UfsClient mRootUfsClient;
-  private UfsClient mJournalUfsClient;
+  private Map<String, UfsClient> mJournalUfsClientMap;
   protected final Closer mCloser;
 
   protected AbstractUfsManager() {
@@ -213,18 +215,23 @@ public abstract class AbstractUfsManager implements UfsManager {
   }
 
   @Override
-  public UfsClient getJournal() {
+  public UfsClient getJournal(URI location) {
     synchronized (this) {
-      if (mJournalUfsClient == null) {
-        addMount(IdUtils.UFS_JOURNAL_MOUNT_ID, new AlluxioURI("/dummy_journal_mount_point"),
+      if (mJournalUfsClientMap == null) {
+        mJournalUfsClientMap = new HashMap<>();
+      }
+      UfsClient ufsClient = mJournalUfsClientMap.get(location.getScheme());
+      if (ufsClient == null) {
+        addMount(IdUtils.UFS_JOURNAL_MOUNT_ID, new AlluxioURI(location.toString()),
             UfsJournal.getJournalUfsConf());
         try {
-          mJournalUfsClient = get(IdUtils.UFS_JOURNAL_MOUNT_ID);
+          ufsClient = get(IdUtils.UFS_JOURNAL_MOUNT_ID);
+          mJournalUfsClientMap.put(location.getScheme(), ufsClient);
         } catch (NotFoundException | UnavailableException e) {
           throw new RuntimeException("We should never reach here", e);
         }
       }
-      return mJournalUfsClient;
+      return ufsClient;
     }
   }
 
--- a/core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
+++ b/core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
@@ -29,7 +29,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URI;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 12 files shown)


core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
```diff
@@ -209,6 +214,27 @@ public UfsClient getRoot() {
     }
   }
 
+  @Override
+  public UfsClient getJournal(URI location) {
+    synchronized (this) {
+      if (mJournalUfsClientMap == null) {
+        mJournalUfsClientMap = new HashMap<>();
+      }
+      UfsClient ufsClient = mJournalUfsClientMap.get(location.getScheme());
```
gpang, line 223: I am confused. Why could there be multiple journal locations?

Shouldn't this look similar to the root ufs client, which there is only 1 of?

gpang (commented):
@maobaolong I had a question.

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
maobaolong: From the master.log, It seems that there are multiple journal for different service need format.

```
2020-12-16 10:48:04,744 INFO  cli.Format (Format.java:format(121)) - Formatting master journal: /Users/mbl/projects/github/alluxio-2/journal/
2020-12-16 10:48:04,825 INFO  extensions.ExtensionFactoryRegistry (ExtensionFactoryRegistry.java:scanLibs(169)) - Loading core jars from /Users/mbl/projects/github/alluxio-2/lib
2020-12-16 10:48:05,000 INFO  extensions.ExtensionFactoryRegistry (ExtensionFactoryRegistry.java:scanExtensions(159)) - Loading extension jars from /Users/mbl/projects/github/alluxio-2/extensions
2020-12-16 10:48:05,032 INFO  ufs.UfsJournal (UfsJournal.java:format(428)) - Formatting /Users/mbl/projects/github/alluxio-2/journal/BlockMaster/v1
2020-12-16 10:48:05,058 INFO  ufs.UfsJournal (UfsJournal.java:format(428)) - Formatting /Users/mbl/projects/github/alluxio-2/journal/TableMaster/v1
2020-12-16 10:48:05,060 INFO  ufs.UfsJournal (UfsJournal.java:format(428)) - Formatting /Users/mbl/projects/github/alluxio-2/journal/FileSystemMaster/v1
2020-12-16 10:48:05,060 INFO  ufs.UfsJournal (UfsJournal.java:format(428)) - Formatting /Users/mbl/projects/github/alluxio-2/journal/MetaMaster/v1
2020-12-16 10:48:05,061 INFO  ufs.UfsJournal (UfsJournal.java:format(428)) - Formatting /Users/mbl/projects/github/alluxio-2/journal/MetricsMaster/v1
2020-12-16 10:48:05,062 INFO  cli.Format (Format.java:main(106)) - Formatting complete
```

Also, from the following Format#format, there are lots of master services, each one need create journal.

```java

public static void format(Mode mode, AlluxioConfiguration alluxioConf) throws IOException {
    NoopUfsManager noopUfsManager = new NoopUfsManager();
    switch (mode) {
      case MASTER:
        URI journalLocation = JournalUtils.getJournalLocation();
        LOG.info("Formatting master journal: {}", journalLocation);
        JournalSystem journalSystem = new JournalSystem.Builder()
            .setLocation(journalLocation).build(CommonUtils.ProcessType.MASTER);
        for (String masterServiceName : ServiceUtils.getMasterServiceNames()) {
          journalSystem.createJournal(new NoopMaster(masterServiceName, noopUfsManager));
        }
        journalSystem.format();
        break;
...
```

maobaolong:
@gpang Thanks for the reply, this PR aimed to fix the issue #12637 which is blocked by my previous PR https://github.com/Alluxio/alluxio/pull/12366 , and it cannot worked if the location is not local filesystem.

LuQQiu:
@maobaolong I didn't get time to take a deeper look into the related PRs, but I ran the related HDFS journal tests with this PR and those tests passed!

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
gpang: @maobaolong I see. Then, does this PR actually help with the original motivation of avoiding loading the UFS journal jars? It seems like this will create a new UFS client for each master service (block, fs, table, etc) to format each, and each one will have a different location, so it will load the jars again?

In this example, I feel like the ideal behavior should be to load the UFS jars once for the location `/Users/mbl/projects/github/alluxio-2/journal/`, and all the master services are going to be sub-directories of that single UFS journal location. Does this current PR already behave in this way?

maobaolong: @gpang Thanks for your review again. I think it just load the jars once during the format, since `location.getScheme()`, I use the scheme of the location as the key mapping to the related Underfilesystem.

And I did a format test using this PR, the behavior running as expected.

maobaolong:
@LuQQiu Thank you very much about passed tests, I'm excited to hear that. Thank you again.

maobaolong:
Hi @gpang I reply your comment inline, PTAL. Thanks.

core/server/common/src/main/java/alluxio/underfs/AbstractUfsManager.java
gpang: @maobaolong O, you are right. the scheme will always be the same. Then I am confused as to why we even need the map to begin with. The scheme for all the journals should always be the same, since they are always a sub-directory of the journal location right?

I feel like the map will always have only 1 entry, so we should just not use a map?

maobaolong: @gpang OK, I remove the map now.

maobaolong:
@gpang Thanks for the review these days, I remove the ufsclient map within the last commit, PTAL.

gpang (approved):
LGTM

maobaolong:
@gpang Thanks for your review.

gpang:

PR #12644, opened 2020-12-15
Reduce cycle times

ljk-1993:
when tier is null, there is no need to go into the following loop.


apc999 (approved):
thanks @ljk-1993 !

apc999:

PR #12646, opened 2020-12-15
Collect the client-side metrics and heartbeat them to master by default

ljk-1993:
If the user is not set the parameter "alluxio. User. The metrics. Collections. Enabled = true", alluxio monitoring page shows short - circuit 0 b, can lead to misunderstanding: alluxio not from the local worker load data.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3770,7 +3770,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey USER_METRICS_COLLECTION_ENABLED =
       new Builder(Name.USER_METRICS_COLLECTION_ENABLED)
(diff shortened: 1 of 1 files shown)


apc999:
@LuQQiu  @calvinjia  any downsides to turn on client log collection by default IYO?

ljk-1993:
@apc999 Operations like Alluxio shell and copyFromLocal will also be monitored

LuQQiu:
@apc999 When there are not many users, client log collection by default should not have problems, but in the real case, there could be a lot of clients, All the clients need to heartbeat to the leading master periodically which may add much overhead to the Alluxio leading master. I am not suggesting enable it by default.

apc999:
@ljk-1993  thanks for the patch. For now we want to avoid creating a lot of performance overhead when a lot of clients connecting to the master as mentioned by Lu. For this single flag, you can definitely turn it on in your site properties. Let me know if you have any questions. I will close this PR but feel free to re-open it if you think differently

PR #12649, opened 2020-12-15
Add clearer ratis flush exception logging

LuQQiu:
Add the debug logging since the actual flush failure will be overwritten by the last meaningless failure like `ratis.protocol.AlreadyClosedException` or `NoMoreRetriesException`.

The actual flush failure will sometimes give us helpful information or annoying meaningless logs, that's why i mark it as debug for now. Will be better to have an exception duplication-checking and only log the no-duplicate exceptions. And also link the exception to meaningful configuration suggestions.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
@@ -30,6 +30,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.SettableFuture;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.grpc.Status;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Iterator;
@@ -49,6 +51,8 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED")
 public final class AsyncJournalWriter {
+  private static final Logger LOG = LoggerFactory.getLogger(AsyncJournalWriter.class);
+
   /**
    * Used to manage and keep track of pending callers of ::flush.
    */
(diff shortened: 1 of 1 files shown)


gpang:
@LuQQiu Is there value to make this a WARN or ERROR? Why should this only be in DEBUG?

core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
```diff
@@ -308,6 +312,10 @@ private void doFlush() {
           }
         }
       } catch (IOException | JournalClosedException exc) {
+        // Add the error logging here since the actual flush error may be overwritten
+        // by the future meaningless ratis.protocol.AlreadyClosedException
+        // use debug level logging because the exception may be annoying in some cases
+        LOG.debug("Failed to flush journal entry", exc);
```
bf8086, line 318: (nit) consider using `SamplingLogger` if the error happens too frequently.

LuQQiu: Good advice! I don't know we have that convenient logging way!

LuQQiu:
@gpang I changed it to be WARN. The logging is helpful in understanding the ratis errors. The previous debug is to avoid unnecessary annoying logs

core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
```diff
@@ -308,6 +316,9 @@ private void doFlush() {
           }
         }
       } catch (IOException | JournalClosedException exc) {
+        // Add the error logging here since the actual flush error may be overwritten
+        // by the future meaningless ratis.protocol.AlreadyClosedException
+        SAMPLING_LOG.warn("Failed to flush journal entry", exc);
```
gpang, line 321: NIT: the sampling logger works by treating every log template as a separate error. I don't know if this will be useful, but if we use the template `SAMPLING_LOG.warn("Failed to flush journal entry: " + exc.getMessage(), exc); `, we would get to see the warning log of each cause, once every 30s. This may help if there are potentially different causes of journal flush errors, and we would still not miss any cause.

gpang (commented):
@LuQQiu Thanks! I just had a minor comment.

core/server/common/src/main/java/alluxio/master/journal/AsyncJournalWriter.java
LuQQiu: Good catch! in the previous way, the different error messages may still be ignored.

gpang (approved):
LGTM

gpang:

PR #12650, opened 2020-12-15
Add ratis size related configuration

LuQQiu:
Add ratis configuration so that users can change it when they have larger journal entries (e.g. introduced by Alluxio catalog service).

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1549,6 +1549,22 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX)
+          .setDefaultValue("64MB")
+          .setDescription("The maximum size in bytes of journal entries allowed "
+              + "in concurrent journal flushing (journal IO to standby masters "
+              + "and IO to local disks.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX)
+          .setDefaultValue("4MB")
+          .setDescription("The maximum single journal entry size allowed to be flushed.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_PORT =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_PORT)
           .setDescription("The port to use for embedded journal communication with other masters.")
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1549,6 +1549,22 @@ public String toString() {
...
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX)
+          .setDefaultValue("64MB")
+          .setDescription("The maximum size in bytes of journal entries allowed "
+              + "in concurrent journal flushing (journal IO to standby masters "
+              + "and IO to local disks.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX =
```
bf8086, line 1561: Do we need two different properties? They looks quite similar in name and description.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -5139,6 +5155,10 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.master.embedded.journal.appender.batch.size";
     public static final String MASTER_EMBEDDED_JOURNAL_CATCHUP_RETRY_WAIT =
         "alluxio.master.embedded.journal.catchup.retry.wait";
+    public static final String MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX =
+        "alluxio.master.embedded.journal.flush.size.max";
+    public static final String MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX =
+        "alluxio.master.embedded.journal.catchup.retry.wait";
```
gpang, line 5161: This name is a duplicate of something else?

gpang (changes requested):

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: The two are different, the journal size max is single journal entry max. The journal flush size max is the total journal entries size. Any advice for the naming?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -5139,6 +5155,10 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.master.embedded.journal.appender.batch.size";
     public static final String MASTER_EMBEDDED_JOURNAL_CATCHUP_RETRY_WAIT =
         "alluxio.master.embedded.journal.catchup.retry.wait";
+    public static final String MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX =
+        "alluxio.master.embedded.journal.flush.size.max";
+    public static final String MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX =
```
gpang, line 5160: Would `MASTER_EMBEDDED_JOURNAL_ENTRY_SIZE_MAX` make more sense?

and `alluxio.master.embedded.journal.entry.size.max`

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1549,6 +1549,22 @@ public String toString() {
...
+          .setDefaultValue("64MB")
+          .setDescription("The maximum size in bytes of journal entries allowed "
+              + "in concurrent journal flushing (journal IO to standby masters "
+              + "and IO to local disks.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_JOURNAL_SIZE_MAX)
+          .setDefaultValue("4MB")
```
gpang, line 1563: Should/can this be larger? Is there any downside for a larger max size?

gpang (commented):
@LuQQiu Thanks! I left a few questions.

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: That's the default value in Ratis. If we want to increase this value, we may hit other thresholds.
For example, FLUSH_SIZE_MAX property should also be increased. ENTRY_SIZE_MAX defines the single journal entry allowed, FLUSH_SIZE_MAX one define the total concurrent journal entry size allowed. Other thresholds are unknown. But i agree that we probably should enlarge the default value of these two properties together

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Yeah, the new name is bettter. Updated, thanks!

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Enlarge the value of both properties, it should not have downside in normal cases. If we figure out any downside, i would move the value back

gpang (approved):
LGTM

gpang:

LuQQiu:

LuQQiu:

PR #12652, opened 2020-12-16
Support Authorization and add ListAllMyBuckets for S3 Rest Api

bradyoo:


bradyoo:
In action:

AccessKeyId is your user name. Secret Access Key is filled but unused.
 
image

image

bradyoo:
User: bradleyyoo

image

image

(Note that listing a bucket itself is currently buggy/incorrect as you can see from the browser view. At the very least 0/0 is not a file.)

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
```diff
@@ -99,6 +104,56 @@ public S3RestServiceHandler(@Context ServletContext context) {
         (FileSystem) context.getAttribute(ProxyWebServer.FILE_SYSTEM_SERVLET_RESOURCE_KEY);
   }
 
+  private static String getUserFromAuthorization(String authorization) {
```
gpang, line 107: Can we make this not `private`, and then add a unittest for this method?

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
```diff
@@ -99,6 +104,56 @@ public S3RestServiceHandler(@Context ServletContext context) {
...
+  public Response listAllMyBuckets(@HeaderParam("Authorization") String authorization) {
+
+    return S3RestUtils.call("", new S3RestUtils.RestCallable<ListAllMyBucketsResult>() {
+      @Override
+      public ListAllMyBucketsResult call() throws S3Exception {
+        String user = getUserFromAuthorization(authorization);
+
+        List<URIStatus> objects;
+        try {
+          objects = mFileSystem.listStatus(new AlluxioURI("/"));
```
gpang, line 143: What exactly is the concept of a bucket in the Alluxio proxy?

gpang (changes requested):
@bradyoo Thanks! I left a few comments.

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
bradyoo: There is a general lack of unit tests for most methods here.. That is partially because I don't have a good understanding of the full API spec quite yet and I plan to add them later as I implement more and have a better understanding of the spec.

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
bradyoo: Bucket in the concept of alluxio is the most outer folder. This means that if there are files in the outer folder. There will be no way to view it using the S3 API.

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
gpang: Ok, as long as we don't forget to add the coverage soon.

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
gpang: We may need to update the behavior in the future.

Another possibility is that we always have a "root" bucket which just holds the files under root, and the directories under root are just separate buckets.

In any case, I don't think there is a clear spec for this currently, so we may have to be flexible in the future.

gpang (approved):
LGTM

gpang:

bradyoo:
@gpang That would actually be a pretty large spec change from what it is from the past and I'm not sure if I'm comfortable making that spec change.

PR #12653, opened 2020-12-16
Add 'attachdb' option for mounting diff prefix partitions

### LINKED ISSUE #12651: Allow mounting Hive partition with different location prefix than that of table
Yes. I'm running Presto queries on Alluxio with the Alluxio Catalog Service setup to keep in sync with metadata stored in Hive. The `attachdb` and `sync` command will fail adding the Hive tables that contain partitions that don't share the same location prefix with that of table. For example, if a Hive table is located on `s3://bucket/root_path`, while one of its partitions is located on `s3://bucket/path1/some_suffix`, the whole table sync will fail.

Add an option for `attachdb` command to allow user determine whether to allow mounting partitions that don't share the same location prefix with that of table, so the table/db sync can be finished.

No.

**Urgency**
We have some important tables that contain partitions with different prefix, and we have to attach those tables if we want to run Presto with Alluxio

**Additional context**
Pasting the result error log that is caused by the above problem:
```
Table TABLE_NAME failed to sync: java.io.IOException: Failed to translate ufs path (UFS_PATH). Mapping missing from translator
```

hycsam:
This PR aims to add an Hive UDB option in `table attachdb` command, to allow mounting partitions in Hive tables that don't share the same location prefix with that of the table.

Fixes #12651

### DIFF
--- a/docs/en/operation/User-CLI.md
+++ b/docs/en/operation/User-CLI.md
@@ -1315,6 +1315,13 @@ This command will attach the database `hive_db_name` (of type `hive`) from the U
 When paths are mounted for `s3a://bucket1`, the mount option `aws.accessKeyId=abc` will be used,
 and when paths are mounted for `s3a://bucket2`, the mount option `aws.accessKeyId=123` will be used.
 
+Besides mount options, there are some additional properties with the `-o` options:
+  * `udb-hive.<UDB_PROPERTY>`: specify the UDB options for the Hive UDB. The options
+  are as follows
+    * `allow.diff.partition.location.prefix`: Whether to mount partitions that do not share
+  the same location prefix with table location(true/false, default false)
+
+
 ### Glue UDB
 For `glue` udb type, there are some additional properties with the `-o` options:
   * `udb-glue.<UDB_PROPERTY>`: specify the UDB options for the Glue UDB. The options
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
@@ -162,8 +162,12 @@ public class HiveDatabase implements UnderDatabase {
 
       for (Partition part : partitions) {
         AlluxioURI partitionUri;
-        if (part.getSd() != null && part.getSd().getLocation() != null
-            && ufsUri.isAncestorOf(partitionUri = new AlluxioURI(part.getSd().getLocation()))) {
+        if (part.getSd() != null && part.getSd().getLocation() != null) {
+          partitionUri = new AlluxioURI(part.getSd().getLocation());
+          if (!mConfiguration.getBoolean(Property.ALLOW_DIFF_PART_LOC_PREFIX)
+              && !ufsUri.isAncestorOf(partitionUri)) {
+            continue;
+          }
           hiveUfsUri = part.getSd().getLocation();
           String partName = part.getValues().toString();
           try {
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/Property.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/Property.java
@@ -11,12 +11,166 @@
 
 package alluxio.table.under.hive;
 
+import alluxio.table.common.udb.UdbProperty;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import org.slf4j.Logger;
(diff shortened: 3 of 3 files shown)


table/server/underdb/hive/src/main/java/alluxio/table/under/hive/Property.java
```diff
@@ -11,12 +11,166 @@
 
 package alluxio.table.under.hive;
```
hycsam, line 12: This file is using the same code structure as `alluxio.table.under.glue`

gpang:
ok to test

docs/en/operation/User-CLI.md
```diff
@@ -1315,6 +1315,13 @@ This command will attach the database `hive_db_name` (of type `hive`) from the U
 When paths are mounted for `s3a://bucket1`, the mount option `aws.accessKeyId=abc` will be used,
 and when paths are mounted for `s3a://bucket2`, the mount option `aws.accessKeyId=123` will be used.
 
+Besides mount options, there are some additional properties with the `-o` options:
+  * `udb-hive.<UDB_PROPERTY>`: specify the UDB options for the Hive UDB. The options
+  are as follows
+    * `allow.diff.partition.location.prefix`: Whether to mount partitions that do not share
```
gpang, line 1321: Can we make this property a general udb property, and not hive specific?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -162,8 +162,12 @@ private PathTranslator mountAlluxioPaths(Table table, List<Partition> partitions
 
       for (Partition part : partitions) {
         AlluxioURI partitionUri;
-        if (part.getSd() != null && part.getSd().getLocation() != null
-            && ufsUri.isAncestorOf(partitionUri = new AlluxioURI(part.getSd().getLocation()))) {
+        if (part.getSd() != null && part.getSd().getLocation() != null) {
+          partitionUri = new AlluxioURI(part.getSd().getLocation());
+          if (!mConfiguration.getBoolean(Property.ALLOW_DIFF_PART_LOC_PREFIX)
+              && !ufsUri.isAncestorOf(partitionUri)) {
```
gpang, line 168: Can this be a general property, since the glue integration also has a similar check: https://github.com/Alluxio/alluxio/blob/master/table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java#L263

gpang (commented):
@hycsam Thanks for this feature! I left a few questions.

docs/en/operation/User-CLI.md
yuzhu: does glue allow partitions to be outside of the table path?  if it does not,  then this property probably does not make sense for glue?

gpang: hrmmm, good question, I'm not sure how that works in glue. If it is possible to edit the glue metadata directly, then maybe?

hycsam: Both the `Property.java` files (Hive and Glue) now share the same code structure. I can move this builder code logic to their parent class, `UdbProperty.java`. We can decide later if necessary to add this property to Glue UDB.

hycsam: Tried abstracting the common code in `UdbProperty.java` but don't think that's a good idea, will leave as such. Also bumping this up, does this PR look OK for Hive UDB? I think we can decided whether to add that for Glue later.

apc999:
@hycsam any update?

hycsam:
@gpang Is there still pending req for adding this property to Glue catalog as well?

apc999:
@gpang  @yuzhu  any further concerns on this PR?
I would like to merge it soon for the next edge release

docs/en/operation/User-CLI.md
yuzhu: What challenges did you run into? I am concerned that we have two separate property classes and they will diverge further in the future.

hycsam: @yuzhu  Both the Builder pattern and the static property map (DEFAULT_KEYS_MAP variable in `Property.java`) are making it hard to abstract the code in the parent class. Don't want to make the code too complicated.

yuzhu: got it @hycsam

yuzhu (approved):
LGTM

hycsam:
@gpang @yuzhu How do I merge this?

apc999:

Issue #12356, opened 2020-10-20
Alluxio master failed to start after stopping and restart with backup


jainanuj07:
**Alluxio Version:**
  alluxio-2.4

**Describe the bug**
 I have to stop alluxio master to change my ufs sync interval. After stooping when I try to start the master is not able to start. 
 I then tried to start it with backup in that case also it fails and not able to start. 

It's giving this error 

java.io.IOException: io.grpc.StatusRuntimeException: CANCELLED: HTTP/2 error code: CANCEL
Received Rst Stream
        at alluxio.master.journal.raft.RaftJournalWriter.flush(RaftJournalWriter.java:100)
        at alluxio.master.journal.AsyncJournalWriter.doFlush(AsyncJournalWriter.java:295)
        at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: io.grpc.StatusRuntimeException: CANCELLED: HTTP/2 error code: CANCEL
Received Rst Stream
        at io.grpc.Status.asRuntimeException(Status.java:533)
        at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:449)
        at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
        at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
        at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
        at alluxio.grpc.GrpcChannel$ChannelResponseTracker$1$1.onClose(GrpcChannel.java:172)
        at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
        at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
        at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
        at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:426)
        at io.grpc.internal.ClientCallImpl.access$500(ClientCallImpl.java:66)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:689)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.access$900(ClientCallImpl.java:577)
[...]

PR #12658, opened 2020-12-17
Change to journal large table partitions separately

LuQQiu:
One of the PRs target to solve issue #12356
In the previous implementation, each table is considered as a single journal entry.
When a table has thousands of partitions, the journal entry size can be bigger than 30MB (an example journal entry with about 4000 partitions takes 31MB disk space). 
Alluxio embedded journal raft library - Ratis currently hard-coded the largest journal entry as 30MB which means that all journal entries bigger than 30MB is not allowed in the read path.

This PR change from journaling a table with a single big entry to multiple smaller entries by journaling table partitions in different entries.

Table entries in backup files created before this change are still big entries with all table partitions. Restoring from those backups or restarting the leading master will not be resolved by this PR. This change mainly takes effect when loading table meta into Alluxio for the first time.


table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -349,6 +355,9 @@ private boolean processJournalEntryInternal(Journal.JournalEntry entry,
     if (entry.hasAddTable()) {
       return applyAddTable(context, entry);
     }
+    if (entry.hasAddTablePartitions()) {
```
gpang, line 358: I think `AlluxioCatalog` also needs this: https://github.com/Alluxio/alluxio/blob/master/table/server/master/src/main/java/alluxio/master/table/AlluxioCatalog.java#L390

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -467,14 +508,55 @@ public boolean hasNext() {
...
-            "GetTableIteratorr#Iterator#remove is not supported.");
+            "GetTableIterator#Iterator#remove is not supported.");
+      }
+    };
+  }
+
+  private Iterator<Journal.JournalEntry> getTablePartitionsIterator() {
+    final Iterator<Table> it = getTables().iterator();
+    return new Iterator<Journal.JournalEntry>() {
+      private List<alluxio.proto.journal.Table.AddTablePartitionsEntry> mPartitionsEntries
```
gpang, line 526: Instead of 1 table iterator and partition list, should we just have a table iterator, and partition iterator? Everything the partition iterator is empty() we just move on to the next table and get the next partition iterator?

It seems like mixing an iterator and a list is a bit confusing to follow.

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -484,7 +566,8 @@ public void remove() {
     Journal.JournalEntry entry = Journal.JournalEntry.newBuilder().setUpdateDatabaseInfo(
         toJournalProto(getDatabaseInfo(), mName)).build();
     return CloseableIterator.noopCloseable(
-        Iterators.concat(Iterators.singletonIterator(entry), getTableIterator()));
+        Iterators.concat(Iterators.singletonIterator(entry),
+            getTableIterator(), getTablePartitionsIterator()));
```
gpang, line 570: Is it perfectly fine output all the tables and then all the individual partition batches?

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -410,6 +419,38 @@ private boolean applyAddTable(@Nullable JournalContext context, Journal.JournalE
...
+    alluxio.proto.journal.Table.AddTablePartitionsEntry addTablePartitions
+        = entry.getAddTablePartitions();
+    if (!addTablePartitions.getDbName().equals(mName)) {
+      return false;
+    }
+
+    mTables.compute(addTablePartitions.getTableName(), (key, existingTable) -> {
+      if (existingTable != null) {
+        if (addTablePartitions.getVersion() == existingTable.getVersion()) {
+          // this table is being removed, and has the expected version
```
gpang, line 433: What does this comment mean? What is being removed?

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -410,6 +419,38 @@ private boolean applyAddTable(@Nullable JournalContext context, Journal.JournalE
...
+          LOG.info("Adding partitions to table {}.{}", mName, addTablePartitions.getTableName());
```
gpang, line 434: can we also add the # partitions we are adding in the message?

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -410,6 +419,38 @@ private boolean applyAddTable(@Nullable JournalContext context, Journal.JournalE
...
+          if (context != null) {
+            context.append(entry);
+          }
+          existingTable.addPartitions(addTablePartitions);
+          return existingTable;
+        }
+        LOG.info("Will not add partitions to table {}.{}, because of mismatched versions. "
+                + "version-to-delete: {} existing-version: {}",
```
gpang, line 442: We are not deleting right?

gpang (changes requested):
@LuQQiu Thanks for this improvement! Are there any tests we can add? Maybe `TableMasterJournalIntegrationTest`?

table/server/master/src/main/java/alluxio/master/table/Database.java
LuQQiu: Added, thanks!

table/server/master/src/main/java/alluxio/master/table/Database.java
LuQQiu: I changed to add partitionIterator directly in the existing getTableIterator to be cleaner and safer. PTAL

table/server/master/src/main/java/alluxio/master/table/Database.java
LuQQiu: The table iterator change to include partition iterators so they are now placed together. This way seems to be safer

table/server/master/src/main/java/alluxio/master/table/Database.java
LuQQiu: Removed, thanks

table/server/master/src/main/java/alluxio/master/table/Database.java
LuQQiu: Added, thanks

table/server/master/src/main/java/alluxio/master/table/Database.java
LuQQiu: Changed, thanks

gpang (commented):
@LuQQiu Thanks! Is there a way to add a test to `TableMasterJournalIntegrationTest` (with a tiny `alluxio.table.journal.partitions.chunk.size`) to validate that journal replay will work?

LuQQiu:
@gpang Added the chunk size test to the TableMasterJournalIntegrationTest, PTAL, thanks!

gpang (approved):
@LuQQiu Thanks for this great improvement!

LGTM

LuQQiu:

PR #12659, opened 2020-12-17
Improve logging for tier mgmt task

jiacheliu3:
This change does 2 things:
1. When a tier mgmt task fails because the destination dir is full, it logs a warning instead of error
2. Adds a unit test that verifies a failed tier move issued by a tier mgmt task fail consistently

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
@@ -363,7 +363,7 @@ public class TieredBlockStore implements BlockStore {
       return;
     }
     throw new WorkerOutOfSpaceException(ExceptionMessage.NO_SPACE_FOR_BLOCK_MOVE,
-            moveOptions.getLocation(), blockId);
+        moveOptions.getLocation(), blockId);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -362,10 +362,8 @@ public void moveBlock(long sessionId, long blockId, BlockStoreLocation oldLocati
       }
       return;
     }
-    // TODO(bin): We are probably seeing a rare transient failure, maybe define and throw some
-    // other types of exception to indicate this case.
```
jiacheliu3, line 366: This comment is from https://github.com/Alluxio/alluxio/commit/530c4c3d5781a386b962c1cc16d24b3d5b249055 it doesn't make sense here anymore. We have changed how the space allocation works largely. When the move task failed, it can be due to many more reasons than just a transient error here, for example: 
1) The target location is full and this move operation does not evict
2) Eviction failed since the target is full of pinned blocks
3) We evict on the target location but the free space is stolen by someone else
4) maybe an unfound bug from concurrency

Basically this TODO can be removed.

gpang (approved):
LGTM

gpang:

PR #12662, opened 2020-12-18
Enable remote logging in k8s

jiacheliu3:
This change enables the Alluxio logserver to run in k8s. Alluxio services like masters and workers on k8s will be able to log to the logserver.


jiacheliu3:
The docker image part will be resolved separately by: https://github.com/Alluxio/alluxio/pull/12663

integration/kubernetes/helm-chart/alluxio/templates/_helpers.tpl
```diff
@@ -119,6 +119,28 @@ resources:
     {{- end }}
 {{- end -}}
 
+{{- define "alluxio.logserver.resources" -}}
```
nirav-chotai, line 122: This function is really not needed.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -310,3 +310,47 @@ fuse:
...
+  # It can be "persistentVolumeClaim" or "hostPath"
+  volumeType: persistentVolumeClaim
+  size: 4Gi
+  # Attributes to use if the log volume is PVC
+  pvcName: alluxio-logserver-logs
+  accessModes:
+    - ReadWriteOnce
+  storageClass: standard
+  # Attributes to use if the log volume is hostPath
+  hostPath: "/tmp/alluxio-logs" # The hostPath directory to use
```
nirav-chotai, line 355: What if I do not want to use any additional volume for logging as my ephermal storage is enough for my requirement, then I do not see any option here. I just want to run a remote logging server without any additional volume.

jiacheliu3: Could you elaborate a bit how you intend to provision your ephemeral storage for the logserver? In other words, i'm curious how much disk is given when you spawn a pod for the log server. 

If you are thinking about test usages where provisioning a PV would be overhead or a user, does supporting `emptyDir` make sense here?

nirav-chotai: @jiacheliu3 I was talking about not adding any volume at all. I just want to run logserver based on given memory (or may be spilling to disk) but not want to attach any volume for storing logs.

```
spec.containers[].resources.limits.ephemeral-storage
spec.containers[].resources.requests.ephemeral-storage
```
Ref: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage

jiacheliu3: @nirav-chotai Sure you do have control to the ephemeral storage, but I wouldn't recommend it because I feel it contradicts (at least partly) with why we want a logserver. To me the benefits of a log server are:
1. We have a centralized location for logs
2. We hold the logs when pods die

If we have the logs on the logserver but never persist that, we can still lose the logs. The logserver may be evicted for whatever reason, and you lose the logs. In the helm chart, the recommended (default) config will force the users to create and use PVs in multiple places. That was intended because we want the users to think about persistence for things like journal and worker's tiered storage. The same idea would apply to logs.

From that perspective, I guess my bottom line would be: it might be okay to allow using no volume at all, but that should be explicitly turned off. In other words, using a volume should be the default to be fool-proof.

Maybe you have your considerations or your design on how to use the logserver in your architecture? What are the benefits for you to only use ephemeral storage for the log server?

nirav-chotai: Yes, you can enforce the default to have volume but there should be a way to use logserver without volume. 

My use case, in dev like environments I would like to run separate logserver so that whenever I rollout my changes (frequent for dev), it will persist good amount of logs in the logserver (based on the capacity of the logserver pod) while recreating only master and worker pods. Production like environments I am pretty sure anyone would go for PVC only. 

IMO, anyone would try and test in dev first and if things work, then they configure PVC (not necessarily available in all environments) to meet production requirements.

nirav-chotai: And yes, your both points are valid even if I deploy logserver without volume

We have a centralized location for logs (yes)
We hold the logs when pods die (yes, logs will be there though master and workers pod die and re-created).

jiacheliu3: Sorry for the late reply. I see your point and that use case is legit. I will add the option to not use a volume for the log server.

integration/kubernetes/helm-chart/alluxio/templates/logging/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+    spec:
+      containers:
+        - name: alluxio-logserver
+          image: {{ .Values.image }}:{{ .Values.imageTag }}
+          imagePullPolicy: {{ .Values.imagePullPolicy }}
+          securityContext:
+            runAsUser: {{ .Values.user }}
+            runAsGroup: {{ .Values.group }}
+          {{- if .Values.logserver.resources  }}
+{{ include "alluxio.logserver.resources" . | indent 10 }}
```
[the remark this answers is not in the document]
jiacheliu3: IMO this makes the code shorter with fewer if-elses that break a future user's indentation. So I lean towards having a function.

integration/kubernetes/helm-chart/alluxio/values.yaml
jiacheliu3: @nirav-chotai I can do as my last [commit ](https://github.com/Alluxio/alluxio/pull/12662/commits/6213d9dc0e26f38bf04879d524faf833aa026cd3) to add a `none` option that creates no volume. But if `emptyDir` can just serve, I lean towards not adding another switch.

jiacheliu3: I think I forgot to add `emptyDir` to the available options. Initially there was only `pvc` and `hostPath`. If there were only `pvc` and `hostPath`, I think it makes much sense to ask for sth that does need a volume like `none`. But if `emptyDir` is available, I think setting this option to `emptyDir` will serve your need.

integration/docker/entrypoint.sh
```diff
@@ -51,6 +51,7 @@ function printUsage {
   echo -e " proxy                        \t Start Alluxio proxy"
   echo -e " fuse [--fuse-opts=opt1,...]  \t Start Alluxio FUSE file system, option --fuse-opts expects a list of fuse options separated by comma"
   echo -e " logserver                    \t Start Alluxio log server"
+  echo -e " logserver [--fuse-opts=opt1,...]  \t Start Alluxio FUSE file system, option --fuse-opts expects a list of fuse options separated by comma"
```
madanadit, line 54: Comment is stale. And why are there 2 logserver commands?

integration/kubernetes/helm-chart/alluxio/values.yaml
nirav-chotai: @jiacheliu3 Sorry, I was not available in the last few days. Yes, `emptyDir` will serve the need.

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -14,7 +14,7 @@
 {{- $hostNetwork := .Values.worker.hostNetwork }}
 {{- $hostPID := .Values.worker.hostPID }}
 {{- $fullName := include "alluxio.fullname" . }}
-
+{{- $logServerEnabled := .Values.logserver.enabled }}
```
nirav-chotai, line 17: Where this is going to be used in the worker daemonset?

integration/kubernetes/helm-chart/alluxio/templates/logserver/log-pvc.yaml
```diff
@@ -0,0 +1,39 @@
...
+    role: alluxio-logserver
+spec:
+  volumeMode: Filesystem
+  resources:
+    requests:
+      storage: {{ .Values.logserver.size }}
+  storageClassName: {{ .Values.logserver.storageClass }}
+  accessModes:
+  {{ toYaml .Values.logserver.accessModes | trim | indent 4 }}
+selector:
```
nirav-chotai, line 33: How this will work in the case of dynamic provisioning of distributed storage? 

With Ceph, I do not provision PV as it will automatically be created. And if that automatically created PV does not have these labels, then it will not be bound to the PVC, right?

integration/kubernetes/helm-chart/alluxio/values.yaml
jiacheliu3: Cool, will remove option `none`

integration/kubernetes/helm-chart/alluxio/templates/logserver/log-pvc.yaml
nirav-chotai: @jiacheliu3 , the above selectors, might fail in the case of Ceph PVC.

jiacheliu3: @nirav-chotai That's a very good point. 

**Static (manual) provision**
If we remove the `matchLabels`, it's harder to manage for users who are using static provisioning (manual). They may be creating a PV and that got mounted by undesired pods.

**Dynamic provision**
I'm not sure if it is possible to request a PV with default labels to a storageClass. Is it possible in your environment? I don't find anything online so I guess not?

If by either having `matchingLabels` or not having it, we have a hard time satisfying some use cases, do you have a better option?
Adding @madanadit to the discussion too.

nirav-chotai: @jiacheliu3 , below is an outcome of the current setup:

```
...
spec:
  volumeMode: Filesystem
  resources:
    requests:
      storage: 1Mi
  storageClassName: standard
  accessModes:
    - ReadWriteOnce
  selector:
    matchLabels:
      app: alluxio
      release: myrelease
      heritage: Helm
      role: alluxio-worker
```
I would take care of both the scenarios with the below change:

In `log-pvc.yaml`:
```
  {{- with .Values.logserver.pvc.selector }}
  selector:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

Now, for the scenario of static/manual provision, in `values.yaml`:
```
# Remote logging server
logserver:
  enabled: true
  ...
  pvc:
    selector:
      matchLabels:
        role: alluxio-logserver
        dc: mydatacenter
        region: myregion
```
The outcome would be like this:
```
...
spec:
  volumeMode: Filesystem
  resources:
    requests:
      storage: 4Gi
  storageClassName: standard
  accessModes:
      - ReadWriteOnce
  selector:
    matchLabels:
      dc: mydatacenter
      region: myregion
      role: alluxio-logserver
```

And, for the scenario of dynamic provision, in `values.yaml`:
```
# Remote logging server
logserver:
  enabled: true
  ...
  pvc:
    selector: {}
```

The outcome would be like this:
```
...
spec:
  volumeMode: Filesystem
  resources:
    requests:
      storage: 4Gi
  storageClassName: standard
  accessModes:
      - ReadWriteOnce
```

jiacheliu3: @nirav-chotai I'm actually working on the same as your suggestion. Will update the PR once I tested it :)

jiacheliu3:
@nirav-chotai PTAL, thanks!

integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
```diff
@@ -0,0 +1,106 @@
...
+          - containerPort: {{ .Values.logserver.ports.logging }}
+            name: logging
+          volumeMounts:
+            {{- if not (eq .Values.logserver.volumeType "none") }}
+          - name: alluxio-logs
+            mountPath: /opt/alluxio/logs
+            {{- end }}
+            {{- if .Values.secrets }}
+              {{- if .Values.secrets.logserver }}
+            {{- include "alluxio.master.secretVolumeMounts" . }}
```
madanadit, line 77: why does the logserver need master secrets?

madanadit:
Thanks @jiacheliu3 and @nirav-chotai. @jiacheliu3 can you please add docs as well before merging?

integration/kubernetes/helm-chart/alluxio/templates/_helpers.tpl
```diff
@@ -333,3 +355,34 @@ livenessProbe:
...
+    medium: {{ .Values.logserver.medium }}
+    sizeLimit: {{ .Values.logserver.size | quote }}
+{{- else }}
+- name: alluxio-logs
+  persistentVolumeClaim:
+    claimName: "{{ .Values.logserver.pvcName }}"
+{{- end }}
+{{- end -}}
+
+{{- define "alluxio.logserver.pvc.selector" -}}
```
nirav-chotai, line 377: Where this is being used?

jiacheliu3:
> Thanks @jiacheliu3 and @nirav-chotai. @jiacheliu3 can you please add docs as well before merging?

@madanadit Sure will do

integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
jiacheliu3: Oops, should be logserver secrets

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -313,6 +313,8 @@ fuse:
 #     alluxio-hdfs-config: hdfsConfig
 #   worker: # Shared by worker and jobWorker containers
 #     alluxio-hdfs-config: hdfsConfig
+#   logserver: # Used by the logserver container
+#     alluxio-hdfs-config: hdfsConfig
```
jiacheliu3, line 317: Not sure if there's a better example

madanadit: logserver doesn't need to talk to hdfs so the hdfs secret is not required, but his placeholder is fine

madanadit (approved):
Thanks @jiacheliu3

jiacheliu3:
@nirav-chotai Do you wanna take a final look before I merge this? Thanks!

jiacheliu3:

jiacheliu3:

jiacheliu3:

jiacheliu3:

jiacheliu3:

PR #12665, opened 2020-12-18
Fix ListBucket in S3 API v1 to return correct values

bradyoo:
The existing implementation is not v1 or v2, so this PR implements the v1 API. v2 will come in a later PR.


bradyoo:
Examples in action in S3Browser after the commit.

image

image

image

bradyoo:
Because S3Browser only supports ListObjectsV1 API. continuationToken changed to Marker. I'll be introducing ListObjectsV2 back soon so that it can support both when list-type=2

gpang:
test failure: 
```S3ClientRestApiTest.putBucketUnderNonMountPointDirectory:189 create bucket under non-mount-point directory should fail```

**Comment by bradyoo** [comment]
Yeah the removal of that restriction was intentional. Let me remove that test.

**Comment by gpang** [comment]
@bradyoo was the previous implementation the v2 listing api, and the new code is the v1 api?

**Comment by bradyoo** [comment]
the old one wasn't really a v2 api either. It had v2 parameters but it wasn't actually giving the correct results for those v2 parameters. This is a more correct implementation of v1 api.

**Comment by gpang** [comment]
@bradyoo I see. So is it fair to say that the old implementation was not a correct implementation for v1 nor v2, so this is now a more correct implementation of v1?

**Comment by bradyoo** [comment]
That is correct. I will have to also support/copy some parameter names around to support v2.

File: core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketOptions.java [hunk 1]
```diff
@@ -18,9 +18,10 @@
  * The options for list bucket operation.
  */
 public final class ListBucketOptions {
-  private String mContinuationToken = null;
+  private String mMarker = null;
   private String mMaxKeys = null;
   private String mPrefix = null;
+  private String mDelimiter = null;
```
gpang, line 24: is the delimiter used anywhere?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketOptions.java
```diff
@@ -93,20 +110,20 @@ public boolean equals(Object o) {
       return false;
     }
     ListBucketOptions that = (ListBucketOptions) o;
-    return Objects.equal(mContinuationToken, that.mContinuationToken)
+    return Objects.equal(mMarker, that.mMarker)
```
gpang, line 113: should delimiter be added to this?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketOptions.java
```diff
@@ -93,20 +110,20 @@ public boolean equals(Object o) {
...
         && Objects.equal(mMaxKeys, that.mMaxKeys)
         && Objects.equal(mPrefix, that.mPrefix);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(mContinuationToken, mMaxKeys, mPrefix);
+    return Objects.hashCode(mMarker, mMaxKeys, mPrefix);
```
gpang, line 120: should delimiter be added to this?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketOptions.java
```diff
@@ -93,20 +110,20 @@ public boolean equals(Object o) {
...
   }
 
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
-        .add("continuationToken", mContinuationToken)
+        .add("continuationToken", mMarker)
```
gpang, line 126: should this field be called marker?

Should delimiter be added to this?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -11,56 +11,49 @@
...
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
- * Get bucket result defined in http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html
+ * Get bucket result defined in https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
```
gpang, line 30: I thought this was supposed to be the v1 version?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -11,56 +11,49 @@
...
+  private String mName;
   /**
    * Returns the number of keys included in the response. The value is always less than or equal
    * to the mMaxKeys value.
    */
-  private int mKeyCount = 0;
-  /* The maximum number of keys returned in the response body. */
-  private int mMaxKeys = S3Constants.S3_DEFAULT_MAX_KEYS;
-  /**
-   * Specifies whether or not all of the results were returned. If the number of
```
gpang, line 53: Can we keep this comment and the comment for continuation? (or an updated version of the comment)

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
-        // If continuation token does not exist in the object list, find the first element which is
-        // greater than the token.
-        startIndex = (-1) * startIndex - 1;
-      }
+    mMarker = options.getMarker();
+
+    Collections.sort(children, new URIStatusComparator());
+
+    final List<URIStatus> keys = children.stream()
+        .filter((status) -> status.getPath().compareTo(mMarker) > 0)
```
gpang, line 81: Will there always be a marker defined?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
+        .limit(mMaxKeys)
+        .collect(Collectors.toList());
+
+    mKeyCount = keys.size();
+    mIsTruncated = mKeyCount == mMaxKeys;
+
+    if (mIsTruncated) {
+      mNextMarker = keys.get(keys.size() - 1).getPath();
```
gpang, line 89: is it ever possible that the `keys.size() == 0`?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
     }
 
-    for (int i = startIndex; i < objectsList.size(); i++) {
+    final Map<Boolean, List<URIStatus>> typeToStatus = keys.stream()
+        .collect(Collectors.groupingBy((status) -> status.isFolder()));
+    final List<URIStatus> objectsList = typeToStatus.getOrDefault(false, new ArrayList<>());
```
gpang, line 94: I think both of these can be `Collections.emptyList()`

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
+    final List<URIStatus> prefixList = typeToStatus.getOrDefault(true, new ArrayList<>());
+
+    mContents = new ArrayList<>();
+    for (int i = 0; i < objectsList.size(); i++) {
```
gpang, line 98: can this loop be `for (URIStatus status : objectsList) {`?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
-      }
-      if (mPrefix != null && !objectKey.startsWith(mPrefix)) {
-        continue;
-      }
-      // TODO(chaomin): set ETag once there's a way to get MD5 hash of an Alluxio file.
-      // TODO(chaomin): construct the response with CommonPrefixes when delimiter support is added.
+
       mContents.add(new Content(
-          status.getPath().substring(mName.length() + 1),
+          status.getPath().substring(mName.length() + 2),
```
gpang, line 102: why is this + 2 and not + 1 as before?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
-          S3Constants.S3_EMPTY_ETAG,
-          String.valueOf(status.getLength()),
-          S3Constants.S3_STANDARD_STORAGE_CLASS));
-      mKeyCount++;
+          String.valueOf(status.getLength())
+      ));
+    }
+
+    final ArrayList<String> commonPrefixes = new ArrayList<>();
+    for (int i = 0; i < prefixList.size(); i++) {
```
gpang, line 109: can this be `for (URIStatus status : prefixList) {`?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -71,58 +64,56 @@ public ListBucketResult() {}
...
+
+    mKeyCount = keys.size();
+    mIsTruncated = mKeyCount == mMaxKeys;
+
+    if (mIsTruncated) {
+      mNextMarker = keys.get(keys.size() - 1).getPath();
     }
 
-    for (int i = startIndex; i < objectsList.size(); i++) {
+    final Map<Boolean, List<URIStatus>> typeToStatus = keys.stream()
```
gpang, line 92: It looks like we group by if each status is a folder or not a folder, then process each group. Instead, can we just loop over each of these status once, and just check if it is a folder or not, and do the appropriate thing? It seems like we are looping over this list (and creating lists) more times than necessary.

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
```diff
@@ -353,228 +331,6 @@ public void putObjectWithNoMD5() throws Exception {
...
-    // Verify
-    HashMap<String, String> maxKeysParam = new HashMap<>();
-    maxKeysParam.put("continuation-token", continuationToken);
-    new TestCase(mHostname, mPort, S3_SERVICE_PREFIX + AlluxioURI.SEPARATOR + bucket, maxKeysParam,
-        HttpMethod.GET, expected,
-        TestCaseOptions.defaults().setContentType(TestCaseOptions.XML_CONTENT_TYPE)).run();
-  }
-
-  @Test
-  public void listEmptyBucket() throws Exception {
```
gpang, line 559: why are all of these tests removed?

gpang (changes requested):
@bradyoo Thanks! I left some comments.

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
bradyoo: Almost none of the tests were correct and as noted in standup. I'm taking the time to write new tests for all of the behaviors here in the next PR.

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
bradyoo: because +1 was wrong and you got an extra character in your S3Browser result :/

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
bradyoo: I thought the code was a lot cleaner when i do it this way because the things that get done for folder vs files are very different.

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
bradyoo: not if mIsTruncated is True unless mMaxKeys is 0.

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
bradyoo: In particular, this needs to remove 2 slashes. The both ends of "/<mName>/"

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
bradyoo: Marker is set to be an empty string is the rest api implementation itself.

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
bradyoo: I just don't think this is much of a big deal because it's only 2 extra lists that is getting created.

bradyoo (commented):
Applied feedback and responded to some other comments.

bradyoo:
Feedback addressed

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
gpang: Can we do this here too?

gpang (commented):
@bradyoo I just had a minor comment.

bradyoo:
@gpang Oops. Now addressed.

gpang (approved):
LGTM

gpang:

PR #12684, opened 2020-12-28
Add code coverage to PR runs

ZacBlanco:
This PR will add code coverage report creation and aggregation via codecov.io.

A few notes:
- Java 11 coverage with JaCoCo didn't seem to generate properly, so we don't collect coverage information on Java 11
- Integration tests from two of the matrix modules on java 8 seem to corrupt the Jacoco output somehow. Unfortunately, we couldn't find a way around this, so we skip coverage report generation and uploading for alluxio.server.ft and alluxio.client.cli integration tests.

Additionally, the actions which do upload coverage reports are also now configured to run once per day on the master branch. This will allow PR coverage reports to be compared to recent master commits.


bradyoo:
This seems to add about 2~3 minutes to the test. @gpang let me know what you think? We could do a few runs with and without this to confirm because it may just be a sample size issue.

ZacBlanco:
One other note - it seems that something with the way our tests are run causes the reports to be corrupted on a somewhat often basis. I've tried debugging to figure out why, but couldn't some up with a solid solution.

```
Error: 8.441 [ERROR] Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.6:report (default-cli) on project alluxio-core-transport: An error has occurred in JaCoCo report generation.: Error while creating report: malformed input around byte 23 -> [Help 1]
```

Errors like the above I've found come up with some kind of regularity and it prevents coverage from being uploaded (as well as failing the build). I have a suspect that the corruption could be a side-effect of powermock, but I don't have concrete proof. I'm not sure what other weird things we do in our tests that could causes this. I've seen it happen across unit tests and the integration tests in `alluxio.client.cli` and `alluxio.server.ft`

gpang:
could the jacoco version be related to the random failures? I don't think I've seen that error before.

ZacBlanco:
It is possible, but I doubt it since there don't seem to be reports of this issue on the jacoco project itself. Given the number of people who rely on jacoco I would have expected bug reports if this is a new bug in a later version

ZacBlanco:
I've downgraded back to 0.8.5 and still see the same errors

gpang:
I don't know if this is the same issue, but could some of these suggestions work? https://stackoverflow.com/questions/55381133/an-error-has-occured-in-jacoco-report-generation

ZacBlanco:
I tried these on my test branch against my own repo and didn't have any success...I'll see if we can get any successful runs here

gpang:
@ZacBlanco This is how we ran it on jenkins in the past: https://github.com/Alluxio/alluxio/blob/master/dev/jenkins/build.sh#L90

Are there differences between that and your command?

ZacBlanco:
@gpang, I think the command is the same for report generation. I'm not sure what the problem is :/ 

Either way, you and @bradyoo can review this now. I think I've covered everything since the last discussion

ZacBlanco:
@gpang @bradyoo friendly reminder about this PR

bradyoo:
LGTM

bradyoo:

PR #12695, opened 2021-01-04
Flush when written journal reach to the given batch size

LuQQiu:
There are always chances that some new big journal entries are added unexpectedly which violates the underlying assumption of the raft implementation. 
This PR reduces the possibilities of flushing a single big journal entry by introducing journal size check and flush when reach journal batch size.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1569,15 +1569,6 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE =
-      new Builder(Name.MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE)
-          .setDefaultValue("1MB")
-          .setDescription("Embedded journal flush journal entries in batches to improve flush performance. "
-              + "When the current journal batch size in bytes is bigger than the configured batch size, "
-              + "an auto journal flush is triggered to avoid journal batch size grows unbounded")
-          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
-          .setScope(Scope.MASTER)
-          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_PORT =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_PORT)
           .setDescription("The port to use for embedded journal communication with other masters.")
@@ -5187,8 +5178,6 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.embedded.journal.entry.size.max";
     public static final String MASTER_EMBEDDED_JOURNAL_FLUSH_SIZE_MAX =
         "alluxio.master.embedded.journal.flush.size.max";
-    public static final String MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE =
-        "alluxio.master.embedded.journal.flush.batch.size";
     public static final String MASTER_EMBEDDED_JOURNAL_HEARTBEAT_INTERVAL =
         "alluxio.master.embedded.journal.heartbeat.interval";
     public static final String MASTER_EMBEDDED_JOURNAL_PORT =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1569,6 +1569,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE)
+          .setDefaultValue("1MB")
+          .setDescription("Embedded journal flush journal entries in batches to improve flush performance. "
+              + "When the current journal batch size in bytes is bigger than the configured batch size, "
+              + "an auto journal flush is triggered to avoid journal batch size grows unbounded")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_PORT =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_PORT)
           .setDescription("The port to use for embedded journal communication with other masters.")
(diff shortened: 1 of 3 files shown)


LuQQiu:
Testing to guarantee this PR does not introduce journal perf downgrade is in progress

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1569,6 +1569,15 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_FLUSH_BATCH_SIZE)
+          .setDefaultValue("1MB")
+          .setDescription("Embedded journal flush journal entries in batches to improve flush performance. "
```
gpang, line 1575: I cannot tell what the difference is between this parameter and `alluxio.master.embedded.journal.flush.size.max`. They look like very similar parameters. Could you add the differences and relationship (if there is any) between the two?

gpang (changes requested):
@LuQQiu Thanks! I left a minor question

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: That's true. Actually, it should be related to the journal entry size max, so i change to directly induct from the journal entry size max.

Journal entry size max defines the hard limit of a single journal entry allowed in ratis.
We need to define the single journal size max to be flushed to the ratis, could be set to 1/2 or 1/3 of the hard limit provided.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
```diff
@@ -116,6 +127,7 @@ public void flush() throws IOException, JournalClosedException {
             mWriteTimeoutMs), e);
       }
       mJournalEntryBuilder = null;
+      mCurrentJournalEntrySize.set(0);
```
bf8086, line 130: (nit) is it necessary since you have set it in line 87?

bf8086 (commented):
@LuQQiu Thanks for the improvement! Would it be possible to add a unit test for this feature?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
LuQQiu: No necessary, removed! Good catch

LuQQiu:
@bf8086 @gpang Addressed the comments and add a new unit test, PTAL

LuQQiu:
Ran some embedded journal related tests, didn't find significant performance differences.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
```diff
@@ -62,25 +64,34 @@ public RaftJournalWriter(long nextSequenceNumberToWrite,
...
     mLastCommittedSequenceNumber = new AtomicLong(-1);
+    mCurrentJournalEntrySize = new AtomicLong(0);
     mClient = client;
     mClosed = false;
     mWriteTimeoutMs =
         ServerConfiguration.getMs(PropertyKey.MASTER_EMBEDDED_JOURNAL_WRITE_TIMEOUT);
+    // journal entry size max is the hard limit set by underlying ratis
+    // use a smaller value to guarantee we don't pass the hard limit
+    mFlushBatchBytes =
+        ServerConfiguration.getBytes(PropertyKey.MASTER_EMBEDDED_JOURNAL_ENTRY_SIZE_MAX) / 3;
```
gpang, line 75: `max / 3` is basically a heuristic, right? What would happen if this was larger, like `max / 2`? What are the downsides?

gpang (commented):
@LuQQiu Thanks! I left a question

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
LuQQiu: Both max/3 and max/2 have no obvious difference in my new flush() in write().
If an entry is bigger than max, i will directly log an error. both max/3 and max/2 cannot avoid the failure in this case.

I set max / 3 mainly because we currently give the entry size max 10MB which is a relatively big number (the default entry size max is 3MB). 3MB journal entry size is big enough for one flush().

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
```diff
@@ -82,16 +85,23 @@ public void write(JournalEntry entry) throws IOException, JournalClosedException
...
       mCurrentJournalEntrySize.set(0);
     }
     LOG.trace("Writing entry {}: {}", mNextSequenceNumberToWrite, entry);
     mJournalEntryBuilder.addJournalEntries(entry.toBuilder()
         .setSequenceNumber(mNextSequenceNumberToWrite.getAndIncrement()).build());
-    if (mCurrentJournalEntrySize.addAndGet(entry.getSerializedSize()) > mFlushBatchBytes) {
-      flush();
+    long size = entry.getSerializedSize();
+    if (size > mEntrySizeMax) {
+      LOG.error("Journal entry size ({}) is bigger than the max allowed size ({}) defined by {}",
```
gpang, line 100: This is confusing, since the max entry size is really `MASTER_EMBEDDED_JOURNAL_ENTRY_SIZE_MAX`, but `mEntrySizeMax` will be 1/3 of that. So, this message will say something like `entry size (150) is bigger than max (100) defined by alluxio.master.embedded.journal.entry.size.max` However, `alluxio.master.embedded.journal.entry.size.max` will be set to 300.

gpang (commented):
@LuQQiu I had a questions

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
LuQQiu:
```
    mEntrySizeMax = ServerConfiguration
        .getBytes(PropertyKey.MASTER_EMBEDDED_JOURNAL_ENTRY_SIZE_MAX);
    mFlushBatchBytes = mEntrySizeMax / 3;
```
entry size max = 10MB
flush batch bytes = 10MB / 3 = 3.3MB
here we complain that a single entry size is bigger than the entry size max 10MB

gpang: ahhh, I see. I think I got confused with entry size and batch size.

Is this the only place we error when the entry is too large? Or will we get another error somewhere else in ratis?

gpang (commented):
@LuQQiu

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
LuQQiu: this place just gives clue for why Ratis flush failed. The actual error is logged by https://github.com/Alluxio/alluxio/pull/12649

gpang (approved):
LGTM

LuQQiu:

PR #12701, opened 2021-01-06
Add S3 api tests and maxKeys

bradyoo:


tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
```diff
@@ -100,74 +116,143 @@ public void before() throws Exception {
...
-    final String bucket = "bucket";
-    createBucketRestCall(bucket);
-    // Verify the directory is created for the new bucket.
-    AlluxioURI uri = new AlluxioURI(AlluxioURI.SEPARATOR + bucket);
-    Assert.assertTrue(mFileSystemMaster
-        .listStatus(uri, ListStatusContext.defaults()).isEmpty());
+  public void listAllMyBuckets() throws Exception {
+    mFileSystem.createDirectory(new AlluxioURI("/bucket0"));
+    SetAttributePOptions setAttributeOptions =
+        SetAttributePOptions.newBuilder().setOwner("user0").build();
```
bradyoo, line 122: I would like to not set owner like this separately but instead have a file system that does this properly for me. (because I'm going to need it to correct various creation apis with authorization). How do I do that?

gpang: Can we use `AuthenticatedUserRule` for this purpose?

gpang (commented):
@bradyoo I left a response

bradyoo:
@gpang Is this the right way to do it?

final AuthenticatedUserRule user0 = new AuthenticatedUserRule("user0", ServerConfiguration.global());

    user0.apply(new Statement() {
      @Override
      public void evaluate() throws Throwable {
        mFileSystem.createDirectory(new AlluxioURI("/bucket0"));
      }
    }, null);

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
gpang: Does it matter that the cluster is using `.setProperty(PropertyKey.SECURITY_AUTHENTICATION_TYPE, AuthType.NOSASL.getAuthName())`? It may have to be changed to SIMPLE.

gpang (commented):
@bradyoo I left a comment

bradyoo:
That may be the reason why the other attempts were failing before. But given this current design, it doesn't really matter.

gpang (approved):
LGTM

gpang:

PR #12710, opened 2021-01-12
Add Directory support for S3 PutObject + Debug Logging Support

bradyoo:
When the Key ends with a separator, it means a directory should be added

### DIFF
--- a/core/server/proxy/src/main/java/alluxio/proxy/s3/logging/RequestLoggingFilter.java
+++ b/core/server/proxy/src/main/java/alluxio/proxy/s3/logging/RequestLoggingFilter.java
@@ -1,0 +1,26 @@
+package alluxio.proxy.s3.logging;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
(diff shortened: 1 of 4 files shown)


core/server/proxy/src/main/java/alluxio/proxy/s3/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Helper class for adding logs to request methods for debugging purposes.
+ */
+@Provider
+public class RequestLoggingFilter implements ContainerRequestFilter {
+  private static final Logger LOG = LoggerFactory.getLogger(RequestLoggingFilter.class);
+
+  @Override
+  public void filter(ContainerRequestContext context) throws IOException {
+    LOG.info("Method: " + context.getMethod());
```
gpang, line 31: Do these need to be info level?

gpang (commented):
@bradyoo just a minor question

core/server/proxy/src/main/java/alluxio/proxy/s3/logging/RequestLoggingFilter.java
bradyoo: This code is not actually included in any of our codebase by default. You have to add it with Logged annotation. So in that case, info is just more convenient.

gpang (approved):
LGTM

gpang:

PR #12711, opened 2021-01-12
Add null check in block ids in MutableInodeFile

bradyoo:


bradyoo:
The fact that this code is nullpointering in Java 8 Integration Tests is concerning.

bradyoo:
@gpang Thought the nullpointers were initially interesting but they seem to happening somewhere completely different. For now, I think this PR is good.

bradyoo:
Take a look, I added some null checks and also guarded the object during newBlockId(). ArrayList adds the item first and then increments size.

PR #12714, opened 2021-01-13
Write async persistent tmp file to sub-directory of mount point

HelloHorizon:
Currently, Alluxio writes the temporary file of async persistence to the path, which is the same as the destination file path. The computing framework, like presto, will pick up the tmp files, which is leftover there for some reason, fail to clean up, etc.

This change writes the tmp file of async persistent to the sub dir of the mount point.

### DIFF
--- a/core/base/src/main/java/alluxio/util/io/PathUtils.java
+++ b/core/base/src/main/java/alluxio/util/io/PathUtils.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -169,6 +170,56 @@ public final class PathUtils {
       return AlluxioURI.SEPARATOR;
     }
     return parent;
+  }
+
+  /**
+   * Join two path elements for ufs, separated by {@link AlluxioURI#SEPARATOR}.
+   *
+   * For example,
+   *
+   * <pre>
+   * {@code
+   * concatUfsPath("s3://myroot/", "filename").equals("s3://myroot/filename");
+   * concatUfsPath("s3://", "filename").equals("s3://filename");
+   * }
+   * </pre>
+   *
+   * @param base base path
+   * @param path path element to concatenate
+   * @return joined path
+   */
+  public static String concatUfsPath(String base, String path) {
+    Pattern basePattern = Pattern.compile("(...)\\/\\/");
+    // use conCatPath to join path if base is not starting with //
+    if (!basePattern.matcher(base).matches()) {
+      return concatPath(base, path);
+    } else {
+      Preconditions.checkArgument(base != null, "Failed to concatPath: base is null");
+      Preconditions.checkArgument(path != null, "Failed to concatPath: a null set of paths");
+      String trimmedPath = SEPARATOR_MATCHER.trimFrom(path);
+      StringBuilder output = new StringBuilder(base.length() + trimmedPath.length());
+      output.append(base);
+      output.append(trimmedPath);
+      return output.toString();
+    }
+  }
+
+  /**
+   * Get temp path for async persistence job.
+   *
+   * "s3://"
+   *
+   * @param path ufs path
+   * @return ufs temp path
+   */
+  public static String getPersistentTmpPath(String path) {
+    String tempPath;
+    String fileName = FilenameUtils.getName(path);
+    String parentPath = path.substring(0, path.length() - fileName.length());
+    tempPath = concatUfsPath(parentPath, "alluxio_persistent_job");
+    tempPath = concatPath(tempPath, fileName);
+    tempPath = temporaryFileName(System.currentTimeMillis(), tempPath);
+    return tempPath;
   }
 
   /**
--- a/core/base/src/main/java/alluxio/util/io/PathUtils.java
+++ b/core/base/src/main/java/alluxio/util/io/PathUtils.java
@@ -174,6 +174,22 @@ public final class PathUtils {
   }
 
   /**
+   * Gets the parent of the the file at a ufs path.
+   *
+   * @param path the ufs path
+   * @return the parent path of the file; this is "/" if the given path is the root
+   */
+  public static String getUfsParent(String path) {
+    String name = FilenameUtils.getName(path);
+    String parent = path.substring(0, path.length() - name.length() - 1);
+    if (parent.isEmpty()) {
+      // The parent is the root path without any prefix.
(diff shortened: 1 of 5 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3797,14 +3798,19 @@ private void handleReady(long fileId) throws AlluxioException, IOException {
...
-          tempUfsPath = PathUtils.temporaryFileName(
-              System.currentTimeMillis(), resolution.getUri().toString());
+          // make temp path for temp file to avoid the
+          // error reading (failure of temp file clean up)
+          String mountPointUri = resolution.getUfsMountPointUri().toString();
+          tempUfsPath = PathUtils.concatUfsPath(mountPointUri,
+              PathUtils.getPersistentTmpPath(resolution.getUri().toString()));
+          LOG.debug("Generate tmp ufs path {} from ufs path {} for persistence.",
+              tempUfsPath, resolution.getUri().toString());
+          LOG.info("Temp file {} for File persistent.", tempUfsPath);
```
calvinjia, line 3808: Include the file name as well?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3963,6 +3969,60 @@ private void handleSuccess(PersistJob job) {
...
+                      if (ufs.isDirectory(dir)) {
+                        LOG.debug("UFS directory {} already exists", dir);
+                      } else {
+                        LOG.info("UFS path {} is an existing file", dir);
+                      }
+                    }
+                  } catch (IOException e) {
+                    LOG.error("Failed to create UFS directory {} with correct permission", dir, e);
+                  }
+                }
```
calvinjia, line 4025: Should this logic be a method?

calvinjia (commented):
Is it guaranteed that `.alluxio_ufs_persistence` will exist and what happens if external services modify it? Can we make the special directory configurable?

HelloHorizon:
@calvinjia The name of `.alluxio_persistence_job` follow the same naming logic as `alluxio.worker.BlockUtils#getUfsBlockPath`. If this path is get removed, the persistence job will try to create this directory when writing the temp file. I can make this directory name configurable, but I think the location of this path should below the mount point. Otherwise, the computing framework will still have a chance to pick up the leftover temp files.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: I meant to have both paths in the `INFO` level log

calvinjia:
@HelloHorizon Yes, the path must be relative to the mount point. If we have the same guarantees as ufsBlocks that's fine.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
HelloHorizon: I see, changed debug log to info log with both path.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
HelloHorizon: Yeah, it make sense, changes to `createParentPath` method.

calvinjia (approved):

core/base/src/main/java/alluxio/util/io/PathUtils.java
```diff
@@ -171,6 +173,58 @@ public static String getParent(String path) throws InvalidPathException {
...
+   * @param path ufs path
+   * @return ufs temp path with UUID
+   */
+  public static String getPersistentTmpPath(String path) {
+    StringBuilder tempFilePath = new StringBuilder();
+    StringBuilder tempFileName = new StringBuilder();
+    String fileName = FilenameUtils.getName(path);
+    tempFilePath.append(".alluxio_ufs_persistence/");
+    String uuid = UUID.randomUUID().toString();
+    tempFileName.append(fileName);
```
gpang, line 220: would it be useful to have the date in the prefix of the filenames? Then, it could be easy to see which files are "old"?

gpang (changes requested):
@HelloHorizon Thanks! I left a few comments.

gpang (approved):
LGTM

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4029,6 +4037,68 @@ private void handleSuccess(PersistJob job) {
...
+     * Create parent path if there isn't exiting ancestors path for final persistence file.
+     *
+     * @param inodes List of inodes
+     * @param ufsPath ufs path
+     * @param ufs under file system
+     */
+    private void createParentPath(List<Inode> inodes, String ufsPath, UnderFileSystem ufs)
+        throws IOException {
+      Stack<Pair<String, Inode>> ancestors = new Stack<>();
+      int curInodeIndex = inodes.size() - 2;
```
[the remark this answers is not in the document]
gpang: I see, so we are assuming it is a file, so it has to have a parent directory, so we are guaranteed the size is >= 2.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4029,6 +4037,68 @@ private void handleSuccess(PersistJob job) {
...
+      // get file path
+      AlluxioURI curUfsPath = new AlluxioURI(ufsPath);
+      // get the parent path of current file
+      curUfsPath = curUfsPath.getParent();
+      // Stop when the directory already exists in UFS.
+      while (!ufs.isDirectory(curUfsPath.toString()) && curInodeIndex >= 0) {
```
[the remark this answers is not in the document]
gpang: Well, what if the UFS path (s3 in this example) is different from the alluxio path? does this still work?

For example:
alluxio `/a/b/c/s3` path is mounted to s3 `s3://bucket/`

Then, the s3 paths are much shorter (no parents) than the alluxio paths (curInodeIndex).

Does this still work?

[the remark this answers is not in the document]
HelloHorizon: In that case, the ufs will return true(will not iterate over s3://bucket/) when check `s3://bucket/`, it will still work

HelloHorizon:

HelloHorizon:

PR #12715, opened 2021-01-13
Make add operations on DefaultBlockDeletionContext thread safe

bradyoo:
Addresses https://github.com/Alluxio/alluxio/issues/12722

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultBlockDeletionContext.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultBlockDeletionContext.java
@@ -19,20 +19,21 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Class for tracking which blocks need to be deleted, and deleting them on close.
  */
 public final class DefaultBlockDeletionContext implements BlockDeletionContext {
   private final List<BlockDeletionListener> mListeners;
-  private final List<Long> mBlocks;
+  private final ConcurrentHashMap.KeySetView<Long, Boolean> mBlocks;
 
   /**
    * @param listeners listeners to call for each deleted block when the context is closed
    */
   public DefaultBlockDeletionContext(BlockDeletionListener... listeners) {
     mListeners = Arrays.asList(listeners);
-    mBlocks = new ArrayList<>(0);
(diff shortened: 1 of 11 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultBlockDeletionContext.java
```diff
@@ -19,20 +19,21 @@
...
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Class for tracking which blocks need to be deleted, and deleting them on close.
  */
 public final class DefaultBlockDeletionContext implements BlockDeletionContext {
   private final List<BlockDeletionListener> mListeners;
-  private final List<Long> mBlocks;
+  private final ConcurrentHashMap.KeySetView<Long, Boolean> mBlocks;
```
gpang, line 29: Can we just use a concurrent list or queue? I think `ConcurrentLinkedQueue` might work?

gpang (commented):
@bradyoo Thanks for this fix!

Can we also get a unit test? I think we could probably have a simple one where we force a parallel sync after we delete all the files in a directory in UFS. We should be able to get it to fail if there are multiple files and multiple blocks per file.

something like: 
- create ufs files (with multiple blocks) in a dir
- list dir (with sync)
- delete all those ufs files
- list dir (with sync)

core/server/master/src/main/java/alluxio/master/file/DefaultBlockDeletionContext.java
bradyoo: I don't think no matter how much we try to write this unit test, it's not really going to fail on the github actions machines.

bradyoo:
I don't think no matter how much we try to write this unit test, it's not really going to fail on the github actions machines.

core/server/master/src/main/java/alluxio/master/file/DefaultBlockDeletionContext.java
bradyoo: I can do ConcurrentLinkedQueue. Usually though the ConcurrentHashMap related implementations tend to be the cleanest.

bradyoo:
The fix you have requested is made though unit test is another thing.

gpang (approved):
LGTM

gpang:

bradyoo:

gpang:

PR #12729, opened 2021-01-15
Fix Python S3 API

bradyoo:


bradyoo:
I mistakenly thought that the accesskey would be empty if user didn't provide anything. In reality, it was something like AWS :<Secret>

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
```diff
@@ -125,7 +125,12 @@ private static String getUserFromAuthorization(String authorization) {
       return null;
     }
 
-    return stripped.substring(0, colonIndex);
+    final String user = stripped.substring(0, colonIndex);
```
gpang, line 128: is there a unittest we can add to verify this behavior?

gpang (commented):
@bradyoo

bradyoo:
Unit test added.

gpang (approved):
LGTM

gpang:

PR #12743, opened 2021-01-20
Fix backup lock acquisition

yuzhu:
This PR disables the interruption threads once we give up on the back up process. Previous behavior causes all future RPCs to fail because it is not able to obtain shared state lock.

This PR also updates the default value for alluxio.master.lost.worker.file.detection.interval to 5mins from 10secs.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1465,7 +1465,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_DAILY_BACKUP_STATE_LOCK_TIMEOUT =
       new Builder(Name.MASTER_DAILY_BACKUP_STATE_LOCK_TIMEOUT)
-          .setDefaultValue("12h")
+          .setDefaultValue("2h")
           .setDescription("The max duration for a grace-cycle.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1822,10 +1822,17 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL =
       new Builder(Name.MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL)
-          .setAlias("alluxio.master.worker.heartbeat.interval")
           .setDefaultValue("5min")
+          .setDescription("The interval between Alluxio master detections to find lost "
+              + "files based on updates from Alluxio workers.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.SERVER)
+          .build();
+  public static final PropertyKey MASTER_LOST_WORKER_DETECTION_INTERVAL =
+      new Builder(Name.MASTER_LOST_WORKER_DETECTION_INTERVAL)
+          .setDefaultValue("10sec")
           .setDescription("The interval between Alluxio master detections to find lost workers "
-              + "and files based on updates from Alluxio workers.")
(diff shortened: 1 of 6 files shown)


gpang:
@yuzhu Can you add some details to the PR description?

gpang:
@yuzhu Also, there are a few test failures, probably from this change?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -207,10 +211,9 @@ public LockResource lockExclusive(StateLockOptions lockOptions)
           throw new TimeoutException(ExceptionMessage.STATE_LOCK_TIMED_OUT
               .getMessage(lockOptions.getGraceCycleTimeoutMs() + mForcedDurationMs));
         }
-      } catch (InterruptedException e) {
+      } finally {
```
bf8086, line 214: Would it be a problem that the success path doe another `deactivateInterruptCycle` at line 224 when the write lock is release?

bf8086 (changes requested):
@yuzhu thanks for the fix!

core/server/common/src/main/java/alluxio/master/StateLockManager.java
yuzhu: Good catch, addressed.

yuzhu:
@gpang @bf8086 PTAL, test passes now.

core/server/master/src/test/java/alluxio/master/journal/JournalContextTest.java
```diff
@@ -149,11 +151,12 @@ public void pauseBlocksJournalContext() throws Exception {
...
       // after un-pausing, new journal contexts can be created
       lock.close();
+      thread2.run();
       CommonUtils.waitFor("journal context created", journalContextCreated::get,
           WaitForOptions.defaults().setTimeoutMs(5 * Constants.SECOND_MS).setInterval(10));
     } finally {
-      thread.interrupt();
-      thread.join();
+      thread2.interrupt();
+      thread2.join();
```
bf8086, line 159: Shall we clean up `thread` here as well just to be safe?

bf8086 (approved):
LGTM.

core/server/master/src/test/java/alluxio/master/journal/JournalContextTest.java
yuzhu: done

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -142,6 +142,10 @@ public LockResource lockShared() throws InterruptedException {
           exclusiveOnlyRemainingMs, PropertyKey.Name.MASTER_BACKUP_STATE_LOCK_EXCLUSIVE_DURATION);
       throw new IllegalStateException(safeModeMsg);
     }
+    // Check if the active interrupt cycle is on, we do not try to obtain the shared lock
```
yuzhu, line 145: @bf8086 should we remove this part for now, because this has not been observed? and probably would not happen because of fair locking?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1822,10 +1822,17 @@ public String toString() {
           .build();
   public static final PropertyKey MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL =
       new Builder(Name.MASTER_LOST_WORKER_FILE_DETECTION_INTERVAL)
-          .setAlias("alluxio.master.worker.heartbeat.interval")
```
bf8086, line 1825: Assign this alias to `MASTER_LOST_WORKER_DETECTION_INTERVAL` for compatibility?

core/server/common/src/main/java/alluxio/master/StateLockManager.java
bf8086: Makes sense.

core/common/src/main/java/alluxio/conf/PropertyKey.java
yuzhu: done

bf8086 (approved):

yuzhu:

yuzhu:

yuzhu:

PR #12744, opened 2021-01-20
Fix table underdb pom

yuzhu:


yuzhu:
Was just trying it out. I guess the right thing to do is to put this copy in the sub projects? This pom does not actually generate a jar file. Will report an error because of it.

PR #12766, opened 2021-01-27
Add debug logging to StateLockManager

bradyoo:


bradyoo:
@bf8086 can you review this? It's really easy.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -131,7 +133,13 @@ public void mastersStartedCallback() {
    */
   public LockResource lockShared() throws InterruptedException {
     if (LOG.isDebugEnabled()) {
-      LOG.debug("Thread-{} entered lockShared().", ThreadUtils.getCurrentThreadIdentifier());
+      SAMPLING_LOG.debug(String.format("Thread-%s entered lockShared().",
```
bradyoo, line 136: This will add this string to a map in sampling logger.. Do you think that's okay?

bf8086: Woudl that be useful though? Sampling log will filter many instances of this log when there are  a lot of them in a short period of time. Since it is indexed by format string, multiple thread logging at the same time will result in only one entry being logged.

bradyoo: That's the intention. I am not that interested in a background thread that is going in and out of this lock repeatedly which can happen a lot in certain scenarios. There is a case for reentrant locks but I don't think that is so extreme to the point where we could hit the lock limit of 60K+.

bf8086: I am not sure if I understand the purpose of this change. Do you mind putting some explanation or example in the description?

bradyoo: The purpose of this change is to reduce the amount of logging in this debug log. The real bit of change is the one below this where it's reporting all the waiters and holders when the number of shared locks is too high.

bf8086: In that case, it makes more sense to just downlevel this log to trace instead of sampling it. Sampling log on locking information isn't very useful in that we actually care about each individual occurrence within a short period of time.

bradyoo: I'll just not change this.

bradyoo: Fixed.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -131,7 +133,13 @@ public void mastersStartedCallback() {
    */
   public LockResource lockShared() throws InterruptedException {
     if (LOG.isDebugEnabled()) {
-      LOG.debug("Thread-{} entered lockShared().", ThreadUtils.getCurrentThreadIdentifier());
+      LOG.debug(String.format("Thread-%s entered lockShared().",
```
bf8086, line 136: Remove `String.format`?

bradyoo: Oops. Done.

core/server/common/src/main/java/alluxio/master/StateLockManager.java
```diff
@@ -50,9 +50,11 @@
  */
 public class StateLockManager {
   private static final Logger LOG = LoggerFactory.getLogger(StateLockManager.class);
+  private static final SamplingLogger SAMPLING_LOG = new SamplingLogger(LOG, 30 * 1000);
```
bf8086, line 53: (nit) 30 * Constants.SECOND_MS

bf8086 (approved):

bradyoo:

bradyoo:

bradyoo:
alluxio-bot, cherry-pick to branch-2.3 please.

bradyoo:

PR #12767, opened 2021-01-27
[WIP] Adjust codecov

bradyoo:


bradyoo:
@ZacBlanco The code coverage number is so volatile in this white space change commit.

ZacBlanco:
Yeah I see. I'm not entirely sure why.

One possible explanation, is that the last code coverage generation was done multiple commits earlier. I think there were two new commits on master between when this PR was created and when code coverage last ran on master. Otherwise, it's possible that the tests don't deterministically cover the paths, but I think that's unlikely since previously when code coverage ran on a nightly basis it seemed pretty stable?

bradyoo:
I merged with master before creating this just to check but also note that unless the 2 commits that were made were gigantic, it's going to be difficult to make a 5% swing. When I made this commit previously it was saying this same commit was causing -8%. The swings are wild.

PR #12770, opened 2021-01-28
Fix backup suspend timeout

bf8086:
Currently delegated backup would timeout on suspend when the primary master takes long time to acquire exclusive lock during busy hours. This change adds a function in state lock manager to allow suspend of journal before every retry of lock, which should greatly improve the chance of keeping secondary master ready during the locking stage.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
+++ b/core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
@@ -56,8 +56,6 @@ import java.util.concurrent.TimeUnit;
 public class BackupWorkerRole extends AbstractBackupRole {
   private static final Logger LOG = LoggerFactory.getLogger(BackupWorkerRole.class);
 
-  // Constant timeout for when suspend request is not followed by a backup request.
-  private static final long BACKUP_ABORT_AFTER_SUSPEND_TIMEOUT_MS = 15000;
   // Constant timeout for journal transition before backup.
   private static final long BACKUP_ABORT_AFTER_TRANSITION_TIMEOUT_MS = 30000;
   // Minimum retry wait time between each connection attempt to leader.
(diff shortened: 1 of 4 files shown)


bf8086:
@yuzhu @calvinjia PTAL.

core/server/master/src/main/java/alluxio/master/backup/BackupWorkerRole.java
```diff
@@ -175,7 +177,7 @@ public BackupStatus getBackupStatus(BackupStatusPRequest statusPRequest) throws
     mBackupTimeoutTask = mTaskScheduler.schedule(() -> {
       LOG.info("Resuming journals as backup request hasn't been received.");
       enforceResumeJournals();
-    }, BACKUP_ABORT_AFTER_SUSPEND_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+    }, mBackupAbortSuspendTimeoutMs, TimeUnit.MILLISECONDS);
```
yuzhu, line 180: Is there a concurrency issue here? since `mBackupTimeoutTask` is a shared state, could a `resume journal`  happen after the next suspend? so when we backup the journal is actually not in a suspended state?  I think we might have to stop the resume journal before we suspend?

bf8086: It can be a concern if a user sets the suspend timeout (default to 60s with this change) to be larger than the retry interval (10min) + the lock wait (30s). This is a constraint without this fix and will require changes on backup worker and journal system to make it reentrance safe. I don't think we need to worry about it unless we are looking for a much shorter retry interval or larger suspend timeout.

yuzhu: discussed offline

yuzhu (approved):
Thanks for the fix, LGTM

bf8086:

bf8086:

yuzhu:

PR #12777, opened 2021-01-29
Add Initial journal replay before ZK connections in ufs journal

LuQQiu:
Currently, all masters register with ZK and listen on state change events (ie. LOST, SUSPENDED) immediately after starting. For large namespaces, there is an expensive journal replay step (> 10 minutes). If there is a state change during this period, the progress is reset. This can lead to an infinite loop, especially since GC pressure is higher during journal replay, leading to longer pauses.

This PR tries to let Alluxio Masters in Zookeeper + Ufs journal fault-tolerant mode replay its journal first and then register with the ZK. Those masters will be in ready mode and able to serve requests or listening to the leader after receiving the leadership information from ZK.

Note that this is the best effort and does not guarantee the master finishes its actual replay process before registering with Zookeeper. 
* If a master sleeps continuously for up to `shutdown wait time` when catching up the journal, we will consider the replay as done. 
* A max replay time configuration is given to avoid the master from continuously catching up during a heavy load environment (leading master writes journal quicker than expected so the standby master is busy catching up all the time, continuous sleep cannot be fulfilled). 
* A flag is added to enable this initial replay functionality.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1880,7 +1880,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_JOURNAL_CATCHUP_PROTECT_ENABLED =
       new Builder(Name.MASTER_JOURNAL_CATCHUP_PROTECT_ENABLED)
-          .setDefaultValue(false)
+          .setDefaultValue(true)
           .setDescription("(Experimental) make sure the journal catchup finish "
               + "before joining the quorum in fault tolerant mode when starting the master process "
               + "and before the current master becoming the leader."
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1878,14 +1878,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDescription("Maximum concurrency level for the lock pool")
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_JOURNAL_INITIAL_REPLAY_ENABLED =
-      new Builder(Name.MASTER_JOURNAL_INITIAL_REPLAY_ENABLED)
+  public static final PropertyKey MASTER_JOURNAL_CATCHUP_PROTECT_ENABLED =
+      new Builder(Name.MASTER_JOURNAL_CATCHUP_PROTECT_ENABLED)
           .setDefaultValue(false)
-          .setDescription("(Experimental) Enable the initial journal replay "
-              + "before joining the quorum in fault tolerant mode "
-              + "when starting the master process. This is added to prevent "
-              + "frequently leadership transition during heavy journal replay stage. "
-              + "Initial replay is only implemented in ufs journal with Zookeeper.")
+          .setDescription("(Experimental) make sure the journal catchup finish "
+              + "before joining the quorum in fault tolerant mode when starting the master process "
+              + "and before the current master becoming the leader."
+              + "This is added to prevent frequently leadership transition "
+              + "during heavy journal catchup stage. "
+              + "Catchup is only implemented in ufs journal with Zookeeper.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
@@ -5190,8 +5191,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.lock.pool.high.watermark";
     public static final String MASTER_LOCK_POOL_CONCURRENCY_LEVEL =
         "alluxio.master.lock.pool.concurrency.level";
-    public static final String MASTER_JOURNAL_INITIAL_REPLAY_ENABLED =
-        "alluxio.master.journal.initial.replay.enabled";
+    public static final String MASTER_JOURNAL_CATCHUP_PROTECT_ENABLED =
+        "alluxio.master.journal.catchup.protect.enabled";
     public static final String MASTER_JOURNAL_FLUSH_BATCH_TIME_MS =
         "alluxio.master.journal.flush.batch.time";
     public static final String MASTER_JOURNAL_FLUSH_TIMEOUT_MS =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1884,12 +1884,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDescription("(Experimental) Enable the initial journal replay "
               + "before joining the quorum in fault tolerant mode "
               + "when starting the master process. This is added to prevent "
-              + "frequently leadership transition during heavy journal replay stage")
+              + "frequently leadership transition during heavy journal replay stage. "
+              + "Initial replay is only implemented in ufs journal with Zookeeper.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME =
-      new Builder(Name.MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME)
+  public static final PropertyKey MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME =
+      new Builder(Name.MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME)
           .setDefaultValue("10min")
           .setDescription("The maximum time to wait for ufs journal catching up "
               + "before listening to Zookeeper state change. This is added to prevent "
@@ -5190,9 +5191,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     public static final String MASTER_LOCK_POOL_CONCURRENCY_LEVEL =
         "alluxio.master.lock.pool.concurrency.level";
     public static final String MASTER_JOURNAL_INITIAL_REPLAY_ENABLED =
-        "alluxio.master.initial.replay.enabled";
-    public static final String MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME =
-        "alluxio.master.max.initial.replay.time";
+        "alluxio.master.journal.initial.replay.enabled";
     public static final String MASTER_JOURNAL_FLUSH_BATCH_TIME_MS =
         "alluxio.master.journal.flush.batch.time";
     public static final String MASTER_JOURNAL_FLUSH_TIMEOUT_MS =
@@ -5349,6 +5348,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.ufs.active.sync.poll.batch.size";
     public static final String MASTER_UFS_BLOCK_LOCATION_CACHE_CAPACITY =
         "alluxio.master.ufs.block.location.cache.capacity";
+    public static final String MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME =
+        "alluxio.master.ufs.journal.max.initial.replay.time";
     public static final String MASTER_UFS_MANAGED_BLOCKING_ENABLED =
         "alluxio.master.ufs.managed.blocking.enabled";
     public static final String MASTER_UFS_PATH_CACHE_CAPACITY =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1881,8 +1881,9 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey MASTER_JOURNAL_INITIAL_REPLAY_ENABLED =
       new Builder(Name.MASTER_JOURNAL_INITIAL_REPLAY_ENABLED)
           .setDefaultValue(false)
-          .setDescription("(Experimental) Enable the initial journal replay before joining the quorum "
-              + "in fault tolerant mode when starting the master process. This is added to prevent "
+          .setDescription("(Experimental) Enable the initial journal replay "
+              + "before joining the quorum in fault tolerant mode "
+              + "when starting the master process. This is added to prevent "
               + "frequently leadership transition during heavy journal replay stage")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1878,6 +1878,24 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDescription("Maximum concurrency level for the lock pool")
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_JOURNAL_INITIAL_REPLAY_ENABLED =
+      new Builder(Name.MASTER_JOURNAL_INITIAL_REPLAY_ENABLED)
+          .setDefaultValue(false)
+          .setDescription("(Experimental) Enable the initial journal replay before joining the quorum "
+              + "in fault tolerant mode when starting the master process. This is added to prevent "
+              + "frequently leadership transition during heavy journal replay stage")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME =
+      new Builder(Name.MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME)
+          .setDefaultValue("10min")
+          .setDescription("The maximum time to wait for ufs journal catching up "
+              + "before listening to Zookeeper state change. This is added to prevent "
+              + "frequently leadership transition during heavy journal replay stage.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_JOURNAL_FLUSH_BATCH_TIME_MS =
       new Builder(Name.MASTER_JOURNAL_FLUSH_BATCH_TIME_MS)
           .setAlias("alluxio.master.journal.flush.batch.time.ms")
@@ -5170,6 +5188,10 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.lock.pool.high.watermark";
     public static final String MASTER_LOCK_POOL_CONCURRENCY_LEVEL =
         "alluxio.master.lock.pool.concurrency.level";
+    public static final String MASTER_JOURNAL_INITIAL_REPLAY_ENABLED =
+        "alluxio.master.initial.replay.enabled";
+    public static final String MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME =
+        "alluxio.master.max.initial.replay.time";
     public static final String MASTER_JOURNAL_FLUSH_BATCH_TIME_MS =
         "alluxio.master.journal.flush.batch.time";
     public static final String MASTER_JOURNAL_FLUSH_TIMEOUT_MS =
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1890,8 +1890,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  public static final PropertyKey MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME =
-      new Builder(Name.MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME)
+  public static final PropertyKey MASTER_UFS_JOURNAL_MAX_CATCHUP_TIME =
+      new Builder(Name.MASTER_UFS_JOURNAL_MAX_CATCHUP_TIME)
           .setDefaultValue("10min")
           .setDescription("The maximum time to wait for ufs journal catching up "
               + "before listening to Zookeeper state change. This is added to prevent "
@@ -5349,8 +5349,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.ufs.active.sync.poll.batch.size";
     public static final String MASTER_UFS_BLOCK_LOCATION_CACHE_CAPACITY =
         "alluxio.master.ufs.block.location.cache.capacity";
-    public static final String MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME =
-        "alluxio.master.ufs.journal.max.initial.replay.time";
+    public static final String MASTER_UFS_JOURNAL_MAX_CATCHUP_TIME =
+        "alluxio.master.ufs.journal.max.catchup.time";
     public static final String MASTER_UFS_MANAGED_BLOCKING_ENABLED =
         "alluxio.master.ufs.managed.blocking.enabled";
     public static final String MASTER_UFS_PATH_CACHE_CAPACITY =
--- a/core/common/src/main/java/alluxio/metrics/MetricKey.java
+++ b/core/common/src/main/java/alluxio/metrics/MetricKey.java
@@ -418,11 +418,6 @@ public final class MetricKey implements Comparable<MetricKey> {
           .setMetricType(MetricType.GAUGE)
           .build();
   // Journal metrics
-  public static final MetricKey MASTER_UFS_JOURNAL_INITIAL_REPLAY_TIME_MS =
-      new Builder(Name.MASTER_UFS_JOURNAL_INITIAL_REPLAY_TIME_MS)
-          .setDescription("The process time of the ufs journal initial replay")
-          .setMetricType(MetricType.GAUGE)
-          .build();
   public static final MetricKey MASTER_JOURNAL_FLUSH_FAILURE =
       new Builder(Name.MASTER_JOURNAL_FLUSH_FAILURE)
           .setDescription("Total number of failed journal flush")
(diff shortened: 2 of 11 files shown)


LuQQiu:
Error: 7.548 [ERROR] alluxio.client.rest.AlluxioMasterRestApiTest.getWorkers(alluxio.client.rest.AlluxioMasterRestApiTest)
Error: 7.548 [ERROR]   Run 1: AlluxioMasterRestApiTest.getWorkers:220->getInfo:83 Request failed: {"statusCode":"INTERNAL","message":"class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')"}
Error: 7.549 [ERROR]   Run 2: AlluxioMasterRestApiTest.getWorkers:220->getInfo:83 Request failed: {"statusCode":"INTERNAL","message":"class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')"}

LuQQiu:
@bf8086 @gpang  The changes are still under testing and need to be discussed, so i didn't add the tests yet.
PTAL the changes of adding initial journal replay before ZK connections in ufs journal, appreciate!

LuQQiu:
@bf8086 @calvinjia PTAL at the PR, thanks!
One thing that may need to be discussed is how we deal with the checkpoint.
Do we want to wait for the triggered checkpoint to be finished before considering the initial replay as DONE?

In the current impl, we only consider initial replay as DONE if continuous sleep up to SHUTDOWN_WAIT_TIME is detected or `MAX_INITIAL_REPLAY_TIME` reached.
* if checkpoint finished within `MAX_INITIAL_REPLAY_WAIT_TIME`, great!
* if checkpoint is in progress when `MAX_INITIAL_REPLAY_WAIT_TIME` is reached, the checkpoint may be interrupted if the master gain leadership.
What your opinion on this approach?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -137,6 +145,32 @@ public CatchupFuture catchup(Map<String, Long> journalSequenceNumbers) throws IO
...
+          UfsJournalCheckpointThread.ReplayState replayState = journal.getReplayState();
+          if (replayState != UfsJournalCheckpointThread.ReplayState.REPLAY_DONE) {
+            // keep flushing entries
+            return false;
+          }
+        }
+        return true;
+      }, WaitForOptions.defaults().setTimeoutMs(
+          (int) ServerConfiguration.getMs(PropertyKey.MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME))
+          .setInterval(30 * Constants.SECOND_MS));
```
bf8086, line 163: Should this be more frequent for faster startup?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -224,6 +246,10 @@ private void runInternal() {
 
       // Sleep for a while if no entry is found.
       if (entry == null) {
+        if (continuousSleepNum >= mReplayDoneSleepThreshold) {
```
bf8086, line 249: Why do we need to wait for multiple sleeps?

bf8086 (changes requested):
Thanks for the improvement! I left a couple comments.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
LuQQiu: Yeah, i don't think it is any harmful if do it at 5-second interval since our shutdown quiet wait time is 5 second.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
LuQQiu: I cannot eliminate the chance that the alluxio cluster has some distance from its journal system (HDFS) and maybe at some period of time the journal transfer is slow. 
Like send ... send ... send ... send ..  (1 sec gap)... send 
waiting for a longer sleep time can better make us confident that the journal replay is actually done.

bf8086: Wouldn't `UfsJournalCheckpointThread` only sleep when it already read the last entry that have been written recently? Is that not enough indicator that it has caught up all the entries that has been written by the primary master at the time of startup?

LuQQiu: Oh you should be correct, i don't need to sleep for that long time, one sleep is enough,

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -224,6 +234,9 @@ private void runInternal() {
 
       // Sleep for a while if no entry is found.
       if (entry == null) {
+        if (mReplayState != ReplayState.REPLAY_DONE) {
```
bf8086, line 237: (nit) Is this condition needed?

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
```diff
@@ -170,6 +170,23 @@
...
+   * in secondary mode. This is best-effort, because even if it did not finish
+   * replaying the journal, the rest of the system will still complete
+   * the replay (journal catch up) in a different phase.
+   *
+   * This can be implemented by a journal type to optimize the journal replay, and avoid getting
+   * interrupted with primary state changes during journal replay.
+   *
+   * @return true if the journal replay is considered caught up, false otherwise
+   */
+  default boolean waitForInitialReplay() {
```
bf8086, line 185: Is the return value used anywhere?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -137,6 +145,32 @@ public CatchupFuture catchup(Map<String, Long> journalSequenceNumbers) throws IO
...
+      }, WaitForOptions.defaults().setTimeoutMs(
+          (int) ServerConfiguration.getMs(PropertyKey.MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME))
+          .setInterval(5 * Constants.SECOND_MS));
+    } catch (InterruptedException | TimeoutException e) {
+      LOG.info("Journal initial replay is interrupted or timeout", e);
+      mInitialCatchUpTimeMs = System.currentTimeMillis() - start;
+      return false;
+    }
+    mInitialCatchUpTimeMs = System.currentTimeMillis() - start;
+    LOG.info("Finish master process ufs journal initial replay in {} ms", mInitialCatchUpTimeMs);
```
bf8086, line 170: (nit) Finished

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
```diff
@@ -73,15 +73,22 @@ protected FaultTolerantAlluxioMasterProcess(JournalSystem journalSystem,
   public void start() throws Exception {
     mRunning = true;
     mJournalSystem.start();
+
+    startMasters(false);
+    LOG.info("Secondary started");
+
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_INITIAL_REPLAY_ENABLED)) {
+      mJournalSystem.waitForInitialReplay();
+      LOG.info("Journal initial replay finished");
```
bf8086, line 82: (nit) This log seems redundant to line 170 in `UfsJournalSystem`.

bf8086 (changes requested):
Mostly LGTM. Would it be possible to add some tests for the new behavior?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
LuQQiu: @bf8086 I tested if i directly set replay == DONE when entry == null, the journal replay will be considered as finished before the master reading any journals. 

The logs look like
```
2021-02-02 03:32:43,590 INFO  ZkMasterInquireClient - All masters: [ReplayUfsJournalZK4-masters-2:19998]
2021-02-02 03:34:05,495 INFO  JournalUtils - Reading journal entries
2021-02-02 03:34:05,497 INFO  JournalUtils - Reading journal entries
2021-02-02 03:34:05,497 INFO  JournalUtils - Reading journal entries
2021-02-02 03:34:05,497 INFO  JournalUtils - Reading journal entries
2021-02-02 03:34:06,277 INFO  UfsJournalSystem - Finish master process ufs journal initial replay in 90528 ms
2021-02-02 03:34:06,277 INFO  FaultTolerantAlluxioMasterProcess - Journal initial replay finished
2021-02-02 03:34:07,211 INFO  UfsJournalReader - Reading journal file hdfs://ReplayUfsJournalZK4-masters-1:9000/alluxio_journal/FileSystemMaster/v1/logs/0x858b9b-0x861492.
2021-02-02 03:34:07,483 INFO  UfsJournalReader - Reading journal file hdfs://ReplayUfsJournalZK4-masters-1:9000/alluxio_journal/FileSystemMaster/v1/logs/0x861492-0x869c2f.
2021-02-02 03:34:07,939 INFO  UfsJournalReader - Reading journal file hdfs://ReplayUfsJournalZK4-masters-1:9000/alluxio_journal/FileSystemMaster/v1/logs/0x869c2f-0x872793.
```

LuQQiu: The UfsJournalCheckpointThread could sleep if it hasn't read the HDFS journal fast enough.

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
LuQQiu: Removed, thanks

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
LuQQiu: Done, thanks

core/server/common/src/main/java/alluxio/master/journal/JournalSystem.java
LuQQiu: Change to void, thanks

LuQQiu:
@bf8086 Add the tests and the comment for replay done behavior, PTAL, thanks!

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
bf8086: The logs look like it got entry=null when it finished reading the checkpoint. So a more accurate indicator for journal read completion is the return value of `mJournalReader.advance()`. It would be something like:
```
      try {
        switch (mJournalReader.advance()) {
          case CHECKPOINT:
            ...
          case ENTRY:
            ...
          case DONE:
            mReplayState = ReplayState.REPLAY_DONE;
            break;
          default:
            break;
        }
```

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
```diff
@@ -73,15 +73,21 @@ protected FaultTolerantAlluxioMasterProcess(JournalSystem journalSystem,
   public void start() throws Exception {
     mRunning = true;
     mJournalSystem.start();
+
+    startMasters(false);
+    LOG.info("Secondary started");
+
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_INITIAL_REPLAY_ENABLED)) {
+      mJournalSystem.waitForInitialReplay();
```
gpang, line 81: I was wondering if this initial replay will only work for the process start? I originally thought it could be ok to do whenever a master process became secondary, basically doing this wait within the while loop below. What are the possible issues with that?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -418,6 +418,11 @@ public MetricKey build() {
           .setMetricType(MetricType.GAUGE)
           .build();
   // Journal metrics
+  public static final MetricKey MASTER_JOURNAL_INITIAL_REPLAY_TIME_MS =
+      new Builder(Name.MASTER_JOURNAL_INITIAL_REPLAY_TIME_MS)
+          .setDescription("The process time of the journal initial repaly")
```
gpang, line 423: repaly -> replay

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1878,6 +1878,25 @@ public String toString() {
...
+              + "before joining the quorum in fault tolerant mode "
+              + "when starting the master process. This is added to prevent "
+              + "frequently leadership transition during heavy journal replay stage")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME =
+      new Builder(Name.MASTER_JOURNAL_MAX_INITIAL_REPLAY_TIME)
+          .setDefaultValue("10min")
+          .setDescription("The maximum time to wait for ufs journal catching up "
```
gpang, line 1894: I know only the UFS journal implements the initial replay, but could this also apply to raft later? Basically, this description mentions ufs journal, but the property name is not ufs specific.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -184,6 +194,7 @@ private void runInternal() {
             LOG.debug("{}: Restoring from checkpoint", mMaster.getName());
             mMaster.restoreFromCheckpoint(mJournalReader.getCheckpoint());
             LOG.debug("{}: Finished restoring from checkpoint", mMaster.getName());
+            //continuousSleepNum = 0;
```
gpang, line 197: Remove this?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -137,6 +145,31 @@ public CatchupFuture catchup(Map<String, Long> journalSequenceNumbers) throws IO
...
 
+  @Override
+  public void waitForInitialReplay() {
+    long start = System.currentTimeMillis();
+    try {
+      CommonUtils.waitFor("journal replay to finish catching up", () -> {
+        for (UfsJournal journal : mJournals.values()) {
+          UfsJournalCheckpointThread.ReplayState replayState = journal.getReplayState();
+          if (replayState != UfsJournalCheckpointThread.ReplayState.REPLAY_DONE) {
+            // keep flushing entries
```
gpang, line 156: I don't think this is "flushing" the journal, but replaying the journal, right?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -205,6 +216,7 @@ private void runInternal() {
             }
             break;
           default:
+            mReplayState = ReplayState.REPLAY_DONE;
```
gpang, line 219: What if the current log file is done (file1), but the next one (file2) was not complete yet. Would the replay be considered done after file1, or after file2 is completed and read by this master?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -228,6 +240,7 @@ private void runInternal() {
         maybeCheckpoint();
         if (mShutdownInitiated) {
           if (quietPeriodWaited || !mWaitQuietPeriod) {
+            mReplayState = ReplayState.REPLAY_DONE;
```
gpang, line 243: If the tailer is shutdown, then this means it is already being signaled to become primary from secondary? Is that transition when we want to set the replay state to done?

gpang (changes requested):
@LuQQiu Thanks! I left some questions/comments.

core/common/src/main/java/alluxio/metrics/MetricKey.java
LuQQiu: DONE, thanks!

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Yeah, change to MASTER_UFS_JOURNAL.
It's interesting that we don't have MASTER_UFS_JOURNAL specific property keys yet.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
LuQQiu: Removed the comment

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
LuQQiu: UfsJournalCheckpointThread keeps catching up the "complete" journal files and ignores "incomplete" files. so only file 1 will be read and replayed. After that, the replay will be considered done and the checkpoint thread may start writing checkpoint. 
We will not catch up all journals, but catch up all complete journals.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
LuQQiu: I don't think this line actually do anything. I add it to be safe.
When initial replaying, the master will not be able to gainPrimacy() or losePrimacy() to shutdown the checkpoint thread. If the AlluxioMaster itself is shutting down, replay is done or not doesn't make a difference.

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
gpang: Yeah, that makes sense, and it should be fine to only have 1 incomplete file to replay afterwards.

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
gpang: Was it already decided that we will only do the replay for the very first replay, and not other replays?

gpang (commented):
@LuQQiu Thanks! Left a minor question.

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
LuQQiu: Sorry for the late reply. I am still getting more information on whether I am able to add the initial replay to the while loop safe and whether the change can bring more benefits. There hasn't been any decision but I first move it out to make sure the behavior is understandable. 

In the current implementation, the initial replay is put before we connect to Zookeeper and join leader election to make sure the initial replay can finish and not affect the other running Alluxio masters.  In the previous implementation which we put the `waitForInitReplay` only in the while loop, the following behavior may happen:
```
For example, 
* Master A is the leading master
* Master B join the leader election
* Master B inits from replay
* During the Master B initial replay, Master B is selected as the leading master
* If Master B replaying for too long, it cannot gain primacy and able to serve requests fast enough, then no Alluxio Master will be able to serve client requests in a long period.
```
As for whether we want to add `waitForCatchup` in the while loop, it may not make a huge difference if all masters already finish the initial replay. Step 3 is not expected to take a long time.
```
1. mJournalSystem.waitForInitialReplay();
2. mLeaderSelector.start(getRpcAddress());
 while (!Thread.interrupted()) {
3. mJournalSystem.waitForCatchup();   // I think this can be added may not make a difference
      mLeaderSelector.waitForState(State.PRIMARY);
      if (!mRunning) {
        break;
      }
      if (gainPrimacy()) {
        if (!mRunning) {
          break;
        }
        losePrimacy();
      }
```

LuQQiu: @gpang PTAL, thanks!

gpang: Yes, if the `waitForCatchup` is in the while loop, the behavior you described would happen. So, the worst case is that for some period of time, there will be no master service requests. However, is that different from any alternative? For example, the leader will only serve after it finishes `gainPrimacy()`.

Isn't the `waitForState()` method the only place where a master will consult the zk state? So, as long as we `waitForCatchup()` before `waitForState()`, we will always finish replaying the journal before checking the zk state.

LuQQiu: Yeah, there should be no difference between catchup and then gainPrimacy or gainPrimacy different when considering no master service period.
Adding waitForCatchup can also make sure the master catches up during frequent leadership change.
Overall, adding waitForCatchup in the while loop sounds to be slightly better than the current implementation. 
@bf8086 What's your opinion about adding the waitForCatchup in the Primary/Seconday while loop
```
 while (!Thread.interrupted()) {
     mJournalSystem.waitForCatchup();  <--- adding this
      mLeaderSelector.waitForState(State.PRIMARY);
      if (!mRunning) {
        break;
      }
      if (gainPrimacy()) {
        if (!mRunning) {
          break;
        }
        losePrimacy();
      }
```

bf8086: Agree it's helpful to add `waitForCatchup` in the while loop for frequent leadership changes. Can you maybe try some failover tests to make sure it does not introduce undesired behaviors?

gpang: Yeah, my intuition feels like it should be correct in the while-loop, but I'm just not completely sure. Running some tests would increase our confidence.

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
```diff
@@ -73,16 +73,25 @@ protected FaultTolerantAlluxioMasterProcess(JournalSystem journalSystem,
   public void start() throws Exception {
     mRunning = true;
     mJournalSystem.start();
+
+    startMasters(false);
+    LOG.info("Secondary started");
+
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_CATCHUP_PROTECT_ENABLED)) {
+      mJournalSystem.waitForCatchup();
```
bf8086, line 81: Do we still need it here now that we do it anyway within the while loop?

LuQQiu: Yeah it's still needed. We want the initial replay (which could be really long in production env) to finish before joining the leader election.
If this is removed, there could be a case that
1. Master A is the previous leading master
2. Master B joins the leader election
3. Master B starts the journal initial replay, during the initial replay, Master B is selected as the primary master
4. All the clients need to wait for master B to finish its initial replay and start serving requests even though other masters are available.

In the beginning, there will be two `wait for catch up` in a row, the second one is usually a noop and finish in one second. This doesn't seem to introduce any issues

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -82,6 +82,15 @@
   /** A supplier of journal sinks for this journal. */
   private final Supplier<Set<JournalSink>> mJournalSinks;
 
+  /**
+   * The state of the journal catchup.
+   */
+  public enum CatchupState {
+    CATCHUP_NOT_STARTED, CATCHUP_IN_PROGRESS, CATCHUP_DONE;
```
calvinjia, line 89: The `CATCHUP` prefix seems a bit redundant?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
```diff
@@ -137,6 +147,35 @@ public CatchupFuture catchup(Map<String, Long> journalSequenceNumbers) throws IO
...
+            return false;
+          }
+        }
+        return true;
+      }, WaitForOptions.defaults().setTimeoutMs(
+          (int) ServerConfiguration.getMs(PropertyKey.MASTER_UFS_JOURNAL_MAX_INITIAL_REPLAY_TIME))
+          .setInterval(Constants.SECOND_MS));
+    } catch (InterruptedException | TimeoutException e) {
+      LOG.info("Journal catchup is interrupted or timeout", e);
+      if (mInitialCatchupTimeMs == -1) {
```
calvinjia, line 168: Why do we need to check that we have not set the catchup time yet? Shouldn't the value reflect the latest catchup time?

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
calvinjia, line 81: If this fails, do we still want to proceed?

calvinjia (changes requested):

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalSystem.java
LuQQiu: There are several catchups in the current implementation:
```
1. Initial catchup
2.start primary selector to join leader election
while {
    3. catchup when step back to become the secondary master and before checking if it's the primary
    mLeaderSelector.waitForState(State.PRIMARY);
    gainPrimacy()
   losePrimacy();
}
```
I have one metric MASTER_UFS_JOURNAL_CATCHUP_TIMER to record the catchup time for all the catchup.
and have a MASTER_UFS_JOURNAL_INITIAL_REPLAY_TIME_MS to record the first initial catchup time.

The initial replay time should only be recorded once and not overridden by future catchups

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
LuQQiu: Yeah, I have a `alluxio.master.ufs.journal.max.catchup.time` for users to set the maximum catchup time they want. If the master hasn't finish catching up, timeout in waiting for catchup, it will
1) continue catching up if it's not selected as the primary
2) continue catching up in the gainPrimacy() if selected as the primary.
So it's actually fine for catchup to fail.

The `waitForCatchup` is the best effort for the master to catch up, but doesn't guarantee catchup can be finished within timeout. Even if it doesn't catchup, in both cases described above, it will not introduce issues.

LuQQiu:
@gpang @bf8086 @calvinjia Thanks for the comments and reviews! I replied the comments, ran some ufs journal related tests with catchup enabled (nothing surprise happens), and modified the unit tests. PTAL, appreciate!

calvinjia (approved):

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
bf8086: That makes sense. Let's keep both then.

gpang: Sounds good.

Can we add a short comment here describing that? Maybe something like `// Perform initial catchup before joining leader election, to avoid potential delay if this process is selected as leader`

gpang (approved):
@LuQQiu Thanks! I just had a minor comment, and then looks good!

LGTM

bf8086 (approved):
LGTM.

LuQQiu:
Thanks for the careful reviews! @gpang @calvinjia @bf8086

LuQQiu:

PR #12783, opened 2021-02-01
Convert the configuration files to mount options when Mount

### LINKED ISSUE #12599: Master should dispatch the configuration files to all worker when Mount command and worker register

Now, we can mount a ufs to alluxio easily by execute a simple command line, for example

```bash
./bin/alluxio fs mount \
  --option alluxio.underfs.hdfs.configuration=<DIR>/ozone-site.xml:<DIR>/core-site.xml \
  /ozone o3fs://<OZONE_BUCKET>.<OZONE_VOLUME>/
```

But we have to dispatch the config files to all of the workers manually, that is Inefficient and error-prone.

Alluxio master can maintain mount table now, so it can maintain the configuration files related each mount point by the way, workers should ask for mount point configuration files update when heartbeat if needed.

Modify the worker module code, when worker need access ufs and found configuration file does not exist, just ask master to request the lost configuration file. During each heartbeat, worker can send its configuration files fingerprint to master to check if the files updated from master.

**Urgency**

LOW

**Additional context**
Improve user interface feelin.

maobaolong:
Fixes #12599

If the `alluxio.underfs.hdfs.configuration` mount option start with "kv@", then extract the key value to mount options.

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
@@ -24,6 +24,8 @@ import com.google.common.collect.Maps;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
 
 import java.io.IOException;
 import java.util.Map;
@@ -62,6 +64,14 @@ public final class MountCommand extends AbstractFileSystemCommand {
           .valueSeparator('=')
           .desc("options associated with this mount point")
           .build();
+  private static final Option CONFIG_FILE_OPTION =
+      Option.builder()
+          .longOpt("cfgfile")
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .desc("config file path associated with this mount point")
+          .build();
 
   /**
    * @param fsContext the filesystem of Alluxio
@@ -78,7 +88,7 @@ public final class MountCommand extends AbstractFileSystemCommand {
   @Override
   public Options getOptions() {
     return new Options().addOption(READONLY_OPTION).addOption(SHARED_OPTION)
-        .addOption(OPTION_OPTION);
+        .addOption(OPTION_OPTION).addOption(CONFIG_FILE_OPTION);
   }
 
   @Override
@@ -99,6 +109,16 @@ public final class MountCommand extends AbstractFileSystemCommand {
     if (cl.hasOption(SHARED_OPTION.getLongOpt())) {
       optionsBuilder.setShared(true);
     }
+    if (cl.hasOption(CONFIG_FILE_OPTION.getLongOpt())) {
+      String[] cfgFiles = cl.getOptionValues(CONFIG_FILE_OPTION.getLongOpt());
+      Configuration hdfsConf = new Configuration(false);
+      for (String path : cfgFiles) {
+        if (!path.isEmpty()) {
+          hdfsConf.addResource(new Path(path));
+        }
+      }
+      optionsBuilder.putAllProperties(hdfsConf.getPropsWithPrefix(""));
+    }
     if (cl.hasOption(OPTION_OPTION.getLongOpt())) {
       Properties properties = cl.getOptionProperties(OPTION_OPTION.getLongOpt());
       optionsBuilder.putAllProperties(Maps.fromProperties(properties));
@@ -110,7 +130,8 @@ public final class MountCommand extends AbstractFileSystemCommand {
 
   @Override
   public String getUsage() {
-    return "mount [--readonly] [--shared] [--option <key=val>] <alluxioPath> <ufsURI>";
+    return "mount [--readonly] [--shared] [--option <key=val>] [--cfgfile <cfgFilePath>] "
+        + "<alluxioPath> <ufsURI>";
   }
 
   @Override
--- a/shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
@@ -64,13 +64,13 @@ public final class MountCommand extends AbstractFileSystemCommand {
           .valueSeparator('=')
           .desc("options associated with this mount point")
           .build();
-  private static final Option CONFIG_FILE_OPTION =
+  private static final Option HDFS_CONF_OPTION =
       Option.builder()
-          .longOpt("cfgfile")
+          .longOpt("hdfsconf")
           .required(false)
           .hasArg(true)
           .numberOfArgs(1)
-          .desc("config file path associated with this mount point")
+          .desc("path to its config files for an HDFS mount point")
           .build();
 
   /**
@@ -88,7 +88,7 @@ public final class MountCommand extends AbstractFileSystemCommand {
   @Override
   public Options getOptions() {
     return new Options().addOption(READONLY_OPTION).addOption(SHARED_OPTION)
-        .addOption(OPTION_OPTION).addOption(CONFIG_FILE_OPTION);
+        .addOption(OPTION_OPTION).addOption(HDFS_CONF_OPTION);
   }
 
   @Override
@@ -109,8 +109,8 @@ public final class MountCommand extends AbstractFileSystemCommand {
     if (cl.hasOption(SHARED_OPTION.getLongOpt())) {
       optionsBuilder.setShared(true);
     }
-    if (cl.hasOption(CONFIG_FILE_OPTION.getLongOpt())) {
-      String[] cfgFiles = cl.getOptionValues(CONFIG_FILE_OPTION.getLongOpt());
+    if (cl.hasOption(HDFS_CONF_OPTION.getLongOpt())) {
+      String[] cfgFiles = cl.getOptionValues(HDFS_CONF_OPTION.getLongOpt());
       Configuration hdfsConf = new Configuration(false);
       for (String path : cfgFiles) {
         if (!path.isEmpty()) {
@@ -130,7 +130,7 @@ public final class MountCommand extends AbstractFileSystemCommand {
 
   @Override
   public String getUsage() {
-    return "mount [--readonly] [--shared] [--option <key=val>] [--cfgfile <cfgFilePath>] "
+    return "mount [--readonly] [--shared] [--option <key=val>] [--hdfsconf <hdfsConfigFilePath>] "
         + "<alluxioPath> <ufsURI>";
   }
 
--- a/pom.xml
+++ b/pom.xml
@@ -470,6 +470,11 @@
       </dependency>
       <dependency>
         <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-common</artifactId>
+        <version>${hadoop.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.hadoop</groupId>
         <artifactId>hadoop-minicluster</artifactId>
         <version>${hadoop.version}</version>
         <exclusions>
--- a/shell/pom.xml
+++ b/shell/pom.xml
@@ -67,6 +67,16 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-compress</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <exclusions>
+        <exclusion>
+          <artifactId>jsr311-api</artifactId>
+          <groupId>javax.ws.rs</groupId>
+        </exclusion>
+      </exclusions>
+    </dependency>
 
     <!-- Internal dependencies -->
     <dependency>


maobaolong:
It seems the failed test MultiProcessClusterTest have no related with this PR.

core/common/pom.xml
```diff
@@ -123,6 +123,11 @@
       <artifactId>curator-test</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
gpang, line 126: Is this supposed to be non-test dependency? Then it should be moved to a different section.

Also, @calvinjia do we want to add a hadoop dep?

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -114,6 +118,17 @@ public UnderFileSystemConfiguration setShared(boolean shared) {
   public UnderFileSystemConfiguration createMountSpecificConf(Map<String, String> mountConf) {
     UnderFileSystemConfiguration ufsConf = new UnderFileSystemConfiguration(mProperties.copy());
     ufsConf.mProperties.merge(mountConf, Source.MOUNT_OPTION);
+    String ufsConfigValue = ufsConf.get(PropertyKey.UNDERFS_HDFS_CONFIGURATION);
+    if (ufsConfigValue.startsWith("kv@")) {
```
gpang, line 122: can we make `"kv@"` a constant?

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -114,6 +118,17 @@ public UnderFileSystemConfiguration setShared(boolean shared) {
...
+      Configuration hdfsConf = new Configuration();
+      for (String path : ufsConfigValue.substring(3).split(":")) {
```
gpang, line 124: If `"kv@"` was a constant, it can be used for the length.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -114,6 +118,17 @@ public UnderFileSystemConfiguration setShared(boolean shared) {
   public UnderFileSystemConfiguration createMountSpecificConf(Map<String, String> mountConf) {
     UnderFileSystemConfiguration ufsConf = new UnderFileSystemConfiguration(mProperties.copy());
     ufsConf.mProperties.merge(mountConf, Source.MOUNT_OPTION);
+    String ufsConfigValue = ufsConf.get(PropertyKey.UNDERFS_HDFS_CONFIGURATION);
```
gpang, line 121: Can you update the property key for `UNDERFS_HDFS_CONFIGURATION` to describe this new feature?

gpang (changes requested):
@maobaolong Thanks! I left a few comments.

core/common/pom.xml
maobaolong: Done, move to the non-test section.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
maobaolong: @gpang Yeah, I defined it in the PropertyKey.Name

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
maobaolong: Sure, use length() to get the prefix length.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
maobaolong: Done.

maobaolong:
@gpang Thank for your review and comments, I pushed a new commit to address the comments. PTAL.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -667,7 +667,9 @@ public String toString() {
               "${%s}/core-site.xml:${%s}/hdfs-site.xml", Name.CONF_DIR, Name.CONF_DIR))
           .setDescription("Location of the HDFS configuration file to overwrite "
               + "the default HDFS client configuration. Note that, these files must be available"
-              + "on every node.")
+              + "on every node, but if there are a "
+              + Name.UNDERFS_HDFS_CONFIGURATION_PREFIX + " prefix, these files just need to be "
```
gpang, line 671: This should explain that the hdfs configuration will be converted into mount properties, so that is why it only needs to be on the masters.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -41,6 +46,7 @@
 @NotThreadSafe
 @PublicApi
 public final class UnderFileSystemConfiguration extends InstancedConfiguration {
+  private static final String UNDERFS_HDFS_CONFIGURATION_PREFIX = "kv@";
```
gpang, line 49: This should use the other constant?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -5005,6 +5007,7 @@ private static String javadocLink(String fullyQualifiedClassname) {
         "alluxio.underfs.gcs.directory.suffix";
     public static final String UNDERFS_GCS_OWNER_ID_TO_USERNAME_MAPPING =
         "alluxio.underfs.gcs.owner.id.to.username.mapping";
+    public static final String UNDERFS_HDFS_CONFIGURATION_PREFIX = "kv@";
```
gpang, line 5010: This is not a PropertyKey name, right? Maybe this should go to Constants.java.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -114,6 +120,19 @@ public UnderFileSystemConfiguration setShared(boolean shared) {
...
+    String ufsConfigValue = ufsConf.get(PropertyKey.UNDERFS_HDFS_CONFIGURATION);
+    if (ufsConfigValue.startsWith(Name.UNDERFS_HDFS_CONFIGURATION_PREFIX)) {
+      Configuration hdfsConf = new Configuration();
+      for (String path :
+          ufsConfigValue.substring(Name.UNDERFS_HDFS_CONFIGURATION_PREFIX.length()).split(":")) {
+        if (!path.isEmpty()) {
+          hdfsConf.addResource(new Path(path));
+        }
+      }
+      ufsConf.merge(hdfsConf.getPropsWithPrefix(Name.UNDERFS_HDFS_CONFIGURATION_PREFIX),
```
gpang, line 132: I don't know how this works.

 I thought this would convert the hdfs conf to key/values. However, this is taking all the property keys which have the prefix of `UNDERFS_HDFS_CONFIGURATION_PREFIX`, which is "kv@". No hdfs properties will have that prefix, right?

gpang (changes requested):
@maobaolong Thanks! I left a few questions.

core/common/src/main/java/alluxio/conf/PropertyKey.java
maobaolong: Done.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
maobaolong: Sorry for that, removed it.

core/common/src/main/java/alluxio/conf/PropertyKey.java
maobaolong: Yeah, done.

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
maobaolong: Oh, sorry, I make a mistake here, I suppose to `getPropsWithPrefix("")` to merge all hdfs configuration from the given files. Thank you for point out this bug.

maobaolong:
@gpang Thanks you for your review, I corrected the related code and push a new commit, PTAL.

maobaolong:
image

image

Through the above debug screenshot, we can see that there are 8 properties from two given hdfs configuration files, and it can be merged into `UnderFileSystemConfiguration` as expected.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -667,7 +667,10 @@ public String toString() {
               "${%s}/core-site.xml:${%s}/hdfs-site.xml", Name.CONF_DIR, Name.CONF_DIR))
           .setDescription("Location of the HDFS configuration file to overwrite "
               + "the default HDFS client configuration. Note that, these files must be available"
-              + "on every node.")
+              + "on every node, but if there are a "
```
apc999, line 670: why not just support this behavior by default? I think it will be strictly better

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
```diff
@@ -11,13 +11,18 @@
...
+import alluxio.Constants;
 import alluxio.annotation.PublicApi;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.AlluxioProperties;
 import alluxio.conf.ConfigurationValueOptions;
 import alluxio.conf.InstancedConfiguration;
+import alluxio.conf.PropertyKey;
 import alluxio.conf.Source;
 
+import org.apache.hadoop.conf.Configuration;
```
apc999, line 23: it can be tricky to introduce dependency of `org.apache.hadoop` in common.
possible to move it to master module?

apc999 (changes requested):

core/common/src/main/java/alluxio/conf/PropertyKey.java
gpang: I see. So, this PR achieves this behavior by reading in the conf files, and converting them to mount options, so when workers get the ufs config, they will get all the hadoop configs too.

This is turned on when "kv@" is prefixed. Are you saying just enable this behavior all the time?

core/common/src/main/java/alluxio/underfs/UnderFileSystemConfiguration.java
maobaolong: Ok, move the logic into master module.

core/common/src/main/java/alluxio/conf/PropertyKey.java
maobaolong: @apc999 @gpang  It can be more simply if we turned on "kv@" all the time, but not sure wether we should support different worker use different hdfs config files? If not, I'm ok to push a new commit to strictly support this behavior only.

maobaolong: I push the last commit  c344662c117aa0d9401ea2e8d9098f39d0699f70

maobaolong:
@apc999 @gpang Thank you for the previous review and suggestion, and sorry for the late last week. I push two new commit to address your comments, PTAL. Thanks

gpang:
@maobaolong Thanks! I have 2 high level questions.

- If we make the "key-value" mode the default way, would there be any sensitive information that gets added to the mount properties?
- Which properties from hdfs are added to mount properties? Is it every hdfs config, or is it just the ones explicitly specified in the files? Also, the hdfs config files could have a lot of properties, so one concern is that the mount properties grows to a very large set of settings.

maobaolong:
@gpang Thanks for your question.

> If we make the "key-value" mode the default way, would there be any sensitive information that gets added to the mount properties?

Yes, all the key-value in the config file will gets added to the mount properties.

> Which properties from hdfs are added to mount properties?

Just the specified config files from the given mount option `alluxio.underfs.hdfs.configuration`

>  Also, the hdfs config files could have a lot of properties, so one concern is that the mount properties grows to a very large set of settings.

We use `new Configuration(false)`, it means we don't load default files, something like `core-default.xml`, `hdfs-default.xml`, we only load the `alluxio.underfs.hdfs.configuration` specified files, so I think the mount properties couldn't be very large. 

there are these high-level question in this approach, I'm open to discuss that if there are another better way to help us don't need to sync hdfs configuration files to all worker nodes.

Another option approach: put the ufs configuration files into a shared file system, alluxio itself is the best choice.

apc999:
as discussed offline, instead of changing the master side, we will enhance MountCommand to support something like `--conf` to take the conf files like core-site

maobaolong:
@apc999 Thank you for your suggestion, now I updated this PR, move the config file extracting logic to shell side. I did a test locally by execute `mount --cfgfile /tmp/a.xml --cfgfile /tmp/b.xml /abc file:///tmp/abc`, the config items inside the config files which I specified can be converted to options, that is expected behaivor.

PTAL.

shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
```diff
@@ -62,6 +64,14 @@
           .valueSeparator('=')
           .desc("options associated with this mount point")
           .build();
+  private static final Option CONFIG_FILE_OPTION =
+      Option.builder()
+          .longOpt("cfgfile")
```
apc999, line 69: how bout
```suggestion
          .longOpt("hdfsconf")
```

shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
```diff
@@ -62,6 +64,14 @@
...
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .desc("config file path associated with this mount point")
```
apc999, line 73:
```suggestion
          .desc("path to its config files for an HDFS mount point")
```

shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
```diff
@@ -110,7 +130,8 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
 
   @Override
   public String getUsage() {
-    return "mount [--readonly] [--shared] [--option <key=val>] <alluxioPath> <ufsURI>";
+    return "mount [--readonly] [--shared] [--option <key=val>] [--cfgfile <cfgFilePath>] "
```
apc999, line 133: change accordingly

apc999 (changes requested):
Minor comments left. Otherwise this PR looks good to me

shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
maobaolong: Done, fix it in the new commit.

shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
maobaolong: Done, fix it in the new commit.

shell/src/main/java/alluxio/cli/fs/command/MountCommand.java
maobaolong: Done, fix it in the new commit.

maobaolong:
@apc999 Thank you for your review, I push a new commit to address the comment. PTAL

apc999 (approved):

apc999:

LuQQiu:
@Xenorith Can you elaborate the issue found?

LuQQiu:
In a sanity check cluster with Alluxio cluster + HDFS as root UFS,
when starting the Alluxio master, error 
```
Caused by: java.lang.IllegalArgumentException: Unable to create an UnderFileSystem instance for path: hdfs://SanityMicrobench-masters-1:9000/alluxio_storage
        at alluxio.underfs.UnderFileSystem$Factory.create(UnderFileSystem.java:122)
        at alluxio.underfs.AbstractUfsManager.getOrAdd(AbstractUfsManager.java:132)
        at alluxio.underfs.AbstractUfsManager.lambda$addMount$0(AbstractUfsManager.java:171)
        at alluxio.underfs.UfsManager$UfsClient.acquireUfsResource(UfsManager.java:61)
        at alluxio.master.file.DefaultFileSystemMaster.getRootMountInfo(DefaultFileSystemMaster.java:502)
        at alluxio.master.file.DefaultFileSystemMaster.<init>(DefaultFileSystemMaster.java:444)
        at alluxio.master.file.DefaultFileSystemMaster.<init>(DefaultFileSystemMaster.java:425)
        at alluxio.master.file.FileSystemMasterFactory.create(FileSystemMasterFactory.java:51)
        at alluxio.master.file.FileSystemMasterFactory.create(FileSystemMasterFactory.java:28)
        at alluxio.master.MasterUtils.lambda$createMasters$0(MasterUtils.java:49)
        Suppressed: java.lang.NoClassDefFoundError: org/apache/commons/configuration/Configuration
                at org.apache.hadoop.metrics2.lib.DefaultMetricsSystem.<init>(DefaultMetricsSystem.java:38)
                at org.apache.hadoop.metrics2.lib.DefaultMetricsSystem.<clinit>(DefaultMetricsSystem.java:36)
                at org.apache.hadoop.security.UserGroupInformation$UgiMetrics.create(UserGroupInformation.java:121)
                at org.apache.hadoop.security.UserGroupInformation.<clinit>(UserGroupInformation.java:237)
                at alluxio.underfs.hdfs.HdfsUnderFileSystem.<init>(HdfsUnderFileSystem.java:150)
                at alluxio.underfs.hdfs.HdfsUnderFileSystem.createInstance(HdfsUnderFileSystem.java:108)
                at alluxio.underfs.hdfs.HdfsUnderFileSystemFactory.create(HdfsUnderFileSystemFactory.java:43)
                at alluxio.underfs.UnderFileSystem$Factory.create(UnderFileSystem.java:107)
                ... 13 more
        Caused by: java.lang.ClassNotFoundException: org.apache.commons.configuration.Configuration
                at java.lang.ClassLoader.findClass(ClassLoader.java:523)
                at alluxio.extensions.ExtensionsClassLoader$DefaultClassLoader.findClass(ExtensionsClassLoader.java:43)
                at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
                at alluxio.extensions.ExtensionsClassLoader$DefaultClassLoader.loadClass(ExtensionsClassLoader.java:53)
                at alluxio.extensions.ExtensionsClassLoader.loadClass(ExtensionsClassLoader.java:84)
                at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
```
shown and master failed to come up, will create an issue for that

LuQQiu:
Will revert this PR for now @apc999 what's your opinion?

PR #12785, opened 2021-02-01
Prevent leader from writing to embedded journal after stepping down

bf8086:
Existing Alluxio master with embedded journal will attempt to write to the quorum even after it stepped down in various failure situation, journaling a bunch of inconsistent master states as a result. Even though the new master has a catchup process to wait for these journal entries before serving request, there is a chance for the old leader to send it after the period ends and corrupt the journal.

This change updates the journal writer to only write to local raft server by default. This should prevent any remote write request from an old leader at the cost of some temporary request failures during the fail over.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1599,7 +1599,9 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED)
           .setDefaultValue(false)
-          .setDescription("Whether the journal writer will write to remote master.")
+          .setDescription("Whether the journal writer will write to remote master. This is "
+              + "disabled by default and should not be turned on unless Alluxio encounters issues "
+              + "with local journal write.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .setIsHidden(true)
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -1596,6 +1596,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED)
+          .setDefaultValue(false)
+          .setDescription("Whether the journal writer will write to remote master.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .setIsHidden(true)
+          .build();
   public static final PropertyKey MASTER_EMBEDDED_JOURNAL_WRITE_TIMEOUT =
       new Builder(Name.MASTER_EMBEDDED_JOURNAL_WRITE_TIMEOUT)
           .setDefaultValue("30sec")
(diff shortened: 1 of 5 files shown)


bf8086:
@LuQQiu PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
```diff
@@ -70,7 +74,7 @@ public LocalFirstRaftClient(RaftServer server, Supplier<RaftClient> clientSuppli
    */
   public CompletableFuture<RaftClientReply> sendAsync(Message message,
       TimeDuration timeout) throws IOException {
-    if (mClient == null) {
+    if (!mEnableRemoteClient || mClient == null) {
```
LuQQiu, line 77: WRITE_REMOTE_ENABLED == false && WRITE_LOCAL_FIRST_ENBLAED == false send local request
WRITE_REMOTE_ENABLED == true &&  WRITE_LOCAL_FIRST_ENBLAED == false send remote request
WRITE_REMOTE_ENABLED == false &&  WRITE_LOCAL_FIRST_ENBLAED == true send local request
WRITE_REMOTE_ENABLED == true &&  WRITE_LOCAL_FIRST_ENBLAED == true send local request

so only WRITE_REMOTE_ENABLED == true &&  WRITE_LOCAL_FIRST_ENBLAED == false send remote request
two properties are somehow a little difficult to understand what's desire, can we consolidate them into one property and add a description

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1596,6 +1596,14 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED =
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED)
+          .setDefaultValue(false)
+          .setDescription("Whether the journal writer will write to remote master.")
```
LuQQiu, line 1602: So currently, the leader will only write to local journal and Raft in the backend is responsible for replicating the leader local journal to all other masters?

During leadership transition, Raft will stop replicating the old leader journal to other masters and start replicating the new leader journal.

Is it true that all the local journal of the previous leading master will be replicated to all other nodes?
and the previous leading master steps down, it will fail to write to local journal because NoLeaderException and related RPCs will be returned as failure?

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
```diff
@@ -66,18 +72,20 @@ public void before() throws Exception {
...
+        () -> mFollowerJournalSystem.isLeader() || mLeaderJournalSystem.isLeader(), mWaitOptions);
     if (journalSystems.get(1).isLeader()) {
       mLeaderJournalSystem = journalSystems.get(1);
       mFollowerJournalSystem = journalSystems.get(0);
     }
     // Transition primary journal to primacy state.
-    mFollowerJournalSystem.notifyLeadershipStateChanged(false);
-    mLeaderJournalSystem.notifyLeadershipStateChanged(true);
+//    mFollowerJournalSystem.notifyLeadershipStateChanged(false);
+//    mLeaderJournalSystem.notifyLeadershipStateChanged(true);
```
LuQQiu, line 83: Can be removed if no needed?

LuQQiu (commented):
Thanks @bf8086 for fixing this critical issue! Left some comments

core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
bf8086: `WRITE_REMOTE_ENABLED` isn't a property intent for user configuration in normal cases. It is only in case that users see issues with local only writes then they can re-enable the old behavior. Once the new behavior is validated over time, we should remove the entire remote write logic and issue only local writes.

core/common/src/main/java/alluxio/conf/PropertyKey.java
bf8086: Once stepped down, the journal writes in previous leader will fail and pending requests will be rejected. The clients will retry and eventually succeed once the new leader is up.

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
bf8086: Done.

core/server/common/src/main/java/alluxio/master/journal/raft/LocalFirstRaftClient.java
LuQQiu: Oh got it, then could you update the description for write_remote_enabled to include more information (like when it should be used), thanks

bf8086: Added some more description.

bf8086:
Thanks @LuQQiu. I added some details in the property description. PTAL.

LuQQiu (approved):
LGTM, thanks!

bf8086:

bf8086:

PR #12792, opened 2021-02-02
Report coverage even when tests fail

bradyoo:


ZacBlanco:
Hi @bradyoo can you provide a short description in the PR to what exactly you're changing in this PR?

.github/workflows/java8_integration_tests.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       - name: "Upload Code Coverage Report"
         uses: codecov/codecov-action@v1
         # skip generating+uploading coverage on these matrix modules because they are flaky
-        if: ${{ matrix.modules != 'alluxio.server.ft.**,!alluxio.server.ft.journal.raft.**,!alluxio.server.ft.journal.ufs.**' && matrix.modules != 'alluxio.client.cli.**,!alluxio.client.cli.fs.**' }}
+        if: ${{ always() && matrix.modules != 'alluxio.server.ft.**,!alluxio.server.ft.journal.raft.**,!alluxio.server.ft.journal.ufs.**' && matrix.modules != 'alluxio.client.cli.**,!alluxio.client.cli.fs.**' }}
```
ZacBlanco, line 93: isn't this essentially the same as saying `(true && condition)` which logically simplifies to `(condition)`? ditto for other line

.github/workflows/java8_integration_tests_webui.yml
```diff
@@ -71,6 +72,7 @@ jobs:
           retention-days: 7
       - name: "Upload Code Coverage Report"
         uses: codecov/codecov-action@v1
+        if: always()
```
ZacBlanco, line 75: if generating fails, then I believe that the uploading task will fail since it won't be able to find a coverage report

.github/workflows/java8_unit_tests.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           retention-days: 7
       - name: "Upload Code Coverage Report"
         uses: codecov/codecov-action@v1
+        if: always()
```
ZacBlanco, line 77: same comment on failure if coverage report isn't generated

.github/workflows/java8_unit_tests.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         timeout-minutes: 60
       - name: Generate coverage report
         id: cov0
+        if: always()
```
ZacBlanco, line 54: if the tests fail, do we really want to generate the coverage report?

bradyoo:
@ZacBlanco So the problem here is that, we have been skipping coverage when the test fails and when it does. Coverage report doesn't get sent. When this happens, the codecov.io still appears to interpret every build as successful (because every coverage report they got was a success) and this can indirectly cause a pretty large difference in coverage report.

.github/workflows/java8_integration_tests.yml
bradyoo: Neh.. This is actually not the same.. Take a look at https://github.community/t/multiple-if-conditional-with-always-is-always-true/128840. The way they handle this is a little odd.

.github/workflows/java8_integration_tests_webui.yml
bradyoo: Unfortunately, when tests fail, generating can still succeed and in that case, you want to upload the report.

.github/workflows/java8_unit_tests.yml
bradyoo: Yes. Because sending the failed report is the main way codecov knows that the ci build failed.

.github/workflows/java8_integration_tests_webui.yml
ZacBlanco: yes, but for failed tests, it is possible to miss coverage which makes the report not so helpful

```
test() {
doSomething() // covers 50% of lines
assert(condition) // <--- test fails here
doSomethingElse() // covers remaining 50%
assert(condition2)
}
```

I guess if you still want to check coverage for a section of code where the test passed, the upload can still be useful, but the overall numbers will probably still differ in a significant manner

.github/workflows/java8_integration_tests.yml
ZacBlanco: weird...

ZacBlanco (approved):

.github/workflows/java8_integration_tests_webui.yml
bradyoo: Obviously the report isn't full. But what's more important is that the codecov knows that there is a failed test in the coverage. That's important because or else, it thinks failed builds are successful builds that are treated as baseline for coverage.

bradyoo:

PR #12794, opened 2021-02-02
Address race condition for async persist

calvinjia:
Fixes a race condition where if a file is scheduled for async persist changes in between submitting the async persist request and starting the async persist request, the wrong data will be written for the original file.

I was able to reproduce this consistently with a unit test and adding a timing delay to the job service.
Not including the test because it would require a hook in the job service for timing delays, perhaps this is something we can add in the future when we have a framework for controlling timings between different services.

For reference the test:

```
  @Test
  public void persistMultiRename() throws Exception {
    String testFilePath = "/testPersist/multiRenameTestFile";
    FileSystemTestUtils.createByteFile(sFileSystem, testFilePath, WritePType.MUST_CACHE, 2);
    sFileSystem.persist(new AlluxioURI(testFilePath));
    CommonUtils.sleepMs(1000);
    sFileSystem.rename(new AlluxioURI(testFilePath), new AlluxioURI(testFilePath + 2));
    FileSystemTestUtils.createByteFile(sFileSystem, testFilePath, WritePType.MUST_CACHE, 1);
    CommonUtils.sleepMs(3000);
    sFileSystem.persist(new AlluxioURI(testFilePath));
    sFileSystem.rename(new AlluxioURI(testFilePath), new AlluxioURI(testFilePath + 1));
    for (int i = 2; i > 0; i--) {
      final AlluxioURI uri = new AlluxioURI(testFilePath + i);
      CommonUtils.waitFor("File to be persisted", () -> {
        try {
          return sFileSystem.getStatus(uri).isPersisted();
        } catch (Exception e) {
          return false;
        }
      }, WaitForOptions.defaults().setTimeoutMs(10000));
      checkFilePersisted(uri, i);
    }
  }
```

Without the fix this will result in writing two 1 byte files. With the fix, one job will fail and retry, on the retry the correct (renamed) path will be picked up and both files will have the appropriate length.

Adding a hidden flag to enable/disable and will remove in a later patch.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -140,6 +140,7 @@ import alluxio.security.authorization.AclEntryType;
 import alluxio.security.authorization.Mode;
 import alluxio.underfs.Fingerprint;
 import alluxio.underfs.MasterUfsManager;
+import alluxio.underfs.UfsFileStatus;
 import alluxio.underfs.UfsManager;
 import alluxio.underfs.UfsMode;
 import alluxio.underfs.UfsStatus;
@@ -3964,6 +3965,23 @@ public final class DefaultFileSystemMaster extends CoreMaster
               String ufsPath = resolution.getUri().toString();
               ufs.setOwner(tempUfsPath, inode.getOwner(), inode.getGroup());
               ufs.setMode(tempUfsPath, inode.getMode());
+
+              // Check if the size is the same to guard against a race condition where the Alluxio
+              // file is mutated in between the persist command and execution
+              if (ServerConfiguration.isSet(PropertyKey.MASTER_ASYNC_PERSIST_SIZE_VALIDATION)
+                  && ServerConfiguration.getBoolean(
+                      PropertyKey.MASTER_ASYNC_PERSIST_SIZE_VALIDATION)) {
+                UfsStatus ufsStatus = ufs.getStatus(tempUfsPath);
+                if (ufsStatus.isFile()) {
+                  UfsFileStatus status = (UfsFileStatus) ufsStatus;
+                  if (status.getContentLength() != inode.getLength()) {
+                    throw new IOException(String.format("%s size does not match. Alluxio expected"
+                        + "length: %d, UFS actual length: %d", tempUfsPath,
+                        inode.getLength(), status.getContentLength()));
+                  }
+                }
+              }
+
               if (!ufsPath.equals(tempUfsPath)) {
                 // Make rename only when tempUfsPath is different from final ufsPath. Note that,
                 // on object store, we take the optimization to skip the rename by having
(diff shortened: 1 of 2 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3964,6 +3965,23 @@ private void handleSuccess(PersistJob job) {
...
+
+              // Check if the size is the same to guard against a race condition where the Alluxio
+              // file is mutated in between the persist command and execution
+              if (ServerConfiguration.isSet(PropertyKey.MASTER_ASYNC_PERSIST_SIZE_VALIDATION)
+                  && ServerConfiguration.getBoolean(
+                      PropertyKey.MASTER_ASYNC_PERSIST_SIZE_VALIDATION)) {
+                UfsStatus ufsStatus = ufs.getStatus(tempUfsPath);
+                if (ufsStatus.isFile()) {
+                  UfsFileStatus status = (UfsFileStatus) ufsStatus;
+                  if (status.getContentLength() != inode.getLength()) {
```
HelloHorizon, line 3977: If UFS file is correct, will it promise that the size of UFS file has the same size in Alluxio?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3964,6 +3965,23 @@ private void handleSuccess(PersistJob job) {
...
+                    throw new IOException(String.format("%s size does not match. Alluxio expected"
+                        + "length: %d, UFS actual length: %d", tempUfsPath,
```
HelloHorizon, line 3979: (nit) add the alluxio file path and file id

HelloHorizon (changes requested):
@calvinjia Thanks! I left some minor comments

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: These will get logged further down in the catch clause.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
calvinjia: Good point. This is why I added the flag. I haven't come across a case where the data size is different between UFS and Alluxio, but I could see that happening if the UFS API is providing true file size as opposed to logical file size.

HelloHorizon: I see, in this case, do we want to set `MASTER_ASYNC_PERSIST_SIZE_VALIDATION ` to false by default?

calvinjia: I'm setting it to true because I don't expect that case to come up. If it does break some special workflow, their persist jobs will fail until this is turned off.

HelloHorizon: ok, do you mind adding some comments here in case someone met inconsistent file size here?

calvinjia: Updated

HelloHorizon (approved):
LGTM

calvinjia:

calvinjia:

PR #12795, opened 2021-02-02
Add an option for active job count

jiacheliu3:
Commands like distributedLoad and distributedCp submits jobs to the cluster job service. The command creates 1 job for each file. By default the command allows up to 3000 jobs to be submitted and waiting for completion. When there are many distributedLoad/Cp commands running by many users, there may be too many jobs in the queue for the job service.

This change intends to add a way to throttle the job submission speed at the command side. Especially when the distributedLoad/Cp jobs are submitted with automated scripts, we may want less than 3000 jobs to run with each command. Maybe find a balance between speed and resource claim, like 1000 or 500.

Volume control on Alluxio job service is beyond the scope of this PR.

Example:
```
$ bin/alluxio fs distributedLoad --job-count 1000 /                      
Allow 1000 active jobs
...
```

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
@@ -49,15 +49,16 @@ import javax.annotation.concurrent.ThreadSafe;
 public class DistributedCpCommand extends AbstractDistributedJobCommand {
   private String mWriteType;
 
-  private static final Option JOB_COUNT_OPTION =
+  private static final Option ACTIVE_JOB_COUNT_OPTION =
       Option.builder()
-          .longOpt("job-count")
+          .longOpt("active-jobs")
           .required(false)
           .hasArg(true)
           .numberOfArgs(1)
           .type(Number.class)
-          .argName("job count")
-          .desc("Number of active jobs to run at the same time"
+          .argName("active job count")
+          .desc("Number of active jobs that can run at the same time. Later jobs must wait. "
+                  + "The default upper limit is "
                   + AbstractDistributedJobCommand.DEFAULT_ACTIVE_JOBS)
           .build();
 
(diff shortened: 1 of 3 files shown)


shell/src/main/java/alluxio/cli/fs/command/DistributedCpCommand.java
```diff
@@ -145,16 +179,6 @@ private void addJob(String srcPath, String dstPath) {
...
-  @Override
-  public String getUsage() {
-    return "distributedCp <src> <dst>";
-  }
-
-  @Override
-  public String getDescription() {
-    return "Copies a file or directory in parallel at file level.";
-  }
-
```
jiacheliu3, line 157: Move up closer to other getters, better for the eyes.

yuzhu:
how does this interact with the `alluxio.job.worker.threadpool.size` property? does the latter not control the former somehow as well? because ultimately if there is no jobworker threads, we can't run the distributed load jobs?

jiacheliu3:
Closed since resource is already managed by `alluxio.job.worker.threadpool.size`

jiacheliu3:
@yuzhu Actually on 2nd thought, I think it's reasonable to have the control on the submitter side anyways.

bradyoo:
this doesn't really do what you think it does. And when users reduce this down to a number like 100, it's going to be really slow because those jobs are going to be really slow even when there is only one job worker that handles at most 20 concurrent jobs because this setting is not really about concurrent jobs. 

This is a very misleading setting.

gpang:
@bradyoo I don't understand why this flag is not what what we think it means? I think it does do what we think, but if tasks are very short compared to the heartbeat time, there will be a lot of unused job worker resources, right?

jiacheliu3:
@bradyoo @gpang I believe a better name would be "active jobs" rather than "concurrent jobs", because the concurrency is provided by the job worker and the corresponding job worker threadpool.

Yea 20 is going to be unreasonably small. But there might be some good numbers like 500 or 1000 that finds a balance between speed and being nice to the other users, right?

bradyoo:
LGTM

jiacheliu3:

jiacheliu3:

jiacheliu3:

Xenorith:

PR #12797, opened 2021-02-02
Use Generic type to make OpenFileEntry extensible

maobaolong:


integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
```diff
@@ -39,7 +39,7 @@
  * Utility methods for Alluxio-FUSE.
  */
 @ThreadSafe
-public final class AlluxioFuseUtils {
+public class AlluxioFuseUtils {
```
jiacheliu3, line 42: Do you want to extend `AlluxioFuseUtils`?

maobaolong: Yeah, I have a project dependent alluxio-fuse, my Class want to extends `AlluxioFuseUtils `

maobaolong:
@jiacheliu3 Thanks for your previous review, please take a look again

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
jiacheliu3: It's not great to extend an utility class, so they are often made `final`. and only have static methods. I think you are not inheriting anything when you extend `AlluxioFuseUtils` right? I would actually recommend just having a separate class to keep the extra util methods.

Everything else looks good to me :)

maobaolong: @jiacheliu3 Thank you for your reply, It make sense to keep final

jiacheliu3 (approved):
@maobaolong Thanks for resolving the comments, LGTM

maobaolong:
@jiacheliu3 I revert to keep final for AlluxioFuseUtils by the last commit. PTAL.

jiacheliu3:

jiacheliu3:

PR #12800, opened 2021-02-03
[SMALLFIX] Add debug log on BlockMaster request size

jiacheliu3:
When the worker has a large number of blocks the request can be big for worker register and block heartbeat. Adding a little debug log for better traffic visibility.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
@@ -143,7 +143,8 @@ public final class BlockMasterWorkerServiceHandler extends
   public void registerWorker(RegisterWorkerPRequest request,
       StreamObserver<RegisterWorkerPResponse> responseObserver) {
     if (LOG.isDebugEnabled()) {
-      LOG.debug("Register worker request is {} bytes, containing {} blocks", request.getSerializedSize(),
+      LOG.debug("Register worker request is {} bytes, containing {} blocks",
+              request.getSerializedSize(),
               request.getCurrentBlocksCount());
     }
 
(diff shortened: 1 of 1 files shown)


jiacheliu3:
@bf8086 anther alternative is to check the message size, and log a warn when this is larger than a threshold of, say 10MB.

gpang:
Why do we need to know the sizes of these messages?

jiacheliu3:
> Why do we need to know the sizes of these messages?

@gpang Some user environments have a huge number of blocks per worker. For example, a worker with 6TB storage in total can have more than 2 million blocks. On registering with the master we see much stress in CPU and network. So I'd like to get more visibility to better understand.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -61,6 +61,12 @@ public BlockMasterWorkerServiceHandler(BlockMaster blockMaster) {
   @Override
   public void blockHeartbeat(BlockHeartbeatPRequest request,
       StreamObserver<BlockHeartbeatPResponse> responseObserver) {
+    if (LOG.isDebugEnabled()) {
```
maobaolong, line 64: Is this log level check necessary?

bf8086:
> @bf8086 anther alternative is to check the message size, and log a warn when this is larger than a threshold of, say 10MB.

yeah you can probably get a rough estimation by checking the block counts instead which has less performance impact.

bf8086 (approved):

jiacheliu3:

jiacheliu3:

jiacheliu3:

jiacheliu3:

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
jiacheliu3: I think we avoid one calculation if the log level is above debug

Issue #11474, opened 2020-05-25
chown -R on Alluxio 2.2.1 does not change all child inodes owner/group


JySongWithZhangCe:
**Alluxio Version:**
What version of Alluxio are you using?
2.2.1
**Describe the bug**
chown -R can not change all subdir or file to correct user/group
**To Reproduce**
use alluxio fs chown -R user:group /path command to change /path owner
where will be 4 combination in subdir and file:
user_new:group_new
user_old:group_old
user_new:group_old
user_old:group_new
and which user/group will a subdir or path belong can not predict

there are no log even in debug mode

jiacheliu3:
I had a session with @JySongWithZhangCe and confirmed on the above behavior.

The directory is not too deeply nested but pretty large each. We tried an example like /A/B1~100/C1~1000 where we did `alluxio fs -R new_user:new_group /A`. And we had the below observations:

1. This is very likely a regression. @JySongWithZhangCe 's team upgraded from 2.1 to 2.2.1 and immediately had this issue. Their team performs chown operations very frequently and they should be able to identify the same issue had it existed in 2.1. @JySongWithZhangCe can you please confirm on the open source version of Alluxio which is verified to be fine?

1. ALL the direct children (B1 ~ B100) are correctly updated. 

1. Many grand children are not successfully updated and end in one of the 4 combinations. 

1. As far as we observe, ALL the children in one particular directory are in one user:group combination. @JySongWithZhangCe can you please confirm on this?

1. The call took only a few seconds to complete and completed successfully. We found no logs on master or client side regarding this chown operation.

JySongWithZhangCe:
> I had a session with @JySongWithZhangCe and confirmed on the above behavior.
> 
> The directory is not too deeply nested but pretty large each. We tried an example like /A/B1-100/C1-1000 where we did `alluxio fs -R new_user:new_group /A`. And we had the below observations:
> 
> 1. This is very likely a regression. @JySongWithZhangCe 's team upgraded from 2.1 to 2.2.1 and immediately had this issue. Their team performs chown operations very frequently and they should be able to identify the same issue had it existed in 2.1. @JySongWithZhangCe can you please confirm on the open source version of Alluxio which is verified to be fine?
> 2. ALL the direct children (B1 ~ B100) are correctly updated.
> 3. Many grand children are not successfully updated and end in one of the 4 combinations.
[...]

PR #12810, opened 2021-02-05
Add statistic info for setAttributeInternal

HelloHorizon:
debug log for troubleshooting https://github.com/Alluxio/alluxio/issues/11474


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -3230,6 +3230,8 @@ private void setAttributeInternal(RpcContext rpcContext, LockedInodePath inodePa
     long opTimeMs = System.currentTimeMillis();
     if (context.getOptions().getRecursive() && targetInode.isDirectory()) {
       try (LockedInodePathList descendants = mInodeTree.getDescendants(inodePath)) {
+        LOG.debug("Set attribute for {} descendants under path {}.",
+            descendants.getInodePathList().size(), inodePath.getUri());
```
gpang, line 3234: Is `getInodePathList()` expensive? If it is, then we should avoid doing this computation if debug is disabled.

HelloHorizon:
Close it since we have a better way to do log

PR #12812, opened 2021-02-05
Fix false positives in dockerfile env.sh

madanadit:
The current logic includes extra variables such as `PATH` or `HOSTNAME` due to regex matching


madanadit:

PR #12813, opened 2021-02-05
Cleanup counters in local client cache

apc999:
two improvements:
1. Added new counter "Client.CachePutInsufficientSpaceErrors" indicating the failed put due to insufficient space made after eviction attempts.
2. Remove unnecessary cleanup which may artificially bump error counters
3. Remove unnecessary directory cleanup which has potential race condition to fail put unnecessarily

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
@@ -23,7 +23,6 @@ import org.slf4j.LoggerFactory;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.RandomAccessFile;
-import java.nio.file.DirectoryStream;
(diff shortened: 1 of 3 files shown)


calvinjia (approved):

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -127,13 +126,6 @@ public void delete(PageId pageId) throws IOException, PageNotFoundException {
       throw new PageNotFoundException(p.toString());
     }
     Files.delete(p);
-    Path parent = Preconditions.checkNotNull(p.getParent(),
```
apc999, line 130: removing parent is unnecessary and introduces race on the parent dir among threads, causing put failure unnecessarily. Remove this logic

apc999:

PR #12824, opened 2021-02-09
Add feature check

HelloHorizon:


core/server/common/src/main/java/alluxio/util/FeatureUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        .equals(ServerConfiguration.get(PropertyKey.MASTER_JOURNAL_TYPE));
+  }
+
+  /**
+   * Utility to check the master metastore is rocks.
+   *
+   * @return true, if running with rocks
+   */
+  public static boolean isRocks() {
+    return "ROCKS".equals(ServerConfiguration.get(PropertyKey.MASTER_METASTORE));
```
Xenorith, line 38: can this be `MetastoreType.ROCKS.toString()`?

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -98,7 +111,46 @@ public static String getUserAgentString(String clusterID) throws IOException {
...
+      featureInfo.add("persistBlackList");
+    }
+    if (FeatureUtils.isUnsafeDirectPersistEnable()) {
+      featureInfo.add("unsafePersist");
+    }
+    if (FeatureUtils.isMasterAuditLoggingEnable()) {
+      featureInfo.add("masterAuditLog");
+    }
+    if (featureInfo.size() == 0) {
+      return null;
```
Xenorith, line 151: should the method be annotated somehow if this method can return null?
is there a unit test for this corner case?

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -84,6 +85,18 @@ public static String getLatestVersion(String clusterID, long connectionRequestTi
    */
   @VisibleForTesting
   public static String getUserAgentString(String clusterID) throws IOException {
```
Xenorith, line 87: update javadoc for this method, specifically the `@return`

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -84,6 +85,18 @@ public static String getLatestVersion(String clusterID, long connectionRequestTi
...
+    Joiner joiner = Joiner.on("; ").skipNulls();
+    String sysInfo = joiner.join(getUserAgentEnvironmentString(clusterID), getFeatureString());
+    return String.format("Alluxio/%s (%s)", ProjectConstants.VERSION, sysInfo);
+  }
+
+  /**
+   * @param clusterID the cluster ID
+   * @return a string representation of the user's environment in the format "key1:value1, key2:
```
Xenorith, line 95: is this javadoc accurate? i don't think they are always key/value pairs and each section is separated by `;`, not `,`

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -98,7 +111,46 @@ public static String getUserAgentString(String clusterID) throws IOException {
...
       }
     }
-    return String.format("Alluxio/%s (%s)", ProjectConstants.VERSION, sysInfo);
+    return sysInfo;
+  }
+
+  /**
+   * Get the features information.
+   *
+   * @return a list of string represent feature names
```
Xenorith, line 120: `a list of strings representing enabled features in the format ...`

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
```diff
@@ -46,86 +53,135 @@ public void before() {
...
     Mockito.when(EnvironmentUtils.getEC2ProductCode()).thenReturn("random123code");
     Mockito.when(EnvironmentUtils.isEMR(Mockito.anyString())).thenReturn(true);
     Mockito.when(EC2MetadataUtils.getUserData()).thenReturn("emr_apps");
 
+    String userAgentString = UpdateCheck.getUserAgentEnvironmentString("cluster1");
+    Assert.assertTrue(userAgentString.equals("cluster1; ProductCode:random123code; emr; ec2"));
+  }
+
+  @Test
+  public void featureString() {
```
Xenorith, line 129: should this be split up into different unit tests for each individual feature to be consistent with the existing unit tests?

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
```diff
@@ -46,86 +53,135 @@ public void before() {
...
+    // Embedded journal
+    ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS);
+    Assert.assertTrue(!UpdateCheck.getFeatureString().contains("embedded"));
```
Xenorith, line 132: using `String.contains` is a weak check since another feature could have the substring `embedded` or have the equivalent value. how can this unit test be reformatted to catch the bad scenario where there's another feature that is also called `embedded`? same for other tests

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
```diff
@@ -46,86 +53,135 @@ public void before() {
...
+    ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.EMBEDDED);
+    Assert.assertTrue(UpdateCheck.getFeatureString().contains("embedded"));
+
+    // Rocks
+    ServerConfiguration.set(PropertyKey.MASTER_METASTORE, MetastoreType.ROCKS);
+    Assert.assertTrue(UpdateCheck.getFeatureString().contains("rocks"));
+    ServerConfiguration.set(PropertyKey.MASTER_METASTORE, MetastoreType.HEAP);
+    Assert.assertTrue(!UpdateCheck.getFeatureString().contains("rocks"));
```
Xenorith, line 140: `Assert.assertTrue(!...)` should be `Assert.assertFalse(...)`

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Utility to check Zookeeper is enabled.
+   *
+   * @return true, if Zookeeper is enabled
+   */
+  public static boolean isZookeeperEnable() {
```
Xenorith, line 46: instead of having a method for each property, can we have a single method for boolean properties that takes in the argument for the property key? this would make it easier to reuse in the future

Xenorith, line 46: isZookeeperEnable -> isZookeeperEnabled, similar for below

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
```diff
@@ -0,0 +1,99 @@
...
+    return ServerConfiguration.getBoolean(PropertyKey.MASTER_DAILY_BACKUP_ENABLED);
+  }
+
+  /**
+   * Utility to check persistence black list is empty.
+   *
+   * @return true, if persistence black list is empty
+   */
+  public static boolean isPersistenceBlacklistEmpty() {
+    if (ServerConfiguration.isSet(PropertyKey.MASTER_PERSISTENCE_BLACKLIST)) {
```
Xenorith, line 74: why split this up into 2 nested if statements? seems like it can be represented as a single line of `return condition1 &&/|| condition2`

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -98,7 +111,46 @@ public static String getUserAgentString(String clusterID) throws IOException {
...
+   */
+  @VisibleForTesting
+  public static String getFeatureString() {
+    Joiner joiner = Joiner.on("; ").skipNulls();
+    List<String> featureInfo = new ArrayList<>();
+    if (FeatureUtils.isEmbeddedJournal()) {
```
Xenorith, line 126: this list looks really long and repetitive. not sure if this is possible in java, but could this be represented as a map of string to function, with a loop iterating over the map?

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
```diff
@@ -46,86 +53,135 @@ public void before() {
...
+    Assert.assertTrue(UpdateCheck.getFeatureString().contains("masterAuditLog"));
+    ServerConfiguration.set(PropertyKey.MASTER_AUDIT_LOGGING_ENABLED, false);
+    Assert.assertTrue(!UpdateCheck.getFeatureString().contains("masterAuditLog"));
+  }
+
+  @Test
+  public void userAgent() throws Exception {
     String userAgentString = UpdateCheck.getUserAgentString("cluster1");
-    Assert.assertTrue(userAgentString.equals(
-        String.format("Alluxio/%s (cluster1; ProductCode:random123code; emr; ec2)",
```
Xenorith, line 128: is this not how the string ends up being anymore? i prefer the strict equals check than a regex check to be able to potentially catch misformatted entries

Xenorith (changes requested):

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
HelloHorizon: addressed

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
HelloHorizon: addressed

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
HelloHorizon: addressed

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
HelloHorizon: addressed

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
```diff
@@ -46,86 +53,148 @@ public void before() {
...
+  public void featureStringEmbeddedJournal() {
+    ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS);
+    Assert.assertFalse(UpdateCheck.getFeatureString().contains("embedded"));
+    ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.EMBEDDED);
+    Assert.assertTrue(UpdateCheck.getFeatureString().contains("embedded"));
+  }
+
+  @Test
+  public void featureStringRocks() {
+    ServerConfiguration.set(PropertyKey.MASTER_METASTORE, MetastoreType.ROCKS);
```
Xenorith, line 138: i'm confused why you can use MetastoreType.ROCKS here in the test file but not the source file, is this expected?

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
Xenorith: sorry, i dont understand the reasoning behind why we shouldn't be more efficient in writing methods in utils class?
what is the issue with having:
```
public static boolean isBooleanPropertyEnabled(String propertyKey) {
  return ServerConfiguration.getBoolean(propertyKey);
}
```
to replace all of the simple boolean check methods?

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
```diff
@@ -46,86 +53,148 @@ public void before() {
...
     Mockito.when(EC2MetadataUtils.getUserData()).thenReturn("emr_apps");
 
+    String userAgentString = UpdateCheck.getUserAgentEnvironmentString("cluster1");
+    Assert.assertTrue(userAgentString.equals("cluster1; ProductCode:random123code; emr; ec2"));
+  }
+
+  @Test
+  public void featureStringEmbeddedJournal() {
+    ServerConfiguration.set(PropertyKey.MASTER_JOURNAL_TYPE, JournalType.UFS);
+    Assert.assertFalse(UpdateCheck.getFeatureString().contains("embedded"));
```
Xenorith, line 131: readding previous comment that got marked as outdated but was not addressed. this is especially important if the `userAgent()` unit test at the bottom of this file is going to do a regex check instead of an exact match.

using `String.contains` is a weak check since another feature could have the substring embedded or have the equivalent value. how can this unit test be reformatted to catch the bad scenario where there's another feature that is also called `embedded`? same for other tests

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -79,11 +80,22 @@ public static String getLatestVersion(String clusterID, long connectionRequestTi
...
    * @param clusterID the cluster ID
-   * @return a string representation of the user's environment in the format "key1:value1, key2:
-   *         value2".
+   * @return a string representation of the user's environment in the format
+   *         "Alluxio/{ALLUXIO_VERSION} (valueA; valueB)"
    */
   @VisibleForTesting
   public static String getUserAgentString(String clusterID) throws IOException {
+    Joiner joiner = Joiner.on("; ").skipNulls();
+    String sysInfo = joiner.join(getUserAgentEnvironmentString(clusterID), getFeatureString());
```
Xenorith, line 89: `getUserAgentEnvironmentString()` always returns a non-empty string but i believe we want `getFeatureString` to return null if there are zero features enabled. correct me if i'm wrong, but a joiner on no elements is going to return `""` and so you will end up joining `"some env string"` with `""` -> `"some env string; "` instead of `"some env string"`

regardless of the behavior of joiner, we should have a unit test for this case as i had asked for previously

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -79,11 +80,22 @@ public static String getLatestVersion(String clusterID, long connectionRequestTi
...
+    return String.format("Alluxio/%s (%s)", ProjectConstants.VERSION, sysInfo);
+  }
+
+  /**
+   * @param clusterID the cluster ID
+   * @return a string representation of the user's environment in the format "docker; kubernetes"
+   */
+  @VisibleForTesting
+  public static String getUserAgentEnvironmentString(String clusterID) throws IOException {
```
Xenorith, line 98: either this method should be named `getEnvironmentString` or the below method should be `getUserAgentFeatureString` to be consistent with each other

Xenorith (changes requested):

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
HelloHorizon: It's because of the dependency of different modules. Both `UpdateCheckTest` and  `JournalType" are under alluxio-core-server-common, while `MetastoreType` is under alluxio-core-server-master. If we want to reuse the enum MetastoreType, we have to make alluxio-core-server-master as a dependency as alluxio-core-server-common, which is not expected. The workaround is using string instead, and check consistency in a unit test.

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
HelloHorizon: I am not sure I understand the difference between using the existing method compare to change to `isBooleanPropertyEnabled`. If we want to abstract a method like `isBooleanPropertyEnabled`, we can directly use `ServerConfiguration.getBoolean(propertyKey)` instead.

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
Xenorith: does the test module have different dependencies than the source module? because you are importing `import alluxio.master.metastore.MetastoreType;` in this test class. again, my confusion is why you can import it here in the test class but not in the utils class

core/server/master/src/test/java/alluxio/master/meta/UpdateCheckTest.java
HelloHorizon: Since if we have more fields, we have to mute all configurations one by one before the test. Just let me know if you have a better idea to do these tests

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
Xenorith: honestly that sounds like a better idea than mine

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
```diff
@@ -98,7 +114,39 @@ public static String getUserAgentString(String clusterID) throws IOException {
...
+    return sysInfo;
+  }
+
+  /**
+   * Get the features information.
+   *
+   * @return a list of strings representing enabled features
+   */
+  @VisibleForTesting
+  public static List<String> getUserAgentFeatureString() {
```
Xenorith, line 126: since this doesn't return a String anymore, rename to `getUserAgentFeatures()`?

Xenorith (approved):
looks good to me other than considering a small rename

core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java
HelloHorizon: addressed

HelloHorizon:

PR #12825, opened 2021-02-09
Replace lock to double check to reduce heap memory cost

maobaolong:
For the existing code of `AlluxioJniFuseFileSystem`, it use a `2048` ReadWrite lock pool as the guards for each fd, there are 3 reason why I purpose to remove it by a double check.

- Reduce the heap memory for the `LOCK_SIZE`, now the constant is `2048`, so the reason seems too weak.
- We can only have 2048 lock in the same time, if there are more other unrelated files operation, they will hang for waiting lock resource other then acquire lock.
- `2048` is a hard code here, and we can use double check to replace these unnecessary locks, this PR will clear the code.
- Synchronized block is recommend to use if we have no strong reason to use `ReentrantReadWriteLock`. We don't use the readlock at all, we also don't use fair lock. So, to simplify code, we'd better to replace it.

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
@@ -612,33 +612,38 @@ public final class AlluxioFuseFileSystem extends FuseStubFS {
       LOG.error("Cannot find fd for {} in table", path);
       return -ErrorCodes.EBADFD();
     }
-
-    int rd = 0;
-    int nread = 0;
     if (oe.getIn() == null) {
       LOG.error("{} was not open for reading", path);
       return -ErrorCodes.EBADFD();
     }
-    try {
-      oe.getIn().seek(offset);
-      final byte[] dest = new byte[sz];
-      while (rd >= 0 && nread < size) {
-        rd = oe.getIn().read(dest, nread, sz - nread);
-        if (rd >= 0) {
-          nread += rd;
+    int rd = 0;
+    int nread = 0;
+    synchronized (oe) {
+      if (!mOpenFiles.contains(fd)) {
+        LOG.error("Cannot find fd for {} in table", path);
+        return -ErrorCodes.EBADFD();
+      }
+      try {
+        oe.getIn().seek(offset);
+        final byte[] dest = new byte[sz];
+        while (rd >= 0 && nread < size) {
+          rd = oe.getIn().read(dest, nread, sz - nread);
+          if (rd >= 0) {
+            nread += rd;
+          }
         }
-      }
 
-      if (nread == -1) { // EOF
-        nread = 0;
-      } else if (nread > 0) {
-        buf.put(0, dest, 0, nread);
+        if (nread == -1) { // EOF
+          nread = 0;
+        } else if (nread > 0) {
+          buf.put(0, dest, 0, nread);
+        }
+      } catch (Throwable t) {
+        LOG.error("Failed to read file={}, offset={}, size={}", path, offset,
+            size, t);
+        return AlluxioFuseUtils.getErrorCode(t);
       }
-    } catch (Throwable t) {
-      LOG.error("Failed to read file={}, offset={}, size={}", path, offset, size, t);
-      return AlluxioFuseUtils.getErrorCode(t);
     }
-
     return nread;
   }
 
@@ -701,13 +706,14 @@ public final class AlluxioFuseFileSystem extends FuseStubFS {
     OpenFileEntry oe;
     final long fd = fi.fh.get();
     oe = mOpenFiles.getFirstByField(ID_INDEX, fd);
-    mOpenFiles.remove(oe);
-    if (oe == null) {
+    if (oe == null || !mOpenFiles.remove(oe)) {
       LOG.error("Cannot find fd for {} in table", path);
       return -ErrorCodes.EBADFD();
     }
     try {
-      oe.close();
+      synchronized (oe) {
+        oe.close();
+      }
     } catch (IOException e) {
       LOG.error("Failed closing {} [in]", path, e);
     }
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
@@ -619,7 +619,7 @@ public final class AlluxioFuseFileSystem extends FuseStubFS {
     int rd = 0;
     int nread = 0;
     synchronized (oe) {
-      if (!mOpenFiles.contains(fd)) {
+      if (!mOpenFiles.contains(oe)) {
         LOG.error("Cannot find fd for {} in table", path);
         return -ErrorCodes.EBADFD();
       }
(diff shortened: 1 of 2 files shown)


integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
```diff
@@ -701,13 +706,14 @@ private int releaseInternal(String path, FuseFileInfo fi) {
...
-    mOpenFiles.remove(oe);
-    if (oe == null) {
+    if (oe == null || !mOpenFiles.remove(oe)) {
       LOG.error("Cannot find fd for {} in table", path);
       return -ErrorCodes.EBADFD();
     }
     try {
-      oe.close();
+      synchronized (oe) {
+        oe.close();
```
gpang, line 715: Do we need to check here that `oe` is still in the index?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -401,18 +400,22 @@ public int release(String path, FuseFileInfo fi) {
 
   private int releaseInternal(String path, FuseFileInfo fi) {
     long fd = fi.fh.get();
-    try (LockResource r1 = new LockResource(mFileLocks.get(fd).writeLock())) {
+    try {
       FileInStream is = mOpenFileEntries.remove(fd);
       FileOutStream os = mCreateFileEntries.remove(fd);
       if (is == null && os == null) {
```
gpang, line 406: Hrmmm, should we treat each input and output streams separately/independently? Then, we could just simply do something like
```
if (is == null) {
  return error;
} else {
  synchronized (is) {
    is.close();
  }
}
```

We could even have a helper method which takes a closable.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -401,18 +400,22 @@ public int release(String path, FuseFileInfo fi) {
...
         LOG.error("Cannot find fd {} for {}", fd, path);
         return -ErrorCodes.EBADFD();
       }
       if (is != null) {
-        is.close();
+        synchronized (is) {
+          is.close();
```
gpang, line 412: Do we have to check if `is` is in the index? (same question for `os` below)

gpang (commented):
@maobaolong Thanks! I left a few comments.

However, is the lock issue a big deal? Is having ~2000 locks a big strain on the heap? Could you add to the PR description about what the issue is?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: if `mOpenFiles.remove(oe)` return true, it means other threads cannot remove successfully and return false, so that other threads will return with `-ErrorCodes.EBADFD()`.

And for the lucky thread which remove the oe successfully and return true, it can step into the synchronized block if no other threads was reading from the inputStream, so `oe.close` can be executed only one time.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: As the previous logic of this method, it return `-ErrorCodes.EBADFD()` only if both is and os are all `null`.

So, I don't know well about how the write the code like your given template, would you please give me some more detail or tips? Thank you very mush!

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
gpang: sounds good

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
gpang: Ok, we don't have to change the existing behavior. I was just wondering if the existing behavior is what we want or not.

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
gpang: What about this?

gpang (commented):
@maobaolong

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
maobaolong: I think we don't need to check oe still in the index here, even, we can ensure that oe is not in the index certainly. Because `mOpenFiles.remove(oe)` must return true, otherwise, it should return `-ErrorCodes.EBADFD()`.

gpang (approved):
LGTM

maobaolong:
@gpang Thanks for your review.

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
```diff
@@ -701,13 +706,14 @@ private int releaseInternal(String path, FuseFileInfo fi) {
     OpenFileEntry oe;
     final long fd = fi.fh.get();
     oe = mOpenFiles.getFirstByField(ID_INDEX, fd);
-    mOpenFiles.remove(oe);
-    if (oe == null) {
+    if (oe == null || !mOpenFiles.remove(oe)) {
```
LuQQiu, line 709: Does the `mOpenFiles.remove(oe)` need to be included in the synchronized(oe)?
I am not 100% sure if the following scenario will happen:
1) read -> enter the synchronized(oe) block, check the mOpenFiles contains the oe and start reading
2) during the reading, release() is called and oe is removed from mOpenFiles.

The above case seems unlikely to happen, but to be safe, can we move the logic into the synchronized block? 

what's your thoughts?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -401,18 +400,22 @@ public int release(String path, FuseFileInfo fi) {
 
   private int releaseInternal(String path, FuseFileInfo fi) {
     long fd = fi.fh.get();
-    try (LockResource r1 = new LockResource(mFileLocks.get(fd).writeLock())) {
+    try {
       FileInStream is = mOpenFileEntries.remove(fd);
```
LuQQiu, line 404: Same comment as above

LuQQiu (commented):
Thanks for improving the code concurrency! Left one minor comment, otherwise LGTM

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseFileSystem.java
maobaolong: I think the  above scenario will happen, but it is safe, because even `oe` is removed, read thread can also read the data from alluxio if the read thread enter the synchronized block. For the release thread, it can close the stream only after enter the synchronized block.

LuQQiu: Make sense, then i am fine with keeping the original logics

LuQQiu (approved):
LGTM, thanks!

LuQQiu:

maobaolong:
@LuQQiu Thank you for your review, suggestion and merge this PR.

PR #12826, opened 2021-02-09
Fix wrong log for fuse chown method

maobaolong:


jiacheliu3 (approved):
LGTM, @maobaolong thanks for the fix! A side note, I see quite some shared code between these 2 classes, can they be refactored? The refactor does not need to happen in this PR.

jiacheliu3:

PR #12828, opened 2021-02-09
Clean up incomplete files if sync is interrupted

yuzhu:
It is possible for syncMetadata process to be interrupted due to user cancellation or IOExceptions. 
We need to recover from incomplete files if we happen to have created the file but not completed the file.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -824,6 +824,8 @@ public class InodeSyncStream {
         LOG.debug("File not created yet, do not need to clean up: {}", ex.toString());
       } catch (DirectoryNotEmptyException ex) {
         LOG.warn("Trying to remove a directory, expecting a file: {}", ex.toString());
+      } catch (Throwable ex) {
+        e.addSuppressed(ex);
       }
       throw e;
     }
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -821,7 +821,7 @@ public class InodeSyncStream {
               .mergeFrom(DeletePOptions.newBuilder().setAlluxioOnly(true)));
         }
       } catch (FileDoesNotExistException ex) {
-        LOG.debug("File not created yet, do not need to clean up: {}", ex.toString());
+        LOG.info("File not created yet, do not need to clean up: {}", ex.toString());
       } catch (DirectoryNotEmptyException ex) {
         LOG.warn("Trying to remove a directory, expecting a file: {}", ex.toString());
       } catch (Throwable ex) {
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -40,7 +40,6 @@ import alluxio.master.file.contexts.GetStatusContext;
 import alluxio.master.file.contexts.LoadMetadataContext;
 import alluxio.master.file.contexts.SetAttributeContext;
 import alluxio.master.file.meta.Inode;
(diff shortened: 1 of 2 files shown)


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -810,6 +812,20 @@ static void loadFileMetadataInternal(RpcContext rpcContext, LockedInodePath inod
...
+    } catch (Throwable e) {
+      // Clean up possibly incomplete files due to any exceptions
+      try (LockedInodePath writeLockedPath = inodePath.lockFinalEdgeWrite()) {
+        FileInfo info = fsMaster.getFileInfoInternal(writeLockedPath);
+        if (!info.isCompleted()) {
+          fsMaster.deleteInternal(rpcContext, inodePath, DeleteContext
+              .mergeFrom(DeletePOptions.newBuilder().setAlluxioOnly(true)));
+        }
+      } catch (FileDoesNotExistException ex) {
+        LOG.debug("File not created yet, do not need to clean up: {}", ex.toString());
```
ZacBlanco, line 824: I'm not sure if DEBUG is good here...I think it's already bad if we've reached this `catch` statement, and then only logging `debug` on the exception within the catch seems like probably something we want in the logs regardless? I would opt for warn

The other log statement below is also a `warn`

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -810,6 +812,20 @@ static void loadFileMetadataInternal(RpcContext rpcContext, LockedInodePath inod
...
+      } catch (DirectoryNotEmptyException ex) {
```
ZacBlanco, line 825: Are there any other exceptions besides these two we should consider catching? I feel `deleteInternal` could fail in many ways, which would cause us to not throw the original exception and possibly cause a loss of the true error information.

I suggest add another `catch` here and use `e.addSuppressed()` in the case of any other failures

ZacBlanco (changes requested):

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: This exception can happen in normal operation, if the end user cancels an RPC, causes an InterruptedException,  before the file is even created.  hence the DEBUG here. I am ok with upgrading it to an INFO. The other log statement is more serious, hence the WARN.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: done

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -810,6 +812,22 @@ static void loadFileMetadataInternal(RpcContext rpcContext, LockedInodePath inod
...
+      try (LockedInodePath writeLockedPath = inodePath.lockFinalEdgeWrite()) {
+        FileInfo info = fsMaster.getFileInfoInternal(writeLockedPath);
+        if (!info.isCompleted()) {
+          fsMaster.deleteInternal(rpcContext, inodePath, DeleteContext
+              .mergeFrom(DeletePOptions.newBuilder().setAlluxioOnly(true)));
+        }
+      } catch (FileDoesNotExistException ex) {
+        LOG.info("File not created yet, do not need to clean up: {}", ex.toString());
+      } catch (DirectoryNotEmptyException ex) {
+        LOG.warn("Trying to remove a directory, expecting a file: {}", ex.toString());
```
ZacBlanco, line 826: I am also wondering whether these should be suppressed. I guess at least there is logging, but I guess it's not the worst if we don't since we're at least logging them

ZacBlanco (approved):

yuzhu:
Offline discussion lead to a cleaner alternative approach, will open a separate PR

PR #12829, opened 2021-02-09
Add support for S3 Northbound API tags

bradyoo:
This adds support for S3 Northbound API for tags enough to be able to be used in S3Browser.

DeleteBucketTagging and DeleteObjectTagging is not in this PR because S3Browser doesn't need it but can be added.


core/server/proxy/src/main/java/alluxio/proxy/s3/tag/Tagging.java
```diff
@@ -0,0 +1,67 @@
...
+   */
+  public Tagging() {}
+
+  /**
+   * Constructs a Tagging object.
+   * @param tagMap map of tag keys to tag values
+   */
+  public Tagging(Map<String, String> tagMap) {
+    final ArrayList<String> keys = new ArrayList<>(tagMap.keySet());
+    Collections.sort(keys);
```
gpang, line 40: Instead of this, can we do something like:
```java
mTagSet = new TagSet(
    tagMap.entrySet().stream().sorted(Comparator.comparing(e -> e.getKey())).collect(Collectors.toList()));
```
(I don't know if this compiles...)

Also, the tags have to be sorted?

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+
+/**
+ * Helper class for handling tags according to https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketTagging.html
+ * and https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectTagging.html.
+ */
+public class TagHelper {
+  private static final Logger LOG = LoggerFactory.getLogger(S3RestServiceHandler.class);
+
+  private static final String TAG_FOLDER = "/_TAG";
```
gpang, line 42: Can you add some comments describing the protocol of how tagging works?

gpang (changes requested):
@bradyoo Thanks! I left a few questions

bradyoo:
Added tests and addressed feedback

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagSet.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class TagSet {
+  private List<Tag> mTags;
+
+  /**
+   * Default Constructor used for XML deserialization.
+   */
+  public TagSet() {}
+
+  TagSet(List<Tag> tags) {
```
gpang, line 31: Does this require the list to be ordered? The other code is ordering the tags when creating a `TagSet`.

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
```diff
@@ -231,6 +236,14 @@ public ListBucketResult call() throws S3Exception {
         final FileSystem fs = getFileSystem(authorization);
         checkPathIsAlluxioDirectory(fs, path);
 
+        if (tagging.isEmpty()) {
```
gpang, line 239: I'm not sure how this works. If the request does set the tagging parameter, what does this tagging variable become?

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
```diff
@@ -190,6 +193,7 @@ public ListAllMyBucketsResult call() throws S3Exception {
    * @param prefixParam the optional prefix param
    * @param delimiterParam the optional delimiter param
    * @param maxKeysParam the optional max keys param
+   * @param tagging the optional tagging param to determine if this is GetBucketTagging
    * @return the response object
```
gpang, line 197: maybe we should add here that it can return a listbuckets result or a tagging result.

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
```diff
@@ -0,0 +1,120 @@
...
+      final FileInStream fis = fs.openFile(new AlluxioURI(tagPath));
+      return new ObjectMapper().readValue(fis, Map.class);
+    } catch (FileDoesNotExistException e) {
+      return Collections.emptyMap();
+    }
+  }
+
+  private static String getTagPath(FileSystem fs, String path)
+      throws IOException, AlluxioException {
+    final URIStatus status = fs.getStatus(new AlluxioURI(path));
```
gpang, line 113: Will this expensive to perform an RPC to convert paths?

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Helper class for handling tags according to https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketTagging.html
+ * and https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectTagging.html.
+ */
+public class TagHelper {
+  private static final Logger LOG = LoggerFactory.getLogger(S3RestServiceHandler.class);
+
+  /**
+   * all tag information will go into TAG_FOLDER (/_TAG), inside of the folder, every file in the
```
gpang, line 46: This folder is in the root of the alluxio filesystem? I think we have a special folder that we can use for alluxio specific things. Lu should have experience with that.

Also, if the root UFS is CACHE_THROUGH, is that ok?

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
```diff
@@ -0,0 +1,120 @@
...
+      fs.delete(tagPathURI);
+    } catch (FileDoesNotExistException e) {
+      // ignore
+    }
+
+    StringWriter writer = new StringWriter();
+    IOUtils.copy(is, writer, Charset.defaultCharset());
+    String theString = writer.toString();
+
+    LOG.info(theString);
```
gpang, line 77: remove debugging output?

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
```diff
@@ -821,6 +823,43 @@ public void completeMultipartUpload() throws Exception {
...
+
+    final HashMap<String, String> tag0 = new HashMap<>();
+    tag0.put("abc", "def");
+    tag0.put("ghi", "jkl");
+    putObjectTagRestCall("test0", createTagXml(tag0));
+
+    final String tag0result = getObjectTagRestCall("test0");
+    assertThat(tag0result, new StringContains("<Key>abc</Key><Value>def</Value>"));
+    assertThat(tag0result, new StringContains("<Key>ghi</Key><Value>jkl</Value>"));
+    assertThat(getObjectTagRestCall("test1"), new StringContains("<TagSet/>"));
```
gpang, line 842: Is this string NOT present when tags are returned?

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
```diff
@@ -821,6 +823,43 @@ public void completeMultipartUpload() throws Exception {
...
+  public void bucketTags() throws Exception {
+    createBucketRestCall("test0");
+    createBucketRestCall("test1");
+
+    assertThat(getObjectTagRestCall("test0"), new StringContains("<TagSet/>"));
+    assertThat(getObjectTagRestCall("test1"), new StringContains("<TagSet/>"));
+
+    final HashMap<String, String> tag0 = new HashMap<>();
+    tag0.put("abc", "def");
+    tag0.put("ghi", "jkl");
```
gpang, line 836: What are valid tags in s3? can we have empty keys or values? Can there be duplicate keys?

Also, is there a test to update the existing tags (and add new ones)?

gpang (changes requested):
@bradyoo Thanks! I left some comments.

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagSet.java
bradyoo: Nothing actually requires it to be ordered. It's only ordered for the display to the user to be ordered.

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
bradyoo: It becomes the default value which is "null". The reason why it's setup this way is because the way this gets sent from the client is ?tagging without any value.

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
bradyoo: This is certainly going to be cheaper than opening a file which will need to happen later.

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
bradyoo: Yeah in particular, <TagSet/> means that TagSet tag is immediately closed. Which wouldn't happen if there are any members inside a TagSet.

tests/src/test/java/alluxio/client/rest/S3ClientRestApiTest.java
bradyoo: there can't be duplicate keys. (it wouldn't error but result is unexpected). Test to update the existing tags and add new ones isn't too useful here because what the api does isn't an update/remove. It just replaces the entire existing tags with new ones.

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
bradyoo: for the root UFS being CACHE_THROUGH. How is this any different from the problem with _internal/ for catalog service.

Also: @LuQQiu

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
bradyoo: Modifying all summaries to mention the modal nature of all these apis.

bradyoo:
feedback

bradyoo:
feedback addressed.

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
LuQQiu: We have a `alluxio.underfs.s3.directory.suffix` to use a special zero-byte object name for the empty folder, besides that, i don't remember we have any special folder for other usage.

gpang: O whoops, I think it was the tmp file issue that @HelloHorizon  worked on. I think there is a special folder for Alluxio to use internally.

gpang (commented):
@bradyoo do we want these tags to be in internal files, or just maintained on the master? I can imagine those are the 2 alternatives for these tags.

core/server/proxy/src/main/java/alluxio/proxy/s3/tag/TagHelper.java
HelloHorizon: The tmp file folder for persistence is the folder under UFS mount point `alluxio.util.io.PathUtils#getPersistentTmpPath`. And we also have a similar implementation for block in `alluxio.worker.BlockUtils#getUfsBlockPath`.

bradyoo:
feedback applied

bradyoo:
I did as suggested but I'm not sure if this is a good idea because I believe I want these tag information to be persisted to the underlying storage and don't want to be dropped by regex patterns.

HelloHorizon:
@bradyoo We don't have any cleanup mechanism for this regex for now I think, and since I don't have any context for tag information, I am not sure if you want to do it. For persistence, the main reason for us to create a temp folder under each mount point is to avoid the clean up failure for the intermediate persistent file.

HelloHorizon:
FYI, the main reason for us to create tmp folder under each mount point for persist file is to avoid compute frameworks pick up tmp persistent file. And I don't think we have any regex related mechanism for tmp folder now.

bradyoo:
@gpang ?

gpang:
My main question was that we already have a "special" directory that Alluxio uses for Alluxio-internal purposes (tmp files, ufs fallback). Can we just make a sub-directory for the tags, under that "special" directory?

bradyoo:
HelloHorizon's method is on each mount point and the folder is literally called ".alluxio_ufs_persistence". I don't think that's a good idea.

HelloHorizon:
We definitely can add one more layer for the "special" directory to include all usage for Alluxio's internal purposes. The name ".alluxio_ufs_persistence" is meaningless, it just follows the same naming logic as `getUfsBlockPath`. We can change it if you have a better idea of these directories.

bradyoo:
I certainly don't think we should be doing that now though.

bradyoo:
I can put it in something like generic _internal/_TAGS so that we can make it easier though.

bradyoo:
@gpang it is still not clear what you are asking and this is a release blocker :( Should I just ask someone else?

calvinjia (commented):
Have some high level questions/comments - 
1. Should/can tags be persisted to the underlying storage?
2. What happens to the tags if the file is renamed / mount is changed?
3. IMO a proxy side implementation of tags seems brittle/not scalable. What was the reason against doing a server side implementation?

bradyoo:
1. How would the tags survive a restart of the server if it is not persisted to the underlying storage?
2. This is a good question. The reason why this is fine for S3 Northbound API is because rename/mount isn't a concept that exists in S3 API.
3. Server side support would likely mean that we would be supporting this feature for any operation that we can do on the filesystem and may conflict with any future implementations just for the sake of supporting S3 APIs. This isolates any issues to S3 better.

calvinjia:
1. I meant persisted as part of the UFS metadata. Currently if we mount S3 and there already is object metadata, we will not surface that?
2. The S3 API is just another API to access the namespace, I don't think we can assume that once part of the namespace has been accessed with the S3 API, no other APIs (ie. ones that can rename) will access it.
3. I imagine this would be something like (or even the same as) the extended attributes we have as part of the file system API?

bradyoo:
Will put it up again later using ExtendedAttribute instead.

Issue #12830, opened 2021-02-09
Support AlluxioFuse embedded in AlluxioWorker


LuQQiu:
**Is your feature request related to a problem? Please describe.**
Performance considerations are the most important consideration that forces us to consider embedding POSIX API into AlluxioWorkers. In multi-threading AI scenarios running against fuse mount point, Alluxio POSIX API integration introduces non-trivial overheads compare to reading from local files directly. Part of those overheads are un-avoided like we need to ship data from/to remote workers, but a non-trivial amount of overheads are introduced by excessive RPC calls to AlluxioMaster and AlluxioWorkers. After analysis, many of those RPCs can be avoided if POSIX API is embedded inside the AlluxioWorker process. Reducing the RPC needed in core paths has been proved to significantly improve the performance of important AI scenarios. 

**Describe the solution you'd like**
Embedded Fuse in Worker process

**Describe alternatives you've considered**
On the Fuse client side, many workarounds have been made to bypass some RPCs between standalone AlluxioFuse and collated AlluxioWorker. Those workarounds only target some specific scenarios and are not guaranteed to be safe.

**Urgency**
This feature is important to improve the AlluxioFuse performance in a safe and clean way.

**Additional context**
[The embedded Fuse in Worker design doc](https://docs.google.com/document/d/1hdI5LPAPN-NPEYhcxdw47yGyP4GdyzJRf98oawJx0rA/edit?usp=sharing) has more design details. Feel free to comment on the design docs and provide suggestions.

PR #12837, opened 2021-02-10
Modify Fuse code for convenient embedded into AlluxioWorker

LuQQiu:
PR1 for #12830 
This PR modifies the Fuse side code for workers to launch Fuse inside the worker processes as non-blocking operations.
`FuseMountInfo` is responsible for providing the Fuse mount information when mounting embedded fuse and providing the existing fuse mount information to clients.
`FuseUnmountable` interface is added for workers to unmount the running Fuse applications.


integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
```diff
@@ -109,25 +109,47 @@ public static void main(String[] args) {
...
               opts.getMountPoint(), fuseOpts.toArray(new String[0]));
-          fuseFs.mount(true, opts.isDebug(), fuseOpts.toArray(new String[0]));
+          fuseFs.mount(blocking, opts.isDebug(), fuseOpts.toArray(new String[0]));
         } catch (FuseException e) {
-          LOG.error("Failed to mount {}", opts.getMountPoint(), e);
           // only try to umount file system when exception occurred.
           // jni-fuse registers JVM shutdown hook to ensure fs.umount()
           // will be executed when this process is exiting.
           fuseFs.umount();
+          throw new IOException(String.format("Failed to mount %s", opts.getMountPoint()), e);
```
gpang, line 150: Will this print out the local machine mount point as well as the alluxio directory?

LuQQiu: Usually for fuse mount, the identifier is the local machine mount point which is unique. Several mount point can point to the same alluxio directory. 
but i can add the alluxio directory to this error message.

LuQQiu:
@apc999 @gpang PTAL the fuse code changes of embeddedFuse impl, thanks!

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
```diff
@@ -109,25 +109,48 @@ public static void main(String[] args) {
           + "Proceed with local configuration for FUSE: {}", e.toString());
     }
     conf = fsContext.getClusterConf();
-    final AlluxioFuseOptions opts = parseOptions(args, conf);
+    final FuseMountInfo opts = parseOptions(args, conf);
     if (opts == null) {
       System.exit(1);
     }
+    try {
+      launchFuse(fsContext, opts, true);
```
apc999, line 117: return value not used? Then do we need a return a `FuseUnmountable ` for this method?
And also do we even need `FuseUnmountable` since it is not used?

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
```diff
@@ -109,25 +109,48 @@ public static void main(String[] args) {
...
+   * Launches Fuse application.
+   *
+   * @param fsContext file system context for Fuse client to communicate to servers
+   * @param opts the fuse mount options
+   * @param blocking whether the Fuse application is blocking or not
+   * @return the Fuse application handler for future unmount operation
+   */
+  public static FuseUnmountable launchFuse(FileSystemContext fsContext,
+      FuseMountInfo opts, boolean blocking) throws IOException {
+    if (opts == null) {
```
apc999, line 133: `Preconditions.checkNotNull`

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
```diff
@@ -109,25 +109,48 @@ public static void main(String[] args) {
...
     conf = fsContext.getClusterConf();
-    final AlluxioFuseOptions opts = parseOptions(args, conf);
+    final FuseMountInfo opts = parseOptions(args, conf);
     if (opts == null) {
       System.exit(1);
     }
+    try {
+      launchFuse(fsContext, opts, true);
+    } catch (IOException e) {
+      LOG.error(e.getMessage());
```
apc999, line 119: do we need to `exit(-1)`

apc999 (changes requested):

integration/fuse/src/main/java/alluxio/fuse/FuseMountInfo.java
```diff
@@ -19,7 +19,7 @@
  * Convenience class to pass around Alluxio-FUSE options.
  */
 @ThreadSafe
-public final class AlluxioFuseOptions {
+public final class FuseMountInfo {
```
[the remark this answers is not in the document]
LuQQiu: Changed back to `FuseMountOptions`, use `FuseMountInfo` before to support user CLI get the fuse mount information from the AlluxioWorker

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
LuQQiu: Remove the `FuseUnmountable` because it's not needed anymore since we decide to remove the support for runtime mount/unmount

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
LuQQiu: Changed, thanks

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
LuQQiu: Add the system exit (-1)

LuQQiu:
@apc999 addressed the comments, PTAL, thanks

apc999 (approved):

LuQQiu:

PR #12838, opened 2021-02-11
Extract the worker read handler logics for general usage

LuQQiu:
PR2 for #12830
This PR extracts the `getBlockReader` and `cleanBlockReader` logics from `Grpc BlockWorkerImpl -> BlockReadHandler` and puts in `DefaultBlockWorker` for both GRPC client and worker internal usage. 

`BlockReadRequest` related classes are put to `alluxio-base` package for usage shared by clients and workers.


LuQQiu:
@apc999 PTAL the worker read logic extraction, thanks!

core/server/worker/src/test/java/alluxio/worker/grpc/BlockReadHandlerTest.java
```diff
@@ -30,16 +31,28 @@
...
+  private DefaultBlockWorker mBlockWorker;
   private BlockReader mBlockReader;
 
   @Before
   public void before() throws Exception {
     ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.ADVANCED);
-    mBlockWorker = mock(BlockWorker.class);
+
+    // Use PowerMockito to mock final class and able to test the method contents
+    mBlockWorker = PowerMockito.mock(DefaultBlockWorker.class);
```
apc999, line 51: if  needed, we can make `DefaultBlockWorker` non final to avoid `PowerMockito`

core/server/worker/src/main/java/alluxio/worker/block/BlockWorker.java
```diff
@@ -423,6 +424,28 @@ void closeUfsBlock(long sessionId, long blockId)
...
+
+  /**
+   * Cleans data reader and related blocks after using the block reader obtained
+   * from {@link #getBlockReader(BlockReadRequest)}.
+   *
+   * @param sessionId the session id which used for getting the block reader
+   * @param blockId the block id this block reader belongs to
+   * @param reader the to be cleaned block reader
+   */
+  void cleanBlockReader(long sessionId, long blockId, BlockReader reader)
```
apc999, line 446: nit: better to put read as the first arg as this method is about `cleanBlockReader`

core/server/worker/src/main/java/alluxio/worker/block/BlockWorker.java
```diff
@@ -423,6 +424,28 @@ void closeUfsBlock(long sessionId, long blockId)
...
       WorkerOutOfSpaceException;
 
+  /**
+   * Gets the block reader to read from Alluxio block or UFS block.
+   * This operation must be paired with {@link #cleanBlockReader(long, long, BlockReader)}.
+   *
+   * @param request the block read request
+   * @return a block reader to read data from
+   */
+  BlockReader getBlockReader(BlockReadRequest request) throws IOException,
```
apc999, line 434: I would suggest to pass `long sessionId, long blockId` instead of `BlockReadRequest`, to keep `getBlockReader` and `cleanBlockReader` in parity. Alternatively, pass `BlockReadRequest` to `cleanBlockReader`

core/server/worker/src/test/java/alluxio/worker/grpc/BlockReadHandlerTest.java
```diff
@@ -30,16 +31,28 @@
 import io.netty.util.ResourceLeakDetector;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
 
+@RunWith(PowerMockRunner.class)
```
apc999, line 40: if we remove `final` for `DefaultBlockWorker` , can we avoid `PowerMockRunner`?
PowerMock introduces a lot limitation and let's avoid to use it

apc999 (approved):
LGTM modulo minor comments

core/server/worker/src/test/java/alluxio/worker/grpc/BlockReadHandlerTest.java
LuQQiu: Add a TODO for making DefaultBlockWorker non-final and avoid PowerMockito.
The test changes will be given in another PR to unblock the current implementation

core/server/worker/src/main/java/alluxio/worker/block/BlockWorker.java
LuQQiu: Moved BlockReader to be the first parameter, thanks

core/server/worker/src/main/java/alluxio/worker/block/BlockWorker.java
LuQQiu: Agreed that we need to keep the two methods consistent.
The `BlockReadRequest` contains too much information, like isPromoted, isPersisted, UfsBlockOptions, etc. Putting them all in the constructor is kind of messy. I changed to pass `BlockReadRequest` to `cleanBlockReader`.

core/server/worker/src/test/java/alluxio/worker/grpc/BlockReadHandlerTest.java
LuQQiu: Add a TODO(lu) in the code and will give another PR to address the PowerMockRunner issue of the tests, thanks!

LuQQiu: I took a deeper look, DefaultBlockWorker is a complicated class with many other classes involved, it's not isolated like the previous BufferCachingDataReader class. The `BlockWorkerTest` itself is using PowerMock runner. Would suggest changing the whole `BlockWorkerTest` as well. `BlockReadHandlerTest` is a small extend of the `BlockWorkerTest`. This test may need more careful design and think through.

core/server/worker/src/test/java/alluxio/worker/grpc/BlockReadHandlerTest.java
LuQQiu: I took a deep look, I can make `DefaultBlockWorker` not final and use `extends DefaultBlockWorker`, but then in order to create the `MockDefaultBlockWorker`, i need to pass in (mock) classes including `BlockMasterClient.class, BlockMasterClientPool,BlockStore, FileSystemMasterClient,Sessions`. Those classes are all final. 

So in order to avoid using powerMockito, all those classes need to change to non-final. Many other mock classes may be needed. And there are maybe more to change in the next layer.

The whole `BlockWorkerTest` relies on PowerMockito, 
```
@RunWith(PowerMockRunner.class)
@PrepareForTest({BlockMasterClient.class, BlockMasterClientPool.class, FileSystemMasterClient.class,
    BlockHeartbeatReporter.class, BlockMetricsReporter.class, BlockMeta.class,
    BlockStoreLocation.class, StorageDir.class, ServerConfiguration.class, UnderFileSystem.class,
    BlockWorker.class, Sessions.class})
```
I couldn't find an easy way now to reconstruct the test to not using PowerMockito.

Would be better to reconsider how to test the whole `BlockWorker`. `BlockReadHandlerTest` can then extend the logic and change.

apc999 (approved):

LuQQiu:

Issue #11729, opened 2020-07-08
Limit the logging for MasterWorkerInfo


jiacheliu3:
**Is your feature request related to a problem? Please describe.**
When a `MasterWorkerInfo` is converted to string, all its blocks are converted to a list of block IDs. When a `MasterWorkerInfo` has too many blocks this wastes resource and take up a very large block in the logs. For example a MasterWorkerInfo of a good 400 blocks can fill my 15-inch screen. I've seen a 10MB log file mostly filled by one `MasterWorkerInfo` object.

**Describe the solution you'd like**
Limit the printout for a `MasterWorkerInfo`.

**Describe alternatives you've considered**

**Urgency**
LOW

**Additional context**
The drawback is we lose the full knowledge about the omitted blocks in the logs. One option is to print the full list in DEBUG and chunk the output in normal mode, like https://github.com/Alluxio/alluxio/pull/11514

jiacheliu3:
@gpang I wanted to discuss with you on this. Do you think it's okay to chunk the full list of blocks in our logs? Do we rely on the logs to trace the existence of certain blocks?

gpang:
What do you mean by "chunk the list"?

Maybe a simple solution is to print the prefix (and the count) of the list at INFO level, and the rest at DEBUG level, so we have access to the full list if we need it.

jiacheliu3:
@gpang I think I meant "truncate" the list. See https://github.com/Alluxio/alluxio/pull/12841, the `blockCount` field contains the count. When log level is above DEBUG, only the prefix of 100 blocks are printed. And when the log level is DEBUG, all the blocks are printed.

jiacheliu3:
Resolved by the PR

PR #12841, opened 2021-02-11
Limit the printed blocks when converting MasterWorkerInfo to string

jiacheliu3:
Addresses https://github.com/Alluxio/alluxio/issues/11729

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
+++ b/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
@@ -379,8 +379,8 @@ public final class MasterWorkerInfo {
   @Override
   public String toString() {
     Collection<Long> blocks = mBlocks;
+    String blockFieldName = "blocks";
     // We truncate the list of block IDs to print, unless it is for DEBUG logs
-    String blockFieldName = "all-blocks";
     if (!LOG.isDebugEnabled() && mBlocks.size() > BLOCK_SIZE_LIMIT) {
       blockFieldName = "blocks-truncated";
       blocks = mBlocks.stream().limit(BLOCK_SIZE_LIMIT).collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -375,10 +378,19 @@ public boolean hasLostStorage() {
...
+    if (!LOG.isDebugEnabled()) {
+      blocks = (mBlocks.size() < BLOCK_SIZE_LIMIT) ? mBlocks :
+              mBlocks.stream().limit(BLOCK_SIZE_LIMIT).collect(Collectors.toList());
+    }
     return MoreObjects.toStringHelper(this).add("id", mId).add("workerAddress", mWorkerAddress)
-        .add("capacityBytes", mCapacityBytes).add("usedBytes", mUsedBytes)
-        .add("lastUpdatedTimeMs", mLastUpdatedTimeMs).add("blocks", mBlocks)
-        .add("lostStorage", mLostStorage).toString();
+            .add("capacityBytes", mCapacityBytes).add("usedBytes", mUsedBytes)
+            .add("lastUpdatedTimeMs", mLastUpdatedTimeMs).add("blocks", mBlocks)
```
gpang, line 389: This adds `blocks` twice?

Also, if the list was actually truncated, can we label it like `blocks-truncated` or something that distinguishes it from the full blocks list?

gpang (changes requested):
@jiacheliu3 Thanks!

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: oops i overlooked that field

Yea makes sense to me `blocks-truncated` would be a better name.

gpang (approved):
LGTM

jiacheliu3:

PR #12844, opened 2021-02-12
Support logLevel for job master/worker

jiacheliu3:
`logLevel` command currently only works for master and workers. This change enables the command to set log level on job master/workers.

### DIFF
--- a/shell/src/main/java/alluxio/cli/LogLevel.java
+++ b/shell/src/main/java/alluxio/cli/LogLevel.java
@@ -16,14 +16,18 @@ import alluxio.Constants;
 import alluxio.annotation.PublicApi;
 import alluxio.client.block.BlockWorkerInfo;
 import alluxio.client.file.FileSystemContext;
+import alluxio.client.job.JobMasterClient;
 import alluxio.conf.AlluxioConfiguration;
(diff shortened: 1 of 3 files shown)


shell/src/main/java/alluxio/cli/LogLevel.java
```diff
@@ -192,6 +236,7 @@ private static void setLogLevel(final TargetInfo targetInfo, String logName, Str
     if (level != null) {
       uriBuilder.addParameter(LEVEL_OPTION_NAME, level);
     }
+    LOG.info("Setting log level on %s%n", uriBuilder.toString());
```
gpang, line 239: Did you want this for log, or for System.out?

gpang (commented):
@jiacheliu3 Thanks!

shell/src/main/java/alluxio/cli/LogLevel.java
jiacheliu3: I think users need not know the exact URL unless something goes wrong, then they can find it in the log. So I'd like this to be in the log.

gpang (approved):
LGTM

jiacheliu3:

PR #12855, opened 2021-02-16

yuzhu:
Before this change, syncMetadata writes an inode journal entry and then two additional updateInode entries to complete the file. This combines three entries into one so that we do not have a failure mode where the first inode entry is inserted and incomplete files are left behind.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/CompleteFileResult.java
+++ b/core/server/master/src/main/java/alluxio/master/file/CompleteFileResult.java
@@ -1,0 +1,49 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


yuzhu:
@bf8086 @calvinjia still cleaning up the PR but would like early feedback. The general idea is to reuse as much of the current completeFile and createFile code as possible.  So createFile now has an optional completeFileContext passed in, and can get completeFile results.  Unfortunately, it is not as clean as i hoped because now InodeTree has a dependency on the FileSystemMaster to complete the file.

bf8086:
> @bf8086 @calvinjia still cleaning up the PR / fix tests but would like some early feedback. The general idea is to reuse as much of the current completeFile and createFile code as possible. So createFile now has an optional completeFileContext passed in, and can get completeFile results. Unfortunately, it is not as clean as i hoped because now InodeTree has a dependency on the FileSystemMaster to complete the file.

I agree we should avoid having InodeTree call back to FSM to handle complete files. Maybe we can have the `RpcContext` allow an optional "preprocessor" that will merge the create entry with complete/update entries as they are appended?

yuzhu:
closing this and open #12875

PR #12857, opened 2021-02-16
Make libjnifuse packaged in distribution

apc999:
This PR packages libjnifuse (both .so and . dylib) into `lib/` dir, and add this dir to library path of fuse process by default.
benefits:
- Users can use libjnifuse without setting java library. When binary tarball is downloaded, JNIFuse is configured and ready to use. There will be no difference for Mac and Linux users.
- No need to separately copy libjnifuse.so in Dockerfile.fuse --- it will be handled as part of Alluxio tarball

@maobaolong @cheyang , FYI


integration/fuse/bin/alluxio-fuse
```diff
@@ -18,21 +18,9 @@ get_env() {
   . ${ALLUXIO_LIBEXEC_DIR}/alluxio-config.sh
 
   ALLUXIO_FUSE_JAR=${SCRIPT_DIR}/../target/alluxio-integration-fuse-${VERSION}-jar-with-dependencies.jar
-  FUSE_MAX_WRITE=131072
```
apc999, line 21: @cheyang  is this used any where?

integration/fuse/bin/alluxio-fuse
```diff
@@ -18,21 +18,9 @@ get_env() {
...
   CLASSPATH=${CLASSPATH}:${ALLUXIO_FUSE_JAR}
 }
 
-check_java_version() {
```
apc999, line 25: duplicated in `libexec/alluxio-config.sh`

integration/fuse/bin/alluxio-fuse
```diff
@@ -42,13 +30,6 @@ check_fuse_jar() {
   fi
 }
 
-set_java_opt() {
```
apc999, line 45: moved to `libexec/alluxio-config.sh` the central place to handle all env var setup

integration/fuse/bin/alluxio-fuse
```diff
@@ -60,9 +41,9 @@ mount_fuse() {
   fi
 
   if [[ ${NO_DAEMON} -eq "1" ]]; then
-    # Assumes that the end of ALLUXIO_FUSE_JAVA_OPTS is -Dalluxio.logger.type=FUSE_LOGGER
```
apc999, line 63: bad idea with tricky assumption. rewritten

libexec/alluxio-config.sh
```diff
@@ -151,6 +151,12 @@ if [[ -n "${ALLUXIO_LOGSERVER_HOSTNAME}" && -n "${ALLUXIO_LOGSERVER_PORT}" ]]; t
     ALLUXIO_WORKER_JAVA_OPTS+=" -Dalluxio.remote.logger.type=REMOTE_WORKER_LOGGER"
 fi
 
+# FUSE specific parameters that will be shared to all workers based on ALLUXIO_JAVA_OPTS.
+ALLUXIO_FUSE_JAVA_OPTS_DEFAULT=" -server -Xms1G -Xmx1G -XX:MaxDirectMemorySize=4g"
+ALLUXIO_FUSE_JAVA_OPTS_DEFAULT+=" -Djava.library.path=${ALLUXIO_HOME}/lib"
+ALLUXIO_FUSE_JAVA_OPTS_DEFAULT+=" -Dalluxio.logger.type=FUSE_LOGGER"
+ALLUXIO_FUSE_JAVA_OPTS="${ALLUXIO_FUSE_JAVA_OPTS_DEFAULT} ${ALLUXIO_JAVA_OPTS} ${ALLUXIO_FUSE_JAVA_OPTS}"
```
apc999, line 158: order here is tricky, typically JVM respect the rightmost setting when a java opt  is passed multiple times.
so let's go this way: from most general to most specific settings. in this way, when users' setting ALLUXIO_FUSE_JAVA_OPTS in alluxio-env.sh will be respected.

@calvinjia  do you agree here? shall we fix other env?

LuQQiu (commented):
Thanks for making life easier for people using the libjnifuse!
I tested on the Linux env, the changes look good.
Left one minor question, otherwise LGTM

libexec/alluxio-config.sh
calvinjia: Yes, most JVMs respect the latter option. I think this is okay and its not worth doing some more bash scripting to try and de-duplicate.

libexec/alluxio-config.sh
```diff
@@ -151,6 +151,12 @@ if [[ -n "${ALLUXIO_LOGSERVER_HOSTNAME}" && -n "${ALLUXIO_LOGSERVER_PORT}" ]]; t
     ALLUXIO_WORKER_JAVA_OPTS+=" -Dalluxio.remote.logger.type=REMOTE_WORKER_LOGGER"
 fi
 
+# FUSE specific parameters that will be shared to all workers based on ALLUXIO_JAVA_OPTS.
+ALLUXIO_FUSE_JAVA_OPTS_DEFAULT=" -server -Xms1G -Xmx1G -XX:MaxDirectMemorySize=4g"
+ALLUXIO_FUSE_JAVA_OPTS_DEFAULT+=" -Djava.library.path=${ALLUXIO_HOME}/lib"
```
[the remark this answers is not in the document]
apc999: that's sounds very nice to have. here I see it is not very necessary as it will complicate substaintially the script here.
I would leave it when we see people are encounting or asking for this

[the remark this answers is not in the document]
LuQQiu: Sounds good!

LuQQiu (approved):
LGTM! Thanks!

integration/fuse/bin/alluxio-fuse
cheyang: I think it's not used. We can remove it.

apc999:

PR #12864, opened 2021-02-17
Move asyncCache logics to DefaultBlockWorker

LuQQiu:
PR3 for #12830
Move asyncCache logics to DefaultBlockWorker so that it can be called by both internal client and outside gRPC client.


LuQQiu:
@apc999 Please take a look the small change to move AsyncCacheManager out of GRPC class

apc999 (approved):

LuQQiu:

PR #12869, opened 2021-02-17
Add cost saving estimation and add timestamp to operations saved

yuzhu:
This PR will export cost saving due to saved metadata operation per UFS, calculated from cloud vendor's cost sheet.


yuzhu:
@tieujason330 FYI,  added a new field in the MasterWebUIMetrics, mUfsOpsCostSaved which is a map mapping UfsURI to cost saved.

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * This class contains utilities to compute cloud cost savings.
+ */
+public final class CloudCostUtils {
+  private CloudCostUtils() {}  // prevent instantiation
+
+  public static final Map<UFSOps, Double> S3COSTMAP =
+      ImmutableMap.of(
+          UFSOps.CREATE_FILE, 0.000005,
```
gpang, line 28: What about cross region costs?

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.Map;
+
+/**
+ * This class contains utilities to compute cloud cost savings.
+ */
+public final class CloudCostUtils {
+  private CloudCostUtils() {}  // prevent instantiation
+
+  public static final Map<UFSOps, Double> S3COSTMAP =
```
gpang, line 26: What are all the units for these values?

Also, should each of these maps be pointing to a website to where the costs are retrieved? How often do they need to be updated?

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
```diff
@@ -0,0 +1,79 @@
...
+          UFSOps.GET_FILE_INFO, 0.0000004,
+          UFSOps.DELETE_FILE, 0.0,
+          UFSOps.LIST_STATUS, 0.000005);
+  public static final Map<UFSOps, Double> OSSCOSTMAP =
+      ImmutableMap.of(
+          UFSOps.CREATE_FILE, 0.000005,
+          UFSOps.GET_FILE_INFO, 0.0000004,
+          UFSOps.DELETE_FILE, 0.0,
+          UFSOps.LIST_STATUS, 0.000005);
+  public static final Map<String, Map<UFSOps, Double>> COSTMAP =
```
gpang, line 50: where is `OSSCOSTMAP`?

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
```diff
@@ -0,0 +1,79 @@
...
+      ImmutableMap.of(
+       "abfs", ABFSCOSTMAP,
+       "gcs", GCSCOSTMAP,
```
gpang, line 53: Are the keys supposed to be the scheme? I think the gcs scheme is `gs`.

Also, I think sometimes s3 can have different schemes like `s3n`, or `s3a`. Is there a way we can just use the UFS factories to match the keys?

gpang (changes requested):
@yuzhu Thanks! I left some comments.

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
yuzhu: cross region access cost is not part of the cost being calculated here. We are only counting saved metadata ops here. not even data access.

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
yuzhu: done

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
yuzhu: done

core/server/master/src/main/java/alluxio/master/metrics/CloudCostUtils.java
yuzhu: keys are the getFsType associated with each connector. It is unique to each connector and covers `s3n` and `s3a` schemes.

yuzhu:
image
@calvinjia @rastogiasr  this is the currently what the user sees, PTAL.

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -993,8 +995,11 @@ public Response getWebUIMetrics() {
             String ufsUnescaped = MetricsSystem.unescape(ufs);
             Map<String, Long> perUfsMap = ufsOpsSavedMap.getOrDefault(
                 ufsUnescaped, new TreeMap<>());
-            perUfsMap.put(alluxio.metrics.Metric.getBaseName(metricName)
-                    .substring(UFS_OP_SAVED_PREFIX.length()),
+            String alluxioOperation = alluxio.metrics.Metric.getBaseName(metricName)
+                .substring(UFS_OP_SAVED_PREFIX.length());
+            perUfsMap.put(String.format("%s (Roughly equivalent to %s operation)", alluxioOperation,
+                DefaultFileSystemMaster.Metrics.UFS_OPS_DESC.getOrDefault(
+                    DefaultFileSystemMaster.Metrics.UFSOps.valueOf(alluxioOperation), "no")),
```
calvinjia, line 1002: Does a default value make sense here?

calvinjia (approved):
Minor comment

yuzhu:

PR #12875, opened 2021-02-18

yuzhu:
Before this change, syncMetadata writes an inode journal entry and then two additional updateInode entries to complete the file. This combines three entries into one so that we do not have a failure mode where the first inode entry is inserted and incomplete files are left behind.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
@@ -13,6 +13,7 @@ package alluxio.master.journal;
 
 import alluxio.AlluxioURI;
 import alluxio.Constants;
+import alluxio.collections.Pair;
 import alluxio.conf.PropertyKey;
 import alluxio.conf.ServerConfiguration;
 import alluxio.master.journal.checkpoint.CheckpointInputStream;
@@ -229,14 +230,18 @@ public final class JournalUtils {
   public static List<alluxio.proto.journal.Journal.JournalEntry> mergeCreateComplete(
       List<alluxio.proto.journal.Journal.JournalEntry> entries) {
     List<alluxio.proto.journal.Journal.JournalEntry> newEntries = new ArrayList<>();
-    Map<Long, File.InodeFileEntry.Builder> fileEntryMap = new HashMap<>();
+    // file id : index in the newEntries, InodeFileEntry
+    Map<Long, Pair<Integer, File.InodeFileEntry.Builder>> fileEntryMap = new HashMap<>();
     for (alluxio.proto.journal.Journal.JournalEntry oldEntry : entries) {
       if (oldEntry.hasInodeFile()) {
+        // Use the old entry as a placeholder, to be replaced later
+        newEntries.add(oldEntry);
         fileEntryMap.put(oldEntry.getInodeFile().getId(),
-            File.InodeFileEntry.newBuilder(oldEntry.getInodeFile()));
+            new Pair<>(newEntries.size() - 1,
+                File.InodeFileEntry.newBuilder(oldEntry.getInodeFile())));
       } else if (oldEntry.hasUpdateInode()) {
         File.UpdateInodeEntry entry = oldEntry.getUpdateInode();
-        File.InodeFileEntry.Builder builder = fileEntryMap.get(entry.getId());
+        File.InodeFileEntry.Builder builder = fileEntryMap.get(entry.getId()).getSecond();
         if (builder == null) {
           newEntries.add(oldEntry);
           continue;
@@ -295,7 +300,7 @@ public final class JournalUtils {
         }
       } else if (oldEntry.hasUpdateInodeFile()) {
         File.UpdateInodeFileEntry entry = oldEntry.getUpdateInodeFile();
-        File.InodeFileEntry.Builder builder = fileEntryMap.get(entry.getId());
+        File.InodeFileEntry.Builder builder = fileEntryMap.get(entry.getId()).getSecond();
         if (builder == null) {
           newEntries.add(oldEntry);
           continue;
@@ -332,8 +337,11 @@ public final class JournalUtils {
         newEntries.add(oldEntry);
       }
     }
-    for (File.InodeFileEntry.Builder builder : fileEntryMap.values()) {
-      newEntries.add(Journal.JournalEntry.newBuilder().setInodeFile(builder).build());
+    for (Pair<Integer, File.InodeFileEntry.Builder> pair : fileEntryMap.values()) {
+      // Replace the old entry place holder with the new entry,
+      // to create the file in the same place in the journal
+      newEntries.set(pair.getFirst(),
+          Journal.JournalEntry.newBuilder().setInodeFile(pair.getSecond()).build());
     }
     return newEntries;
   }
--- a/core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
@@ -22,6 +22,8 @@ import alluxio.master.journal.checkpoint.Checkpointed;
 import alluxio.master.journal.checkpoint.CompoundCheckpointFormat.CompoundCheckpointReader;
 import alluxio.master.journal.checkpoint.CompoundCheckpointFormat.CompoundCheckpointReader.Entry;
 import alluxio.master.journal.sink.JournalSink;
+import alluxio.proto.journal.File;
+import alluxio.proto.journal.Journal;
 import alluxio.proto.journal.Journal.JournalEntry;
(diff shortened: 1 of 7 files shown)


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -69,6 +71,7 @@
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+
```
bf8086, line 74: (nit) extra new line

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
```diff
@@ -214,5 +220,123 @@ public static void sinkFlush(Supplier<Set<JournalSink>> journalSinks) {
...
+  public static List<alluxio.proto.journal.Journal.JournalEntry> mergeCreateComplete(
+      List<alluxio.proto.journal.Journal.JournalEntry> entries) {
+    List<alluxio.proto.journal.Journal.JournalEntry> newEntries = new ArrayList<>();
+    Map<Long, File.InodeFileEntry.Builder> fileEntryMap = new HashMap<>();
+    for (alluxio.proto.journal.Journal.JournalEntry oldEntry : entries) {
+      if (oldEntry.hasInodeFile()) {
+        fileEntryMap.put(oldEntry.getInodeFile().getId(),
+            File.InodeFileEntry.newBuilder(oldEntry.getInodeFile()));
+      } else if (oldEntry.hasUpdateInode()) {
+        File.UpdateInodeEntry entry = oldEntry.getUpdateInode();
```
bf8086, line 238: Would it be possible to reuse the update logic from `MutableInode::updateFromEntry`? It can be challenging to keep multiple places in sync and might result in missing updates when later on someone add a new field to the entry.

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
```diff
@@ -214,5 +220,123 @@ public static void sinkFlush(Supplier<Set<JournalSink>> journalSinks) {
...
+          builder.putAllXAttr(new HashMap<>(entry.getXAttrMap()));
+        }
+      } else if (oldEntry.hasUpdateInodeFile()) {
+        File.UpdateInodeFileEntry entry = oldEntry.getUpdateInodeFile();
+        File.InodeFileEntry.Builder builder = fileEntryMap.get(entry.getId());
+        if (builder == null) {
+          newEntries.add(oldEntry);
+          continue;
+        }
+        if (entry.hasPersistJobId()) {
```
bf8086, line 303: Similarly, can we reuse the logic from `MutableInodeFile.updateFromEntry`?

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
```diff
@@ -214,5 +220,123 @@ public static void sinkFlush(Supplier<Set<JournalSink>> journalSinks) {
...
+        if (entry.getSetBlocksCount() > 0) {
+          builder.clearBlocks();
+          builder.addAllBlocks(entry.getSetBlocksList());
+        }
+      } else {
+        newEntries.add(oldEntry);
+      }
+    }
+    for (File.InodeFileEntry.Builder builder : fileEntryMap.values()) {
+      newEntries.add(Journal.JournalEntry.newBuilder().setInodeFile(builder).build());
```
bf8086, line 336: Shall we keep the `InodeFile` entries in the original position in the list?

It is probably not an issue for now but might break natural dependency order if we later use this in other places or add more dependency in inode creation.

core/server/common/src/main/java/alluxio/master/journal/MergeJournalContext.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Context for merging journal entries together for a wrapped journal context.
+ */
+@NotThreadSafe
+public final class MergeJournalContext implements JournalContext {
+  private static final Logger LOG = LoggerFactory.getLogger(MergeJournalContext.class);
+  private final JournalContext mJournalContext;
+  private final UnaryOperator<List<JournalEntry>> mMergeOperator;
+  private final List<JournalEntry> mJournalEntries;
```
bf8086, line 35: Can we add a warning message when we have too many entries in this list? Just so that we can be aware of potential memory pressure in case it is used for large list of entries accidentally.

bf8086 (changes requested):
Thanks @yuzhu! Overall LGTM. Would it be possible to add some unit tests?

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
yuzhu: yeah i thought about that as well, but it is actually a bit tricky because of the subtle difference between the inode class and the inode proto.  They do not share anything in the inheritance tree either, so that makes it difficult to factor out this part of the logic.  any suggestions?

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
yuzhu: see above

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
yuzhu: done

core/server/common/src/main/java/alluxio/master/journal/MergeJournalContext.java
yuzhu: done

yuzhu (commented):
@bf8086 responded to your comments, PTAL

core/server/common/src/main/java/alluxio/master/journal/JournalUtils.java
bf8086: Would it be easier to convert entry to `MutableInodeFile` first? Something like

```
    Map<Long, Pair<Integer, MutableInodeFile>> fileEntryMap = new HashMap<>();
    for (alluxio.proto.journal.Journal.JournalEntry oldEntry : entries) {
      if (oldEntry.hasInodeFile()) {
        ...
        fileEntryMap.put(oldEntry.getInodeFile().getId(),
            new Pair<>(newEntries.size() - 1,
                MutableInodeFile.fromJournalEntry(oldEntry.getInodeFile())));
      } else if (oldEntry.hasUpdateInode()) {
        ...
        MutableInodeFile inode = fileEntryMap.get(entry.getId()).getSecond();
        inode.updateFromEntry(entry);
      } else if (oldEntry.hasUpdateInodeFile()) {
        ...
        MutableInodeFile inode = fileEntryMap.get(entry.getId()).getSecond();
        inode.updateFromEntry(entry);
```

yuzhu:
@bf8086 PTAL, it was a little more complicated than initially thought. Turns out our fromJournal and toJournalEntry are not inverse functions exactly.. so i had to force it to reuse the file id.

yuzhu:
@bf8086 PTAL

core/server/common/src/main/java/alluxio/master/journal/MergeJournalContext.java
```diff
@@ -0,0 +1,78 @@
...
+    mJournalContext = journalContext;
+    mMergeOperator = merger;
+    mJournalEntries = new ArrayList<>();
+  }
+
+  @Override
+  public void append(JournalEntry entry) {
+    mJournalEntries.add(entry);
+    if (mJournalEntries.size() > MAX_ENTRIES) {
+      LOG.debug("MergeJournalContext has " + mJournalEntries.size()
```
bf8086, line 67: (nit) consider using sample logging or log once at closing to avoid excessive logging.

bf8086 (approved):
LGTM. Thanks!

LuQQiu (approved):

LuQQiu:
03:55:16.676 [INFO] Running alluxio.master.transport.GrpcMessagingTransportTest
Error: 7.771 [ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.092 s <<< FAILURE! - in alluxio.master.transport.GrpcMessagingTransportTest
Error: 7.772 [ERROR] testServerClosed(alluxio.master.transport.GrpcMessagingTransportTest)  Time elapsed: 0.84 s  <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at alluxio.master.transport.GrpcMessagingTransportTest.testServerClosed(GrpcMessagingTransportTest.java:153)

yuzhu:

yuzhu:

Issue #12882, opened 2021-02-19
UnsatisfiedLinkError: sun.management.OperatingSystemImpl.getTotalPhysicalMemorySize() running Alluxio client in Spark


jiacheliu3:
**Alluxio Version:**
2.5.0-SNAPSHOT commit 3ef0b4d719

**Describe the bug**
Spark driver log
```
21/02/19 01:48:33 INFO StateStoreCoordinatorRef: Registered StateStoreCoordinator endpoint
21/02/19 01:48:33 INFO AbstractFileSystem: Creating Alluxio configuration from Hadoop configuration {}, uri configuration {alluxio.zookeeper.address=null, alluxio.zookeeper.enabled=false, alluxio.master.hostname=alluxio-master-0, alluxio.master.rpc.addresses=null, alluxio.master.embedded.journal.addresses=null, alluxio.master.rpc.port=19998}
Exception in thread "main" java.lang.UnsatisfiedLinkError: sun.management.OperatingSystemImpl.getTotalPhysicalMemorySize()J
	at sun.management.OperatingSystemImpl.getTotalPhysicalMemorySize(Native Method)
	at alluxio.conf.PropertyKey.lambda$static$8(PropertyKey.java:3044)
	at alluxio.DefaultSupplier.get(DefaultSupplier.java:34)
	at alluxio.conf.PropertyKey.getDefaultValue(PropertyKey.java:7821)
	at alluxio.conf.InstancedConfiguration.validate(InstancedConfiguration.java:549)
	at alluxio.util.ConfigurationUtils.reloadProperties(ConfigurationUtils.java:456)
	at alluxio.util.ConfigurationUtils.defaults(ConfigurationUtils.java:404)
	at alluxio.hadoop.AbstractFileSystem.initialize(AbstractFileSystem.java:512)
	at alluxio.hadoop.AbstractFileSystem.initialize(AbstractFileSystem.java:474)
	at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:2669)
        ...
org.apache.spark.deploy.SparkSubmit.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:845)
	...
```

**To Reproduce**
Following doc 
The spark-alluxio image is built against spark-2.4.4 with the Alluxio client jar.

Run a spark-submit with a word count job:
```
/tmp/alluxio/integration/spark/bin/spark-submit --master k8s://https://blah:6443 \
[...]

PR #12884, opened 2021-02-19
Catch UnsatisfiedLinkError for alluxio.worker.ramdisk.size, improve desc

jiacheliu3:
When running Alluxio client in a environment without com.sun.management, an UnsatisfiedLinkError will be thrown. The current code only catches `Exception` so not able to catch `UnsatisfiedLinkError`.

Also the current implementation can give incorrect ram disk size in a containerized environment. Improve the description to reflect that.

Related to https://github.com/Alluxio/alluxio/issues/12882

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2819,7 +2819,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               return "1GB";
             }
           }, "2/3 of total system memory, or 1GB if system memory size cannot be determined")
-          .setDescription("Memory capacity of each worker node.")
+          .setDescription("Memory capacity of each worker node. "
+                  + "It is recommended to set this value explicitly. "
+                  + "If you are running Alluxio workers in a containerized environment, "
+                  + "your Java runtime may return incorrect system memory "
+                  + "(i.e. the host memory instead of the container memory). "
+                  + "You should upgrade your JDK version according to "
+                  + "https://bugs.openjdk.java.net/browse/JDK-8226575.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2813,7 +2813,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               OperatingSystemMXBean operatingSystemMXBean =
                   (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
               return operatingSystemMXBean.getTotalPhysicalMemorySize() * 2 / 3;
-            } catch (Exception e) {
+            } catch (Throwable e) {
               // The package com.sun.management may not be available on every platform.
               // fallback to a reasonable size.
               return "1GB";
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2813,13 +2813,19 @@ public String toString() {
...
               return operatingSystemMXBean.getTotalPhysicalMemorySize() * 2 / 3;
-            } catch (Exception e) {
+            } catch (Throwable e) {
               // The package com.sun.management may not be available on every platform.
               // fallback to a reasonable size.
               return "1GB";
             }
           }, "2/3 of total system memory, or 1GB if system memory size cannot be determined")
-          .setDescription("Memory capacity of each worker node.")
+          .setDescription("Memory capacity of each worker node. "
```
jiacheliu3, line 2822: @LuQQiu Please see if this makes sense to you

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2813,13 +2813,19 @@ public String toString() {
...
+                  + "It is recommended to set this value explicitly. "
+                  + "If you are running Alluxio workers in a containerized environment, "
+                  + "your Java runtime may return incorrect system memory "
+                  + "(i.e. the host memory instead of the container memory). "
+                  + "You should upgrade your JDK version according to "
```
LuQQiu, line 2827: Why higher version JDK doesn't have this issue? I don't think the link recommend the upgrade, or I just didn't find the right place?

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Why higher version JDK doesn't have this issue? I don't think the link recommend the upgrade, or I just didn't find the right place?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2813,13 +2813,19 @@ public String toString() {
...
+                  + "https://bugs.openjdk.java.net/browse/JDK-8226575.")
```
LuQQiu, line 2828: I preferred not mention this line since the link isn't suggested the upgrade clearly and it's not guaranteed that the user will not see the issue in newer version (unless we test it)

LuQQiu (commented):
Left one small comment, otherwise LGTM

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: I think we should mention this pitfall, so that if anyone happens to meet this issue we can refer to the doc. I agree seeing is believing (and I have not tested different java versions). But testing this is non-trivial 0.0
@apc999 What do you think?

LuQQiu: @jiacheliu3 After some investigation which I mentioned in the issue, I will prefer to not mention the detailed reason here but only change the Exception to Throwable.
I doubt if the reason is true that the Java runtime return the host memory. From what I see, the JVM loads the correct jdk file [Loaded sun.management.OperatingSystemImpl from /usr/lib/jvm/java-1.8-openjdk/jre/lib/rt.jar] but the class somehow doesn't contain that method.

apc999: I'm thinking this might also be related to classloading issues in your container environment.
even the jdk files are loaded, if by the wrong classloader, we may still the link error.

apc999: agree, let's not mention JDK upgrade for now.
we can leave a TODO in src code

jiacheliu3: @apc999 @LuQQiu Sorry about the confusion.

The comments here are related to the known bug in JDK when calling `operatingSystemMXBean.getTotalPhysicalMemorySize()`, not related to #12882. Using a higher version JDK will give you "correct" physical memory size, but NOT resolve the UnsatisfiedLinkError.

But it makes sense we discuss separately so I removed this part from the PR.

LuQQiu (approved):

jiacheliu3:

PR #12890, opened 2021-02-19
Add worker Fuse manager and client read logics

LuQQiu:
PR4 for #12830
This PR adds the Fuse manager to control the worker internal Fuse application lifecycle
and add the internal client read logics.

This PR depends on #12837 #12838 #12864


LuQQiu:
@apc999 PTAL the embedded fuse worker/client side main logics, thanks!

core/client/fs/src/main/java/alluxio/client/block/stream/WorkerInternalClientDataReader.java
```diff
@@ -0,0 +1,158 @@
...
+      mWorkerInternalBlockWorker = context.acquireWorkerInternalBlockWorkerClient();
+    }
+
+    @Override
+    public DataReader create(long offset, long len) throws IOException {
+      mReadRequestPartial = mReadRequestPartial.toBuilder()
+          .setOffset(offset).setLength(len).build();
+      mBlockReadRequest = new BlockReadRequest(mReadRequestPartial);
+      try {
+        mReader = mWorkerInternalBlockWorker.getBlockReader(mBlockReadRequest);
```
apc999, line 132: is it necessary to create `mReadRequestPartial` in this Factory? I actually don't see how a proto is needed except to create  `mBlockReadRequest` to pass to `getBlockReader` and `cleanBlockReader`. If I recall correctly, we don't require a proto to be passed there, right?

core/common/src/main/java/alluxio/worker/block/io/WorkerInternalBlockWorker.java
```diff
@@ -0,0 +1,56 @@
...
+ * without going through external RPC frameworks.
+ */
+public interface WorkerInternalBlockWorker {
+
+  /**
+   * Submits the async cache request to the worker to execute.
+   *
+   * @param request the async cache request
+   */
+  void asyncCache(AsyncCacheRequest request);
```
apc999, line 34: when client is reading from a worker at the same process, we don't need to call `asyncCache` anymore. It will be a redundant operation b/c worker will have already get the data cached. Async cache operation is only needed if client on node A is reading from worker B and now wants the worker on the same node A to cache this block

core/common/src/main/java/alluxio/worker/block/io/WorkerInternalBlockWorker.java
```diff
@@ -0,0 +1,56 @@
...
+import alluxio.grpc.AsyncCacheRequest;
+import alluxio.wire.BlockReadRequest;
+
+import java.io.IOException;
+
+/**
+ * An interface for worker internal clients to interact with worker operations directly
+ * without going through external RPC frameworks.
+ */
+public interface WorkerInternalBlockWorker {
```
apc999, line 27: - Why this is called `WorkerInternalBlockWorker`, the name is confusing
- actually we do we need an interface with a subset of methods of `BlockWorker`?

core/server/worker/src/main/java/alluxio/worker/block/BlockWorker.java
```diff
@@ -438,20 +438,30 @@ void closeUfsBlock(long sessionId, long blockId)
...
+   *         because the block exists in the Alluxio block store after opening the ufs block reader
+   * @throws BlockDoesNotExistException if the requested block does not exist in this worker
+   * @throws InvalidWorkerStateException if blockId does not belong to sessionId
+   * @throws WorkerOutOfSpaceException if there is no enough space
+   * @throws IOException if it fails to get block reader
    */
-  BlockReader getBlockReader(BlockReadRequest request) throws IOException,
-      BlockDoesNotExistException, InvalidWorkerStateException,
-      BlockAlreadyExistsException, WorkerOutOfSpaceException;
+  BlockReader getBlockReader(BlockReadRequest request) throws
```
apc999, line 448: shall this `BlockWorker` interface to extend `WorkerInternalBlockWorker`?

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -255,6 +257,9 @@ public void start(WorkerNetAddress address) throws IOException {
               (int) ServerConfiguration.getMs(PropertyKey.WORKER_BLOCK_HEARTBEAT_INTERVAL_MS),
                   ServerConfiguration.global(), ServerUserState.global()));
     }
+
+    // Mounts the embedded Fuse application if configured
+    mFuseManager.start();
```
apc999, line 262: shall this be protected by a flag or config?

apc999 (changes requested):
Thanks @LuQQiu  for your PR! Comments left

core/client/fs/src/main/java/alluxio/client/block/stream/WorkerInternalClientDataReader.java
LuQQiu: The BlockReadRequest is heavily bundled with Proto files:
```
public final class BlockReadRequest extends ReadRequest {
  private final Protocol.OpenUfsBlockOptions mOpenUfsBlockOptions;

  /**
   * Creates an instance of {@link BlockReadRequest}.
   *
   * @param request the block read request
   */
  public BlockReadRequest(alluxio.grpc.ReadRequest request) {
    super(request.getBlockId(), request.getOffset(), request.getOffset() + request.getLength(),
        request.getChunkSize());

    if (request.hasOpenUfsBlockOptions()) {
      mOpenUfsBlockOptions = request.getOpenUfsBlockOptions();
    } else {
      mOpenUfsBlockOptions = null;
    }
    mPromote = request.getPromote();
    mPositionShort = request.getPositionShort();
    // Note that we do not need to seek to offset since the block worker is created at the offset.
  }
```

I didn't fully move the proto logics out because I think that will impact more files and could be done later after merging all the fuse code back to the master branch if needed.

core/common/src/main/java/alluxio/worker/block/io/WorkerInternalBlockWorker.java
LuQQiu: The reason i call it `WorkerInternalBlockWorker` is because i think `EmbeddedBlockWorker` is confusing as well. The code is client code but it calls `Embedded...`, I don't know what it embedded to, so I give this name, any suggestion for the naming?

core/common/src/main/java/alluxio/worker/block/io/WorkerInternalBlockWorker.java
LuQQiu: Yeah, the client on node A is reading from worker B and now wants the embedded worker A to cache this block? I think the above case is still valid?

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -52,7 +52,10 @@
 
   /** the source tracking where the block is from. */
   public enum BlockInStreamSource {
-    LOCAL, REMOTE, UFS
+    // Reads from local worker or reads from local worker storage directly
```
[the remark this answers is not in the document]
LuQQiu: Add a new EMBEDDED type, agree that we should diff these two types

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
```diff
@@ -171,6 +175,10 @@ public static void launchFuse(FileSystemContext fsContext,
       }
     } catch (Throwable e) {
       throw new IOException("Failed to mount Alluxio file system", e);
+    } finally {
+      if (blocking) {
```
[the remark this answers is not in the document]
LuQQiu: If blocking is false, i previously rely on process kill to close fs, now change to close fs in FuseManager.close() so that in the future, we can have more control of the embedded fuse manager

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -116,6 +119,12 @@ public static BlockInStream create(FileSystemContext context, BlockInfo info,
     boolean sourceSupportsDomainSocket = NettyUtils.isDomainSocketSupported(dataSource);
     boolean sourceIsLocal = dataSourceType == BlockInStreamSource.LOCAL;
 
+    if (sourceIsLocal && context.acquireWorkerInternalBlockWorkerClient() != null) {
```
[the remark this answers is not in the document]
LuQQiu: Add a `isWorkerInternalClient` to directly tell if the client is in worker, thanks!
```
 /**
   * @return if the current client is embedded in a worker
   */
  public synchronized boolean isWorkerInternalClient() {
    return mWorkerInternalBlockWorker != null;
  }

```

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -140,6 +149,27 @@ public static BlockInStream create(FileSystemContext context, BlockInfo info,
...
+   * @param context the file system context
+   * @param address the network address of the gRPC data server to read from
+   * @param blockId the block ID
+   * @param length the block length
+   * @param options the in stream options
+   * @return the {@link BlockInStream} created
+   */
+  private static BlockInStream createWorkerInternalBlockInStream(FileSystemContext context,
+      WorkerNetAddress address, long blockId, long length, InStreamOptions options) {
+    long chunkSize = context.getClusterConf().getBytes(
```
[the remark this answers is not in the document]
LuQQiu: Removed the old property, the usage is the same, just worried about the property naming may make users confusing

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -153,6 +183,7 @@ public static BlockInStream create(FileSystemContext context, BlockInfo info,
   private static BlockInStream createLocalBlockInStream(FileSystemContext context,
       WorkerNetAddress address, long blockId, long length, InStreamOptions options)
       throws IOException {
+    // TODO(lu) should i change this property key?
```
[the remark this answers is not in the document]
LuQQiu: Sounds good

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -488,6 +516,18 @@ public void close() {
...
 
+  /**
+   * Acquires the worker internal block worker client
+   * as a gateway for worker internal clients to communicate with
+   * the local worker directly without going through external RPC frameworks.
+   *
+   * @return the acquired worker internal block worker client
+   */
+  @Nullable
+  public WorkerInternalBlockWorker acquireWorkerInternalBlockWorkerClient() {
```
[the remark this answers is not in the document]
LuQQiu: Change to 
```  public WorkerInternalBlockWorker getInternalBlockWorker() {
    return mWorkerInternalBlockWorker;
  }
```

core/server/worker/src/main/java/alluxio/worker/block/BlockWorker.java
LuQQiu: Removed the WorkerInternalBlockWorkerImpl and let the BlockWorker interface directly extend the WorkerInternalBlockWorker

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
LuQQiu: Add a new configuration `WORKER_FUSE_ENABLED` to decide whether to start fuse manager, thanks!

LuQQiu:
@apc999 Addressed the comments, PTAL,
on my side, i plan to do more testing first, if i have spare time, i will add the related tests

LuQQiu:
Error: 9.094 [ERROR] alluxio.client.cli.fs.command.ChownCommandIntegrationTest.alluxio.client.cli.fs.command.ChownCommandIntegrationTest
Error: 9.095 [ERROR]   Run 1: ChownCommandIntegrationTest>AbstractFileSystemShellTest.beforeClass:123 » Runtime
Error: 9.095 [ERROR]   Run 2: ChownCommandIntegrationTest>AbstractFileSystemShellTest.afterClass:132 » NullPointer
05:17:49.095 [INFO]

LuQQiu:
Errors: 
Error: 7.653 [ERROR]   FileSystemIntegrationTest.alluxio.client.hadoop.FileSystemIntegrationTest » Runtime

apc999 (approved):

apc999:

PR #12891, opened 2021-02-19
Add the metrics for embedded journal snapshot

LuQQiu:
Add metrics for better understanding the embedded journal snapshot behaviors


core/server/common/src/main/java/alluxio/master/journal/raft/SnapshotReplicationManager.java
```diff
@@ -171,8 +174,11 @@ public SnapshotReplicationManager(RaftJournalSystem journalSystem,
       String address = String.valueOf(client.getAddress());
       SnapshotDownloader<DownloadSnapshotPRequest, DownloadSnapshotPResponse> observer =
           SnapshotDownloader.forFollower(mStorage, address);
+      Timer.Context ctx = MetricsSystem
+          .timer(MetricKey.MASTER_EMBEDDED_JOURNAL_SNAPSHOT_DOWNLOAD_TIMER.getName()).time();
       client.downloadSnapshot(observer);
       return observer.getFuture().thenApplyAsync((termIndex) -> {
+        ctx.close();
```
gpang, line 181: Will this get closed twice, once here and once in `whenComplete`?

gpang (commented):
@LuQQiu a minor question

core/server/common/src/main/java/alluxio/master/journal/raft/SnapshotReplicationManager.java
LuQQiu: Good catch, the timer got close() twice which is incorrect. Removed the close() in completeRequest. Thanks!

gpang (approved):
LGTM

Have you done a sanity check for these metrics?

yuzhu:
@LuQQiu are these metrics ready to go? almost always good to have more metrics as long as they are not expensive..

LuQQiu:
@yuzhu Yeah, i do a small empty lines fix and wait for the test build, i think it's ready to merge

LuQQiu:

PR #12896, opened 2021-02-20
[DOCFIX] Adjust Spark on Alluxio in K8s docs

ZhuTopher:
The default k8s API server port is (for recent version of Kubernetes) 6443, not 8443.

This PR also adjusts and adds notes.

### DIFF
--- a/docs/en/compute/Spark-On-Kubernetes.md
+++ b/docs/en/compute/Spark-On-Kubernetes.md
@@ -121,9 +121,9 @@ You can find more about how to mount volumes to Spark executors in Spark
   {% endnavtab %}
 {% endnavtabs %}
 
-Note: 
-- Volume support in Spark was added in version 2.4.0.
-- You may observe a performance hit when not using short-circuit access via a domain socket.
+> Note: 
+> - Volume support in Spark was added in version 2.4.0.
+> - You may observe a performance hit when not using short-circuit access via a domain socket.
 
 ### Run a Spark job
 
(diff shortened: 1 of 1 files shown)


docs/en/compute/Spark-On-Kubernetes.md
```diff
@@ -145,9 +140,19 @@ $ kubectl create clusterrolebinding spark-role --clusterrole=edit \
...
+
+The output and time taken can be seen in the logs for Spark driver pod. Refer to
+the Spark [documentation](https://spark.apache.org/docs/latest/running-on-kubernetes.html)
+for further details about running Spark on Kubernetes. For example, you may find
+additional details on some of the flags used in this command
+[here](https://spark.apache.org/docs/latest/running-on-kubernetes.html?q=cluster-info#cluster-mode).
+
 Run the job from the Spark distribution directory
 ```console
-$ ./bin/spark-submit --master k8s://https://<kubernetes-api-server>:8443 \
```
jiacheliu3, line 150: nit: Yea it is indeed 6443 now but I'm not sure why this changed. If you happen to figure out could you let me know?
One possibility is the k8s version change.
Another possibility was when the doc was first written it was using minikube which defaults to 8443.

docs/en/compute/Spark-On-Kubernetes.md
```diff
@@ -160,10 +165,15 @@ $ ./bin/spark-submit --master k8s://https://<kubernetes-api-server>:8443 \
 local:///opt/spark/examples/jars/spark-examples_2.11-2.4.4.jar \
 alluxio://<alluxio-master>:19998/LICENSE
 ```
-> Note: You can find the address of the Kubernetes API server by running `kubectl cluster-info`.
-You can find more details in Spark [documentation](https://spark.apache.org/docs/latest/running-on-kubernetes.html?q=cluster-info#cluster-mode).
-You should also using the properties corresponding to your 
-[domain socket volume type]({{ '/en/compute/Spark-On-Kubernetes.html#short-circuit-operations' | relativize_url }}).
+
+Note:
```
jiacheliu3, line 169: Does `>` work with `-` so you can list items in a note? That way you stay consistent for the note format. @Xenorith

jiacheliu3 (approved):
@ZhuTopher Thanks! I left a few comments but this can be taken as approval, once the nit items are fixed.

docs/en/compute/Spark-On-Kubernetes.md
ZhuTopher: From what I saw, the `>` syntax didn't render properly into a note on the webpage. I also don't think the list syntax would render as a bullet list if prepended by the `>` syntax like the following:
```
> Note:
> - item 1
> - item 2
```

docs/en/compute/Spark-On-Kubernetes.md
ZhuTopher: I can't really seem to find links to older version of the Kubernetes docs on the matter. I'd guess the docs probably did use minikube which defaults to 8443 rather than somebody configuring `kubeadm` to use port 8443 over 6443. Should I make a note of this in the docs?

jiacheliu3: Feel free to add to the notes below sth like `The port for the Kubernetes API server may be different depending on your Kubernetes configuration.` Does that make sense?

LuQQiu:

LuQQiu:

PR #12897, opened 2021-02-20

yuzhu:
Before this change, syncMetadata writes an inode journal entry and then
two additional updateInode entries to complete the file. This combines
three entries into one so that we do not have a failure mode where the
first inode entry is inserted and incomplete files are left behind.

pr-link: Alluxio/alluxio#12875
change-id: cid-bdfcf60239984c608870b47f50a9c3e74f4c1b9f


yuzhu:
@bf8086 cherrypicked from #12875

yuzhu:

PR #12899, opened 2021-02-22
Fix journal error message when RPCs are cancelled

yuzhu:
When a RPC is cancelled, we interrupt InodeSyncStreams that are doing the sync, which might interrupt journal writes in progress. 
Before this change, this would produce a rather alarming Journal Writes failed message, now we show journal writes interrupted instead.


bf8086 (approved):

yuzhu:

LuQQiu:

PR #12901, opened 2021-02-22
Report error if any of the sync paths failed in syncMetadata

yuzhu:
Before this change, if any path was successfully synced, we return success from syncMetadata. 
After this change, we only return success if there was no failures and all paths have been synced without interruption.

This also fixes an issue where an interrupted syncmetadata is considered synced by the file system master and will not be synced again until sync interval passes.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -396,6 +396,10 @@ public class InodeSyncStream {
               elapsedTime.toMillis(), mRootScheme);
     }
     boolean success = syncPathCount > 0;
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_METADATA_SYNC_REPORT_FAILURE)) {
+      // There should not be any failed or outstanding jobs
+      success = (failedSyncPathCount == 0) && mSyncPathJobs.isEmpty() && mPendingPaths.isEmpty();
+    }
     if (success) {
       // update the sync path cache for the root of the sync
       // TODO(gpang): Do we need special handling for failures and thread interrupts?
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2453,6 +2453,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setScope(Scope.MASTER)
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
+  public static final PropertyKey MASTER_METADATA_SYNC_REPORT_FAILURE =
+      new Builder(Name.MASTER_METADATA_SYNC_REPORT_FAILURE)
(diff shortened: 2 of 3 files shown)


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -396,6 +396,10 @@ public boolean sync() throws AccessControlException, InvalidPathException {
               elapsedTime.toMillis(), mRootScheme);
     }
     boolean success = syncPathCount > 0;
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_METADATA_SYNC_REPORT_FAILURE)) {
+      // There should not be any failed or outstanding jobs
+      success = (failedSyncPathCount == 0) && mSyncPathJobs.isEmpty() && mPendingPaths.isEmpty();
+    }
     if (success) {
       // update the sync path cache for the root of the sync
```
bf8086, line 404: What is the side effect now that we skip the sync path cache update on partial syncs?

yuzhu: if we did a `ls -R /path` and failed, we should not update the sync path cache. Otherwise, any path under /path will be considered synced when they are in fact not in sync.

bf8086 (approved):
LGTM. Thanks!

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
gpang: Is there some way we can have a unittest which confirms this behavior?

gpang (commented):
@yuzhu I left a minor question.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: added

yuzhu: added the test which does three things:
1. simulates an interrupted syncmetadata
2. check to make sure it leaves no incomplete files
3. check to make sure the files are not considered synced by the file master

yuzhu:
i just realized that this PR does more than returning errors to the user. It also fixes a bug where a part of the tree that is incompletely synced could be incorrectly considered synced. but the user proably won’t notice unless their sync interval is something other than 0 or -1.   The newly added test fails without this change and passes with this change. @bf8086 @gpang

gpang:
@yuzhu If this fixes bugs, then why do we need a flag?

yuzhu:
@gpang the flag is because of a prior discussion. Since it significantly changes the behavior of how ufs sync error is handled, we want to be able to work around it at runtime if, for example.  ufs sync always fails for a particular file.

yuzhu:
@gpang any more comments? trying to close this out today

gpang (approved):
LGTM

yuzhu:

LuQQiu:

PR #12911, opened 2021-02-23
Reduce the overhead of operations saved metrics

yuzhu:


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4402,7 +4399,7 @@ private void checkUfsMode(AlluxioURI alluxioPath, OperationType opType)
         = MetricsSystem.counter(MetricKey.MASTER_SET_ATTRIBUTE_OPS.getName());
     private static final Counter UNMOUNT_OPS
         = MetricsSystem.counter(MetricKey.MASTER_UNMOUNT_OPS.getName());
-    private static final Map<String, Map<UFSOps, Counter>> SAVED_UFS_OPS
+    private static final Map<AlluxioURI, Map<UFSOps, Counter>> SAVED_UFS_OPS
         = new ConcurrentHashMap<>();
```
ZacBlanco, line 4403: I see we've got two CHMs here. Based on the flamegraphs, locking within the CHM was contributing to increased CPU utilization (latency).

Do you think it would be better to remove the CHM all together and use some kind of wait-free queue to queue the increment operations, and then serialize and execute the increments asynchronously in an executor service somewhere? the counters would then become eventually consistent in a sense, but it could decrease the overhead of needing to access the CHM for latency-sensitive RPC operations.

ZacBlanco (commented):
Just one comment regarding using the ConcurrentHashMap. It would be a little more complex of an implementation, but could probably shave off some more latency

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
yuzhu: yeah, but let's address that in a separate PR. 

For this PR, AFAIK, CHM performs ok when there is little contention, it is only a problem when it is heavily contended. So I propose that we precompute the counter object and pass it to `getFileInfoInternal` when it is used in the context of listStatus. This way we avoid multiple look ups and reduces contention significantly.

ZacBlanco: I think that's a good optimization for listStatus. I'm fine kicking something async later down the road

ZacBlanco (approved):

yuzhu:

yuzhu:

LuQQiu:

PR #12944, opened 2021-02-25
Skip object store parent creation

bf8086:
Skip creating parent directory for object stores to improve performance. 

Passed UFS contract tests and other object store related tests.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -751,6 +751,16 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_OBJECT_STORE_SKIP_PARENT_DIRECTORY_CREATION =
+      new Builder(Name.UNDERFS_OBJECT_STORE_SKIP_PARENT_DIRECTORY_CREATION)
+          .setDefaultValue(true)
+          .setDescription("Do not create parent directory for new files. Object stores generally "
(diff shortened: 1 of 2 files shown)


bf8086:
@LuQQiu PTAL.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -751,6 +751,16 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_OBJECT_STORE_SKIP_PARENT_DIRECTORY_CREATION =
+      new Builder(Name.UNDERFS_OBJECT_STORE_SKIP_PARENT_DIRECTORY_CREATION)
+          .setDefaultValue(true)
+          .setDescription("Do not create parent directory for new files. Object stores generally "
+              + "uses prefix which is not required for creating new files. Skipping parent "
+              + "directory is recommended for better performance. Set this to false if the "
+              + "object store requires prefix creation for new files.")
```
LuQQiu, line 760: Does any object store require prefix creation as you know?

LuQQiu (commented):
Overal LGTM, left one question, thanks

core/common/src/main/java/alluxio/conf/PropertyKey.java
bf8086: I am not aware of any. The S3 APIs only uses prefix and there is no need to create a prefix before creating an object under the prefix.

LuQQiu: we can do more testing to make sure that

LuQQiu (approved):
LGTM, thanks for fixing it

LuQQiu:

LuQQiu:

PR #12948, opened 2021-02-26
Add logging on slow UFS calls

apc999:
Add information on UFS call logging:
(1) time elapsed for each UFS call
(2) add a new flag to control log UFS calls that takes time longer than defined threshold.

### DIFF
--- a/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
+++ b/core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
@@ -16,6 +16,8 @@ import alluxio.Constants;
 import alluxio.SyncInfo;
 import alluxio.collections.Pair;
 import alluxio.conf.AlluxioConfiguration;
+import alluxio.conf.InstancedConfiguration;
+import alluxio.conf.PropertyKey;
 import alluxio.exception.status.UnimplementedException;
 import alluxio.security.authorization.AccessControlList;
 import alluxio.metrics.Metric;
@@ -30,6 +32,7 @@ import alluxio.underfs.options.FileLocationOptions;
 import alluxio.underfs.options.ListOptions;
 import alluxio.underfs.options.MkdirsOptions;
 import alluxio.underfs.options.OpenOptions;
+import alluxio.util.ConfigurationUtils;
 import alluxio.util.SecurityUtils;
 
 import com.codahale.metrics.Timer;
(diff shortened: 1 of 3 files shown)


core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
```diff
@@ -52,7 +55,8 @@
  */
 public class UnderFileSystemWithLogging implements UnderFileSystem {
   private static final Logger LOG = LoggerFactory.getLogger(UnderFileSystemWithLogging.class);
-
+  private static final long THRESHOLD = new InstancedConfiguration(ConfigurationUtils.defaults())
```
gpang, line 58: Should this just use the `mConf` to get the property?

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
```diff
@@ -1203,14 +1207,27 @@ public String toString() {
...
+        LOG.warn("{}({}) returned OK in {} ms (>={} ms)", methodName,
+            callable, durationMs, THRESHOLD);
+      }
       return ret;
     } catch (IOException e) {
+      durationMs = System.currentTimeMillis() - startMs;
       MetricsSystem.counter(getQualifiedFailureMetricName(methodName)).inc();
-      LOG.debug("Exit (Error): {}: {}, Error={}", methodName, callable, e.getMessage());
+      LOG.debug("Exit (Error): {}({}) in {} ms, Error={}",
+          methodName, callable, e.getMessage(), durationMs);
```
gpang, line 1226: I think the arguments are in the wrong order. `e.getMessage()` and `durationMs` should be switched.

core/common/src/main/java/alluxio/underfs/UnderFileSystemWithLogging.java
```diff
@@ -1203,14 +1207,27 @@ public String toString() {
...
+      if (durationMs >= THRESHOLD) {
+        LOG.warn("{}({}) returned {} in {} ms (>={} ms)", methodName,
```
gpang, line 1228: NIT: maybe enclose the error message in "" or ``?

gpang (changes requested):
@apc999 Thanks!

What about any data read/write operations? This won't be able to log any of those UFS calls.

gpang (approved):
LGTM

apc999:

PR #12953, opened 2021-02-26
Keep enum configuration backwards compatible

apc999:
Fix https://alluxio.atlassian.net/browse/ALLUXIO-3402

### DIFF
--- a/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
+++ b/core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
@@ -304,14 +304,12 @@ public class InstancedConfiguration implements AlluxioConfiguration {
     Preconditions.checkArgument(delimiter != null,
         "Illegal separator for Alluxio properties as list");
     String rawValue = get(key);
-
     return ConfigurationUtils.parseAsList(rawValue, delimiter);
   }
 
   @Override
   public <T extends Enum<T>> T getEnum(PropertyKey key, Class<T> enumType) {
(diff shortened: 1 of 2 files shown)


core/common/src/main/java/alluxio/conf/InstancedConfiguration.java
```diff
@@ -304,14 +304,12 @@ public boolean getBoolean(PropertyKey key) {
...
     String rawValue = get(key);
-
     return ConfigurationUtils.parseAsList(rawValue, delimiter);
   }
 
   @Override
   public <T extends Enum<T>> T getEnum(PropertyKey key, Class<T> enumType) {
-    String rawValue = get(key);
-
+    String rawValue = get(key).toUpperCase();
```
maobaolong, line 312: Not sure is there some corner case that the member of T enumType is lowerCase or mix lowerCase and upperCase?

apc999: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

enumType should be in upper case as they are constant to follow Java convention. 
Also in here,  it only works for a few enum type propertyKey. I believe they are all in upper case

maobaolong: OK, sounds good.

maobaolong (approved):
LGTM, thanks for the backwards compatible improvement.

apc999:

PR #12981, opened 2021-03-03
Add NullCacheEvictor for debugging purpose

apc999:


calvinjia (approved):
Perhaps change the name to `UnevictableCacheEvictor`? From what I understand, returning `null` on `evict()` means we cannot find a page to evict.

core/client/fs/src/main/java/alluxio/client/file/cache/evictor/FIFOEvictor.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Override
+  public void updateOnDelete(PageId pageId) {
+    int idx = mQueue.indexOf(pageId);
+    if (idx >= 0) {
+      mQueue.remove(idx);
+    }
+  }
+
+  @Nullable
```
calvinjia, line 55: Do we need nullable since interface is already annotated?

apc999:

PR #13007, opened 2021-03-05
Move BlockWorker interface to core/common module

apc999:
This refactor will help https://github.com/Alluxio/alluxio/pull/12890/


LuQQiu:
@apc999 You may need to pull the latest master branch with my updated commit to revert the changes which introduce the test failures

apc999:
@LuQQiu  updated

LuQQiu:
@apc999 A high-level confusion is the role of  `alluxio-core-base` and `alluxio-core-common`.
In PR  #12838, i put all the BlockReadRequest, ReadRequest related concepts in `alluxio-core-base`.

From the previous idea of introducing `alluxio-core-base`, it act as a single definition of option classes shared b/w client and server modules. Define interfaces for all services.  (which i think is the role of `alluxio-core-common`, not knowing why we introduce it)

Not sure BlockWorker related interfaces will be better to go in `alluxio-core-base` or `alluxio-core-common`.

core/common/src/main/java/alluxio/worker/block/BlockStoreMeta.java
```diff
@@ -17,50 +17,19 @@
...
-     * @param manager the BlockMetadata Manager
-     * @return BlockStoreMeta instance
-     */
-    public static BlockStoreMeta createFull(BlockMetadataManager manager) {
-      return new DefaultBlockStoreMeta(manager, true);
-    }
-  }
   /**
-   * Note: This is only available in {@link BlockStoreMeta.Factory#createFull}.
+   * Note: This is only available when blocks are included.
```
LuQQiu, line 25: Not very clear what when blocks are included mean?

LuQQiu (approved):
LGTM
the new structure is clear, maybe run some sanity tests to make sure nothing breaks.
The only left issue is the `alluxio-core-base` vs `alluxio-core-common` issue. Maybe @calvinjia @madanadit have better understanding of whether those interfaces should be moved to?

core/common/src/main/java/alluxio/worker/block/BlockStoreMeta.java
apc999: @LuQQiu  agree. I may submit another PR to improve it. But I didn't do it here to avoid further increase the footprint of this PR

apc999:
> The only left issue is the `alluxio-core-base` vs `alluxio-core-common` issue. Maybe @calvinjia @madanadit have better understanding of whether those interfaces should be moved to?

I believe there is no real reason to keep both `alluxio-core-base` & `alluxio-core-common`, as every module depending on `alluxio-core-base` also depends on `alluxio-core-common`. @madanadit @calvinjia  what's your thought

apc999:

PR #13017, opened 2021-03-07
Combine core/base with core/common

apc999:
`alluxio-core-base` is not really needed - as all modules depending on `alluxio-core-base` also depends on `alluxio-core-common`. Thus merging these two modules.


core/common/src/main/java/alluxio/util/network/NetworkAddressUtils.java
```diff
@@ -667,13 +665,13 @@ public static SocketAddress getDataPortSocketAddress(WorkerNetAddress netAddress
...
    * Test if the input address is serving an Alluxio service. This method make use of the
    * gRPC protocol for performing service communication.
+   * This methods throws UnauthenticatedException if the user is not authenticated,
+   * StatusRuntimeException If the host not reachable or does not serve the given service.
    *
    * @param address the network address to ping
    * @param serviceType the Alluxio service type
    * @param conf Alluxio configuration
    * @param userState the UserState
-   * @throws UnauthenticatedException If the user is not authenticated
```
madanadit, line 675: why this change? I understand `StatusRuntimeException` might be remove grpc dependency?

madanadit (commented):
Thanks a lot Bin! Left 1 minor comment.

core/common/src/main/java/alluxio/util/network/NetworkAddressUtils.java
apc999: it seems we hit some maven plugin bug that with these two `@throws`, style check will fail

madanadit (approved):

apc999:

PR #13043, opened 2021-03-10
Do not always load metadata on webui browse

bradyoo:


core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -459,7 +459,7 @@ public Response getWebUIBrowse(@DefaultValue("/") @QueryParam("path") String req
         }
 
         filesInfo = mFileSystemMaster.listStatus(currentPath, ListStatusContext.mergeFrom(
-            ListStatusPOptions.newBuilder().setLoadMetadataType(LoadMetadataPType.ALWAYS)));
+            ListStatusPOptions.newBuilder()));
```
yuzhu, line 462: maybe use `ListStatusContext.defaults()` ?

yuzhu (commented):
@bradyoo Minor nit. 
Ideally we should be using the stream based listStatus call here since we paginate on the web Ui side as well. Otherwise the web ui will wait forever for the listStatus to fetch all the directories. but we can address that in a separate PR.

yuzhu:
is the test failure relevant?

bradyoo:

bradyoo:
@yuzhu Should we cherry-pick this? if so, how far back?

PR #13047, opened 2021-03-11
Remove dependency on Java JPS

peterroelants:
[`jps`](https://docs.oracle.com/javase/7/docs/technotes/tools/share/jps.html) is part of the JDK, but not of the JRE. By removing this dependency and only relying on `ps` Alluxio FUSE can be run with Java JRE, which is much lighter than the JDK.

Another reason to remove the `jps` depencency is that the jps documentation mentions that:
> _NOTE: This utility is unsupported and may not be available in future versions of the JDK._

I tested this with both OpenJDK 8 JRE and OpenJDK 11 JRE.

### DIFF
--- a/integration/fuse/bin/alluxio-fuse
+++ b/integration/fuse/bin/alluxio-fuse
@@ -82,11 +82,7 @@ umount_fuse() {
 }
 
 fuse_stat() {
-  local jps="jps"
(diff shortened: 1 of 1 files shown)


apc999 (approved):

apc999:
Thanks @peterroelants  for the improvement!

apc999:

PR #13049, opened 2021-03-12
Reuse refcount=0 lock in the LockPool

### LINKED ISSUE #13048: Reuse lock resource in lockPool to reduce high frequency new object

Now, alluxio LockPool is a great feature to reduce heap space while invent fine grain lock, but it can new lock resource and drop it in a high frequency.

Use Resource pool to solve this issue.

Add a Resource pool in the lockPool, let's say it `resuePool`.

**Urgency**
Normal

**Additional context**
NO

maobaolong:
Fix #13048

### DIFF
--- a/core/common/src/main/java/alluxio/collections/LockPool.java
+++ b/core/common/src/main/java/alluxio/collections/LockPool.java
@@ -270,6 +270,9 @@ public class LockPool<K> implements Closeable {
         return v;
       }
       Resource res = mResueLockPool.acquireWithoutBlocking();
+      if (res == null) {
+        res = new Resource(mDefaultLoader.get());
+      }
       res.reset();
       return res;
     });
--- a/core/common/src/main/java/alluxio/collections/LockPool.java
+++ b/core/common/src/main/java/alluxio/collections/LockPool.java
@@ -269,7 +269,7 @@ public class LockPool<K> implements Closeable {
         v.mIsAccessed = true;
         return v;
       }
-      Resource res = mResueLockPool.acquire();
+      Resource res = mResueLockPool.acquireWithoutBlocking();
       res.reset();
       return res;
     });
--- a/core/common/src/main/java/alluxio/collections/LockPool.java
+++ b/core/common/src/main/java/alluxio/collections/LockPool.java
@@ -16,6 +16,7 @@ import alluxio.concurrent.LockMode;
 import alluxio.resource.LockResource;
 import alluxio.resource.RWLockResource;
 import alluxio.resource.RefCountLockResource;
+import alluxio.resource.ResourcePool;
 import alluxio.util.ThreadFactoryUtils;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


core/common/src/main/java/alluxio/collections/LockPool.java
```diff
@@ -256,7 +269,9 @@ private Resource getResource(K key) {
         v.mIsAccessed = true;
         return v;
       }
-      return new Resource(mDefaultLoader.apply(k));
+      Resource res = mResueLockPool.acquire();
```
apc999, line 272: the behavior will be changed here, right? if we run out of the `mReuseLockPool`, will this `getResource` block?

maobaolong: Yeah, it could be blocked

maobaolong:
@apc999 Thank you for your review. 

I add a new api `acquireWithoutBlocking` to the `ResourcePool`, so that it can acquire object without blocking.

core/common/src/main/java/alluxio/resource/ResourcePool.java
```diff
@@ -141,6 +141,22 @@ public T acquire(long time, TimeUnit unit) {
...
+   * @return a resource taken from the pool
+   */
+  public T acquireWithoutBlocking() {
+    // Try to take a resource without blocking
+    T resource = mResources.poll();
+    if (resource != null) {
+      return resource;
+    }
+    return createNewResource();
+  }
```
apc999, line 158: Move the creation logic to caller side?
In this way, this `ResourcePool` provides a more isolated and independent function, just dealing with a pool of resources with different ways to acquire these resource (blocking or non-blocking)

```suggestion
  @Null
  public T acquireWithoutBlocking() {
    // Try to take a resource without blocking
    T resource = mResources.poll();
    if (resource != null) {
      return resource;
    }
    return null;
  }
```

core/common/src/main/java/alluxio/collections/LockPool.java
```diff
@@ -58,7 +59,18 @@
   private static final String EVICTOR_THREAD_NAME = "LockPool Evictor";
 
   private final Map<K, Resource> mPool;
-  private final Function<? super K, ? extends ReentrantReadWriteLock> mDefaultLoader;
+  // TODO(baoloongmao): make it configurable before merge this PR.
```
apc999, line 62: it's fine to hard code this constant for now.
unless you feel strong there is a need to configure this parameter in practise

core/common/src/main/java/alluxio/collections/LockPool.java
```diff
@@ -256,7 +269,9 @@ private Resource getResource(K key) {
         v.mIsAccessed = true;
         return v;
       }
-      return new Resource(mDefaultLoader.apply(k));
+      Resource res = mResueLockPool.acquireWithoutBlocking();
```
apc999, line 272: Once we change the `acquireWithoutBlocking` based on my suggestion, we should have it here:

```java
Resource res = mResueLockPool.acquireWithoutBlocking();
if (res == null) {
  res = mDefaultLoader.get();
}

core/common/src/main/java/alluxio/collections/LockPool.java
maobaolong: OK

core/common/src/main/java/alluxio/collections/LockPool.java
maobaolong: Done

apc999 (approved):

apc999:

PR #13084, opened 2021-03-16
Fix file's owner and group out of sync in activeSync

fuchaohong:


gpang (commented):
@fuchaohong Thanks for this fix! Could you add a simple unit test which exercises the fix?

yuzhu:
already fixed in #13805

PR #13102, opened 2021-03-19
[SMALLFIX] Hide test only property

jiacheliu3:
Property `alluxio.user.block.remote.read.buffer.size.bytes` is only used by tests. Hide it in the doc.


yuzhu:
does this property not have performance implications？ if so, does it make sense to hide it?

jiacheliu3:
Screen Shot 2021-03-20 at 11 24 12 AM
This is what I see from the code. AFAIS it's only used by a few tests now. I'm not sure if it is used to be used more.

yuzhu (approved):

jiacheliu3:

PR #13104, opened 2021-03-19
Add resource loading fallback logic to ExtensionClassLoader

bf8086:
Some extensions perform resource loading during initialization which can fail in the URLClassLoader logic. This change adds a fallback logic to look for resource in the default class loader.

### DIFF
--- a/core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
+++ b/core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
@@ -90,6 +90,7 @@ public class ExtensionsClassLoader extends URLClassLoader {
     Enumeration<URL> resources = super.getResources(name);
     // Falls back to default class loader if not found in the URL class loader
     if (!resources.hasMoreElements()) {
+      LOG.debug("Falling back to default class loader for loading resource {}", name);
       return mDefaultClassloader.getResources(name);
     }
     return resources;
--- a/core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
+++ b/core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
@@ -14,8 +14,10 @@ package alluxio.extensions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.Enumeration;
 
 /**
  * An isolated {@link ClassLoader} for loading extensions to core Alluxio. This class loader first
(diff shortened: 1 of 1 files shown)


bf8086:
@calvinjia PTAL.

@ns1123 FYI.

core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
```diff
@@ -82,4 +84,14 @@ public ExtensionsClassLoader(URL[] urls, ClassLoader defaultClassLoader) {
       return mDefaultClassloader.loadClass(name, resolve);
     }
   }
+
+  @Override
+  public Enumeration<URL> getResources(String name) throws IOException {
+    Enumeration<URL> resources = super.getResources(name);
```
calvinjia, line 90: What's the implication of `super.getResources` returning no resources but `mDefaultClassloader` having the requested resource? Is it possible to get the wrong resource if they have resources of the same name?

core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
```diff
@@ -82,4 +84,14 @@ public ExtensionsClassLoader(URL[] urls, ClassLoader defaultClassLoader) {
...
+    // Falls back to default class loader if not found in the URL class loader
+    if (!resources.hasMoreElements()) {
+      return mDefaultClassloader.getResources(name);
```
calvinjia, line 93: Is there a rare enough/significant enough event that we want to log something?

calvinjia (changes requested):

core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
bf8086: Since it can be used by 3rd party component we can't really assume the pattern/frequency. I can add a debug log in case we want to track it.

core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
bf8086: This can be used by `ServiceLoader` for loading service classes. 

Theoretically this does gives more resources comparing to the original implementation. But I think it aligns with the purpose of this class and is similar to what other methods do in this class. 

I did run it through UFS related tests for HDFS and S3 and haven't observe any issue.

core/common/src/main/java/alluxio/extensions/ExtensionsClassLoader.java
calvinjia: Got it, let's leave something at DEBUG level at least so we have insight if something is unexpected. If you think the frequency is likely minimal, we can bump it up to INFO.

calvinjia (approved):

bf8086:

PR #13107, opened 2021-03-21
Add replay statistics for journal startup

ZacBlanco:
This change adds some new logging messages to the master while
replaying the  journal. Specifically, it will now log the rate of
journal entries applied to the state machine during the startup process.
The number of entries in a particular interval are logged along with the
current commit index and the expected entries remaining and estimated
time left.

In the embedded journal Instead, we use the Ratis "Commit Index"
which represents batches of journal entries applied to the raft log. We are
able to get the last commit index as well as track the most recently
applied index which gives us the ability to provide the estimated time
remaining. The last SN is more readily available in the UFSJournal
since we control that implementation.

A new function for the UfsJournalReader needed to be added in order to
find the highest sequence number is within a particular journal. This
allows us to get the estimated entries remaining and time left during
the replay. The method should take O(N) with the number of journal
log files. It does not read data from the log files. In the case where
there is an error determining the highest SN, an empty optional is
returned in which case the information about expected finish time and
number of entries remaining is omitted.

There is a case where an error occurs when retrieving the final commit
index. In this case, the optional variable with the index will not be
populated and the estimated time remaining statistics will not be
calculated, only the measured rate of entries being applied.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -583,9 +583,9 @@ public class RaftJournalSystem extends AbstractJournalSystem {
       // If for some reason there is more than one..it should be fine as it only
       // affects the completion time estimate in the logs.
       synchronized (this) { // synchronized to appease findbugs; shouldn't make any difference
+        RaftPeerId serverId = mServer.getId();
         Optional<RaftProtos.CommitInfoProto> commitInfo = getGroupInfo().getCommitInfos().stream()
-            .filter(commit ->
-                mServer.getId().equals(RaftPeerId.valueOf(commit.getServer().getId())))
+            .filter(commit -> serverId.equals(RaftPeerId.valueOf(commit.getServer().getId())))
             .findFirst();
         if (commitInfo.isPresent()) {
           endCommitIndex = OptionalLong.of(commitInfo.get().getCommitIndex());
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -81,9 +81,9 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Set;
-import java.util.StringJoiner;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -579,8 +579,16 @@ public class RaftJournalSystem extends AbstractJournalSystem {
         WaitForOptions.defaults().setTimeoutMs(10 * Constants.MINUTE_MS));
     OptionalLong endCommitIndex = OptionalLong.empty();
     try {
-      endCommitIndex = getGroupInfo().getCommitInfos().stream()
-          .mapToLong(RaftProtos.CommitInfoProto::getCommitIndex).min();
+      // raft peer IDs are unique, so there should really only ever be one result.
+      // If for some reason there is more than one..it should be fine as it only
+      // affects the completion time estimate in the logs.
+      Optional<RaftProtos.CommitInfoProto> commitInfo = getGroupInfo().getCommitInfos().stream()
+              .filter(commit ->
+                  mServer.getId().equals(RaftPeerId.valueOf(commit.getServer().getId())))
+              .findFirst();
+      if (commitInfo.isPresent()) {
+        endCommitIndex = OptionalLong.of(commitInfo.get().getCommitIndex());
+      }
     } catch (IOException e) {
       LogUtils.warnWithException(LOG, "Failed to get raft log information before replay."
           + " Replay statistics will not be available", e);
@@ -589,6 +597,8 @@ public class RaftJournalSystem extends AbstractJournalSystem {
     long lastMeasuredTime = startTime;
     long lastCommitIdx = 0L;
     long logCount = 0;
+    RaftJournalProgressLogger progressLogger =
+        new RaftJournalProgressLogger(mStateMachine, endCommitIndex);
 
     // Loop until we lose leadership or convince ourselves that we are caught up and we are the only
     // master serving. To convince ourselves of this, we need to accomplish three steps:
@@ -631,24 +641,7 @@ public class RaftJournalSystem extends AbstractJournalSystem {
             JOURNAL_STAT_LOG_MAX_INTERVAL_MS);
         if (ex instanceof LeaderNotReadyException
             && (now - lastMeasuredTime) > nextLogTime) {
-          logCount++;
-          long currCommitIdx = stateMachine.getLastAppliedCommitIndex();
-          long timeSinceLastMeasure = (now - lastMeasuredTime);
-          long commitIdxRead = currCommitIdx - lastCommitIdx;
-          double commitIdxRateS = 1000 * ((double) commitIdxRead) / timeSinceLastMeasure;
-          StringJoiner logMsg = new StringJoiner("|");
-          logMsg.add(String.format("current SN: %d", currCommitIdx));
-          logMsg.add(String.format("entries in last %dms=%d", timeSinceLastMeasure, commitIdxRead));
-          if (endCommitIndex.isPresent()) {
-            long commitsRemaining = endCommitIndex.getAsLong()
-                - stateMachine.getLastAppliedCommitIndex();
-            double expectedTimeRemaining = ((double) commitsRemaining) / commitIdxRateS;
-            logMsg.add(String.format("est. commits left: %d", commitsRemaining));
-            logMsg.add(String.format("est. time remaining: %.2fms", expectedTimeRemaining));
-          }
-          LOG.info(logMsg.toString());
-          lastMeasuredTime = now;
-          lastCommitIdx = currCommitIdx;
+          progressLogger.logProgress();
         } else if (!(ex instanceof LeaderNotReadyException)) {
           LOG.info("Exception submitting term start entry: {}", ex.toString());
         }
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -83,6 +83,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.OptionalLong;
 import java.util.Set;
+import java.util.StringJoiner;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -531,10 +532,6 @@ public class RaftJournalSystem extends AbstractJournalSystem {
         new JournalEntryCommand(entry).getSerializedJournalEntry()));
   }
 
-  static GroupInfoRequest toRaftMessage(org.apache.ratis.protocol.GroupInfoRequest request) {
-    return request;
-  }
-
   @Override
   public synchronized void checkpoint() throws IOException {
     // TODO(feng): consider removing this once we can automatically propagate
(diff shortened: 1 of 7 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -565,13 +568,24 @@ public synchronized void checkpoint() throws IOException {
    * The caller is responsible for detecting and responding to leadership changes.
    */
   private void catchUp(JournalStateMachine stateMachine, LocalFirstRaftClient client)
-      throws TimeoutException, InterruptedException {
+          throws TimeoutException, InterruptedException {
```
calvinjia, line 571: Spacing

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -565,13 +568,24 @@ public synchronized void checkpoint() throws IOException {
...
     // Wait for any outstanding snapshot to complete.
     CommonUtils.waitFor("snapshotting to finish", () -> !stateMachine.isSnapshotting(),
         WaitForOptions.defaults().setTimeoutMs(10 * Constants.MINUTE_MS));
+    OptionalLong endCommitIndex = OptionalLong.empty();
+    try {
+      endCommitIndex = getGroupInfo().getCommitInfos().stream()
+          .mapToLong(RaftProtos.CommitInfoProto::getCommitIndex).min();
+    } catch (IOException e) {
+      LOG.warn("Failed to get raft log information before replay."
+          + " Replay statistics will not be available:", e);
```
calvinjia, line 584: I think we want to avoid dumping a full stack trace on warn, this can be pretty concerning for admins when it is a recoverable issue.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -565,13 +568,24 @@ public synchronized void checkpoint() throws IOException {
...
     long startTime = System.currentTimeMillis();
     long waitBeforeRetry = ServerConfiguration.global()
         .getMs(PropertyKey.MASTER_EMBEDDED_JOURNAL_CATCHUP_RETRY_WAIT);
     // Wait for any outstanding snapshot to complete.
     CommonUtils.waitFor("snapshotting to finish", () -> !stateMachine.isSnapshotting(),
         WaitForOptions.defaults().setTimeoutMs(10 * Constants.MINUTE_MS));
+    OptionalLong endCommitIndex = OptionalLong.empty();
+    try {
+      endCommitIndex = getGroupInfo().getCommitInfos().stream()
+          .mapToLong(RaftProtos.CommitInfoProto::getCommitIndex).min();
```
calvinjia, line 581: Do we choose minimum commit index as an approximation, the set of quorum acknowledged messages can be further than the min commit index?

calvinjia (changes requested):
Could we expose the replay stats as metrics?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: Good point. I'll change this to `LogUtils.warn` which should just grab the exception message instead, unless debug logs are enabled

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: This is the minimum commit index of all the maximum commits indexes which exist between each journal in the quorum. I opted for the lower bound. It's possible that some nodes in the quorum may have a higher commit index. I think min() should still give a good approximation because we don't expect nodes in the quorum to be many commits behind? I actually don't know the behavior of the system if that occurs.

We can change this to `max()` if you feel using the pure maximum bound of all journals is more appropriate. I don't have an opinion either way

ZacBlanco:
> Could we expose the replay stats as metrics?

I think we _potentially_ could, but I think they would be of limited use since they would only really be valid during the startup phase of the master. The original ask was to implement these as log messages. Metrics could another way to expose this information but would likely require more thought and I would probably separate that into another PR/ticket.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: fixed

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: fixed

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: One case might affect the estimation is when there is a new master added to the quorum. The new node will probably get zero commit index initially and catch up slowly to the leader.

Would it be possible to use the current leader commit index? An elected leader is guaranteed to have the latest commit of the quorum.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -604,8 +621,37 @@ private void catchUp(JournalStateMachine stateMachine, LocalFirstRaftClient clie
...
+          long currCommitIdx = stateMachine.getLastAppliedCommitIndex();
+          long timeSinceLastMeasure = (now - lastMeasuredTime);
+          long commitIdxRead = currCommitIdx - lastCommitIdx;
+          double commitIdxRateS = 1000 * ((double) commitIdxRead) / timeSinceLastMeasure;
+          StringJoiner logMsg = new StringJoiner("|");
+          logMsg.add(String.format("current SN: %d", currCommitIdx));
+          logMsg.add(String.format("entries in last %dms=%d", timeSinceLastMeasure, commitIdxRead));
+          if (endCommitIndex.isPresent()) {
+            long commitsRemaining = endCommitIndex.getAsLong()
+                - stateMachine.getLastAppliedCommitIndex();
```
bf8086, line 644: Would this be negative after we submit a new catchup entry?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: That's a good point. I will look into whether or not I can get the commit index for the current leader. I tried finding this but if I recall the API surface is quite limited getting information out of ratis which is why I've implemented this way.

If I can't find a way to get the current leader index I'll change this to max() to use the upper-bound for time estimation

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
```diff
@@ -168,12 +174,57 @@ public long getNextSequenceNumber() {
...
+          double remainingTime = ((double) remaining) / entryRateS;
+          if (remaining > 0) {
+            logMsg.add(String.format("est. entries left: %d", remaining));
+          }
+          if (!Double.isNaN(remainingTime) && !Double.isInfinite(remainingTime)) {
+            logMsg.add(String.format("est. time remaining %.2fms", remainingTime));
+          }
+        }
+
+        LOG.info(logMsg.toString());
```
bf8086, line 207: Would it be possible to extract the calculation and logging of the replay stats to a util method and reuse it in both UFS and embedded journals?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalReader.java
```diff
@@ -270,4 +272,30 @@ public State advance() throws IOException {
...
+    try (UfsJournalReader reader = new UfsJournalReader(journal, 0, false)) {
+      reader.updateInputStream();
+      while (!reader.mFilesToProcess.isEmpty()) {
+        UfsJournalFile file = reader.mFilesToProcess.poll();
+        endSN = Math.max(endSN, file.getEnd());
+        reader.mInputStream = new JournalInputStream(file, reader.mUfs);
+        reader.updateInputStream();
+      }
+    } catch (IOException e) {
+      LOG.warn("Failed to get last SN from journal: ", e);
```
bf8086, line 296: (nit) use `LogUtils.warnWithException` instead?

bf8086 (changes requested):
Overall LGTM. I left some minor comments inline.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: The catchup entry value != commit index. The commit indexes come from ratis and as far as I can tell are not negative

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
ZacBlanco: Hmm..I can try this. It might be messy though :upside_down_face:

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalReader.java
ZacBlanco: updated

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournalCheckpointThread.java
ZacBlanco: added

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: I addressed this issue in the latest commit. You can see the new logic here:
https://github.com/Alluxio/alluxio/pull/13107/commits/99fe56c3510e89a8085f7270acaa5c46d1ae9f02#diff-37edc9788907ad25cdb171ff28732c44ba2bc66325be1a97f573b71beadfd092R585

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Sorry about the confusion. I was thinking about the whole expression `endCommitIndex.getAsLong() - stateMachine.getLastAppliedCommitIndex()` being negative due to the LastAppliedCommitIndex move beyond initial endCommitIndex with multiple catchup entries submitted in the while loop.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -572,6 +577,28 @@ private void catchUp(JournalStateMachine stateMachine, LocalFirstRaftClient clie
...
+      // raft peer IDs are unique, so there should really only ever be one result.
+      // If for some reason there is more than one..it should be fine as it only
+      // affects the completion time estimate in the logs.
+      Optional<RaftProtos.CommitInfoProto> commitInfo = getGroupInfo().getCommitInfos().stream()
+              .filter(commit ->
+                  mServer.getId().equals(RaftPeerId.valueOf(commit.getServer().getId())))
+              .findFirst();
+      if (commitInfo.isPresent()) {
+        endCommitIndex = OptionalLong.of(commitInfo.get().getCommitIndex());
+      }
```
bf8086, line 591: (nit) shall we log a warning if `commitInfo` is not present?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: Oh, I see what you're saying. Yes, it's potentially negative if the journal replay takes that long.

How long do we have to replay until that really matters..well..let's do some math.

Max logging interval = 30sec
Journal replay rate = 20k/s (this is approximate from my local testing. I hit 28k, but let's say it's slower on a real cluster)

catchup entries submitted before we log a negative "commits remaining" (20k/s * 30s) = 600,000 entries
Say we submit 1 catchup entry/second (current default afaik)
Then it would take 600,000 seconds of journal replay before we logged a negative value one time. This translates to 166.66 hours of replay or about 6.9 days of  the master simply replaying the journal to start up. 

Frankly, I don't think its a huge issue if this is negative. I can add in some defensive statements just in case to prevent logging negative values. But generally the amount of time we would need to replay for this becoming an issue is somewhat ridiculous and I hope no one would ever let their journal get that large before taking a checkpoint.

ZacBlanco: updated

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: updated

bf8086 (approved):

ZacBlanco:
ping @calvinjia

calvinjia (approved):
Some metrics exposed would be nice, though I understand it isn't as straightforward as a counter / timer. I think this change itself is sufficiently helpful, we can do any other metrics separately.

calvinjia:

PR #13109, opened 2021-03-22
Add configuration to stop master process on demotion

ZacBlanco:
The code changes here are relatively simple. When the
FaultTolerantAlluxioMasterProcess changes state from primary to
secondary, then instead of switching states and continuing operation, we
should trigger the JVM to exit instead.

The reasoning behind this change is that the implementation to clean up
between state changes is not sound and leaves additional resources
hanging around which accumulate if enough state changes occur.
Defaulting to this behavior places the restart responsibility elsewhere,
but guarantees a clean slate and no resource leaks upon startup.

This is implemented with little change by simply placing a call to
stop() once the primary selector changes to the SECONDARY state
within the FTAMP. A simple new test is added to verify this behavior

The behavior by default is hidden behind the flag
alluxio.master.journal.exit.on.demotion. It is false by default and when
set to true, the masters will exhibit the behavior described above.

Additional considerations may want to be taken into account when this
is made the default behavior. We may want to consider additional
recommended guidelines for process supervision on bare metal machines
using systemd. K8s deployments should not be affected as the default
behavior is usually to restart the pod anyway.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
+++ b/core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
@@ -105,6 +105,11 @@ final class FaultTolerantAlluxioMasterProcess extends AlluxioMasterProcess {
         mLeaderSelector.waitForState(State.SECONDARY);
         if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_EXIT_ON_DEMOTION)) {
           stop();
+        } else {
+          if (!mRunning) {
+            break;
+          }
+          losePrimacy();
         }
       }
     }
--- a/core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
+++ b/core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
@@ -103,7 +103,9 @@ final class FaultTolerantAlluxioMasterProcess extends AlluxioMasterProcess {
       }
       if (gainPrimacy()) {
         mLeaderSelector.waitForState(State.SECONDARY);
-        stop();
+        if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_EXIT_ON_DEMOTION)) {
+          stop();
+        }
       }
     }
   }
--- a/core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
+++ b/core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
@@ -91,6 +91,9 @@ final class FaultTolerantAlluxioMasterProcess extends AlluxioMasterProcess {
     }
 
     while (!Thread.interrupted()) {
+      if (!mRunning) {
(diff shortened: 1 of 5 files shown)


calvinjia (commented):
Thanks for the fix! Should we update some of the documentation? Also, in Embedded mode, will this cause a problem in the quorum since a node will leave and need to rejoin every time leadership changes - if the user does not have something that brings the process back up, the quorum will be lost after a few failures (similar for ZK but ZK only need 1 node to be alive).

ZacBlanco:
I'll update the relevant parts of the documentation that I can locate. If you have particular places you think I missed you can point them out after I push the next commit.

Regarding your other questions:

>  Also, in Embedded mode, will this cause a problem in the quorum since a node will leave and need to rejoin every time leadership changes

When a node is shutdown (in this case because it's been demoted to a secondary) my presumption is that it's either planned (via some kind of `kill` or something errors in the master causing it to relinquish its primacy. In either case, I don't think we want the master to specifically leave the quorum as doing so could then leave an even number of nodes which leaves the potential for a split-brain cluster if more errors occur.

>  if the user does not have something that brings the process back up, the quorum will be lost after a few failures

Sudden failures or masters shutting down are part of operating the distributed cluster, so yes, if many shut down in a row we lose the quorum and the system will no longer operate until a majority of the nodes come back up. I think this behavior is acceptable.

calvinjia:
@ZacBlanco Got it. I think my main concern was around letting users know that they would need something to handle their master life cycles (and perhaps in a perfect world we would have something for them). Although they should already have something since processes do go down for various reasons, that isn't always the case.

ZacBlanco:
Per some offline discussion, this behavior is now hidden behind a configuration flag

ZacBlanco:
Documentation added in https://github.com/Alluxio/alluxio/pull/13109/commits/9fdc84658f11885f53b46f170d7c3349d6e59c31

calvinjia (approved):

core/server/master/src/main/java/alluxio/master/FaultTolerantAlluxioMasterProcess.java
```diff
@@ -100,10 +103,9 @@ public void start() throws Exception {
       }
       if (gainPrimacy()) {
         mLeaderSelector.waitForState(State.SECONDARY);
-        if (!mRunning) {
-          break;
+        if (ServerConfiguration.getBoolean(PropertyKey.MASTER_JOURNAL_EXIT_ON_DEMOTION)) {
+          stop();
         }
-        losePrimacy();
```
bf8086, line 106: Do we not want to call `losePrimacy()` any more?

ZacBlanco: good catch. Updated the logic

ZacBlanco:
ping @bf8086 . The test failures are unrelated to my changes (FUSE)

bf8086 (approved):
LGTM. Thanks!

ZacBlanco:
I don't have permission to merge. @calvinjia or @bf8086 could you merge this? Thanks!

calvinjia:

PR #13044, opened 2021-03-11
Support Fuse Mount with Worker to Reduce RPCs

apc999:
Fix https://github.com/Alluxio/alluxio/issues/12830

Co-authored-by: Lu Qiu <luqiujob@gmail.com>

PR #13114, opened 2021-03-23
Wait file closed before unmount fuse

Binyang2014:
Since release function is async, we need to wait for all out stream closed before umount the FUSE


Binyang2014:
@apc999 @LuQQiu Please help to review

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -466,14 +468,17 @@ private int releaseInternal(String path, FuseFileInfo fi) {
...
         synchronized (ce) {
           ce.getOut().close();
         }
       }
     } catch (Throwable e) {
       LOG.error("Failed closing {}", path, e);
       return -ErrorCodes.EIO();
+    } finally {
+      if (ce != null) {
+        mCreateFileEntries.remove(ce);
```
LuQQiu, line 480: why we need to do it in finally?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -657,6 +662,32 @@ public String getFileSystemName() {
     return mFsName;
   }
 
+  @Override
+  public void umount() {
+    // Release operation is async, we need to make sure
+    // all out stream is closed before umount the fuse
+    LOG.info("Waiting for all file out stream closed");
+    long startTime = System.currentTimeMillis();
+    while (!mCreateFileEntries.isEmpty()) {
```
LuQQiu, line 671: Can we use CommonUtils.waitFor() method?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -657,6 +662,32 @@ public String getFileSystemName() {
...
+      try {
+        if (System.currentTimeMillis() - startTime >= MAX_UMOUNT_WAITTIME_MS) {
+          throw new TimeoutException("reach the max umount waiting time");
+        }
+        Thread.sleep(100);
```
LuQQiu, line 676: Try not to use Thread.sleep() in the source code, thanks!

LuQQiu (commented):
Thanks for adding the logics! Left some comments

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: `CommonUtils.waitFor()` Only can wait one url, we need to wait all pending file out stream here

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: 1. Previous logic remove ce before call `ce.getOut().close();`, I move it after this function to make sure it will be cleaned after the RPC return. So I can check if all close ops are returned based on the size of `mCreateFileEntries`
2. To respect previous logic: ce will be removed no matter `ce.getOut().close();` be executed successfully or not, I put it into finally block

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: Fixed, thx

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -657,6 +664,26 @@ public String getFileSystemName() {
     return mFsName;
   }
 
+  @Override
+  public void umount() {
+    // Release operation is async, we need to make sure
+    // all out stream is closed before umount the fuse
+    LOG.info("Waiting for all file out stream closed");
+    long startTime = System.currentTimeMillis();
```
LuQQiu, line 672: Is this line still needed?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -657,6 +664,26 @@ public String getFileSystemName() {
     return mFsName;
   }
 
+  @Override
+  public void umount() {
+    // Release operation is async, we need to make sure
+    // all out stream is closed before umount the fuse
+    LOG.info("Waiting for all file out stream closed");
```
LuQQiu, line 671: Can we wrap the try{} block if a mCreateFileEntries.isEmpty() check? so that we can log the info when we actually need to wait?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: Got it, make sense!

LuQQiu (approved):
Overal LGTM, left some minor comments, thanks!

LuQQiu:
Fixes of the failure JNIFuse test is in https://github.com/Alluxio/alluxio/pull/13044, no need to worry in this PR

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: fixed

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: fixed, thx

apc999:
failing tests "WorkerFuseIntegrationTest>AbstractFuseIntegrationTest" have been fixed in master. 
it is safe to merge this PR

thanks for your contribution @Binyang2014 !

apc999:

PR #13115, opened 2021-03-23
Reduce default inode cache size

bradyoo:
Each Inode Cache entry takes up about 300 ~ 350 bytes. So a 10 million default max size is already 3 ~ 3.5 GB when full.

This is also not accounting for the EdgeCache which has the same default max size as inode cache and while smaller than inode cache in average. Isn't that much smaller.


calvinjia (commented):
Would it be possible to base this size off the configured JVM heap by default?

bradyoo:
Updated default to something a bit more dynamic.

bradyoo:
@calvinjia ping? The tests are sadly very flaky :(

calvinjia (approved):

calvinjia:

PR #13119, opened 2021-03-24
Convert AsyncCacheExecutor to use UniqueBlockingQueue 

yuzhu:
This is done so that duplicated tasks are not added to the queue, causing the queue to fill up and eventually start rejecting legitimate async caching requests.

### DIFF
--- a/core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
+++ b/core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
@@ -27,7 +27,11 @@ import java.util.concurrent.TimeUnit;
  */
 public class UniqueBlockingQueue<T> extends AbstractQueue<T> implements BlockingQueue<T> {
   private ConcurrentHashSet<T> mElementSet = new ConcurrentHashSet<>();
-  private BlockingQueue<T> mBlockingQueue = new LinkedBlockingQueue<>();
+  private BlockingQueue<T> mBlockingQueue;
+
+  public UniqueBlockingQueue(int capacity) {
+    mBlockingQueue = new LinkedBlockingQueue<>(capacity);
+  }
 
   @Override
   public synchronized void put(T e) throws InterruptedException {
--- a/core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
+++ b/core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
@@ -1,0 +1,149 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.util.executor;
+
+import alluxio.collections.ConcurrentHashSet;
+
+import java.util.AbstractQueue;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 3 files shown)


yuzhu:
@calvinjia I will continue to add more unit testing and some benchmarking results, but the PR itself is ready for review.

core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
```diff
@@ -0,0 +1,161 @@
...
+  private ConcurrentHashSet<T> mElementSet = new ConcurrentHashSet<>();
+  private BlockingQueue<T> mBlockingQueue;
+
+  /**
+   * Constructor for a UniqueBlockingQueue.
+   *
+   * @param capacity capacity of the blocking queue
+   */
+  public UniqueBlockingQueue(int capacity) {
+    mBlockingQueue = new LinkedBlockingQueue<>(capacity);
```
calvinjia, line 41: Should we initialize `mElementSet` here with the capacity as well? Although the concept of capacity (fixed vs initial) is different, I think the set is bounded by capacity as well.

core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @Override
+  public synchronized void put(T e) throws InterruptedException {
+    if (!mElementSet.contains(e)) {
+      mElementSet.add(e);
+      mBlockingQueue.put(e);
```
calvinjia, line 48: Do we need to clean up `mElementSet` if `put` throws an interrupted exception (or do this in the other order)?

core/server/worker/src/main/java/alluxio/worker/block/AsyncCacheRequestManager.java
```diff
@@ -233,6 +202,88 @@ private boolean cacheBlockFromRemoteWorker(long blockId, long blockSize,
...
+    @Override
+    public void run() {
+      boolean result = false;
+      long blockId = mRequest.getBlockId();
+      long blockLength = mRequest.getLength();
+      try {
+        boolean isSourceLocal = mLocalWorkerHostname.equals(mRequest.getSourceHost());
+        // Check if the block has already been cached on this worker
+        if (mBlockWorker.hasBlockMeta(mRequest.getBlockId())) {
+          ASYNC_CACHE_DUPLICATE_REQUESTS.inc();
```
calvinjia, line 258: Does our tracking of duplicate requests need to change now that we do deduplication at the queue level?

calvinjia (changes requested):
Thanks for the improvement! Had a few comments

core/server/worker/src/main/java/alluxio/worker/block/AsyncCacheRequestManager.java
yuzhu: even though we do dedup at the queue level, there could still be blocks cached either through explicit loading? I think this metric is much more meaningful now and reflects real issues if it is unusually high..

core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
yuzhu: good catch, fixed

yuzhu (commented):
@calvinjia PTAL

core/common/src/main/java/alluxio/util/executor/UniqueBlockingQueue.java
calvinjia: Fair point, let's keep it as is then.

core/server/worker/src/main/java/alluxio/worker/block/AsyncCacheRequestManager.java
calvinjia: I was thinking of preserving `ASYNC_CACHE_REQUESTS` = sum of all the other result request types, but I agree that this metric is a lot more useful than the old one. I'm fine either way.

calvinjia (approved):

yuzhu:

PR #13126, opened 2021-03-25
Handle client cache failure on "no space left on device"

apc999:
This happens when (1) cache is configured with wrong and smaller capacity, (2) for some reason disk capacity allocated to Alluxio was "stolen" by other applications, or (3) alluxio cache is under-counting the real space used on disk (e.g., due to bugs, inode overhead and etc).

This PR aim to
1. add a metrics to monitor and report when `no space left on device` happens
2. force evicting items in a retry in this case
3.  add a new flag `alluxio.user.client.cache.store.overhead` to allow users to set the space overhead when writing data to disk

Fix https://github.com/prestodb/presto/issues/15972

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
@@ -23,6 +23,7 @@ import alluxio.collections.Pair;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.PropertyKey;
 import alluxio.exception.PageNotFoundException;
+import alluxio.exception.status.ResourceExhaustedException;
 import alluxio.metrics.MetricKey;
 import alluxio.metrics.MetricsSystem;
 import alluxio.resource.LockResource;
@@ -217,6 +218,7 @@ public class LocalCacheManager implements CacheManager {
     OK,
     INSUFFICIENT_SPACE,
     BENIGN_RACING,
+    NO_SPACE,
     OTHER,
   }
 
@@ -300,6 +302,9 @@ public class LocalCacheManager implements CacheManager {
           // note that, we only evict one item a time in putAttempt. So it is possible the evicted
           // page is not large enough to cover the space needed by this page. Try again
           continue;
+        case NO_SPACE:
+          // fall through intentionally
+          // TODO(binfan): smart way to handle this case
         case OTHER:
           // fall through intentionally
         default:
@@ -348,9 +353,14 @@ public class LocalCacheManager implements CacheManager {
           mPageStore.put(pageId, page);
           Metrics.BYTES_WRITTEN_CACHE.mark(page.length);
           return PutResult.OK;
+        } catch (ResourceExhaustedException e) {
+          undoAddPage(pageId);
+          LOG.error("Failed to add page {} to pageStore: ", pageId, e);
+          Metrics.PUT_STORE_WRITE_NO_SPACE_ERRORS.inc();
+          return PutResult.NO_SPACE;
         } catch (IOException e) {
           undoAddPage(pageId);
-          LOG.error("Failed to add page {}: {}", pageId, e);
+          LOG.error("Failed to add page {} to pageStore: ", pageId, e);
           Metrics.PUT_STORE_WRITE_ERRORS.inc();
           return PutResult.OTHER;
         }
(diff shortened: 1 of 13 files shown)


core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -65,21 +66,24 @@ public LocalPageStore(LocalPageStoreOptions options) {
...
+        Files.createFile(p);
+      }
       // extra try to ensure output stream is closed
       try (FileOutputStream fos = new FileOutputStream(p.toFile(), false)) {
         fos.write(page);
       }
     } catch (Exception e) {
       Files.deleteIfExists(p);
+      if (e.getMessage().contains("No space left on device")) {
+        throw new ResourceExhaustedException("No space left on device");
```
calvinjia, line 85: Make the string a constant? Also is it guaranteed or is there a better way of referencing it?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -348,9 +353,14 @@ private PutResult putAttempt(PageId pageId, byte[] page, CacheScope scope, Cache
...
           return PutResult.OK;
+        } catch (ResourceExhaustedException e) {
+          undoAddPage(pageId);
+          LOG.error("Failed to add page {} to pageStore: ", pageId, e);
+          Metrics.PUT_STORE_WRITE_NO_SPACE_ERRORS.inc();
+          return PutResult.NO_SPACE;
         } catch (IOException e) {
           undoAddPage(pageId);
-          LOG.error("Failed to add page {}: {}", pageId, e);
+          LOG.error("Failed to add page {} to pageStore: ", pageId, e);
```
calvinjia, line 363: Is the notation to have `: ` before the exception, I don't think we do this in other places?

calvinjia (changes requested):

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
gpang, line 85: Would it be helpful to also state the device (`rootDir`) that is full?

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: I was using this convention in this class

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
apc999: done.
it is not guaranteed, this is best effort for monitoring and recover from this error

apc999: done

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -308,13 +318,14 @@ private boolean putInternal(PageId pageId, byte[] page, CacheScope scope, CacheQ
...
+    } else if (result == PutResult.INSUFFICIENT_SPACE_EVICTED) {
       Metrics.PUT_INSUFFICIENT_SPACE_ERRORS.inc();
     }
     return false;
   }
 
-  private PutResult putAttempt(PageId pageId, byte[] page, CacheScope scope, CacheQuota quota) {
+  private PutResult putAttempt(PageId pageId, byte[] page, CacheScope scope, CacheQuota quota,
+      boolean forcedToEvict) {
     LOG.debug("putInternal({},{} bytes) enters", pageId, page.length);
```
calvinjia, line 329: Should we also log `forcedToEvict` or have some metrics around that?

calvinjia (changes requested):
Had a minor comment

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
apc999: @calvinjia  tracked by metrics `PUT_STORE_WRITE_NO_SPACE_ERROR`

apc999:
@calvinjia  PTAL

calvinjia (approved):

apc999:

PR #13132, opened 2021-03-27
Add journal space monitor

ZacBlanco:
This new monitor tracks utilization of disks when Alluxio is
 is configured to use the embedded journal. The monitor logs messages
 when available disk space falls below a configurable percentage. The
 disk utilization metrics are exposed in the logs when the threshold is
 reached, but can also now be found as metrics as well as in the Alluxio
 master's web interface.

 Implementation details:
 The monitor is implemented as a heartbeat thread in the
 DefaultMetaMaster. It gets disk utilization by using the `df` command
 that is by default included with nearly all linux distributions, including
 within bare containers such as alpine, ubuntu, and centos. The command
 and arguments used are all POSIX-compatible, so this should be portable
 but for now is limited to Linux.

 For metrics, the actual disk values are not retrieved on every call to
 the metrics endpoint but rather they are only updated on the configured
 heartbeat interval.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2047,8 +2047,9 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey MASTER_JOURNAL_SPACE_MONITOR_INTERVAL =
       new Builder(Name.MASTER_JOURNAL_SPACE_MONITOR_INTERVAL)
       .setDefaultValue("10min")
-      .setDescription("How often to check and update information on space utilization of the"
-          + " journal disk. This is currently only compatible with linux-based systems")
+      .setDescription(String.format("How often to check and update information on space "
+          + "utilization of the journal disk. This is currently only compatible with linux-based"
+          + "systems and when %s is configured to EMBEDDED", Name.MASTER_JOURNAL_TYPE))
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2006,6 +2006,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_JOURNAL_SPACE_MONITOR_PERCENT_FREE_THRESHOLD =
+      new Builder(Name.MASTER_JOURNAL_SPACE_MONITOR_PERCENT_FREE_THRESHOLD)
+          .setDefaultValue(10)
+          .setDescription("When the percent of free space on any disk which backs the journal "
+              + "falls below this percentage, begin logging warning messages to let "
+              + "administrators know the journal disk(s) may be running low on space.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
   public static final PropertyKey MASTER_JOURNAL_TOLERATE_CORRUPTION =
       new Builder(Name.MASTER_JOURNAL_TOLERATE_CORRUPTION)
           .setDefaultValue(false)
@@ -2032,6 +2041,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDefaultValue("10MB")
           .setDescription("If a log file is bigger than this value, it will rotate to next "
               + "file.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_JOURNAL_SPACE_MONITOR_INTERVAL =
+      new Builder(Name.MASTER_JOURNAL_SPACE_MONITOR_INTERVAL)
+      .setDefaultValue("10min")
+      .setDescription("How often to check and update information on space utilization of the"
+          + " journal disk. This is currently only compatible with linux-based systems")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
@@ -5328,6 +5345,10 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     public static final String MASTER_JOURNAL_FOLDER = "alluxio.master.journal.folder";
     public static final String MASTER_JOURNAL_INIT_FROM_BACKUP =
         "alluxio.master.journal.init.from.backup";
+    public static final String MASTER_JOURNAL_SPACE_MONITOR_INTERVAL =
+        "alluxio.master.journal.space.monitor.interval";
+    public static final String MASTER_JOURNAL_SPACE_MONITOR_PERCENT_FREE_THRESHOLD
+        = "alluxio.master.journal.space.monitor.percent.free.threshold";
     public static final String MASTER_JOURNAL_TOLERATE_CORRUPTION
         = "alluxio.master.journal.tolerate.corruption";
     public static final String MASTER_JOURNAL_TYPE = "alluxio.master.journal.type";
--- a/core/common/src/main/java/alluxio/heartbeat/HeartbeatContext.java
+++ b/core/common/src/main/java/alluxio/heartbeat/HeartbeatContext.java
@@ -32,28 +32,29 @@ public final class HeartbeatContext {
   public static final String JOB_MASTER_LOST_WORKER_DETECTION = "Job Master Lost Worker Detection";
   public static final String JOB_WORKER_COMMAND_HANDLING =
       "Job Worker Command Handling";
-  public static final String MASTER_PERSISTENCE_CHECKER = "Master Persistence Checker";
-  public static final String MASTER_PERSISTENCE_SCHEDULER = "Master Persistence Scheduler";
-  public static final String MASTER_REPLICATION_CHECK = "Master Replication Check";
+  public static final String MASTER_ACTIVE_UFS_SYNC = "Master Active UFS Sync";
   public static final String MASTER_BLOCK_INTEGRITY_CHECK = "Master Block Integrity Check";
-  public static final String MASTER_CLUSTER_METRICS_UPDATER = "Master Cluster Metrics Updater";
   public static final String MASTER_CHECKPOINT_SCHEDULING = "Master Checkpoint Scheduling";
+  public static final String MASTER_CLUSTER_METRICS_UPDATER = "Master Cluster Metrics Updater";
   public static final String MASTER_DAILY_BACKUP = "Master Daily Backup";
   public static final String MASTER_FILE_RECOMPUTATION = "Master File Recomputation";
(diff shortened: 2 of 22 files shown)


core/common/src/main/java/alluxio/heartbeat/HeartbeatContext.java
```diff
@@ -32,28 +32,29 @@
...
-  public static final String MASTER_ACTIVE_UFS_SYNC = "Master Active UFS Sync";
+  public static final String MASTER_ORPHANED_METRICS_CLEANER = "Master Orphaned Metrics Cleaner";
+  public static final String MASTER_PERSISTENCE_CHECKER = "Master Persistence Checker";
+  public static final String MASTER_PERSISTENCE_SCHEDULER = "Master Persistence Scheduler";
+  public static final String MASTER_REPLICATION_CHECK = "Master Replication Check";
   public static final String MASTER_TABLE_TRANSFORMATION_MONITOR =
       "Master Table Transformation Monitor";
+  public static final String MASTER_TTL_CHECK = "Master TTL Check";
+  public static final String MASTER_UFS_CLEANUP = "Master Ufs Cleanup";
+  public static final String MASTER_UPDATE_CHECK = "Master Update Check";
```
ZacBlanco, line 57: Alphabetized all of the names here

ZacBlanco:
Sample warning message on main screen:
image

Sample table from metrics page:
image

ZacBlanco:
Adding @tieujason330 as a reviewer for the UI code

webui/master/src/containers/pages/Metrics/Metrics.tsx
```diff
@@ -290,6 +291,35 @@ export class MetricsPresenter extends React.Component<AllProps> {
             </Table>
           </div>
         ))}
+        {Object.keys(data.journalDiskMetrics).map((key: string) => (
```
tieujason330, line 294: Since `journalDiskMetrics` is an array, it might not be necessary to use `map` here since you're already creating the table rows starting line 307. Mapping here will cause a new table w/ header "Alluxio Master Journal Disk Status" to be created for each array item. You want to just create one table w/ rows mapped from the array right?

ZacBlanco: I think it would be better design to only have one "journal disk warning" and the rest of the rows not have the header. I was aware of the problem you pointed out before but I was a little lazy to make it better. I'll spend some time to implement a version which doesn't repeat the Message

ZacBlanco:
ping @bf8086

webui/master/src/containers/pages/Metrics/Metrics.tsx
ZacBlanco: Here's a screenshot after the latest commit (e20c835):
image

core/server/master/src/main/java/alluxio/master/meta/JournalSpaceMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+    // command allowed by the POSIX specification
+    CommandReturn output = getRawDiskInfo();
+    if (output.getExitCode() != 0) {
+      throw new IOException("Command failed with exit code " + output.getExitCode());
+    }
+    List<JournalDiskInfo> infos = Arrays.stream(output.getOutput().split("\n"))
+        .skip(1)
+        .map(infoLine -> Arrays.stream(infoLine.split(" "))
+              .map(String::trim).filter(s -> !s.isEmpty()).collect(Collectors.toList()))
+        .filter(data -> data.size() >= 6)
```
gpang, line 112: What does the output look like?

Also, I'm worried about parsing `df` output, because it could be fragile?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2035,6 +2044,14 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_JOURNAL_SPACE_MONITOR_INTERVAL =
+      new Builder(Name.MASTER_JOURNAL_SPACE_MONITOR_INTERVAL)
+      .setDefaultValue("10min")
+      .setDescription("How often to check and update information on space utilization of the"
+          + " journal disk. This is currently only compatible with linux-based systems")
```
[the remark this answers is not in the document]
ZacBlanco: No, this only applies to the embedded journal and when running on Linux. I can add this to the description

core/server/master/src/main/java/alluxio/master/meta/JournalSpaceMonitor.java
ZacBlanco: df output is not fragile. The output is actually defined by a POSIX standard: https://pubs.opengroup.org/onlinepubs/000095399/utilities/df.html

Additionally, it's part of coreutils which is generally unchanging. I think that generally, this should be safe to do. I looked into the coreutils source code for `df` and found that there's a lot of corner cases that the implementation accounts for that I would unlikely be able to do. Since `df` is included in essentially all distributions and is defined by the POSIX standard my conclusion was parsing the output should be safe.

webui/master/src/containers/pages/Metrics/Metrics.tsx
tieujason330: The `Alluxio Master Journal Disk Status` table (including table headers) will be rendered for each item in `data.journalDiskMetrics`. did you want it this way? or just to show a new table row per item?

ZacBlanco: Oh wait...my previous comment I was referring to the overview page. For some reason I didn't realize this comment was on the metrics page 🤦 I will fix this part tonight as well....

ZacBlanco: ook.. should be fixed now

core/common/src/main/java/alluxio/conf/PropertyKey.java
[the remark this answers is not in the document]
ZacBlanco: updated

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -424,6 +443,11 @@ public ConfigHash getConfigHash() {
     return new ConfigHash(ServerConfiguration.hash(), mPathProperties.hash());
   }
 
+  @Override
+  public Optional<JournalSpaceMonitor> getJournalSpaceMonitor() {
+    return mJournalSpaceMonitor == null ? Optional.empty() : Optional.of(mJournalSpaceMonitor);
```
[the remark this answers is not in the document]
ZacBlanco: huh. don't know why I wasn't aware of this method. Fixed.

tieujason330 (approved):
ui changes lgtm, thanks

gpang (approved):
LGTM

ZacBlanco:
@bf8086 can we get this merged?

webui/master/src/containers/pages/Metrics/Metrics.tsx
```diff
@@ -290,6 +291,33 @@ export class MetricsPresenter extends React.Component<AllProps> {
...
+                <th>Percent Available</th>
+                <th>Mount Path</th>
+              </tr>
+              {data.journalDiskMetrics.map((info: IJournalDiskInfo, idx: number) => {
+                return (
+                  <tr key={idx}>
+                    <td>{info.diskPath}</td>
+                    <td>{info.totalAllocatedBytes}</td>
+                    <td>{info.usedBytes}</td>
+                    <td>{info.availableBytes}</td>
```
bf8086, line 312: (nit) It would be nice to have a more human readable format (auto converting to MB,GB) when displaying the numbers. Maybe a follow up PR?

core/server/master/src/main/java/alluxio/master/meta/JournalSpaceMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+          .filter(info -> info.getPercentAvailable() < mWarnCapacityPercentThreshold)
+          .map(info -> String.format(
+                  "The journal disk %s backing the journal has only %.2f%% space left - %s",
+              info.getDiskPath(), info.getPercentAvailable(), info))
+          .collect(Collectors.toList());
+    } catch (IOException e) {
+      LogUtils.warnWithException(LOG, "Failed to get journal disk information. Critical warnings "
+          + "about journal disk space may not appear in the logs.", e);
+    }
+    return Collections.emptyList();
```
bf8086, line 164: Should we include the "failed to get journal disk info" warning in the returned warning list?

bf8086 (approved):
mostly LGTM. Left a couple minor comments.

core/server/master/src/main/java/alluxio/master/meta/JournalSpaceMonitor.java
ZacBlanco: I am personally not so sure. If returned, the warning would show up in the UI which might indicate to users there is a problem with the cluster; where in-fact, it could be a buggy problem in our code..maybe parsing the output, or something else weird with `df` on the system.

I feel it would be bad for the UX to show a warning/error in the case nothing is _actually_ wrong which is why I opted just for the log message originally, so that this message could be picked up by us, or some automated log analysis. I don't know what the best solution for the user would be in this situation. I just don't want to show a scary orange warning in the UI when nothing is actually wrong.

Thoughts?

webui/master/src/containers/pages/Metrics/Metrics.tsx
ZacBlanco: updated in latest commit:
image

@tieujason330 FYI since I changed some UI code

core/server/master/src/main/java/alluxio/master/meta/JournalSpaceMonitor.java
bf8086: I see, I was originally anticipate that the warning would mostly be environment issues not a bug in our code. If the warning message is something actionable to the admin then we should show it on UI for them. If we are not confident that this feature will work consistently then let's not include it for now. Let's revisit it some time later once we are more confident on the quality and update the UI if we see specific types of issues that are useful to the admin users.

ZacBlanco: I'm confident that this feature will work for our supported platforms (e.g. Linux -- CentOS 6+ and Ubuntu/Debian-based distros/containers). It's really impossible to test every environment this will be executed in. So if some distro isn't POSIX compliant with the `df` output spec, or maybe a transient error occurs executing `df` (potentially failed syscall, other system issue, etc), or maybe someone is using a home-baked container in k8s/docker using some weird base image is the case that I expect us to reach this particular `return`.

Those potentially transient errors or un-comformity to the POSIX specification is what would result in us hitting this point. I am confident in the logic which parses the output that _is_ POSIX-confirming. Since we can't possibly determine all scenarios which these errors might occur, and the fact that those errors may not be indicative of an issue with the system, which is why I am weary about showing a warning in the UI which may not necessitate any user action (or even be actionable by the user). If you think it's best to just put all the warnings in the UI regardless of the type of error that's ok. I just wanted to have a discussion since it _does_ affect the UX and isn't just some internal log message that we will look for in the future.

bf8086: Makes sense. Let's aim to improve the compatibility as usage grows and add UI warnings case by case if we find them useful.

bf8086:

PR #13139, opened 2021-03-30
[DOCFIX] Adjust Alluxio properties for HDFS

ZhuTopher:
Added the `alluxio.master.mount.table.root` prefix for clarity and a relative link to HDFS version mounting.

### DIFF
--- a/docs/en/ufs/HDFS.md
+++ b/docs/en/ufs/HDFS.md
@@ -65,11 +65,17 @@ address and the HDFS directory you want to mount to Alluxio. For example, the un
 can be `hdfs://localhost:8020` if you are running the HDFS namenode locally with default port and
 mapping HDFS root directory to Alluxio, or `hdfs://localhost:8020/alluxio/data` if only the HDFS
 directory `/alluxio/data` is mapped to Alluxio.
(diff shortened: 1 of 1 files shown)


Xenorith (approved):
do you know if there's any mention in this doc about using the `alluxio.master.mount.table.root.option.` property key prefix to set properties relating to the root mount? all your additions/fixes relate to using this prefix

ZhuTopher:
In the [Alluxio HDFS docs](https://docs.alluxio.io/os/user/stable/en/ufs/HDFS.html) there isn't really a mention of prefixing properties with `alluxio.master.mount.table.root.option` that I could find.

Notably the prefix is present in the (somewhat confusingly named) ["Using Site Properties"](https://docs.alluxio.io/os/user/stable/en/ufs/HDFS.html#using-site-properties) section while being absent from the ["Specify HDFS Configuration Location"](https://docs.alluxio.io/os/user/stable/en/ufs/HDFS.html#specify-hdfs-configuration-location) section.

Xenorith:
hmm well i do think it belongs somewhere more prominently but i'm not as familiar with all the implications. maybe @ns1123 can suggest a good place to mention the `alluxio.master.mount.table.root.option.` property key prefix. but it'd be a separate change anyways, so merging this

Xenorith:

PR #13146, opened 2021-03-31
Improve distributedLoad

apc999:
Improve efficiency of `distributedLoad` in two fold:
1. calling `BlockInStream.create` to obtain the InputStream without RPC, where the original  `blockStore.getOutStream(blockId,...)`  makes one RPC to master to get blockInfo.
2. removing using `BlockOutStream` and only reads the block data to alluxio job worker without writing it back to the block store on worker. This will further save at least two RPCs and the overhead to transfer data back to worker.

Note that, the original code is not removed to handle the case Pinned Media, because currently reading a block does not support promoting this block to a medium but only a tier. After discussed with @calvinjia , we will leave this once we consolidate medium and tier on worker

### DIFF
--- a/job/server/src/main/java/alluxio/job/util/JobUtils.java
+++ b/job/server/src/main/java/alluxio/job/util/JobUtils.java
@@ -11,11 +11,13 @@
 
 package alluxio.job.util;
 
+import alluxio.Constants;
 import alluxio.client.Cancelable;
 import alluxio.client.block.AlluxioBlockStore;
 import alluxio.client.block.BlockWorkerInfo;
 import alluxio.client.block.policy.BlockLocationPolicy;
 import alluxio.client.block.policy.LocalFirstPolicy;
+import alluxio.client.block.stream.BlockInStream;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.URIStatus;
 import alluxio.client.file.options.InStreamOptions;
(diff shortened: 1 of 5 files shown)


job/server/src/main/java/alluxio/job/util/JobUtils.java
```diff
@@ -128,13 +129,32 @@ public static void loadBlock(URIStatus status, FileSystemContext context, long b
...
+    if (pinnedLocation.isEmpty()) {
+      OpenFilePOptions openOptions =
+          OpenFilePOptions.newBuilder().setReadType(ReadPType.CACHE_PROMOTE).build();
+      InStreamOptions inOptions = new InStreamOptions(status, openOptions, conf);
+      BlockInfo info = status.getBlockInfo(blockId);
+      if (info == null) {
+        // we shall not reach here
+        throw new AlluxioException("Invalid block Id " + blockId + ", status " + status);
+      }
+      try (InputStream inputStream = BlockInStream.create(context, info, localNetAddress,
```
apc999, line 142: Use `BlockInStream.create` will further save one RPC compared to `blockStore.getOutStream` which fetches `blockInfo` from master

bradyoo:
What was the problem and how does this address it?

apc999:
> What was the problem (that the previous DistributedLoad PR didn't fix) and how does this address it?

I put the description https://github.com/Alluxio/alluxio/pull/13146#issue-605154372

> EDIT: also, according to https://app.codecov.io/gh/Alluxio/alluxio/compare/13146/diff#D1L135 . There are no tests that cover this line and the only test that would have covered this line has been modified to not cover it. Can you add a test to cover them?

Will do

bradyoo:
Sounds good. Originally, the description said it fixes the issue that was linked and I couldn't see how this could possibly do that.

apc999:
> Sounds good. Originally, the description said it fixes the issue that was linked and I couldn't see how this could possibly do that.

This is because i spin off the real fix to be a separate PR https://github.com/Alluxio/alluxio/pull/13151 from here to keep each change incremental. Sorry for the confusion

apc999:
@bradyoo  Added test coverage. PTAL

bradyoo:
LGTM after tests pass.

apc999:

PR #13151, opened 2021-03-31
Remove property of server-side buffer size when writing blocks

apc999:
Fix https://github.com/fluid-cloudnative/fluid/issues/633,
When property `alluxio.worker.file.buffer.size` is set to a large value (e.g., 320MB in that case),
`distributedLoad` or writing large files may fail often due to each block is allocating  such a big space (compared to much smaller block size 16MB really needed in https://github.com/fluid-cloudnative/fluid/issues/633)

On the other hand, the reserved bytes is calculated on block size today accurately and efficiently, 
https://github.com/Alluxio/alluxio/blob/c939e33a5ca3c4decd27fe7b6080a0feb82d3b9d/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataWriter.java#L131 
asking users to configure this property serves no purpose. Therefore we don't need to take the risk of misconfiguration by users.

This PR removed this property and hardcoded the value as 1MB.

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
@@ -11,7 +11,7 @@
 
 package alluxio.worker.grpc;
 
-import alluxio.conf.PropertyKey;
+import alluxio.Constants;
 import alluxio.conf.ServerConfiguration;
 import alluxio.exception.WorkerOutOfSpaceException;
 import alluxio.exception.status.NotFoundException;
@@ -52,8 +52,7 @@ import javax.annotation.concurrent.NotThreadSafe;
 public final class UfsFallbackBlockWriteHandler
     extends AbstractWriteHandler<BlockWriteRequestContext> {
   private static final Logger LOG = LoggerFactory.getLogger(UfsFallbackBlockWriteHandler.class);
-  private static final long FILE_BUFFER_SIZE =
-      ServerConfiguration.getBytes(PropertyKey.WORKER_FILE_BUFFER_SIZE);
+  private static final long FILE_BUFFER_SIZE = Constants.MB;
 
   /** The Block Worker which handles blocks stored in the Alluxio storage of the worker. */
   private final BlockWorker mWorker;
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
@@ -11,7 +11,6 @@
 
 package alluxio.worker.grpc;
 
-import alluxio.Constants;
 import alluxio.conf.ServerConfiguration;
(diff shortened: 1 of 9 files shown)


tests/src/test/java/alluxio/client/fs/io/UfsFallbackFileOutStreamIntegrationTest.java
```diff
@@ -87,12 +84,12 @@ protected void customizeClusterResource(LocalAlluxioClusterResource.Builder reso
   @Parameterized.Parameter(2)
   public int mUserFileBufferSize;
 
+  @Ignore
```
calvinjia, line 87: Should this be ignored?

core/server/worker/src/main/java/alluxio/worker/grpc/UfsFallbackBlockWriteHandler.java
```diff
@@ -52,8 +52,7 @@
 public final class UfsFallbackBlockWriteHandler
     extends AbstractWriteHandler<BlockWriteRequestContext> {
   private static final Logger LOG = LoggerFactory.getLogger(UfsFallbackBlockWriteHandler.class);
-  private static final long FILE_BUFFER_SIZE =
-      ServerConfiguration.getBytes(PropertyKey.WORKER_FILE_BUFFER_SIZE);
+  private static final long FILE_BUFFER_SIZE = Constants.MB;
```
calvinjia, line 55: Should this be a constant since we use the same value in both write handlers?

apc999: done, moved to `AbstractWriteHandler.java`

tests/src/test/java/alluxio/client/fs/io/UfsFallbackFileOutStreamIntegrationTest.java
apc999: this test is hard to fix, I'd like not spending time in fixing this test

apc999:
```
Error: 0.185 [ERROR] alluxio.client.fs.io.FileOutStreamIntegrationTest.cancelWrite[0]
Error: 0.185 [ERROR]   Run 1: FileOutStreamIntegrationTest.cancelWrite:248 expected:<0> but was:<3000>
Error: 0.186 [ERROR]   Run 2: FileOutStreamIntegrationTest.cancelWrite:248 expected:<0> but was:<2000>
```

tests/src/test/java/alluxio/client/fs/io/UfsFallbackFileOutStreamIntegrationTest.java
calvinjia: Does this mean there is regression introduced for UFS Fallback, and is it a functionality regression or an artifact of how we wrote this test (ie. expected certain buffer size)?

apc999: it is more related to how the test is written

apc999:

PR #13160, opened 2021-04-01
Fix Fuse write then read problem in async release and support umount

LuQQiu:
Previous commit https://github.com/Alluxio/alluxio/commit/8f618358891ca89f753b369316d58221e8fb9b74
enlarges the race condition window in a access pattern of `write()` - `release()` (async unfortunately) - `getAttr()`.

In `getAttr()`, a special logic is added to wait for the writing file to completed before getting file size information.
`getAttr()` can happen in any cases. When it's in `write()` - `getAttr()` - `write()` process, we don't want it to block waiting file completed. When it's written by other threads or it's in `write()` - `release()` - `getAttr()`, we do want it to block waiting file completed.

This PR introducing new concurrent hash maps for the releasing entries to fulfill the `getAttr()` logics and `umount` logics().

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -4611,6 +4611,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey FUSE_UMOUNT_WAITTIME =
+      new Builder(Name.FUSE_UMOUNT_WAITTIME)
+          .setDefaultValue("1min")
+          .setDescription("The maximum wait time to wait for all file in stream and out stream "
+              + "closed in Fuse umount() operation.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
+          .setScope(Scope.CLIENT)
+          .build();
   public static final PropertyKey FUSE_USER_GROUP_TRANSLATION_ENABLED =
       new Builder(Name.FUSE_USER_GROUP_TRANSLATION_ENABLED)
           .setDefaultValue(false)
@@ -5925,6 +5933,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         = "alluxio.fuse.shared.caching.reader.enabled";
     public static final String FUSE_LOGGING_THRESHOLD = "alluxio.fuse.logging.threshold";
     public static final String FUSE_MAXWRITE_BYTES = "alluxio.fuse.maxwrite.bytes";
+    public static final String FUSE_UMOUNT_WAITTIME =
+        "alluxio.fuse.umount.waittime";
     public static final String FUSE_USER_GROUP_TRANSLATION_ENABLED =
         "alluxio.fuse.user.group.translation.enabled";
 
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -4611,11 +4611,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.CLIENT)
           .build();
-  public static final PropertyKey FUSE_UMOUNT_WAITTIME =
-      new Builder(Name.FUSE_UMOUNT_WAITTIME)
+  public static final PropertyKey FUSE_UMOUNT_TIMEOUT =
+      new Builder(Name.FUSE_UMOUNT_TIMEOUT)
           .setDefaultValue("1min")
-          .setDescription("The maximum wait time to wait for all file in stream and out stream "
-              + "closed in Fuse umount() operation.")
+          .setDescription("The timeout to wait for all in progress file read and write to finish "
+              + "before unmounting the Fuse filesystem. After the timeout, "
+              + "all in progress file read will be forced to stop "
+              + "and all in progress file write will be abandoned.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.CLIENT)
           .build();
@@ -5933,8 +5935,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         = "alluxio.fuse.shared.caching.reader.enabled";
     public static final String FUSE_LOGGING_THRESHOLD = "alluxio.fuse.logging.threshold";
     public static final String FUSE_MAXWRITE_BYTES = "alluxio.fuse.maxwrite.bytes";
-    public static final String FUSE_UMOUNT_WAITTIME =
-        "alluxio.fuse.umount.waittime";
+    public static final String FUSE_UMOUNT_TIMEOUT =
+        "alluxio.fuse.umount.timeout";
     public static final String FUSE_USER_GROUP_TRANSLATION_ENABLED =
         "alluxio.fuse.user.group.translation.enabled";
 
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -27,6 +27,7 @@ import alluxio.grpc.CreateFilePOptions;
 import alluxio.grpc.SetAttributePOptions;
 import alluxio.jnifuse.AbstractFuseFileSystem;
 import alluxio.jnifuse.ErrorCodes;
+import alluxio.jnifuse.FuseException;
 import alluxio.jnifuse.FuseFillDir;
 import alluxio.jnifuse.struct.FileStat;
 import alluxio.jnifuse.struct.FuseContext;
@@ -731,37 +732,10 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
       }
     }
 
-    // Force close all opened stream
-    if (!mOpenFileEntries.isEmpty()) {
-      LOG.info("Unmounting {}. Force closing all remaining file read", mMountPoint);
-      for (Map.Entry<Long, FileInStream> entry : mOpenFileEntries.entrySet()) {
-        FileInStream is = entry.getValue();
-        try {
-          synchronized (is) {
-            // Files have a small chance to be double closed in umount() and release()
-            // but AlluxioFileInStream supports double closed.
-            is.close();
-          }
-        } catch (Throwable t) {
-          LOG.error("Failed to close opened file with id {} when unmounting {}",
-              entry.getKey(), mMountPoint, t);
-        }
-      }
-    }
-    if (!mCreateFileEntries.isEmpty()) {
-      LOG.info("Unmounting {}. Force abandoning all remaining file write", mMountPoint);
-      for (CreateFileEntry ce : mCreateFileEntries) {
-        try {
-          synchronized (ce) {
-            ce.getOut().cancel();
-            LOG.info("Abandoned file write with id {} and path {} when unmounting {}",
-                ce.getId(), ce.getPath(), mMountPoint);
-          }
-        } catch (Throwable t) {
-          LOG.error("Failed to abandon file write with id {} and path{} when unmounting {}: {}",
-              ce.getId(), ce.getPath(), mMountPoint, t);
-        }
-      }
+    if (!(mCreateFileEntries.isEmpty() && mOpenFileEntries.isEmpty())) {
+      // TODO(lu) consider the case that client application may not call release()
+      // for all open() or create(). Force closing those operations.
+      throw new FuseException("Cannot unmount Fuse, device or resource busy");
     }
     super.umount();
   }
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -12,7 +12,6 @@
 package alluxio.fuse;
 
 import alluxio.AlluxioURI;
-import alluxio.Constants;
 import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileOutStream;
 import alluxio.client.file.FileSystem;
@@ -64,7 +63,6 @@ import javax.annotation.concurrent.ThreadSafe;
 public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
     implements FuseUmountable {
   private static final Logger LOG = LoggerFactory.getLogger(AlluxioJniFuseFileSystem.class);
-  private static final int MAX_UMOUNT_WAITTIME_MS = Constants.MINUTE_MS;
   private final FileSystem mFileSystem;
   private final AlluxioConfiguration mConf;
   // base path within Alluxio namespace that is used for FUSE operations
@@ -72,12 +70,13 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
   // is /users/foo, then an operation on /mnt/alluxio/bar will be translated on
   // an action on the URI alluxio://<master>:<port>/users/foo/bar
   private final Path mAlluxioRootPath;
+  private final String mFsName;
   // Keeps a cache of the most recently translated paths from String to Alluxio URI
   private final LoadingCache<String, AlluxioURI> mPathResolverCache;
   private final LoadingCache<String, Long> mUidCache;
   private final LoadingCache<String, Long> mGidCache;
+  private final int mMaxUmountWaitTime;
   private final AtomicLong mNextOpenFileId = new AtomicLong(0);
-  private final String mFsName;
 
   private final Map<Long, FileInStream> mOpenFileEntries = new ConcurrentHashMap<>();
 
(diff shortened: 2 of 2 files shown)


integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -98,6 +103,10 @@ public String getFieldValue(CreateFileEntry o) {
       = new IndexedSet<>(ID_INDEX, PATH_INDEX);
   private final boolean mIsUserGroupTranslation;
 
+  // Map for holding the async releasing entries for proper umount
+  private final Map<Long, FileInStream> mReleasingReadEntries = new ConcurrentHashMap<>();
```
Binyang2014, line 107: Do we need to check `FileInStream` in release function? For in stream, the file should ready in `completed` status and no need extra action in server side

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -471,15 +480,27 @@ private int releaseInternal(String path, FuseFileInfo fi) {
...
-          is.close();
-        }
-      }
       if (ce != null) {
+        // Remove earlier to try best effort to avoid write() - async release() - getAttr()
+        // without waiting for file completed and return 0 bytes file size error
         mCreateFileEntries.remove(ce);
-        synchronized (ce) {
-          ce.getOut().close();
+        mReleasingWriteEntries.put(fd, ce);
```
Binyang2014, line 487: Seems the `fd` and `ce` will be kept in mReleasingWriteEntries until we umount the fuse daemon. Should we check and clean  the `mReleasingWriteEntries` periodically? Or move the logic of "put ce into mReleasingWriteEntries" in `create` function and remove it after close is done in server side

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -668,6 +689,68 @@ public String getFileSystemName() {
...
+          }
+        } catch (Throwable t) {
+          LOG.error("Failed to close opened file with id {} in umount operation",
+              entry.getKey(), t);
+        }
+      }
+    }
+    // Release operation is async, we need to make sure
+    // all out stream is closed before umount the fuse
+    if (!mCreateFileEntries.isEmpty()) {
```
Binyang2014, line 737: If we already have `mReleasingReadEntries`, why we need to check `mCreateFileEntries` here?

Binyang2014 (commented):
Thanks for this fix, several comments and questions about this PR

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: The fileInStream also needs to be closed to make sure resources are released, e.g. the opened RandomAccessFile and the FileChannel used for reading from local worker storage..

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: They will only be kept in `mReleasingWriteEntries` when they are releasing. It should be a short period. No need to clean it periodically.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: you may talk about why we need wait for mCreateFileEntries cleanup here?
The reason is that release() and umount() operation could be called by our users at any given time. 

The mReleasingWriteEntries, take care of the following sequence 
release() started - umount() - release() finished

The mCreateFileEntries wait take care of the following sequence
umount() - release() started and finished.

Same for mReleasingReadEntries and mOpenFileEntries

LuQQiu:
@Binyang2014 @apc999 updated the PR and replied to the comments, PTAL, thanks!

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -5925,6 +5933,8 @@ private static String javadocLink(String fullyQualifiedClassname) {
         = "alluxio.fuse.shared.caching.reader.enabled";
     public static final String FUSE_LOGGING_THRESHOLD = "alluxio.fuse.logging.threshold";
     public static final String FUSE_MAXWRITE_BYTES = "alluxio.fuse.maxwrite.bytes";
+    public static final String FUSE_UMOUNT_WAITTIME =
```
gpang, line 5936: I think the part `waittime` is not standard? Could this just be called `alluxio.fuse.umount.timeout`? That name looks like a timeout value for the umount, which is what we want right?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4611,6 +4611,14 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey FUSE_UMOUNT_WAITTIME =
+      new Builder(Name.FUSE_UMOUNT_WAITTIME)
+          .setDefaultValue("1min")
+          .setDescription("The maximum wait time to wait for all file in stream and out stream "
+              + "closed in Fuse umount() operation.")
```
gpang, line 4618: Could you describe what happens after this waiting period?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -668,6 +689,77 @@ public String getFileSystemName() {
     return mFsName;
   }
 
+  @Override
+  public void umount() {
+    // Release operation is async, we will try our best efforts to
+    // close all opened file in/out stream before umounting the fuse
+    if (!mCreateFileEntries.isEmpty() || !mOpenFileEntries.isEmpty()) {
+      LOG.info("Waiting for all file out stream closed");
```
gpang, line 697: Can we add something to this message, like the path we are umounting?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -668,6 +689,77 @@ public String getFileSystemName() {
...
+      try {
+        CommonUtils.waitFor("all opened file in stream and out stream closed",
+            () -> mCreateFileEntries.isEmpty() && mOpenFileEntries.isEmpty(),
+                WaitForOptions.defaults().setTimeoutMs(mMaxUmountWaitTime));
+      } catch (InterruptedException e) {
+        LOG.error("unmount interrupted");
```
gpang, line 703: Can we add the path we are umounting?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -668,6 +689,77 @@ public String getFileSystemName() {
...
+        Thread.currentThread().interrupt();
+      } catch (TimeoutException e) {
+        LOG.error("Timeout when waiting all file in stream and file out stream to close."
```
gpang, line 706: Can we add the path we are umounting? This should apply to other log messages in this method.

gpang (changes requested):
@LuQQiu Thanks! I left a few comments.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: For this change, I think `mReleasingWriteEntries` will only be cleaned when `umount` function is called.
In some deployment, fuse daemon will run as a service. So umount will no be called until admin stop the service. 
For this case, all closed file entries will be kept in `mReleasingWriteEntries`. If we read/write more than 100 million files (which is the common  for imageNet dataset),  then  `mReleasingWriteEntries` will consume much memory resource.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: Thanks. I will make my question more clearly. What I though is:
1. For writing operations, when close the `FileOutStream`. We should wait for the operation completed because  we need to make sure the resource is server (worker or master) side is released. Otherwise, file will be in an `uncompleted` status
2. For reading operations, the `FileInStream` only hold the resources in client side. After umount is called and fuse daemon is down. All resources used in client side will be released, so we don' t need to release the resource explicitly here. (Maybe server keeps grpc connection with the client, but the resource will be released after reach the timeout)

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: Thanks for the explanation, but I'm not sure when `umount` function will be called before `release` function.
If one process still hold a `fd`. Call `umount` will caused `failed to unmount /mnt/fuse: Device or resource busy`. Refer :https://man7.org/linux/man-pages/man2/umount2.2.html.

So we need to close the fd first then umount the fuse?

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Make sense. It's a timeout then waittime. Modified to timeout.

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Added the description, PTAL

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: Yeah, added

LuQQiu:
@gpang Addressed the comments, PTAL

gpang:
Looks good after other reviews!

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -668,6 +691,81 @@ public String getFileSystemName() {
...
+        Thread.currentThread().interrupt();
+      } catch (TimeoutException e) {
+        LOG.error("Timeout when waiting in progress file read/write closing to finish "
+            + "when unmounting {}. {} fileInStream and {} fileOutStream "
+            + "are still in closing process.",
+            mMountPoint, mReleasingReadEntries.size(), mReleasingWriteEntries.size());
+      }
+    }
+
+    // Force close all opened stream
```
apc999, line 734: I don't think we shall force close all opened streams. Might be better to 
`throw new FuseException("Device or resource busy")` as pointed out by @Binyang2014

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -471,15 +482,27 @@ private int releaseInternal(String path, FuseFileInfo fi) {
...
+        }
+      }
+      if (is != null) {
+        mReleasingReadEntries.put(fd, is);
+        try {
+          synchronized (is) {
+            is.close();
+          }
+        } finally {
+          mReleasingWriteEntries.remove(fd);
```
apc999, line 505: `mReleasingReadEntries.remove(fd)` ?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
apc999: @Binyang2014 ` mReleasingWriteEntries.remove(fd);` in line 495 after `ce.close()` completes

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -471,15 +482,27 @@ private int releaseInternal(String path, FuseFileInfo fi) {
...
+        // without waiting for file completed and return 0 bytes file size error
         mCreateFileEntries.remove(ce);
-        synchronized (ce) {
-          ce.getOut().close();
+        mReleasingWriteEntries.put(fd, ce);
+        try {
+          synchronized (ce) {
+            ce.close();
+          }
+        } finally {
```
apc999, line 494: shall we log error?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -471,15 +482,27 @@ private int releaseInternal(String path, FuseFileInfo fi) {
...
+          mReleasingWriteEntries.remove(fd);
+        }
+      }
+      if (is != null) {
+        mReleasingReadEntries.put(fd, is);
+        try {
+          synchronized (is) {
+            is.close();
+          }
+        } finally {
```
apc999, line 504: shall we log error?

apc999:
@LuQQiu  I left some comments. PTAL
Thanks @Binyang2014 for your review!

apc999 (changes requested):
comments left

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: We throw FuseException to not allow umount?
Not sure if it's correct. I think we should always allow umount, otherwise the fuse mount state may be strange or we cannot kill the process?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: We log in the outter catch function
```
    } catch (Throwable e) {
      LOG.error("Failed closing {}", path, e);
      return -ErrorCodes.EIO();
    }
```

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: It's possible that some open()/create() may not be properly release() as client operation crash?
In that case, i would not want to prevent umount from happening.
As for whether we want to force close opened streams, we can discuss that part.

apc999: @LuQQiu  in reality, umount a busy mount point is expect to fail there. This is often useful to help avoid data loss by disallowing unmouting a filesystem when it is being used.

apc999: maybe leave a TODO here to implement umount with force ?

apc999:
only comment left https://github.com/Alluxio/alluxio/pull/13160#discussion_r612907091

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: Sounds good. I will do more testing for it tomorrow morning. Feeling a little unsafe to throw the exception

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: Got it thanks. I miss this piece

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
Binyang2014: Actually, I think OS don't let we umount a busy device. Throw exception is good for me. Thanks

Binyang2014 (approved):

apc999 (approved):

apc999:

LuQQiu:
Using fuse on worker to test out the behaviors. I stopped the worker process when reading files from the fuse mount point. Umount will error out and the fuse application will be left in the operating system. Client read operation error out. The worker process killed.

Binyang2014:
Hi @LuQQiu Seem `alluxio-stop.sh worker` and `alluxio-fuse umount /mnt/fuse` will kill the fuse daemon. Even though we failed to umount.
Please try to use `umount <mount-point>` or `fusermount -u <mountpoint>` directly. From the error log, you will get
```
ERROR AbstractFuseFileSystem - Failed to umount /var/lib/kubelet/pods/.../mount
alluxio.jnifuse.FuseException: Unable to umount FS with exit code 1
        at alluxio.jnifuse.AbstractFuseFileSystem.umountInternal(AbstractFuseFileSystem.java:165)
        at alluxio.jnifuse.AbstractFuseFileSystem.umount(AbstractFuseFileSystem.java:121)
        at alluxio.fuse.AlluxioJniFuseFileSystem.umount(AlluxioJniFuseFileSystem.java:685)

```

From my test, umount will failed and we can still reading/writing data through the mount point

PR #13168, opened 2021-04-03
Reduce Mem Copy in client

apc999:


integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -352,11 +353,11 @@ public int read(String path, ByteBuffer buf, long size, long offset, FuseFileInf
   }
 
   private int readInternal(String path, ByteBuffer buf, long size, long offset, FuseFileInfo fi) {
+    final int sz = (int) size;
     int nread = 0;
-    int rd = 0;
     long fd = fi.fh.get();
     try {
-      FileInStream is = mOpenFileEntries.get(fd);
+      AlluxioFileInStream is = (AlluxioFileInStream) mOpenFileEntries.get(fd);
```
calvinjia, line 360: Is it worth doing this cast later and reverting to the old behavior so it is compatible with the `FileInStream` API?

calvinjia (commented):
Do you have any data on how much of an impact this makes? It seems to be at least 1 memory copy.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
apc999: done

LuQQiu (approved):
LGTM, thanks!

core/common/src/main/java/alluxio/network/protocol/databuffer/NioDataBuffer.java
```diff
@@ -68,7 +68,14 @@ public void readBytes(OutputStream outputStream, int length) throws IOException
 
   @Override
   public void readBytes(ByteBuffer outputBuf) {
-    outputBuf.put(mBuffer);
+    if (mBuffer.remaining() <= outputBuf.remaining()) {
+      outputBuf.put(mBuffer);
+    } else {
+      int oldLimit = mBuffer.limit();
```
gpang, line 74: This looks tricky. Could you comment why we have to do this? It looks like we change the limit of the output, write the data, then change the limit back? Is that safe to do for the output?

gpang (commented):
@apc999 Just a minor question/comment.

core/common/src/main/java/alluxio/network/protocol/databuffer/NioDataBuffer.java
apc999: in the case where `outputBuf` is smaller than the this databuffer,
it will fail when `outputBuf.put(mBuffer);`. 
So we need to tentatively make sure `outputBuf.put(mBuffer);` can proceed

apc999:

PR #13169, opened 2021-04-03
Skip caching large files

yabola:
Memory is relatively small, in order to avoid misreading large files, resulting in frequent memory swaps and improve overall cache hit rate , I want to have a configuration to limit caching behavior. 
For example, I have five 2G files and one 5G file in 10G  of  memory space. If Alluxio does not cache large files, it will increase the overall hit rate.

It can also limit the threshold of small file size as well (I will add such configuration later).If this PR makes sense I can add UT later.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
@@ -50,6 +50,7 @@ import alluxio.grpc.ListStatusPOptions;
 import alluxio.grpc.LoadMetadataPType;
 import alluxio.grpc.MountPOptions;
 import alluxio.grpc.OpenFilePOptions;
+import alluxio.grpc.ReadPType;
 import alluxio.grpc.RenamePOptions;
 import alluxio.grpc.ScheduleAsyncPersistencePOptions;
 import alluxio.grpc.SetAclAction;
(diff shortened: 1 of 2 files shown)


jiacheliu3:
@apc999 @calvinjia What do you think about this property?

jiacheliu3:
@yabola Could you add more context why you need this property and how you are using it?

yabola:
> @yabola Could you add more context why you need this property and how you are using it?

I want to add a feature to kylin(our product) that can load data into Alluxio in advance when  building data is completed (not when querying) to improve overall query speed . But I only want to cache the appropriate size of data. So I want to add this property in Alluxio.This is just my idea.

yabola:
> @yabola Could you add more context why you need this property and how you are using it?

This is just my thought.I am not sure if it makes sense in normal environment.I think it will be useful when memory is small.

calvinjia:
Alluxio already caches at a block level, so a 5 GB file and a 2 GB file will likely be cached at the same granularity (ie. in 128 MB blocks). Our cache policy (`alluxio.user.client.cache.evictor.class`) will handle determining which blocks should be cached when there is contention.

jiacheliu3:
In my opinion, it makes sense sometimes you want to skip caching certain files. For example if you already know that the big files belong to time-insensitive batch jobs and small files belong to real time operations, you want to save the cache to the small files. 

However, it's arguable if telling from the file size is the best way to decide whether we want to cache a file or not. Maybe there are better ways like deciding based on the name pattern or the file path/directory. Honestly I think the latter one is more preferable. The user should better organize the files based on whether they want to cache or not, instead of inferring from the file size.

Back to the original discussion, I think if there's a real use case, we can further discuss if deciding on the file size is the best way. And if there currently no use case and we are designing as a general use, I think it's actually harder for us to know if this is the best option.

yuzhu:
what about making it an extension of the `pin` command?  pin means to always cache, pin -n never cache?

yabola:
Thanks for all replies. Yeah, caching based on file size may not be a common usage.If caching based on file path pattern (or other caching strategy) makes sense, I can try to write it.

PR #13171, opened 2021-04-05
Change System.getCurrentTimeMillis to Instant.now().toEpochMilli()

jiacheliu3:
This changes `System.getCurrentTimeMillis()` to use Java 8 new time implementation for better precision thread safety. Many classes are using `CommonUtils.getCurrentMs` to calculate time elapsed. So we should move to this new implementation for better calculation precision.

Ref: https://stackify.com/heres-how-to-calculate-elapsed-time-in-java/

### DIFF
--- a/core/common/src/main/java/alluxio/util/CommonUtils.java
+++ b/core/common/src/main/java/alluxio/util/CommonUtils.java
@@ -44,6 +44,7 @@ import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/util/CommonUtils.java
```diff
@@ -571,7 +572,7 @@ public void remove() {
    */
   public static <T> void invokeAll(ExecutorService service, List<Callable<T>> callables,
       long timeoutMs) throws TimeoutException, ExecutionException {
-    long endMs = System.currentTimeMillis() + timeoutMs;
+    long endMs = getCurrentMs() + timeoutMs;
```
jiacheliu3, line 575: `System.currentTimeMillis()` should all be changed to use the util method to hide implementation

gpang (commented):
I think this looks good. Is there a perf difference with the new API?

jiacheliu3:
> I think this looks good. Is there a perf difference with the new API?

@gpang I did a bunch of tests like running `System.currentTimeMillis()` vs `Instant.now().toEpochMillis()` 1 million times and compare the time taken. The difference in performance is very small. For 1 million executions the latter one is taking 1ms~6ms more than the former. So I would say the performance is just acceptable.

gpang (approved):
LGTM

jiacheliu3:

PR #13175, opened 2021-04-05
Add RocksDB.open retry logic

bradyoo:
When doing a quick restart of an alluxio master process with RocksDB, the previous RocksDB may not have been fully cleared yet if the previous process didn't terminate gracefully. Add a timeout around connection so the previous connection lock times out before.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
+++ b/core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
@@ -149,6 +149,7 @@ public final class RocksStore implements Closeable {
     while (retryPolicy.attempt()) {
       try {
         mDb = RocksDB.open(mDbOpts, mDbPath, cfDescriptors, columns);
+        break;
       } catch (RocksDBException e) {
         // sometimes the previous terminated process's lock may not have been fully cleared yet
         // retry until timeout to make sure that isn't the case
--- a/core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
+++ b/core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
@@ -15,6 +15,8 @@ import alluxio.Constants;
 import alluxio.master.journal.checkpoint.CheckpointInputStream;
 import alluxio.master.journal.checkpoint.CheckpointOutputStream;
 import alluxio.master.journal.checkpoint.CheckpointType;
+import alluxio.retry.TimeoutRetry;
+import alluxio.util.CommonUtils;
 import alluxio.util.TarUtils;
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
```diff
@@ -49,6 +50,7 @@
 @ThreadSafe
 public final class RocksStore implements Closeable {
   private static final Logger LOG = LoggerFactory.getLogger(RocksStore.class);
+  public static final int ROCKS_OPEN_RETRY_TIMEOUT = 20 * 1000;
```
calvinjia, line 53: Use `Constants`?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
```diff
@@ -142,7 +144,20 @@ private void createDb() throws RocksDBException {
...
+    while (retryPolicy.attempt()) {
+      try {
+        mDb = RocksDB.open(mDbOpts, mDbPath, cfDescriptors, columns);
+      } catch (RocksDBException e) {
+        // sometimes the previous terminated process's lock may not have been fully cleared yet
+        // retry until timeout to make sure that isn't the case
+        lastException = e;
+      }
+    }
+    if (mDb == null && lastException != null) {
```
calvinjia, line 158: Is it possible for `mDb == null` and `lastException != null`?

calvinjia (changes requested):
It seems like there can be a constant overhead of closing the last RocksDB instance, is there a way to avoid this (ie. use new dir)?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksStore.java
bradyoo: Yeah when RocksDB.open always throws a RocksDBException. lastException will not be null and mDB will never be populated.

bradyoo:
feedback applied

calvinjia (approved):
LGTM, any thoughts on an alternative to waiting on the previous clean up, or is the clean up generally time bound and only an issue when we are running tests?

bradyoo:
the previous cleanup is odd. Mostly because it's coming from a process that is already dead.

calvinjia:
@bradyoo do you mean its likely not waiting for the clean up to complete (since there is no one to clean it up), but the lockfile is hanging around for some other reason? Ideally we can remove the potential 20 extra seconds for start up.

bradyoo:
There was a very obvious bug here that has now been fixed.

bradyoo:

PR #13181, opened 2021-04-06
Fix helm-chart fuse hostpath type from File to CharDevice

cdmikechen:
In k8s 1.19+ , `hostPathType` needs to be more accurate, otherwise the following error will be reported:
```
MountVolume.SetUp failed for volume "alluxio-fuse-device" : hostPath type check failed: /dev/fuse is not a file
```
In Node use `file` command to check `/dev/fuse` type:
```shell
$ file fuse
fuse: character special (10/229)
```

So that Maybe we should change `hostPathType` from  `File` to `CharDevice`

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
+++ b/integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
@@ -155,3 +155,7 @@
 0.6.15
 
 - Fix incorrect indentation in logserver secret volume mount
+
+0.6.16
+
+- Change helm-chart fuse hostPath type from File to CharDevice
--- a/integration/kubernetes/helm-chart/alluxio/Chart.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/Chart.yaml
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.6.15
+version: 0.6.16
 home: https://www.alluxio.io/
 maintainers:
 - name: Adit Madan
--- a/integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/fuse/daemonset.yaml
@@ -130,7 +130,7 @@ spec:
         - name: alluxio-fuse-device
           hostPath:
             path: /dev/fuse
-            type: File
+            type: CharDevice
         - name: alluxio-fuse-mount
           hostPath:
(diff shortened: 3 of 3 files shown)


jiacheliu3 (commented):
@cdmikechen Thanks for the fix! Could you update the helm chart version just like https://github.com/Alluxio/alluxio/pull/13193#issuecomment-815281941?
This fix looks good to me but I have 2 questions:
1. I don't know much about `CharDevice` vs `BlockDevice` in this field. Does Fuse work if you set this to `BlockDevice`? Will it work or will anything be different?
2. Will setting this to `CharDevice` work with K8s <1.19?

cdmikechen:
@jiacheliu3 
Hi~ Thanks for your reviewed !
I've changed `CharDevice` to `BlockDevice` and fuse can not started:
```
MountVolume.SetUp failed for volume "alluxio-fuse-device" : hostPath type check failed: /dev/fuse is not a block device
```

`CharDevice` is not a new type in k8s 1.19+, so that I think it can works in k8s below 1.19.

integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
```diff
@@ -155,3 +155,7 @@
 0.6.15
 
 - Fix incorrect indentation in logserver secret volume mount
+
+0.6.16
+
+- Fix helm-chart fuse hostPath type from File to CharDevice
```
jiacheliu3, line 161:
```suggestion
- Change helm-chart fuse hostPath type from File to CharDevice
```

integration/kubernetes/helm-chart/alluxio/Chart.yaml
```diff
@@ -21,3 +21,6 @@ maintainers:
   email: jiacheng@alluxio.com
 - name: Yang Che
   email: cheyang@163.com
+- name: Xiang chen
+  email: cdmikechen@hotmail.com
```
jiacheliu3, line 25: This list is the major code maintainers. We have the commitment on maintaining the helm chart and resolving issues. If you volunteer to be one, you are more than welcome! But you don't have to put your name here just because you submitted a bug fix :)

jiacheliu3 (approved):
@cdmikechen Thanks! LGTM except one minor wording suggestion.

integration/kubernetes/helm-chart/alluxio/Chart.yaml
cdmikechen: @jiacheliu3  OK~ I‘ll remove this

jiacheliu3:

PR #13183, opened 2021-04-06
Fix helm-chart logserver secretVolumeMounts space Indent

cdmikechen:
`alluxio.logserver.secretVolumeMounts` in `_helpers.tpl` has wrong space Indent. If use hadoop config, it will report this error:
```
Error: YAML parse error on alluxio/templates/logserver/deployment.yaml: error converting YAML to JSON: yaml: line 69: did not find expected key
```

Now, try to fix this.

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/templates/_helpers.tpl
+++ b/integration/kubernetes/helm-chart/alluxio/templates/_helpers.tpl
@@ -181,9 +181,9 @@ resources:
 
 {{- define "alluxio.logserver.secretVolumeMounts" -}}
   {{- range $key, $val := .Values.secrets.logserver }}
-            - name: secret-{{ $key }}-volume
-              mountPath: /secrets/{{ $val }}
-              readOnly: true
+          - name: secret-{{ $key }}-volume
+            mountPath: /secrets/{{ $val }}
+            readOnly: true
   {{- end -}}
 {{- end -}}
 


jiacheliu3 (approved):
LGTM, I reproduced this issue and the solution works. Thanks for the fix!

jiacheliu3:

jiacheliu3:
@cdmikechen BTW, the log server is just writing logs to a local path. It does not really connect to the HDFS or anything. So actually it does not need the HDFS secret. Yea maybe in the `values.yaml` I could give a better example.

cdmikechen:
@jiacheliu3  Thanks for your reviewed~
I just found this problem when testing the latest version by HDFS, and I found some other problems. I've seen every time updating helm, I need to update the version information of chart. Should I merge multiple issues into a PR to submit?

jiacheliu3:
> @jiacheliu3 Thanks for your reviewed~
> I just found this problem when testing the latest version by HDFS, and I found some other problems. I've seen every time updating helm, I need to update the version information of chart. Should I merge multiple issues into a PR to submit?

Thanks for brining that up. I would suggest creating one PR for each issue, so we keep track of it individually. Please also update the Alluxio Helm chart version together with the code change like https://github.com/Alluxio/alluxio/pull/13193. Thanks!

PR #13188, opened 2021-04-07
[WIP] Add scalability test

jiacheliu3:
This branch is used for scalability related analysis and investigations. This PR is currently only for CI/CD.

When we are done with the scalability analysis, the testing code here may be organized into an instrumentation framework. By then we will decide whether or not to use the same PR or create different PRs to merge the relevant code into Alluxio.

The new benchmark tests are (examples below)
```
$ bin/alluxio runClass alluxio.stress.cli.RpcBench --duration 30s --rpc registerWorker --block-count 50000 --fake-same-worker --concurrency 20

$ bin/alluxio runClass alluxio.stress.cli.RpcBench --duration 30s --rpc blockHeartbeat--block-count 50000 --concurrency 20
```

See RpcBench for code.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
@@ -86,6 +86,8 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.UnknownHostException;
 import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -821,7 +823,14 @@ public final class DefaultBlockMaster extends CoreMaster implements BlockMaster 
         continue;
       }
 
+      Instant waitWorkerBlock = Instant.now();
       synchronized (worker) {
+        Instant inWorkerBlock = Instant.now();
+        Duration waitTime = Duration.between(waitWorkerBlock, inWorkerBlock);
+        if (waitTime.compareTo(Duration.ofMillis(1)) > 0) {
+          LOG.warn("Wait time {}ms to enter the registerWorkerInternal synchronized section", waitTime.toMillis());
+        }
+
         worker.updateLastUpdatedTimeMs();
         mWorkers.add(worker);
         workers.remove(worker);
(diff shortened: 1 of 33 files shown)


jiacheliu3:
@HelloHorizon Let's use this branch for the tests.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -821,7 +823,14 @@ private MasterWorkerInfo registerWorkerInternal(long workerId) {
         continue;
       }
 
+      Instant waitWorkerBlock = Instant.now();
       synchronized (worker) {
+        Instant inWorkerBlock = Instant.now();
```
jiacheliu3, line 828: This synchronized section should have very little contention.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -888,7 +897,16 @@ public void workerRegister(long workerId, List<String> storageTiers,
       blocks.addAll(blockIds);
     }
 
+    Instant waitWorkerBlock = Instant.now();
+    // This synchronized section is guarding two workers registering with the same NetAddress
+    // This should happen with a very low possibility
+    // And should not be the performance bottleneck
     synchronized (worker) {
```
jiacheliu3, line 904: This synchronized section should have very little contention.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1140,6 +1158,7 @@ private void processLostWorker(MasterWorkerInfo worker) {
   }
 
   private LockResource lockBlock(long blockId) {
+    // TODO(jiacheng): when will this be bottleneck?
```
jiacheliu3, line 1161: There are 10K locks provided by `Striped.lock`. When will there be contention?

stress/shell/src/main/java/alluxio/stress/cli/RpcBench.java
```diff
@@ -0,0 +1,273 @@
...
+    // Stop after certain time has elapsed
+    int startPort = 9999;
+    long i = 0;
+    while (Instant.now().isBefore(endTime)) {
+      Instant s = Instant.now();
+
+      try {
+        String hostname = NetworkAddressUtils.getLocalHostName(500);
+        LOG.info("Detected local hostname {}", hostname);
+        WorkerNetAddress address = new WorkerNetAddress().setHost(hostname).setDataPort(startPort++).setRpcPort(startPort++);
```
jiacheliu3, line 143: A new address every time so it's a new worker

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Unless we trigger contention over the same worker object. The possible situations include:
1. register the same worker in a high volume.
This test makes no sense in reality. However it's probably the easiest way if we want to poke the bottleneck on this critical section. 

2. In many other calls they also do `synchronized(worker)`, for example `commitBlock`
It can be harder to simulate those calls. If what matters is only this synchronized lock, then we do Point 1.

jiacheliu3:
The changes in this draft have been split into separate PRs and merged into 2.6. Closing this.

PR #13201, opened 2021-04-08
Add metrics to JNIFuse and StackFS

LuQQiu:
Add some Fuse metrics for easy debugging and monitoring

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -22,6 +22,8 @@ import alluxio.conf.PropertyKey;
 import alluxio.exception.PreconditionMessage;
 import alluxio.exception.status.NotFoundException;
 import alluxio.grpc.ReadRequest;
+import alluxio.metrics.MetricKey;
+import alluxio.metrics.MetricsSystem;
 import alluxio.network.protocol.databuffer.DataBuffer;
 import alluxio.proto.dataserver.Protocol;
 import alluxio.util.io.BufferUtils;
@@ -30,6 +32,7 @@ import alluxio.util.network.NetworkAddressUtils;
 import alluxio.wire.BlockInfo;
 import alluxio.wire.WorkerNetAddress;
 
+import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.slf4j.Logger;
@@ -301,7 +304,9 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
       return -1;
     }
     int toRead = Math.min(len, mCurrentChunk.readableBytes());
-    mCurrentChunk.readBytes(b, off, toRead);
+    try (Timer.Context ctx = MetricsSystem.timer(MetricKey.CLIENT_BLOCK_READ_FROM_CHUNK.getName()).time()) {
+      mCurrentChunk.readBytes(b, off, toRead);
+    }
     mPos += toRead;
     return toRead;
   }
@@ -456,7 +461,9 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
       mCurrentChunk = null;
     }
     if (mCurrentChunk == null) {
-      mCurrentChunk = mDataReader.readChunk();
+      try (Timer.Context ctx = MetricsSystem.timer(MetricKey.CLIENT_BLOCK_READ_CHUNK.getName()).time()) {
+        mCurrentChunk = mDataReader.readChunk();
+      }
     }
   }
 
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -61,6 +61,7 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
   }
 
   private final WorkerNetAddress mAddress;
+  private final boolean mDetailedMetricsEnabled;
   private final BlockInStreamSource mInStreamSource;
   /** The id of the block or UFS file to which this instream provides access. */
   private final long mId;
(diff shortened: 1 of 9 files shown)


integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
```diff
@@ -274,13 +271,15 @@ public static int call(Logger logger, FuseCallable callable, String methodName,
       String description, Object... args) {
     String debugDesc = logger.isDebugEnabled() ? String.format(description, args) : null;
     logger.debug("Enter: {}({})", methodName, debugDesc);
-    long startMs = System.currentTimeMillis();
-    int ret = callable.call();
-    long durationMs = System.currentTimeMillis() - startMs;
-    logger.debug("Exit ({}): {}({}) in {} ms", ret, methodName, debugDesc, durationMs);
-    if (durationMs >= THRESHOLD) {
-      logger.warn("{}({}) returned {} in {} ms (>={} ms)", methodName,
-          String.format(description, args), ret, durationMs, THRESHOLD);
```
apc999, line 283: can we still keep the threshold and this warning?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4608,13 +4608,6 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.IGNORE)
           .setScope(Scope.CLIENT)
           .build();
-  public static final PropertyKey FUSE_LOGGING_THRESHOLD =
```
apc999, line 4611: prefer to keep it. it turned out useful

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
```diff
@@ -274,13 +271,15 @@ public static int call(Logger logger, FuseCallable callable, String methodName,
...
+    int ret = -1;
+    try (Timer.Context ctx = MetricsSystem.timer("Fuse" + methodName).time()) {
```
apc999, line 275: string calculation can be expensive, let's avoid having them on critical path. e.g., passing string "Fuse.Read" etc in to `call`?

integration/fuse/src/main/java/alluxio/fuse/StackFS.java
```diff
@@ -86,7 +87,11 @@ private int getMode(Path path) throws IOException {
 
   @Override
   public int getattr(String path, FileStat stat) {
-    MetricsSystem.counter("Client.GetattroOps").inc();
+    return AlluxioFuseUtils.call(
+        LOG, () -> getattrInternal(path, stat), STACK_FS_PREFIX + "Getattr", "path=%s", path);
```
apc999, line 91: same here, this can be expensive

apc999 (changes requested):

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: I agree that it's useful. The reason that I remove it is because
(1) duplicate time recording one by metrics.timer and one by manually getting the start/end time.
(2) The functionalities can also be served by Timer metrics P75, P90, P95, P99. If the P90 number is larger than 10S, then 10% of this operation is taking longer 10s. Although I agree that the logging may be easier to be understand by end users.

Whats your thoughts @apc999？

LuQQiu: Ping @apc999

integration/fuse/src/main/java/alluxio/fuse/StackFS.java
LuQQiu: DONE

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
LuQQiu: DONE

core/common/src/main/java/alluxio/conf/PropertyKey.java
apc999: it is often more useful when we know the time sequence in the log which helps debugging in my past experience.
e.g., knowing the location of expensive FUSE calls can help checkout the abnormals in other components.

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -301,7 +304,10 @@ public int read(byte[] b, int off, int len) throws IOException {
       return -1;
     }
     int toRead = Math.min(len, mCurrentChunk.readableBytes());
-    mCurrentChunk.readBytes(b, off, toRead);
+    try (Timer.Context ctx = MetricsSystem
```
apc999, line 307: @LuQQiu , I am a bit concerned about the overhead introduced here.
Can you add a flag to turn on and off whether we are tracking these operations by timers?

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -456,7 +462,10 @@ private void readChunk() throws IOException {
       mCurrentChunk = null;
     }
     if (mCurrentChunk == null) {
-      mCurrentChunk = mDataReader.readChunk();
+      try (Timer.Context ctx = MetricsSystem
```
apc999, line 465: same here, let's use the same flag to turn it on and off ?

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
apc999: still `"Fuse" + methodName`?

also, given all these methods are known, maybe create timers in the caller classes like `mFuseGetAttrTimer` and pass in the timer directly in this util call?
e.g.:

```java
public static int call(Logger logger, FuseCallable callable, String methodName, Timer timer,
      String description, Object... args) {
  ...
  Timer.Context ctx = timer.time();
  int ret = callable.call();
  long durationNs = ctx.stop();
  // then we can still make the timer but also track the duration and keep the original logic
  if (durationMs >= THRESHOLD_IN_NS) {
     logger.warn(...);
  } 
}

```

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: Make sense, will add the log back

apc999 (changes requested):

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
LuQQiu: Oops, changed to another wrong metrics branch. Updated the PR.
Timer instance may be changed during reset all metrics functionality. If we save the timer instance, the future metrics after resetting metrics may be lost.

LuQQiu:
@apc999 Sorry for updating the wrong branch previously. Updated this PR with the latest changes, PTAL, thanks

apc999 (approved):

LuQQiu:

PR #13204, opened 2021-04-09
Support FUSE on Worker in docker file

apc999:
Add libfuse to Alluxio server dockerfile

### DIFF
--- a/integration/docker/Dockerfile
+++ b/integration/docker/Dockerfile
@@ -79,14 +79,16 @@ RUN if [ ${ALLUXIO_USERNAME} != "root" ] \
     && [ ${ALLUXIO_GROUP} != "root" ] \
     && [ ${ALLUXIO_UID} -ne 0 ] \
     && [ ${ALLUXIO_GID} -ne 0 ]; then \
-      addgroup -g ${ALLUXIO_GID} -S ${ALLUXIO_GROUP} && \
-      adduser -u ${ALLUXIO_UID} -S ${ALLUXIO_USERNAME} -G ${ALLUXIO_GROUP} && \
+      addgroup --gid ${ALLUXIO_GID} ${ALLUXIO_GROUP} && \
+      adduser --system --uid ${ALLUXIO_UID} --gid ${ALLUXIO_GID} ${ALLUXIO_USERNAME} && \
       usermod -a -G root ${ALLUXIO_USERNAME} && \
       mkdir -p /journal && \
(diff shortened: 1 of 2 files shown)


integration/docker/Dockerfile
```diff
@@ -32,7 +32,10 @@ RUN if [ ${ENABLE_DYNAMIC_USER} = "true" ] ; then \
        chmod -R 777 /opt/* ; \
     fi
 
-FROM openjdk:8-jdk-alpine
+# instead of ubuntu:16.04, use alluxio/alluxio-ubuntu:ubuntu1604-customize
+# applies optimizations in libfuse on top of ubuntu:16.04
+# it is based off cheyang/fuse2:ubuntu1604-customize
+FROM alluxio/alluxio-ubuntu:1604-customize
```
cheyang, line 38: I think there are some differences between debian and alpine. For example, it uses 'apt' in debian while 'apk' in alpine

cheyang (changes requested):

cheyang:
And you also need to install fuse lib in worker dockerfile

integration/docker/Dockerfile
apc999: commented out

apc999:
> And you also need to install fuse lib in worker dockerfile

In `/libfuse`, as part of image
`alluxio/alluxio-ubuntu:1604-customize` which is based off `cheyang/fuse2:ubuntu1604-customize`

integration/docker/Dockerfile
```diff
@@ -73,6 +85,8 @@ RUN if [ ${ALLUXIO_USERNAME} != "root" ] \
       mkdir -p /journal && \
```
cheyang, line 85: The command to adduser is also different between debian and alpine.

ZhuTopher: @apc999 Maybe this is the problem you're running into?

ZhuTopher (commented):
The changes LGTM from what I understand about this image

cheyang:
May I ask how to set mount opts in alluxio worker?

ZhuTopher (approved):
LGTM from what I can tell

apc999:
> May I ask how to set mount opts in alluxio worker?

by `alluxio.worker.fuse.mount.options`. I will create a docfix PR to tell how to set these in docker

apc999:
Tested this PR in docker and macos . it works now. thanks for your help and reivew @cheyang @ZhuTopher

apc999:

PR #13208, opened 2021-04-09
Update docs with latest docGen run

ns1123:
Docs were out of sync. This PR is just to bring things back in line so that future commits can include their doc changes.


docs/_data/table/en/common-configuration.yml
```diff
@@ -344,3 +356,113 @@ fs.swift.tenant:
...
+user.dir:
+  'N/A'
+user.home:
+  'N/A'
+user.language:
+  'N/A'
+user.name:
+  'N/A'
+user.timezone:
+  'N/A'
```
calvinjia, line 468: Do you know what added all these?

apc999 (approved):

docs/_data/table/en/common-configuration.yml
apc999: we need to investigate this. it is also generated on my side for unknown reasons. 
I removed them manually for now

apc999:

PR #13210, opened 2021-04-09
Cancel sync job in middle of ufs calls if client cancelled

bradyoo:


core/server/master/src/main/java/alluxio/underfs/UfsStatusCache.java
```diff
@@ -210,22 +213,28 @@ public UfsStatus fetchStatusIfAbsent(AlluxioURI path, MountTable mountTable)
...
-            path, e);
-        if (e instanceof InterruptedException) {
-          Thread.currentThread().interrupt();
-          throw (InterruptedException) e;
+      while (true) {
+        try {
+          return prefetchJob.get(100, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException e) {
+          if (rpcContext != null) {
+            rpcContext.throwIfCancelled();
```
yuzhu, line 226: is there a way we can test the cancellation and how our system responds to a runtimeException being thrown in the middle of a sync?

yuzhu (commented):
@bradyoo Thanks for fixing this! left some comments

bradyoo:
@yuzhu Hmm.. I was in the middle of turning the LinkedBlockingQueue into some PriorityQueue for the Executor but I'm beginning to realize that that's not a good idea because while it moves everything I want to the front of the queue, it causes everything else to not be in submission order unless I give unique priorities to everything so I'm scrapping the implementation. I'm thinking of just having yet another Executor to handle this. What do you think?

yuzhu:
@bradyoo i think a separate executor makes sense here.

bradyoo:
Comments discussed and tests added

yuzhu (approved):

bradyoo:

bradyoo:

bradyoo:

bradyoo:
this is having an adverse effect in sync performance. Reverting.

Issue #13203, opened 2021-04-09
Helm chart add support for toleration


jiacheliu3:
**Is your feature request related to a problem? Please describe.**
Currently there are only fields for `nodeSelector`. We should have the same fields for `tolerations`

**Describe the solution you'd like**
Add the fields similar to `nodeSelector`

**Describe alternatives you've considered**

**Urgency**
Explain why the feature is important

**Additional context**

ssz1997:
This is solved by https://github.com/Alluxio/alluxio/pull/13214

PR #13214, opened 2021-04-12
Add Helm chart support for k8s node tolerations

ZhuTopher:
This PR is being made to address git [issue #13203](https://github.com/Alluxio/alluxio/issues/13203).

I tested that the Helm chart rendered using the following in my `config.yaml`
```
tolerations: [ {"key": "test", "operator": "Equal", "value": "test", "effect": "NoSchedule"} ]
```

and that the tolerations worked as intended on k8s nodes I tainted using `kubectl taint nodes <node> test=test:NoSchedule`

**Note:** I tested the Helm chart using all default values; so no FUSE nor logging-server deployments.

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
@@ -41,6 +41,18 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-logserver
     spec:
+      nodeSelector:
+      {{- if .Values.logserver.nodeSelector }}
+{{ toYaml .Values.logserver.nodeSelector | trim | indent 8  }}
+      {{- else if .Values.nodeSelector }}
+{{ toYaml .Values.nodeSelector | trim | indent 8  }}
+      {{- end }}
+      tolerations:
+      {{- if .Values.logserver.tolerations }}
+{{ toYaml .Values.logserver.tolerations | trim | indent 8  }}
+      {{- else if .Values.tolerations }}
+{{ toYaml .Values.tolerations | trim | indent 8  }}
+      {{- end }}
       containers:
         - name: alluxio-logserver
           image: {{ .Values.image }}:{{ .Values.imageTag }}
--- a/integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
@@ -50,7 +50,8 @@ spec:
       tolerations:
       {{- if .Values.logserver.tolerations }}
 {{ toYaml .Values.logserver.tolerations | trim | indent 8  }}
-      {{- else if .Values.tolerations }}
+      {{- end }}
+      {{- if .Values.tolerations }}
 {{ toYaml .Values.tolerations | trim | indent 8  }}
       {{- end }}
       containers:
--- a/integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
@@ -41,12 +41,6 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-logserver
     spec:
-      nodeSelector:
-      {{- if .Values.logserver.nodeSelector }}
-{{ toYaml .Values.logserver.nodeSelector | trim | indent 8  }}
-      {{- else if .Values.nodeSelector }}
-{{ toYaml .Values.nodeSelector | trim | indent 8  }}
-      {{- end }}
       tolerations:
       {{- if .Values.logserver.tolerations }}
 {{ toYaml .Values.logserver.tolerations | trim | indent 8  }}
--- a/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
@@ -73,7 +73,8 @@ spec:
       tolerations:
       {{- if .Values.master.tolerations }}
 {{ toYaml .Values.master.tolerations | trim | indent 8  }}
-      {{- else if .Values.tolerations }}
+      {{- end }}
+      {{- if .Values.tolerations }}
 {{ toYaml .Values.tolerations | trim | indent 8  }}
       {{- end }}
       securityContext:
--- a/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
@@ -70,6 +70,12 @@ spec:
       {{- else if .Values.nodeSelector }}
 {{ toYaml .Values.nodeSelector | trim | indent 8  }}
       {{- end }}
+      tolerations:
+      {{- if .Values.master.tolerations }}
+{{ toYaml .Values.master.tolerations | trim | indent 8  }}
+      {{- else if .Values.tolerations }}
+{{ toYaml .Values.tolerations | trim | indent 8  }}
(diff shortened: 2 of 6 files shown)


jiacheliu3:
@ZhuTopher Thanks for adding this! You will need to rebase when https://github.com/Alluxio/alluxio/pull/13226 is in. It took 0.6.17 so you'll need to be 0.6.18.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -46,6 +46,11 @@ properties:
 # Use labels to run Alluxio on a subset of the K8s nodes
 # nodeSelector: {}
 
+# A list of k8s Node taints to allow scheduling on.
+# See the Kubernetes docs for more info:
+# - https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
+# tolerations: []
```
jiacheliu3, line 52: Does it make sense to add an example in the comment?

jiacheliu3:
@nirav-chotai PTAL too, thanks!

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -69,6 +69,12 @@ spec:
 {{ toYaml .Values.master.nodeSelector | trim | indent 8  }}
       {{- else if .Values.nodeSelector }}
 {{ toYaml .Values.nodeSelector | trim | indent 8  }}
+      {{- end }}
+      tolerations:
+      {{- if .Values.master.tolerations }}
+{{ toYaml .Values.master.tolerations | trim | indent 8  }}
+      {{- else if .Values.tolerations }}
```
jiacheliu3, line 76: How do you like `.Values.master.tolerations` + `.Values.tolerations`? In other words I think this should be a union instead of if-else.

jiacheliu3 (commented):
Thanks @ZhuTopher ! I left some comments

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
nirav-chotai: It would be good if `.Values.tolerations` is there which can add tolerations to all the Alluxio components rather than specifying them individually. And if someone wants to run only some components on dedicated nodes, then they can use `.Values.XYZ.tolerations`

integration/kubernetes/helm-chart/alluxio/values.yaml
nirav-chotai: yes, as many make a mistake between `list` and `maps` while passing values here.

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
ZhuTopher: Yeah that makes sense to me, I can change it to a union of the tolerations.

integration/kubernetes/helm-chart/alluxio/values.yaml
ZhuTopher: Ah okay, I'll specify an example

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -64,6 +64,13 @@ spec:
 {{ toYaml .Values.worker.nodeSelector | trim | indent 8  }}
       {{- else if .Values.nodeSelector }}
 {{ toYaml .Values.nodeSelector | trim | indent 8  }}
+      {{- end }}
+      tolerations:
+      {{- if .Values.worker.tolerations }}
+{{ toYaml .Values.worker.tolerations | trim | indent 8  }}
+      {{- end }}
+      {{- if .Values.tolerations }}
+{{ toYaml .Values.tolerations | trim | indent 8  }}
```
ZhuTopher, line 73: @jiacheliu3 Is this what you meant by union? This will include duplicate definitions if the same toleration is in both `.Values.tolerations` and `.Values.xxx.tolerations` but I think that's on the user if they do that. Also I tested duplicate toleration definitions and I don't think it matters to Kubernetes anyway

jiacheliu3: Exactly this is what I meant

jiacheliu3 (approved):
Thanks! LGTM except for the comment below. Please remove the change I mentioned in the comment then I'll merge it.

integration/kubernetes/helm-chart/alluxio/templates/logserver/deployment.yaml
```diff
@@ -41,6 +41,19 @@ spec:
         heritage: {{ .Release.Service }}
         role: alluxio-logserver
     spec:
+      nodeSelector:
+      {{- if .Values.logserver.nodeSelector }}
+{{ toYaml .Values.logserver.nodeSelector | trim | indent 8  }}
+      {{- else if .Values.nodeSelector }}
+{{ toYaml .Values.nodeSelector | trim | indent 8  }}
+      {{- end }}
```
jiacheliu3, line 49: I can see you choose `if`+`else if` because the other code are doing this way. However I think `if`+`if` is a better option because it allows you to have general node selectors and more specific ones. 

I think this PR (and helm chart version) is not intended for this nodeSelector change so let's remove this few lines. In a separate PR you can address this one and if you get a chance, change all node selectors to be `if` + `if`.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -416,7 +416,6 @@ logserver:
   # dnsPolicy: ClusterFirst
   # JVM options specific to the logserver container
   jvmOptions:
-  nodeSelector: {}
```
ZhuTopher, line 409: @jiacheliu3 I removed this because it was (and will still be) unused

jiacheliu3: okay, let's deal with the `nodeSelector` in a separate PR after discussion

jiacheliu3: Actually, let's keep it this way (don't remove this line).

One use case I can imagine is, there are nodes that have storage and nodes that do not. By defining a selector, the logserver is scheduled to a node that has storage, so that the logs are persisted to the storage.

Another argument is, let's do not take away what's there, unless the whole PR is intended for that. In this PR we add toleration, so let's not touch unrelated stuff. In the next PR just for nodeSelector, we can do the refactor as we redesign.

jiacheliu3: @ZhuTopher I'll merge after this is reverted. Thanks!

jiacheliu3: @ZhuTopher ooops sorry I seem to have clicked resolve unintendedly. Could you revert this line and then I'll merge? Thanks!

ZhuTopher: Oh sorry, I didn't notice this comment previously my bad! The reason I removed this from the Helm chart `values.yaml` is because this field is unused in `templates/logserver/deployment.yaml`.

I tried to add the `nodeSelector` behaviour from our other templates to it but since we decided to leave it out of this PR I also removed the related chart value to avoid confusion (chart implies you _can_ specify a `nodeSelector` when you actually can't).

Anyway, again we can leave all this to a separate 'nodeSelector'-PR so I'll revert it for now.

jiacheliu3:
the failed test is due to flaky codecov, irrelevant to the change

jiacheliu3:

jiacheliu3:
Same as the other PRs please add to the doc page too. Let's do the doc change in a separate PR in order not to block yourself. Thanks!

PR #13228, opened 2021-04-14
Remove extra HashSet creation for pinned list

jiacheliu3:
In `InodeTreePersistentState` we return a `Collections.unmodifiableSet` wrapped around the pinned list. Then in `InodeTree` we create a HashSet copy from it yet again. This is redundant.

Eventually the `getPinnedFileIds` call should be changed to a list-watch mechanism where the worker lists all the pinned IDs then just incrementally watch the changes. But that is not trivial and is not done in this PR.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
@@ -1094,7 +1094,7 @@ public class InodeTree implements DelegatingJournaled {
    * @return the set of file ids which are pinned
    */
   public Set<Long> getPinIdSet() {
-    return new HashSet<>(mState.getPinnedInodeFileIds());
+    return mState.getPinnedInodeFileIds();
   }
 
   /**


core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -1094,7 +1094,7 @@ public void setReplication(RpcContext rpcContext, LockedInodePath inodePath,
    * @return the set of file ids which are pinned
    */
   public Set<Long> getPinIdSet() {
-    return new HashSet<>(mState.getPinnedInodeFileIds());
+    return mState.getPinnedInodeFileIds();
```
jiacheliu3, line 1097: This `mState.getPinnedInodeFileIds()` creates a `Collections.unmodifieableSet()` over the set of pinned file IDs. 
https://github.com/Alluxio/alluxio/blob/4c4ec92c9a117dc5cebd668a0d3e5803c645e87e/core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java#L133

I think we don't need to create a new HashSet over the UnmodifiableSet here. The only upside of having a `HashSet` is being modifiable. The downside is a copy of a whole set. Currently all the references of this method are read-only. This copy is unnecessary.

ggezer (approved):

apc999 (approved):

apc999:

PR #13231, opened 2021-04-14
Change throw FuseException to log error in jni-fuse umount

LuQQiu:
If the Fuse device is busy reading/writing files, we want umount to wait for those operations to happen.
If timeout exceeds, we want umount to be an no-op.
The current implementation is incorrect and doesn't achieve our target.
Add a TODO and change to log error for a workaround.


apc999 (approved):

apc999:

PR #13232, opened 2021-04-14
[SMALLFIX] Add tini to Docker image

ZhuTopher:
This change is being made to fix PR #13204 where we switched the Alluxio Docker base image from Alpine to Ubuntu.

Alpine has [tini](https://github.com/krallin/tini) included as a [package](https://pkgs.alpinelinux.org/package/edge/community/x86/tini) that we previously installed via `apk`.

Since our Helm charts rely on `tini` for the `command: ["tini", "--", "/entrypoint.sh"]`  (see [here](https://github.com/Alluxio/alluxio/blob/master/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml#L91) for an example), the quickest fix for now is to just add `tini` to the Docker image again.

### DIFF
--- a/integration/docker/Dockerfile
+++ b/integration/docker/Dockerfile
@@ -42,15 +42,15 @@ ARG ALLUXIO_UID=1000
 ARG ALLUXIO_GID=1000
 ARG ENABLE_DYNAMIC_USER=false
 
-# Add Tini since it isn't included in Ubuntu by default
-# - https://github.com/krallin/tini
-ENV TINI_VERSION v0.19.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /usr/local/bin/tini
-RUN chmod +x /usr/local/bin/tini
-
 # Specify versions for known vulnerabilities reported by trivy scan
 # RUN apk --no-cache --update add bash libc6-compat shadow tini 'libbz2>=1.0.6-r7' 'sqlite-libs>=3.28.0-r3' && \
 #    rm -rf /var/cache/apk/*
+
+# Add Tini
+# - https://github.com/krallin/tini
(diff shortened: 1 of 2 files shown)


Xenorith:
is this the same tini version that was available in alpine before the base image change?

ZhuTopher:
The `tini` version in our Alluxio image is 0.18.0:
```
/opt/alluxio-2.5.0 $ tini
tini (tini version 0.18.0)
Usage: tini [OPTIONS] PROGRAM -- [ARGS] | --version
Execute a program under the supervision of a valid init process (tini)
Command line options:
  --version: Show version and exit.
  -h: Show this help message and exit.
  -s: Register as a process subreaper (requires Linux >= 3.4).
  -p SIGNAL: Trigger SIGNAL when parent dies, e.g. "-p SIGKILL".
  -v: Generate more verbose output. Repeat up to 3 times.
  -w: Print a warning when processes are getting reaped.
  -g: Send signals to the child's process group.
  -e EXIT_CODE: Remap EXIT_CODE (from 0 to 255) to 0.
  -l: Show license and exit.
Environment variables:
  TINI_SUBREAPER: Register as a process subreaper (requires Linux >= 3.4).
  TINI_VERBOSITY: Set the verbosity level (default: 1).
  TINI_KILL_PROCESS_GROUP: Send signals to the child's process group.
```

It turns out the `openjdk:8-jdk-alpine` image doesn't have `tini` either. It was previously added in our Dockerfile through `apk`:
```
RUN apk --no-cache --update add ... tini ...
```

For now I'll lock the version to 0.18.0

apc999 (approved):

Xenorith (approved):

Xenorith:

PR #13233, opened 2021-04-14
Enable /mnt/alluxio-fuse as default worker fuse mount

LuQQiu:
Set the default value of worker fuse mount point to reduce user configuration in using worker fuse.


apc999 (approved):

apc999:

PR #13235, opened 2021-04-15
Prevent loading child inode metadata when not required

yuzhu:
Previously, loadMetadata on a single level of directory would load that directory and its children, causing unnecessary delay. 
This prevents that from happening by changing the descendant type when we are loading a non-base path.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -522,7 +522,8 @@ public class InodeSyncStream {
     // The requested path already exists in Alluxio.
     Inode inode = inodePath.getInode();
     // initialize sync children to true if it is a listStatus call on a directory
-    boolean syncChildren = inode.isDirectory() && !mIsGetFileInfo;
+    boolean syncChildren = inode.isDirectory() && !mIsGetFileInfo
+        && !mDescendantType.equals(DescendantType.NONE);
 
     // if the lock pattern is WRITE_EDGE, then we can sync (update or delete). Otherwise, if it is
     // we can only load metadata.
@@ -629,15 +630,11 @@ public class InodeSyncStream {
       }
     }
 
-    // Only sync children when
-    // (1) DescendantType.ALL or (2) syncing root of this stream && DescendantType.ONE
+    // sync plan does not know about the root of the sync
+    // if DescendantType.ONE we only sync children if we are syncing root of this stream
     if (mDescendantType == DescendantType.ONE) {
       syncChildren =
-          syncChildren && inode.isDirectory() && mRootScheme.getPath().equals(inodePath.getUri());
-    } else if (mDescendantType == DescendantType.ALL) {
-      syncChildren = syncChildren && inode.isDirectory();
-    } else {
-      syncChildren = false;
+          syncChildren && mRootScheme.getPath().equals(inodePath.getUri());
     }
 
     Map<String, Inode> inodeChildren = new HashMap<>();
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -491,8 +491,10 @@ public class InodeSyncStream {
       InterruptedException {
     if (!inodePath.fullPathExists()) {
       loadMetadataForPath(inodePath);
+      // skip the load metadata step in the sync if it has been just loaded
+      syncExistingInodeMetadata(inodePath, true);
     } else {
-      syncExistingInodeMetadata(inodePath);
+      syncExistingInodeMetadata(inodePath, false);
     }
   }
 
@@ -501,7 +503,7 @@ public class InodeSyncStream {
    *
    * This method expects the {@code inodePath} to already exist in the inode tree.
    */
-  private void syncExistingInodeMetadata(LockedInodePath inodePath)
+  private void syncExistingInodeMetadata(LockedInodePath inodePath, boolean skipLoad)
       throws AccessControlException, BlockInfoException, FileAlreadyCompletedException,
       FileDoesNotExistException, InvalidFileSizeException, InvalidPathException, IOException,
       InterruptedException {
@@ -634,6 +636,8 @@ public class InodeSyncStream {
           syncChildren && inode.isDirectory() && mRootScheme.getPath().equals(inodePath.getUri());
     } else if (mDescendantType == DescendantType.ALL) {
       syncChildren = syncChildren && inode.isDirectory();
+    } else {
+      syncChildren = false;
     }
 
     Map<String, Inode> inodeChildren = new HashMap<>();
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -491,8 +491,9 @@ private void syncInodeMetadata(LockedInodePath inodePath)
       InterruptedException {
     if (!inodePath.fullPathExists()) {
       loadMetadataForPath(inodePath);
+    } else {
+      syncExistingInodeMetadata(inodePath);
```
yuzhu, line 495: if a path does not exist in alluxio yet, load metadata should be equivalent to loading + syncing.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -691,11 +692,16 @@ private void loadMetadataForPath(LockedInodePath inodePath)
       throws InvalidPathException, AccessControlException, IOException, FileDoesNotExistException,
       FileAlreadyCompletedException, InvalidFileSizeException, BlockInfoException {
     UfsStatus status = mStatusCache.fetchStatusIfAbsent(inodePath.getUri(), mMountTable);
+    DescendantType descendantType = mDescendantType;
+    if (descendantType.equals(DescendantType.ONE)
```
yuzhu, line 696: we need to change descendantType to ONE if the original descendantType was not recursive

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -751,7 +757,8 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
             }
             LoadMetadataContext loadMetadataContext =
                 LoadMetadataContext.mergeFrom(LoadMetadataPOptions.newBuilder()
-                    .setLoadDescendantType(LoadDescendantPType.NONE)
+                    .setLoadDescendantType(type.equals(LoadDescendantPType.ONE)
```
yuzhu, line 760: we should set loadDescendantType to all if it was all to begin with.. setting it to NONE disables recursive loading of the paths.

yuzhu (commented):
@ZacBlanco I made several changes here to further improve the performance, PTAL. commented on the changes to explain them in a bit more detail.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ZacBlanco: If `loadDescendantType` is `ALL` won't this trigger recursive calls to this function on L769?
And if we trigger recursive calls, then won't be also be losing the parallelism of the inode metadata loading? It seems like if it's set to `ALL` this function now loads everything in a single-threaded manner?

ZacBlanco (commented):
I left a question about one of the changes

tests/src/test/java/alluxio/client/fs/LoadMetadataIntegrationTest.java
```diff
@@ -214,6 +214,16 @@ public void syncOverrideLoadMetadata() throws Exception {
     checkGetStatus("/mnt/dir1/dirA/file", options, true, true, 1);
   }
 
+  @Test
+  public void loadMetadataListDir() throws Exception {
```
gpang, line 218: What is this behavior before the change?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
gpang: Can you add this as a comment?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
gpang: Can you add an explanation as a comment, since all the syncing logic is difficult to follow sometimes.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
gpang: Please add a comment here, so it is easier to understand.

gpang (commented):
@yuzhu Thanks! I just left a few minor questions/comments.

tests/src/test/java/alluxio/client/fs/LoadMetadataIntegrationTest.java
yuzhu: when doing a ls dir triggering loadmetadata , it does ufs `ls dir` and ufs `ls dir/dira`, `dir/dirb`, which should not be necessary.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: done

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: LoadMetadataForPath does a maximum of one level and a minimum of 0 level now.

yuzhu:
helps with #13259

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -633,6 +636,8 @@ private void syncExistingInodeMetadata(LockedInodePath inodePath)
           syncChildren && inode.isDirectory() && mRootScheme.getPath().equals(inodePath.getUri());
     } else if (mDescendantType == DescendantType.ALL) {
       syncChildren = syncChildren && inode.isDirectory();
+    } else {
+      syncChildren = false;
```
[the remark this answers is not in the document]
yuzhu: looked at it again, it is not necessary to set syncChildren explicitly here. we only need to address the situation where RootScheme is not the path being considered.

yuzhu:
@gpang @ZacBlanco PTAL, addressed some comments.

gpang:
I think this works, but could you add more information to the PR description, as to what was the undesired behavior, and how this PR fixes it?

ZacBlanco (approved):
Thanks for the improvements :rocket:

ZacBlanco:
One request to change the title of the PR to be a little more descriptive. I propose:

> Prevent loading child inode metadata when not required

tests/src/test/java/alluxio/client/fs/LoadMetadataIntegrationTest.java
```diff
@@ -214,6 +214,16 @@ public void syncOverrideLoadMetadata() throws Exception {
...
+    ListStatusPOptions listOptions =
+        ListStatusPOptions.newBuilder().setRecursive(false)
+            .setLoadMetadataType(LoadMetadataPType.ONCE)
+            .setCommonOptions(FileSystemMasterCommonPOptions.newBuilder().setSyncIntervalMs(-1)
+            ).build();
+    checkListStatus("/mnt/", listOptions, true, true, 3);
```
gpang, line 224: Could you add a comment on what the "undesired" behavior would have done? I think it would have performed more UFS operations, right? How many more?

gpang (approved):
@yuzhu left a minor comment, but looks good.

LGTM

yuzhu:

yuzhu:

yuzhu:

PR #13236, opened 2021-04-15
Support open file for override

### LINKED ISSUE #13179: Support fuse open for overwrite
In our use case. We need to overwrite file due to job retry.

Currently, open an existing file for write will throw exception, maybe we can check the 
`(fi.flags.get() & OPEN_FLAG_MASK) == O_WRONLY.intValue()`
to support open for overwrite (remove existing file and create a new one).

maobaolong:
Fixes #13179
With this PR, the following bash command can execute successfully.
```bash
echo hello>mountPoint/file
echo world>mountPoint/file
```

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -343,6 +343,7 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
     try {
       long fd = mNextOpenFileId.getAndIncrement();
       if ((flags & 0b11) != 0) {
+        mFileSystem.delete(uri);
         FileOutStream os = mFileSystem.createFile(uri);
         long fid = mNextOpenFileId.getAndIncrement();
         mCreateFileEntries.add(new CreateFileEntry(fid, path, os));
@@ -689,8 +690,8 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
 
   @Override
   public int truncate(String path, long size) {
-    LOG.error("Truncate is not supported {}", path);
-    return -ErrorCodes.EOPNOTSUPP();
+    LOG.info("truncate {} to {}", path, size);
+    return 0;
   }
 
   @Override
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -338,20 +338,30 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
 
   private int openInternal(String path, FuseFileInfo fi) {
     final AlluxioURI uri = mPathResolverCache.getUnchecked(path);
+    final int flags = fi.flags.get();
+    LOG.trace("open({}, 0x{}) [target: {}]", path, Integer.toHexString(flags), uri);
     try {
       long fd = mNextOpenFileId.getAndIncrement();
-      FileInStream is;
-      try {
-        is = mFileSystem.openFile(uri);
-      } catch (FileIncompleteException e) {
-        if (AlluxioFuseUtils.waitForFileCompleted(mFileSystem, uri)) {
+      if ((flags & 0b11) != 0) {
+        FileOutStream os = mFileSystem.createFile(uri);
+        long fid = mNextOpenFileId.getAndIncrement();
+        mCreateFileEntries.add(new CreateFileEntry(fid, path, os));
+        fi.fh.set(fid);
+        setUserGroupIfNeeded(uri);
+      } else {
+        FileInStream is;
+        try {
           is = mFileSystem.openFile(uri);
-        } else {
-          throw e;
+        } catch (FileIncompleteException e) {
+          if (AlluxioFuseUtils.waitForFileCompleted(mFileSystem, uri)) {
+            is = mFileSystem.openFile(uri);
+          } else {
+            throw e;
+          }
         }
+        mOpenFileEntries.put(fd, is);
+        fi.fh.set(fd);
       }
-      mOpenFileEntries.put(fd, is);
-      fi.fh.set(fd);
       return 0;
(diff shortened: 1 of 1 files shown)


integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -338,20 +338,31 @@ public int open(String path, FuseFileInfo fi) {
...
+    final int flags = fi.flags.get();
+    LOG.trace("open({}, 0x{}) [target: {}]", path, Integer.toHexString(flags), uri);
     try {
       long fd = mNextOpenFileId.getAndIncrement();
-      FileInStream is;
-      try {
-        is = mFileSystem.openFile(uri);
-      } catch (FileIncompleteException e) {
-        if (AlluxioFuseUtils.waitForFileCompleted(mFileSystem, uri)) {
+      if ((flags & 0b11) != 0) {
```
LuQQiu, line 345: What's the meaning of this flag? so Fuse actually telling us it's an overwrite?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -338,20 +338,31 @@ public int open(String path, FuseFileInfo fi) {
...
+        mFileSystem.delete(uri);
+        FileOutStream os = mFileSystem.createFile(uri);
+        long fid = mNextOpenFileId.getAndIncrement();
+        mCreateFileEntries.add(new CreateFileEntry(fid, path, os));
```
LuQQiu, line 349: @maobaolong On a high level, what's the overwrite workflow in Linux and mac?
Is it open() - write(） - write() - release()?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -338,20 +338,31 @@ public int open(String path, FuseFileInfo fi) {
 
   private int openInternal(String path, FuseFileInfo fi) {
     final AlluxioURI uri = mPathResolverCache.getUnchecked(path);
+    final int flags = fi.flags.get();
+    LOG.trace("open({}, 0x{}) [target: {}]", path, Integer.toHexString(flags), uri);
     try {
       long fd = mNextOpenFileId.getAndIncrement();
```
LuQQiu, line 344: Move the `fd` to else block only since it's only used when we open a file for read

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -338,20 +338,31 @@ public int open(String path, FuseFileInfo fi) {
...
+          }
         }
+        mOpenFileEntries.put(fd, is);
+        fi.fh.set(fd);
       }
-      mOpenFileEntries.put(fd, is);
-      fi.fh.set(fd);
       return 0;
     } catch (Throwable e) {
       LOG.error("Failed to open {}: ", path, e);
```
LuQQiu, line 368: Would be better to have a more detailed exception about whether failed to open a file for reading or open a file for overwriting.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -679,8 +690,8 @@ private int chownInternal(String path, long uid, long gid) {
 
   @Override
   public int truncate(String path, long size) {
```
LuQQiu, line 692: Why we change truncate to 0 return?
what's the return int means in this function? it means the operation succeeds or not or how many bytes we truncate?

LuQQiu (commented):
Thanks for supporting opening file for overwriting! Left some comments

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: Ok, I will use the OpenFlags enum.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: AFAIK, it is open(openflag = 0b2) - truncate - write - flush - release

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: Sure, thanks

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: The return of `truncate` is error code, 0 stand for success. It would be better to avoid non-zero size truncate.

LuQQiu (commented):
Thanks for addressing the comments. LGTM overall, except one comment to change back to CreateFileEntries and OpenFileEntries

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -250,7 +250,7 @@ private int getattrInternal(String path, FileStat stat) {
       if (!status.isCompleted()) {
         // Always block waiting for file to be completed except when the file is writing
         // We do not want to block the writing process
-        if (!mCreateFileEntries.contains(PATH_INDEX, path)
+        if (!mOpenFiles.contains(PATH_INDEX, path)
```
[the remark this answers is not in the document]
maobaolong: OK, got it, thank you for your explanation.

maobaolong:
@LuQQiu Revert last commit to original OpenFileEntries and CreateFileEntries. PTAL.

LuQQiu (approved):
LGTM! Thanks for addressing the comments

LuQQiu:

PR #13237, opened 2021-04-15
[DOCFIX] Update README alluxio version

liupan664021:
I notice that in file ./dev/scripts/update-versions.sh, here are lines to update the version of README when updating project version. However, the alluxio version in README is still 2.3.0, because the update version script failed to update README since it has the wrong 'old version'. So I update the version of README to latest version of the project, and then the update version script will correctly recognize the old version of README and update it.
```bash
# Arguments:
#  $1: old version
#  $2: new version
function update_readme() {
    perl -pi -e "s/${1}/${2}/g" README.md
}
```


apc999 (approved):

apc999:

PR #13238, opened 2021-04-15
Use Generic type to make CreateFileEntry extensible

maobaolong:


apc999:
a high level question @maobaolong  , why do we want to make `CreateFileEntry` extensible. Any possibility to add other OutputStream?

maobaolong:
@apc999 I work for the `HCFSFuse`, it is a fuse project based on Alluxio-FUSE, so it use `CreateFileEntry` but not FsOutputStream.

apc999 (approved):

apc999:

PR #13241, opened 2021-04-15
Split read and write timeouts

calvinjia:
In certain environments, the read and write timeout may not make sense to be equal. For example, it may be desirable to fail fast on reads and retry while allowing for a long timeout on writes because retries cannot be attempted.


ns1123 (approved):
LGTM. Not sure if you want to regen the docs?

calvinjia:
@ns1123 Since we are so backlogged let's do it separately

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4101,15 +4101,25 @@ public String toString() {
...
               + "the configuration.")
           .build();
-  public static final PropertyKey USER_STREAMING_DATA_TIMEOUT =
-      new Builder(Name.USER_STREAMING_DATA_TIMEOUT)
-          .setAlias("alluxio.user.network.data.timeout.ms", Name.USER_NETWORK_DATA_TIMEOUT)
+  public static final PropertyKey USER_STREAMING_DATA_READ_TIMEOUT =
+      new Builder(Name.USER_STREAMING_DATA_READ_TIMEOUT)
+          .setAlias("alluxio.user.network.data.timeout.ms", Name.USER_NETWORK_DATA_TIMEOUT,
+              Name.USER_STREAMING_DATA_TIMEOUT)
           .setDefaultValue("1h")
```
ggezer, line 4108: Isn't 1h quite long for both read and write?

ggezer (approved):

core/common/src/main/java/alluxio/conf/PropertyKey.java
calvinjia: Yeah, we made the decision to increase the timeouts here (182e08bfe20073f275ed28731401b8dbbf291072) to avoid spurious issues but I agree it is a case by case decision since in certain environments there are benefits to having short timeouts and retrying.

calvinjia: For writes, I think a longer timeout is fine because we cannot retry.

calvinjia:

calvinjia:

PR #13243, opened 2021-04-16
Track ReplicationChecker's Replication submissions

bradyoo:


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2171,7 +2171,7 @@ public String toString() {
           .build();
   public static final PropertyKey MASTER_REPLICATION_CHECK_INTERVAL_MS =
       new Builder(Name.MASTER_REPLICATION_CHECK_INTERVAL_MS)
-          .setDefaultValue("1min")
+          .setDefaultValue("10sec")
```
yuzhu, line 2174: any particular reason why we want to update this?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -57,7 +59,7 @@
 public final class ReplicationChecker implements HeartbeatExecutor {
   private static final Logger LOG = LoggerFactory.getLogger(ReplicationChecker.class);
   private static final Logger SAMPLING_LOG = new SamplingLogger(LOG, 10L * Constants.MINUTE_MS);
-  private static final long MAX_QUIET_PERIOD_SECONDS = 64;
+  private static final Integer MAX_ACTIVE_JOBS = 10000;
```
yuzhu, line 62: can we make this a fraction of max jobs allowed in job service?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -130,7 +131,22 @@ public void heartbeat() throws InterruptedException {
...
+      final Long jobId = jobIterator.next();
+      try {
+        final Status jobStatus = mReplicationHandler.getJobStatus(jobId);
+        if (jobStatus.isFinished()) {
+          final long inodeId = mActiveJobToInodeID.get(jobId);
+          mActiveInodeIDs.remove(inodeId);
+          jobIterator.remove();
+        }
+      } catch (IOException e) {
+        e.printStackTrace();
```
yuzhu, line 146: LOG instead of printstack?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -200,6 +216,12 @@ public void close() {
   private void checkMisreplicated(Set<Long> inodes, ReplicationHandler handler)
       throws InterruptedException {
     for (long inodeId : inodes) {
+      if (mActiveJobToInodeID.size() >= MAX_ACTIVE_JOBS) {
```
yuzhu, line 219: one thing i am concerned about, if we have 20k files that need replication, the first 10k files are missing in the ufs. This implementation will keep trying those 10k and never get to the second half?

yuzhu (changes requested):
@bradyoo thanks for implementing this.   left some comments

core/common/src/main/java/alluxio/conf/PropertyKey.java
bradyoo: Yeah. The reason why this number had to be so high before was because it is scheduling all of the jobs at once until it's full. Now that it doesn't have to be that, it can run much more frequently.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
bradyoo: There is a DO NOT MERGE here because this was never intended to be merged. Just wanted some feedback.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
bradyoo: Look at mActiveInodeIDS.contains(inodeId) check.. 

On first run of this, it's going to schedule the first 10k files. On the 2nd run, if 3k of them are completed, it will remove the inodeids that are completed off the list and schedule new ones. (though while saying this, I noticed a bug in my code)

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -130,7 +131,22 @@ public void heartbeat() throws InterruptedException {
...
     }
 
-    TimeUnit.SECONDS.sleep(mQuietPeriodSeconds);
+    final Iterator<Long> jobIterator = mActiveJobToInodeID.keySet().iterator();
+
+    while (jobIterator.hasNext()) {
+      final Long jobId = jobIterator.next();
+      try {
+        final Status jobStatus = mReplicationHandler.getJobStatus(jobId);
+        if (jobStatus.isFinished()) {
```
yuzhu, line 140: This includes failed jobs as well. so we will remove failed jobs and cause checker to rerun them. 
Or maybe this is the bug that you mentioned that you saw?

bradyoo: It will absolutely rerun them. Is that bad?

bradyoo:
I turned the tracking from a Set and a Map into a BiMap so that it detects cases when there are multiple jobs for a single inode properly.

bradyoo:
Is the best way to trigger this in testing by using setReplication?

yuzhu:
there is a `ReplicateIntegrationTest` test? perhaps we can add a test there?

bradyoo:
@yuzhu I will but I was talking just testing the code before I write actual tests.

core/common/src/main/java/alluxio/conf/PropertyKey.java
yuzhu: each check has non trivial amount of overhead (checking all pinned files) and can be quite large if the number of pinned files is large. can we leave this at 1min?

core/server/master/src/test/java/alluxio/master/file/replication/ReplicationCheckerTest.java
```diff
@@ -418,4 +440,44 @@ public void heartbeatFilesUnderAndOverReplicated() throws Exception {
     Assert.assertEquals(expected2, mMockReplicationHandler.getEvictRequests());
     Assert.assertEquals(expected1, mMockReplicationHandler.getReplicateRequests());
   }
```
yuzhu, line 442: could we add a testcase where the job keeps failing? after all, this was the failure mode that prompted this change.

yuzhu (changes requested):
@bradyoo looks good overall, left a couple of comments

core/server/master/src/test/java/alluxio/master/file/replication/ReplicationCheckerTest.java
bradyoo: The job failing case wasn't that interesting. It was really the job service being overloaded with too many jobs which was much more interesting.

core/common/src/main/java/alluxio/conf/PropertyKey.java
bradyoo: This is going to have a side effect of replication taking a much longer time than before for a large number of files. That is okay?

bradyoo: In the worst case, this will end up equating to being about 10x slower since we are using at most 10% of the job service capacity. The other thing here is that if job service is at capacity for other reasons, our job retention time for finished jobs is 60 seconds so this needs to be less than 60 seconds if you are writing code that looks for job status. I wrote code such that it assumes it has finished if the job status is unavailable but it shouldn't get there by default.

bradyoo:
Applied feedback but note the warning I commented on moving the replication checker interval to 1 minute.

yuzhu (approved):
LGTM

bradyoo:

bradyoo:

bradyoo:

PR #13247, opened 2021-04-18
Improve check consistency with under storage behavior

yabola:
Each path in Alluxio will be compared with the underlying storage recursively. 
This PR aims to reduce unnecessary comparisons  with under storage path status (maybe an expansive operation) in nested directorys. 
E.g.
1. `/a/b/c` in Alluxio and UFS , then delete UFS directory `/a`, we can assert that `b`, `c` must be inconsistent with UFS. 
2. `/a/b/c` is only created in Alluxio, not on underlying storage , we can assert that  `b`, `c` must be inconsistent with UFS. 
3. .....
we can assert that the parent node is inconsistent, then the child nodes must be also inconsistent. 
remind : dir is considered consistent even if the dir's fingerprint was changed.

Existing test ：alluxio.client.fs.CheckConsistencyIntegrationTest

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1275,7 +1275,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
           auditContext.setAllowed(false);
           throw e;
         }
-        checkConsistencyRecursive(parent, inconsistentUris);
+        checkConsistencyRecursive(parent, inconsistentUris, false);
 
         auditContext.setSucceeded(true);
       }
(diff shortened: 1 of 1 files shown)


yabola:
@jiacheliu3 please take a look~

yabola:

yabola:

yabola:
@gpang @calvinjia please take a look,thx.

yuzhu:
@yabola thanks for the change. one comment, looks like the only time we can have MINOR_INCONSISTENT is when we have files.  so perhaps we can say that if a dir is inconsistent we can assert the children is inconsistent?  so we avoid introducing the concept of MAJOR and MINOR inconsistency?

yabola:
@yuzhu  I agree with you. If a dir is inconsistent, then we can assert the children is inconsistent. I had modified the code .But I want to confirm that dir is considered consistent even if the dir's fingerprint was changed.
Please review again, thx.

calvinjia:
@yabola Thanks for the optimization!
@yuzhu raises a good point, I think in the case we shadow a directory, it is reasonable to state that the files underneath are also inconsistent, so I'm okay with this change. Though this should be a pretty rare case where consistency doesn't make that much sense in the first place.

yuzhu (approved):
LGTM, thanks

yabola:

yuzhu:

PR #13248, opened 2021-04-18
Make MaxFreeAllocator in tiers select the last available block

yabola:
`MaxFreeAllocator` should select the last available block. Suppose there are two tiers, allocate block in any 
tiers
**Before** 
available block-1 in tier1 , reviewer rejected -> no available block in tier2 -> allocate null block (this may lead to evicting block cache)

available block-1 in tier1 , reviewer rejected  -> available block-2 in tier2 , reviewer rejected  ->  allocate block-2
Therefore I think we should choose the last available block

**After this PR**
available block-1 in tier1 , reviewer rejected -> no available block in tier2 -> allocate available block-1

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
@@ -68,19 +68,7 @@ public final class MaxFreeAllocator implements Allocator {
     StorageDirView candidateDirView = null;
 
     if (location.equals(BlockStoreLocation.anyTier())) {
-      for (StorageTierView tierView : mMetadataView.getTierViews()) {
-        candidateDirView = getCandidateDirInTier(tierView, blockSize,
-            BlockStoreLocation.ANY_MEDIUM);
-        if (candidateDirView != null) {
-          if (skipReview || mReviewer.acceptAllocation(candidateDirView)) {
-            break;
-          }
-          // We tried the dir on this tier with max free bytes but that is not good enough.
-          // So we move on to the lower tier.
-          LOG.debug("Allocation rejected for anyTier: {}",
-                  candidateDirView.toBlockStoreLocation());
-        }
-      }
+      candidateDirView = getAvailDirInAnyTier(blockSize, skipReview, BlockStoreLocation.ANY_MEDIUM);
     } else if (location.equals(BlockStoreLocation.anyDirInTier(location.tierAlias()))) {
       StorageTierView tierView = mMetadataView.getTierView(location.tierAlias());
       candidateDirView = getCandidateDirInTier(tierView, blockSize, BlockStoreLocation.ANY_MEDIUM);
@@ -94,18 +82,7 @@ public final class MaxFreeAllocator implements Allocator {
       }
     } else if (location.equals(BlockStoreLocation.anyDirInAnyTierWithMedium(
             location.mediumType()))) {
-      for (StorageTierView tierView : mMetadataView.getTierViews()) {
-        candidateDirView = getCandidateDirInTier(tierView, blockSize, location.mediumType());
-        if (candidateDirView != null) {
-          if (skipReview || mReviewer.acceptAllocation(candidateDirView)) {
-            break;
-          }
-          // We tried the dir on this tier with max free bytes but that is not good enough.
-          // So we move on to the lower tier.
-          LOG.debug("Allocation rejected for anyDirInTierWithMedium: {}",
-                  candidateDirView.toBlockStoreLocation());
-        }
-      }
+      candidateDirView = getAvailDirInAnyTier(blockSize, skipReview, location.mediumType());
     } else {
       // For allocation in a specific directory, we are not checking the reviewer,
       // because we do not want the reviewer to reject it.
@@ -117,6 +94,33 @@ public final class MaxFreeAllocator implements Allocator {
     }
 
     return candidateDirView;
+  }
+
+  /**
+   * Find available dir in tiers for a block with blockSize.
+   *
+   * @param blockSize  the requested block size
+   * @param skipReview wether need to review
+   * @param mediumType the medium type to find a dir
+   * @return available dir; null if fail to find a dir
+   */
+  public StorageDirView getAvailDirInAnyTier(long blockSize, boolean skipReview,
+                                             String mediumType) {
+    StorageDirView finalCandidate = null;
+    for (StorageTierView tierView : mMetadataView.getTierViews()) {
+      StorageDirView currentCandidate = getCandidateDirInTier(tierView, blockSize, mediumType);
+      if (currentCandidate != null) {
+        finalCandidate = currentCandidate;
+        if (skipReview || mReviewer.acceptAllocation(currentCandidate)) {
+          break;
+        }
+        // We tried the dir on this tier with max free bytes but that is not good enough.
+        // So we move on to the lower tier.
+        LOG.debug("Allocation rejected for anyTier: {}",
+                currentCandidate.toBlockStoreLocation());
+      }
+    }
+    return finalCandidate;
   }
 
   /**
--- a/core/server/worker/src/test/java/alluxio/worker/block/reviewer/MockReviewer.java
+++ b/core/server/worker/src/test/java/alluxio/worker/block/reviewer/MockReviewer.java
@@ -1,0 +1,39 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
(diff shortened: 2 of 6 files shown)


core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
```diff
@@ -119,6 +96,32 @@ private StorageDirView allocateBlock(long sessionId, long blockSize,
...
+   * @param skipReview wether need to review
+   * @param mediumType the medium type to find a dir
+   * @return available dir; null if fail to find a dir
+   */
+  public StorageDirView getAvailDirInTiers(long blockSize, boolean skipReview, String mediumType) {
+    StorageDirView finalCandidate = null;
+    for (StorageTierView tierView : mMetadataView.getTierViews()) {
+      StorageDirView currentCandidate = getCandidateDirInTier(tierView, blockSize, mediumType);
+      if (currentCandidate != null) {
+        finalCandidate = currentCandidate;
```
yabola, line 112: **finalCandidate** will mark the last available candidate.

yabola:

yabola:
@ggezer  please take a look, thx

apc999:
@ggezer  can you help take a look at this PR?

jiacheliu3:
hi @yabola A quick comment is, `MaxFreeAllocator` is not the only `Allocator` implementation. Does your change apply to the rest of them (`RoundRobinAllocator` and `GreedyAllocator`)?

yabola:
Hi, @jiacheliu3. I have checked `RoundRobinAllocator ` and `GreedyAllocator `. I found that these two allocators will not allocate block when all reviewer rejected. ( If in  specific tier , `MaxFreeAllocator ` will also like this )
I think we should unify the behavior here.

yabola:
@jiacheliu3 But at least I think behavior should to be consistent in  one allocator.
The current behavior is inconsistent in `MaxFreeAllocator`: 
1. available block-1 in tier1 , reviewer rejected -> no available block in tier2                                -> allocate null block
2. available block-1 in tier1 , reviewer rejected -> available block-2 in tier2 , reviewer rejected -> allocate block-2

I add some more UT for mock Reviewer, please take a look~

LuQQiu:
@jiacheliu3 PTAL the updated PR and comments, thanks!

jiacheliu3:
@yabola @ggezer Actually I prefer the existing approach and behavior. 

**Before**
Case 1
available block-1 in tier1 , reviewer rejected -> no available block in tier2 -> allocate null block -> evict a block and try again if 
`option.isEvictionAllowed()` -> available block-1 in tier1, reviewer rejected -> available block in tier 2

Case 2
available block-1 in tier1 , reviewer rejected -> available block-2 in tier2 , reviewer rejected -> same way of evicting and allocating in tier 2

The motivation of having the Reviewer is, sometimes although there are room in a certain tier, we don't want the allocation to happen. One example is, for an expanding block we normally allocate 1MB first, then when the block expands, it is going to take more space on the tier (and consequently evict other blocks). We prefer not to have this behavior, and that's why we brought in the [ProbablisticBufferReviewer](https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/block/reviewer/ProbabilisticBufferReviewer.java) so that when there's little space on one tier, we try to stop allocating new blocks. We imagine there are similar use cases where we want to stop the allocation even though there is enough space for the allocation (and people will implement `Reviewer` interface to make that happen).

This is my justification why we don't want to select the last available block and choose to evict instead.

**After this PR**
available block-1 in tier1 , reviewer rejected -> no available block in tier2 -> allocate available block-1

Back to the example above, if the availability on tier 1 is just enough for your expanding block (in this call you ask for 1MB but then you will write more, and allocate more space). What happens is you stick your block into tier 1 and then you will incur eviction when your block grows.

I like the way you check all the rest of tiers and if they are all not available, then you go back to the option you skipped. In this way you avoid the issue (if you don't check lower tiers). **Imagine if you don't check lower tiers**, all your expanding blocks (remember they are 1MB initially) will find their way into tier 1. Your tier 1 will always be 99% and all the evictions, and your lower tiers are almost 0%. That would be bad.

Back to my point, this PR does make some sense but I'm not sure the behavior will be better. @ggezer what do you think?

I do have some other comments on the code, but let's settle on the desired behavior first.

yabola:
@jiacheliu3 Thank you for explaining the idea of realization! 
The motivation of  `Reviewer` is  to avoid consequently evict other blocks when the block expands and users can define more flexible Reviewer strategies. this is good for frequently swapping memory. It can also  but I have some confusion. 

I don't understand case2 very well. 
available block-1 in tier1 , reviewer rejected -> available block-2 in tier2 , reviewer rejected -> allocate block-2 
I confused that is it better to assign null block and evict block in tier1 to free more space.
According to the LRU , the priority of the most recent visit is generally higher. we'd better assign it on tier1.

what's more in our production case, there is only one tier, then the behavior is:
available block in tier1 , reviewer rejected -> allocate block
`Review` does not seem to work here.

actually I think if it is better to make the behavior uniform here.

yabola:
I found that 
https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java#L634
is it NPE here?

jiacheliu3:
> I found that
> https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java#L634
> is it NPE here?

@yabola  Good catch!

LuQQiu:
@yabola Can you help fix the conflicts?
@jiacheliu3 can you review this PR or you think it's better for @ggezer to review it?

core/server/worker/src/test/java/alluxio/worker/block/reviewer/MockReviewer.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * An implementation of {@link Reviewer}. This is used to Reviewer unit test.
+ * If a directory's `availableBytes` matches the bytes in `BYTES_TO_REJECT`,
+ * allocation will be rejected.
+ */
+public class MockReviewer implements Reviewer {
+
+  private static final Set<Long> BYTES_TO_REJECT = Sets.newHashSet();
+
+  public static void resetBytesToReject(Set<Long> bytes) {
```
ggezer, line 29: Could you please add comment on this method, otherwise the working of this class is not obvious at first sight.

core/server/worker/src/test/java/alluxio/worker/block/reviewer/MockReviewer.java
```diff
@@ -0,0 +1,39 @@
...
+import com.google.common.collect.Sets;
+
+import java.util.Set;
+
+/**
+ * An implementation of {@link Reviewer}. This is used to Reviewer unit test.
+ * If a directory's `availableBytes` matches the bytes in `BYTES_TO_REJECT`,
+ * allocation will be rejected.
+ */
+public class MockReviewer implements Reviewer {
```
ggezer, line 25: nit: TestReviewer?

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
```diff
@@ -119,6 +96,33 @@ private StorageDirView allocateBlock(long sessionId, long blockSize,
...
+   * Find available dir in tiers for a block with blockSize.
+   *
+   * @param blockSize  the requested block size
+   * @param skipReview wether need to review
+   * @param mediumType the medium type to find a dir
+   * @return available dir; null if fail to find a dir
+   */
+  public StorageDirView getAvailDirInAnyTier(long blockSize, boolean skipReview,
+                                             String mediumType) {
+    StorageDirView finalCandidate = null;
```
ggezer, line 109: In the spirit of MaxFree allocation, I believe first candidate is a better choice for when all candidates are rejected.

ggezer (changes requested):
Thanks @yabola for this fix, and sorry for late review.
Can you also update the title? (Since this change effects MaxFreeAllocator only.)

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
yabola: > In the spirit of MaxFree allocation, I believe first candidate is a better choice for when all candidates are rejected.
@ggezer hi~
behavior now:
available block-1 in tier1 , reviewer rejected -> no available block in tier2 -> allocate null block
available block-1 in tier1 , reviewer rejected -> available block-2 in tier2 , reviewer rejected -> allocate block-2

so you think it is better that?
available block-1 in tier1 , reviewer rejected -> no available block in tier2 -> allocate  block-1 
available block-1 in tier1 , reviewer rejected -> available block-2 in tier2 , reviewer rejected -> allocate  block-1

ggezer: Exactly. Effectively, ignoring the reviewer when it rejects everything and make the allocation as if reviewer was not consulted.

yabola: @jiacheliu3 What do you think is more appropriate behavior?  I can modify the existing implementation if you agree with that.

PR #13253, opened 2021-04-20
Fix alluxio-fuse stat for Alluxio-Worker-FUSE

peterroelants:
Fix `alluxio-fuse stat` which did not pick up the worker process creating the FUSE mount in the new Alluxio-Worker-FUSE integration that can be enabled with `alluxio.worker.fuse.enabled=true`.

I tested this code snipped on 2.5.0-2

### DIFF
--- a/integration/fuse/bin/alluxio-fuse
+++ b/integration/fuse/bin/alluxio-fuse
@@ -108,6 +108,7 @@ fuse_stat() {
     fuse_alluxio_path="$(echo ${fuse_info} | grep -Po '(?<=alluxio\.worker\.fuse\.mount\.alluxio\.path=)\S+(?=\s)')"
     echo -e "pid\tmount_point\talluxio_path"
     echo -e "${fuse_worker_pid}\t${fuse_mount_point}\t${fuse_alluxio_path}"
+    return 0
   fi
   fuse_info=$(ps aux | grep [S]tackMain)
   if [[ -n ${fuse_info} ]]; then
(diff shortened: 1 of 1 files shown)


apc999 (approved):

apc999:
thanks @peterroelants  for the improvement!

apc999:

integration/fuse/bin/alluxio-fuse
```diff
@@ -101,6 +101,15 @@ fuse_stat() {
     echo -e "$(ps aux | grep [A]lluxioFuse | awk -F' ' '{print $2 "\t" $(NF-2) "\t" $NF}')"
     return 0
   fi
+  fuse_info=$(ps aux | grep [A]lluxioWorker | grep alluxio\.worker\.fuse\.enabled=true)
```
LuQQiu, line 104: Alluxio configuration can be passed in via jvm opts or conf/alluxio-site.properties. In this way, we can find out the alluxio mount point and alluxio path configured via jvm opts but not via conf/alluxio-site.properties. Considering adding support for the site properties in the next PR?

peterroelants: I wasn't aware.

Is it possible to get the `conf/alluxio-site.properties` information from ps (proc)? Otherwise we need to parse the `conf/alluxio-site.properties` file.

As an alternative (which will work for all variants), could the FUSE process write its state (mountpoint, pid, alluxio_path) somewhere? This could then be verified against local `mount` and running processes `ps` to ensure it's up and running.

PR #13256, opened 2021-04-20
Improve overwrite logs and enable touch test in JNIFuse

LuQQiu:
A follow-up PR for #13236 to change the logs and enable the touch test back.

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -335,16 +335,17 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
 
   @Override
   public int open(String path, FuseFileInfo fi) {
-    return AlluxioFuseUtils.call(LOG, () -> openInternal(path, fi), "Fuse.Open", "path=%s", path);
+    final int flags = fi.flags.get();
+    boolean overwrite = OpenFlags.valueOf(flags) == OpenFlags.O_WRONLY;
+    String methodName = overwrite ? "Fuse.OpenOverwrite" : "Fuse.Open";
+    return AlluxioFuseUtils.call(LOG, () -> openInternal(path, fi, overwrite),
+        methodName, "path=%s,flags=%s", path, Integer.toHexString(flags));
   }
 
-  private int openInternal(String path, FuseFileInfo fi) {
+  private int openInternal(String path, FuseFileInfo fi, boolean overwrite) {
     final AlluxioURI uri = mPathResolverCache.getUnchecked(path);
-    final int flags = fi.flags.get();
-    OpenFlags openFlag = OpenFlags.valueOf(flags);
-    LOG.trace("open({}, 0x{}) [target: {}]", path, Integer.toHexString(flags), uri);
     try {
-      if (openFlag == OpenFlags.O_WRONLY) {
+      if (overwrite) {
         if (mFileSystem.exists(uri)) {
           mFileSystem.delete(uri);
         }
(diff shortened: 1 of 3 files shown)


integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -335,16 +335,17 @@ private int readdirInternal(String path, long buff, long filter, long offset,
 
   @Override
   public int open(String path, FuseFileInfo fi) {
-    return AlluxioFuseUtils.call(LOG, () -> openInternal(path, fi), "Fuse.Open", "path=%s", path);
+    final int flags = fi.flags.get();
+    boolean overwrite = OpenFlags.valueOf(flags) == OpenFlags.O_WRONLY;
+    String methodName = overwrite ? "Fuse.OpenOverwrite" : "Fuse.Open";
```
apc999, line 340: we can tell if this is `Fuse.OpenOverwrite` from the flag you are logging now, right?
Not sure if it more helpful to differentiate "Fuse.OpenOverwrite" and "Fuse.Open".

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -335,16 +335,17 @@ private int readdirInternal(String path, long buff, long filter, long offset,
...
+    return AlluxioFuseUtils.call(LOG, () -> openInternal(path, fi, overwrite),
+        methodName, "path=%s,flags=%s", path, Integer.toHexString(flags));
```
apc999, line 342:
```suggestion
    return AlluxioFuseUtils.call(LOG, () -> openInternal(path, fi, overwrite),
        methodName, "path=%s,flags=0x%x", path, flags);
```
to avoid calling `Integer.toHexString(flags))` which can be expensive

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -695,11 +696,17 @@ private int chownInternal(String path, long uid, long gid) {
 
   @Override
```
apc999, line 697: can you add comments here saying we only support truncate for special cases like xxx

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: Logging and metrics are two systems. From logging, we can know it's open or openOverwrite, but from metrics, the previous method name cannot differentiate these two circumstances.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
LuQQiu: Make sense!

LuQQiu:
@apc999 Addressed the comments, PTAL, thanks!

tests/src/test/java/alluxio/client/fuse/JNIFuseIntegrationTest.java
```diff
@@ -45,10 +43,4 @@ public void mountFuse(FileSystem fileSystem, String mountPoint, String alluxioRo
   public void umountFuse(String mountPath) throws Exception {
     mFuseFileSystem.umount();
   }
-
-  @Ignore
-  @Override
-  public void touchAndLs() throws Exception {
-    // TODO(lu) Enable the test after https://github.com/Alluxio/alluxio/issues/13090 solved
-  }
```
apc999, line 53: shall we apply the same to `alluxio.server.worker.WorkerFuseIntegrationTest#touchAndLs` ?

apc999 (approved):
LGTM modulo one minor comment left

tests/src/test/java/alluxio/client/fuse/JNIFuseIntegrationTest.java
LuQQiu: good catch!

LuQQiu:
Crashed tests:
Error: 6.582 [ERROR] alluxio.server.ft.journal.MultiMasterJournalTest

apc999:
faling flaky test is unrelated. LGTM and merging your PR

apc999:

PR #13262, opened 2021-04-21
Support overwrite in MigrateConfig

bradyoo:
MigrateConfig has an overwrite flag that didn't do anything.

This has the side effect where distributedCp and distributedMv are now going to overwrite and the retries are going to work in more cases (when destination file is partially written).


job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
```diff
@@ -160,17 +164,26 @@ private static void migrate(MigrateCommand command, WritePType writeType, FileSy
...
+          } catch (Throwable t2) {
+            t.addSuppressed(t2);
+          }
+          throw t;
         }
-        throw t;
+      }
+    } catch (FileAlreadyExistsException e) {
+      if (overwrite) {
+        fileSystem.delete(destinationURI);
```
yuzhu, line 184: @bradyoo a couple of high level questions.  if this does not throw an exception, would it be considered successful and not retried? also if a user specified overwrite, maybe we should delete and then copy and not rely on a retry?

bradyoo: Originally this code was the intuitive version of 

if (exists) { delete } but then I did a couple experiments and it appears that for small files, it's slower than doing this.

yuzhu: i see.  So if we look at MigrateDefinition by itself, it does not actually overwrite any files. Its caller is causing the overwrite behavior, by itself it simply deletes it. 

Does it make sense to have a retry within the MigrateDefinition code, so that any (future/ other) caller of MigrateDefinition will get the overwrite behavior?

bradyoo: Oh nevermind I see the bug. fixing.

bradyoo:
Fixed. Sorry. That was a silly bug :/

LuQQiu:
@bradyoo Can you help add a description for this PR? thanks

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
```diff
@@ -160,19 +164,33 @@ private static void migrate(MigrateCommand command, WritePType writeType, FileSy
...
+        if (overwrite) {
+          fileSystem.delete(destinationURI);
+          retry = true;
+        } else {
+          throw e;
         }
-        throw t;
       }
-    }
+    } while (retry);
```
calvinjia, line 193: Should we limit the number of retries? Would it be too expensive to first check if the file exists, delete on override = true, and leave the retry handling to the job service instead of at the task level?

calvinjia (changes requested):

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
```diff
@@ -160,19 +164,33 @@ private static void migrate(MigrateCommand command, WritePType writeType, FileSy
...
+      retry = false;
       try {
-        IOUtils.copyLarge(in, out, new byte[8 * Constants.MB]);
-      } catch (Throwable t) {
-        try {
-          out.cancel();
-        } catch (Throwable t2) {
-          t.addSuppressed(t2);
+        try (FileInStream in = fileSystem.openFile(new AlluxioURI(source), openFileOptions);
+             FileOutStream out = fileSystem.createFile(destinationURI, createOptions)) {
```
bf8086, line 173: Can we merge the two try-blocks?

bradyoo: You can't. The first try block can't be merged with the 2nd try block because the 2nd try block doesn't reach the catch block when the try statement fails. The 2nd try block can't be merged with the 3rd try block because out variable will not be in scope in the catch statement.

job/server/src/main/java/alluxio/job/plan/migrate/MigrateDefinition.java
bradyoo: In order for one to reach retry repeatedly, one must hit FileAlreadyExistsException repeatedly. The only way you can do that is for fileSystem.delete to fail which would exit the retry loop.

This isn't a real retry. This retry is only a retry when the destination file happens to already exists. The only reason why it's designed this way is because this is faster than doing a file exists check in the common case of when it's not overwriting.

bradyoo (commented):
Both the comments I got seem to be coming from confusion of the code.

Should I just do the simple 

if(overwrite & destination file exists)

check even though it will be slower?

calvinjia:
@bradyoo How much overhead is the existence check? If it is nontrivial (>5% on average with a small file) I'm okay with the current approach.

bradyoo:
I think it's non-trivial but there was enough noise when testing locally to
really be sure.

bradyoo:
Added overwrite option to command

calvinjia (approved):
Could we change overwrite default to false to preserve the old behavior if we cherry-pick to any older branches?

bradyoo:
That is what will happen @calvinjia

calvinjia:

bradyoo:

PR #13267, opened 2021-04-21
Add Search Properties bar to the Web UI Configuration tab

yashnak:
A Search bar is added in order to filter the data by property name. There is a button to enter the search keyword and another to clear the search. Backspacing the whole search key will also clear the search.

### DIFF
--- a/webui/master/src/containers/pages/Configuration/Configuration.scss
+++ b/webui/master/src/containers/pages/Configuration/Configuration.scss
@@ -1,0 +1,22 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+#searchConfig {
+  width: 25%;
+}
+
+.search-container {
+  position: relative;
+  bottom: 10px;
+  left: 70%;
+}
+
+
--- a/webui/master/src/containers/pages/Configuration/Configuration.scss
+++ b/webui/master/src/containers/pages/Configuration/Configuration.scss
@@ -10,13 +10,11 @@
  */
 
 #searchConfig {
-  width: 25%;
+  width: 50%;
 }
 
 .search-container {
-  position: relative;
-  bottom: 10px;
-  left: 70%;
+  align-content: center;
+  left: 20%;
+  bottom: 5px;
 }
-
-
--- a/webui/master/src/containers/pages/Configuration/Configuration.scss
+++ b/webui/master/src/containers/pages/Configuration/Configuration.scss
@@ -11,10 +11,10 @@
 
 #searchConfig {
   width: 50%;
+  margin-left: 50%;
 }
 
 .search-container {
   align-content: center;
-  left: 20%;
   bottom: 5px;
 }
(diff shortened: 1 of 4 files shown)


yashnak:
Screen Shot 2021-04-21 at 4 43 04 PM

webui/master/src/containers/pages/Configuration/Configuration.tsx
```diff
@@ -82,6 +135,30 @@ export class ConfigurationPresenter extends React.Component<AllProps> {
       </div>
     );
   }
+
+  private __searchInputHandler = (e: React.ChangeEvent<HTMLInputElement>): void => {
```
tieujason330, line 139: since the filtering isn't reliant on making any calls to the backend, i think it may worthwhile to do the filtering in realtime as the user types. That'll help simplify the logic, and you wouldn't need to keep track of the clearSearch button status since it'd just be dependent on if there's any input in the search bar atm. ideally you'd still trim and make the input lower-case only (could be done with regex or converted in the inputHandler).

what do you think about doing it this way?

tieujason330 (commented):
can you make the search bar in-line with the `Alluxio Configuration` header? since it's on a separate line, it makes it look mis-aligned

webui/master/src/containers/pages/Configuration/Configuration.tsx
yashnak: Yeah, I think this is a better approach. I think better user experience too.

webui/master/src/containers/pages/Configuration/Configuration.scss
```diff
@@ -10,13 +10,11 @@
...
-  position: relative;
-  bottom: 10px;
-  left: 70%;
-}
-
-
+  align-content: center;
+  left: 20%;
+  bottom: 5px;
+}
```
tieujason330, line 20: EOF newline

tieujason330 (changes requested):
i noticed that the search input div is really wide, so a width scrollbar always appears in the browser. can you try reducing the size of that div to just fit the search input component?

webui/master/src/containers/pages/Configuration/Configuration.test.tsx
```diff
@@ -57,5 +57,18 @@ describe('Configuration', () => {
     it('Matches snapshot', () => {
       expect(shallowWrapper).toMatchSnapshot();
     });
+
+    it('Searches data with valid search key', () => {
+      shallowWrapper.setState({ searchConfig: 'alluxio.' });
+      expect(shallowWrapper.find('filtered-data-body'));
```
tieujason330, line 63: you'll need to add some condition at the end, otherwise i think this `expect` always returns true. when looking for an element, you can do something like `expect(shallowWrapper.find('filtered-data-body').length).toEqual(1);`

webui/master/src/containers/pages/Configuration/Configuration.test.tsx
```diff
@@ -57,5 +57,18 @@ describe('Configuration', () => {
...
+    });
+
+    it('Searches data with invalid search key', () => {
+      const randomStr = Math.random()
+        .toString(36)
+        .substring(20);
+      shallowWrapper.setState({ searchConfig: randomStr });
+      expect(shallowWrapper.find('no-data-body'));
```
tieujason330, line 71: same as above

tieujason330 (changes requested):

tieujason330 (approved):
lgtm, thanks

gpang:
Some lint errors?

```
/usr/src/alluxio/webui/master/src/containers/pages/Configuration/Configuration.test.tsx
Error:   12:21  error  'mount' is defined but never used         @typescript-eslint/no-unused-vars
Error:   12:28  error  'ReactWrapper' is defined but never used  @typescript-eslint/no-unused-vars
Error:   18:10  error  'initialState' is defined but never used  @typescript-eslint/no-unused-vars
```

yashnak:

tieujason330:

PR #13268, opened 2021-04-21
Use default Uid/Gid on FUSE create/mkdir when translation is off

### LINKED ISSUE #13254: fuse with translation disabled still requires username lookup
**Alluxio Version:**
2.5.0-2

**Describe the bug**
I have a FUSE process with user translation turned off by setting `alluxio.fuse.user.group.translation.enabled=false`. This fuse process is running in a Docker container as `alluxio` user, and is sharing the FUSE mount via a bind-mount on the host.
 However when trying to write to the FUSE folder from a different service I get the following error:
 ```
 ExitCodeException exitCode=1: id: ‘1548’: no such user

	at alluxio.shell.ShellCommand.run(ShellCommand.java:71)
	at alluxio.util.ShellUtils.execCommand(ShellUtils.java:238)
	at alluxio.fuse.AlluxioFuseUtils.getUserName(AlluxioFuseUtils.java:108)
	at alluxio.fuse.AlluxioJniFuseFileSystem.setUserGroupIfNeeded(AlluxioJniFuseFileSystem.java:196)
	at alluxio.fuse.AlluxioJniFuseFileSystem.createInternal(AlluxioJniFuseFileSystem.java:232)
	at alluxio.fuse.AlluxioJniFuseFileSystem.lambda$create$0(AlluxioJniFuseFileSystem.java:213)
	at alluxio.fuse.AlluxioFuseUtils.call(AlluxioFuseUtils.java:280)
	at alluxio.fuse.AlluxioJniFuseFileSystem.create(AlluxioJniFuseFileSystem.java:213)
	at alluxio.jnifuse.AbstractFuseFileSystem.createCallback(AbstractFuseFileSystem.java:253)
```

This seems to imply that the FUSE container still needs to be aware of the user with UID 1548. I was hoping that by turning the translation off all files would be owned by by the `alluxio` user, and the FUSE container would not need to be aware of other users.

**To Reproduce**
- Launch Alluxio FUSE in Docker running as `alluxio` user with UID `1001`. Make sure FUSE is mounted in a bind mount to the host. Make sure that `alluxio` user is the only user on this container.
- From the host with a different user with UID `1548`: Write a file to the FUSE mount.
- Result: above error.

Please note that the exact user name and UID does not matter, what matters is that the FUSE mount is being written to by a user that the FUSE container is not aware of.

**Expected behavior**
- Writing by a different user working out-of-the box when user translation is turned off.

**Urgency**
low,
If I write to the fuse folder with a user that has the same UID it works. The error prevents me to seperate user spaces between services.

**Additional context**

apc999:
This PR 
1. use default uid/gid on open/create when user translation is off . This will fix #13254 
2. add username/groupname cache to avoid frequent shell command that get username/groupname from ids.

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -78,8 +78,11 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
   private final String mFsName;
   // Keeps a cache of the most recently translated paths from String to Alluxio URI
   private final LoadingCache<String, AlluxioURI> mPathResolverCache;
+  // Cache Uid<->Username and Gid<->Groupname mapping for local OS
   private final LoadingCache<String, Long> mUidCache;
   private final LoadingCache<String, Long> mGidCache;
+  private final LoadingCache<Long, String> mUsernameCache;
+  private final LoadingCache<Long, String> mGroupnameCache;
   private final int mMaxUmountWaitTime;
   private final AtomicLong mNextOpenFileId = new AtomicLong(0);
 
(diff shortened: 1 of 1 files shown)


LuQQiu (approved):
LGTM！ Looks like we don't have related tests cover the given issue

apc999:
> LGTM！ Looks like we don't have related tests cover the given issue

I don't find a good way to mkdir or create a file as a different local OS user in UT.

apc999:

PR #13270, opened 2021-04-22
Change load command readType to default readType

jiacheliu3:
Seems the load command is using CACHE_PROMOTE no matter what

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
@@ -15,9 +15,11 @@ import alluxio.AlluxioURI;
 import alluxio.Constants;
 import alluxio.annotation.PublicApi;
 import alluxio.cli.CommandUtils;
+import alluxio.client.ReadType;
 import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.URIStatus;
+import alluxio.conf.PropertyKey;
 import alluxio.exception.AlluxioException;
 import alluxio.exception.status.InvalidArgumentException;
 import alluxio.grpc.OpenFilePOptions;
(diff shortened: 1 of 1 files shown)


shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -96,8 +98,11 @@ private void load(AlluxioURI filePath, boolean local) throws AlluxioException, I
         load(newPath, local);
       }
     } else {
+      ReadPType readType = mFsContext.getPathConf(filePath)
```
gpang, line 101: Will `getPathConf` incur an RPC each time?

yuzhu:
actually, i think it makes sense to use CACHE_PROMOTE for load.. I mean, how would load work if the user specified something like NO_CACHE?

jiacheliu3:
> actually, i think it makes sense to use CACHE_PROMOTE for load.. I mean, how would load work if the user specified something like NO_CACHE?

In that case we should hard code `CACHE` I think?

yuzhu:
either CACHE or CACHE_PROMOTE should be fine.  I think when a user explicitly loads something, he/she probably expect the best performance for those files, so moving to top tier aka CACHE_PROMOTE makes sense.  @calvinjia any opinion on this?

jiacheliu3:
> either CACHE or CACHE_PROMOTE should be fine. I think when a user explicitly loads something, he/she probably expect the best performance for those files, so moving to top tier aka CACHE_PROMOTE makes sense. @calvinjia any opinion on this?

In https://github.com/Alluxio/alluxio/commit/3762d75eb611f388895ad9bea0746b17afc7cdd1 the `CACHE` implementation is changed and we changed `readType.default` from `CACHE_PROMOTE` to `CACHE` to avoid moving blocks around while serving a read request. Changing to `CACHE` to be consistent with that.

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jiacheliu3: changed to hard code `CACHE` so this is avoided

jiacheliu3:
Added commit https://github.com/jiacheliu3/alluxio/commit/81d1b758b6614138cab5eda0a63c3c62a27e5e7c but it's not yet reflected on this PR...

jiacheliu3:
Somehow this PR does not work. Created https://github.com/Alluxio/alluxio/pull/13281 instead

PR #13277, opened 2021-04-22
Do not wait for web ui to be ready to be ready in tests

bradyoo:
I believe this is going to be greatly increase the stability of the unit tests.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/MasterProcess.java
+++ b/core/server/common/src/main/java/alluxio/master/MasterProcess.java
@@ -15,6 +15,7 @@ import static alluxio.util.network.NetworkAddressUtils.ServiceType;
 
 import alluxio.Process;
 import alluxio.conf.InstancedConfiguration;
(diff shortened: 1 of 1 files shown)


bradyoo:
@calvinjia What do you think about dropping Webserver serving requirement from waitForReady for Master processes?

totongcute (approved):

calvinjia (commented):
@bradyoo If its a source of test instability, I would prefer checking for the test flag and deciding whether or not to wait for the web server. For production deployments we should ensure the webserver is running.

bradyoo:
Changed to only be the case for tests based on @calvinjia's suggestion

gpang (commented):
Thanks! Does this also work for webui or rest tests?

calvinjia (approved):

bradyoo:
if the webui or rest tests didn't work. The tests for the ci wouldn't be passing.

bradyoo:

PR #13282, opened 2021-04-23
Fix the worker hang issue

### LINKED ISSUE #13255: Alluxio-Worker-FUSE integration hangs
**Alluxio Version:**
2.5.0-2

**Describe the bug**
The [Alluxio Worker-FUSE integration](https://docs.alluxio.io/os/user/edge/en/api/POSIX-API.html#fuse-on-worker-process) hangs after fetching a file from cache.

**To Reproduce**
- Start Alluxio Master
- Start Alluxio Worker with FUSE integration
  - Worker has multi-tiered caching with a tier 0 in MEM and tier 1 on disk.
- On worker container:
  - Load file from S3 (770MB file in this case) `alluxio fs load <alluxio_path>`
  - Read file a few times: `cat ${MOUNT_POINT}/<alluxio_path> > /dev/null`, it usually hangs after 2-4 tries
  
For example, these are some timings I got when running it on my alluxio-worker with FUSE:
```
alluxio@7f23ccee2029:/opt/alluxio$ alluxio fs load <alluxio_path>
<alluxio_path> loaded
alluxio@7f23ccee2029:/opt/alluxio$ date
Tue Apr 20 14:50:23 UTC 2021
alluxio@7f23ccee2029:/opt/alluxio$ date && time cat ${MOUNT_POINT}/<alluxio_path> > /dev/null
Tue Apr 20 14:50:57 UTC 2021

real	0m0.797s
user	0m0.003s
sys	0m0.219s
alluxio@7f23ccee2029:/opt/alluxio$ date && time cat ${MOUNT_POINT}/<alluxio_path> > /dev/null
Tue Apr 20 14:51:03 UTC 2021

real	0m0.497s
user	0m0.004s
sys	0m0.269s
alluxio@7f23ccee2029:/opt/alluxio$ date && time cat ${MOUNT_POINT}/<alluxio_path> > /dev/null
Tue Apr 20 14:51:09 UTC 2021

.... (Hangs here for more than  2 hours now)
```
I added date references to be able to lookup things against the logs. However, I don't notice anything weird in the logs

Note that I can actually read the file two times in less than a second before it hangs the 3rd time. I have been able to replicate this each time I tried it (4 times) after loading the file 2-4 times in a row.

Please find the following logs attached:
- [master.log](https://github.com/Alluxio/alluxio/files/6344709/master.log)
- [job_master.log](https://github.com/Alluxio/alluxio/files/6344712/job_master.log)
- [worker.log](https://github.com/Alluxio/alluxio/files/6344713/worker.log)
- [worker/user_alluxio.log](https://github.com/Alluxio/alluxio/files/6344715/worker-user_alluxio.log)
- [output of `ps -aux` in worker-fuse container](https://github.com/Alluxio/alluxio/files/6344721/worker_ps_aux.log)

**Expected behavior**
Alluxio Worker-FUSE integration to not hang

**Urgency**
High, This makes the Alluxio Worker-FUSE integration unworkable.

**Additional context**
- Services running in custom Docker containers
- Relevant FUSE options:
  ```
  alluxio.fuse.user.group.translation.enabled=false
  alluxio.fuse.jnifuse.enabled=true
  alluxio.user.file.readtype.default=CACHE
  alluxio.fuse.maxwrite.bytes=131072
  alluxio.fuse.shared.caching.reader.enabled=false
  alluxio.user.metadata.cache.enabled=false
  ```
- Worker: `MAX_IDLE_THREADS: 64`
- fusermount version: 2.9.5 (Patched version)
 - Worker `ALLUXIO_JAVA_OPTS`:
   ```
    -Dalluxio.worker.hostname=alluxio-worker-fuse
    -Dalluxio.worker.fuse.enabled=true
    -Dalluxio.worker.fuse.mount.alluxio.path=${ALLUXIO_UFS_PATH_MOUNTED_IN_FUSE}
    -Dalluxio.worker.fuse.mount.point=/mnt/fuse_parent_volume/${ALLUXIO_FUSE_MNT_DIR_NAME}
    -Dalluxio.worker.fuse.mount.options=entry_timeout=7200,attr_timeout=7200
    -Dalluxio.worker.ramdisk.size=${ALLUXIO_WORKER_RAMDISK_SIZE_GB}G
    -Dalluxio.worker.tieredstore.levels=2
    -Dalluxio.worker.tieredstore.level0.alias=MEM
    -Dalluxio.worker.tieredstore.level0.dirs.path=/mnt/tmpfs
    -Dalluxio.worker.tieredstore.level0.dirs.mediumtype=MEM
    -Dalluxio.worker.tieredstore.level0.dirs.quota=${ALLUXIO_WORKER_RAMDISK_SIZE_GB}G
    -Dalluxio.worker.tieredstore.level1.alias=HDD
    -Dalluxio.worker.tieredstore.level1.dirs.path=/mnt/hdd-cache
    -Dalluxio.worker.tieredstore.level1.dirs.mediumtype=HDD
    -Dalluxio.worker.tieredstore.level1.dirs.quota=${ALLUXIO_WORKER_HDD_SIZE_GB}G
  ```

LuQQiu:
Fixes #13255
We observe the behavior that reading a large file will read lock the same block for a large amount times (e.g. 9990) but seldom release the blocks (e.g. 130).
each create(offset,len) in BlockWorkerDataReader.factory will create the block reader and lock the block.  However, the created reader will be lost (will not be close) if there is another create(offset, len) operation. We only close the last created reader left many reader unclosed.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataReader.java
@@ -11,7 +11,6 @@
 
 package alluxio.client.block.stream;
 
-import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.options.InStreamOptions;
 import alluxio.grpc.ReadPType;
 import alluxio.metrics.MetricKey;
@@ -109,19 +108,20 @@ public final class BlockWorkerDataReader implements DataReader {
     /**
      * Creates an instance of {@link Factory}.
      *
-     * @param context the file system context
+     * @param blockWorker the block worker
      * @param blockId the block ID
      * @param chunkSize chunk size in bytes
      * @param options the instream options
      */
-    public Factory(FileSystemContext context, long blockId,
+    public Factory(BlockWorker blockWorker, long blockId,
         long chunkSize, InStreamOptions options)  {
+      Preconditions.checkNotNull(blockWorker);
       mBlockId = blockId;
+      mBlockWorker = blockWorker;
       mChunkSize = chunkSize;
       mIsPromote = options.getOptions().getReadType() == ReadPType.CACHE_PROMOTE;
       mIsPositionShort = options.getPositionShort();
       mOpenUfsBlockOptions = options.getOpenUfsBlockOptions(blockId);
-      mBlockWorker = context.getProcessLocalWorker();
     }
 
     @Override
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataReader.java
@@ -83,9 +83,12 @@ public final class BlockWorkerDataReader implements DataReader {
   }
 
   @Override
-  public void close() {
+  public void close() throws IOException {
     if (mClosed) {
       return;
+    }
+    if (mReader != null) {
+      mReader.close();
     }
     mClosed = true;
   }
@@ -102,8 +105,6 @@ public final class BlockWorkerDataReader implements DataReader {
     private final boolean mIsPositionShort;
     private final Protocol.OpenUfsBlockOptions mOpenUfsBlockOptions;
     private BlockReadRequest mBlockReadRequest;
-    private boolean mClosed;
-    private BlockReader mReader;
 
     /**
      * Creates an instance of {@link Factory}.
@@ -117,7 +118,6 @@ public final class BlockWorkerDataReader implements DataReader {
         long chunkSize, InStreamOptions options)  {
       mBlockId = blockId;
       mChunkSize = chunkSize;
-      mClosed = false;
       mIsPromote = options.getOptions().getReadType() == ReadPType.CACHE_PROMOTE;
       mIsPositionShort = options.getPositionShort();
       mOpenUfsBlockOptions = options.getOpenUfsBlockOptions(blockId);
@@ -129,8 +129,8 @@ public final class BlockWorkerDataReader implements DataReader {
       mBlockReadRequest = new BlockReadRequest(mBlockId, offset, offset + len, mChunkSize,
           mIsPromote, mIsPositionShort, mOpenUfsBlockOptions);
       try {
-        mReader = mBlockWorker.createBlockReader(mBlockReadRequest);
-        return new BlockWorkerDataReader(mReader, offset, len, mChunkSize);
+        BlockReader reader = mBlockWorker.createBlockReader(mBlockReadRequest);
+        return new BlockWorkerDataReader(reader, offset, len, mChunkSize);
       } catch (Exception e) {
         throw new IOException(e);
       }
@@ -142,15 +142,7 @@ public final class BlockWorkerDataReader implements DataReader {
     }
 
     @Override
-    public void close() throws IOException {
-      if (mClosed) {
-        return;
-      }
-      if (mReader != null) {
-        mReader.close();
-      }
-      mClosed = true;
-    }
+    public void close() throws IOException {}
   }
 }
 
--- a/core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
+++ b/core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
@@ -13,7 +13,6 @@ package alluxio.client.block.stream;
 
 import static org.mockito.ArgumentMatchers.any;
 
-import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.URIStatus;
 import alluxio.client.file.options.InStreamOptions;
 import alluxio.conf.AlluxioConfiguration;
@@ -49,16 +48,14 @@ public class BlockWorkerDataReaderTest {
   @Before
   public void before() throws Exception {
     mBlockWorker = Mockito.mock(BlockWorker.class);
-    FileSystemContext context = Mockito.mock(FileSystemContext.class);
-    Mockito.when(context.getProcessLocalWorker()).thenReturn(mBlockWorker);
-
     URIStatus dummyStatus = new URIStatus(new FileInfo()
         .setBlockIds(Collections.singletonList(BLOCK_ID)));
     AlluxioConfiguration conf
         = new InstancedConfiguration(ConfigurationUtils.defaults());
     InStreamOptions options = new InStreamOptions(dummyStatus,
         FileSystemOptions.openFileDefaults(conf), conf);
-    mDataReaderFactory = new BlockWorkerDataReader.Factory(context, BLOCK_ID, CHUNK_SIZE, options);
+    mDataReaderFactory = new BlockWorkerDataReader
+        .Factory(mBlockWorker, BLOCK_ID, CHUNK_SIZE, options);
   }
 
   @Test
--- a/core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
+++ b/core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
@@ -95,7 +95,7 @@ public class BlockWorkerDataReaderTest {
     DataReader dataReader = mDataReaderFactory.create(0, len);
     validateBuffer(dataReader.readChunk(), 0, CHUNK_SIZE);
     Assert.assertEquals(CHUNK_SIZE, dataReader.pos());
-    validateBuffer(dataReader.readChunk(), CHUNK_SIZE, len);
+    validateBuffer(dataReader.readChunk(), CHUNK_SIZE, CHUNK_SIZE);
     Assert.assertEquals(len, dataReader.pos());
     dataReader.close();
   }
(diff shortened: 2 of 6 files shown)


LuQQiu:
WIP again, thinking that we should only have one blockreader for reading chunks inside one thread. That reader should be in factory level instead of creating one for reading each chunk

core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataReader.java
```diff
@@ -49,18 +49,19 @@
...
    */
-  private BlockWorkerDataReader(BlockReader reader,
-      long offset, long len, long chunkSize) {
-    mReader = reader;
-    Preconditions.checkArgument(chunkSize > 0);
-    mPos = offset;
-    mEnd = Math.min(mReader.getLength(), offset + len);
-    mChunkSize = chunkSize;
+  private BlockWorkerDataReader(BlockWorker blockWorker,
+      BlockReadRequest request) throws IOException {
```
apc999, line 56: we shall avoid throw exceptions in a constructor.
a constructor is meant to be something light weight.
If an exception is potentially thrown, move it to the factory method (that's what we did)

core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataReader.java
```diff
@@ -83,10 +84,13 @@ public long pos() {
...
 
   @Override
-  public void close() {
+  public void close() throws IOException {
     if (mClosed) {
       return;
     }
+    if (mReader != null) {
+      mReader.close();
+    }
```
apc999, line 93: isn't this enough to ensure `mReader` is closed?
why do we need to move `mBlockWorker.createBlockReader()` to constructor of `BlockWorkerDataReader`?

LuQQiu:
@apc999 Move the createBlockReader back to factory, PTAL.
I am considering should we reuse the blockReader for all readChunks?
Otherwise we keep creating new BlockReader for each  readChunk

apc999 (approved):
Can you add a UT to cover this case? thanks

LuQQiu:
@apc999 Added the unit test for BlockWorkerDataReader, cover the close() and general readChunk(), PTAL, thanks!

core/common/src/main/java/alluxio/util/io/BufferUtils.java
```diff
@@ -328,6 +328,7 @@ public static boolean equalIncreasingByteBuffer(int start, int len, ByteBuffer b
       return false;
     }
     buf.rewind();
+    int wvl = buf.remaining();
```
apc999, line 331: why is this needed? please comment

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
```diff
@@ -0,0 +1,124 @@
...
+  private static final Random RANDOM = new Random();
+
+  private BlockWorker mBlockWorker;
+  private BlockWorkerDataReader.Factory mDataReaderFactory;
+
+  @Before
+  public void before() throws Exception {
+    mBlockWorker = Mockito.mock(BlockWorker.class);
+    FileSystemContext context = Mockito.mock(FileSystemContext.class);
+    Mockito.when(context.getProcessLocalWorker()).thenReturn(mBlockWorker);
```
apc999, line 53: code like this will be quite challenging to maintain as it requires an understanding of how `BlockWorkerDataReader.Factory` uses `context`, which is totally internal and subject to change.

I would suggest to modify `BlockWorkerDataReader.Factory` to take `BlockWorker` as an argument rather than `Context`. In this way we don't need set up tests based on the understanding of the internal implementation of `BlockWorkerDataReader`, which will be very hard to maintain

core/common/src/main/java/alluxio/util/io/BufferUtils.java
LuQQiu: oh, a mistake, will remove it

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
LuQQiu: Yeah make sense, that will make the code cleaner and no need to use FileSystemContext

LuQQiu:
Addressed the comments, PTAL, thanks! @apc999

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
```diff
@@ -0,0 +1,121 @@
...
+    mDataReaderFactory = new BlockWorkerDataReader
+        .Factory(mBlockWorker, BLOCK_ID, CHUNK_SIZE, options);
+  }
+
+  @Test
+  public void createAndClose() throws Exception {
+    byte[] bytes = new byte[128];
+    RANDOM.nextBytes(bytes);
+
+    MockBlockReader blockReader = new MockBlockReader(bytes);
```
apc999, line 66:
```suggestion
    BlockReader blockReader = new MockBlockReader(bytes);
```

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
```diff
@@ -0,0 +1,121 @@
...
+    Mockito.when(mBlockWorker.createBlockReader(any(BlockReadRequest.class)))
```
apc999, line 67: consider using `NoopBlockWorker` to inject your logic?
I am really not a big fan of using `Mockito` which is hard to maintain

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
```diff
@@ -0,0 +1,121 @@
...
+        .thenReturn(blockReader);
+    DataReader dataReader = mDataReaderFactory.create(1, 100);
+
+    MockBlockReader blockReaderTwo = new MockBlockReader(bytes);
```
apc999, line 71:
```suggestion
    BlockReader blockReader2 = new MockBlockReader(bytes);
```

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
```diff
@@ -0,0 +1,121 @@
...
+  private static final long BLOCK_ID = 1L;
+  private static final int CHUNK_SIZE = 128;
+  private static final Random RANDOM = new Random();
+
+  private BlockWorker mBlockWorker;
+  private BlockWorkerDataReader.Factory mDataReaderFactory;
+
+  @Before
+  public void before() throws Exception {
+    mBlockWorker = Mockito.mock(BlockWorker.class);
```
apc999, line 50: Consider using `NoopBlockWorker`?  Perhaps we should rename it to `MockBlockWorker`

core/client/fs/src/test/java/alluxio/client/block/stream/BlockWorkerDataReaderTest.java
```diff
@@ -0,0 +1,121 @@
...
+        .thenReturn(blockReaderTwo);
+    DataReader dataReaderTwo = mDataReaderFactory.create(10, 30);
+
+    Assert.assertFalse(blockReader.isClosed());
+    Assert.assertFalse(blockReaderTwo.isClosed());
+    dataReaderTwo.close();
+    Assert.assertFalse(blockReader.isClosed());
+    Assert.assertTrue(blockReaderTwo.isClosed());
+    dataReader.close();
+    Assert.assertTrue(blockReader.isClosed());
```
apc999, line 82: I think these are internal logic w.r.t. `BlockWorkerDataReader`. I don't like unit tests relying on internal logic because that logic is always subject to change, making the unit tests hard to maintain. A unit test of `BlockWorkerDataReader` should focus on its public method, setting expectation and check results.

apc999 (changes requested):
I have concerns on the unit tests. basically it is not a good practise to write unit tests that sets up tests with target class's internal member and its behavior, and checking the internal member's state after test. Tests like this are very challenging to maintain. Unit test of class foo should be testing its public method on the promise and checking the results, and it should be totally independent from its internal implementation.

apc999:
How about creating a `DefaultBlockWorker` instance, and pass it to  `BlockWorkerDataReader.Factory`. in order to do so, you probably need to make this test class in `core/server/worker` module rather than `core/client/fs` module.  To setup a instance  of `DefaultBlockWorker`, you can check out `alluxio/worker/block/DefaultBlockWorkerTest.java`

After having a real and working `DefaultBlockWorker`, you can simply check basic behavior, as well as creating/destroying `BlockWorkerDataReader` and make sure it is not hanging.

LuQQiu:
For the test side, we want to avoid things like the bug happening again. The best practice is adding a test similar to the bug behavior. Like using actual Fuse on worker and read data through the mount point. I will add this part in our AbstractFuseIntegrationTest. In this way, we don't test any internal logics but test the desire behavior we want.

I couldn't think of a good way to do what we want to test in unit test without touching internal logics.
Our DefaultBlockWorkerTest is a disaster.... it internally contains a lot of PowerMockito logics. So it's not a real DefaultBlockWorker. It uses the `Mockito.spy(BlockStore)` so it will not reach the actually code to create 1000 block read locks.

Agree that i change the MockIito.mock(BlockWorker.class) to a class extending NoopBlockWorker.

LuQQiu:
Changed Mockito.mock(BlockWorker) to use MockBlockWorker extends NoopBlockWorker.
WIP adding integration test for reading large files with WorkerFuse

apc999 (approved):
LGTM for this PR given we want this fix merged ASAP. in terms of adding integration test, let's discuss offline on the unit test and my view is if it does not require master setup, we can still better to make it as unit test (less expensive, less flaky).

apc999:

PR #13292, opened 2021-04-26
Include custom jars to lib by flag when generating tarballs

liupan664021:
`./dev/scripts/generate-tarballs` is really convinient for alluxio developers to generate custom alluxio tarballs. But as a custom version, not all underfs and tool jars under ./lib are necessary to be added to the tarballs. Maybe a custom `libjars` flag is helpful to slim the tarball by size when generating custom tarballs.


dev/scripts/src/alluxio.org/build-distribution/cmd/root.go
```diff
@@ -48,4 +49,7 @@ func additionalFlags(cmd *flag.FlagSet) {
 	cmd.BoolVar(&debugFlag, "debug", false, "whether to run this tool in debug mode to generate additional console output")
 	cmd.StringVar(&ufsModulesFlag, "ufs-modules", strings.Join(defaultModules(ufsModules), ","),
 		fmt.Sprintf("a comma-separated list of ufs modules to compile into the distribution tarball(s). Specify 'all' to build all ufs modules. Supported ufs modules: [%v]", strings.Join(validModules(ufsModules), ",")))
+	cmd.StringVar(&includedLibJarsFlag, "lib-jars", "all",
+		"a comma-separated list of jars under lib/ needed to included except underfs-hdfs modules. All jars under lib/ will be included by default."+
```
Xenorith, line 53: slight rewording of the flag description:
`a comma-separted list of jars under lib/ to include in addition to all underfs-hdfs modules. All jars under lib/ will be included by default.`

also move this flag declaration up a line for alphabetical order in declaration of flags

dev/scripts/src/alluxio.org/build-distribution/cmd/root.go
```diff
@@ -21,6 +21,7 @@ var (
 	customUfsModuleFlag string
 	debugFlag           bool
 	ufsModulesFlag      string
+	includedLibJarsFlag string
```
Xenorith, line 24: move up a line to maintain alphabetical order

Xenorith (changes requested):
thanks for this change; this flag makes a lot of sense for building a small custom tarball. just a few small changes but otherwise looks good

also dont worry about retriggering the build to pass all the unit tests, as there are a few flakey ones. the most critical job for this PR is the `Checkstyle, Findbugs, Doc Check, etc.` one; the other ones aren't as relevant since the PR doesn't change any java code

dev/scripts/src/alluxio.org/build-distribution/cmd/root.go
liupan664021: Thanks so much for the suggessions.

dev/scripts/src/alluxio.org/build-distribution/cmd/common.go
```diff
@@ -63,6 +63,25 @@ var ufsModules = map[string]module{
 	"ufs-hadoop-3.3": {"hadoop-3.3", false, "-pl underfs/hdfs -Pufs-hadoop-3 -Dufs.hadoop.version=3.3.0 -PhdfsActiveSync"},
 }
 
+var libJars = map[string]struct{}{
+	"integration-tools-hms":        {},
```
Xenorith, line 67: i just noticed that the list of jars names here do not match with the list of jars removed from generate-tarball.go

this new list is missing ozone, cephfs, and cephfs-hadoop but adds kodo. is this change intentional? i haven't taken a look at the list of published jars for a while so my knowledge may be out of date

liupan664021: Sorry, my fault. I think I just missed `ozone, cephfs, cephfs-hadoop`. 

As far as `kodo`, it is a new underfs? I don't know. I notice that `kodo` is included in the `underfs/pom.xml`, so it is supposed to be included in a release version. But `kodo` is missed in the original generate-tarball.go. So I added it to the libjar names.

Xenorith: for this PR, can we keep all the original ufs jars and not make any modifications to the list? i don't think there is any issue in adding the kodo jar, but i would like to get the opinion of other people, so it would be better suited as a separate PR

liupan664021: Of course. I'll fix it back.

Xenorith (approved):

Xenorith:

PR #13298, opened 2021-04-26
Validate flag for docGen command

Limuzhou:
If validate flag added, generate temporary doc files and compare with existing doc files to see if there are changes.

### DIFF
--- a/shell/src/main/java/alluxio/cli/DocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/DocGenerator.java
@@ -35,6 +35,7 @@ public class DocGenerator {
 
   private static final String METRIC_OPTION_NAME = "metric";
   private static final String CONF_OPTION_NAME = "conf";
+  private static final String VALIDATE_OPTION_NAME = "validate";
 
   private static final Option METRIC_OPTION =
       Option.builder().required(false).longOpt(METRIC_OPTION_NAME).hasArg(false)
@@ -42,9 +43,11 @@ public class DocGenerator {
   private static final Option CONF_OPTION =
       Option.builder().required(false).longOpt(CONF_OPTION_NAME).hasArg(false)
           .desc("the configuration properties used by the master.").build();
-
+  private static final Option VALIDATE_OPTION =
+      Option.builder().required(false).longOpt(VALIDATE_OPTION_NAME).hasArg(false)
+          .desc("the configuration properties used by the master.").build();
   private static final Options OPTIONS =
-      new Options().addOption(METRIC_OPTION).addOption(CONF_OPTION);
+      new Options().addOption(METRIC_OPTION).addOption(CONF_OPTION).addOption(VALIDATE_OPTION);
 
   /**
    * Main entry for this util class.
@@ -52,6 +55,7 @@ public class DocGenerator {
    * @param args arguments for command line
    */
   public static void main(String[] args) throws IOException {
+    boolean validate = false;
     if (args.length != 0) {
       CommandLineParser parser = new DefaultParser();
       CommandLine cmd;
@@ -61,15 +65,21 @@ public class DocGenerator {
         printHelp("Unable to parse input args: " + e.getMessage());
         return;
       }
+      if (cmd.hasOption(VALIDATE_OPTION_NAME)){
+        validate = true;
+      }
       if (cmd.hasOption(METRIC_OPTION_NAME)) {
-        MetricsDocGenerator.generate();
+        System.out.println("Validate flag is: " + validate);
+        MetricsDocGenerator.generate(validate);
       }
       if (cmd.hasOption(CONF_OPTION_NAME)) {
-        ConfigurationDocGenerator.generate();
+        System.out.println("Validate flag is: " + validate);
+        ConfigurationDocGenerator.generate(validate);
       }
     } else {
-      MetricsDocGenerator.generate();
-      ConfigurationDocGenerator.generate();
+      System.out.println("Validate flag is: " + validate);
+      MetricsDocGenerator.generate(validate);
+      ConfigurationDocGenerator.generate(validate);
     }
   }
 
--- a/shell/src/main/java/alluxio/cli/DocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/DocGenerator.java
@@ -45,7 +45,8 @@ public class DocGenerator {
           .desc("set true to specifically generate config files").build();
   private static final Option VALIDATE_OPTION =
       Option.builder().required(false).longOpt(VALIDATE_OPTION_NAME).hasArg(false)
-          .desc("set true to validate that the existing generated output files are up to date").build();
+          .desc("set true to validate that the existing generated output files are up to date")
+          .build();
   private static final Options OPTIONS =
       new Options().addOption(METRIC_OPTION).addOption(CONF_OPTION).addOption(VALIDATE_OPTION);
 
--- a/shell/src/main/java/alluxio/cli/DocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/DocGenerator.java
@@ -39,13 +39,13 @@ public class DocGenerator {
 
   private static final Option METRIC_OPTION =
       Option.builder().required(false).longOpt(METRIC_OPTION_NAME).hasArg(false)
-          .desc("the configuration properties used by the master.").build();
+          .desc("set true to specifically generate metric files.").build();
   private static final Option CONF_OPTION =
       Option.builder().required(false).longOpt(CONF_OPTION_NAME).hasArg(false)
-          .desc("the configuration properties used by the master.").build();
+          .desc("set true to specifically generate config files").build();
   private static final Option VALIDATE_OPTION =
       Option.builder().required(false).longOpt(VALIDATE_OPTION_NAME).hasArg(false)
-          .desc("the configuration properties used by the master.").build();
+          .desc("set true to validate that the existing generated output files are up to date").build();
   private static final Options OPTIONS =
       new Options().addOption(METRIC_OPTION).addOption(CONF_OPTION).addOption(VALIDATE_OPTION);
 
--- a/shell/src/main/java/alluxio/cli/DocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/DocGenerator.java
@@ -65,10 +65,10 @@ public class DocGenerator {
         printHelp("Unable to parse input args: " + e.getMessage());
         return;
       }
-      if (cmd.hasOption(VALIDATE_OPTION_NAME)){
+      if (cmd.hasOption(VALIDATE_OPTION_NAME)) {
         validate = true;
       }
-      if ((!cmd.hasOption(METRIC_OPTION_NAME)) && (!cmd.hasOption(CONF_OPTION_NAME))){
+      if ((!cmd.hasOption(METRIC_OPTION_NAME)) && (!cmd.hasOption(CONF_OPTION_NAME))) {
         MetricsDocGenerator.generate(validate);
         ConfigurationDocGenerator.generate(validate);
       }
--- a/shell/src/main/java/alluxio/cli/DocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/DocGenerator.java
@@ -68,16 +68,17 @@ public class DocGenerator {
       if (cmd.hasOption(VALIDATE_OPTION_NAME)){
         validate = true;
       }
+      if ((!cmd.hasOption(METRIC_OPTION_NAME)) && (!cmd.hasOption(CONF_OPTION_NAME))){
+        MetricsDocGenerator.generate(validate);
+        ConfigurationDocGenerator.generate(validate);
+      }
       if (cmd.hasOption(METRIC_OPTION_NAME)) {
-        System.out.println("Validate flag is: " + validate);
         MetricsDocGenerator.generate(validate);
       }
       if (cmd.hasOption(CONF_OPTION_NAME)) {
-        System.out.println("Validate flag is: " + validate);
         ConfigurationDocGenerator.generate(validate);
       }
     } else {
-      System.out.println("Validate flag is: " + validate);
       MetricsDocGenerator.generate(validate);
       ConfigurationDocGenerator.generate(validate);
     }
--- a/shell/src/main/java/alluxio/cli/DocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/DocGenerator.java
@@ -69,7 +69,7 @@ public class DocGenerator {
       if (cmd.hasOption(VALIDATE_OPTION_NAME)) {
         validate = true;
       }
-      if ((!cmd.hasOption(METRIC_OPTION_NAME)) && (!cmd.hasOption(CONF_OPTION_NAME))) {
+      if (!cmd.hasOption(METRIC_OPTION_NAME) && !cmd.hasOption(CONF_OPTION_NAME)) {
         MetricsDocGenerator.generate(validate);
         ConfigurationDocGenerator.generate(validate);
       }
--- a/shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
@@ -46,17 +46,24 @@ public final class ConfigurationDocGenerator {
   private static final String CSV_FILE_DIR = "docs/_data/table/";
   private static final String YML_FILE_DIR = "docs/_data/table/en/";
   public static final String CSV_FILE_HEADER = "propertyName,defaultValue";
-  public static final String TEMP_PREFIX = "temp-";
+  private static final String TEMP_PREFIX = "temp-";
 
-  public static final String[] CSV_FILE_NAMES = {"user-configuration.csv",
-      "master-configuration.csv", "worker-configuration.csv",
-      "security-configuration.csv", "common-configuration.csv",
-      "cluster-management-configuration.csv"};
+  public static final String[] CSV_FILE_NAMES = {
+      "cluster-management-configuration.csv",
+      "common-configuration.csv",
+      "master-configuration.csv",
+      "security-configuration.csv",
+      "user-configuration.csv",
+      "worker-configuration.csv"
+  };
 
-  public static final String[] YML_FILE_NAMES = {"user-configuration.yml",
-      "master-configuration.yml", "worker-configuration.yml",
-      "security-configuration.yml", "common-configuration.yml",
-      "cluster-management-configuration.yml"
+  public static final String[] YML_FILE_NAMES = {
+      "cluster-management-configuration.yml",
+      "common-configuration.yml",
+      "master-configuration.yml",
+      "security-configuration.yml",
+      "user-configuration.yml",
+      "worker-configuration.yml"
   };
 
   private ConfigurationDocGenerator() {} // prevent instantiation
@@ -149,16 +156,16 @@ public final class ConfigurationDocGenerator {
       try {
         closer.close();
         if (validate) {
-          compareFiles(CSV_FILE_NAMES, filePath);
+          compareFiles(CSV_FILE_NAMES, filePath, "CSV");
         }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing Property Key CSV FileWriter", e);
-        for (String fileName : CSV_FILE_NAMES) {
-          String fileNameTemp = TEMP_PREFIX.concat(fileName);
-          File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
-          if (tempFile.exists()) {
-            tempFile.delete();
-          }
+      }
+      for (String fileName : CSV_FILE_NAMES) {
+        String fileNameTemp = TEMP_PREFIX.concat(fileName);
+        File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
+        if (tempFile.exists()) {
+          tempFile.delete();
         }
       }
     }
@@ -235,17 +242,17 @@ public final class ConfigurationDocGenerator {
       try {
         closer.close();
         if (validate) {
-          compareFiles(YML_FILE_NAMES, filePath);
+          compareFiles(YML_FILE_NAMES, filePath, "YML");
         }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing YML files for description of Property Keys "
             + "FileWriter", e);
-        for (String fileName : CSV_FILE_NAMES) {
-          String fileNameTemp = TEMP_PREFIX.concat(fileName);
-          File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
-          if (tempFile.exists()) {
-            tempFile.delete();
-          }
+      }
+      for (String fileName : YML_FILE_NAMES) {
+        String fileNameTemp = TEMP_PREFIX.concat(fileName);
+        File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
+        if (tempFile.exists()) {
+          tempFile.delete();
         }
       }
     }
@@ -256,8 +263,10 @@ public final class ConfigurationDocGenerator {
    *
    * @param fileNames the name of the file
    * @param filePath the path to the file
+   * @param fileType the type of the file that we are comparing (CSV/YML)
    */
-  public static void compareFiles(String[] fileNames, String filePath) throws IOException {
+  public static void compareFiles(String[] fileNames, String filePath, String fileType)
+      throws IOException {
     boolean hasDiff = false;
 
     for (String fileName : fileNames) {
@@ -267,11 +276,9 @@ public final class ConfigurationDocGenerator {
         hasDiff = true;
         System.out.println("Config file " + fileName + " changed.");
       }
-      File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
-      tempFile.delete();
     }
     if (!hasDiff) {
-      System.out.println("No change in config CSV file detected.");
+      System.out.println("No change in config " + fileType + " files detected.");
     }
   }
 
--- a/shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
@@ -161,8 +161,8 @@ public final class ConfigurationDocGenerator {
           compareFiles(CSV_FILE_NAMES, filePath, CSV_SUFFIX);
         }
       } catch (IOException e) {
-        LOG.error("Error while flushing/closing Property Key CSV FileWriter " +
-            "or compiling generated files to the existing files if validate flag is set", e);
+        LOG.error("Error while flushing/closing Property Key CSV FileWriter "
+            + "or compiling generated files to the existing files if validate flag is set", e);
       }
       if (validate) {
         for (String fileName : CSV_FILE_NAMES) {
@@ -251,7 +251,8 @@ public final class ConfigurationDocGenerator {
         }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing YML files for description of Property Keys "
-            + "FileWriter or compiling generated files to the existing files if validate flag is set", e);
+            + "FileWriter or compiling generated files to the existing files if validate flag "
+            + "is set", e);
       }
       if (validate) {
         for (String fileName : YML_FILE_NAMES) {
--- a/shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
+++ b/shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
@@ -20,10 +20,13 @@ import alluxio.util.io.PathUtils;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.Closer;
 import org.apache.commons.lang3.StringEscapeUtils;
+import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.FileWriter;
+import java.io.File;
+import java.nio.charset.StandardCharsets;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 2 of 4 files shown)


LuQQiu:
@Limuzhou Please fix the checkstyle issues found by running `mvn checkstyle:checkstyle` locally, thanks!

Xenorith:
these build failures also need to be addressed

2 checkstyle failures on the test file:
```
2021-04-28T01:44:25.5484603Z /usr/src/alluxio/tests/src/test/java/alluxio/client/cli/fs/ConfigurationDocGeneratorTest.java:123:67: error: ',' is not followed by whitespace.
2021-04-28T01:44:25.5487162Z /usr/src/alluxio/tests/src/test/java/alluxio/client/cli/fs/ConfigurationDocGeneratorTest.java:144:67: error: ',' is not followed by whitespace.
```

and 12 unit test failures in `ConfigurationDocGeneratorTest`

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -149,7 +160,13 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
       // HashMap for FileWriter per each category
       Map<String, FileWriter> fileWriterMap = new HashMap<>();
       for (String fileName : fileNames) {
-        fileWriter = new FileWriter(PathUtils.concatPath(filePath, fileName));
+        if (validate) {
```
Xenorith, line 163: this looks identical to the newly added section in `writeCSVFile`. refactor to a helper function?

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
     String homeDir = new InstancedConfiguration(ConfigurationUtils.defaults())
         .get(PropertyKey.HOME);
     // generate CSV files
     String filePath = PathUtils.concatPath(homeDir, CSV_FILE_DIR);
-    writeCSVFile(defaultKeys, filePath);
+    writeCSVFile(defaultKeys, filePath, validate);
+
+    if (validate) {
+      boolean hasDiff = false;
+      String[] fileNames = {"user-configuration.csv", "master-configuration.csv",
```
Xenorith, line 239: this list is identical to the one in `writeCSVFile`. this should not be copied; if someone adds a new csv file, there's a good chance that it will be added onto one list but not the other. as much as possible, there should always be a "single source of truth", especially when there are constants involved

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+          "worker-configuration.csv", "security-configuration.csv",
+          "common-configuration.csv", "cluster-management-configuration.csv"};
+
+      for (String fileName : fileNames) {
+        String tempPrefix = "temp-";
```
Xenorith, line 244: this `temp-` prefix must match the same `temp-` prefix set in the `writeCSVFile` and `writeYMLFile`. if any one of these string prefixes changed, they must all be changed -> create a constant for this

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+        String fileNameTemp = tempPrefix.concat(fileName);
+
+        // create the ProcessBuilder for the diff command
+        ProcessBuilder pb = new ProcessBuilder("diff", PathUtils.concatPath(filePath, fileName),
```
Xenorith, line 248: the `diff` command is generally available in most operating systems but it would be more robust if we did not call this as a system command. i would much prefer to reimplement this to ensure the validate functionality works in any environment

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+        // redirect the diff output to a string
+        String output = IOUtils.toString(pb.start().getInputStream(), StandardCharsets.UTF_8);
+
+        if (output.length() > 0) {
+          System.out.println(output);
+          hasDiff = true;
+          System.out.println("Config file " + fileName + " changed.");
+        }
+        File f = new File(PathUtils.concatPath(filePath, fileName));
+        // f = new File(PathUtils.concatPath(filePath, fileName));
```
Xenorith, line 259: leftover comment?

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+    }
+
     // generate YML files
     filePath = PathUtils.concatPath(homeDir, YML_FILE_DIR);
-    writeYMLFile(defaultKeys, filePath);
+    writeYMLFile(defaultKeys, filePath, validate);
+
+    if (validate) {
+      boolean hasDiff = false;
+      String[] fileNames = {"user-configuration.yml", "master-configuration.yml",
```
Xenorith, line 274: this section is also very very similar to the validate section for csv files and should be refactored into a common function rather than copied

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+            PathUtils.concatPath(filePath, fileNameTemp));
+        // redirect the diff output to a string
+        String output = IOUtils.toString(pb.start().getInputStream(), StandardCharsets.UTF_8);
+
+        if (output.length() > 0) {
+          System.out.println(output);
+          hasDiff = true;
+          System.out.println("Config file " + fileName + " changed.");
+        }
+        File f = new File(PathUtils.concatPath(filePath, fileName));
```
Xenorith, line 258: this section renames the temp generated file into the actual generated file, but a validation should not touch any files.

on the other hand, the generated files do need to get deleted

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
     Collection<? extends PropertyKey> defaultKeys = PropertyKey.defaultKeys();
     defaultKeys.removeIf(key -> key.isHidden());
     String homeDir = new InstancedConfiguration(ConfigurationUtils.defaults())
         .get(PropertyKey.HOME);
     // generate CSV files
     String filePath = PathUtils.concatPath(homeDir, CSV_FILE_DIR);
-    writeCSVFile(defaultKeys, filePath);
+    writeCSVFile(defaultKeys, filePath, validate);
+
+    if (validate) {
```
Xenorith, line 237: since you have all the info you need within the writeCSVFile and writeYAMLFile functions, i think the validation logic should also belong within each respective function. since they would be handling the write and optionally the validate step, i would also rename them from write*File to handle*File

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -206,17 +223,82 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
 
   /**
    * Generates the configuration docs.
+   * @param validate validate flag
    */
-  public static void generate() throws IOException {
+  public static void generate(boolean validate) throws IOException {
```
Xenorith, line 228: note that the generate function can throw an IOException. if the validate flag is true and a temp file is generated, it needs to be deleted even if an exception is thrown, otherwise the git workspace will be dirty. there should be some exception handling to handle this scenario and ensure that the generated file is deleted even if an exception is thrown

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -68,19 +72,26 @@ public static void generate() throws IOException {
...
       FileWriter ymlFileWriter;
       for (String category : CATEGORIES) {
-        csvFileWriter = new FileWriter(PathUtils
-            .concatPath(csvFolder, category + "-metrics." + CSV_SUFFIX));
-        csvFileWriter.append(CSV_FILE_HEADER + "\n");
-        ymlFileWriter = new FileWriter(PathUtils
-            .concatPath(ymlFolder, category + "-metrics." + YML_SUFFIX));
+        if (validate) {
+          csvFileWriter = new FileWriter(PathUtils
+              .concatPath(csvFolder, ("temp-" + category + "-metrics." + CSV_SUFFIX)));
```
Xenorith, line 84: similar comment regarding the use of `temp-` string prefix; if it needs to be the same across several places, share the same reference

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -110,6 +120,39 @@ public static void generate() throws IOException {
             key, StringEscapeUtils.escapeHtml4(metricKey.getDescription().replace("'", "''"))));
       }
     }
+    if (validate) {
+      boolean hasDiff = false;
+
+      for (String category : CATEGORIES) {
```
Xenorith, line 126: can this part be refactored and not copied; looks similar to the validate section for the configuration docs

Xenorith (changes requested):

shell/src/main/java/alluxio/cli/DocGenerator.java
```diff
@@ -35,23 +35,27 @@
...
-      new Options().addOption(METRIC_OPTION).addOption(CONF_OPTION);
+      new Options().addOption(METRIC_OPTION).addOption(CONF_OPTION).addOption(VALIDATE_OPTION);
 
   /**
    * Main entry for this util class.
    *
    * @param args arguments for command line
    */
   public static void main(String[] args) throws IOException {
+    boolean validate = false;
```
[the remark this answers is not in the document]
Limuzhou: cmd is created below in the try block. I feel like having this as the variable place holder and change it with other flags after cmd is defined is alright too?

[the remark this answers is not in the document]
Xenorith: ah yes that makes sense, let's leave it as is

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -128,8 +145,18 @@ public static void writeCSVFile(Collection<? extends PropertyKey> defaultKeys,
     } finally {
       try {
         closer.close();
+        if (validate) {
+          compareFiles( CSV_FILE_NAMES,filePath);
+        }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing Property Key CSV FileWriter", e);
+        for (String fileName : CSV_FILE_NAMES) {
```
Xenorith, line 153: does this mean the temp files only get deleted if there's an error in closing the file? the temp files should be removed regardless of how close behaves so i think this block should be in the outer finally block

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -214,10 +231,44 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys,
     } finally {
       try {
         closer.close();
+        if (validate) {
+          compareFiles( YML_FILE_NAMES, filePath);
+        }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing YML files for description of Property Keys "
             + "FileWriter", e);
+        for (String fileName : CSV_FILE_NAMES) {
```
Xenorith, line 240: ditto for this

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -214,10 +231,44 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys,
     } finally {
       try {
         closer.close();
+        if (validate) {
+          compareFiles( YML_FILE_NAMES, filePath);
```
Xenorith, line 235: extra space before `YML_FILE_NAMES`

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
Xenorith: oh now i see the delete file call in the compare files function. i think it's cleaner to have a single delete call since we can have both scenarios, with or without throwing an error, traverse through the finally block. otherwise it's easy to fall into the assumption i made where i did not think the deletion was handled properly

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -214,10 +231,44 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys,
...
+      if (! FileUtils.contentEquals(new File(filePath, fileName),
+          new File(filePath, fileNameTemp))) {
+        hasDiff = true;
+        System.out.println("Config file " + fileName + " changed.");
       }
+      File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
+      tempFile.delete();
+    }
+    if (!hasDiff) {
+      System.out.println("No change in config CSV file detected.");
```
Xenorith, line 271: in config CSV file`s`

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -52,6 +51,22 @@
...
   private static final String YML_SUFFIX = "yml";
   private static final String CSV_FILE_HEADER = "metricName,metricType";
+  public static final String TEMP_PREFIX = "temp-";
+
+  /**
+   * Compare the temp files and the committed files if validate flag is set
+   *
+   */
+  public static boolean compareFiles(String folder, String suffix, String category, boolean hasDiff) throws IOException{
+    if (! FileUtils.contentEquals(new File(folder, (TEMP_PREFIX + category + "-metrics." + suffix)),
```
Xenorith, line 61: extra space after `!`

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -47,9 +46,36 @@
   private static final String CSV_FILE_DIR = "docs/_data/table/";
   private static final String YML_FILE_DIR = "docs/_data/table/en/";
   public static final String CSV_FILE_HEADER = "propertyName,defaultValue";
+  public static final String TEMP_PREFIX = "temp-";
```
Xenorith, line 49: private instead of public?
same for the identical variable in the MetricsDocGenerator

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -124,30 +139,8 @@ public static void generate(boolean validate) throws IOException {
...
-        output = IOUtils.toString(pb.start().getInputStream(), StandardCharsets.UTF_8);
-        if (output.length() > 0) {
-          hasDiff = true;
-          System.out.println("Metrics file " + category + "-metrics." + YML_SUFFIX + " changed.");
-        }
-        f = new File(ymlFolder, category + "-metrics." + YML_SUFFIX);
-        f2 = new File(ymlFolder, "temp-" + category + "-metrics." + YML_SUFFIX);
-        f2.renameTo(f);
+        hasDiff = compareFiles(csvFolder,CSV_SUFFIX,category,hasDiff);
+        hasDiff = compareFiles(ymlFolder,YML_SUFFIX,category,hasDiff);
```
Xenorith, line 143: the yml check overwrites the result of the csv check here? set different output booleans for each and join them

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -47,9 +46,36 @@
...
+
+  public static final String[] CSV_FILE_NAMES = {"user-configuration.csv", "master-configuration.csv",
```
Xenorith, line 51: nit: when listing a long series of strings (ex. 3 or more), i like to put them each on their own line and sort them alphabetically to be able to quickly scan for a specific file. most of the alluxio codebase does not follow this but i'm slowly making this a pattern whenever i see the opportunity

Xenorith (changes requested):

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -44,33 +46,67 @@
...
+   * @param validate the validate flag of the command
+   * @return a FileWriter associated with the file
+   */
+  public static FileWriter createFileWriter(String filePath , String fileName , boolean validate)
+      throws IOException {
+    if (validate) {
+      String fileNameTemp = TEMP_PREFIX.concat(fileName);
+      return new FileWriter(PathUtils.concatPath(filePath, fileNameTemp));
+    } else {
+      return new FileWriter((PathUtils.concatPath(filePath, fileName)));
```
Xenorith, line 85: extra set of parentheses around PathUtils.concatPath()

tests/src/test/java/alluxio/client/cli/fs/ConfigurationDocGeneratorTest.java
```diff
@@ -141,7 +141,7 @@ public void checkYMLFile() throws Exception {
     String description = pKey.getDescription();
     defaultKeys.add(pKey);
 
-    ConfigurationDocGenerator.writeYMLFile(defaultKeys, mLocation);
+    ConfigurationDocGenerator.handleYMLFile(defaultKeys, mLocation, true);
```
Xenorith, line 144: add a new parameter for this validate parameter; should be testing both the true and false cases

shell/src/main/java/alluxio/cli/DocGenerator.java
```diff
@@ -61,15 +66,22 @@ public static void main(String[] args) throws IOException {
         printHelp("Unable to parse input args: " + e.getMessage());
         return;
       }
+      if (cmd.hasOption(VALIDATE_OPTION_NAME)) {
+        validate = true;
+      }
+      if ((!cmd.hasOption(METRIC_OPTION_NAME)) && (!cmd.hasOption(CONF_OPTION_NAME))) {
```
Xenorith, line 72: nit: unnecessary parenthesis surrounding the two parts?
`if (!cmd.hasOption(METRIC_OPTION_NAME) && !cmd.hasOption(CONF_OPTION_NAME)) {`

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -119,9 +155,19 @@ public static void writeCSVFile(Collection<? extends PropertyKey> defaultKeys, S
     } finally {
       try {
         closer.close();
+        if (validate) {
+          compareFiles(CSV_FILE_NAMES, filePath, "CSV");
```
Xenorith, line 159: similar to how MetricsDocGenerator defines constants for CSV and YML suffixes, replace "CSV" with a constant. in general, be wary about using strings within code, unless they have little significance (ex. whitespace, newlines)

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -119,9 +155,19 @@ public static void writeCSVFile(Collection<? extends PropertyKey> defaultKeys, S
...
+        }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing Property Key CSV FileWriter", e);
       }
+      for (String fileName : CSV_FILE_NAMES) {
+        String fileNameTemp = TEMP_PREFIX.concat(fileName);
+        File tempFile = new File(PathUtils.concatPath(filePath , fileNameTemp));
```
Xenorith, line 166: nit: remove space after `filePath`

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -119,9 +155,19 @@ public static void writeCSVFile(Collection<? extends PropertyKey> defaultKeys, S
     } finally {
       try {
         closer.close();
+        if (validate) {
+          compareFiles(CSV_FILE_NAMES, filePath, "CSV");
+        }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing Property Key CSV FileWriter", e);
       }
+      for (String fileName : CSV_FILE_NAMES) {
```
Xenorith, line 164: the following deletion operation only needs to occur if validate is true. if validate is false, the temp files are known not to exist.

same for the yml files iteration

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -119,9 +155,19 @@ public static void writeCSVFile(Collection<? extends PropertyKey> defaultKeys, S
     } finally {
       try {
         closer.close();
+        if (validate) {
+          compareFiles(CSV_FILE_NAMES, filePath, "CSV");
+        }
       } catch (IOException e) {
         LOG.error("Error while flushing/closing Property Key CSV FileWriter", e);
```
Xenorith, line 162: this log message is not accurate anymore after adding additional code in the try block, should be updated. same for the yml files section

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -197,26 +241,72 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+      if (!committedFile.exists()) {
+        System.out.println("Committed file does not exists");
+        continue;
+      }
+      if (!tempFile.exists()) {
+        System.out.println("Temporary generated file does not exists");
+        continue;
+      }
+      if (!FileUtils.contentEquals(committedFile,
+          tempFile)) {
```
Xenorith, line 285: this line doesn't need to be separated into a new line?

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -197,26 +241,72 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+  public static void compareFiles(String[] fileNames, String filePath, String fileType)
+      throws IOException {
+    boolean hasDiff = false;
+
+    for (String fileName : fileNames) {
+      String fileNameTemp = TEMP_PREFIX.concat(fileName);
+      File committedFile = new File(filePath, fileName);
+      File tempFile = new File(filePath, fileNameTemp);
+      if (!committedFile.exists()) {
+        System.out.println("Committed file does not exists");
```
Xenorith, line 277: when either file doesn't exist, something is probably very very wrong. currently the function continues to the next file as if nothing had happened, but i think this should throw an exception and fail the entire validate operation

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -44,33 +46,67 @@
...
 
+  /**
+   * Create the FileWriter object based on if validate flag is set to true.
+   *
+   * @param filePath path for the csv/yml file
+   * @param fileName name of the csv/yml file
+   * @param validate the validate flag of the command
+   * @return a FileWriter associated with the file
+   */
+  public static FileWriter createFileWriter(String filePath , String fileName , boolean validate)
```
Xenorith, line 79: nit: remove extra spaces before the commas

shell/src/main/java/alluxio/cli/docgen/ConfigurationDocGenerator.java
```diff
@@ -197,26 +241,72 @@ public static void writeYMLFile(Collection<? extends PropertyKey> defaultKeys, S
...
+        continue;
+      }
+      if (!tempFile.exists()) {
+        System.out.println("Temporary generated file does not exists");
+        continue;
+      }
+      if (!FileUtils.contentEquals(committedFile,
+          tempFile)) {
+        hasDiff = true;
+        System.out.println("Config file " + fileName + " changed.");
```
Xenorith, line 287: use printf instead of println to take advantage of string substitution. add `%n` to append a newline at the end to mimic what println would have done
`System.out.println("Config file %s changed.%n", fileName);`

and same for all other `System.out.println`s that concats a string variable value

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -42,18 +44,48 @@
...
+   * @param suffix the file types (CSV/YML)
+   * @param category the category in the file names
+   * @param hasDiff a boolean indicating has there been changes
+   *
+   * @return a boolean indicating if there is anything difference
+   * between the temp files and committed files
+   */
+  public static boolean compareFiles(String folder, String suffix, String category,
+                                     boolean hasDiff) throws IOException {
+    if (!FileUtils.contentEquals(new File(folder, (TEMP_PREFIX + category + "-metrics." + suffix)),
```
Xenorith, line 74: use String.format() to join string variables and string constants. ex. replace `(TEMP_PREFIX + category + "-metrics." + suffix)` with `String.format("%s%s-metrics.%s", TEMP_PREFIX, category, suffix)`

and same for other strings being joined with `+`

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -42,18 +44,48 @@
...
+        new File(folder, (category + "-metrics." + suffix)))) {
+      hasDiff = true;
+      System.out.println("Metrics file " + category + "-metrics." + suffix + " changed.");
+    }
+    File tempFile = new File(folder, TEMP_PREFIX + category + "-metrics." + suffix);
+    tempFile.delete();
```
Xenorith, line 80: compareFiles shouldn't be responsible for deleting the temp file...

okay i see that the logic that calls the metrics version of compareFiles doesn't have the same try/finally block as in the configuration version. the try/finally is actually critical to proper control flow. if compareFile throws an exception, such as in the case where one of the files doesn't exist and `FileUtils.contentEquals` throws an IOException, it will leave all the temporary files undeleted

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -42,18 +44,48 @@
...
+   *
+   * @param folder the location of the files to be compared
+   * @param suffix the file types (CSV/YML)
+   * @param category the category in the file names
+   * @param hasDiff a boolean indicating has there been changes
+   *
+   * @return a boolean indicating if there is anything difference
+   * between the temp files and committed files
+   */
+  public static boolean compareFiles(String folder, String suffix, String category,
```
Xenorith, line 72: i would've expected this compareFiles to be very similar to the compareFiles in MetricsDocGenerator.java. i'm a bit confused why they are so different; can they be implemented similarly? being able to consolidate similar logic is important; for example, the problem below about deleting the temp file would not have been made if there was only one version of compareFiles

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -110,6 +148,17 @@ public static void generate() throws IOException {
             key, StringEscapeUtils.escapeHtml4(metricKey.getDescription().replace("'", "''"))));
       }
     }
+    if (validate) {
+      boolean hasDiffCSV = false;
+      boolean hasDiffYML = false;
+      for (String category : CATEGORIES) {
+        hasDiffCSV = compareFiles(csvFolder, CSV_SUFFIX, category, hasDiffCSV);
```
Xenorith, line 155: rather than passing the boolean variable into the function, i find it more intuitive to do the boolean logic outside the function:
`hasDiffCSV = compareFiles(csvFolder, CSV_SUFFIX, category, hasDiffCSV) || hasDiffCSV;`
note that the order is important; i want to avoid short circuting because i always want to run compareFiles to print the file names of nonmatching files if hasDiffCSV has already been set to true

also if going along with this pattern, there is no longer a need for 2 separate boolean variables

shell/src/main/java/alluxio/cli/docgen/MetricsDocGenerator.java
```diff
@@ -110,6 +148,17 @@ public static void generate() throws IOException {
...
+        hasDiffYML = compareFiles(ymlFolder, YML_SUFFIX, category, hasDiffYML);
+      }
+      if (!hasDiffCSV && !hasDiffYML) {
+        System.out.println("No change in metric file detected.");
```
Xenorith, line 159: metric files

Xenorith (changes requested):
the more i read this code, the more i realized this section of code is messy and unorganized. as you've been experiencing, refactoring this code to introduce more logic is now a huge pain. there are similar operations across the two generators, each with different file types, and instead of properly funneling the logic to go through the same set of shared operations, there is lots of copy/paste code. this mess forces the next person to make a larger mess when introducing new logic and it'll only continue to compound and become even harder for the next person that comes across this code

LuQQiu:
@Xenorith Any updates for this PR?

Xenorith:
this PR is being put on hold for now, no ETA

PR #13306, opened 2021-04-27
Enable HTTPS when SSE is enabled on S3

ns1123:
When enabling SSE, PUT/GET requests to S3 must use HTTPS or the response will be a 403. Instead of having the user specify the need for HTTPS manually, we should enable it by default whenever SSE is specified.


bf8086 (approved):

bf8086:

PR #13307, opened 2021-04-27
Deflake stopAfterSecondaryTransition test

bradyoo:


ZacBlanco (changes requested):
Can we _not_ ignore this test since it is the only test providing critical coverage for the stop on demotion flag? Please try removing the `assertFalse` on lines 130/131 but leave the `assertFalse(m.isRunning())`. It should be more reliable since it won't be relying on checking the ports.

bradyoo:
@ZacBlanco Okay. Will test with that.

tieujason330:
i removed myself from review since i don't know much about this, but thanks for making this change for the master builds

ZacBlanco (approved):

gpang (approved):
LGTM

bradyoo:

Issue #13309, opened 2021-04-28
Cephfs mvn build doc command error out


LuQQiu:
**Alluxio Version:**
Alluxio master 2.6.0-SNAPSHOT-c6582b0cdcb83e41209cf00b809690588434a106

**Describe the bug**
Doc generation failed after PR https://github.com/Alluxio/alluxio/commit/3ef4728df196c17e4eb397f6612e41b68fe26791 merged. 
After removing the whole underfs/cephfs, the `mvn javadoc:aggregate` command passed.

**To Reproduce**
In linux environment, git clone https://github.com/alluxio/alluxio
git init
git fetch origin
git checkout origin/master
mvn clean install --no-transfer-progress -T 2C -DskipTests -Dmaven.javadoc.skip -Dcheckstyle.skip -Dlicense.skip -Dfindbugs.skip
mvn javadoc:aggregate --no-transfer-progress -T 2C -Dcheckstyle.skip -Dlicense.skip -Dfindbugs.skip

the javadoc:aggregate command error out
```
[INFO] Forking Alluxio Shaded Libraries - cephfs-hadoop 2.6.0-SNAPSHOT
[WARNING] Invalid POM for com.ceph.fs:cephfs-hadoop:jar:0.80.6, transitive dependencies (if any) will not be available, enable debug logging for more details
[INFO] Forking Alluxio Under File System - CephFS 2.6.0-SNAPSHOT
02:24:43  [WARNING] The POM for com.ceph:libcephfs:jar:0.80.5 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
[INFO] Forking Alluxio UI 2.6.0-SNAPSHOT
[WARNING] Invalid POM for com.ceph.fs:cephfs-hadoop:jar:0.80.6, transitive dependencies (if any) will not be available, enable debug logging for more details

02:27:14 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.9:aggregate (default-cli) on project alluxio-parent: Execution default-cli of goal org.apache.maven.plugins:maven-javadoc-plugin:2.9:aggregate failed: For artifact {com.ceph:libcephfs:null:jar}: The version cannot be empty. -> [Help 1]
```

**Expected behavior**
Expect the cephfs not affecting the docs publish mvn build.

**Urgency**
High, it blocks the alluxio doc generation

**Additional context**

LuQQiu:
[...]

PR #13310, opened 2021-04-28
Remove ceph module from underfs temporarily

LuQQiu:
Remove ceph module from underfs pom so that it will be skipped for maven build. Temporary workaround for #13309


LuQQiu:
@maobaolong This is a temporary workaround. After the issue solved, will add the module back

apc999 (approved):

apc999:
@maobaolong FYI

apc999:

PR #13315, opened 2021-04-29
[SMALLFIX] Change alluxio.logger.type property to hidden

jiacheliu3:
This property is only used in tests. It also should not be set explicitly by the users. It's confusing to display the property on the doc page.
Screen Shot 2021-04-29 at 4 53 54 PM

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -333,7 +333,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey LOGGER_TYPE =
       new Builder(Name.LOGGER_TYPE)
           .setDefaultValue("Console")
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -333,7 +333,8 @@ public String toString() {
   public static final PropertyKey LOGGER_TYPE =
       new Builder(Name.LOGGER_TYPE)
           .setDefaultValue("Console")
-          .setDescription("The type of logger.")
+          .setDescription("This controls which logger the process uses. This is only set by code.")
```
gpang, line 336: Do you want to say "test code"?

jiacheliu3: updated

gpang (approved):
LGTM

jiacheliu3:

PR #13319, opened 2021-04-29
[SMALLFIX] Fix NPE in Allocator and add some UT for Reviewer

yabola:
1. Allocator may  allocate null dirView ,`dirView.getAvailableBytes()`  is NPE and dirView has no toString implementation.
2. Add some UT for `Reviewer`

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
@@ -629,10 +629,8 @@ public class TieredBlockStore implements BlockStore {
           // Block expansion are forcing the location. We do not want the review's opinion.
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               options.getLocation(), allocatorView.refreshView(), true);
-          if (LOG.isDebugEnabled()) {
-            LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
-                dirView.getAvailableBytes());
-          }
+          LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
+                  dirView == null ? 0 : dirView.getAvailableBytes());
           if (dirView == null) {
             LOG.error("Target tier: {} has no evictable space to store {} bytes for session: {}",
                 options.getLocation(), options.getSize(), sessionId);
@@ -671,7 +669,8 @@ public class TieredBlockStore implements BlockStore {
           // Skip the review as we want the allocation to be in the place we just freed
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               BlockStoreLocation.anyTier(), allocatorView.refreshView(), true);
-          LOG.debug("Allocation after freeing space for block creation: {}", dirView);
+          LOG.debug("Allocation after freeing space for block creation. availableBytes: {}",
+                  dirView == null ? 0 : dirView.getAvailableBytes());
         }
       }
       if (dirView == null) {
--- a/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
@@ -629,8 +629,10 @@ public class TieredBlockStore implements BlockStore {
           // Block expansion are forcing the location. We do not want the review's opinion.
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               options.getLocation(), allocatorView.refreshView(), true);
-          LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
-                  dirView == null ? 0 : dirView.getAvailableBytes());
+          if (LOG.isDebugEnabled()) {
+            LOG.debug("Allocation after freeing space for block expansion, {}", dirView == null
+                ? "no available dir." : "available bytes in dir: " + dirView.getAvailableBytes());
+          }
           if (dirView == null) {
             LOG.error("Target tier: {} has no evictable space to store {} bytes for session: {}",
                 options.getLocation(), options.getSize(), sessionId);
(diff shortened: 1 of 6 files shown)


yabola:
@jiacheliu3 please take a look,thanks~

yabola:
@jiacheliu3 please take a look~

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -629,10 +629,8 @@ private StorageDirView allocateSpace(long sessionId, AllocateOptions options)
           // Block expansion are forcing the location. We do not want the review's opinion.
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               options.getLocation(), allocatorView.refreshView(), true);
-          if (LOG.isDebugEnabled()) {
-            LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
-                dirView.getAvailableBytes());
-          }
+          LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
+                  dirView == null ? 0 : dirView.getAvailableBytes());
```
jiacheliu3, line 633: nit: reduce the indent by 2 spaces to be aligned to previous lines

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -671,7 +669,8 @@ private StorageDirView allocateSpace(long sessionId, AllocateOptions options)
           // Skip the review as we want the allocation to be in the place we just freed
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               BlockStoreLocation.anyTier(), allocatorView.refreshView(), true);
-          LOG.debug("Allocation after freeing space for block creation: {}", dirView);
+          LOG.debug("Allocation after freeing space for block creation. availableBytes: {}",
+                  dirView == null ? 0 : dirView.getAvailableBytes());
```
jiacheliu3, line 673: nit: reduce the indent by 2 spaces to be aligned to previous lines

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -629,10 +629,8 @@ private StorageDirView allocateSpace(long sessionId, AllocateOptions options)
           // Block expansion are forcing the location. We do not want the review's opinion.
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               options.getLocation(), allocatorView.refreshView(), true);
-          if (LOG.isDebugEnabled()) {
-            LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
-                dirView.getAvailableBytes());
-          }
+          LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
```
jiacheliu3, line 632: I don't think you should print 0 when this is null. As far as i remember the `dirView` being null means there's no available allocation or the allocation is rejected. The dir is not necessarily 0 byte availability. Therefore printing 0 here may be confusing.

I will suggest we do a null check and just `LOG.debug` a message saying there's no available allocation found if it's null. When it's not null we just print the same message.

jiacheliu3 (changes requested):
Thanks @yabola for the improvement. The fix and tests make sense to me. I like how you mocked the Reviewer class. If you resolve my comment on the logging, this will look good to me.

One comment is I'm not sure about rejecting when the available bytes is in the Set, it looks a bit strange...But yea I don't have a better idea so I'm fine with that.

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
yabola: done

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
yabola: done

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
yabola: Thanks for review!
How do you feel the way now? I found there are lots of  `if` conditions here. so i try not to introduce introduce another `if` statement.

jiacheliu3 (approved):
@yabola Thanks for the improvement! LGTM

yabola:
@jiacheliu3 sorry, my mistake, lost `{}` in log. please review again and help merge this PR, thanks.

jiacheliu3:
/ping @yuzhu I'll merge if this looks good to you :)

LuQQiu:
@yuzhu PTAL this PR, thanks!

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -629,10 +629,8 @@ private StorageDirView allocateSpace(long sessionId, AllocateOptions options)
           // Block expansion are forcing the location. We do not want the review's opinion.
           dirView = mAllocator.allocateBlockWithView(sessionId, options.getSize(),
               options.getLocation(), allocatorView.refreshView(), true);
-          if (LOG.isDebugEnabled()) {
-            LOG.debug("Allocation after freeing space for block expansion. availableBytes: {}",
-                dirView.getAvailableBytes());
-          }
+          LOG.debug("Allocation after freeing space for block expansion, {}", dirView == null
```
yuzhu, line 632: there is actually a reason to do this check before the debug statement. because it saves the cost of computing parameters to the debug statement. in this case `dirView.getAvailableBytes` is not called.

yuzhu (changes requested):
I don't actually have the proper context to review this PR, just a minor comment about the debug logging statement.

LuQQiu:
@jiacheliu3 do you feel comfortable merging this PR? since both @yuzhu and I don't have much context to review this PR

jiacheliu3:
> @jiacheliu3 do you feel comfortable merging this PR? since both @yuzhu and I don't have much context to review this PR
> or who do you think would be better to review this PR?

@LuQQiu I'm good with merging after the comment from @yuzhu is resolved. @yabola could you update the logging as suggested? thanks

yabola:
thanks for advice. I have updated~

yuzhu (approved):

yuzhu:

PR #13321, opened 2021-04-29
Make connection pools per connection instead of per db for hive udb

### LINKED ISSUE #13227: HiveClientPool causes the number of hive metastore connections to be too high
**Alluxio Version:**
2.5.0-1

**Describe the bug**
HiveClientPool causes the number of hive metastore connections to be too high, when there are many databases to synchronize

**To Reproduce**
We have more than 100 databases. Alluxio connected to hive metastore and generated more than 400 connections.

**Expected behavior**
There should not be so many connections. 

**Urgency**
MEDIUM

**Additional context**
The alluxio source code uses a database-level connection pool. In fact, it should use a hive metastore url connection pool.

yuzhu:
fixes #13227

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -5167,7 +5167,20 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-
+  public static final PropertyKey TABLE_UDB_HIVE_CLIENTPOOL_MIN =
+      new Builder(Name.TABLE_UDB_HIVE_CLIENTPOOL_MIN)
+          .setDefaultValue("16")
+          .setDescription("The minimum capacity of the hive client pool per hive metastore")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey TABLE_UDB_HIVE_CLIENTPOOL_MAX =
+      new Builder(Name.TABLE_UDB_HIVE_CLIENTPOOL_MAX)
+          .setDefaultValue("256")
+          .setDescription("The maximum capacity of the hive client pool per hive metastore")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
   /**
    * @deprecated This key is used for testing. It is always deprecated.
    */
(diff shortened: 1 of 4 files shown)


table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HiveClientPool.java
```diff
@@ -79,20 +73,7 @@ protected IMetaStoreClient createNewResource() throws IOException {
 
       IMetaStoreClient client = HMSClientFactory.newInstance(
           RetryingMetaStoreClient.getProxy(conf, NOOP_HOOK, HiveMetaStoreClient.class.getName()));
```
yuzhu, line 75: I think it works just as well to let other exception handling handle db not exist errors than explicitly checking for it. 
@gpang what do you think?

gpang: I thought maybe there was a reasons to explicitly check for the db, but I forgot why...

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/util/HiveClientPool.java
```diff
@@ -41,24 +40,19 @@
...
    * Creates a new hive client client pool.
    *
    * @param connectionUri the connect uri for the hive metastore
-   * @param hiveDbName the db name in hive
    */
-  public HiveClientPool(String connectionUri, String hiveDbName) {
+  public HiveClientPool(String connectionUri) {
     super(Options.defaultOptions()
         .setMinCapacity(16)
         .setMaxCapacity(128)
```
gpang, line 52: If we are re-using pools to the same hive db, we may need to change this to a configuration parameter?

gpang (commented):
@yuzhu

yuzhu:
@gpang  PTAL

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -5167,7 +5167,20 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-
+  public static final PropertyKey TABLE_UDB_HIVE_CLIENTPOOL_MIN =
+      new Builder(Name.TABLE_UDB_HIVE_CLIENTPOOL_MIN)
+          .setDefaultValue("16")
+          .setDescription("The minimum capacity of the hive client pool per hive metastore")
```
gpang, line 5173: If these pools are per UDB, then should they be UDB properties instead? (meaning, each `attach`ed db can have a different sized pool.)

gpang (approved):
@yuzhu Had a minor question.

LGTM

LuQQiu:
@yuzhu Please help update the status of this PR? is it ready to merge

core/common/src/main/java/alluxio/conf/PropertyKey.java
yuzhu: took another closer look, these pools are per connection, not per UDB. since we don't have properties associated with connectionUris, i am inclined to keep them as global property keys for now. we can address this in a separate PR if needed.

yuzhu:

PR #13326, opened 2021-04-30
Make runUfsIOTest run in a subdir

jiacheliu3:
The motivations is sometimes when the user runs this command, they expect alluxio to create a child dir in the `--path` so they just give a parent dir.


stress/shell/src/main/java/alluxio/stress/cli/UfsIOBench.java
```diff
@@ -43,6 +45,7 @@
 public class UfsIOBench extends Benchmark<IOTaskResult> {
   private static final Logger LOG = LoggerFactory.getLogger(UfsIOBench.class);
   private static final int BUFFER_SIZE = 1024 * 1024;
+  private static final String TEST_DIR_NAME = "iotest";
```
JySongWithZhangCe, line 48: Maybe the `TEST_DIR_NAME` should be a special value ? `"iotest"` will be already exists.

yuzhu (approved):
LGTM thanks

jiacheliu3:

PR #13334, opened 2021-05-04
Cancel sync job in middle of ufs calls if client cancelled

bradyoo:
The previous version slowed down the sync because the number of prefetch threads were much smaller than the number of rpc threads that were performing a sync in many cases.


yuzhu:
@bradyoo is the only difference from the prev version the change in default thread count?

bradyoo:
@yuzhu That is correct though I can't figure out why the test is failing so I'm reverting that bit back for a moment to see if the tests succeed.

bradyoo:
had to reduce the number for the tests to pass. I think the way the test is passing right now (and the way it not have been passing before may not be the way we think it is... The test is too timing based to debug properly.)

yuzhu:
which test was it?

bradyoo:
repeatedDirectorySync and repeatedChildrenSync

yuzhu:
LGTM

yuzhu (approved):

bradyoo:

PR #13340, opened 2021-05-05
Attempt to fix RocksDB SegFaulting Occasionally

bradyoo:
RocksDB engineer claims that it is possible for the readOptions to get garbage collected during iteration because RocksIterator doesn't hold a reference to the object but instead a native resource owned by readOptions.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
+++ b/core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
@@ -151,8 +151,8 @@ public class RocksBlockStore implements BlockStore {
     byte[] startKey = RocksUtils.toByteArray(id, 0);
     byte[] endKey = RocksUtils.toByteArray(id, Long.MAX_VALUE);
 
-    try (RocksIterator iter = db().newIterator(mBlockLocationsColumn.get(),
-        new ReadOptions().setIterateUpperBound(new Slice(endKey)))) {
+    final ReadOptions readOptions = new ReadOptions().setIterateUpperBound(new Slice(endKey));
+    try (RocksIterator iter = db().newIterator(mBlockLocationsColumn.get(), readOptions)) {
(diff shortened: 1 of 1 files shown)


apc999 (approved):

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -151,8 +151,8 @@ public void close() {
     byte[] startKey = RocksUtils.toByteArray(id, 0);
     byte[] endKey = RocksUtils.toByteArray(id, Long.MAX_VALUE);
 
-    try (RocksIterator iter = db().newIterator(mBlockLocationsColumn.get(),
-        new ReadOptions().setIterateUpperBound(new Slice(endKey)))) {
+    final ReadOptions readOptions = new ReadOptions().setIterateUpperBound(new Slice(endKey));
+    try (RocksIterator iter = db().newIterator(mBlockLocationsColumn.get(), readOptions)) {
```
ggezer, line 155: How does this change prevent segsiv? or else why does the original cause one?

bradyoo: Read the description.

bradyoo:

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
bradyoo: The claim is that because there is no hard reference to ReadOptions in Java. The garbage collector may collect the object away even though the RocksDB native code still needs it.

apc999: @bradyoo  can you add some javadoc before line 154 to prevent regression by accident in the future ?

bradyoo:

PR #13345, opened 2021-05-06
[Helm] Add Master StatefulSet podManagementPolicy Parallel

cdmikechen:
Sometimes two masters will run on the same node. So that we can add `podantiaffinity` to avoid this situation.

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/Chart.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/Chart.yaml
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.6.20
+version: 0.6.21
 home: https://www.alluxio.io/
 maintainers:
 - name: Adit Madan
--- a/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
@@ -64,6 +64,15 @@ spec:
       {{- include "alluxio.hostAliases" . | nindent 6 }}
       {{ end }}
       hostNetwork: {{ $hostNetwork }}
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - topologyKey: "kubernetes.io/hostname"
+              labelSelector:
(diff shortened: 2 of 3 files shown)


maobaolong:
@jiacheliu3 Would you like to take a look at this PR?

integration/kubernetes/helm-chart/alluxio/Chart.yaml
```diff
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.6.20
+version: 0.6.21
```
jiacheliu3, line 15: Please rebase on the latest `master` as the version has gone up after a few changes, thanks!

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
```diff
@@ -64,6 +64,15 @@ spec:
       {{- include "alluxio.hostAliases" . | nindent 6 }}
       {{ end }}
       hostNetwork: {{ $hostNetwork }}
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
jiacheliu3, line 69: I prefer the "softer" version `preferredDuringSchedulingIgnoredDuringExecution`. I know that some users do not mind running multiple master pods on the same node (for example, 2 nodes for 3 master pods). This deployment makes sense for test and dev environments. 

Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

jiacheliu3 (changes requested):
@cdmikechen Sorry for the late review. I missed the last notification. PTAL at my comments, thanks!

integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
cdmikechen: @jiacheliu3 
I think I can move `affinity` to values.yaml, so that users  can decide whether to use k8s's affinity. Many helm charts come with this optional configuration.

jiacheliu3: Sure

PR #13348, opened 2021-05-06
Fix process local block in stream

LuQQiu:
When data locate in other nodes, use the GrpcBlockingStream to read from other nodes instead of reading from the local node and refetching data from UFS.


apc999 (approved):

LuQQiu:

PR #13355, opened 2021-05-06
Check consistency should check file length

yuzhu:


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1354,7 +1354,9 @@ private boolean checkConsistencyInternal(LockedInodePath inodePath) throws Inval
       } else {
         String ufsFingerprint = Fingerprint.create(ufs.getUnderFSType(), ufsStatus).serialize();
         return ufsStatus.isFile()
-            && (ufsFingerprint.equals(inode.asFile().getUfsFingerprint()));
+            && (ufsFingerprint.equals(inode.asFile().getUfsFingerprint()))
+            && ufsStatus instanceof UfsFileStatus
```
ggezer, line 1358: Is this required with `ufsStatus.isFile()` above in the statement?

ggezer: Can it be another class? And would it invalidate this check if so?

yuzhu: it should not be, unless we have a bug. If it happens to be another class, then we should report the file as inconsistent anyways, because we want the user to know something is wrong

ggezer (approved):

yuzhu:

yuzhu:

PR #13363, opened 2021-05-07
Supply Alluxio csi implementation

maobaolong:


maobaolong:
@Binyang2014 Thanks for your reply. 

I decided to do this effort by the following reason

- After some research, I found two great implementations wrote by java, and these two open source project are all make csi as a module of them, few works to do just following and back port it to alluxio repo can satisfy our request. We can reference the yarn and ozone implementations, in fact we are using ozoen csi on some of our customers.
  - [Apache hadoop yarn csi](https://github.com/apache/hadoop/tree/trunk/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-csi/src/main)
  - [Apache Ozone csi](https://github.com/apache/ozone/tree/master/hadoop-ozone/csi/src/main)
- Running on JVM is easy to debug this csi driver remotely, it can dependencies modules of Alluxio directly, something like `alluxio-core-client-fs` and `alluxio-core-common`, we can define the configure items in the `alluxio.properties` like the alluxio master worker and fuse. 
- I tried to using `https://github.com/alluxio/alluxio-csi` and hope it can satisfy our request, after some effort, we found that this is a part implementation of CSI, it cannot helps us to supply a volume dynamically, so we have to turn to cost our resource to implement it. Before this work started, I have a discussion with @apc999 , he have no objection idea about this.
- ` if some CSI spec update, we need to reimplement them and can not just update dependencies`, given that we just maintain the `csi.proto` other than the generated code, so it's easy to update `csi.proto`.
- A separate repository can make `alluxio-csi` independent, but lack of attention, now I propose to bring alluxio-csi into alluxio as a module of alluxio, can let more contributors join us to maintain the alluxio-csi.

Some following-up works
- Alluxio csi yaml files preview
[alluxio-csi-yaml.zip](https://github.com/Alluxio/alluxio/files/6456991/alluxio-csi-yaml.zip)
- Mount alluxio ufs when publish a csi volume

Binyang2014:
BTW, I did some test and find current go implementation support dynamic provisioning, here is a draft PR you can refer: https://github.com/Alluxio/alluxio-csi/pull/6

maobaolong:
@Binyang2014 @apc999 As we have a discussion on line for the alluxio csi, I decided to close this PR after
- Move alluxio-csi as a module of alluxio repository
- Implements the dynamic provisioning and generate csi related yaml files and examples when using `helm-generate.sh`
- Using an all-in-one image in whole alluxio.

Binyang2014:
Link draft PR here: https://github.com/Alluxio/alluxio/pull/13435

Issue #11197, opened 2020-03-20
Prompt "Cannot connect to master [Address]" when client failed to connect to Master


apc999:
**Is your feature request related to a problem? Please describe.**

```console
$ ./bin/alluxio fs mount /mnt/s3 s3://analytics-zoo-data
Failed to connect to FileSystemMasterClient @ localhost:19998 after 8 attempts
```
The error message is not very clear what happened without hint for users further investigate the root cause.

**Describe the solution you'd like**

Something like:

```console
$ ./bin/alluxio fs mount /mnt/s3 s3://analytics-zoo-data
Failed to connect to Alluxio master (localhost:19998) after 8 attempts
Please check if you Alluxio master is currently running on server "localhost".
```

We may even provide more detailed instructions like "log into master running "jps" to see if process "AlluxioMaster" is running, or check master log at ${ALLUXIO_HOME}/logs/master.log"

**Describe alternatives you've considered**
NA

**Urgency**

**Additional context**

githezhihao:
@apc999 hi~,I had the same error here when  i use presto to execute a query through alluxio.
the presto server log is following:
`io.prestosql.spi.PrestoException: Error opening Hive split alluxio:/beh001/yarn/warehouse/alx_yarn.db/store_sales/000094_0 (offset=100663296, length=33554432): Failed to connect to FileSystemMasterClient @ bonc-bdc-11.bonc.com:19998 after 44 attempts
Caused by: alluxio.exception.status.UnavailableException: Failed to connect to FileSystemMasterClient @ bonc-bdc-11.bonc.com:19998 after 44 attempts
Caused by: alluxio.exception.status.UnavailableException: Failed to handshake with master bonc-bdc-11.bonc.com:19998 to load cluster default configuration values: UNAVAILABLE: io exception
Caused by: alluxio.shaded.client.io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
Caused by: alluxio.shaded.client.io.netty.channel.AbstractChannel.AnnotatedConnectException: finishConnect(..) failed: Connection refused: bonc-bdc-11.bonc.com/192.168.27.220:19998
Caused by: java.net.ConnectException: finishConnect(..) failed: Connection refused`

here is my alluxio configuration :
`
#worker配置
alluxio.worker.ramdisk.size=25GB
alluxio.worker.tieredstore.levels=1
alluxio.worker.tieredstore.level0.alias=MEM
[...]

PR #13370, opened 2021-05-07
Improve logging message on exceptions

apc999:
Improve logging messages to be consistent with our documentation: 
https://docs.alluxio.io/os/user/edge/en/contributor/Code-Conventions.html#log-level-guidelines

Essentially, we 
1. prefer `e.toString()` to `e.getMessage()` in warn logging
2. prefer leaving exception `e` as the last argument for `LOG.error` to print the stack trace

Also fix some incorrect usage of logging,

- Fix `logger.warn("Exit(stream) (Error): {}: {}, Error={}", methodName, String.format(description, args), e);`
which will give output of:

```
2021-05-07 20:06:31,676 WARN  ShortCircuitBlockReadHandler - Exit(stream) (Error): OpenBlock: Session=0, Request=null, Error={}
alluxio.exception.BlockDoesNotExistException: blockId 5553258511 not found
	at alluxio.worker.grpc.ShortCircuitBlockReadHandler$1.call(ShortCircuitBlockReadHandler.java:90)
	at alluxio.worker.grpc.ShortCircuitBlockReadHandler$1.call(ShortCircuitBlockReadHandler.java:69)
	at alluxio.RpcUtils.streamingRPCAndLog(RpcUtils.java:163)
	at alluxio.worker.grpc.ShortCircuitBlockReadHandler.onNext(ShortCircuitBlockReadHandler.java:69)
	at alluxio.worker.grpc.ShortCircuitBlockReadHandler.onNext(ShortCircuitBlockReadHandler.java:36)
	at io.grpc.stub.ServerCalls$StreamingServerCallHandler$StreamingServerCallListener.onMessage(ServerCalls.java:255)
...
```

- Fix https://github.com/Alluxio/alluxio/issues/11197

### DIFF
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -275,9 +275,10 @@ public abstract class AbstractClient implements Client {
           new ServiceNotFoundException(lastConnectFailure.getMessage(), lastConnectFailure));
     }
 
-    throw new UnavailableException(String.format("Failed to connect to master (%s) after %s "
-            + "attempts of %s",
-        mAddress, retryPolicy.getAttemptCount(), getServiceName()), lastConnectFailure);
+    throw new UnavailableException(String.format(
+        "Failed to connect to master (%s) after %s attempts." +
+        "Please check if Alluxio master is currently running on \"%s\". Service=\"%s\"",
+        mAddress, retryPolicy.getAttemptCount(), mAddress, getServiceName()), lastConnectFailure);
   }
 
   /**
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -264,7 +264,7 @@ public abstract class AbstractClient implements Client {
               retryPolicy.getAttemptCount()));
     }
 
-    /**
+    /*
      * Throw as-is if {@link UnauthenticatedException} occurred.
      */
     if (lastConnectFailure instanceof UnauthenticatedException) {
@@ -275,10 +275,12 @@ public abstract class AbstractClient implements Client {
           new ServiceNotFoundException(lastConnectFailure.getMessage(), lastConnectFailure));
     }
 
-    throw new UnavailableException(String.format(
-        "Failed to connect to master (%s) after %s attempts." +
-        "Please check if Alluxio master is currently running on \"%s\". Service=\"%s\"",
-        mAddress, retryPolicy.getAttemptCount(), mAddress, getServiceName()), lastConnectFailure);
+    throw new UnavailableException(
+        String.format(
+            "Failed to connect to master (%s) after %s attempts."
+                + "Please check if Alluxio master is currently running on \"%s\". Service=\"%s\"",
+            mAddress, retryPolicy.getAttemptCount(), mAddress, getServiceName()),
+        lastConnectFailure);
   }
 
   /**
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -238,8 +238,8 @@ public abstract class AbstractClient implements Client {
             getServiceName(), mAddress);
         return;
       } catch (IOException e) {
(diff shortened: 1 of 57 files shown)


apc999:
@ggezer @bf8086  please take a look at this PR and ensure the logging style consistent with the logging convention in our documentation.

core/common/src/main/java/alluxio/AbstractClient.java
```diff
@@ -238,8 +238,8 @@ public synchronized void connect() throws AlluxioStatusException {
             getServiceName(), mAddress);
         return;
       } catch (IOException e) {
-        LOG.debug("Failed to connect ({}) with {} @ {}: {}", retryPolicy.getAttemptCount(),
-            getServiceName(), mAddress, e.getMessage());
+        LOG.debug("Failed to connect ({}) with {} @ {}", retryPolicy.getAttemptCount(),
+            getServiceName(), mAddress, e);
```
ggezer, line 242: I vaguely remember @bf8086 discovering this doesn't print the error as expected. Or was that something else?

shell/src/main/java/alluxio/cli/fs/command/AbstractFileSystemCommand.java
```diff
@@ -96,7 +96,7 @@ protected void runWildCardCmd(AlluxioURI wildCardPath, CommandLine cl) throws IO
         runPlainPath(path, cl);
       } catch (AlluxioException | IOException e) {
         LOG.error(String.format("error processing path: %s", path), e);
-        errorMessages.add(e.getMessage() != null ? e.getMessage() : e.toString());
+        errorMessages.add(e.toString());
```
ggezer, line 99: This might break cli tests.

core/common/src/main/java/alluxio/AbstractClient.java
apc999: I have an issue created on this error message: https://github.com/Alluxio/alluxio/issues/11197
It shows very confusing messages, like
```
$ ./bin/alluxio fs mount /mnt/s3 s3://analytics-zoo-data
Failed to connect to FileSystemMasterClient @ localhost:19998 after 8 attempts
```
Let me improve the message here too

shell/src/main/java/alluxio/cli/fs/command/AbstractFileSystemCommand.java
apc999: reverted

apc999:
@ggezer  thanks! Please take another look.

ggezer:
@apc999 do you mind fast-forwarding your commits on top of current master to avoid merging with master?

I'm aware we don't have any restrictions on that but I believe it's a better practice to avoid merging merge-commits unless it's coming from a big feature branch. That way we can keep git history lookups cleaner.

apc999:
> @apc999 do you mind fast-forwarding your commits on top of current master to avoid merging with master?
> 
> I'm aware we don't have any restrictions on that but I believe it's a better practice to avoid merging merge-commits unless it's coming from a big feature branch. That way we can keep git history lookups cleaner.

Fully agree with you. Rebased with upstream master. PTAL @ggezer

apc999:
@ggezer @bf8086  can you take a look? thanks

ggezer (approved):

ggezer:
@bf8086 I vaguely remember you mentioned either LOG.debug or LOG.error not printing error as expected when the last argument is a throwable. PTAL if these changes are safe against that.

apc999:
> vaguely remember you mentioned either LOG.debug or LOG.error not printing error as expected when the last argument is a throwable. PTAL if these changes are safe against that.

When the last argument is a throwable for `LOG.debug` or `LOG.error`, the stack trace will be printed.
Like `LOG.error("{}, {}", var1, var2, e)`, we have been using patterns like this in this code base extensively.

core/common/src/main/java/alluxio/AbstractClient.java
bf8086: IIRC the issue was more about we lose important information when reporting a server error. It was not related to the server log.

apc999:

Issue #13239, opened 2021-04-15
Properly umount fuse when killing the process


LuQQiu:
**Alluxio Version:**
Alluxio 2.6.0-SNAPSHOT

**Describe the bug**
In the PR #13160 , we throw `FuseException("Device is busy")` to help prevent the umount from happening when device is busy.
We expect the umount will be treated as an n-op and the client operations can continue running.

according to @Binyang2014 , if we use `umount <mount-point>` or `fusermount -u <mountpoint>` directly, `umount` will fail, alluxio logs contain error 
```
ERROR AbstractFuseFileSystem - Failed to umount /var/lib/kubelet/pods/.../mount
alluxio.jnifuse.FuseException: Unable to umount FS with exit code 1
```
and reading/writing data through the mount point can be continued.

However, our way to stop the fuse mount is killing the process of running the fuse application through `bin/alluxio-stop.sh worker` (for fuse on worker process) or `integration/fuse/bin/alluxio-fuse umount /mnt/alluxio-fuse`. In this way, the process will be killed, the fuse mount will be left in the operating system but not function anymore. Client operations will fail.

**To Reproduce**
- Set the fuse umount timeout to a relatively short period
- Starting the Fuse on Worker process or Standalone Fuse.
- Run a long-running read operation against the fuse mount point
- During the read, I alluxio-stop.sh worker or alluxio-fuse umount /mnt/fuse to stop the Fuse application.
- The results are similar. The worker/fuse process is killed. Fuse mount remains in the operating system. The client operations error out (failed to open/ failed to read). The fuse mount is inaccessible and nonfunctional anymore.

**Expected behavior**
TBD. 
Unable to kill worker/fuse process when fuse device is busy so that fuse mount can function as expected?

**Urgency**
Medium level urgency

**Additional context**

LuQQiu:
@apc999 please provide your investigation for this issue

PR #13373, opened 2021-05-08
Make umount fuse properly when killing the process

yabola:
It’s just a draft for fast review in #13239

### DIFF
--- a/bin/alluxio
+++ b/bin/alluxio
@@ -40,7 +40,7 @@ function printUsage {
   echo -e "                \t NOTE: This command requires a valid ufs path to run tests against."
   echo -e "  readJournal   \t Read an Alluxio journal file from stdin and write a human-readable version of it to stdout."
   echo -e "  upgradeJournal\t Upgrade an Alluxio journal version 0 (Alluxio version < 1.5.0) to an Alluxio journal version 1 (Alluxio version >= 1.5.0)."
-  echo -e "  killAll <WORD>\t Kill processes containing the WORD."
+  echo -e "  killAll [-s] <WORD>\t Kill processes containing the WORD (if '-s' specified, processes won't be forcibly killed even if operation is timeout)."
   echo -e "  copyDir <PATH>\t Copy the PATH to all master/worker nodes."
   echo -e "  clearCache    \t Clear OS buffer cache of the machine."
   echo -e "  docGen        \t Generate docs automatically."
@@ -79,31 +79,52 @@ EOF
 }
 
 function killAll {
-  if [[ $# -ne 1 ]]; then
-    echo "Usage: alluxio killAll <WORD>" >&2
+  if [[ $# -gt 2  ]]; then
+    echo "Usage: alluxio killAll [-s] <WORD> (if '-s' specified, processes won't be forcibly killed even if operation is timeout.)" >&2
     exit 2
   fi
-
-  keyword=$1
-  count=0
+  if [[ "$1" == "-s" ]]; then
+    keyword=$2
+    soft_kill_option=true
+  else
+    keyword=$1
+    soft_kill_option=false
+  fi
+  local success_count=0
+  local failed_count=0
   for pid in $(ps -Aww -o pid,command | grep -i "[j]ava" | grep ${keyword} | awk '{print $1}'); do
     kill -15 ${pid} > /dev/null 2>&1
     local maxWait=120
     local cnt=${maxWait}
+    fail_to_kill=false
     while kill -0 ${pid} > /dev/null 2>&1; do
       if [[ ${cnt} -gt 1 ]]; then
         # still not dead, wait
         cnt=$(expr ${cnt} - 1)
         sleep 1
+      elif [[ "${soft_kill_option}" = true ]] ; then
+        # waited long enough, give up killing the process
+        echo "Process [${pid}] did not complete after "${maxWait}" seconds, failing to kill it." >&2
+        fail_to_kill=true
+        break;
       else
-        # waited long enough, kill the process
-        echo "Process did not complete after "${maxWait}" seconds, killing." >&2
+        # waited long enough, forcibly kill the process
+        echo "Process [${pid}] did not complete after "${maxWait}" seconds, forcibly killing it." >&2
         kill -9 ${pid} 2> /dev/null
       fi
     done
-    count=$(expr ${count} + 1)
+    if [[ "$fail_to_kill" = true ]]; then
+      failed_count=$(expr ${failed_count} + 1)
+    else
+      success_count=$(expr ${success_count} + 1)
+    fi
   done
-  echo "Killed ${count} process(es) on $(hostname)"
+  if [[ $success_count -gt 0 ]]; then
+    echo "Successfully Killed ${success_count} process(es) successfully on $(hostname)"
+  fi
+  if [[ $failed_count -gt 0 ]]; then
+    echo "Soft kill option -s specified, failed to kill ${failed_count} process(es) on $(hostname)"
+  fi
 }
 
 function copyDir {
--- a/integration/fuse/bin/alluxio-fuse
+++ b/integration/fuse/bin/alluxio-fuse
@@ -88,9 +88,22 @@ umount_fuse() {
     echo "${mount_point} not mounted" >&2
     return 1
   else
-    echo "Unmount fuse at ${mount_point} (PID: ${fuse_pid})."
     kill ${fuse_pid}
-    return $?
+    local maxWait=60
+    local cnt=${maxWait}
+    while kill -0 ${fuse_pid} > /dev/null 2>&1; do
+      if [[ ${cnt} -gt 1 ]]; then
+        # still not dead, wait
+        cnt=$(expr ${cnt} - 1)
+        sleep 1
+      else
+        # waited long enough, giving up umount
+        echo "Failed to umount fuse process ["${fuse_pid}"] after waiting for "${maxWait}" seconds." >&2
+        return 1
+      fi
+    done
+    echo "Successfully unmount fuse at ${mount_point} (PID: ${fuse_pid})."
+    return 0
   fi
 }
 
--- a/integration/fuse/src/main/java/alluxio/fuse/FuseSignalHandler.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/FuseSignalHandler.java
@@ -1,0 +1,54 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.fuse;
+
+import alluxio.jnifuse.FuseException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.misc.Signal;
+import sun.misc.SignalHandler;
+
+/**
+ * Respond to the kill command when in mount fuse JVM.
+ * If it does not respond, the subsequent shutdown hook will not be triggered.
+ */
+public class FuseSignalHandler implements SignalHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(FuseSignalHandler.class);
+
+  /**
+   * Use to umount Fuse application during stop.
+   */
+  private FuseUmountable mFuseUmountable;
+
+  /**
+   * Constructs the new {@link FuseSignalHandler}.
+   * @param fuseUmountable mounted fuse application
+   */
+  public FuseSignalHandler(FuseUmountable fuseUmountable) {
+    mFuseUmountable = fuseUmountable;
+  }
+
+  @Override
+  public void handle(Signal signal) {
+    int number = signal.getNumber();
+    if (number == 15) {
+      try {
+        mFuseUmountable.umount(false);
+      } catch (FuseException e) {
+        LOG.error("unable to umount fuse.", e);
+        return;
+      }
+    }
+    System.exit(0);
+  }
+}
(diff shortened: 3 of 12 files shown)


apc999:
@yabola  are you still working on this PR?

yabola:
@apc999 right, but I want to confirm whether the idea of intercepting kill requests is appropriate or any other new idea on this issue.I will continue to complete this PR if it is an appropriate way.

LuQQiu:
@apc999 I remember that you took a look before and think that another function other than umount should be used?

LuQQiu:
@yabola 
We don't have a pretty good idea of how to properly umount fuse, so would like to know what's the behavior before and after your changes? Have you tested your changes out in standalone fuse and worker fuse? what's the behavior looks like.

yabola:
@LuQQiu 
My test step:
set properties` alluxio.fuse.umount.timeout=5s` ->  mount fuse dir ->
vi fuse_dir/b.txt ->  find fuse JVM Pid -> kill pid

Before this PR:
fuse JVM is killed.  the file which i vi before show`"fuse_dir/b.txt" E212: Can't open file for writing` (i use :wq command)

After this PR:
fuse JVM is not killed.The fuse JVM acts normally as before.

now this PR only catch kill signal 15 (kill pid). it hadn't catch other kill signal like kill -9.

LuQQiu:
@yabola this behavior looks good for the standalone fuse. When fuse is in the worker process, what's the behavior before and after your changes?
Basically we would want a solution that properly umount fuse in both standalone way and worker fuse. Ideally, when fuse device is busy, standalone fuse umount should be tricked as noop. Fuse in worker is a little tricky, we don't want the worker process components to be killed while the fuse mount is still there leaving the fuse mount in unusable state. We would rather force the fuse to be unmounted or prevent the whole worker process from exiting (trick worker process kill as noop). Not sure if it's possible

yabola:
@LuQQiu 
When fuse is in the worker process, it can act like standalone fuse . when use kill AlluxioWorker pid command, worker process kill as noop.  
But    `alluxio-stop.sh worker` command will kill -9 pid if kill pid failed after 120s. (https://github.com/Alluxio/alluxio/blob/master/bin/alluxio#L101). this command will finally kill worker process.
so if we need change `kill -9 worker pid` behavior here?

yabola:
@LuQQiu @apc999 
what's your opinions about that?

LuQQiu:
@yabola 
```
When fuse is in the worker process, it can act like standalone fuse . when use kill AlluxioWorker pid command, worker process kill as noop.
```
Can you elaborate more on this line?  why worker process kill will be treated as noop? Fuse will directly capture the kill signal and prevent that? it will prevent Alluxio worker stop all other functionalities?

yabola:
@LuQQiu 
 worker and  fuse is in one JVM when embedded-fuse worker started.`SignalHandler` is also registered in this JVM. So embedded-fuse worker JVM will listen to kill signal as standalone fuse do. when user use kill pid command, it will act as noop if fuse device is busy. However, it can't catch kill -9 pid signal.

LuQQiu:
As long as we get the desired behavior, your design and initial draft LGTM.
Thanks for the detailed description.
One small question is about the umount(force), the force parameter doesn't look like doing anything useful. 
For alluxio-stop.sh worker, we can add a force parameter, by default turn on to sync with the previous behavior, but user can pass in force=false to not trigger kill -9

yabola:
@LuQQiu fine,I will complete my PR later~

yabola:
Sorry for late. I have updated alluxio shell command as you said. 
This is my test evidence.


1. 
mount fuse dir in worker
```
alluxio.worker.fuse.enabled=true
alluxio.worker.fuse.mount.alluxio.path=/fuse_data
alluxio.worker.fuse.mount.point=/Users/chenliang.lu/soft/alluxio_data/fuse_data
```
2.  
open operation on mount file
`less /Users/chenliang.lu/soft/alluxio_data/fuse_data/b.txt`
image

3. 
 `sh bin/alluxio-stop.sh -s worker ` ---> won't kill worker
error message:
```
 Error, process [70125] did not complete after 120 seconds, failing to kill it.
Successfully Killed 0 process(es) successfully on ChengliangLudeMacBook-Pro.local
Failed to kill 1 process(es) due to soft kill option on ChengliangLudeMacBook-Pro.local
```
image

 4.
 `sh bin/alluxio-stop.sh  worker ` ---> will kill worker forcily finally
 
image
image

yabola:
Test Evidence in `alluxio-fuse`shell command：
When using the umount command, the fuse application does not respond.The AlluxioFuse process still exists and can be used normally.

image

LuQQiu:
Awesome!! @yabola Thanks for doing all the testing!!
The behavior LGTM!

tests/src/test/java/alluxio/client/fuse/JNIFuseIntegrationTest.java
```diff
@@ -41,6 +41,6 @@ public void mountFuse(FileSystem fileSystem, String mountPoint, String alluxioRo
 
   @Override
   public void umountFuse(String mountPath) throws Exception {
-    mFuseFileSystem.umount();
+    mFuseFileSystem.umount(false);
```
LuQQiu, line 44: why false here? will it cause the test to leave the fuse process running while closing?

bin/alluxio
```diff
@@ -79,31 +79,50 @@ EOF
 }
 
 function killAll {
-  if [[ $# -ne 1 ]]; then
-    echo "Usage: alluxio killAll <WORD>" >&2
+  if [[ $# -gt 2  ]]; then
+    echo "Usage: alluxio killAll [-s] <WORD>" >&2
```
LuQQiu, line 83: Did the `-s` only apply to AlluxioWorker?
In this usage, i will feel -s the soft kill option apply to all processes but looks like it only use in AlluxioWorker?

LuQQiu (commented):
@yabola Can you help fix the `integration/fuse/bin/alluxio-fuse umount` output message while saying that the fuse is mounted but actually fuse is not mounted because fuse divice is busy?

Another question is about the `-s` soft kill option. I will prefer to apply it to all processes not only AlluxioWorker. @apc999 @ggezer what's your advice about adding a soft kill option that do not force all processes to be killed? do not kill -9 if kill -15 doesn't work

tests/src/test/java/alluxio/client/fuse/JNIFuseIntegrationTest.java
yabola: done.

yabola:
@LuQQiu 
My test evidence in alluxio-fuse shell command:
when device is busy,fail to kill:
image
when device is free, successfully kill it:
image

I am not sure if `-s` should apply to all processes. If it needs, I'm willing to update it.

bin/alluxio
```diff
@@ -79,31 +79,50 @@ EOF
...
     local cnt=${maxWait}
+    fail_to_kill=false
     while kill -0 ${pid} > /dev/null 2>&1; do
       if [[ ${cnt} -gt 1 ]]; then
         # still not dead, wait
         cnt=$(expr ${cnt} - 1)
         sleep 1
+      elif [[ "${soft_kill_option}" = true ]] ; then
+        # waited long enough, give up killing the process
+        echo "Error, process ["${pid}"] did not complete after "${maxWait}" seconds, failing to kill it." >&2
```
LuQQiu, line 107:
```suggestion
        echo "Process ${pid} did not complete after "${maxWait}" seconds, failing to kill it." >&2
```

bin/alluxio
```diff
@@ -79,31 +79,50 @@ EOF
...
+    if [[ "$fail_to_kill" = true ]]; then
+      failed_count=$(expr ${failed_count} + 1)
+    else
+      success_count=$(expr ${success_count} + 1)
+    fi
   done
-  echo "Killed ${count} process(es) on $(hostname)"
+  echo "Successfully Killed ${success_count} process(es) successfully on $(hostname)"
+  if [[ $failed_count -gt 0 ]]; then
+    echo "Failed to kill ${failed_count} process(es) due to soft kill option on $(hostname)"
```
LuQQiu, line 124:
```suggestion
    echo "Soft kill option -s specified, failed to kill ${failed_count} process(es) on $(hostname)"
```

integration/fuse/bin/alluxio-fuse
```diff
@@ -88,9 +88,22 @@ umount_fuse() {
...
+    local maxWait=30
+    local cnt=${maxWait}
+    while kill -0 ${fuse_pid} > /dev/null 2>&1; do
+      if [[ ${cnt} -gt 1 ]]; then
+        # still not dead, wait
+        cnt=$(expr ${cnt} - 1)
+        sleep 1
+      else
+        # waited long enough, give up umount
+        echo "Error, fuse process ["${fuse_pid}"] did not complete after "${maxWait}" seconds, failing to umount fuse." >&2
```
LuQQiu, line 101:
```suggestion
        echo "Failed to umount fuse process ["${fuse_pid}"] after waiting for "${maxWait}" seconds." >&2
```

integration/fuse/src/main/java/alluxio/fuse/FuseSignalHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.LoggerFactory;
+import sun.misc.Signal;
+import sun.misc.SignalHandler;
+
+/**
+ * Respond to the kill command when in mount fuse JVM.
+ * If it does not respond, the subsequent shutdown hook will not be triggered.
+ */
+public class FuseSignalHandler implements SignalHandler {
+
```
LuQQiu, line 26: Remove extra empthy line

integration/fuse/src/main/java/alluxio/fuse/FuseSignalHandler.java
```diff
@@ -0,0 +1,55 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(FuseSignalHandler.class);
+
+  /**
+   * Use to umount Fuse application during stop.
+   */
+  private FuseUmountable mFuseUmountable;
+
+  /**
+   * init.
```
LuQQiu, line 35:
```suggestion
   * Constructs the new {@link FuseSingalHandler}.
   *
```

integration/fuse/src/main/java/alluxio/fuse/FuseUmountable.java
```diff
@@ -11,12 +11,16 @@
...
+import alluxio.jnifuse.FuseException;
+
 /**
  * An interface for unmounting the mounted Fuse applications.
  */
 public interface FuseUmountable {
   /**
    * Unmounts the Fuse application.
+   * @param force when the umount times out due to the device is busy,
+   *              it will throw exception (catched by {@link FuseSignalHandler})
```
LuQQiu, line 23:
```
   * @param force true to throw an exception and treat umount as a no-op 
   *            when umount timeouts due to the fuse device is busy.
```
modify the previous line to fit the indentation codestyle

LuQQiu (commented):
Left some minor comments. For the -s, can you add a description in bin/alluxio getUsage and telling that it is only used by AlluxioWorker.

LuQQiu:
can you help fix the conflicts as well? thanks!

LuQQiu:
@yabola Please modify the bin/alluxio-stop.sh usage as well, and add some related docs on the Alluxio POSIX API doc?

ggezer:
> @yabola Can you help fix the `integration/fuse/bin/alluxio-fuse umount` output message while saying that the fuse is mounted but actually fuse is not mounted because fuse divice is busy?
> 
> Another question is about the `-s` soft kill option. I will prefer to apply it to all processes not only AlluxioWorker. @apc999 @ggezer what's your advice about adding a soft kill option that do not force all processes to be killed? do not kill -9 if kill -15 doesn't work

@LuQQiu I think that's fair. However, I'd like to get that done outside of this PR. ( Ideally this PR just rebases after that PR is merged.)

LuQQiu:
Sounds good @ggezer, do you prefer @yabola to work on adding the -s to all processes or other engineers?

ggezer:
> Sounds good @ggezer, do you prefer @yabola to work on adding the -s to all processes or other engineers?

Whoever has bandwidth really :D 
@yabola We'd appreciate if you could isolate this change into a separate PR complete with all processes.

yabola:
> > Sounds good @ggezer, do you prefer @yabola to work on adding the -s to all processes or other engineers?
> 
> Whoever has bandwidth really :D
> @yabola We'd appreciate if you could isolate this change into a separate PR complete with all processes.

@ggezer @LuQQiu Do you mean first pull a PR for `-s` option, then  to compete umount fuse option PR?  I am willing to compete it~~

bin/alluxio
```diff
@@ -79,31 +79,50 @@ EOF
...
     done
-    count=$(expr ${count} + 1)
+    if [[ "$fail_to_kill" = true ]]; then
+      failed_count=$(expr ${failed_count} + 1)
+    else
+      success_count=$(expr ${success_count} + 1)
+    fi
   done
-  echo "Killed ${count} process(es) on $(hostname)"
+  echo "Successfully Killed ${success_count} process(es) successfully on $(hostname)"
```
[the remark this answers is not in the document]
yabola: done

bin/alluxio
yabola: done

bin/alluxio
yabola: done

integration/fuse/bin/alluxio-fuse
yabola: done

integration/fuse/src/main/java/alluxio/fuse/FuseSignalHandler.java
yabola: done

integration/fuse/src/main/java/alluxio/fuse/FuseUmountable.java
yabola: done

integration/fuse/src/main/java/alluxio/fuse/FuseSignalHandler.java
yabola: done

yabola:
I have updated my PR to add soft kill option to all process. I will separate PR into two PR.

PR #13374, opened 2021-05-08
Fix possible NullPointerException when calling e.getCause

zuston:


LuckyNemo:
In my eye，when using e.getCause().toString() method,should think twice, casue e.getCause() may be "null"!
Plz fix it asap.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -685,7 +685,7 @@ public synchronized void startInternal() throws InterruptedException, IOExceptio
       mServer.start();
     } catch (IOException e) {
       String errorMessage = ExceptionMessage.FAILED_RAFT_BOOTSTRAP
-          .getMessage(Arrays.toString(clusterAddresses.toArray()), e.getCause().toString());
+          .getMessage(Arrays.toString(clusterAddresses.toArray()), e);
```
gpang, line 688: I'm not sure what the original intent was, but the new code changes the original behavior. The original would print the cause, but this PR will print the original exception.

Maybe we print the `e.getCause()` if it is not `null`, and `e` otherwise?

gpang (commented):
@zuston Thanks for the improvement! I left a question.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
zuston: Sounds good.

LuckyNemo: > I'm not sure what the original intent was, but the new code changes the original behavior. The original would print the cause, but this PR will print the original exception.
> 
> Maybe we print the `e.getCause()` if it is not `null`, and `e` otherwise?

Right! It is the right way.It can give users more useful info about the exception rather than null pointer exception.

LuQQiu:
@apc999 PTAL, thanks

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
apc999, line 688:
```suggestion
          .getMessage(Arrays.toString(clusterAddresses.toArray()), 
              e.getCause() == null ? e : e.getCause().toString());
```

apc999 (changes requested):

apc999:
@zuston can you address the comments left by me or Gene?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
zuston: Done

apc999 (approved):

LuQQiu:
@zuston Please fix the `mvn checkstyle:checkstyle` issue so that we can merge this PR

apc999:

zuston:
Sorry for the late reply. @apc999  Thanks for fixing it.

PR #13384, opened 2021-05-10
Refactor MasterWorkerInfo

jiacheliu3:
Currently all the worker metadata are wrapped in `MasterWorkerInfo` object on the master side. This includes worker metadata, usage and the various block lists. The existing `DefaultBlockMaster` does a `synchronized (MasterWorkerInfo)` locking on this object on every read or write operation on any field in this object. This locking is excessive.

By refactoring this `MasterWorkerInfo` object and refactoring the locking mechanism, the `DefaultBlockMaster` concurrency will be improved, as multiple operations will be able to happen on the same worker.

That said, this PR aims at the following:
1. Refactor `MasterWorkerInfo` object. Break this into multiple parts which can be individually locked and updated.
2. Refactor the locking to lock individual parts on the object instead of  the whole object. RW Locking should also be applied.


jiacheliu3:
The change to stream the registerWorker RPC call will be removed from this PR and moved to a separate PR, so this PR will be more standalone. We will be able to focus on this part of the improvement and move on to the next stage faster.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -814,25 +890,23 @@ private MasterWorkerInfo findUnregisteredWorker(long workerId) {
...
-  private MasterWorkerInfo registerWorkerInternal(long workerId) {
+  private MasterWorkerInfo recordWorkerRegistration(long workerId) {
     for (IndexedSet<MasterWorkerInfo> workers: Arrays.asList(mTempWorkers, mLostWorkers)) {
       MasterWorkerInfo worker = workers.getFirstByField(ID_INDEX, workerId);
       if (worker == null) {
         continue;
       }
 
-      synchronized (worker) {
-        worker.updateLastUpdatedTimeMs();
```
jiacheliu3, line 825: This update is not really needed here. I removed this line to the end of the whole register logic, so we don't need to worry about locking in this method just for updating this TS.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -814,25 +890,23 @@ private MasterWorkerInfo findUnregisteredWorker(long workerId) {
    * @param workerId the worker id to register
    */
   @Nullable
-  private MasterWorkerInfo registerWorkerInternal(long workerId) {
+  private MasterWorkerInfo recordWorkerRegistration(long workerId) {
```
jiacheliu3, line 893: This method just updates the `DefaultBlockMaster` internal worker list, adding this worker to the internal available worker list. This methods only needs 2 fields from the `MasterWorkerInfo`:
1. `worker.getAddress()` which is a final field
2. `worker.updateLastUpdatedTimeMs()` which is removed from this method(see comment below).

Therefore this methods does not need locking.

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
```diff
@@ -380,10 +416,130 @@ public void getBlockInfo() throws Exception {
     assertEquals(expectedBlockInfo, mBlockMaster.getBlockInfo(blockId));
   }
 
+  /**
+   * RW contention
+   * Concurrent commit and readers
+   * Signal in commit and the readers inquire the state
+   * */
+  @Test
+  public void concurrentCommitWithReaders() throws Exception {
```
jiacheliu3, line 425: @apc999 This is the test I wanted to discuss about. In this PR I change the locking in `DefaultBlockMaster` from a big `synchronized` section to using RW locks. So the testing on RW and WW contentions are more necessary than before.

In short the goal is to deterministically trigger an R or W2 operation while the W1 operation is running. In this way the reader or writer will test if W1 operation is protecting the internal metadata properly.

The `SignalBlockMaster` extends the `DefaultBlockMaster` but sends a signal in the middle of a write operation. The signal is sent on `createJournalContext()` because both `commit` and `remove` will call this in the middle, and this is part of the public API for the `BlockMaster` so it's better than hacking a private method.

jiacheliu3: Do you think this kind of injection is acceptable? The interface is mostly intact. The cons might be: (1) I removed `final` on `DefaultBlockMaster`; (2) `workerRegister` and `workerHeartbeat` in `DefaultBlockMaster` cannot be injected the same way because they don't create a journal context.

If not, are there better ways?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -108,47 +100,18 @@ public MasterWorkerInfo(long id, WorkerNetAddress address) {
    * @param blocks set of block ids on this worker
    * @return A Set of blocks removed (or lost) from this worker
    */
+  // TODO(jiacheng): Do I need to write GuardedBy for member methods like this?
```
jiacheliu3, line 103: I have 2 questions:
1. Do I need to write GuardedBy on this kind of member methods, given that I declared this object needs locking logic externally? I guess the most informative approach is yes.
2. AFAIK `GuardedBy` doesn't allow specifying multiple locks, what should I do?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -108,47 +100,18 @@ public MasterWorkerInfo(long id, WorkerNetAddress address) {
...
-    mTotalBytesOnTiers = new HashMap<>(totalBytesOnTiers);
-    mUsedBytesOnTiers = new HashMap<>(usedBytesOnTiers);
-    mCapacityBytes = 0;
-    for (long bytes : mTotalBytesOnTiers.values()) {
-      mCapacityBytes += bytes;
-    }
-    mUsedBytes = 0;
-    for (long bytes : mUsedBytesOnTiers.values()) {
-      mUsedBytes += bytes;
-    }
```
jiacheliu3, line 145: Moved into a method

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -108,47 +100,18 @@ public MasterWorkerInfo(long id, WorkerNetAddress address) {
...
+    mUsage.updateUsage(globalStorageTierAssoc, storageTierAliases,
+            totalBytesOnTiers, usedBytesOnTiers);
 
     Set<Long> removedBlocks;
-    if (mIsRegistered) {
+    if (mMeta.mIsRegistered) {
```
jiacheliu3, line 111: I'm not using getters and setters in the `MasterWorkerInfo` to access internal object fields like this. Is it okay?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -157,9 +120,9 @@ public MasterWorkerInfo(long id, WorkerNetAddress address) {
     }
 
     // Set the new block information.
-    mBlocks = new HashSet<>(blocks);
+    mBlocks.addAll(blocks);
```
jiacheliu3, line 123: so that the `mBlocks` can be final

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -366,17 +331,18 @@ public boolean isRegistered() {
...
   /**
    * @return true if this worker has lost storage, false otherwise
    */
   public boolean hasLostStorage() {
-    return mLostStorage.size() > 0;
+    return mUsage.mLostStorage.size() > 0;
   }
 
   @Override
+  // TODO(jiacheng): is this locking
```
jiacheliu3, line 345: Actually when we are not locking on `toString` properly....
https://github.com/Alluxio/alluxio/blob/9f5673eccf6452c13493ae3bcd38f6203714f16a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java#L912
@ggezer

core/server/master/src/main/java/alluxio/master/block/meta/WorkerUsageMeta.java
```diff
@@ -0,0 +1,82 @@
...
+    mUsedBytesOnTiers = new HashMap<>(usedBytesOnTiers);
+    mCapacityBytes = 0;
+    for (long bytes : mTotalBytesOnTiers.values()) {
+      mCapacityBytes += bytes;
+    }
+    mUsedBytes = 0;
+    for (long bytes : mUsedBytesOnTiers.values()) {
+      mUsedBytes += bytes;
+    }
+  }
```
jiacheliu3, line 74: This is just moved here from `MasterWorkerInfo`

core/server/master/src/test/java/alluxio/master/block/SingalBlockMaster.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Override
+  public JournalContext createJournalContext() throws UnavailableException {
+    JournalContext context = super.createJournalContext();
+
+    // The reader will wait for this signal and start to read
+    mLatch.countDown();
+
+    return context;
+  }
```
jiacheliu3, line 81: All this extension does is to fire a signal in the middle of a write operation, so the test code knows it's a good time to fire read events to test against the locking.

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMeta.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class WorkerMeta {
+  /** Worker's address. */
+  public final WorkerNetAddress mWorkerAddress;
+  /** The id of the worker. */
+  public final long mId;
+  /** Start time of the worker in ms. */
+  public final long mStartTimeMs;
+  /** Worker's last updated time in ms. */
+  public final AtomicLong mLastUpdatedTimeMs;
```
jiacheliu3, line 19: I'm using Atomic here because I want to make it easier to read and update this field without locking. The RW lock on WorkerMeta is actually just for the `mIsRegistered` field. It might be harder to get rid of that lock totally, because when we write lock the `WorkerMeta` we want the `mIsRegistered` field to be mutex. But honestly the only corner case that can lead to this race condition is two parallel requests registering the same worker. I'm not sure if that should be handled (now the behavior is not too well defined).

jiacheliu3:
@ggezer PTAL for the 1st round review. The missing parts are formating and unit tests. There are 2 or 3 TODOs for myself, because I wanted to discuss with you on those first. Thanks!

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -679,28 +705,31 @@ public void commitBlock(long workerId, long usedBytesOnTier, String tierAlias,
...
           // Update the block metadata with the new worker location.
           mBlockStore.addLocation(blockId, BlockLocation.newBuilder()
-              .setWorkerId(workerId)
-              .setTier(tierAlias)
-              .setMediumType(mediumType)
-              .build());
+                  .setWorkerId(workerId)
+                  .setTier(tierAlias)
+                  .setMediumType(mediumType)
+                  .build());
```
apc999, line 732: i see multiple unnecessary indent change in this PR, please fix them

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -116,7 +116,7 @@
  * This block master manages the metadata for all the blocks and block workers in Alluxio.
```
apc999, line 116: can you specify the updated locking hierarchy here? like what we did in javadoc of `DefaultFileSystemMaster`

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -518,32 +522,50 @@ public long getUsedBytes() {
...
-      }
+      // extractWorkerInfo handles the locking internally
+      workerInfoList.add(extractWorkerInfo(worker, options.getFieldRange(), false));
     }
     return workerInfoList;
   }
 
+  /**
+   * Lock the {@link MasterWorkerInfo} properly and convert it to a {@link WorkerInfo}
+   * */
```
apc999, line 537:
```suggestion
   * Locks the {@link MasterWorkerInfo} properly and converts it to a {@link WorkerInfo}.
   **/
```

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -518,32 +522,50 @@ public long getUsedBytes() {
...
           Map<String, StorageList> lostStorage = worker.getLostStorage().entrySet()
-              .stream().collect(Collectors.toMap(Map.Entry::getKey,
-                  e -> StorageList.newBuilder().addAllStorage(e.getValue()).build()));
+                  .stream().collect(Collectors.toMap(Map.Entry::getKey,
+                          e -> StorageList.newBuilder().addAllStorage(e.getValue()).build()));
           workerLostStorageList.add(WorkerLostStorageInfo.newBuilder()
-              .setAddress(GrpcUtils.toProto(worker.getWorkerAddress()))
-              .putAllLostStorage(lostStorage).build());
+                  .setAddress(GrpcUtils.toProto(worker.getWorkerAddress()))
+                  .putAllLostStorage(lostStorage).build());
```
apc999, line 563: revert unnecessary indent change

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
```diff
@@ -230,8 +230,8 @@ public void register(final long workerId, final List<String> storageTierAliases,
         .setOptions(options).build();
 
     retryRPC(() -> {
-      mClient.registerWorker(request);
-      return null;
+        mClient.registerWorker(request);
+        return null;
```
apc999, line 234: revert

core/server/master/src/test/java/alluxio/master/block/SingalBlockMaster.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * When the writer is writing, issue a signal so the reader will start reading
+ * */
```
apc999, line 49:
```suggestion
 **/
```

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -49,33 +54,21 @@
...
-  /** Mapping from storage tier alias to total bytes. */
-  private Map<String, Long> mTotalBytesOnTiers;
-  /** Mapping from storage tier alias to used bytes. */
-  private Map<String, Long> mUsedBytesOnTiers;
+  /** worker metadata */
+  private final WorkerMeta mMeta;
+  /** lock the worker metadata */
+  private final ReentrantReadWriteLock mMetaLock;
+  /** worker usage data */
+  private final WorkerUsageMeta mUsage;
```
apc999, line 62: `GuardedBy("mUsageLock")` annotation?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -49,33 +54,21 @@
...
-  /** If true, the worker is considered registered. */
-  private boolean mIsRegistered;
-  /** Worker-specific mapping between storage tier alias and storage tier ordinal. */
-  private StorageTierAssoc mStorageTierAssoc;
-  /** Mapping from storage tier alias to total bytes. */
-  private Map<String, Long> mTotalBytesOnTiers;
-  /** Mapping from storage tier alias to used bytes. */
-  private Map<String, Long> mUsedBytesOnTiers;
+  /** worker metadata */
+  private final WorkerMeta mMeta;
```
apc999, line 58: `GuardedBy("mMetaLock")` annotation?
Also, not sure if you really need to protect `WorkerMeta` which has only one mutable info `mLastUpdatedTimeMs`.
In that case, you can move the `mLastUpdatedTimeMs` to `UsageMeta` (or rename it to something better)

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -49,33 +54,21 @@
...
+  /** lock the worker metadata */
+  private final ReentrantReadWriteLock mMetaLock;
+  /** worker usage data */
+  private final WorkerUsageMeta mUsage;
+  /** lock the worker usage data */
+  private final ReentrantReadWriteLock mUsageLock;
 
   /** ids of blocks the worker contains. */
-  private Set<Long> mBlocks;
+  private final Set<Long> mBlocks;
```
apc999, line 67: `GuardedBy`

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMeta.java
apc999: it looks to me this is the only mutable part. How about making this meta info constant, like `StaticWorkerMeta` or even putting them back to `MasterWorkerInfo`, and moving `mLastUpdatedTimeMs` out to other meta part? In this way, accessing these readonly fields requires no locks, eliminating one RW lock.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -404,8 +404,11 @@ public int getLostWorkerCount() {
   public long getCapacityBytes() {
     long ret = 0;
     for (MasterWorkerInfo worker : mWorkers) {
-      synchronized (worker) {
+      worker.getUsageLock().readLock().lock();
```
apc999, line 407: how about following the code convention in `DefautlFileSystemMaster` and returning a `LockResrouce` instead of the lock, so the code is more error-prune 
( I don't have to check the `finally` block and the order)

```
try (LockResource r = new LockResource(worker.getUsageLock().readLock())) {
  ret += worker.getCapacityBytes();
}
```

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
apc999: - removing `final` is actually good. I would even suggest removing most `final` modifiers in our code base for important classes.
- another way you can do is to overriding `alluxio.master.block.DefaultBlockMaster#lockBlock` which returns a modified `LockResource` --- in your modified `LockResource`,  you can signal other threads and wait something to happen before calling `unlock`

apc999 (changes requested):

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Oh I didn't know that. Thanks will do!

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
jiacheliu3: @apc999 My only concern about overriding `alluxio.master.block.DefaultBlockMaster#lockBlock` is, that's a private method. I can make it inheritable, of course, but is inheriting a public `createJournalContext` more elegant than `lockBlock` or they are actually the same? Is it fine to rely on the fact that `remove` is calling either `lockBlock` or `createJournalContext`? Would you consider the test rely on method internals?

Other than that, inheriting `lockBlock` is a better choice because that enables me to throw signals from worker register and heartbeat.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: Yea I'm thinking about doing that too

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: @apc999  I've applied `LockResource` whenever I only have a single lock. For multiple locks the order does matter I think. The order or acquiring multiple locks should be uniformed across functions and the order or releasing should be the opposite, in order to guard deadlocks. How do I use the `LockResource` to achieve that?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Will do at the end of the review. Once we are settled on the implementations I will make sure the checkstyle and the style are happy.

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMeta.java
jiacheliu3: I managed to make this atomic so the locking on this field is unnecessary. The `mIsRegistered` field has a check-then-act section so still the lock is needed. I moved `mIsRegistered` to the upper level `MasterWorkerInfo` and use an exclusive lock for that now. Now the `WorkerMeta` class is thread safe (not static because of the last updated ts).

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Updated @apc999 Please check the section below. In `DefaultFileSystemMaster` it's the same pattern where these comments are not in the class commend, but somewhere in the class. I'm just following that w/o knowing why.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -782,6 +839,7 @@ public BlockInfo getBlockInfo(long blockId) throws BlockInfoException, Unavailab
    */
   @Nullable
   private MasterWorkerInfo findUnregisteredWorker(WorkerNetAddress workerNetAddress) {
+    // TODO(jiacheng): Is this causing mem usage? Arrays.asList(mTempWorkers, mLostWorkers)
```
jiacheliu3, line 842: Left a TODO for future optimization, not in this PR

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -924,29 +999,56 @@ public Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnT
...
 
+      // Technically, 'worker' should be confirmed to still be in the data structure. Lost worker
+      // detection can remove it. However, we are intentionally ignoring this race, since the worker
+      // will just re-register regardless.
+
+      // This changes the block list
+      processWorkerRemovedBlocks(worker, removedBlockIds);
+      processWorkerAddedBlocks(worker, addedBlocks);
+      // TODO(jiacheng): If the worker registers with a after the block is freed, the copy will not be removed
+      //  from this worker.
```
jiacheliu3, line 1029: @apc999 @ggezer Should this be the defined behavior?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: @apc999 @ggezer PTAL I have some small questions on the coding style, thanks!

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -366,17 +346,18 @@ public boolean isRegistered() {
...
   /**
    * @return true if this worker has lost storage, false otherwise
    */
   public boolean hasLostStorage() {
-    return mLostStorage.size() > 0;
+    return mUsage.mLostStorage.size() > 0;
   }
 
   @Override
+  // TODO(jiacheng): Should this be locking?
```
jiacheliu3, line 360: @ggezer @apc999 This is potentially harmful but is it too heavy to lock here?

jiacheliu3:
TODO:
1. Fill in the blanks for unit tests (they are just a lot!)
2. Rebase to resolve conflicts (those are also my PRs..)
3. Fix indentation etc

jiacheliu3:
> TODO:
> 
> 1. Fill in the blanks for unit tests (they are just a lot!)
> 2. Rebase to resolve conflicts (those are also my PRs..)
> 3. Fix indentation etc

The above TODOs are all resolved. There are a few `TODO(jiacheng)` in the code. Those are either things I want to bring to your notice or I plan to improve in the future.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: See the big comment block below

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -926,30 +999,57 @@ public Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnT
...
+      worker.getBlockListLock().writeLock().unlock();
+      worker.getUsageLock().writeLock().unlock();
     }
+
+    // Update the TS again
+    worker.updateLastUpdatedTimeMs();
+
+    // Should not reach here
+    if (workerCommand == null) {
+      LOG.error("Worker heartbeat response command is null!");
```
jiacheliu3, line 1049: Better way to handle this like `Preconditions.checkNotNull`?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -965,7 +1065,7 @@ private void processWorkerMetrics(String hostname, List<Metric> metrics) {
    * @param workerInfo The worker metadata object
    * @param removedBlockIds A list of block ids removed from the worker
    */
-  @GuardedBy("workerInfo")
+  @GuardedBy("workerInfo.mBlockListLock")
```
jiacheliu3, line 1068: Do I refer to the field or getter method `workerInfo.getBlockListLock()`?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -907,7 +983,16 @@ public void workerRegister(long workerId, List<String> storageTiers,
       }
     }
 
-    registerWorkerInternal(workerId);
+    /**
+     * No locking on the {@link MasterWorkerInfo} is required in this method
+     * as it does not update the object fields.
```
ggezer, line 988: This is a bit verbose to be included in the call-site.
For this and some others, could you move such internal info to the method comments itself?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -907,7 +983,16 @@ public void workerRegister(long workerId, List<String> storageTiers,
...
+     */
+    recordWorkerRegistration(workerId);
+
+    // Update the TS at the end of the process
+    // This operation is atomic, no need for locking
```
ggezer, line 993: same here.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -926,30 +1011,57 @@ public Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnT
...
+    // Update the TS before the heartbeat so even if the worker heartbeat processing
+    // is time-consuming or triggers GC, the worker does not get marked as lost
+    // by the LostWorkerDetectionHeartbeatExecutor.
+    worker.updateLastUpdatedTimeMs();
+
+    // The address is final, no need for locking
+    processWorkerMetrics(worker.getWorkerAddress().getHost(), metrics);
 
+    worker.getUsageLock().writeLock().lock();
+    worker.getBlockListLock().writeLock().lock();
```
ggezer, line 1023: another target for unification of lock management.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -926,30 +1011,57 @@ public Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnT
...
 
+      // Technically, 'worker' should be confirmed to still be in the data structure. Lost worker
+      // detection can remove it. However, we are intentionally ignoring this race, since the worker
+      // will just re-register regardless.
+
+      // This changes the block list
+      processWorkerRemovedBlocks(worker, removedBlockIds);
+      processWorkerAddedBlocks(worker, addedBlocks);
+      // TODO(jiacheng): If the worker registers with a after the block is freed,
+      //  the copy will not be removed from this worker.
```
ggezer, line 1041: will this be addressed in this PR?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
ggezer: If the code should not reach here, then why not throw with `Preconditions.checkNotNull`?
It's more like `Postconditions` though :/

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
ggezer: I didn't know you could attach this to a method. 
Honestly, it doesn't make sense to me.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
ggezer: 1- methods doesn't make sense to me for this annotation.
2- I think every individual member should be protected by a single lock only. So unless applying on a method, we shouldn't need to specify multiple locks.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -189,9 +161,9 @@ public void removeBlock(long blockId) {
    * @param dirPath the lost storage path
    */
   public void addLostStorage(String tierAlias, String dirPath) {
-    List<String> paths = mLostStorage.getOrDefault(tierAlias, new ArrayList<>());
+    List<String> paths = mUsage.mLostStorage.getOrDefault(tierAlias, new ArrayList<>());
```
ggezer, line 164: is this method called with usage-lock held? If so can you update the method comment?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -201,9 +173,9 @@ public void addLostStorage(String tierAlias, String dirPath) {
    */
   public void addLostStorage(Map<String, StorageList> lostStorage) {
     for (Map.Entry<String, StorageList> entry : lostStorage.entrySet()) {
-      List<String> paths = mLostStorage.getOrDefault(entry.getKey(), new ArrayList<>());
+      List<String> paths = mUsage.mLostStorage.getOrDefault(entry.getKey(), new ArrayList<>());
```
ggezer, line 176: is this method called with usage-lock held? If so can you update the method comment?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -221,23 +193,23 @@ public WorkerInfo generateWorkerInfo(Set<WorkerInfoField> fieldRange, boolean is
     for (WorkerInfoField field : checkedFieldRange) {
       switch (field) {
```
ggezer, line 194: is this method called with usage and meta lock held? 
If so can you update the method comment?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -284,21 +258,25 @@ public long getAvailableBytes() {
    * @return the capacity of the worker in bytes
    */
   public long getCapacityBytes() {
-    return mCapacityBytes;
+    return mUsage.mCapacityBytes;
```
ggezer, line 261: does this and such methods require locking outside?
can you update comments if so?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
ggezer: ideally with read locking, or string could be inconsistent.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -425,11 +407,12 @@ public void updateToRemovedBlock(boolean add, long blockId) {
...
-    mCapacityBytes = 0;
-    mTotalBytesOnTiers = capacityBytesOnTiers;
-    for (long t : mTotalBytesOnTiers.values()) {
-      mCapacityBytes += t;
+    long capacityBytes = 0;
+    mUsage.mTotalBytesOnTiers = capacityBytesOnTiers;
+    for (long t : mUsage.mTotalBytesOnTiers.values()) {
+      capacityBytes += t;
     }
+    mUsage.mCapacityBytes = capacityBytes;
```
ggezer, line 415: should this be locked outside?

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMeta.java
```diff
@@ -0,0 +1,54 @@
...
+@ThreadSafe
+public class WorkerMeta {
+  /** Worker's address. */
+  final WorkerNetAddress mWorkerAddress;
+  /** The id of the worker. */
+  final long mId;
+  /** Start time of the worker in ms. */
+  final long mStartTimeMs;
+  /** Worker's last updated time in ms. */
+  final AtomicLong mLastUpdatedTimeMs;
```
ggezer, line 36: Would this fit better to `MasterWorkerInfo` since it doesn't require lock like its neighbors?

core/server/master/src/test/java/alluxio/master/block/ConcurrentBlockMasterTest.java
```diff
@@ -0,0 +1,1189 @@
...
+import java.util.Map;
+import java.util.Queue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class ConcurrentBlockMasterTest {
```
ggezer, line 63: This is very comprehensive. Thanks!

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Yea exactly why i didn't put a `Preconditions` here :)

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -811,29 +878,29 @@ private MasterWorkerInfo findUnregisteredWorker(long workerId) {
...
 
-      synchronized (worker) {
-        worker.updateLastUpdatedTimeMs();
-        mWorkers.add(worker);
-        workers.remove(worker);
-        if (workers == mLostWorkers) {
-          for (Consumer<Address> function : mLostWorkerFoundListeners) {
-            function.accept(new Address(worker.getWorkerAddress().getHost(),
+      mWorkers.add(worker);
+      workers.remove(worker);
```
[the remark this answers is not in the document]
jiacheliu3: After staring at this a bit I realized we are iterating a list of Sets
```
for (IndexedSet<MasterWorkerInfo> workers: Arrays.asList(mTempWorkers, mLostWorkers))
```
So updating `workers` in the loop is fine

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: This will not be addressed in this PR.

On 2nd thought, I'm not sure if this behavior should be marked as "expected". 

The sequence is like this:
1. `removeBlocks(blockList, false)` is called where the boolean field means only freeing the blocks but not deleting metadata
2. `removeBlocks` gets the current block locations from the `BlockStore`. This step holds the on block lock. The block lock is now released. The locations are stored in variable `workerIds`.
3. Worker comes in with a `register`. This gets the block lock and updates the location in `BlockStore`.
4. `removeBlocks` loops `workerIds` and adds this blockId to the corresponding `MasterWorkerInfo`. This step holds the `MasterWorkerInfo` locks.
5. `removeBlocks` completes, the copy on the worker is never marked as to-be-removed.

Step 3 and 4 are concurrent and orthogonal.

jiacheliu3: To clarify more, when the `registerWorker` is concurrent with `removeBlocks(blockList, true)` where `true` indicates the block metadata are removed, the block on the worker WILL be removed correctly. The behavior above only happens when the operations is a `free`.

jiacheliu3: I feel this behavior is actually okay. The concurrent `registerWorker` is not very different from a worker that registers way after the `free`. The block can stay. We don't have a `free-and-no-cache` semantic.

@ggezer What do you think?

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMeta.java
jiacheliu3: Moved into `MasterWorkerInfo`

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -680,10 +738,13 @@ public void commitBlock(long workerId, long usedBytesOnTier, String tierAlias,
       throw new NotFoundException(ExceptionMessage.NO_WORKER_FOUND.getMessage(workerId));
     }
 
-    // Lock the worker metadata first.
     try (JournalContext journalContext = createJournalContext()) {
-      synchronized (worker) {
-        try (LockResource lr = lockBlock(blockId)) {
+      // Lock the worker metadata here to preserve the lock order
+      // The worker metadata must be locked before the blocks
+      worker.getUsageLock().writeLock().lock();
```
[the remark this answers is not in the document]
jiacheliu3: Changed to using the new utility method in `MasterWorkerInfo` instead of `LockResource`. See comments below.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Cool. Removed from the `DefaultBlockMaster` methods

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMeta.java
jiacheliu3: Moved `mLastUpdatedTimsMs` out of `WorkerMeta` to `MasterWorkerInfo`

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
jiacheliu3: Changed to throw a signal out of `lockBlock`

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: Added comments to this method and the neighbors that require external locking. (Looking a bit verbose now :/)

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -422,37 +549,106 @@ public void updateToRemovedBlock(boolean add, long blockId) {
...
+   * Locks the corresponding locks on the metadata groups.
+   * The locks will be acquired in order.
+   * The locks can either be shared or exclusive.
+   * The isShared flag will apply to all the locks acquired here.
+   *
+   * @param lockTypes the locks
+   * @param isShared if false, the locking is exclusive
+   */
+  // TODO(jiacheng): refactor this to support LockResource instead of try-finally
+  public void lock(EnumSet<LockType> lockTypes, boolean isShared) {
```
jiacheliu3, line 620: I wasn't able to achieve `lock(STATUS_LOCK  | USAGE_LOCK, true)` because java doesn't support using enum like integers (unlike Go). This seems to be the recommended way https://stackoverflow.com/questions/6282619/is-there-a-way-to-bitwise-or-enums-in-java

Compared to add an int value to each enum and bitwise with `lock(STATUS_LOCK.value | USAGE_LOCK.value)`, I prefer using `EnumSet` because it guards the argument to be enum instead of a random int. 

I guess you were thinking about sth like [this](https://github.com/Alluxio/alluxio/blob/e6b7900126ebb8d236410f35ad69b6ffdb258470/core/client/fs/src/main/java/alluxio/client/ReadType.java#L77)?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -154,23 +154,73 @@ public WorkerNetAddress getFieldValue(MasterWorkerInfo o) {
...
+   *   }
+   * </pre></blockquote>
+   *
+   * When multiple locks are needed, the locks must be acquired in order
+   * and released in the opposite order. The acquisition order is the order above (1 to 4).
+   * For example, if you need to check the worker register status then update the worker
+   * resource usage and block list, the usage will look like:
+   *
+   * <blockquote><pre>
+   *    MasterWorkerInfo worker = ...;
```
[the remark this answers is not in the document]
jiacheliu3: This has changed much due to the new utility in `MasterWorkerInfo`. Please see the updated code.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -154,23 +154,73 @@ public WorkerNetAddress getFieldValue(MasterWorkerInfo o) {
...
+   *     This is guarded by {@link MasterWorkerInfo#getBlockListLock()}.
+   *     The lock is a {@link ReentrantReadWriteLock}.
+   *
+   * While accessing or updating the fields in the {@link MasterWorkerInfo}, relevant locks must
+   * be acquired first.
+   *
+   * When there is only one lock, {@link LockResource} can be used to simplify the lock management:
+   * <blockquote><pre>
+   *   MasterWorkerInfo worker = ...;
+   *   try (LockResource r = new LockResource(worker.getUsageLock().readLock()) {
```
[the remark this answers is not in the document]
jiacheliu3: See comment https://github.com/Alluxio/alluxio/pull/13384/files#r640406297

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -154,23 +154,73 @@ public WorkerNetAddress getFieldValue(MasterWorkerInfo o) {
...
+   * When multiple locks are needed, the locks must be acquired in order
+   * and released in the opposite order. The acquisition order is the order above (1 to 4).
+   * For example, if you need to check the worker register status then update the worker
+   * resource usage and block list, the usage will look like:
+   *
+   * <blockquote><pre>
+   *    MasterWorkerInfo worker = ...;
+   *    worker.getRegisterLock().lock();
+   *    worker.getUsageLock.writeLock().lock();
+   *    worker.getBlockListLock.writeLock().lock();
```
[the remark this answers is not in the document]
jiacheliu3: See comment below https://github.com/Alluxio/alluxio/pull/13384/files#r640406297

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
jiacheliu3: The only toString conversion seems to be [here](https://github.com/Alluxio/alluxio/blob/e6b7900126ebb8d236410f35ad69b6ffdb258470/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java#L914) but i'm not sure if there are more that I don't catch. I'm actually leaving a TODO here because:
1. This PR doesn't change the existing behavior of a unlocked toString()
2. How to guard properly and return a String with as little locking as possible is not on the top of my head. Let's not block this PR with that discussion.

jiacheliu3: Updated the TODO message btw.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -455,8 +456,13 @@ public int getLostWorkerCount() {
   public long getCapacityBytes() {
     long ret = 0;
     for (MasterWorkerInfo worker : mWorkers) {
-      try (LockResource r = new LockResource(worker.getUsageLock().readLock())) {
+      EnumSet<MasterWorkerInfo.LockType> lockTypes =
+          EnumSet.of(MasterWorkerInfo.LockType.USAGE_LOCK);
```
ggezer, line 423: can you make this inline where-ever used?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -956,9 +991,11 @@ public void workerRegister(long workerId, List<String> storageTiers,
...
 
     // Lock all the locks
-    worker.getRegisterLock().writeLock().lock();
-    worker.getUsageLock().writeLock().lock();
-    worker.getBlockListLock().writeLock().lock();
+    EnumSet<MasterWorkerInfo.LockType> lockTypes = EnumSet.of(
+        MasterWorkerInfo.LockType.STATUS_LOCK,
+        MasterWorkerInfo.LockType.USAGE_LOCK,
+        MasterWorkerInfo.LockType.BLOCKS_LOCK);
+    worker.lock(false, lockTypes);
```
ggezer: Would it be OK to make this inline, if worker.lock() returns a `LockResource`?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -464,12 +608,16 @@ public void updateUsedBytes(String tierAlias, long usedBytesOnTier) {
...
+   * The locks will be acquired in order.
+   * The locks can either be shared or exclusive.
+   * The isShared flag will apply to all the locks acquired here.
    *
-   * @param isShared if false, the locking is exclusive
    * @param lockTypes the locks
+   * @param isShared if false, the locking is exclusive
    */
-  public void lock(boolean isShared, EnumSet<LockType> lockTypes) {
+  // TODO(jiacheng): refactor this to support LockResource instead of try-finally
```
ggezer, line 621: This is ideally done earlier than later.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -441,29 +449,49 @@ public void updateUsedBytes(String tierAlias, long usedBytesOnTier) {
...
 
   /**
-   * Returns the lock for worker register status.
-   *
-   * @return a mutex lock
+   * Enum for the locks in {@link MasterWorkerInfo}.
    */
-  public ReentrantLock getRegisterLock() {
-    return mRegisterLock;
+  public enum LockType {
```
ggezer, line 604: Can you move this into its own class, so clients don't need to type encompassing classes?

ggezer (changes requested):
Thanks for the updates. Few more cosmetic requests...

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
ggezer: I see. I think `EnumSet` is neat.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: I get really paranoid when I lock and unlock typing the `EnumSet` twice and fear that they don't match. I believe many people  are like this too.

`LockResource` will then be a lifesaver. I will add that in this PR.

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMetaLock.java
```diff
@@ -0,0 +1,67 @@
...
+        } else {
+          mWorker.getLock(t).writeLock().unlock();
+        }
+      }
+    }
+  }
+
+  @Override
+  public void lockInterruptibly() throws InterruptedException {
+    throw new UnsupportedOperationException();
```
jiacheliu3, line 50: I don't know how to properly implement these and don't think these are needed by the callers at this moment. So instead of inheriting `ReentrantLock` which implements these with internal logics, I'd rather they just throw when the callers are not calling in the way we support.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
ggezer: I see. It looks like we could avoid this if we acquire the block lock outside of process* methods. (Actually it might not be a bad idea to merge them into a single method as register and heartbeat have the same signature.)

However it's too much effort for something that is inevitable under different scenarios as you pointed out.
Please update this TODO with a summary of what you mentioned, or it might not make sense later.

ggezer (changes requested):

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMetaLock.java
```diff
@@ -0,0 +1,67 @@
...
+  private final MasterWorkerInfo mWorker;
+  WorkerMetaLock(EnumSet<WorkerMetaLockType> lockTypes, boolean isShared, MasterWorkerInfo worker) {
+    mLockTypes = lockTypes;
+    mIsShared = isShared;
+    mWorker = worker;
+  }
+
+  @Override
+  public void lock() {
+    for (WorkerMetaLockType t : ImmutableList.of(WorkerMetaLockType.STATUS_LOCK,
```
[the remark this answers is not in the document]
jiacheliu3: On 2nd thought, I'm moving the order into `static` instead of generating new lists every time.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
jiacheliu3: Added TODO to `removeBlocks()`

core/server/master/src/main/java/alluxio/master/block/meta/WorkerMetaLock.java
jiacheliu3: Yea I think `LockResource` supporting multiple locks would be more elegant. Adding TODO there.

ggezer (approved):
Thanks!

ggezer:

PR #13385, opened 2021-05-10
Remove webui unit tests

bradyoo:
There are no webui unit tests.


bradyoo:
@yuzhu This is a fairly easy one. Can you take a look?

gpang:
I think I split up the webui intentionally, because then we don't have to build the webui in some cases. Do we want to force all the runs to build the webui?

bradyoo:
It's a small addition to the beginning of the test for the unit tests and in return you don't need another. The reality is that technically, I don't even need to include webui in the other one because there are 0 webui unit tests and build/checkstyle should cover the case anyways.

gpang:
hrmmm, I see. Yeah, it looks like we don't even need to add it back to the other ones. This should be good then. Thanks!

bradyoo:

PR #13389, opened 2021-05-11
Fix stale raft journal group

bf8086:
For snapshot replication we have been using initial raft group which will be wrong if the quorum has changed. Updated the logic to always get the latest raft group.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -272,6 +272,12 @@ public class JournalStateMachine extends BaseStateMachine {
     mJournalSystem.notifyLeadershipStateChanged(false);
   }
 
+  @Override
+  public void notifyIndexUpdate(long term, long index) {
+    super.notifyIndexUpdate(term, index);
+    CompletableFuture.runAsync(mJournalSystem::updateGroup);
+  }
+
   private long getNextIndex() {
     try {
       return ((RaftServerProxy) mServer).getImpl(mRaftGroupId).getState().getLog().getNextIndex();
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -78,6 +78,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -213,7 +214,7 @@ public class RaftJournalSystem extends AbstractJournalSystem {
    * Should be used for any raft API call that requires a callId.
    **/
   private final ClientId mRawClientId = ClientId.randomId();
-  private RaftGroup mRaftGroup;
+  private RaftGroup mInitialRaftGroup;
   private RaftPeerId mPeerId;
 
   static long nextCallId() {
@@ -294,7 +295,7 @@ public class RaftJournalSystem extends AbstractJournalSystem {
     Set<RaftPeer> peers = addresses.stream()
         .map(addr -> new RaftPeer(RaftJournalUtils.getPeerId(addr), addr))
         .collect(Collectors.toSet());
-    mRaftGroup = RaftGroup.valueOf(RAFT_GROUP_ID, peers);
+    mInitialRaftGroup = RaftGroup.valueOf(RAFT_GROUP_ID, peers);
 
     RaftProperties properties = new RaftProperties();
     Parameters parameters = new Parameters();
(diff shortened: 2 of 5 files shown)


bf8086:
@ggezer @LuQQiu PTAL.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -375,14 +376,32 @@ private synchronized void initServer() throws IOException {
...
         .setParameters(parameters)
         .build();
   }
 
-  private RaftClient createClient() {
+  /**
+   * @return current raft group
+   */
+  @VisibleForTesting
+  public synchronized RaftGroup getCurrentGroup() {
```
LuQQiu, line 390: Is there any callback hook or some ways we can get the current group automatically instead of checking for each gainPrimacy or checkpoint?

LuQQiu (commented):
Thanks for fixing the raft group, left one high level question

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -375,14 +376,32 @@ private synchronized void initServer() throws IOException {
...
+      Preconditions.checkState(group.getGroupId() == RAFT_GROUP_ID,
+          String.format("Invalid group id %s, expecting %s", group.getGroupId(), RAFT_GROUP_ID));
+      return group;
+    } catch (IOException | IllegalStateException e) {
+      LogUtils.warnWithException(LOG, "Failed to get raft group, falling back to initial group", e);
+      return mInitialRaftGroup;
+    }
+  }
+
+  private RaftClient createClient(boolean updateGroup) {
```
ggezer, line 404: Can you add a comment here on what updateGroup do?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -393,7 +412,7 @@ private RaftClient createClient() {
         .setMaxSleepTime(TimeDuration.valueOf(mConf.getElectionTimeoutMs(), TimeUnit.MILLISECONDS))
         .build();
     return RaftClient.newBuilder()
-        .setRaftGroup(mRaftGroup)
+        .setRaftGroup(updateGroup ? getCurrentGroup() : mInitialRaftGroup)
```
ggezer, line 415: Why do we ever want to work with the old quorum info?

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -170,8 +170,38 @@ public synchronized void start() throws Exception {
...
+  }
+
+  /**
+   * Start a number of new master nodes.
+   *
+   * @param count the count of the masters to start
+   * @param format whether to format the master first
+   * @throws Exception if any error occurs
+   */
+  public synchronized void startNewMasters(int count, boolean format) throws Exception {
```
ggezer, line 194: Thank you!

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -701,7 +720,7 @@ private void joinQuorum() {
             .setHost(localAddress.getHostString())
             .setRpcPort(localAddress.getPort()))
         .build();
-    RaftClient client = createClient();
+    RaftClient client = createClient(false);
```
ggezer, line 723: Is this `false` because `getCurrentGroup` will not work until joining?

ggezer (approved):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Yes. The API call will not work before cluster is up and thus a waste of time during startup.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bf8086: Usually during the start up before a raft quorum is formed.

LuQQiu (approved):
Thanks for improving the group info update way! LGTM

bf8086:
Disabling new test which passes in dev environment but fails in GitHub actions when running along with other tests.

bf8086:

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -272,6 +272,12 @@ public void notifyNotLeader(Collection<TransactionContext> pendingEntries) {
     mJournalSystem.notifyLeadershipStateChanged(false);
   }
 
+  @Override
+  public void notifyIndexUpdate(long term, long index) {
+    super.notifyIndexUpdate(term, index);
```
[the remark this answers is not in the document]
jenoudet: If another member joins the quorum in the middle of a term, wouldn't this change prevent the quorum from noticing the new member?
Also, why do you believe there is a need to reduce the update frequency?

PR #13390, opened 2021-05-11
Move block list merging from master to worker

jiacheliu3:
Currently on worker register and heartbeat, the worker will flatten a `Map<BlockStoreLocation, List<Long>>` to a `List<LocationBlockIdListEntry>` and send to the master. `LocationBlockIdListEntry` is a key-value pair `BlockStoreLocationProto:BlockIdList`.  

The problem with this approach is `BlockStoreLocation` is a tuple `<tier,medium,dirIndex>` whereas `BlockStoreLocationProto` is a pair `<tier,medium>`. Basically in map flattening, the `List<LocationBlockIdListEntry>` contains more than one `LocationBlockIdListEntry` with the same `BlockStoreLocationProto:BlockIdList`. When the master receives the list, it will reconstruct the map using `BlockStoreLocationProto` as the key. At this step, there will be key collision because the worker can have more directories with the same `<tier, medium>`. On collision, the master will merge the block lists, which can induce huge list reconstructions.

This change ensures the block lists of the same `<tier,medium>` (from different directories) are merged before sending to the master.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
@@ -15,6 +15,7 @@ import alluxio.RpcUtils;
 import alluxio.grpc.BlockHeartbeatPRequest;
 import alluxio.grpc.BlockHeartbeatPResponse;
 import alluxio.grpc.BlockMasterWorkerServiceGrpc;
+import alluxio.grpc.BlockStoreLocationProto;
 import alluxio.grpc.CommitBlockInUfsPRequest;
 import alluxio.grpc.CommitBlockInUfsPResponse;
 import alluxio.grpc.CommitBlockPRequest;
@@ -174,12 +175,18 @@ public final class BlockMasterWorkerServiceHandler extends
                  * Therefore we just fail on merging.
                  */
                 (e1, e2) -> {
-                  throw new AssertionError(
-                      String.format(
-                          "One registerWorker request contains two block id lists for the "
-                                  + "same BlockLocation.%n"
-                                  + "Existing: %s%n New: %s",
-                          e1, e2));
+//                  System.out.println(entries);
+//                  LocationBlockIdListEntry k1 = entries.get(0);
+//                  LocationBlockIdListEntry k2 = entries.get(1);
+//                  BlockStoreLocationProto p1 = k1.getKey();
+//                  throw new AssertionError(
+//                      String.format(
+//                          "One registerWorker request contains two block id lists for the "
+//                                  + "same BlockLocation.%n"
+//                                  + "Existing: %s%n New: %s",
+//                          e1, e2));
+                  e1.addAll(e2);
+                  return e1;
                 }));
   }
 }
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
@@ -22,6 +22,7 @@ import alluxio.grpc.CommitBlockPResponse;
 import alluxio.grpc.GetWorkerIdPRequest;
 import alluxio.grpc.GetWorkerIdPResponse;
 import alluxio.grpc.GrpcUtils;
+import alluxio.grpc.LocationBlockIdListEntry;
 import alluxio.grpc.RegisterWorkerPOptions;
 import alluxio.grpc.RegisterWorkerPRequest;
 import alluxio.grpc.RegisterWorkerPResponse;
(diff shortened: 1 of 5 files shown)


core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -84,10 +84,14 @@ public void blockHeartbeat(BlockHeartbeatPRequest request,
...
+                     * The merger function is invoked on key collisions to merge the values.
+                     * In fact this merger should never be invoked because the list is deduplicated
+                     * by {@link BlockMasterClient#heartbeat} before sending to the master.
+                     */
                     (e1, e2) -> {
-                      List<Long> e3 = new ArrayList<>(e1);
-                      e3.addAll(e2);
-                      return e3;
+                      e1.addAll(e2);
+                      return e1;
```
jiacheliu3, line 94: Although this is currently not invoked at all, I think the small improvement is needed to guard future changes. There's no need to create a new list and copy, we can just use the existing list.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -84,10 +84,14 @@ public void blockHeartbeat(BlockHeartbeatPRequest request,
                     e -> Block.BlockLocation.newBuilder().setTier(e.getKey().getTierAlias())
                         .setMediumType(e.getKey().getMediumType()).build(),
                     e -> e.getValue().getBlockIdList(),
+                    /**
+                     * The merger function is invoked on key collisions to merge the values.
+                     * In fact this merger should never be invoked because the list is deduplicated
+                     * by {@link BlockMasterClient#heartbeat} before sending to the master.
+                     */
```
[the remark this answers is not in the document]
jiacheliu3: I'm just trying to add some comments here so the caveat can be seen.

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
```diff
@@ -139,6 +139,10 @@ public long getId(final WorkerNetAddress address) throws IOException {
     }, LOG, "GetId", "address=%s", address);
   }
 
+  /**
+   * Converts the block list map to proto.
+   * The output list will not contain duplicate keys.
+   * */
```
gpang, line 145: I think this comment line looks off?

gpang (commented):
@jiacheliu3 just a minor comment.

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
jiacheliu3: Well yeah it looks a bit strange where a "list" has "keys". I'm just trying to say that the map is flattened into a list so when the list is converted to a map on the other side, there will be no key collision, and this is the guaranteed behavior.

jiacheliu3: Well yeah it looks a bit strange where a "list" has "keys". I'm just trying to say that the map is flattened into a list so when the list is converted to a map on the other side, there will be no key collision, and this is the guaranteed behavior.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -84,10 +84,14 @@ public void blockHeartbeat(BlockHeartbeatPRequest request,
...
                     (e1, e2) -> {
-                      List<Long> e3 = new ArrayList<>(e1);
-                      e3.addAll(e2);
-                      return e3;
+                      e1.addAll(e2);
```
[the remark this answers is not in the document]
jiacheliu3: @ggezer Added a test, PTAL, thanks!

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
jiacheliu3: @gpang I tried to rephrase, does it make more sense now?

jiacheliu3:
Paused for the review because I found this is not true. We are using the merger when there are multiple dirs on one tier.

jiacheliu3:
@ggezer Updated the worker side logic to ensure deduplication and added unit tests. PTAL, thanks!

ggezer (approved):

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
ggezer: This looks OK now.

jiacheliu3:

PR #13397, opened 2021-05-12
Add additional metrics throughout Alluxio system

yuzhu:


yuzhu:
Still adding a couple more, but the majority is here. 
Will also add test coverage for the new metrics. 

@bf8086 could you take a look at the journal related metrics? 
@ggezer could you take a look at the others?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -541,6 +548,7 @@ public synchronized void checkpoint() throws IOException {
       mSnapshotAllowed.set(true);
       catchUp(mStateMachine, client);
       mStateMachine.takeLocalSnapshot();
+      mLastCheckPointTime = System.currentTimeMillis();
```
ZacBlanco, line 551: This is the external checkpoint mechanism. If you to record snapshots, you'll need to move this into the `JournalStateMachine#takeSnapshot` function. That function isn't called anywhere internally (Alluxio code) but is called externally by the ratis library when the number of journal entries exceeds the Alluxio PK `MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES`

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -187,6 +198,7 @@ public URI getLocation() {
   @VisibleForTesting
   synchronized void write(JournalEntry entry) throws IOException, JournalClosedException {
     writer().write(entry);
+    mEntriesSinceLastCheckPoint++;
```
ZacBlanco, line 201: will this every need to be an atomic increment? (It probably doesn't matter that much?)

ZacBlanco (changes requested):

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
```diff
@@ -116,6 +118,9 @@
   /** Used to stop catching up when cancellation requested.  */
   private volatile boolean mStopCatchingUp = false;
 
+  private long mLastCheckPointTime = -1;
+  private long mEntriesSinceLastCheckPoint = 0;
```
bf8086, line 122: We might be able to reuse checkpoint stats stored in `UfsJournalCheckpointThread`.

bf8086 (commented):
Thanks for the improvement! Journal metrics mostly LGTM.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
yuzhu: i guess for the ticket I was working on, it was specifically looking for metrics on the checkpoint. maybe we also need it for snapshots?

core/server/common/src/main/java/alluxio/master/journal/ufs/UfsJournal.java
yuzhu: i am guessing not since the method is synchronized?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ZacBlanco: From my understanding of the terminology, checkpoint is the rough equivalent of snapshot? In this implementation, we would only get the metrics for when a user explicitly initiated the checkpoint (snapshot) with `alluxio fsadmin journal checkpoint` or something like that.

The master also takes automatic checkpoints (snapshots) at specified intervals determined by the Alluxio PK mentioned earlier. The snapshots initiated by the automatic threshold mechanism won't have the metrics recorded properly if we only record the timing here

ZacBlanco (approved):
LGTM.

FYI if you try testing some snapshot metrics on a local master I found a bug earlier filed as #13417

core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -109,6 +109,7 @@ private RpcUtils() {} // prevent instantiation
     // avoid string format for better performance if debug is off
     String debugDesc = logger.isDebugEnabled() ? String.format(description, args) : null;
     try (Timer.Context ctx = MetricsSystem.timer(getQualifiedMetricName(methodName)).time()) {
+      MetricsSystem.counter(getQualifiedInProgressMetricName(methodName)).inc();
```
ggezer, line 112: nit: how about moving timer and counter calls together above and close them together on finally below?

core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -187,6 +193,10 @@ private static String getQualifiedFailureMetricName(String methodName) {
     return getQualifiedMetricNameInternal(methodName + "Failures");
   }
 
+  private static String getQualifiedInProgressMetricName(String methodName) {
```
ggezer, line 196: Can you update `getQualifiedMetricNameInternal` to accept `String...components` to attach to the name instead?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -941,6 +945,7 @@ public Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnT
       worker.updateLastUpdatedTimeMs();
 
       List<Long> toRemoveBlocks = worker.getToRemoveBlocks();
+      Metrics.TOTAL_BLOCKS.inc(addedBlocks.size() - removedBlockIds.size());
```
ggezer, line 946: Are we accounting for lost workers somewhere?

core/server/common/src/main/java/alluxio/RpcUtils.java
yuzhu: then we run into situation where ctx might not be initialized before close is called?. i think try-with-resources is the standard way of doing this, and we are not really closing counter right?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
yuzhu: good point, fixed

core/server/common/src/main/java/alluxio/RpcUtils.java
yuzhu: done

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
ggezer: In addition to this, should we do subtract when a worker gets lost?

ggezer (approved):

LuQQiu:
@yuzhu Please fix the merge conflict, thanks

yuzhu:

PR #13406, opened 2021-05-13
Support read(Bytebuffer buf) method in LocalCacheFileInStream

neuyilan:
1. HDFS supports the NIO  read (bytebyffer buffer) function, and we'd better be compatible with HDFS API.
2. The `libhdfs.c `client  uses the read (bytebyffer buffer) method to access the HDFS data[1], "**By using Java DirectByteBuffers we can avoid copying the bytes onto the Java heap.  Instead the data will be directly copied from kernel space to the C heap"**. 
[1] https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfs/hdfs.c#L50


LuQQiu:
@neuyilan Can you add a description of why this change is needed?

neuyilan:
> @neuyilan Can you add a description of why this change is needed?

I have added the description above, PTAL

core/client/fs/src/main/java/alluxio/client/ByteBufferReadable.java
```diff
@@ -0,0 +1,38 @@
...
+ * Implementers of this interface provide a read API that writes to a ByteBuffer, not a byte[].
+ */
+public interface ByteBufferReadable {
+
+  /**
+   * Reads up to buf.remaining() bytes into buf. Callers should use buf.limit(..) to control the
+   * size of the desired read.
+   * <p/>
+   * After a successful call, buf.position() and buf.limit() should be unchanged, and therefore any
+   * data can be immediately read from buf. buf.mark() may be cleared or updated.
```
apc999, line 26: this part is different from Hadoop `ByteBufferReadable` (i.e., `After a successful call, {@code buf.position()} will be advanced by the  number of bytes read and {@code buf.limit()} will be unchanged.`) , any reason for that?

core/client/fs/src/main/java/alluxio/client/ByteBufferReadable.java
```diff
@@ -0,0 +1,38 @@
...
+   * In the case of an exception, the values of buf.position() and buf.limit() are undefined, and
+   * callers should be prepared to recover from this eventuality.
+   * <p/>
+   *
+   * @param buf the ByteBuffer to receive the results of the read operation. Up to buf.limit() -
+   *            buf.position() bytes may be read.
+   * @return the number of bytes available to read from buf
+   * @throws IOException if there is some error performing the read
+   */
+  int read(ByteBuffer buf) throws IOException;
```
apc999, line 37: you can leave a default implementation here, like

```suggestion
default int read(ByteBuffer buf) throws IOException {
  throw new UnsupportedOperationException("read(ByteBuffer buf) not implemented");
}
```

Then you don't need to implement individuals in other classes

core/client/fs/src/main/java/alluxio/client/ByteBufferReadable.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package alluxio.client;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * Implementers of this interface provide a read API that writes to a ByteBuffer, not a byte[].
+ */
+public interface ByteBufferReadable {
```
apc999, line 19: Check `alluxio.client.file.FileInStream` where we already define and provided the default implementation of `  public int read(ByteBuffer byteBuffer, int off, int len) throws IOException`.

You can either move this method there (and remove interface `ByteBufferReadable`), or move that methd here. I don't have a strong opinion. But I do think they should be put together

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -346,4 +347,80 @@ private static void registerGauges() {
...
+          mCacheManager.put(pageId, page, mCacheScope, mCacheQuota);
+        }
+      }
+    }
+    if (totalBytesRead > len || (totalBytesRead < len && remaining() > 0)) {
+      throw new IOException(String.format("Invalid number of bytes read - "
+              + "bytes to read = %d, actual bytes read = %d, bytes remains in file %d",
+          len, totalBytesRead, remaining()));
+    }
+    buf.put(b, off, len);
```
apc999, line 423: it looks like `buf.position()` will be changed after a successful call, right?

apc999 (changes requested):
Left some questions. I think this can be a very useful optimization and let's iterate on this.

Thanks!

core/client/fs/src/main/java/alluxio/client/ByteBufferReadable.java
neuyilan: Maybe it's written by mistake. In fact, when the call is successful, `buf.position()` will change, I have altered the comment

core/client/fs/src/main/java/alluxio/client/ByteBufferReadable.java
neuyilan: Sure, Thanks.

core/client/fs/src/main/java/alluxio/client/ByteBufferReadable.java
neuyilan: Yeah, as there already one `public int read(ByteBuffer byteBuffer, int off, int len) throws IOException` method in `alluxio.client.file.FileInStream`, so I think it's ok to move the interface's logic to `alluxio.client.file.FileInStream` class.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
neuyilan: Yes :)

neuyilan:
> Left some questions. I think this can be a very useful optimization and let's iterate on this.
> 
> Thanks!

Thank you for your review. I have fixed some of your comments, sorted out the redundant code, and added some UTs, PTAL.

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -287,19 +265,41 @@ private FileInStream getExternalFileInStream(long pos) throws IOException {
...
+        page = readExternalPageIntoByteBuffer(pos);
+        break;
+      default:
+        throw new IOException("unsupported read type = " + readType);
+    }
+    return page;
+  }
+
+  /**
+   * Using read(byte b[], int off, int len) method to read data from external storage which contains
```
LuQQiu, line 296: Using -> Uses

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -320,6 +320,46 @@ private FileInStream getExternalFileInStream(long pos) throws IOException {
     return page;
   }
 
+  /**
+   * Using read(ByteBuffer buf) method to read data from external storage which contains the
```
LuQQiu, line 324: Using -> Uses

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -346,4 +386,16 @@ private static void registerGauges() {
           });
     }
   }
+
```
LuQQiu, line 389: Remove extra line

LuQQiu (commented):
left some minor codestyle suggestions

neuyilan (commented):
> left some minor codestyle suggestions

Thanks @LuQQiu, I found that we have our code style file in `build/checkstyle/alluxio_checks.xml`, but those may still not check the indentations format problems like [1], and every code developer may have its own code styles, if all the checks need to be protected by the person who reviews the code, there will inevitably be loopholes. So do we considering using automatic code checking and code collation like spotless[2]?
It's just a discussion, any comments are welcome!

[1] https://github.com/Alluxio/alluxio/pull/13406/files/61f6121d395af64bcc1088bcd6cbbeaef33164ad#diff-ab09d305553e29cf84278bdb023cb71470918975d4ac0f725084e38bb0c8a317
[2]https://github.com/diffplug/spotless

calvinjia (commented):
Would it be possible to check for the preferred API and use that if available instead of having a switch based on readtype?

calvinjia (approved):
Discussed offline - this PR is just to add the interface so applications can start integrating, optimization of the interface with the streams underneath (ie. external stream) will happen later.

apc999 (approved):

apc999:

PR #13407, opened 2021-05-13
Remove cephfs ufs related code temporarily

maobaolong:
@LuQQiu Sorry for missing this batch of code need to be removed

### DIFF
--- a/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
+++ b/underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
@@ -762,13 +762,15 @@ public class HdfsUnderFileSystem extends ConsistentUnderFileSystem
   private FileStatus[] listStatusInternal(String path) throws IOException {
     FileStatus[] files;
     FileSystem hdfs = getFs();
+    Path thePath = new Path(path);
     try {
-      files = hdfs.listStatus(new Path(path));
+      files = hdfs.listStatus(thePath);
     } catch (FileNotFoundException e) {
       return null;
     }
     // Check if path is a file
-    if (files != null && files.length == 1 && files[0].getPath().toString().equals(path)) {
+    if (files != null && files.length == 1
+        && files[0].getPath().toUri().getPath().equals(thePath.toUri().getPath())) {
       return null;
     }
     return files;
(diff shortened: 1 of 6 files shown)


underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
```diff
@@ -762,13 +762,15 @@ private boolean delete(String path, boolean recursive) throws IOException {
...
     try {
-      files = hdfs.listStatus(new Path(path));
+      files = hdfs.listStatus(thePath);
     } catch (FileNotFoundException e) {
       return null;
     }
     // Check if path is a file
-    if (files != null && files.length == 1 && files[0].getPath().toString().equals(path)) {
+    if (files != null && files.length == 1
+        && files[0].getPath().toUri().getPath().equals(thePath.toUri().getPath())) {
```
maobaolong, line 773: @LuQQiu This is the bad condition which make the listStatus  & listRecursive failed, when I use `ceph:///` as the path string, i think the original purpose of this condition suppose to verify if path is a file, so we can using the <PATH> of `<SCHEME>://<AUTHORITY>/<PATH>` to check if the result is the same as the given path.

Luckily, I can work around to run the `runUfsTests` successfully if I use `ceph://<MON_IP_LIST>` instead.

Another hand, the bad root cause is `DistributedFileSystem` `AliyunOSSFileSystem` `CosFileSystem` `CephFileSystem` and many other  HCFS implementation, they implemented the `initialize` method like this

```java
  public void initialize(URI name, Configuration conf) throws IOException {
    ...
    uri = java.net.URI.create(name.getScheme() + "://" + name.getAuthority());
    ...
  }
```

When the uri is `ceph:///` `hdfs:///` `cosn:///` or `oss:///`, it means that the authority is null, so the `name.getAuthority()` will be `"null"`,  the result will be `ceph://null/` `hdfs://null/` `cosn://null/` or `oss://null/`.

To correct these HCFS implementations will be a long term purpose, but we can still make the file verification condition simple and valid like the changes of this part.

maobaolong:
@LuQQiu After some debug effort, we found the behavior of CephFilesystem#listStatus didn't respect the HCFS.

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
LuQQiu: @maobaolong Great findings! Could you help create new-contributor-tasks for supporting ceph:/// format in other UFSes?

LuQQiu:
@maobaolong I don't think this PR is needed, let's keep it easy to include all Cephfs and Ceph-hadoop changes in one PR

maobaolong:
@LuQQiu OK, will close this PR.

Issue #13411, opened 2021-05-14
Passive TTL on files does not take effect


BUPTAnderson:
Alluxio Version:
2.3.0 2.4.0 2.5.0

Describe the bug
Passive TTL on files in Alluxio [https://docs.alluxio.io/ee/user/stable/en/core-services/Caching.html?q=ttl#passive-ttl-on-files-in-alluxio](url).
when i adds a new file to the Alluxio namespace with configuration options: `alluxio.user.file.create.ttl` and `alluxio.user.file.create.ttl.action`, Passive TTL does not work.
for example:
list a new file
```
bin/alluxio fs ls -Dalluxio.user.file.create.ttl=12h -Dalluxio.user.file.create.ttl.action=FREE /data/LICENSE.txt
-rw-r--r--  anderson       supergroup              150569       PERSISTED 04-01-2021 21:30:33:931   0% /data/LICENSE.txt

bin/alluxio fs stat /data/LICENSE.txt
/data/LICENSE.txt is a file path.
FileInfo{fileId=33554431, fileIdentifier=null, name=LICENSE.txt, path=/data/LICENSE.txt, ufsPath=hdfs://localhost:9000/data/LICENSE.txt, length=150569, blockSizeBytes=134217728, creationTimeMs=1620920757297, completed=true, folder=false, pinned=false, pinnedlocation=[], cacheable=true, persisted=true, blockIds=[16777216], inMemoryPercentage=0, lastModificationTimesMs=1617283833931, ttl=-1, lastAccessTimesMs=1617283833931, ttlAction=DELETE, owner=anderson, group=supergroup, mode=420, persistenceState=PERSISTED, mountPoint=false, replicationMax=-1, replicationMin=0, fileBlockInfos=[FileBlockInfo{blockInfo=BlockInfo{id=16777216, length=150569, locations=[]}, offset=0, ufsLocations=[]}], mountId=4303832726592734076, inAlluxioPercentage=0, ufsFingerprint=TYPE|FILE UFS|hdfs OWNER|anderson GROUP|supergroup MODE|420 CONTENT_HASH|(len:150569,_modtime:1617283833931) , acl=user::rw-,group::r--,other::r--, defaultAcl=, cacheQuota=null, cacheScope=null}
Containing the following blocks:
BlockInfo{id=16777216, length=150569, locations=[]}
```
read a new file:
```
[...]

PR #13412, opened 2021-05-14
Fix Passive TTL on files does not take effect #13411

BUPTAnderson:
Fix Passive TTL on files does not take effect #13411.

When client exec stat/list/read operations first time, make sure alluxio.user.file.create.ttl and alluxio.user.file.create.ttl.action options from client take effect.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -137,11 +137,6 @@ public class InodeSyncStream {
 
   private static final Logger LOG = LoggerFactory.getLogger(InodeSyncStream.class);
 
-  private static final FileSystemMasterCommonPOptions NO_TTL_OPTION =
-      FileSystemMasterCommonPOptions.newBuilder()
-          .setTtl(-1)
-          .build();
-
   /** The root path. Should be locked with a write lock. */
   private final LockingScheme mRootScheme;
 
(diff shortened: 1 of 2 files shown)


BUPTAnderson:

BUPTAnderson:
@LuQQiu @ggezer please, review

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -720,7 +715,7 @@ private void loadMetadataForPath(LockedInodePath inodePath)
     }
     LoadMetadataContext ctx = LoadMetadataContext.mergeFrom(
         LoadMetadataPOptions.newBuilder()
-            .setCommonOptions(NO_TTL_OPTION)
+            .setCommonOptions(mSyncOptions)
```
ZacBlanco, line 718: I don't think this is the correct way to go about implementing this

This will cause files loaded from `ls` and other operations you listed to gain the TTL for files loaded into alluxio that may have already been created outside of alluxio. Even if that is the desired behavior, it breaks backwards compatbility as far as I am aware.

The file TTL options are only supposed to take effect for files created through alluxio. hence alluxio.user.file.**create**.ttl. Using the TTL from client options is something that should be discussed amongst some other maintainers and would need to be configurable behavior via _another_ configuration. Otherwise, it should wait until a 3.0 release.

When we re-wrote a bunch of the sync code, the behavior you're trying to implement here actually appeared and was labeled as a bug (and then subsequently fixed), so I think this PR needs further discussion.

ZacBlanco: @LuQQiu FYI

BUPTAnderson: @ZacBlanco @LuQQiu In other words, TTL only valid when we create new file in alluxio. My usage scenario is in Ad-hoc,  the client only care about data reading, the data automatically expires when the alluxio.user.file.create.ttl is set. @LuQQiu  What do you think？

LuQQiu: I agree that `create ttl` should be only valid when files are created from Alluxio namespace.
i am not quite sure whether we have something like `load ttl` and whether we want to have something like `load ttl`? @ZacBlanco  do you aware of such things?

PR #13375, opened 2021-05-08
Improve DefaultBlockWorkerTest

LuQQiu:

PR #13419, opened 2021-05-15
[SMALLFIX] Add description in method and add some UT in storeMeta

yabola:
1. add some description on methond (`getBlockList()` and `getBlockListByStorageLocation` ) in `BlockStoreMeta` to make it more clearer.
2. add some UT for `BlockStoreMeta`

fix
 https://github.com/Alluxio/new-contributor-tasks/issues/588
 https://github.com/Alluxio/new-contributor-tasks/issues/589

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
@@ -61,6 +61,11 @@ public final class DefaultBlockStoreMeta implements BlockStoreMeta {
 
   @Override
   public Map<String, List<Long>> getBlockList() {
+    /**
+     * This method is only available when using method
+     * `DefaultBlockStoreMeta(manager, true)` to init, otherwise `mBlockIdsOnTiers`
+     * will not be initialized and be `NULL` value.
+     */
     Preconditions.checkNotNull(mBlockIdsOnTiers, "mBlockIdsOnTiers");
 
     return mBlockIdsOnTiers;
@@ -68,6 +73,11 @@ public final class DefaultBlockStoreMeta implements BlockStoreMeta {
 
   @Override
   public Map<BlockStoreLocation, List<Long>> getBlockListByStorageLocation() {
+    /**
+     * This method is only available when using method
+     * `DefaultBlockStoreMeta(manager, true)` to init, otherwise `mBlockIdsOnLocations`
+     * will not be initialized and be `NULL` value.
+     */
     Preconditions.checkNotNull(mBlockIdsOnLocations, "mBlockIdsOnLocations");
 
     return mBlockIdsOnLocations;
--- a/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
@@ -61,11 +61,6 @@ public final class DefaultBlockStoreMeta implements BlockStoreMeta {
 
   @Override
   public Map<String, List<Long>> getBlockList() {
-    /**
-     * This method is only available when using method
-     * `DefaultBlockStoreMeta(manager, true)` to init, otherwise `mBlockIdsOnTiers`
-     * will not be initialized and be `NULL` value.
-     */
     Preconditions.checkNotNull(mBlockIdsOnTiers, "mBlockIdsOnTiers");
 
     return mBlockIdsOnTiers;
@@ -73,11 +68,6 @@ public final class DefaultBlockStoreMeta implements BlockStoreMeta {
 
   @Override
   public Map<BlockStoreLocation, List<Long>> getBlockListByStorageLocation() {
-    /**
-     * This method is only available when using method
-     * `DefaultBlockStoreMeta(manager, true)` to init, otherwise `mBlockIdsOnLocations`
-     * will not be initialized and be `NULL` value.
-     */
     Preconditions.checkNotNull(mBlockIdsOnLocations, "mBlockIdsOnLocations");
 
     return mBlockIdsOnLocations;
--- a/core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
+++ b/core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
@@ -176,9 +176,15 @@ public class DefaultBlockWorkerTest {
   public void createBlock() throws Exception {
     long blockId = mRandom.nextLong();
     long sessionId = mRandom.nextLong();
-    long initialBytes = 1;
+    long initialBytes = 1L;
     String path = mBlockWorker.createBlock(sessionId, blockId, 0, "", initialBytes);
-    assertTrue(path.startsWith(mMemDir)); // tier 0 is mem
+    assertTrue(path.startsWith(mMemDir));
+
+    long blockId2 = mRandom.nextLong();
+    long initialBytes2 = 1L * Constants.GB;
+    String path2 = mBlockWorker.createBlock(sessionId, blockId2, 0, "", initialBytes2);
+    // The available space in MEM is not enough. Allocator will allocate in HDD.
+    assertTrue(path2.startsWith(mHddDir));
   }
 
   @Test
@@ -211,11 +217,26 @@ public class DefaultBlockWorkerTest {
   }
 
   @Test
-  public void getStoreMeta() {
+  public void getStoreMeta() throws Exception {
     mBlockWorker.getStoreMeta();
-    mBlockWorker.getStoreMetaFull();
     verify(mBlockStore, times(2)).getBlockStoreMeta(); // 1 is called at metrics registration
-    verify(mBlockStore).getBlockStoreMetaFull();
+    long blockId1 = mRandom.nextLong();
+    long blockId2 = mRandom.nextLong();
+    long sessionId = mRandom.nextLong();
+    mBlockWorker.createBlock(sessionId, blockId1, 0, "", 1L);
+    // Available space in MEM is not enough. Block will be allocated in HDD.
+    mBlockWorker.createBlock(sessionId, blockId2, 0, "", 2L * Constants.GB);
+    BlockStoreMeta storeMeta = mBlockWorker.getStoreMetaFull();
+    assertEquals(2, storeMeta.getBlockList().size());
+    assertEquals(2, storeMeta.getBlockListByStorageLocation().size());
+    assertEquals(0, storeMeta.getBlockList().get("MEM").size());
+    assertEquals(3L * Constants.GB, storeMeta.getCapacityBytes());
+    assertEquals(1L + 2L * Constants.GB, storeMeta.getUsedBytes());
+    mBlockWorker.commitBlock(sessionId, blockId1, true);
+    mBlockWorker.commitBlock(sessionId, blockId2, true);
+    storeMeta = mBlockWorker.getStoreMetaFull();
+    assertEquals(1, storeMeta.getBlockList().get("MEM").size());
+    assertEquals(1, storeMeta.getBlockList().get("HDD").size());
   }
 
   @Test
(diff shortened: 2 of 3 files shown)


core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -90,6 +90,7 @@
           .put(PropertyKey.WORKER_TIERED_STORE_LEVEL1_DIRS_QUOTA, "2GB")
           .put(PropertyKey.WORKER_TIERED_STORE_LEVEL1_DIRS_PATH, mHddDir)
           .put(PropertyKey.WORKER_RPC_PORT, "0")
+          .put(PropertyKey.WORKER_MANAGEMENT_TIER_ALIGN_RESERVED_BYTES, "512MB")
```
yabola, line 93: The default value of 1024M is not obvious enough , which will affect the allocation of blocks

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -212,11 +213,28 @@ public void getReport() {
...
+    assertEquals(2, storeMeta.getBlockListByStorageLocation().size());
+    assertEquals(0, storeMeta.getBlockList().get("MEM").size());
+    // 1GB (`MEM`) + 2GB (`HDD`)
+    assertEquals(3221225472L, storeMeta.getCapacityBytes());
+    // 512MB * 2 (reserved space) + 1bit (`MEM`) + 1GB (`HDD`)
+    assertEquals(2147483649L, storeMeta.getUsedBytes());
+    mBlockWorker.commitBlock(sessionId, blockId1, true);
+    mBlockWorker.commitBlock(sessionId, blockId2, true);
+    storeMeta = mBlockWorker.getStoreMetaFull();
+    assertEquals(1, storeMeta.getBlockList().get("MEM").size());
```
yabola, line 236: we can't assert usedBytes after commit block cause the real path size is 0 in UT.
see `BlockMetadataManager#commitTempBlockMeta()` -> `dir.addBlockMeta(block) // block size is 0 here`

yabola, line 236: we can't assert usedBytes after commit block cause the real path size is 0 in UT.
see BlockMetadataManager#commitTempBlockMeta() -> dir.addBlockMeta(block) // block size is 0 here

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola, line 93: The default value of 1024M is not obvious enough , which will affect the allocation of blocks

yabola:
@LuQQiu please take a look~

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
```diff
@@ -61,13 +61,23 @@
 
   @Override
   public Map<String, List<Long>> getBlockList() {
```
LuQQiu, line 63: @yabola it's always to put the comment in the getBlockList() method declaration. That's where people will usually view to understand the method. And the comment should be implementation independent and as a contract between caller and implementor.
For example, you can change the BlockStoreMeta.getBlockList description:
```
This method is only available when users initialize the blockStoreMeta with full block meta requirement
```

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
```diff
@@ -61,13 +61,23 @@
...
+     * will not be initialized and be `NULL` value.
+     */
     Preconditions.checkNotNull(mBlockIdsOnTiers, "mBlockIdsOnTiers");
 
     return mBlockIdsOnTiers;
   }
 
   @Override
   public Map<BlockStoreLocation, List<Long>> getBlockListByStorageLocation() {
+    /**
```
LuQQiu, line 76: Same for this method description

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -180,9 +181,9 @@ public void commitBlockOnRetry() throws Exception {
   public void createBlock() throws Exception {
     long blockId = mRandom.nextLong();
     long sessionId = mRandom.nextLong();
-    long initialBytes = 1;
+    long initialBytes = 1024 * 1024 * 1024;
```
LuQQiu, line 184: Can you use Constants.MB? Can we not initialize a large block store  e.g. https://github.com/Alluxio/alluxio/pull/13375?

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -212,11 +213,28 @@ public void getReport() {
...
+  public void getStoreMeta() throws Exception {
     mBlockWorker.getStoreMeta();
-    mBlockWorker.getStoreMetaFull();
     verify(mBlockStore, times(2)).getBlockStoreMeta(); // 1 is called at metrics registration
-    verify(mBlockStore).getBlockStoreMetaFull();
+    long blockId1 = mRandom.nextLong();
+    long blockId2 = mRandom.nextLong();
+    long sessionId = mRandom.nextLong();
+    mBlockWorker.createBlock(sessionId, blockId1, 0, "", 1L);
+    // Available space in `MEM` is not enough. Block will be allocated in `HDD`.
```
LuQQiu, line 223: That's not true. It's related to the `put(PropertyKey.WORKER_MANAGEMENT_TIER_ALIGN_RESERVED_BYTES, "0")` which  if reserved bytes is bigger than mem tier size, the blocks will not be allocated in the mem tier

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -212,11 +213,28 @@ public void getReport() {
...
+    mBlockWorker.createBlock(sessionId, blockId2, 0, "", 1024 * 1024 * 1024L);
+    BlockStoreMeta storeMeta = mBlockWorker.getStoreMetaFull();
+    assertEquals(2, storeMeta.getBlockList().size());
+    assertEquals(2, storeMeta.getBlockListByStorageLocation().size());
+    assertEquals(0, storeMeta.getBlockList().get("MEM").size());
+    // 1GB (`MEM`) + 2GB (`HDD`)
+    assertEquals(3221225472L, storeMeta.getCapacityBytes());
```
LuQQiu, line 230: Please use some easy to understanding values, e.g.Constants.GB * 3

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -212,11 +213,28 @@ public void getReport() {
...
+    // 512MB * 2 (reserved space) + 1bit (`MEM`) + 1GB (`HDD`)
+    assertEquals(2147483649L, storeMeta.getUsedBytes());
```
LuQQiu, line 232: Same here and remove the comment

LuQQiu (commented):
Thanks for improving the comments and tests, left some comments

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockStoreMeta.java
yabola: done

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola: done

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola: done

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola: I check in `DefaultStorageDir` -> `mAvailableBytes = capacityBytes - reservedBytes` ,
available space means that the reserved space is removed. allocator will judge whether there is enough space based on AvailableBytes. so i think this description is right...

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -211,11 +217,26 @@ public void getReport() {
   }
 
   @Test
-  public void getStoreMeta() {
+  public void getStoreMeta() throws Exception {
     mBlockWorker.getStoreMeta();
-    mBlockWorker.getStoreMetaFull();
     verify(mBlockStore, times(2)).getBlockStoreMeta(); // 1 is called at metrics registration
```
LuQQiu, line 222: No need to do in this PR, but will be better if you can do the similar changes against getStoreMeta just like what you did for `getStoreMetaFull`

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -176,9 +176,15 @@ public void commitBlockOnRetry() throws Exception {
...
     long blockId = mRandom.nextLong();
     long sessionId = mRandom.nextLong();
-    long initialBytes = 1;
+    long initialBytes = 1L;
     String path = mBlockWorker.createBlock(sessionId, blockId, 0, "", initialBytes);
-    assertTrue(path.startsWith(mMemDir)); // tier 0 is mem
+    assertTrue(path.startsWith(mMemDir));
+
+    long blockId2 = mRandom.nextLong();
+    long initialBytes2 = 1L * Constants.GB;
```
LuQQiu, line 184: In this test, try not to set a large initial bytes number. This test will actually create a block store. With the large initial bytes, the worker will actually try to allocate 1GB or 2GB in the system. When this test is running in a env with little disk space available, this test may be failed.

All the unit tests are better to be designed as light-weighted tests

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -211,11 +217,26 @@ public void getReport() {
...
     mBlockWorker.getStoreMeta();
-    mBlockWorker.getStoreMetaFull();
     verify(mBlockStore, times(2)).getBlockStoreMeta(); // 1 is called at metrics registration
-    verify(mBlockStore).getBlockStoreMetaFull();
+    long blockId1 = mRandom.nextLong();
+    long blockId2 = mRandom.nextLong();
+    long sessionId = mRandom.nextLong();
+    mBlockWorker.createBlock(sessionId, blockId1, 0, "", 1L);
+    // Available space in MEM is not enough. Block will be allocated in HDD.
+    mBlockWorker.createBlock(sessionId, blockId2, 0, "", 2L * Constants.GB);
```
LuQQiu, line 228: Same, don't set initial size as a large number

LuQQiu (changes requested):
Thanks for improving the description and docs! 

Can you modify the PR description to include the main points of what you did in this PR?
The PR description will be included in the log commit message for other people to know what's the content for the commit.

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola: done

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola: done

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
yabola: done

LuQQiu (approved):
LGTM, thanks for improving the description and the unit tests!

LuQQiu:

PR #13424, opened 2021-05-15
Add a test option to accept running test name

maobaolong:
Provide an option to help debug a specific test when running runUfsTests
For example, we can pass `--test` to focus test only `test` and `test2`

```
alluxio runUfsTests --test test1 --test test2
```

### DIFF
--- a/integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
+++ b/integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
@@ -60,7 +60,7 @@ public final class UnderFileSystemContractTest {
   private String mUfsPath;
 
   @Parameter(names = {"--test"}, required = false,
-      description = "Test name, use this option multiply time to run multiply test")
+      description = "Test name, this option can be passed multiple times to indicate multiply tests")
   private List<String> mTestList = new ArrayList<>();
 
   @Parameter(names = {"--help"}, help = true)
--- a/integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
+++ b/integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
@@ -35,6 +35,8 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


maobaolong:
@LuQQiu I create this PR aim to focus on debug a specific test when running `runUfsTests`, please take a look, thank you.

integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -57,6 +59,10 @@
       description = "The under filesystem path to run tests against.")
   private String mUfsPath;
 
+  @Parameter(names = {"--test"}, required = false,
+      description = "Test name, use this option multiply time to run multiply test")
```
apc999, line 63: `use this option multiply time` what does this mean?

maobaolong: It mean that, we can use it something like this,
```
alluxio runUfsTests --test test1 --test test2 --test test3
```

apc999, line 63:
```suggestion
      description = "Test name, this option can be passed multiple times to indicate multiply tests")
```

apc999: multiply -> multiple

maobaolong:
@apc999 Thanks for the suggestion commit, accepted and merged it. PTAL.

integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
```diff
@@ -228,6 +235,9 @@ private int loadAndRunTests(Object operations, String testDir, PrintStream msgSt
       for (Method test : tests) {
         String testName = test.getName();
         if (testName.endsWith("Test")) {
+          if (!(mTestList.isEmpty() || mTestList.contains(testName))) {
```
LuQQiu, line 238: will be better for us to check if the input test names are valid. if not, print out all valid test names. it doesn't need to be done if current PR

LuQQiu (approved):
LGTM, thanks for adding this functionality

integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
LuQQiu: FYI, if you don't have time to do this, can create a contributor task for other people to work on it

LuQQiu:

integration/tools/validation/src/main/java/alluxio/cli/UnderFileSystemContractTest.java
maobaolong: https://github.com/Alluxio/new-contributor-tasks

@LuQQiu Good suggestion, created a new contributor task.

PR #13428, opened 2021-05-16
[DOCFIX] Modify wordings in Prepare Docker Volume to Persist Data

ssz1997:
The original wordings in section 2 Prepare Docker Volume to Persist Data in Deploy Alluxio on Docker (https://docs.alluxio.io/os/user/stable/en/deploy/Running-Alluxio-On-Docker.html#prepare-docker-volume-to-persist-data), where they instruct users to "run" the commands to create a host/named volume, may lead users to actually run the command. Instead they should be included in the complete command which is introduced in section 3.

Fix typo where there are two consecutive "be".

The "Therefore" in the later part of the same section can be a little bit strong, considering that both the host directory and the container location are just examples.


apc999 (approved):
Thanks for improving the doc!

apc999:

PR #13429, opened 2021-05-17
Complete symlink operation in JNI-fuse with n-op

yabola:
This PR will complete unsupported symlink operation in libfuse.
This can avoid execution errors, be convenient for developers to debug JNI-fuse and  logic implementation later if need.
I will complete other operation later according to https://github.com/libfuse/libfuse/blob/fuse_2_9_5/include/fuse.h#L88


integration/jnifuse/fs/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
```diff
@@ -327,6 +327,11 @@ public int removexattrCallback(String path, String name) {
     return 0;
   }
 
+  public int symlinkCallback(String linkname, String path) {
+    LOG.debug("Ignore unsupported symlink operation, linkname {}, path{}", linkname, path);
```
maobaolong, line 331: If we don't implement this function, thinking about return an error code instead of swallow the error only left a line of log.

maobaolong (changes requested):
Thank you for the contribution to the Alluxio FUSE module, this is a good catch to handle the `symlinkOper`, look good overall, left some comments inline. BTW, after we are ready to merge this PR, please commit both the Linux and Macosx version of `libjnifuse` binary file.

yabola:
thanks for review, I will update recently.

integration/jnifuse/native/src/main/native/libjnifuse/jnifuse_impls.cc
```diff
@@ -135,3 +135,7 @@ int write_wrapper(const char *path, const char *buf, size_t size, off_t off,
   return jnifuse::JniFuseFileSystem::getInstance()->writeOper->call(
       path, buf, size, off, fi);
 }
+
+int symlink_wrapper(const char *linkname, const char *path) {
+  return jnifuse::JniFuseFileSystem::getInstance()->symlinkOper->call(linkname, path);
+}
```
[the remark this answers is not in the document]
yabola: done

integration/jnifuse/native/src/main/native/libjnifuse/jnifuse_impls.h
```diff
@@ -47,5 +47,5 @@ int unlink_wrapper(const char *path);
 int utimens_wrapper(const char *path, const struct timespec ts[2]);
 int write_wrapper(const char *path, const char *buf, size_t size, off_t off,
                   struct fuse_file_info *fi);
-
+int symlink_wrapper(const char *linkname, const char *path);
```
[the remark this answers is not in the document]
yabola: done

integration/jnifuse/native/src/main/native/libjnifuse/jnifuse_impls.cc
```diff
@@ -135,3 +135,7 @@ int write_wrapper(const char *path, const char *buf, size_t size, off_t off,
   return jnifuse::JniFuseFileSystem::getInstance()->writeOper->call(
       path, buf, size, off, fi);
 }
+
+int symlink_wrapper(const char *linkname, const char *path) {
```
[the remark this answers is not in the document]
yabola: done

integration/jnifuse/fs/src/main/java/alluxio/jnifuse/AbstractFuseFileSystem.java
yabola: I have modified it. how do you feel the current way~

maobaolong (commented):
Thanks for the update, Look good overall, please update the macos and linux share lib binary file

yabola:
@maobaolong I have uploaded the Linux binary lib. please take a look~

maobaolong (approved):
Thank you for this improvement, it LGTM, @LuQQiu please take another look?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -754,6 +754,12 @@ public int utimensCallback(String path, long aSec, long aNsec, long mSec, long m
     return 0;
   }
 
+  @Override
+  public int symlink(String linkname, String path) {
+    LOG.debug("Ignore unsupported symlink operation, linkname {}, path{}", linkname, path);
```
LuQQiu, line 759: Do we want an unsupported operation symlink to return 0?
In any case (workloads) will unsupported operation symlink return 0 helps? 
Normally for an unsupported operation, we log WARN message and return error code.

LuQQiu (commented):
Thanks for adding the symlink operation to JNI-FUSE! It really helps debugging and trouble-shooting. Left some questions

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
yabola: @LuQQiu done, please take a look~

LuQQiu (approved):
LGTM! Left one minor question, is the .so file suppose to be removed?

yabola:
> LGTM! Left one minor question, is the .so file suppose to be removed?

@LuQQiu I compiled the Alluxio JNI-fuse code in the Linux environment. We don't need so library now?

LuQQiu:
smaller or bigger should be fine, as long as the compile is correct. @maobaolong do we have some jnifuse file compile convention? or just directly compile in different env is fine

LuQQiu:

PR #13433, opened 2021-05-18
[SMALLFIX] Improve of set java home

hddong:
Due to no-login shell does not source `/etc/profile`, we always get `Cannot find 'java'` error when we just set `JAVA_HOME` in `/etc/profile`.
It's better to have a more friendly tip.


apc999 (approved):

apc999:

PR #13446, opened 2021-05-19
Fix throw CCE when using Fuse while enable client cache bug

bzheng888:


bzheng888:
Hi, @maobaolong, review this pr please

PR #13447, opened 2021-05-19
Fix throw CCE when using Fuse while enable client cache bug

### LINKED ISSUE #13442: Throw CCE when using Fuse while enable client cache
**Alluxio Version:**
Alluxio master branch, HEAD

**Describe the bug**

image

```
java.lang.ClassCastException: alluxio.client.file.cache.LocalCacheFileInStream cannot be cast to alluxio.client.file.AlluxioFileInStream
        at alluxio.fuse.AlluxioJniFuseFileSystem.readInternal(AlluxioJniFuseFileSystem.java:429)
        at alluxio.fuse.AlluxioJniFuseFileSystem.lambda$read$17(AlluxioJniFuseFileSystem.java:404)
        at alluxio.fuse.AlluxioFuseUtils.call(AlluxioFuseUtils.java:280)
        at alluxio.fuse.AlluxioJniFuseFileSystem.read(AlluxioJniFuseFileSystem.java:404)
        at alluxio.jnifuse.AbstractFuseFileSystem.readCallback(AbstractFuseFileSystem.java:180)
```

**To Reproduce**

Enable client cache in `alluxio-site.properties`

```
alluxio.user.client.cache.enabled=true
```

And then start Alluxio-fuse, read a file

**Expected behavior**

get the content of the file.

**Urgency**

Urgent, block to use Alluxio Local Cache.

**Additional context**
No.

bzheng888:
Fix #13442

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
@@ -15,7 +15,9 @@ import alluxio.Seekable;
 import alluxio.client.BoundedStream;
 import alluxio.client.PositionedReadable;
 
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
 
 /**
  * A streaming API to read a file. This API represents a file as a stream of bytes and provides a
@@ -24,4 +26,31 @@ import java.io.InputStream;
  */
 public abstract class FileInStream extends InputStream implements BoundedStream, PositionedReadable,
     Seekable {
+  /**
+   * Reads up to len bytes of data from the input stream into the byte buffer.
+   *
+   * @param byteBuffer the buffer into which the data is read
+   * @param off the start offset in the buffer at which the data is written
+   * @param len the maximum number of bytes to read
+   * @return the total number of bytes read into the buffer, or -1 if there is no more
+   *         data because the end of the stream has been reached
+   */
+  public int read(ByteBuffer byteBuffer, int off, int len) throws IOException {
+    int nread = 0;
+    int rd = 0;
+    final int sz = len;
+    final byte[] dest = new byte[sz];
+    while (rd >= 0 && nread < sz) {
+      rd = read(dest, nread, sz - nread);
+      if (rd >= 0) {
+        nread += rd;
+      }
+    }
+    if (nread == -1) { // EOF
+      nread = 0;
+    } else if (nread > 0) {
+      byteBuffer.put(dest, 0, nread);
+    }
+    return nread;
+  }
 }
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -12,7 +12,6 @@
 package alluxio.fuse;
 
 import alluxio.AlluxioURI;
-import alluxio.client.file.AlluxioFileInStream;
 import alluxio.client.file.FileInStream;
 import alluxio.client.file.FileOutStream;
 import alluxio.client.file.FileSystem;
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -402,7 +402,7 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
         if (offset - is.getPos() < is.remaining()) {
           is.seek(offset);
           while (rd >= 0 && nread < sz) {
-            rd = ((AlluxioFileInStream) is).read(buf, nread, sz - nread);
+            rd = is.read(buf, nread, sz - nread);
             if (rd >= 0) {
               nread += rd;
             }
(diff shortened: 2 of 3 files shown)


core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -24,4 +26,31 @@
...
+     * Reads up to len bytes of data from the input stream into the byte buffer.
+     *
+     * @param byteBuffer the buffer into which the data is read
+     * @param off the start offset in the buffer at which the data is written
+     * @param len the maximum number of bytes to read
+     * @return the total number of bytes read into the buffer, or -1 if there is no more data because
+     *         the end of the stream has been reached
+     */
+    public int read(ByteBuffer byteBuffer, int off, int len) throws IOException {
+        int nread = 0;
```
maobaolong, line 39: Please fix the wrap line ident from 4 to 2.

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -24,4 +26,31 @@
  */
 public abstract class FileInStream extends InputStream implements BoundedStream, PositionedReadable,
     Seekable {
+    /**
+     * Reads up to len bytes of data from the input stream into the byte buffer.
+     *
+     * @param byteBuffer the buffer into which the data is read
+     * @param off the start offset in the buffer at which the data is written
+     * @param len the maximum number of bytes to read
+     * @return the total number of bytes read into the buffer, or -1 if there is no more data because
```
maobaolong, line 35: keep line length less than 100 chars.

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -24,4 +26,31 @@
...
+     *         the end of the stream has been reached
+     */
+    public int read(ByteBuffer byteBuffer, int off, int len) throws IOException {
+        int nread = 0;
+        int rd = 0;
+        final int sz = len;
+        final byte[] dest = new byte[sz];
+        while (rd >= 0 && nread < sz) {
```
maobaolong, line 43: Is this while loop necessary? @LuQQiu @apc999 , this code fragment is copied from `AlluxioFuseFileSystem#copyInternal`.

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -402,7 +402,7 @@ private int readInternal(String path, ByteBuffer buf, long size, long offset, Fu
         if (offset - is.getPos() < is.remaining()) {
           is.seek(offset);
           while (rd >= 0 && nread < sz) {
```
maobaolong, line 404: The same question, is this while loop necessary? @LuQQiu @apc999

maobaolong (changes requested):
Thank you for fix this `CCE` and let the `Local Cache` can be used in Alluxio-Fuse.
Please fix the style issue and pass all of the ci checks.

apc999:
@bzheng888  can you add some more info into the PR description, so we know the context of this fix?

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
maobaolong: After an offline discussion with @LuQQiu , I knew the the reason is that we want keep read a specific size of input stream or touch the EOF, when the actual read length is smaller that the specific length, using while-read mode can continue to read the inputstream.

bzheng888:
@apc999 Sorry, I forgot to mention the issue, this pr is to Fix #13442.

maobaolong (approved):
It LGTM, @apc999 Would you please take another look, this blocker make `LocalCache` cannot be used in Alluxio Fuse.

apc999 (approved):

apc999:

core/client/fs/src/main/java/alluxio/client/file/FileInStream.java
```diff
@@ -24,4 +26,31 @@
...
+    final int sz = len;
+    final byte[] dest = new byte[sz];
+    while (rd >= 0 && nread < sz) {
+      rd = read(dest, nread, sz - nread);
+      if (rd >= 0) {
+        nread += rd;
+      }
+    }
+    if (nread == -1) { // EOF
+      nread = 0;
```
dbw9580, line 50: These lines don't look right: `nread` is only updated by adding `rd`, where `rd` is non-negative. `nread` is 0 at the beginning. Therefore `nread` is non-negative. Why do we check if it is -1?

bzheng888: @dbw9580 Good catch! This code is obtained from AlluxioFuseFileSystem.java->readInternal, so it seems both of them have similar problems, can you fix this?

dbw9580: This line was added in this commit https://github.com/Alluxio/alluxio/commit/0d30732ba71674357d9d30c3c34fa701efaacfda so it's almost been here since like genesis. 🤦

bzheng888: And this is indeed an obvious mistake!

PR #13448, opened 2021-05-19
Add Meter metric for block eviction rate

ZhuTopher:
This PR adds a `Meter` type metric to measure the worker block eviction rate. The implementation leverages the existing `Counter` metric for the worker block eviction count, and so the two metrics should always be in-sync.


ZhuTopher:
I was only able to verify that the metric value appears in the metric sink, but wasn't able to force a block eviction. Does anyone have a set-up/command suggestion to do that?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -807,6 +807,12 @@ public MetricKey build() {
           .setMetricType(MetricType.COUNTER)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey WORKER_BLOCKS_EVICTED_RATE =
+      new Builder(Name.WORKER_BLOCKS_EVICTED_RATE)
+          .setDescription("Block eviction rate in this worker.")
+          .setMetricType(MetricType.METER)
+          .setIsClusterAggregated(false)
```
ZhuTopher, line 814: @yuzhu Should this metric (Eviction rate) be per-worker or across the entire cluster?

yuzhu: if you set ClusterAggregated to true, it will produce an aggregated cluster metric at the master.

yuzhu: as well as per worker stats

ZhuTopher: So is there any reason to not have it be `true` in this case? Is there a performance overhead or is that negligable?

yuzhu: yeah it will increase the metrics aggregation rpc that is sent to the master.  but i think in this case, we want the aggregate eviction rate, so we would want it set to true.

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -807,6 +807,12 @@ public MetricKey build() {
           .setMetricType(MetricType.COUNTER)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey WORKER_BLOCKS_EVICTED_RATE =
```
ggezer, line 810: nit: EVICTION_RATE

ggezer (approved):

core/common/src/main/java/alluxio/metrics/MetricKey.java
ZhuTopher: I was torn between naming this `WORKER_BLOCK_EVICTION_RATE` and `WORKER_BLOCKS_EVICTED_RATE`, I can make it `WORKER_BLOCKS_EVICTION_RATE` if we don't care about the phonetic reading too?

ggezer: Both works but I guess I can't see why making it eviction_rate is a careless act phonetically :/

ZhuTopher: Haha actually it'd depend on how you interpret the whole phrase so actually it could work too.

I thought of another possible reason to keep it as is, keeping `WORKER_BLOCKS_EVICTED` as a prefix lets you auto-complete both the original `Counter` and also the new `Meter` metrics. I think it'd make sense to keep the two as close as possible in naming since they are literally updated at the same location.

ggezer:

PR #13450, opened 2021-05-20
Fix variable reference and exit status

ljl1988com:


bin/alluxio
```diff
@@ -223,11 +223,11 @@ function main {
           shift # remove -s param
         fi
       else
-        echo "${Usage} $0 format [-s]"
+        echo "Usage $0 format [-s]"
```
LuQQiu, line 226: Can you help change to to the format similar to 
```
    echo "Usage: alluxio killAll <WORD>"
```

bin/alluxio
```diff
@@ -223,11 +223,11 @@ function main {
...
         exit 2
       fi
     elif [[ $# -gt 1 ]]; then
-      echo "${Usage} $0 format [-s]"
```
LuQQiu, line 230: Same here

LuQQiu (changes requested):
Thanks for fixing the bug in the bash script! Left some minor comments, otherwise LGTM

bin/alluxio
```diff
@@ -81,7 +81,7 @@ EOF
 function killAll {
   if [[ $# -ne 1 ]]; then
     echo "Usage: alluxio killAll <WORD>"
-    exit
+    exit 1
```
ljl1988com, line 84: https://github.com/Alluxio/alluxio/blob/837a41c68b956ded56ae0fd41f0bf26b52ffbf49/bin/alluxio#L112
but
https://github.com/Alluxio/alluxio/blob/837a41c68b956ded56ae0fd41f0bf26b52ffbf49/bin/alluxio#L227
and in file `integration/emr/alluxio-emr.sh` all exit 2.
Which one is better?@LuQQiu

apc999: perhaps exit 2 is better here
according to https://tldp.org/LDP/abs/html/exitcodes.html

apc999:
thanks for our contribution @ljl1988com  . One minor comment left.

LuQQiu (approved):
LGTM, thanks for improving the shell error message and return code!

LuQQiu:

PR #13453, opened 2021-05-20
Save one copy in MasterWorkerInfo.register

jiacheliu3:
Separated this optimization into one standalone PR

https://github.com/Alluxio/alluxio/pull/13384 will be rebased after this

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
+++ b/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
@@ -157,7 +157,7 @@ public final class MasterWorkerInfo {
     }
 
     // Set the new block information.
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -157,7 +157,7 @@ public MasterWorkerInfo(long id, WorkerNetAddress address) {
     }
 
     // Set the new block information.
-    mBlocks = new HashSet<>(blocks);
+    mBlocks = blocks;
```
jiacheliu3, line 160: The `blocks` variable is a newly copied `HashSet` already, passed to this method. There's no benefit copying again.

ggezer (approved):

jiacheliu3:

PR #13454, opened 2021-05-20
Save one copy in DefaultBlockMaster.getLostBlocks

jiacheliu3:


ggezer (commented):
It looks like this will introduce iteration side effects on `DefaultFileSystemMaster.getLostFiles()` as the backing set can be modified at any time.

jiacheliu3:
Closed because https://github.com/Alluxio/alluxio/pull/13467 is better

PR #13455, opened 2021-05-20
Support bypass table when attach and sync db

maobaolong:
This PR supply a new option `catalog.db.config.file` to describe detail information or concrete special behavior for the database. Now, i add a json format file which contains `bypassSet`, if tables are in it, they will be bypassed.

Bypass here means return the udb location directly to client, so that client can read data through ufs directly.

This is an example of db config file.

```json
{
  "bypassSet" : ["person1", "person2"]
}

```

We can use the db config file when `attachdb` and `sync`.

```
bin/alluxio table attachdb hive thrift://localhost:9083 test -o catalog.db.config.file=/etc/alluxio/conf/db-test.json
bin/alluxio table sync test
```

I create 3 table named `person` `person1` and `person2`, now I bypass `person1` and `person2`, left `person` in Alluxio.

As a test result, I show the table content by execute SQL `select * from <TABLE_NAME>` in alluxio cli. It's clearly show that the person and person1 are different, because I modified the content to distinguish whether the data is from Alluxio or ufs.

image

### DIFF
--- a/docs/en/core-services/Catalog.md
+++ b/docs/en/core-services/Catalog.md
@@ -102,10 +102,22 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
     thrift://metastore_host:9083 default
 ```
 
+If you want specify config file for the UDB, please add `attachdb` option `-o catalog.db.config.file`.
+Each time you want to affect the config file, you can use `alluxio table sync`.
+
+If you want to write a configuration file for the UDB, please configure it like the following given example.
+It must contain an array `bypassSet`, the comma-separated table names will be bypassed.
+
+```json
+{
+  "bypassSet" : ["table1", "table2"]
+}
+```
+
 > **Note:** When databases are attached, all tables are synced from the configured UDB.
 If out-of-band updates occur to the database or table and the user wants query results to reflect
 the updates, the database must be synced. See [Syncing Databases](#syncing-databases) for more
-information.
+information. 
 
 ### Exploring Attached Databases
 
@@ -212,7 +224,7 @@ $ ${ALLUXIO_HOME}/bin/alluxio table sync alluxio_db
 ```
 
 This sync command will update the Alluxio catalog metadata according to the updates, deletions,
-and additions in the UDB tables.
+and additions in the UDB tables, it will update db config as well.
 
 ## Using Alluxio Structured Data with Presto
 
--- a/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
@@ -68,7 +68,7 @@ public class PathTranslator {
   public String toAlluxioPath(String ufsPath) throws IOException {
     for (BiMap.Entry<String, String> entry : mPathMap.entrySet()) {
       if (ufsPath.startsWith(entry.getValue())) {
-        // return ufsPath if key and value are same.
+        // return ufsPath if set the key and value to be same when bypass path.
         if (entry.getKey().equals(entry.getValue())) {
           return ufsPath;
         }
--- a/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
@@ -68,6 +68,10 @@ public class PathTranslator {
   public String toAlluxioPath(String ufsPath) throws IOException {
     for (BiMap.Entry<String, String> entry : mPathMap.entrySet()) {
       if (ufsPath.startsWith(entry.getValue())) {
+        // return ufsPath if key and value are same.
+        if (entry.getKey().equals(entry.getValue())) {
+          return ufsPath;
+        }
         String alluxioPath = entry.getKey() + ufsPath.substring(entry.getValue().length());
         if (alluxioPath.startsWith("/")) {
           // scheme/authority are missing, so prefix with the scheme and authority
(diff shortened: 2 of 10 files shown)


apc999:
@maobaolong would you also update the SDS docs?

maobaolong:
@apc999 Sure, updated the related docs, PTAL.

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -68,6 +68,10 @@ public PathTranslator addMapping(String alluxioPath, String ufsPath) {
   public String toAlluxioPath(String ufsPath) throws IOException {
     for (BiMap.Entry<String, String> entry : mPathMap.entrySet()) {
       if (ufsPath.startsWith(entry.getValue())) {
+        // return ufsPath if key and value are same.
```
HelloHorizon, line 71: nit: Add some description under which case the key and value will be same.

table/server/master/src/main/java/alluxio/master/table/CatalogProperty.java
```diff
@@ -38,4 +38,7 @@ private CatalogProperty(String name, String description, String defaultValue) {
               + "overload the UDB, and if set too low, syncing a database with many tables may "
               + "take a long time.",
           Integer.toString(DEFAULT_DB_SYNC_THREADS));
+  public static final CatalogProperty DB_CONFIG_FILE =
+      new CatalogProperty("catalog.db.config.file",
+          "The config file for the UDB.", "<catalog.db.config.file>");
```
HelloHorizon, line 43: It seems there is only one add on configuration for by pass, it will be better to use single configuration with helper function. Check `alluxio.conf.InstancedConfiguration::getList` for reference.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -151,6 +152,10 @@ private PathTranslator mountAlluxioPaths(Table table, List<Partition> partitions
 
     try {
       PathTranslator pathTranslator = new PathTranslator();
+      if (bypass) {
+        pathTranslator.addMapping(hiveUfsUri, hiveUfsUri);
+        return pathTranslator;
+      }
```
HelloHorizon, line 158: Add bypass check for GlueDatabase?

HelloHorizon (changes requested):
@maobaolong Thanks for the improvement! Left some questions.

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
maobaolong: Ok

table/server/master/src/main/java/alluxio/master/table/CatalogProperty.java
maobaolong: I'm afraid that I have to keep using `config.file` since I want to refresh the bypass list by `alluxio table sync` command, further, I want to add more config item in this file.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
maobaolong: Sure.

maobaolong:
@HelloHorizon Thank you for the review for me, I'm sorry for missing the GlueDatabase bypass check, I push a new commit to address your comments, PTAL.

table/server/master/src/main/java/alluxio/master/table/CatalogProperty.java
HelloHorizon: I see, that make sense to me.

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -201,6 +209,11 @@ public SyncStatus sync(JournalContext context) throws IOException {
     // Synchronization is necessary if accessed concurrently from multiple threads
     SyncStatus.Builder builder = SyncStatus.newBuilder();
 
+    if (Files.exists(Paths.get(mConfigPath))) {
+      ObjectMapper mapper = new ObjectMapper();
+      DbConfig dbConfig = mapper.readValue(new File(mConfigPath), DbConfig.class);
+      mBypassSet.addAll(dbConfig.getBypassSet());
+    }
```
HelloHorizon, line 216: I think if we follow the format of configuration as:
`{  "bypassSet" : ["table1", "table2"]}`, it will have problem in some cases.
For example, we have db1:{table1, table2, table3}, db2:{table1, table2, table3}. If we only want to bypass the table1 in db1, but we will also bypass table1 in db2?

maobaolong: The configuration is related each database instance. So, for db1, we use a configuration file db1.config.json, for db2 is db2.config.json.

docs/en/core-services/Catalog.md
```diff
@@ -102,10 +102,22 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
+{
+  "bypassSet" : ["table1", "table2"]
+}
+```
+
 > **Note:** When databases are attached, all tables are synced from the configured UDB.
 If out-of-band updates occur to the database or table and the user wants query results to reflect
 the updates, the database must be synced. See [Syncing Databases](#syncing-databases) for more
-information.
+information.
```
apc999, line 120: nit: extra whitespace

apc999 (approved):
Adding a minor comment. Once @HelloHorizon is done with the review, I will merge it

HelloHorizon (approved):
LGTM

maobaolong:
@HelloHorizon Thank you for your review.

HelloHorizon:

PR #13461, opened 2021-05-21
Save one copy of BlockLocation

jiacheliu3:
In `DefaultBlockMaster` we create extra copies of `BlockLocation` while we can just use the one given.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
@@ -143,7 +143,8 @@ public final class BlockMasterWorkerServiceHandler extends
     final Map<String, StorageList> lostStorageMap = request.getLostStorageMap();
 
     final Map<Block.BlockLocation, List<Long>> currBlocksOnLocationMap =
-            reconstructBlocksOnLocationMap(request.getCurrentBlocksList(), workerId);
+        reconstructBlocksOnLocationMap(request.getCurrentBlocksList(), workerId);
+
 
     RegisterWorkerPOptions options = request.getOptions();
     RpcUtils.call(LOG,
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
@@ -145,7 +145,6 @@ public final class BlockMasterWorkerServiceHandler extends
     final Map<Block.BlockLocation, List<Long>> currBlocksOnLocationMap =
         reconstructBlocksOnLocationMap(request.getCurrentBlocksList(), workerId);
 
-
     RegisterWorkerPOptions options = request.getOptions();
     RpcUtils.call(LOG,
         (RpcUtils.RpcCallableThrowsIOException<RegisterWorkerPResponse>) () -> {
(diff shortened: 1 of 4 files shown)


core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -995,12 +995,7 @@ private void processWorkerAddedBlocks(MasterWorkerInfo workerInfo,
           Optional<BlockMeta> block = mBlockStore.getBlock(blockId);
           if (block.isPresent()) {
             workerInfo.addBlock(blockId);
-            BlockLocation blockLocation = BlockLocation.newBuilder()
-                .setWorkerId(workerInfo.getId())
-                .setTier(entry.getKey().getTier())
-                .setMediumType(entry.getKey().getMediumType())
-                .build();
-            mBlockStore.addLocation(blockId, blockLocation);
+            mBlockStore.addLocation(blockId, entry.getKey());
```
jiacheliu3, line 998: Now that the `BlockLocation` contains the worker ID, I don't see a reason why we don't use that.

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
```diff
@@ -76,8 +76,6 @@
   private static final Map<Block.BlockLocation, List<Long>> NO_BLOCKS_ON_LOCATION
       = ImmutableMap.of();
   private static final Map<String, StorageList> NO_LOST_STORAGE = ImmutableMap.of();
-  private static final Block.BlockLocation BLOCK_LOCATION = Block.BlockLocation.newBuilder()
-      .setTier(Constants.MEDIUM_MEM).setMediumType(Constants.MEDIUM_MEM).build();
```
jiacheliu3, line 80: The manually simulated `BlockLocation` objects now should contain the correct worker IDs. There can be more tests that I haven't found, that need to be updated.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -81,7 +81,8 @@ public void blockHeartbeat(BlockHeartbeatPRequest request,
             .stream()
             .collect(
                 Collectors.toMap(
-                    e -> Block.BlockLocation.newBuilder().setTier(e.getKey().getTierAlias())
+                    e -> Block.BlockLocation.newBuilder().setWorkerId(workerId)
+                        .setTier(e.getKey().getTierAlias())
```
jiacheliu3, line 85: We have the worker ID here already, so we can just set it to the `BlockLocation` and pass it to the `processWorkerAddedBlocks()` method. The worst case is, the blocks are unknown and we just ignore them.

I don't really see a reason why we ignore the worker ID here. The worker IDs don't change anywhere in the `register()` or `heartbeat()`. We just look for the `MasterWorkerInfo` with the ID and then get the ID from the `MasterWorkerInfo` again... Another reason might be we don't add the worker ID here so the hash code doesn't contain worker ID, so the `BlockLocation` items are together with other `BlockLocation` from other workers in some map. But that also doesn't make sense...

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
ggezer: This cut seems promising!
Can you also assert that workerIds match as expected?

ggezer (changes requested):

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -995,7 +995,14 @@ private void processWorkerAddedBlocks(MasterWorkerInfo workerInfo,
           Optional<BlockMeta> block = mBlockStore.getBlock(blockId);
           if (block.isPresent()) {
             workerInfo.addBlock(blockId);
-            mBlockStore.addLocation(blockId, entry.getKey());
+            BlockLocation location = entry.getKey();
```
ggezer, line 998: nit: Preconditions.checkState()

ggezer (approved):

jiacheliu3:
rebased to resolve conflicts with the other PR https://github.com/Alluxio/alluxio/pull/13390 Now this is ready to merge.

jiacheliu3:

PR #13466, opened 2021-05-21
Allow load to read from another worker and force to local worker

yuzhu:
The earlier change can potentially cause undesired load balance among workers.
Because it always respects the alluxio configuration instead of LocalFirst.

Also it always loads from the ufs instead of potentially from another worker when
it is already in alluxio memory. (e.g. When replication checker runs load)

### DIFF
--- a/job/server/src/main/java/alluxio/job/util/JobUtils.java
+++ b/job/server/src/main/java/alluxio/job/util/JobUtils.java
@@ -39,6 +39,7 @@ import alluxio.wire.FileBlockInfo;
 import alluxio.wire.WorkerNetAddress;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.io.ByteStreams;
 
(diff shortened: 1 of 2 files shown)


job/server/src/main/java/alluxio/job/util/JobUtils.java
```diff
@@ -130,21 +131,23 @@ public static void loadBlock(URIStatus status, FileSystemContext context, long b
           ExceptionMessage.PINNED_TO_MULTIPLE_MEDIUMTYPES.getMessage(status.getPath()));
     }
 
-    // when the data to load is persisted, simply use local worker to load from UFS
+    // when the data to load is persisted, simply use local worker to load from UFS or an worker
```
apc999, line 134:
```suggestion
    // when the data to load is persisted, simply use local worker to load from UFS (e.g. diistributedLoad),
    // or from a remote worker (e.g., setReplication)
```

apc999 (approved):
LGTM modulo my minor comments in javadoc

yuzhu:

PR #13467, opened 2021-05-22
Refactor BlockMaster interface to avoid copying the lost block set

jiacheliu3:
This is a better approach of achieving https://github.com/Alluxio/alluxio/pull/13454

The current implementation copies the `DefaultBlockMaster.mLostBlocks` to whoever is using the set. This change improves the `BlockMaster` interface to avoid the need to copy that collection. The `mLostBlocks` set can be huge when there are workers lost. Also the cost is very high because we just copy on every read.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
@@ -30,6 +30,7 @@ import alluxio.wire.BlockInfo;
 import alluxio.wire.WorkerInfo;
 import alluxio.wire.WorkerNetAddress;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -211,9 +212,22 @@ public interface BlockMaster extends Master, ContainerIdGenerable {
       List<Metric> metrics);
 
   /**
-   * @return the block ids of lost blocks in Alluxio
-   */
-  Set<Long> getLostBlocks();
+   * @return whether the block is considered lost in Alluxio
+   * */
+  boolean hasLostBlock(long blockId);
+
+  /**
+   * Returns an {@link Iterator} over the lost blocks.
+   * Note that the iterator should not be shared across threads.
+   *
+   * @return an Iterator
+   * */
+  Iterator<Long> getLostBlocksIterator();
+
+  /**
+   * @return the number of lost blocks in Alluxio
+   * */
+  int getLostBlocksCount();
 
   /**
    * Reports the ids of the blocks lost on workers.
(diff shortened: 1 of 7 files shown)


core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
```diff
@@ -211,9 +212,23 @@ Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnTiers,
...
+   * Note that the iterator should not be shared across threads.
+   *
+   * @return an Iterator
+   * */
+  Iterator<Long> getLostBlocksIterator();
+
+  /**
+   * @return the number of lost blocks in Alluxio
+   * */
+  int getLostBlocksCount();
```
ggezer, line 231: This seems to be test only.
Can you replace this with guava's Iterators.size() for the sake of reducing methods? Or mark it with @VisibleForTest?

core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
```diff
@@ -211,9 +212,23 @@ Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnTiers,
       List<Metric> metrics);
 
   /**
-   * @return the block ids of lost blocks in Alluxio
-   */
-  Set<Long> getLostBlocks();
+   * @param blockId the block ID
+   * @return whether the block is considered lost in Alluxio
+   * */
+  boolean hasLostBlock(long blockId);
```
ggezer, line 218: nit: isLostBlock() ?

ggezer (approved):

apc999 (approved):
minor comments left

core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
jiacheliu3: Actually I started with not having this method and using `Iterators.size()` to loop this iterator. But later I realized if we need the size we would have to do the same (or worse, loop the iterator to get the size in something else like a heartbeating `ReplicationChecker`), so I added this method.

I prefer not reducing this method and marking `@VisibleForTest` for now. And later if something else starts to rely on this method, this annotation can be removed.

core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
```diff
@@ -211,9 +212,23 @@ Command workerHeartbeat(long workerId, Map<String, Long> capacityBytesOnTiers,
       List<Metric> metrics);
 
   /**
-   * @return the block ids of lost blocks in Alluxio
-   */
-  Set<Long> getLostBlocks();
+   * @param blockId the block ID
+   * @return whether the block is considered lost in Alluxio
+   * */
```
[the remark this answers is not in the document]
jiacheliu3: not sure why... my Idea just keeps using that format. Rectified.

jiacheliu3:

PR #13472, opened 2021-05-24
Increase thread.join timeout for stopAfterSecondaryTransition test

ZacBlanco:


ggezer:
@ZacBlanco Can you try relaunching tests few times and see if this change improves test stability?

PR #13478, opened 2021-05-24
[SMALLFIX] Check file equality by comparing URI

dbw9580:
Fixes alluxio/new-contributor-tasks#594.


apc999:
@maobaolong  can you elaborate why we have this fix?
Is it because `files[0].getPath()` may return null?

LuQQiu:
@maobaolong

yyongycy:
so this change is pending due to the check error shown in above comment, is it?

maobaolong:
because the authority of default uri could be different with the current one, and if the authority is null, it may be "null", so we don't care the authority but the path.

apc999:
close as this PR for now as it is duplicated from https://github.com/Alluxio/alluxio/pull/13560

Issue #13480, opened 2021-05-24
Add some metrics to measure the performance of master RPC


maobaolong:
**Is your feature request related to a problem? Please describe.**

Now, we lack of some metrics to show the performance or rpc server request throughput of alluxio master.

**Describe the solution you'd like**

Add some metrics.

**Describe alternatives you've considered**

Modify the rpc communicate layer.

**Urgency**

Middle

**Additional context**

No

maobaolong:
Reference the HDFS namemode `call per second`

PR #13481, opened 2021-05-24
Add metrics for Worker RPC counts

ZhuTopher:
This PR aims at adding metrics for tracking the number of active RPCs handled by the Alluxio worker.

### DIFF
--- a/core/common/src/main/java/alluxio/metrics/MetricKey.java
+++ b/core/common/src/main/java/alluxio/metrics/MetricKey.java
@@ -1035,6 +1035,18 @@ public final class MetricKey implements Comparable<MetricKey> {
           .setMetricType(MetricType.GAUGE)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey WORKER_RPC_READ_COUNT =
+      new Builder("Worker.RpcReadCount")
+          .setDescription("The number of active read-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
+  public static final MetricKey WORKER_RPC_WRITE_COUNT =
+      new Builder("Worker.RpcWriteCount")
+          .setDescription("The number of active write-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
 
   // Client metrics
   public static final MetricKey CLIENT_BLOCK_READ_CHUNK =
--- a/core/common/src/main/java/alluxio/metrics/MetricKey.java
+++ b/core/common/src/main/java/alluxio/metrics/MetricKey.java
@@ -625,6 +625,16 @@ public final class MetricKey implements Comparable<MetricKey> {
           .build();
 
   // Cluster metrics
+  public static final MetricKey CLUSTER_ACTIVE_RPC_READ_COUNT =
+      new Builder("Cluster.ActiveRpcReadCount")
+          .setDescription("The number of active read-RPCs managed by workers")
+          .setMetricType(MetricType.COUNTER)
+          .build();
+  public static final MetricKey CLUSTER_ACTIVE_RPC_WRITE_COUNT =
+      new Builder("Cluster.ActiveRpcWriteCount")
+          .setDescription("The number of active write-RPCs managed by workers")
+          .setMetricType(MetricType.COUNTER)
+          .build();
   public static final MetricKey CLUSTER_BYTES_READ_DIRECT =
       new Builder("Cluster.BytesReadDirect")
           .setDescription("Total number of bytes read from Alluxio storage managed by workers "
@@ -790,6 +800,18 @@ public final class MetricKey implements Comparable<MetricKey> {
           .build();
 
   // Worker metrics
+  public static final MetricKey WORKER_ACTIVE_RPC_READ_COUNT =
+      new Builder("Worker.ActiveRpcReadCount")
+          .setDescription("The number of active read-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
+  public static final MetricKey WORKER_ACTIVE_RPC_WRITE_COUNT =
+      new Builder("Worker.ActiveRpcWriteCount")
+          .setDescription("The number of active write-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
   public static final MetricKey WORKER_ASYNC_CACHE_DUPLICATE_REQUESTS =
       new Builder("Worker.AsyncCacheDuplicateRequests")
           .setDescription("Total number of duplicated async cache request received by this worker")
@@ -1034,18 +1056,6 @@ public final class MetricKey implements Comparable<MetricKey> {
               + "by asynchronous block remover.")
           .setMetricType(MetricType.GAUGE)
           .setIsClusterAggregated(false)
-          .build();
-  public static final MetricKey WORKER_RPC_READ_COUNT =
-      new Builder("Worker.RpcReadCount")
-          .setDescription("The number of active read-RPCs managed by this worker")
-          .setMetricType(MetricType.COUNTER)
-          .setIsClusterAggregated(true)
-          .build();
-  public static final MetricKey WORKER_RPC_WRITE_COUNT =
-      new Builder("Worker.RpcWriteCount")
-          .setDescription("The number of active write-RPCs managed by this worker")
-          .setMetricType(MetricType.COUNTER)
-          .setIsClusterAggregated(true)
           .build();
 
   // Client metrics
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
@@ -92,8 +92,8 @@ public class BlockReadHandler implements StreamObserver<alluxio.grpc.ReadRequest
       ServerConfiguration.getMs(PropertyKey.WORKER_REMOTE_IO_SLOW_THRESHOLD);
   /** Metrics. */
   private static final Counter RPC_READ_COUNT =
-      MetricsSystem.counterWithTags(MetricKey.WORKER_RPC_READ_COUNT.getName(),
-            MetricKey.WORKER_RPC_READ_COUNT.isClusterAggregated());
+      MetricsSystem.counterWithTags(MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.getName(),
+            MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.isClusterAggregated());
 
   /** The executor to run {@link DataReader}. */
   private final ExecutorService mDataReaderExecutor;
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
@@ -90,6 +90,9 @@ public class BlockReadHandler implements StreamObserver<alluxio.grpc.ReadRequest
   private static final Logger SLOW_BUFFER_LOG = new SamplingLogger(LOG, Constants.MINUTE_MS);
   private static final long SLOW_BUFFER_MS =
       ServerConfiguration.getMs(PropertyKey.WORKER_REMOTE_IO_SLOW_THRESHOLD);
+  /** Metrics */
+  private static final Counter RPC_READ_COUNT =
+      MetricsSystem.counter(MetricKey.WORKER_RPC_READ_COUNT.getName());
 
   /** The executor to run {@link DataReader}. */
   private final ExecutorService mDataReaderExecutor;
@@ -283,6 +286,7 @@ public class BlockReadHandler implements StreamObserver<alluxio.grpc.ReadRequest
       context.setMeter(MetricsSystem
           .meter(MetricKey.WORKER_BYTES_READ_REMOTE_THROUGHPUT.getName()));
     }
+    RPC_READ_COUNT.inc();
     return context;
   }
 
@@ -484,6 +488,7 @@ public class BlockReadHandler implements StreamObserver<alluxio.grpc.ReadRequest
         }
       } finally {
         context.setBlockReader(null);
+        RPC_READ_COUNT.dec();
       }
     }
 
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
@@ -92,7 +92,8 @@ public class BlockReadHandler implements StreamObserver<alluxio.grpc.ReadRequest
       ServerConfiguration.getMs(PropertyKey.WORKER_REMOTE_IO_SLOW_THRESHOLD);
   /** Metrics. */
   private static final Counter RPC_READ_COUNT =
-      MetricsSystem.counter(MetricKey.WORKER_RPC_READ_COUNT.getName());
+      MetricsSystem.counterWithTags(MetricKey.WORKER_RPC_READ_COUNT.getName(),
+            MetricKey.WORKER_RPC_READ_COUNT.isClusterAggregated());
 
   /** The executor to run {@link DataReader}. */
   private final ExecutorService mDataReaderExecutor;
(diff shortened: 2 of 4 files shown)


core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -1035,6 +1035,18 @@ public MetricKey build() {
...
+          .setDescription("The number of active read-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
+  public static final MetricKey WORKER_RPC_WRITE_COUNT =
+      new Builder("Worker.RpcWriteCount")
+          .setDescription("The number of active write-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
```
ZhuTopher, line 1049: Any suggestions on how to aggregate these two into a single `WORKER_RPC_COUNT`?

Also I assume we'd want these metrics to be cluster aggregated?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
```diff
@@ -90,6 +90,9 @@
   private static final Logger SLOW_BUFFER_LOG = new SamplingLogger(LOG, Constants.MINUTE_MS);
   private static final long SLOW_BUFFER_MS =
       ServerConfiguration.getMs(PropertyKey.WORKER_REMOTE_IO_SLOW_THRESHOLD);
+  /** Metrics */
+  private static final Counter RPC_READ_COUNT =
+      MetricsSystem.counter(MetricKey.WORKER_RPC_READ_COUNT.getName());
```
ZhuTopher, line 95: Is this the correct way to 'register' a metric? When I compiled and tested I didn't see this get reported in the `/metrics/json` sink.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -35,6 +36,9 @@
 @NotThreadSafe
 public final class BlockWriteHandler extends AbstractWriteHandler<BlockWriteRequestContext> {
   private static final Logger LOG = LoggerFactory.getLogger(BlockWriteHandler.class);
+  /** Metrics */
+  private static final Counter RPC_WRITE_COUNT =
+      MetricsSystem.counter(MetricKey.WORKER_RPC_WRITE_COUNT.getName());
```
ZhuTopher, line 41: Ditto here.

ZhuTopher (commented):
@yuzhu Any advice?

core/common/src/main/java/alluxio/metrics/MetricKey.java
yuzhu: the only way i know is to have another metric gauge which is the sum of these two counter values.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
yuzhu: did you go to the `master/metrics/json` or the `worker/metrics/json`

ZhuTopher: Apparently the metric wasn't appearing because I didn't trigger the `Counter.inc()` calls in the code during my test.

In a subsequent test (read a file from S3 bucket UFS, not cached in Alluxio FS), I saw the `Worker.RpcReadCount` metric appear at the `<worker>:30000/metrics/json` endpoint

core/common/src/main/java/alluxio/metrics/MetricKey.java
ZhuTopher: Note-to-self; use `counterWithTag` for cluster-aggregated `Counter`s

@yuzhu Any suggestions on what parent class should own the gauge?

maobaolong:
https://github.com/Alluxio/alluxio/issues/13480

Could you please also add `Alluxio master related rpc metrics`?

ggezer (changes requested):
@ZhuTopher @yuzhu 
Since we are targeting to implement low level stream tracking (ufs/remote/local), I'm not sure if we need these higher level metrics. wdyt?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -103,6 +110,7 @@ protected void cleanupRequest(BlockWriteRequestContext context) throws Exception
       context.getBlockWriter().close();
     }
     mWorker.cleanupSession(context.getRequest().getSessionId());
+    RPC_WRITE_COUNT.dec();
```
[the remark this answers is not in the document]
ZhuTopher: I wasn't able to tell from the code, but I think I can modify it and test whether or not that occurs. Will get back to this comment once I've done that test.

yuzhu:
@ZhuTopher could you verify that we don't already get some RPC statistics for free from this line https://github.com/Alluxio/alluxio/blob/696647645c30d3a63b4a197aa0dfce9e11df3c39/core/server/common/src/main/java/alluxio/RpcUtils.java#L111 


@ggezer AFAIK, these counters are meant to measure how many concurrent requests we are serving at a particular instant in time, not an aggregate count of how many reads and writes we did.  

@calvinjia could you confirm that is the requirement here?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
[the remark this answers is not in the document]
ZhuTopher: I added log statements to each of those blocks, and here was the result of running `./bin/alluxio fs -Dalluxio.user.short.circuit.enabled=false copyFromLocal ./LICENSE /LICENSE`:
```
2021-05-25 18:47:14,240 INFO  BlockWriteHandler - [czhu] WRITE createRequestContext
2021-05-25 18:47:14,244 INFO  BlockWriteHandler - [czhu] WRITE completeRequest
2021-05-25 19:40:33,333 INFO  JvmPauseMonitor - JVM paused 1574ms
No GCs detected 
max memory = 3641M total memory = 148M free memory = 60M
2021-05-25 19:44:13,730 INFO  JvmPauseMonitor - JVM paused 1547ms
No GCs detected 
max memory = 3641M total memory = 148M free memory = 58M
```

ZhuTopher:
@yuzhu Referring to https://github.com/Alluxio/alluxio/blob/696647645c30d3a63b4a197aa0dfce9e11df3c39/core/server/common/src/main/java/alluxio/RpcUtils.java#L111

The only worker-owned pieces of code calling `RpcUtils` which would use that metric is [BlockWorkerClientService](https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/grpc/BlockWorkerClientServiceHandler.java#L152). The methods/operations are the following:
- `asyncCache`
- `removeBlock`
- `moveBlock`
- `clearMetrics`

From what I can tell, the `readBlock` and `writeBlock` methods in this class flow down into the `BlockReadHandler` and `BlockWriteHandler` where I added the RPC metrics. So I think the answers to your question is no?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -1035,6 +1035,18 @@ public MetricKey build() {
...
+      new Builder("Worker.RpcReadCount")
+          .setDescription("The number of active read-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
+          .build();
+  public static final MetricKey WORKER_RPC_WRITE_COUNT =
+      new Builder("Worker.RpcWriteCount")
+          .setDescription("The number of active write-RPCs managed by this worker")
+          .setMetricType(MetricType.COUNTER)
+          .setIsClusterAggregated(true)
```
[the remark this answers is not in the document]
ZhuTopher: @ggezer From what I understand being cluster aggregated would allow you to see the total number of active worker RPCs (reads/writes) from a single metrics sink (i.e: a master node). I think that'd be worthwhile to include?

I'm not 100% clear on the implementation, but I believe the metric on the master would get updated each time any individual worker's counter gets updated (so asynchronously). @yuzhu do you know?

core/common/src/main/java/alluxio/metrics/MetricKey.java
ZhuTopher: I added `counterWithTag` and tested it but wasn't able to get the master metric sink (`:19999/metrics/json`) to show `Cluster.RpcReadCount`). Notably the `WORKER_BYTES_READ_REMOTE` metric `Cluster.BytesReadRemote` _did_ get incremented as a result of the test, so I'm not sure why the RPC counter didn't show up there too.

core/common/src/main/java/alluxio/metrics/MetricKey.java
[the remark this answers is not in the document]
yuzhu: I think there is some value in this, because a very high worker read count would indicate workers are overloaded and may serve as an indicator to the users to increase the number of workers. 

aggregated metrics are updated with the heartbeat messages i think.

yuzhu:
@ZhuTopher i see, yeah streamingRPCs do not use RPCUtils. 
https://github.com/Alluxio/alluxio/blob/3762d75eb611f388895ad9bea0746b17afc7cdd1/core/transport/src/main/proto/grpc/block_worker.proto

core/common/src/main/java/alluxio/metrics/MetricKey.java
ZhuTopher: Fixed in latest commit: https://github.com/Alluxio/alluxio/pull/13481/commits/fe321bd9e7648e0c4a06db88d90ad42b9dacbe3b

core/common/src/main/java/alluxio/metrics/MetricKey.java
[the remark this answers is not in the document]
ZhuTopher: @ggezer Can you re-review this PR?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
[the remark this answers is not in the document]
ggezer: This could be an issue on cancel path, regular completion will be clean afaict.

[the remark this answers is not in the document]
ggezer: https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html
will provide insights into how these handlers interact with clients...

core/common/src/main/java/alluxio/metrics/MetricKey.java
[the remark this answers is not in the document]
ggezer: Main concern I have was to avoid providing redundant metrics, as this particular one will be categorized further with load tracking metrics. However, it's OK to get this in if we're not having those soon.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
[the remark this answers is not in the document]
ZhuTopher: From the `StreamObserver` interface, it says either of `onCompleted()` or `onError()` must be the last method called.

`BlockWriteHandler` extends `AbstractWriteHandler` which has [the following calls](https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/grpc/AbstractWriteHandler.java#L401) for `mResponseObserver`:
```
  /**
   * Writes a response to signify the success of the write request.
   */
  private void replySuccess() {
    mContext.setDoneUnsafe(true);
    mResponseObserver.onCompleted();
  }

  /**
   * Writes a response to signify the successful cancellation of the write request.
   */
  private void replyCancel() {
    mContext.setDoneUnsafe(true);
    mResponseObserver.onCompleted();
  }

  /**
   * Writes an error response.
   */
  private void replyError() {
    replyError(Preconditions.checkNotNull(mContext.getError()));
  }

  /**
   * Writes an error response.
   */
  private void replyError(Error error) {
    if (error.isNotifyClient()) {
      mResponseObserver.onError(error.getCause().toGrpcStatusException());
    }
  }
```

So I agree `BlockWriteHandler#cancelRequest()` is probably not supposed to decrement the counter.

core/common/src/main/java/alluxio/metrics/MetricKey.java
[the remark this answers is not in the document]
ZhuTopher: From discussion with @ggezer we'll be forgoing adding load tracking metrics for the time being; so I think these metrics do not overlap with the metrics we discussed about adding Counters to `BlockReader` implementations right?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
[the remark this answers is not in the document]
ggezer: It seems, this will be called when complete/cancel has trouble. Depending on the place of error, this could cause decrementing the counter twice. (If `replySuccess()` or `replyCancel()` fails)

So ideally each write-handler should remember its increment/decrement status in a boolean flag (if can't reuse an existing one) and don't decrement here if it has already been decremented.

ggezer (changes requested):

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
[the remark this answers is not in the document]
ZhuTopher: Flag has been added, additionally from `AbstractWriteHandler`:
```
  /**
   * Completes this write. This is called when the write completes.
   *  @param context context of the request to complete
   */
  protected abstract void completeRequest(T context) throws Exception;

  /**
   * Cancels this write. This is called when the client issues a cancel request.
   *
   * @param context context of the request to complete
   */
  protected abstract void cancelRequest(T context) throws Exception;

  /**
   * Cleans up this write. This is called when the write request is aborted due to any exception
   * or session timeout.
   *
   * @param context context of the request to complete
   */
  protected abstract void cleanupRequest(T context) throws Exception;
```

I can't tell whether or not `cancelRequest()` results in `completeRequest()` getting called too from the description. The implementation with the `StreamObserver` with `replyCancel()` seems to imply Cancel -> Completed, but that may just be for the `StreamObserver`.

core/common/src/main/java/alluxio/metrics/MetricKey.java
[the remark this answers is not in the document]
ZhuTopher: Here is the PR for the changes mentioned in my previous comment: https://github.com/Alluxio/alluxio/pull/13502

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -103,6 +123,11 @@ protected void cleanupRequest(BlockWriteRequestContext context) throws Exception
       context.getBlockWriter().close();
     }
     mWorker.cleanupSession(context.getRequest().getSessionId());
+
+    if (!mRpcCompleted) {
```
ggezer, line 127: if(!mContext.isDoneUnsafe()) ?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -95,6 +110,11 @@ protected void cancelRequest(BlockWriteRequestContext context) throws Exception
       context.getBlockWriter().close();
     }
     mWorker.abortBlock(request.getSessionId(), request.getId());
+
+    if (!mRpcCompleted) {
```
ggezer, line 114: no need to check the flag here.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -86,6 +96,11 @@ protected void completeRequest(BlockWriteRequestContext context) throws Exceptio
       context.getBlockWriter().close();
     }
     mWorker.commitBlock(request.getSessionId(), request.getId(), request.getPinOnCreate());
+
+    if (!mRpcCompleted) {
```
ggezer, line 100: no need to check the flag here.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -54,6 +62,7 @@
     super(responseObserver, userInfo);
     mWorker = blockWorker;
     mDomainSocketEnabled = domainSocketEnabled;
+    mRpcCompleted = false;
```
ggezer, line 65: It looks you can use `mContext.isDoneUnsafe()` as a completion flag instead of introducing a new flag.

ggezer (changes requested):

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
```diff
@@ -54,6 +59,7 @@
...
```
ggezer, line 62: please remove the left-over field.

ggezer (approved):

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWriteHandler.java
ZhuTopher: Oops, will do

ggezer:

PR #13487, opened 2021-05-26
Add rpc logging for big/long request/response

jiacheliu3:
An experience from observing into RPC is, it's hard to foresee what RPCs are big and costly. This change adds two things:
1. Extra logging when the RPC request/response is larger than a threshold
2. Extra logging when RPCs take longer than a threshold to finish

The extra logs are on the master side because whenever we experience a master-side bottleneck, we will always look into the master log. It can be hard to aggregate all client/worker logs to identify slowness/overhead.


core/server/common/src/main/java/alluxio/ServerRpcUtils.java
```diff
@@ -33,8 +37,11 @@
  * For each of these, if failureOk is set, non-fatal errors will only be logged at the DEBUG
  * level and failure metrics will not be recorded.
  */
-public final class RpcUtils {
-  private RpcUtils() {} // prevent instantiation
+public final class ServerRpcUtils {
```
jiacheliu3, line 40: Renamed `RpcUtils` to `ServerRpcUtils` because this is the server side logic (various `XXXServiceHandler` and `XXXReadHandler` classes). This class is in `alluxio-core-server-common` so I think it should be made obvious that this is the server RPC utilities.

core/server/master/src/main/java/alluxio/master/block/BlockMasterClientServiceHandler.java
```diff
@@ -120,35 +124,47 @@ public void getCapacityBytes(GetCapacityBytesPOptions options,
...
-        "GetWorkerInfoList", "options=%s", responseObserver, options);
+    ServerRpcUtils.call(LOG,
+        () -> {
+          GetWorkerInfoListPResponse response = GetWorkerInfoListPResponse.newBuilder()
+              .addAllWorkerInfos(mBlockMaster.getWorkerInfoList().stream().map(GrpcUtils::toProto)
+                  .collect(Collectors.toList())).build();
+          if (response.getSerializedSize() > RPC_RESPONSE_SIZE_WARNING_THRESHOLD) {
+            LOG.warn("getWorkerInfoList response has size {}, {} WorkerInfo",
+                response.getSerializedSize(), response.getWorkerInfosCount());
+          }
```
apc999, line 143: possibly to squeeze this into `ServerRpcUtils.call` after `T res = callable.call();` 
if we enforce  `T extending com.google.protobuf.GeneratedMessageV3` ?
then we don't need to insert this in every single case to keep your change clean

e.g. 

```
  public static <T extends GeneratedMessageV3> T callAndReturn(Logger logger, RpcCallableThrowsIOException<T> callable,
      String methodName, boolean failureOk, String description, Object... args) {
  ...
  T res = callable.call();
  if (res.getSerializedSize() > RPC_RESPONSE_SIZE_WARNING_THRESHOLD) {
    ...
  }
}

```

core/server/common/src/main/java/alluxio/ServerRpcUtils.java
apc999: can we have a separate PR for the rename and leave only real change in this PR?
This is because this PR has 600+lines of change and most of them are because of rename. Ideally, let's have incremental changes once a time to reduce review complexity.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -61,11 +67,13 @@ public BlockMasterWorkerServiceHandler(BlockMaster blockMaster) {
...
-              request.getAddedBlocksCount(),
-              request.getRemovedBlockIdsCount());
+    if (request.getSerializedSize() > RPC_REQUEST_SIZE_WARNING_THRESHOLD) {
+      LOG.debug("blockHeartbeat request is {} bytes, {} added blocks, "
+                    + "{} removed blocks, {} metrics",
+          request.getSerializedSize(),
+          request.getAddedBlocksCount(),
+          request.getRemovedBlockIdsCount(),
+          request.getOptions().getMetricsCount());
     }
```
apc999, line 77: can we refactor this into a Util method? e.g.

```
public static void logRequestIfTooLarge(GeneratedMessageV3 request, String message, Object... args) {
   if (request.getSerializedSize() > RPC_REQUEST_SIZE_WARNING_THRESHOLD) {
      LOG.warn(messages, args);
   }
}
```

and, why it is LOG.warn for response but LOG.debug for request?

apc999 (changes requested):
Thanks for the improvement!
I left some comments

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
jiacheliu3: why it is LOG.warn for response but LOG.debug for request? 
 - That's my typo :(

core/server/common/src/main/java/alluxio/ServerRpcUtils.java
jiacheliu3: I began to regret immediately after I realized how much I updated a few commits later TAT
Reverted

core/server/master/src/main/java/alluxio/master/block/BlockMasterClientServiceHandler.java
jiacheliu3: Actually I thought about it and chose to make the messages more specific to each RPC, because sometimes it's hard to know where the repeated fields are. For example `BlockHeartbeatPRequest` has a list of `Metric` in the options so I have to 
do `request.getOptions().getMetricsCount()`.

I think it's kind of hard to write a generic utility method that walks the request/response recursively and report the size of  each repeated field.

ggezer: Right. We should try to avoid making this a call site change.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
ggezer: I feel like `SamplingLogger` is a better fit for these logs in general.

core/server/master/src/main/java/alluxio/master/block/BlockMasterClientServiceHandler.java
apc999: maybe you can pass a string supplier to generate the warning message for each different calls. 
I will leave this up to you

apc999 (changes requested):
last two minor comments

core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -140,6 +148,11 @@ private RpcUtils() {} // prevent instantiation
       throw new InternalException(e).toGrpcStatusException();
     } finally {
       MetricsSystem.counter(getQualifiedInProgressMetricName(methodName)).dec();
+      Instant endTime = Instant.now();
+      long timeElapsed = Duration.between(startTime, endTime).toMillis();
+      if (timeElapsed > RPC_TIME_WARNING_THESHOLD) {
```
maobaolong, line 153: If you want to calculate the cost for each rpc call, `us` would be more accurate.

From this PR, we can create a new metrics, lets say it `rpcProcessingTime`

Reference https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/ipc/metrics/RpcMetrics.java

maobaolong:
Is there any updates?

yuzhu:
@jiacheliu3 ping? is there any update on this?

jiacheliu3:
I haven't been able to spend more time on this but I will in the coming 2 weeks, i think

PR #13488, opened 2021-05-26
Implement cluster aggregation for Worker block eviction rate metric

ZhuTopher:
The previous PR #13448 was merged already, but the aggregation of the meter wasn't properly implemented.

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/BlockMetricsReporter.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/BlockMetricsReporter.java
@@ -36,12 +36,14 @@ public final class BlockMetricsReporter extends AbstractBlockStoreEventListener 
       = MetricsSystem.counter(MetricKey.WORKER_BLOCKS_DELETED.getName());
   private static final Counter BLOCKS_EVICTED
       = MetricsSystem.counter(MetricKey.WORKER_BLOCKS_EVICTED.getName());
-  private static final Meter BLOCKS_EVICTION_RATE
-      = MetricsSystem.meter(MetricKey.WORKER_BLOCKS_EVICTION_RATE.getName());
   private static final Counter BLOCKS_CANCELLED
       = MetricsSystem.counter(MetricKey.WORKER_BLOCKS_CANCELLED.getName());
   private static final Counter BLOCKS_LOST
       = MetricsSystem.counter(MetricKey.WORKER_BLOCKS_LOST.getName());
+
+  private static final Meter BLOCKS_EVICTION_RATE =
+      MetricsSystem.meterWithTags(MetricKey.WORKER_BLOCKS_EVICTION_RATE.getName(),
+        MetricKey.WORKER_BLOCKS_EVICTION_RATE.isClusterAggregated());
 
   /**
    * Creates a new instance of {@link BlockMetricsReporter}.


core/server/worker/src/main/java/alluxio/worker/block/BlockMetricsReporter.java
```diff
@@ -36,13 +36,15 @@
...
-  private static final Meter BLOCKS_EVICTION_RATE
-      = MetricsSystem.meter(MetricKey.WORKER_BLOCKS_EVICTION_RATE.getName());
   private static final Counter BLOCKS_CANCELLED
       = MetricsSystem.counter(MetricKey.WORKER_BLOCKS_CANCELLED.getName());
   private static final Counter BLOCKS_LOST
       = MetricsSystem.counter(MetricKey.WORKER_BLOCKS_LOST.getName());
 
+  private static final Meter BLOCKS_EVICTION_RATE =
+      MetricsSystem.meterWithTags(MetricKey.WORKER_BLOCKS_EVICTION_RATE.getName(),
+        MetricKey.WORKER_BLOCKS_EVICTION_RATE.isClusterAggregated());
```
ZhuTopher, line 46: @yuzhu Do you know if anything else needs to be done for cluster-aggregated meters? I didn't find anything along the lines of `registerMetersIfAbsent` in the codebase, but maybe that needs to be added?

ZhuTopher:

core/server/worker/src/main/java/alluxio/worker/block/BlockMetricsReporter.java
yuzhu: as long as `METRIC_REGISTRY.meter` is called i think we are good.

yuzhu: but you do need to trigger the meter by calling `meter.mark()`

ZhuTopher: I don't seem to be able to see the `Worker.BlockEvictionRate` metric in either master/worker sinks, but that could be due to `meter.mark()` never getting called in my tests (no blocks get evicted).

However I also looked at another metric which uses `meterWithTags`: `Worker.BytesReadUfsThroughput`
```
  public static final MetricKey WORKER_BYTES_READ_UFS_THROUGHPUT =
      new MetricKey.Builder("Worker.BytesReadUfsThroughput")
          .setDescription("Bytes read throughput from all Alluxio UFSes by this worker")
          .setMetricType(MetricType.METER)
          .build();
```
- Note that this Meter is not cluster aggregated

Here is it's definition in [BlockReadHandler](https://github.com/Alluxio/alluxio/blob/master/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java#L572)
```
        MetricKey counterKey = MetricKey.WORKER_BYTES_READ_UFS;
        MetricKey meterKey = MetricKey.WORKER_BYTES_READ_UFS_THROUGHPUT;
        context.setCounter(MetricsSystem.counterWithTags(counterKey.getName(),
            counterKey.isClusterAggregated(), MetricInfo.TAG_UFS, ufsString));
        context.setMeter(MetricsSystem.meterWithTags(meterKey.getName(),
            meterKey.isClusterAggregated(), MetricInfo.TAG_UFS, ufsString));
```

This metric (and the counter `WORKER_BYTES_READ_UFS`) does not appear at all in the worker sink.


In the master sink you see cluster aggregated versions with those names, but they are from completely separate metric definitions (the throughput is actually a Gauge which uses a separate cluster-Counter to manually calculate throughput):
```
  public static final MetricKey CLUSTER_BYTES_READ_UFS_ALL =
      new Builder("Cluster.BytesReadUfsAll")
          .setDescription("Total number of bytes read from a all Alluxio UFSes by all workers")
          .build();
  public static final MetricKey CLUSTER_BYTES_READ_UFS_THROUGHPUT =
      new Builder("Cluster.BytesReadUfsThroughput")
          .setDescription("Bytes read per minute throughput from all Alluxio UFSes by all workers")
          .build();
```
- The throughput gauge is calculated [here](https://github.com/Alluxio/alluxio/blob/master/core/server/master/src/main/java/alluxio/master/metrics/DefaultMetricsMaster.java#L142) where it calls `MetricsSystem.registerGaugeIfAbsent()`

`registerGaugeIfAbsent()` looks like this:
```
  public static synchronized <T> void registerGaugeIfAbsent(String name, Gauge<T> metric) {
    if (!METRIC_REGISTRY.getMetrics().containsKey(name)) {
      METRIC_REGISTRY.register(name, metric);
    }
  }
```

Since neither `meterWithTags()` shows up; I assume we need to call `METRIC_REGISTRY.register()` on them. Hence why I asked if we need to make a `registerMetersIfAbsent()` or equivalent.

@yuzhu What do you think?

ZhuTopher: According to @yuzhu the codepath for `WORKER_BYTES_READ_UFS_THROUGHPUT` never gets called, so that's why it never appeared in the Worker metric sink.

A quick test ensuring `meter.mark()` got called showed `Worker.BlocksEvictionRate` in the worker sink.

yuzhu (approved):

yuzhu:

PR #13498, opened 2021-05-27
Add an RPC benchmark tool

jiacheliu3:
Implements an RPC stress test benchmarking tool based on the `StressBench` framework. The tool provides a way to simulate a specified RPC load on the master. The load is specified by the number of files/blocks and concurrency. The benchmark result will be the time taken in each RPC, and a summary(stat analysis) based on all the RPC times.

One example of this tool is:
```
$ bin/alluxio runClass alluxio.stress.cli.RegisterWorkerBench --tiers "1000,1000,1000;2000,2000;5000" --concurrency 4 --cluster
{
  "mCount" : 4,
  "mTotalDurationMs" : 332,
  "mAvgDurationMs" : 83.0,
  "mPercentile5th" : 82.0,
  "mPercentile25th" : 82.0,
  "mMedian" : 83.0,
  "mPercentile75th" : 84.0,
  "mPercentile95th" : 84.0,
  "parameters" : null,
  "baseParameters" : null,
  "errors" : [ ],
  "points" : [ {
    "mDurationMs" : 84,
    "className" : "alluxio.stress.rpc.RpcTaskResult$Point"
  }, {
    "mDurationMs" : 82,
    "className" : "alluxio.stress.rpc.RpcTaskResult$Point"
  }, {
    "mDurationMs" : 84,
    "className" : "alluxio.stress.rpc.RpcTaskResult$Point"
  }, {
    "mDurationMs" : 82,
    "className" : "alluxio.stress.rpc.RpcTaskResult$Point"
  } ],
  "className" : "alluxio.stress.rpc.RpcTaskSummary"
}
```

The tool currently contains stress test on 3 RPCs:
1. RegisterWorker
2. WorkerHeartbeat
3. GetPinnedFileIds


stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  protected void afterConnect() throws IOException {
+    mClient = FileSystemMasterWorkerServiceGrpc.newBlockingStub(mChannel);
+  }
+
+  public int getPinListLength() throws IOException {
+    return retryRPC(() -> mClient.withDeadlineAfter(mContext.getClusterConf()
+            .getMs(PropertyKey.WORKER_MASTER_PERIODICAL_RPC_TIMEOUT), TimeUnit.MILLISECONDS)
+            .getPinnedFileIds(GetPinnedFileIdsPRequest.newBuilder().build())
+            .getPinnedFileIdsCount(),
```
jiacheliu3, line 48: wrong indentation: 4 spaces more than the previous line

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -163,12 +165,12 @@ private RpcTaskResult runRPC() throws Exception {
...
         pointStopwatch.start();
 
-        Set<Long> pinned = client.getPinList();
+        int numPinnedFiles = client.getPinListLength();
 
         pointStopwatch.stop();
 
-        if (pinned.size() != mParameters.mNumFiles) {
-          result.addError(String.format("Unexpected number of files: %d", pinned.size()));
+        if (numPinnedFiles != mParameters.mNumFiles) {
```
jiacheliu3, line 172: If any of your files failed to generate due to random string collision, this will be an error?

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -163,12 +165,12 @@ private RpcTaskResult runRPC() throws Exception {
...
+          result.addError(String.format("Unexpected number of files: %d", numPinnedFiles));
```
jiacheliu3, line 173: When A mismatches with B, better print both A and B

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Helper client that is used to minimizes the deserialization overhead by calling
+ * getPinnedFileIdsCount instead of getPinnedFileIdsList
+ */
+public class PinListFileSystemMasterClient extends FileSystemMasterClient {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(PinListFileSystemMasterClient.class);
+  private FileSystemMasterWorkerServiceGrpc.FileSystemMasterWorkerServiceBlockingStub mClient =
```
jiacheliu3, line 32: why do you need this `mClient`?

jiacheliu3, line 32: why do you need this `mClient`?

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
```diff
@@ -0,0 +1,51 @@
...
+import alluxio.master.MasterClientContext;
+import alluxio.worker.file.FileSystemMasterClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Helper client that is used to minimizes the deserialization overhead by calling
```
jiacheliu3, line 26: nit: grammar

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
```diff
@@ -0,0 +1,51 @@
...
+ * getPinnedFileIdsCount instead of getPinnedFileIdsList
```
jiacheliu3, line 27: further explain why you only want the count instead of the full list, and why the new `HashSet` is avoided, and why you want to avoid creating those in general

jiacheliu3:
@ggezer PTAL at the interfacing and see if the skeleton makes sense. Thanks!

At the same time we are doing:
1. Test running these RPC stress tests and understand what we wrote
2. Fixing the style and minorities

stress/shell/src/main/java/alluxio/stress/cli/GenerateBlockIdUtils.java
```diff
@@ -0,0 +1,154 @@
...
+    List<Long> list = new ArrayList<>();
+    for (long i = 0; i < count; i++) {
+      list.add(start + i);
+    }
+    return list;
+  }
+
+
+
+  public static void main(String[] args) throws Exception {
```
ggezer: This is called utils but has a main method?

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -0,0 +1,136 @@
...
+             mFileSystemContext.acquireMasterClientResource()) {
+      client.get().createDirectory(mBaseUri,
+              CreateDirectoryPOptions.newBuilder().setAllowExists(true).build());
+    }
+    int numFiles = mParameters.mNumFiles;
+    int fileNameLength = (int) Math.max(8, Math.log10(numFiles));
+
+    CompletableFuture<Void>[] futures = new CompletableFuture[numFiles];
+    LOG.info("Generating {} test files", numFiles);
+    for (int i = 0; i < numFiles; i++) {
```
ggezer, line 52: Is this a sequential step to create all pinned files?
How long does it take to run for creating X files?

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -0,0 +1,136 @@
...
+      AlluxioURI fileUri = mBaseUri.join(CommonUtils.randomAlphaNumString(fileNameLength));
+      CompletableFuture<Void> future = CompletableFuture.supplyAsync((Supplier<Void>) () -> {
+        try (CloseableResource<FileSystemMasterClient> client =
+                 mFileSystemContext.acquireMasterClientResource()) {
+          client.get().createFile(fileUri,
+              CreateFilePOptions
+                  .newBuilder()
+                  .setBlockSizeBytes(mConf.getBytes(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT))
+                  .build());
```
ggezer, line 61: setRecursive(true) for the next run?
Since you're removing the base dir below.

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -0,0 +1,136 @@
...
+        MasterClientContext.newBuilder(ClientContext.create(mConf)).build());
+    RpcTaskResult result = new RpcTaskResult();
+    result.setBaseParameters(mBaseParameters);
+    result.setParameters(mParameters);
+
+    long durationMs = FormatUtils.parseTimeSize(mParameters.mDuration);
+    Stopwatch durationStopwatch = Stopwatch.createStarted();
+    Stopwatch pointStopwatch = Stopwatch.createUnstarted();
+
+    while (durationStopwatch.elapsed(TimeUnit.MILLISECONDS) < durationMs) {
```
ggezer, line 99: what does this test do?

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,208 @@
...
+      Map<String, Long> capMap = ImmutableMap.of("MEM", cap);
+      Map<String, Long> usedMap = ImmutableMap.of("MEM", 0L);
+
+      Instant s = Instant.now();
+      client.register(workerId,
+              tierAliases,
+              capMap,
+              usedMap,
+              // Will use the prepared block list instead of converting this one
+              ImmutableMap.of(),
```
ggezer, line 153: is this why we're subclassing the client?
do we need to avoid this? Won't this effect timings?

stress/shell/src/main/java/alluxio/stress/cli/WorkerHeartbeatBench.java
```diff
@@ -0,0 +1,182 @@
...
+        LOG.error("Failed to run blockHeartbeat {}", i, e);
+        result.addError(e.getMessage());
+      }
+    }
+
+    return result;
+  }
+
+  @Override
+  public void prepare() throws Exception {
```
ggezer, line 119: nit: move it up?

ggezer (commented):
looks almost there. left some comments/questions.

stress/common/src/main/java/alluxio/stress/rpc/RpcBenchParameters.java
```diff
@@ -0,0 +1,14 @@
+package alluxio.stress.rpc;
+
+import alluxio.stress.Parameters;
+import com.beust.jcommander.Parameter;
+
+public class RpcBenchParameters extends Parameters {
+  @Parameter(names = {"--concurrency"},
+          description = "simulate this many workers on one machine")
```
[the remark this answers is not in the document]
jiacheliu3: The RPCBench should cover both worker->master and client->master RPCs. But the existing ones are all worker->master RPCs. 

Yea the description should be changed to `simulate this many workers/clients on one machine`.

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
jiacheliu3: The `prepare()` step is not included in timing so this will not affect the result precision, per se. It does take much time generating a big number of files, but based on the current API we don't have a good way to avoid that. Based on my experience running these tests, it takes a few minutes to generate a few million of files/blocks, on a 4-core/8-core master machine. So I would say the usability is okay for the current scale in testing. 

If we think of a billion files/blocks, that will probably take a beefy machine (24-core and up) and probably 1hr to prepare.

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
jiacheliu3: The principles when we are designing the tests are:
1. Generate as much pressure on the master as possible, so theoretically the only thing that the client should do is to send RPC requests (instead of doing some preprocessing and preparation on the client side each run).
2. Make the measured number as close to the master-side effort as possible. When an RPC takes X millisecond for the round time trip, we'd like the client to NOT waste time generating data in the X millisecond and NOT deserialize the data, as much as possible. 
3. Also the client-side work will take CPU and memory. When the concurrency is high (many client threads on one node), the stress test may be CPU-bounded on the client side, which makes the results meaningless. We are manually validating this as well. For example, manually testing different client-side concurrency to find a good number which introduces enough stress on master and not introduce client-side CPU bound.

Back to this particular line, this is relevant to Principle 2. We want the client to NOT spend time on converting data when doing the RPC. Same applies to `PinListFileSystemMasterClient`.

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
dbw9580: Technically yes, but the probability of such a collision is negligible, even given very large number of files (>1B). We could abort the benchmark during `prepare()` in this case, but the gain will be also very marginal.

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
dbw9580: there's one in the super class, but it's private. I need a grpc stub to do the rpc.

https://github.com/Alluxio/alluxio/blob/d5f9af6fa3de3acdb2ecbd803e18ce8003a67b64/core/server/worker/src/main/java/alluxio/worker/file/FileSystemMasterClient.java#L43-L44

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
dbw9580: Interestingly enough, I copied those lines from https://github.com/Alluxio/alluxio/blob/d5f9af6fa3de3acdb2ecbd803e18ce8003a67b64/core/server/worker/src/main/java/alluxio/worker/file/FileSystemMasterClient.java#L88-L94

dbw9580: Fixed in 6fb2ad27114d5689b74afbc20e9e325ebecd9322

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
dbw9580: 6fb2ad27114d5689b74afbc20e9e325ebecd9322

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
dbw9580: Clarified in 6fb2ad27114d5689b74afbc20e9e325ebecd9322

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class PinListFileSystemMasterClient extends FileSystemMasterClient {
```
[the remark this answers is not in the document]
dbw9580: I moved the class into `GetPinnedFileIdsBench` as an inner class, since it's unlikely there would be anywhere else where a `.getPinListLength()` call would more useful than `.getPinList().size()`.

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
dbw9580: `durationMs` is the time how long the benchmark should run, we check the running time against this to know when we should stop the benchmark.

stress/common/src/main/java/alluxio/stress/PinListFileSystemMasterClient.java
dbw9580: fixed in 6fb2ad2

stress/common/src/main/java/alluxio/stress/CachingBlockMasterClient.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Use this class to avoid map-list conversion at the client side
+ * So if you are running all processes on the same machine,
+ * the client side work affect the performance less.
+ * */
+public class CachingBlockMasterClient extends BlockMasterClient {
```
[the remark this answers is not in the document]
jiacheliu3: In `BlockMasterClient.heartbeart()` and `BlockMasterClient.register()`, the worker calls `convertBlockListMapToProto` to convert its block data structure into a proto. When we are doing a stress test (using simulated workers), this happens in every simulated RPC. I try to avoid this because now the RTT for that RPC includes a non-trivial client-side work and this noise can get very big if we are testing with more than millions of blocks.

I called this `Caching` because it "caches" the converted proto. Yea the name is not that good... I'll give it more thoughts

stress/common/pom.xml
```diff
@@ -54,5 +54,10 @@
       <artifactId>alluxio-job-common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
jiacheliu3: Is there a better place for this dependency? I do see duplicate dependencies in `stress/shell` and `stress/common` modules, but it seems we push all dependencies to the module level (same in the `core` modules). Not sure if that's what you mean.

stress/common/src/main/java/alluxio/stress/CachingBlockMasterClient.java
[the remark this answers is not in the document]
jiacheliu3: I feel like controlling `convertBlockListMapToProto` is better than probably mocking `register` and `heartbeat` methods, as the latter logic are more subject to change.

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
```diff
@@ -153,7 +153,7 @@ public long getId(final WorkerNetAddress address) throws IOException {
    * @return a flattened and deduplicated list
    */
   @VisibleForTesting
-  List<LocationBlockIdListEntry> convertBlockListMapToProto(
+  public List<LocationBlockIdListEntry> convertBlockListMapToProto(
```
jiacheliu3, line 156: Added access for `CachingBlockMasterClient` to override

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -69,15 +69,28 @@
    */
   public abstract void prepare() throws Exception;
 
+  /**
+   * Perform post-run cleanups.
+   */
+  public void cleanup() throws Exception {}
```
jiacheliu3, line 78: Added to the interface as some tests will include a cleanup phase like removing test files.

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -129,4 +131,34 @@ public GetPinnedFileIdsParameters getParameters() {
...
+    public PinListFileSystemMasterClient(MasterClientContext conf) {
+      super(conf);
+    }
+
+    @Override
+    protected void afterConnect() throws IOException {
+      mClient = FileSystemMasterWorkerServiceGrpc.newBlockingStub(mChannel);
+    }
+
+    public int getPinListLength() throws IOException {
```
ggezer, line 198: I'm still not convinced this method should be implemented here. Why not call FileSystemMasterClient.getPinlist() on the call site and call `size()` on it? That way we don't have to complicate the code by subclassing.

dbw9580: I added some comments on why I'm doing this in 6fb2ad27114d5689b74afbc20e9e325ebecd9322, which you might have missed. Could you take a look at the comments of this class?

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -0,0 +1,207 @@
...
+        int numPinnedFiles = mWorkerClient.getPinListLength();
+
+        mPointStopwatch.stop();
+
+        // TODO(bowen): assertion may fail when running in cluster mode due to
+        //  unwanted extra runs of prepare()
+        if (numPinnedFiles != mParameters.mNumFiles) {
+          result.addError(String.format("Unexpected number of files: %d, expected %d",
+              numPinnedFiles, mParameters.mNumFiles));
+          return result;
```
dbw9580, line 148: Don't abort early as in the case of an exception.

stress/common/src/main/java/alluxio/stress/rpc/RpcTaskResult.java
```diff
@@ -0,0 +1,180 @@
...
+   */
+  public RpcTaskResult() {
+    mPoints = new ArrayList<>();
+    mErrors = new ArrayList<>();
+  }
+
+  /**
+   * @return the {@link BaseParameters}
+   */
+  public BaseParameters getBaseParameters() {
```
dbw9580, line 49: Annotate with `@Nullable`. The same for `getParameters`.

stress/common/src/main/java/alluxio/stress/rpc/RpcTaskSummary.java
```diff
@@ -0,0 +1,161 @@
...
+   * @param errors the errors
+   * */
+  public void setErrors(List<String> errors) {
+    mErrors = errors;
+  }
+
+  /**
+   * @return the {@link BaseParameters}
+   * */
+  public BaseParameters getBaseParameters() {
```
dbw9580, line 137: `@Nullable`

stress/common/src/main/java/alluxio/stress/rpc/RpcTaskSummary.java
```diff
@@ -0,0 +1,161 @@
...
+   * @param baseParameters the {@link BaseParameters}
+   * */
+  public void setBaseParameters(BaseParameters baseParameters) {
+    mBaseParameters = baseParameters;
+  }
+
+  /**
+   * @return the task specific {@link UfsIOParameters}
+   * */
+  public Parameters getParameters() {
```
dbw9580, line 151: `@Nullable`

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,224 @@
...
+      aliases.add(TIER_ALIASES.get(i));
+    }
+    return aliases;
+  }
+
+  private Map<String, Long> simulateCapacityMap(List<String> tierAliases) {
+    Map<String, Long> capMap = new HashMap<>();
+    for (String t : tierAliases) {
+      capMap.put(t, CAPACITY);
+    }
```
dbw9580, line 149: More concise and readable using Guava Maps:

`Maps.toMap(tierAliases, (tier) -> CAPACITY)`

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,224 @@
...
+    return result;
+  }
+
+  private List<String> simulateTierAliases(String tierConfig) {
+    String[] parts = tierConfig.split(";");
+    LOG.info("Simulate {} tiers with config {}", parts.length, tierConfig);
+    List<String> aliases = new ArrayList<>();
+    for (int i = 0; i < parts.length; i++) {
+      aliases.add(TIER_ALIASES.get(i));
+    }
```
dbw9580, line 141: `TIER_ALIASES.subList(0, parts.length)`

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,224 @@
...
+      capMap.put(t, CAPACITY);
+    }
+    return capMap;
+  }
+
+  private Map<String, Long> simulateUsedMap(List<String> tierAliases) {
+    Map<String, Long> usedMap = new HashMap<>();
+    for (String t : tierAliases) {
+      usedMap.put(t, 0L);
+    }
```
dbw9580, line 157: `Maps.toMap(tierAliases, (tier) -> 0L)`

dbw9580 (changes requested):
Thanks for the hard work! Left a bunch of comments. Let's find a time to discuss some of the commented issues. @jiacheliu3

ggezer (approved):
Thanks!

stress/common/src/main/java/alluxio/stress/rpc/RpcBenchParameters.java
```diff
@@ -0,0 +1,29 @@
...
+ * Parameters for the RPC benchmark test.
+ */
+public class RpcBenchParameters extends Parameters {
+  @Parameter(names = {"--concurrency"},
+      description = "simulate this many clients/workers on one machine")
+  public int mConcurrency = 2;
+
+  @Parameter(names = {"--duration"},
+      description = "The length of time to run the benchmark. (1m, 10m, 60s, 10000ms, etc.)")
+  public String mDuration = "30s";
```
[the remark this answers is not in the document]
jiacheliu3: Sounds good

stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
```diff
@@ -67,17 +67,33 @@
   /**
    * Prepares to run the test.
    */
+  // TODO(bowen): When the test runs in cluster mode, the prepare step will execute
+  //  both in the command side and on each job worker side. We should separate the logic
+  //  into two different calls instead of relying on the same prepare().
   public abstract void prepare() throws Exception;
 
+  /**
+   * Perform post-run cleanups.
```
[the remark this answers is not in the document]
jiacheliu3: Sounds good

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -0,0 +1,207 @@
...
+ *    --cluster-limit 2 --num-files 100000 --duration 5s
+ */
+public class GetPinnedFileIdsBench extends RpcBench<GetPinnedFileIdsParameters> {
+  private static final Logger LOG = LoggerFactory.getLogger(GetPinnedFileIdsBench.class);
+
+  @ParametersDelegate
+  private final GetPinnedFileIdsParameters mParameters = new GetPinnedFileIdsParameters();
+
+  private final InstancedConfiguration mConf = InstancedConfiguration.defaults();
+  private final FileSystemContext mFileSystemContext = FileSystemContext.create(mConf);
```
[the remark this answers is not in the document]
jiacheliu3: I don't think these are static because they better belong to the object instead of the class

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,224 @@
...
+  private static final Map<String, Long> USED_MEM_EMPTY = ImmutableMap.of("MEM", 0L);
+  private static final List<String> TIER_ALIASES = ImmutableList.of("MEM", "SSD", "HDD");
+  private static final Map<String, List<String>> LOST_STORAGE =
+      ImmutableMap.of("MEM", new ArrayList<>());
+  private static final List<ConfigProperty> EMPTY_CONFIG = ImmutableList.of();
+
+  @ParametersDelegate
+  private BlockMasterBenchParameters mParameters = new BlockMasterBenchParameters();
+
+  private final InstancedConfiguration mConf = InstancedConfiguration.defaults();
```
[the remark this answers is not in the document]
jiacheliu3: Same argument. I think this fits better as a field of an object, instead of a property of a class. This is purely philosophical though....

stress/shell/src/main/java/alluxio/stress/cli/RpcBenchPreparationUtils.java
```diff
@@ -0,0 +1,190 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(RpcBenchPreparationUtils.class);
+
+  private static final String[] TIER_NAMES = new String[]{"MEM", "SSD", "HDD"};
+  private static final long CAPACITY = 20L * 1024 * 1024 * 1024; // 20GB
+  private static final Map<String, Long> CAPACITY_MEM = ImmutableMap.of("MEM", CAPACITY);
+  private static final Map<String, Long> USED_MEM_EMPTY = ImmutableMap.of("MEM", 0L);
+  private static final BlockStoreLocation BLOCK_LOCATION_MEM =
+      new BlockStoreLocation("MEM", 0, "MEM");
+
+  private static InstancedConfiguration sConf = InstancedConfiguration.defaults();
```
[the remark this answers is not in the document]
jiacheliu3: Didn't make it final as I did not want to change the name to `CONF`, which is more like a constant value. I'm not sure if there are implicit rules behind but seems our code prefers `sConf` slightly...

stress/shell/src/main/java/alluxio/stress/cli/RpcBenchPreparationUtils.java
```diff
@@ -0,0 +1,190 @@
...
+
+    return blockMap;
+  }
+
+  private static List<Long> generateDecreasingNumbers(long start, long count) {
+    LOG.info("Generating block Ids ({}, {}]", start - count, start);
+    List<Long> list = new ArrayList<>();
+    for (long i = 0; i < count; i++) {
+      list.add(start - i);
+    }
```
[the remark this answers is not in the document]
jiacheliu3: Good question. I believe `int` will just suffice for count. Fixed.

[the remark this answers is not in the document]
jiacheliu3: I don't want to support a block count larger than `Integer.MAX`. The `prepare()` will not finish in a day anyways...

stress/shell/src/main/java/alluxio/stress/cli/WorkerHeartbeatBench.java
```diff
@@ -0,0 +1,192 @@
...
+      new BlockStoreLocation("MEM", 0, "MEM");
+  private static final Map<String, List<String>> LOST_STORAGE =
+      ImmutableMap.of("MEM", new ArrayList<>());
+  private static final List<Metric> EMPTY_METRICS = ImmutableList.of();
+  private static final List<Long> EMPTY_REMOVED_BLOCKS = ImmutableList.of();
+
+  @ParametersDelegate
+  private BlockMasterBenchParameters mParameters = new BlockMasterBenchParameters();
+
+  private final InstancedConfiguration mConf = InstancedConfiguration.defaults();
```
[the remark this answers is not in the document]
jiacheliu3: Same argument as above

stress/common/src/main/java/alluxio/stress/rpc/BlockMasterBenchParameters.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Parameters for generating different stress on the BlockMaster.
+ */
+public class BlockMasterBenchParameters extends RpcBenchParameters {
+  @Parameter(names = {"--tiers"},
+      description = "The number of blocks in each storage dir and tier. "
+          + "Use semi-colon to separate tiers, use commas to separate dirs. "
+          + "Example: \"100,200,300;1000,1500;2000\"")
+  public String mTiers;
```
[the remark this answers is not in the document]
dbw9580: Fixed in 11be64bcf32899d3faf94e3968529219c568f7be

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+    return usedMap;
+  }
+
+  private void runOnce(alluxio.worker.block.BlockMasterClient client,
+                       RpcTaskResult result, long i, long workerId) {
+    List<String> tierAliases = simulateTierAliases(mParameters.mTiers);
+    LOG.info("Tier aliases: {}", tierAliases);
+    Map<String, Long> capacityMap = simulateCapacityMap(tierAliases);
+    Map<String, Long> usedMap = simulateUsedMap(tierAliases);
```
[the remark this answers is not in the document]
dbw9580: Fixed in 11be64bcf32899d3faf94e3968529219c568f7be

stress/shell/src/main/java/alluxio/stress/cli/RegisterWorkerBench.java
```diff
@@ -0,0 +1,224 @@
...
+    Instant endTime = startTime.plus(durationMs, ChronoUnit.MILLIS);
+    LOG.info("Start time {}, end time {}", startTime, endTime);
+
+    RpcTaskResult result = new RpcTaskResult();
+    result.setBaseParameters(mBaseParameters);
+    result.setParameters(mParameters);
+
+    RpcTaskResult taskResult = simulateRegisterWorker(client);
+    LOG.info("Received task result {}", taskResult);
+    result.merge(taskResult);
```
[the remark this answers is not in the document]
dbw9580: Fixed in 5afed843cf8ce615f60bc752abd5640aa13fcfc4

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
dbw9580: Fixed in 5e19bbe56c4d8e6826c6dff58fee87582d81071a

dbw9580:
I was thinking of making the time resolution configurable: instead of storing a `long` in `Point`, store a `Duration`, and convert to milliseconds or nanoseconds or whatever when exporting the result. However, I was stuck at the point where I want to do the conversion by calling `Duration.get()` method, which surprisingly turned out to not support `ChronoUnit.MILLIS`. Anyways, this improvement is good to have, but the gain is close to marginal. Just FYI, @jiacheliu3.

stress/shell/src/main/java/alluxio/stress/cli/RpcBenchPreparationUtils.java
```diff
@@ -0,0 +1,190 @@
...
+    LOG.info("Collect all results");
+    CompletableFuture.allOf(futures).join();
+  }
+
+  /**
+   * Reserves a list of worker IDs from the master.
+   */
+  static Deque<Long> prepareWorkerIds(BlockMasterClient client, int numWorkers) throws IOException {
+    Deque<Long> workerPool = new ArrayDeque<>();
+    int freePort = 40000;
```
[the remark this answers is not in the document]
jiacheliu3: Good question. I believe the port numbers here are not really allocating ports from the OS so there's no limit like 65535 or `ulimit`. The real upper limit is the max of integer which is... a little far away from real :)

stress/common/src/main/java/alluxio/stress/rpc/RpcTaskResult.java
```diff
@@ -0,0 +1,180 @@
...
+  private Parameters mParameters;
+
+  /**
+   * Constructor.
+   */
+  public RpcTaskResult() {
+    mPoints = new ArrayList<>();
+    mErrors = new ArrayList<>();
+  }
+
```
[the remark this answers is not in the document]
jiacheliu3: good call

stress/shell/src/main/java/alluxio/stress/cli/RpcBench.java
```diff
@@ -0,0 +1,118 @@
...
+      // Collect the result
+      CompletableFuture[] cfs = futures.toArray(new CompletableFuture[0]);
+      List<RpcTaskResult> results = CompletableFuture.allOf(cfs)
+              .thenApply(f -> futures.stream()
+                      .map(CompletableFuture::join)
+                      .collect(Collectors.toList())
+              ).get();
+      LOG.info("{} futures collected: {}", results.size(),
+              results.size() > 0 ? results.get(0) : "[]");
+      return RpcTaskResult.Aggregator.reduceList(results);
```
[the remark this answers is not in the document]
jiacheliu3: yea you know more about lambdas than me

stress/common/src/main/java/alluxio/stress/rpc/RpcTaskResult.java
```diff
@@ -0,0 +1,180 @@
...
+     * Reduce a list of {@link RpcTaskResult} into one.
+     *
+     * @param results a list of results to combine
+     * @return the combined result
+     * */
+    public static RpcTaskResult reduceList(Iterable<RpcTaskResult> results) {
+      RpcTaskResult aggreResult = new RpcTaskResult();
+      for (RpcTaskResult r : results) {
+        aggreResult.merge(r);
+      }
```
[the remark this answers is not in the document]
dbw9580: I added the missing parameters during merge in cda33124db52d1b3f1a5ed331ce7a1fcfd2988f3. If this is not what you want, feel free to revert the commit @jiacheliu3

jiacheliu3:
@dbw9580 Thanks for the improvements! Merging as all tests look good now and all comments are resolved.

jiacheliu3:

PR #13506, opened 2021-05-31
Add a option in distributedLoad job to exclude some workers

### LINKED ISSUE #13505: DistributedLoad Job add a option to exclude some jobWorker
Add option in distributedLoad job: --exclude-hosts and --exclude-hosts-file, 
distributedLoad [--replication <num>] [--active-jobs <num>] [--index] [--hosts <host1,host2,...,hostN>] [--host-file <hostFilePath>] [--exclude-hosts <exclHost1,exclHost2,...,exclHostN>] [--exclude-hosts-file <exclHostFilePath>] [--locality <locality1,locality2,...,localityN>] [--locality-file <localityFilePath>] <path>

kuszz:
Fix #13505


maobaolong (commented):
Before insight review, I have a high-level concern about that this PR involve two new options.

Do you think it would be better if we use a `!` as exclude magic leading char to identify this host is a excluded host?

kuszz:
> Before insight review, I have a high-level concern about that this PR involve two new options.
> 
> Do you think it would be better if we use a `!` as exclude magic leading char to identify this host is a excluded host?

Thanks for your great suggestion, I will use ! as exclude magic leading in the next commit.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -157,53 +159,64 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
 
-  private void readItemsFromOptionString(Set<String> localityIds,
+  private void readItemsFromOptionString(Set<String> localityIds, Set<String> excludedSet,
       String argOption) {
     for (String locality : StringUtils.split(argOption, ",")) {
       locality = locality.trim().toUpperCase();
       if (!locality.isEmpty()) {
-        localityIds.add(locality);
+        if (locality.charAt(0) == '!' && excludedSet != null) {
+          excludedSet.add(locality);
```
maobaolong, line 201: if the `locality` starts with `!` , than you added into the excludeSet is started with `!`, so I guess you want to remove the magic char before put it into the excludedSet?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -157,53 +159,64 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
+      readItemsFromOptionString(workerSet, excludeWorkerSet, argOption);
     }
     if (cl.hasOption(LOCALITY_FILE_OPTION.getLongOpt())) {
       String localityFile = cl.getOptionValue(LOCALITY_FILE_OPTION.getLongOpt()).trim();
-      readLinesToSet(localityIds, localityFile);
+      readLinesToSet(localityIds, null, localityFile);
     } else if (cl.hasOption(LOCALITY_OPTION.getLongOpt())) {
       String argOption = cl.getOptionValue(LOCALITY_OPTION.getLongOpt()).trim();
-      readItemsFromOptionString(localityIds, argOption);
+      readItemsFromOptionString(localityIds, null, argOption);
```
maobaolong, line 176: For the LOCALITY_OPTION, we can also exclude the `localityId` which starts with `!`

maobaolong (changes requested):
Left a couple a comments.

job/server/src/main/java/alluxio/job/plan/load/LoadDefinition.java
```diff
@@ -74,22 +74,37 @@ public LoadDefinition() {
...
         boolean match = false;
-        if (!isEmptySet(config.getLocalityIds())) {
-          if (worker.getNetAddress().getTieredIdentity().getTiers() != null) {
+        if (worker.getNetAddress().getTieredIdentity().getTiers() != null) {
+          if (!(isEmptySet(config.getLocalityIds())
+              && isEmptySet(config.getExcludeLocalityIds()))) {
+            boolean exclude = false;
             for (LocalityTier tier : worker.getNetAddress().getTieredIdentity().getTiers()) {
-              if (config.getLocalityIds().contains(tier.getValue().toUpperCase())) {
+              if (!isEmptySet(config.getExcludeLocalityIds())
```
maobaolong, line 89: L86 `&& isEmptySet(config.getExcludeLocalityIds()` means `config.getExcludeLocalityIds()` is empty,
so this check must be false right?

LuQQiu:
@kuszz @maobaolong The feature is great! The only thing I am worried about is the backward compatibility. Basically unless Alluxio jumps from 2.5(2.6) directly to 3.0 which is the major version upgrade, in all other cases we need to make sure the changes don't break backward compatibility. 

does the older version Job master works with the newer version job worker?
does the older version job worker works with the newer version job master?
does the newer version job services work with the older version Alluxio master/workers?
does the newer version Alluxio servers work with the older version job services?

Can you help figure out the answers to the previous questions, appreciate!

kuszz:
> @kuszz @maobaolong The feature is great! The only thing I am worried about is the backward compatibility. Basically unless Alluxio jumps from 2.5(2.6) directly to 3.0 which is the major version upgrade, in all other cases we need to make sure the changes don't break backward compatibility.
> 
> does the older version Job master works with the newer version job worker?
> does the older version job worker works with the newer version job master?
> does the newer version job services work with the older version Alluxio master/workers?
> does the newer version Alluxio servers work with the older version job services?
> 
> Can you help figure out the answers to the previous questions, appreciate!

Thanks for pointing out  backward compatibility question, the answers of question 1,2 are no, and 3,4 are yes.
Master and worker should have same LoadConfig, and commit not involve question 3,4

kuszz:
> @kuszz @maobaolong The feature is great! The only thing I am worried about is the backward compatibility. Basically unless Alluxio jumps from 2.5(2.6) directly to 3.0 which is the major version upgrade, in all other cases we need to make sure the changes don't break backward compatibility.
> 
> does the older version Job master works with the newer version job worker?
> does the older version job worker works with the newer version job master?
> does the newer version job services work with the older version Alluxio master/workers?
> does the newer version Alluxio servers work with the older version job services?
> 
> Can you help figure out the answers to the previous questions, appreciate!

Thanks for pointing out backward compatibility question, the answer may be no.
The commit does not involve question 3,4, and the question 1,2 can be considered as below:
>
> does the older version Job master works with the newer version job worker?  -> Yes
      > the shell is older version -> Yes, the LoadConfig has serialVersionUID, the serialize and deserialize are not affected, and it always check empty before use set in config
      > the shell is newer version -> Yes, same as above
> does the older version job worker works with the newer version job master? -> No
      > the shell is older version -> Yes, same as above
      > the shell is newer version -> No, the exclude func will not take effect

LuQQiu:
We can assume job master/worker will be of the same version which is not that important when considering backward compatibility. I am more worried about the client/server in different versions issue. Can you explain more about the shell and job servers have different versions case? I am a little confused for this part

kuszz:
> We can assume job master/worker will be of the same version which is not that important when considering backward compatibility. I am more worried about the client/server in different versions issue. Can you explain more about the shell and job servers have different versions case? I am a little confused for this part

The client/server question involve two process, the answer is not affect
>
> RetryHandlingJobMasterClient send config to JobMasterClientServiceHandler: The config transmit between client and handler as JobConfig rather than LoadConfig, it don't affect, and master then get it from handler, transform it to LoadConfig and select worker.
> RetryHandlingJobMasterClient ask config from JobMasterWorkerServiceHandler: Same as above, and then worker get JobConfig, transform it to LoadConfig and execute task.

LuQQiu:
@kuszz if i understand correctly, the changes in this PR only affects the communications between job master and job worker. If job servers are in different versions (which is rare), there might be some issues. But if job servers are in the same version, The communications between client and job servers remain unchanged. This PR will not change the client behaviors. Can you confirm that

kuszz:
> @kuszz if i understand correctly, the changes in this PR only affects the communications between job master and job worker. If job servers are in different versions (which is rare), there might be some issues. But if job servers are in the same version, The communications between client and job servers remain unchanged. This PR will not change the client behaviors. Can you confirm that

Yes, the client and server don't care the config content whether changed or not, beacuse they always treat config as JobConfig rather than LoadConfig(LoadConfig implements Jobconfig), and transmit config in byte, only master and worker use config as LoadConfig.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -78,7 +78,8 @@
           .hasArg(true)
           .numberOfArgs(1)
           .argName("hosts")
-          .desc("A list of worker hosts separated by comma")
+          .desc("A list of worker hosts separated by comma, and add ! before host"
```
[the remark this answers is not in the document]
LuQQiu: I looked into the select worker logics, it's a little complicated. 
Basically, 
if nothing is specified in workers, we choose all.
If a worker is specified, it's choosen
if a worker is specified in excluded list, it will not be chosen for sure.

Then I would recommend adding the excludedHosts option and the excludedLocality option, which should be separated from the current hosts, locality options. Add parameter description saying that when hosts is empty, all hosts will be selected unless explicitly excluded by setting excludedHosts. hosts and excludedHosts should not be set together. Same for locality. 

The example ` [--hosts <host1,host2,...,hostN>] [--locality <locality1,locality2,...,localityN>]` is not a typical use case.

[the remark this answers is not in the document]
LuQQiu: Those changes are recommended for user usability and understability

LuQQiu (changes requested):
Thanks for giving the explanations, left some comments

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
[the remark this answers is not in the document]
kuszz: Thansk, I will modify code in the next commit, and I see it have <--host-file> option, should I set <--excluded-host-file>?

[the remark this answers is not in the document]
LuQQiu: Sounds good, we can also add that option.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -78,7 +78,9 @@
           .hasArg(true)
           .numberOfArgs(1)
           .argName("hosts")
-          .desc("A list of worker hosts separated by comma")
+          .desc("A list of worker hosts separated by comma, when host and locality lists are empty,"
+              + " all hosts will be selected unless excluded by setting excluded-hosts or"
```
LuQQiu, line 82: Can you help add one line to all the hosts related options:
Only one of the 'hosts', 'host-file', 'excluded-hosts', and 'excluded-host-file' should be set.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -87,7 +89,28 @@
           .hasArg(true)
           .numberOfArgs(1)
           .argName("host-file")
-          .desc("Host File contains worker hosts, each line has a worker host")
+          .desc("Host File contains worker hosts, each line has a worker host, if empty,"
```
LuQQiu, line 92: worker host. If empty,

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -87,7 +89,28 @@
...
+              + " all hosts will be selected unless explicitly excluded by setting excluded-hosts")
```
LuQQiu, line 93: excluded-hosts or excluded-host-file

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -87,7 +89,28 @@
...
+          .build();
+  private static final Option EXCLUDED_HOSTS_OPTION =
+      Option.builder()
+          .longOpt("excluded-hosts")
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .argName("excluded-hosts")
+          .desc("A list of excluded worker hosts separated by comma,"
+              + " it should not be set together with hosts")
```
LuQQiu, line 103: Only one of the 'hosts', 'host-file', 'excluded-hosts', and 'excluded-host-file' should be set.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -87,7 +89,28 @@
...
+          .build();
+  private static final Option EXCLUDED_HOST_FILE_OPTION =
+      Option.builder()
+          .longOpt("excluded-host-file")
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .argName("excluded-host-file")
+          .desc("Host File contains excluded worker hosts, each line has a worker host,"
```
LuQQiu, line 112: has a worker host. Only one of the 'hosts', 'host-file', 'excluded-hosts', and 'excluded-host-file' should be set.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -105,8 +131,29 @@
...
           .desc("Locality File contains worker localities, each line has a worker locality")
           .build();
+  private static final Option EXCLUDED_LOCALITY_OPTION =
+      Option.builder()
+          .longOpt("excluded-locality")
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .argName("excluded-locality")
+          .desc("A list of excluded worker locality separated by comma,"
```
LuQQiu, line 144: by comma. It should not be

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -105,8 +131,29 @@
...
+              + "it should not be set together with locality")
+          .build();
+  private static final Option EXCLUDED_LOCALITY_FILE_OPTION =
+      Option.builder()
+          .longOpt("excluded-locality-file")
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .argName("excluded-locality-file")
+          .desc("Locality File contains excluded worker localities,"
```
LuQQiu, line 154: a worker locality. It should

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
LuQQiu: Add the following line to all the locality related options:
Only one of the 'locality', 'locality-file', 'excluded-locality', and 'excluded-locality-file' should be set.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
```diff
@@ -78,7 +78,9 @@
           .hasArg(true)
           .numberOfArgs(1)
           .argName("hosts")
-          .desc("A list of worker hosts separated by comma")
+          .desc("A list of worker hosts separated by comma, when host and locality lists are empty,"
+              + " all hosts will be selected unless excluded by setting excluded-hosts or"
+              + " excluded-locality")
```
LuQQiu, line 83: What's the relationship between the hosts options and locality options? are they actually affect each other?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadUtils.java
```diff
@@ -45,12 +45,16 @@ private DistributedLoadUtils() {} // prevent instantiation
    * @param filePath The {@link AlluxioURI} path to load into Alluxio memory
    * @param replication Number of block replicas of each loaded file
    * @param workerSet A set of worker hosts to load data
+   * @param excludeWorkerSet A set of worker hosts can not to load data
```
LuQQiu, line 48: A set of worker hosts excluded from the loading job

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadUtils.java
```diff
@@ -45,12 +45,16 @@ private DistributedLoadUtils() {} // prevent instantiation
...
    * @param localityIds The locality identify set
+   * @param excludeLocalityIds A set of worker locality identify can not to load data
```
LuQQiu, line 50: A set of worker locality identity excluded from the loading job

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadUtils.java
```diff
@@ -62,22 +66,26 @@ public static void distributedLoad(AbstractDistributedJobCommand command,
    * @param filePath The {@link AlluxioURI} path to load into Alluxio memory
    * @param replication Number of block replicas of each loaded file
    * @param workerSet A set of worker hosts to load data
+   * @param excludeWorkerSet A set of worker hosts can not to load data
```
LuQQiu, line 69: Same here

LuQQiu (commented):
Thanks for modify the distributed options, left some comments

HelloHorizon (changes requested):
@kuszz Thanks for the improvement! left some comments. BTW, can you also update the distributedLoad section in `docs/en/operation/User-CLI.md`?

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadCommand.java
kuszz: No, they don't affect each other. When they are not set, all worker will be chosen.

job/server/src/test/java/alluxio/job/plan/load/LoadDefinitionTest.java
```diff
@@ -195,41 +195,65 @@ public void loadedBySpecifiedLocalityIdentity() throws Exception {
...
-  private void loadedBySpecifiedHost(Set<String> workerSet,
-      Set<String> localityIds, Set<Long> workerIds) throws Exception {
+  private void loadedBySpecifiedHost(Set<String> workerSet, Set<String> excludeWorkerSet,
+      Set<String> localityIds, Set<String> exclLocalityIds, Set<Long> workerIds) throws Exception {
     int numBlocks = 10;
     createFileWithNoLocations(TEST_URI, numBlocks);
-    LoadConfig config = new LoadConfig(TEST_URI, 1,
-        workerSet, localityIds);
+    LoadConfig config = new LoadConfig(TEST_URI, 1, workerSet, excludeWorkerSet,
+        localityIds, exclLocalityIds);
```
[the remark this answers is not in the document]
kuszz: Thanks for suggest, I have modified the code.

shell/src/main/java/alluxio/cli/fs/command/DistributedLoadUtils.java
kuszz: Thanks for suggest, I have modified the code.

LuQQiu (approved):
LGTM, thanks for contributing this useful feature!

@HelloHorizon PTAL, thanks!

LuQQiu:
Ping @HelloHorizon @apc999 PTAL, thanks!

docs/en/operation/User-CLI.md
```diff
@@ -842,6 +842,12 @@ Later jobs must wait until some earlier jobs to finish. The default value is `30
...
 * `--hosts`: Specifies a list of worker hosts separated by comma to load target data.
+* `--excluded-host-file <host-file>`: Specifies a file contains worker hosts which shouldn't load target data, each line has a worker host.
+* `--excluded-hosts`: Specifies a list of worker hosts separated by comma which shouldn't load target data.
+* `--locality-file <locality-file>`: Specifies a file contains worker locality to load target data, each line has a locality.
+* `--locality`: Specifies a list of worker locality separated by comma to load target data.
+* `--excluded-locality-file <locality-file>`: Specifies a file contains worker locality which shouldn't load target data, each line has a worker locality.
+* `--excluded-locality`: Specifies a list of worker locality separated by comma which shouldn't load target data.
 * `--index`: Specifies a file that lists all files to be loaded
 
 ```console
```
HelloHorizon, line 853: (nit)Overall looks good to me, do you mind also add one example here with new options?

HelloHorizon (approved):

LuQQiu:
@apc999 you mentioned that you want to take a final around of look for PRs related to user experience, PTAL!

docs/en/operation/User-CLI.md
kuszz: distributedLoad --replication 2 --active-jobs 2000 --excluded-hosts <host > --excluded-locality <workerTierIdentity > /data/today

apc999 (approved):

apc999:

PR #13514, opened 2021-05-31
Support update config

### LINKED ISSUE #13513: Dynamically update the Configuration
Without this feature, any configuration change affected have to through restart the service.

Create an API to handle the property changed event.

**Urgency**
Normal

maobaolong:
Fix #13513

bin/alluxio fsadmin updateConfig alluxio.master.worker.timeout=4min

image

image

Now support dynamic update properties have been tested are:

- alluxio.master.unsafe.direct.persist.object.enabled
- alluxio.master.worker.timeout
- alluxio.master.audit.logging.enabled
- alluxio.master.ufs.managed.blocking.enabled
- alluxio.master.metastore.inode.inherit.owner.and.group

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -81,6 +81,7 @@ public interface MetaMasterConfigClient extends Client {
    * Update properties.
    *
    * @param propertiesMap the properties map to be updated
+   * @return the update properties status map
    */
   Map<String, UpdatePropertyPStatus> updateConfiguration(
       Map<String, String> propertiesMap) throws IOException;
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -75,4 +75,11 @@ public interface MetaMasterConfigClient extends Client {
    * @param path the path
    */
   void removePathConfiguration(AlluxioURI path) throws IOException;
+
+  /**
+   * Update properties.
+   *
+   * @param propertiesMap the properties map to be updated
+   */
+  void updateConfig(Map<String, String> propertiesMap) throws IOException;
 }
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -15,7 +15,6 @@ import alluxio.AlluxioURI;
 import alluxio.Client;
 import alluxio.conf.PropertyKey;
 import alluxio.grpc.GetConfigurationPOptions;
-import alluxio.grpc.UpdateConfigurationPResponse.UpdatePropertyPStatus;
 import alluxio.wire.ConfigHash;
 import alluxio.wire.Configuration;
 
@@ -83,6 +82,6 @@ public interface MetaMasterConfigClient extends Client {
    * @param propertiesMap the properties map to be updated
    * @return the update properties status map
    */
-  Map<String, UpdatePropertyPStatus> updateConfiguration(
+  Map<String, Boolean> updateConfiguration(
       Map<String, String> propertiesMap) throws IOException;
 }
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -83,5 +83,5 @@ public interface MetaMasterConfigClient extends Client {
    * @return the update properties status map
    */
   Map<PropertyKey, Boolean> updateConfiguration(
-      Map<String, String> propertiesMap) throws IOException;
+      Map<PropertyKey, String> propertiesMap) throws IOException;
 }
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -82,6 +82,6 @@ public interface MetaMasterConfigClient extends Client {
    * @param propertiesMap the properties map to be updated
    * @return the update properties status map
    */
-  Map<String, Boolean> updateConfiguration(
+  Map<PropertyKey, Boolean> updateConfiguration(
       Map<String, String> propertiesMap) throws IOException;
 }
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -15,6 +15,7 @@ import alluxio.AlluxioURI;
 import alluxio.Client;
 import alluxio.conf.PropertyKey;
 import alluxio.grpc.GetConfigurationPOptions;
+import alluxio.grpc.UpdateConfigurationPResponse.UpdatePropertyPStatus;
 import alluxio.wire.ConfigHash;
 import alluxio.wire.Configuration;
 
@@ -81,5 +82,6 @@ public interface MetaMasterConfigClient extends Client {
    *
    * @param propertiesMap the properties map to be updated
    */
-  void updateConfig(Map<String, String> propertiesMap) throws IOException;
+  Map<String, UpdatePropertyPStatus> updateConfiguration(
+      Map<String, String> propertiesMap) throws IOException;
 }
--- a/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
@@ -77,7 +77,7 @@ public interface MetaMasterConfigClient extends Client {
   void removePathConfiguration(AlluxioURI path) throws IOException;
 
   /**
-   * Update properties.
+   * Updates properties.
    *
    * @param propertiesMap the properties map to be updated
    * @return the update properties status map
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -309,6 +309,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.ALL)
           .build();
+  public static final PropertyKey CONF_DYNAMIC_UPDATE_ENABLED =
+      new Builder(Name.CONF_DYNAMIC_UPDATE_ENABLED)
+          .setDefaultValue(false)
+          .setDescription("Whether to support dynamic update property.")
+          .setIgnoredSiteProperty(true)
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.ALL)
+          .build();
   public static final PropertyKey CONF_VALIDATION_ENABLED =
       new Builder(Name.CONF_VALIDATION_ENABLED)
           .setDefaultValue(true)
@@ -5259,6 +5267,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   @ThreadSafe
   public static final class Name {
     public static final String CONF_DIR = "alluxio.conf.dir";
+    public static final String CONF_DYNAMIC_UPDATE_ENABLED =
+        "alluxio.conf.dynamic.update.enabled";
     public static final String CONF_VALIDATION_ENABLED = "alluxio.conf.validation.enabled";
     public static final String DEBUG = "alluxio.debug";
     public static final String EXTENSIONS_DIR = "alluxio.extensions.dir";
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -313,7 +313,6 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       new Builder(Name.CONF_DYNAMIC_UPDATE_ENABLED)
           .setDefaultValue(false)
           .setDescription("Whether to support dynamic update property.")
-          .setIgnoredSiteProperty(true)
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.ALL)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -124,6 +124,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     private ConsistencyCheckLevel mConsistencyCheckLevel = ConsistencyCheckLevel.IGNORE;
     private Scope mScope = Scope.ALL;
     private DisplayType mDisplayType = DisplayType.DEFAULT;
+    private boolean mDynamic = true;
 
     /**
      * @param name name of the property
@@ -250,6 +251,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     }
 
     /**
+     * @param dynamic whether the property could be updated dynamically
+     * @return the updated builder instance
+     */
+    public Builder setDynamic(boolean dynamic) {
+      mDynamic = dynamic;
+      return this;
+    }
+
+    /**
      * Creates and registers the property key.
      *
      * @return the created property key instance
@@ -277,7 +287,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
 
       PropertyKey key = new PropertyKey(mName, mDescription, defaultSupplier, mAlias,
           mIgnoredSiteProperty, mIsHidden, mConsistencyCheckLevel, mScope, mDisplayType,
-          mIsBuiltIn);
+          mIsBuiltIn, mDynamic);
       return key;
     }
 
@@ -4823,6 +4833,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               + "authentication is enabled. Server trusts whoever the client claims to be.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.ALL)
+          .setDynamic(false)
           .build();
   public static final PropertyKey SECURITY_AUTHORIZATION_PERMISSION_ENABLED =
       new Builder(Name.SECURITY_AUTHORIZATION_PERMISSION_ENABLED)
@@ -4830,6 +4841,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDescription("Whether to enable access control based on file permission.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.ALL)
+          .setDynamic(false)
           .build();
   public static final PropertyKey SECURITY_AUTHORIZATION_PERMISSION_SUPERGROUP =
       new Builder(Name.SECURITY_AUTHORIZATION_PERMISSION_SUPERGROUP)
@@ -6598,6 +6610,9 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   /** The displayType which indicates how the property value should be displayed. **/
   private final DisplayType mDisplayType;
 
+  /** Whether the property could be updated dynamically. */
+  private final boolean mDynamic;
+
   /**
    * @param name String of this property
    * @param description String description of this property key
@@ -6614,7 +6629,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   private PropertyKey(String name, String description, DefaultSupplier defaultSupplier,
       String[] aliases, boolean ignoredSiteProperty, boolean isHidden,
       ConsistencyCheckLevel consistencyCheckLevel, Scope scope, DisplayType displayType,
-      boolean isBuiltIn) {
+      boolean isBuiltIn, boolean dynamic) {
     mName = Preconditions.checkNotNull(name, "name");
     // TODO(binfan): null check after we add description for each property key
     mDescription = Strings.isNullOrEmpty(description) ? "N/A" : description;
@@ -6626,6 +6641,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     mScope = scope;
     mDisplayType = displayType;
     mIsBuiltIn = isBuiltIn;
+    mDynamic = dynamic;
   }
 
   /**
@@ -6633,7 +6649,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
    */
   private PropertyKey(String name) {
     this(name, null, new DefaultSupplier(() -> null, "null"), null, false, false,
-        ConsistencyCheckLevel.IGNORE, Scope.ALL, DisplayType.DEFAULT, true);
+        ConsistencyCheckLevel.IGNORE, Scope.ALL, DisplayType.DEFAULT, true, true);
   }
 
   /**
@@ -6781,6 +6797,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
    */
   public boolean isBuiltIn() {
     return mIsBuiltIn;
+  }
+
+  /**
+   * @return true if this property support updated dynamically
+   */
+  public boolean isDynamic() {
+    return mDynamic;
   }
 
   /**
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -124,7 +124,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     private ConsistencyCheckLevel mConsistencyCheckLevel = ConsistencyCheckLevel.IGNORE;
     private Scope mScope = Scope.ALL;
     private DisplayType mDisplayType = DisplayType.DEFAULT;
-    private boolean mDynamic = true;
+    private boolean mIsDynamic = true;
 
     /**
      * @param name name of the property
@@ -254,8 +254,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
      * @param dynamic whether the property could be updated dynamically
      * @return the updated builder instance
      */
-    public Builder setDynamic(boolean dynamic) {
-      mDynamic = dynamic;
+    public Builder setIsDynamic(boolean dynamic) {
+      mIsDynamic = dynamic;
       return this;
     }
 
@@ -287,7 +287,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
 
       PropertyKey key = new PropertyKey(mName, mDescription, defaultSupplier, mAlias,
           mIgnoredSiteProperty, mIsHidden, mConsistencyCheckLevel, mScope, mDisplayType,
-          mIsBuiltIn, mDynamic);
+          mIsBuiltIn, mIsDynamic);
       return key;
     }
 
@@ -4833,7 +4833,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               + "authentication is enabled. Server trusts whoever the client claims to be.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.ALL)
-          .setDynamic(false)
+          .setIsDynamic(false)
           .build();
   public static final PropertyKey SECURITY_AUTHORIZATION_PERMISSION_ENABLED =
       new Builder(Name.SECURITY_AUTHORIZATION_PERMISSION_ENABLED)
@@ -4841,7 +4841,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setDescription("Whether to enable access control based on file permission.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.ALL)
-          .setDynamic(false)
+          .setIsDynamic(false)
           .build();
   public static final PropertyKey SECURITY_AUTHORIZATION_PERMISSION_SUPERGROUP =
       new Builder(Name.SECURITY_AUTHORIZATION_PERMISSION_SUPERGROUP)
@@ -6800,7 +6800,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   }
 
   /**
-   * @return true if this property support updated dynamically
+   * @return true if this property can be updated dynamically during runtime
    */
   public boolean isDynamic() {
     return mDynamic;
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -4559,6 +4559,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
    */
   private FileSystemMasterAuditContext createAuditContext(String command, AlluxioURI srcPath,
       @Nullable AlluxioURI dstPath, @Nullable Inode srcInode) {
+    // Audit log may be enabled during runtime
     AsyncUserAccessAuditLogWriter auditLogWriter = null;
     if (ServerConfiguration.getBoolean(PropertyKey.MASTER_AUDIT_LOGGING_ENABLED)) {
       auditLogWriter = mAsyncAuditLogWriter;
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -4559,9 +4559,13 @@ public final class DefaultFileSystemMaster extends CoreMaster
    */
   private FileSystemMasterAuditContext createAuditContext(String command, AlluxioURI srcPath,
       @Nullable AlluxioURI dstPath, @Nullable Inode srcInode) {
+    AsyncUserAccessAuditLogWriter auditLogWriter = null;
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_AUDIT_LOGGING_ENABLED)) {
+      auditLogWriter = mAsyncAuditLogWriter;
+    }
     FileSystemMasterAuditContext auditContext =
-        new FileSystemMasterAuditContext(mAsyncAuditLogWriter);
-    if (mAsyncAuditLogWriter != null) {
+        new FileSystemMasterAuditContext(auditLogWriter);
+    if (auditLogWriter != null) {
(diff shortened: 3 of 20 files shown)


core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,16 @@ public void resetState() {
     mPathProperties.resetState();
   }
 
+  @Override
+  public void updateConfig(Map<String, String> propertiesMap) {
+    for (Map.Entry<String, String> entry : propertiesMap.entrySet()) {
+      PropertyKey propertyKey = PropertyKey.fromString(entry.getKey());
+      if (propertyKey.isDynamic()) {
+        ServerConfiguration.set(propertyKey, entry.getValue());
```
apc999, line 614: nit: i know this is equivalent, but let's be a bit more explicit here
```suggestion
        ServerConfiguration.set(propertyKey, entry.getValue(), Source.RUNTIME);
```

core/transport/src/main/proto/grpc/meta_master.proto
```diff
@@ -244,6 +246,12 @@ message MasterHeartbeatPResponse {
   optional MetaCommand command = 1;
 }
 
+message UpdateConfigPRequest {
+  map<string, string> properties = 1;
+}
+
+message UpdateConfigPResponse {}
```
apc999, line 253: shall we return the successfully updated keys instead of making this `void` ?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -6783,6 +6799,13 @@ public boolean isBuiltIn() {
     return mIsBuiltIn;
   }
 
+  /**
+   * @return true if this property support updated dynamically
```
apc999, line 6803:
```suggestion
   * @return true if this property can be updated dynamically during runtime
```

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,16 @@ public void resetState() {
     mPathProperties.resetState();
   }
 
+  @Override
+  public void updateConfig(Map<String, String> propertiesMap) {
```
apc999, line 610: as mentioned below, it might be helpful to return the set of successfully updated keys?

core/transport/src/main/proto/grpc/meta_master.proto
```diff
@@ -244,6 +246,12 @@ message MasterHeartbeatPResponse {
   optional MetaCommand command = 1;
 }
 
+message UpdateConfigPRequest {
+  map<string, string> properties = 1;
```
apc999, line 250:
```suggestion
  optional map<string, string> properties = 1;
```

good practise to keep every field optional for future reference

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfigCommand.java
```diff
@@ -0,0 +1,86 @@
...
+      System.out.println("Updated " + properties.size() + " configs");
+    } else {
+      System.out.println("No config to update");
+    }
+    return 0;
+  }
+
+  @Override
+  public String getUsage() {
+    return "updateConfig [--option <key=val>] ";
```
apc999, line 79: how about calling this command `updateConf` as we have `setConf` already?

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfigCommand.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public String getDescription() {
+    return "Update attributes for a ufs path.";
```
apc999, line 84: please update

apc999 (changes requested):
Thanks for the improvement! A bunch of comments are left

apc999:
also, can you add a integration test ?

core/transport/src/main/proto/grpc/meta_master.proto
maobaolong: Sorry, Map fields cannot have labels

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfigCommand.java
maobaolong: Done

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfigCommand.java
maobaolong: Done

core/common/src/main/java/alluxio/conf/PropertyKey.java
maobaolong: Done

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
maobaolong: Done

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -249,6 +250,15 @@ public Builder setDisplayType(DisplayType displayType) {
       return this;
     }
 
+    /**
+     * @param dynamic whether the property could be updated dynamically
+     * @return the updated builder instance
+     */
+    public Builder setDynamic(boolean dynamic) {
```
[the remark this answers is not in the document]
maobaolong: Done

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -124,6 +124,7 @@
     private ConsistencyCheckLevel mConsistencyCheckLevel = ConsistencyCheckLevel.IGNORE;
     private Scope mScope = Scope.ALL;
     private DisplayType mDisplayType = DisplayType.DEFAULT;
+    private boolean mDynamic = true;
```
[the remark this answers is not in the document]
maobaolong: Done

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -75,4 +75,11 @@ void setPathConfiguration(AlluxioURI path, Map<PropertyKey, String> properties)
    * @param path the path
    */
   void removePathConfiguration(AlluxioURI path) throws IOException;
+
+  /**
+   * Update properties.
+   *
+   * @param propertiesMap the properties map to be updated
+   */
+  void updateConfig(Map<String, String> propertiesMap) throws IOException;
```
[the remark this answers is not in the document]
maobaolong: Because the implementation will use `retryRPC` which could throw IOException.

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,16 @@ public void resetState() {
...
+    for (Map.Entry<String, String> entry : propertiesMap.entrySet()) {
+      PropertyKey propertyKey = PropertyKey.fromString(entry.getKey());
+      if (propertyKey.isDynamic()) {
```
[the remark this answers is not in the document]
maobaolong: Done

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -75,4 +76,13 @@ void setPathConfiguration(AlluxioURI path, Map<PropertyKey, String> properties)
...
   void removePathConfiguration(AlluxioURI path) throws IOException;
+
+  /**
+   * Update properties.
+   *
+   * @param propertiesMap the properties map to be updated
+   * @return the update properties status map
+   */
+  Map<String, UpdatePropertyPStatus> updateConfiguration(
+      Map<String, String> propertiesMap) throws IOException;
```
apc999, line 87: How about the following?

```suggestion
  Map<PropertyKey, Boolean> updateConfiguration(
      Map<PropertyKey, String> propertiesMap) throws IOException;
```

1. using `PropertyKey` to indicate key as other methods in this interface to keep consistent, and more type-safe
2. let's not expose Proto-generated type to upper layer, to avoid using `UpdatePropertyPStatus`. It is very hard to read and understand the type when it is a proto

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+    } else {
+      System.out.println("No config to update");
+    }
+    return 0;
+  }
+
+  @Override
+  public String getUsage() {
+    return "updateConf [--option <key=val>] ";
```
apc999, line 88: can we just have the following syntax for this CLI?

```
fsadmin updateConf key1=val1 key2=val2
```

Having `--option` is typically when passing something that is optional.

core/transport/src/main/proto/grpc/meta_master.proto
```diff
@@ -244,6 +246,19 @@ message MasterHeartbeatPResponse {
...
+message UpdateConfigurationPRequest {
+  map<string, string> properties = 1;
+}
+
+message UpdateConfigurationPResponse {
+  enum UpdatePropertyPStatus {
+    SUCCESS = 0;
+    BANNED = 1;
+    Exception = 2;
+  }
```
apc999, line 258: proto enum is hard to read and maintain.
how about just use boolean to indicate success or failure?

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+  @Test
+  public void unknownKey() {
+    int ret = mFsAdminShell.run("updateConf", "--option", "unknown-key=unknown-value");
+    Assert.assertEquals(-1, ret);
+  }
+
+  @Test
+  public void updateNormal() {
+    int ret = mFsAdminShell.run("updateConf", "--option", "alluxio.master.worker.timeout=4min");
+    Assert.assertEquals(0, ret);
```
apc999, line 45: run 
`mFsShell.run("getConf", "alluxio.master.worker.timeout");` and assert the returned value?

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Test
+  public void unknownKey() {
+    int ret = mFsAdminShell.run("updateConf", "--option", "unknown-key=unknown-value");
+    Assert.assertEquals(-1, ret);
+  }
+
+  @Test
+  public void updateNormal() {
```
apc999, line 43: add another test case with non-dynamic key?

apc999 (changes requested):
thanks for the quick interation! I get my last batch of comments

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
maobaolong: Sure

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: OK

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
maobaolong: Sure

core/transport/src/main/proto/grpc/meta_master.proto
maobaolong: Sure

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: OK

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
maobaolong: For 2, it make sense.

For 1, It could not easy to use `PropertyKey` here, because, if user update some bad key which is not exist in PropertyKey, so we cannot get a PropertyKey at all, but we should return the result and tell the user failed to update that key.

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
maobaolong: OK

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
maobaolong: `getConf ` is not a subCommand of `FsShell`, so how to run it through a shell?

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Map.Entry;
+
+/**
+ * Update config for an existing service.
+ */
+@ThreadSafe
+@PublicApi
+public final class UpdateConfCommand extends AbstractFsAdminCommand {
+
+  private static final Option OPTION_OPTION =
```
apc999, line 35: no more needed?

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
apc999: update the usage string?

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
apc999: In that case, we can return to the caller of this client earlier when failed to convert a input string to `PropertyKey`, without really calling RPC. does that work?

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -75,4 +75,13 @@ void setPathConfiguration(AlluxioURI path, Map<PropertyKey, String> properties)
...
    */
   void removePathConfiguration(AlluxioURI path) throws IOException;
+
+  /**
+   * Update properties.
+   *
+   * @param propertiesMap the properties map to be updated
+   * @return the update properties status map
+   */
+  Map<String, Boolean> updateConfiguration(
```
apc999, line 85: check my comments https://github.com/Alluxio/alluxio/pull/13514#discussion_r645950175 , will that work?
I want to prevent the confusion in the future when adding new methods in this class.

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -75,4 +75,13 @@ void setPathConfiguration(AlluxioURI path, Map<PropertyKey, String> properties)
    * @param path the path
    */
   void removePathConfiguration(AlluxioURI path) throws IOException;
+
+  /**
+   * Update properties.
```
apc999, line 80: nit:
```suggestion
   * Updates properties.
```

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Override
+  public int run(CommandLine cl) throws IOException {
+    int errorCode = 0;
+    Map<String, String> properties = new HashMap<>();
+    cl.getArgList().forEach(arg -> {
+      if (arg.contains("=")) {
+        String[] kv = arg.split("=");
+        if (arg.length() == 2) {
+          properties.put(kv[0], kv[1]);
```
apc999, line 72: Will this work?

`Map<PropertyKey, String> properties = new HashMap<>();`

```suggestion
          properties.put(PropertyKey.fromString(kv[0]), kv[1]);
```

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+      }
+    });
+
+    if (properties.size() > 0) {
```
apc999, line 77: This is minor, up to you.

For readability, we prefer to have error handling first and return on errors,
followed by the normal case, rather than leaving the error handling in an `else` branch.
In this case, it will be like

```java
if (properties.size() == 0) {
  System.out.println("No config to update");
  return -1;
}
Map<String, Boolean> result =
  ...
...
```

I know logically they are equivalent. It is more like a convention to ease other maintainers

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,112 @@
...
+    return "updateConf [--option <key=val>] ";
+  }
+
+  /**
+   * @return command's description
+   */
+  @VisibleForTesting
+  public static String description() {
+    return "Update config for alluxio master.";
+  }
```
apc999, line 106: why not a public static string constant? any reason to have this in a getter method?

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
apc999: you are right

tests/src/test/java/alluxio/client/cli/fsadmin/command/UpdateConfIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+  @Test
+  public void unknownOption() {
+    int ret = mFsAdminShell.run("updateConf", "--unknown");
+    Assert.assertEquals(-1, ret);
+    String output = mOutput.toString();
+    Assert.assertEquals(UpdateConfCommand.description(), lastLine(output));
+  }
+
+  @Test
+  public void unNormalKey() {
```
apc999, line 34: you mean update `updateUnknownKey`?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4823,13 +4833,15 @@ public String toString() {
               + "authentication is enabled. Server trusts whoever the client claims to be.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.ALL)
+          .setIsDynamic(false)
```
apc999, line 4836: there shall be some other configurations, like master RPC port, `MASTER_HOSTNAME`, `MASTER_METASTORE`, journal type (UFS, Raft), `HOME`, and many others to be non-dynamic. Now I actually feel it is safer to disable updating a property, unless we are sure it can be set dynamically. (e.g., these timeout). We can discuss it offline

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,33 @@ public void resetState() {
...
+          successCount++;
+          LOG.debug("Update property {} from {} to {}",
+              key.getName(), oldValue, ServerConfiguration.get(key));
+        } else {
+          LOG.debug("Update a non-dynamic property {} is banned", key.getName());
+          result.put(entry.getKey(), false);
+        }
+      } catch (Exception e) {
+        result.put(entry.getKey(), false);
+        LOG.error("Update property {} to {} met exception", e);
```
apc999, line 629: unbalanced arguments

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,33 @@ public void resetState() {
...
+    for (Map.Entry<String, String> entry : propertiesMap.entrySet()) {
+      try {
+        PropertyKey key = PropertyKey.fromString(entry.getKey());
+        if (key.isDynamic()) {
+          String oldValue = ServerConfiguration.get(key);
+          ServerConfiguration.set(key, entry.getValue(), Source.RUNTIME);
+          result.put(entry.getKey(), true);
+          successCount++;
+          LOG.debug("Update property {} from {} to {}",
+              key.getName(), oldValue, ServerConfiguration.get(key));
```
apc999, line 622: Think this twice, I actually think we shall `LOG.info` this change as it may help debug.
Without knowing what configuration is changed during runtime, it can be hard to reason.

```suggestion
          LOG.info("Property {} has been updated to \"{}\" from \"{}\"",
              key.getName(), entry.getValue(), oldValue);
```

core/server/master/src/main/java/alluxio/master/meta/MetaMasterConfigurationServiceHandler.java
```diff
@@ -127,4 +129,17 @@ public void removePathConfiguration(RemovePathConfigurationPRequest request,
...
       }, "removePathConfiguration", "request=%s", responseObserver, request);
   }
+
+  @Override
+  public void updateConfiguration(
+      UpdateConfigurationPRequest request,
+      StreamObserver<UpdateConfigurationPResponse> responseObserver) {
+    RpcUtils.call(LOG, () -> {
+      Map<String, Boolean> result
+          = mMetaMaster.updateConfiguration(request.getPropertiesMap());
```
apc999, line 139: nit: wrong linebreak

https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break

```suggestion
      Map<String, Boolean> result =
            mMetaMaster.updateConfiguration(request.getPropertiesMap());
```

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Override
+  public int run(CommandLine cl) throws IOException {
+    int errorCode = 0;
+    Map<String, String> properties = new HashMap<>();
+    cl.getArgList().forEach(arg -> {
+      if (arg.contains("=")) {
+        String[] kv = arg.split("=");
+        if (arg.length() == 2) {
```
apc999, line 71: when `arg.length() != 2`, output error message?

apc999 (changes requested):
some more questions. We are getting there. Thanks for this useful feature!

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
maobaolong: OK

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
maobaolong: I have a case, if an old version shell client `Alluxio-V0` and a new version master `Alluxio-V1`, the V0 to V1 contains a new PropertyKey added, so we cannot update the new added PropertyKey through the  `Alluxio-V0` shell client. That make things getting hard to maintain the client and master to same version all the time.

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: Done

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: Yeah, removed it.

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
maobaolong: I think the strict type-safe protocol can lead we have to keep use same version both client and master.

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
maobaolong: Done

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: This is what i want to discuss, more strict type-safe, more consistent work we have to do meanwhile.

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: Yeah, make sense.

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: To be frank, I just reference the exist code, something like this

image

core/common/src/main/java/alluxio/conf/PropertyKey.java
maobaolong: OK

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
maobaolong: Good catch, thank you to show me this error, I just want to display the key and value.

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
maobaolong: Ok

core/server/master/src/main/java/alluxio/master/meta/MetaMasterConfigurationServiceHandler.java
maobaolong: Yeah

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: OK

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
apc999: OK, we can change this together later.

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
apc999: I'm a bit confused here. This API is currently called by Alluxio CLI, not by Alluxio client (e.g., inside Presto or FUSE). 
Do you see the possibility that even the Alluxio shell is in a different version of Alluxio deployment in this case?
We can discuss this offline

apc999: You can use `PropertyKey.getOrBuildCustom` to create `PropertyKey`

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+
+    if (properties.size() == 0) {
+      System.out.println("No config to update");
+      return -1;
+    }
+
+    Map<String, Boolean> result
+        = mMetaConfigClient.updateConfiguration(properties);
```
apc999, line 67:
```suggestion
    Map<String, Boolean> result =
          mMetaConfigClient.updateConfiguration(properties);
```

apc999 (commented):
Last batch of comments

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
maobaolong: Done

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
maobaolong: Done

maobaolong:
@apc999 Thanks for the review and help, PTAL.

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
```diff
@@ -75,4 +75,13 @@ void setPathConfiguration(AlluxioURI path, Map<PropertyKey, String> properties)
...
   void removePathConfiguration(AlluxioURI path) throws IOException;
+
+  /**
+   * Updates properties.
+   *
+   * @param propertiesMap the properties map to be updated
+   * @return the update properties status map
+   */
+  Map<PropertyKey, Boolean> updateConfiguration(
+      Map<String, String> propertiesMap) throws IOException;
```
apc999, line 86: Can we change input argument to String to `PropertyKey` too?
```suggestion
      Map< PropertyKey, String> propertiesMap) throws IOException;
```

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -4559,9 +4559,13 @@ private Metrics() {} // prevent instantiation
    */
   private FileSystemMasterAuditContext createAuditContext(String command, AlluxioURI srcPath,
       @Nullable AlluxioURI dstPath, @Nullable Inode srcInode) {
+    AsyncUserAccessAuditLogWriter auditLogWriter = null;
+    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_AUDIT_LOGGING_ENABLED)) {
```
apc999, line 4563:
```suggestion
    // Audit log may be enabled during runtime
    AsyncUserAccessAuditLogWriter auditLogWriter = null;
    if (ServerConfiguration.getBoolean(PropertyKey.MASTER_AUDIT_LOGGING_ENABLED)) {
```

core/client/fs/src/main/java/alluxio/client/meta/MetaMasterConfigClient.java
apc999, line 86: can we have `Map<PropertyKey, String> propertiesMap` as input argument?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,34 @@ public void resetState() {
...
+        if (ServerConfiguration.getBoolean(PropertyKey.CONF_DYNAMIC_UPDATE_ENABLED)
+            && key.isDynamic()) {
+          String oldValue = ServerConfiguration.get(key);
+          ServerConfiguration.set(key, entry.getValue(), Source.RUNTIME);
+          result.put(entry.getKey(), true);
+          successCount++;
+          LOG.info("Property {} has been updated to \"{}\" from \"{}\"",
+              key.getName(), entry.getValue(), oldValue);
+        } else {
+          LOG.debug("Update a non-dynamic property {} is banned", key.getName());
```
apc999, line 625:
```suggestion
          LOG.debug("Update a non-dynamic property {} is not allowed", key.getName());
```

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -606,6 +606,34 @@ public void resetState() {
...
+          result.put(entry.getKey(), false);
+        }
+      } catch (Exception e) {
+        result.put(entry.getKey(), false);
+        LOG.error("Update property {} to {} met exception", entry.getKey(), entry.getValue(), e);
```
apc999, line 630:
```suggestion
        LOG.error("Failed to update property {} to {}", entry.getKey(), entry.getValue(), e);
```

core/server/master/src/main/java/alluxio/master/meta/MetaMasterConfigurationServiceHandler.java
```diff
@@ -127,4 +129,17 @@ public void removePathConfiguration(RemovePathConfigurationPRequest request,
...
+  public void updateConfiguration(
+      UpdateConfigurationPRequest request,
+      StreamObserver<UpdateConfigurationPResponse> responseObserver) {
+    RpcUtils.call(LOG, () -> {
+      Map<String, Boolean> result =
+          mMetaMaster.updateConfiguration(request.getPropertiesMap());
+      return UpdateConfigurationPResponse.newBuilder()
+          .putAllStatus(result)
+          .build();
+    }, "updateConfig", "request=%s", responseObserver, request);
```
apc999, line 143:
```suggestion
    }, "updateConfiguration", "request=%s", responseObserver, request);
```

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,97 @@
...
+        errorCode = -2;
+      }
+    }
+
+    return errorCode;
+  }
+
+  @Override
+  public String getUsage() {
+    return "updateConf key1=val1 key2=val2 ... keyn=valn";
```
apc999, line 82:
```suggestion
    return "updateConf key1=val1 [key2=val2 ...]";
```

shell/src/main/java/alluxio/cli/fsadmin/command/UpdateConfCommand.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  public int run(CommandLine cl) throws IOException {
+    int errorCode = 0;
+    Map<String, String> properties = new HashMap<>();
+    for (String arg : cl.getArgList()) {
+      if (arg.contains("=")) {
+        String[] kv = arg.split("=");
+        if (kv.length != 2) {
+          System.err.println("Each argument must exactly contain one `=`");
```
apc999, line 55:
```suggestion
          System.err.println("Failed to parse %s, expecting argument in the format of \"key=val\", arg);
```

apc999 (changes requested):
Mostly good now. Please help address this batch of comments and I will merge the PR afterwards.

Thanks for the improvement!

apc999 (approved):

apc999:

PR #13519, opened 2021-06-01
[DO NOT MERGE] BlockMaster benchmark 2.5

jiacheliu3:
This branch is only used for benchmarking BlockMaster from 2.5-RC2. 
The benchmarking is against 2.6.


jiacheliu3:
Closed because the benchmark is done

Issue #13512, opened 2021-05-31
The LeaderCommand can be removed?


kuszz:
**Describe the solution you'd like**
It seems the leader command can be removed, the masterInfo command also return the leader master address.

maobaolong:
@kuszz  So do I. I think so too.

apc999:
@kuszz @maobaolong , this may potentially break users' backward compatibility.
e.g., users may already have script built based on this comment. Then this will break their process.
How about adding an Java annotation saying the leader command will be removed after Alluxio 3.0 ?

Xenorith:
welp it broke backwards compatibility for my tests because i parse the output directly and now there's a warning string. i guess the warning worked as intended

PR #13524, opened 2021-06-02
Remove the LeaderCommand.java and LeaderCommandIntegrationTest.java f…

kuszz:
#13512


LuQQiu:
Same backward compatibility issue, instead of removing a shell command directly, can you deprecated this command and mention that this command will be removed in Alluxio 3.0. There should be some examples in the code base about how to deprecate a command that is no longer needed. and point users using this command to new commands.

kuszz:
> Same backward compatibility issue, instead of removing a shell command directly, can you deprecated this command and mention that this command will be removed in Alluxio 3.0. There should be some examples in the code base about how to deprecate a command that is no longer needed. and point users using this command to new commands.

Thanks, I will reference examples and modify the code

apc999:
same [comment](https://github.com/Alluxio/alluxio/issues/13512#issuecomment-854153335) as @LuQQiu

apc999 (changes requested):
Deprecate the command rather than removing it directly

apc999:
close this PR for now. You can update the code offline and re-open it once you are done

PR #13526, opened 2021-06-02
[DOCFIX] Improve remote logging doc

jiacheliu3:
2 changes:
1. Change the port from number to string, as in a K8s configmap the number format will lead to a parsing error, only the string format will work.
2. Add a link to remote logging setup in k8s as that's much different

### DIFF
--- a/docs/en/operation/Remote-Logging.md
+++ b/docs/en/operation/Remote-Logging.md
@@ -67,7 +67,7 @@ The following lines would need to be added to `conf/alluxio-env.sh` to enable re
 
 ```bash
 ALLUXIO_LOGSERVER_HOSTNAME=AlluxioLogServer
-ALLUXIO_LOGSERVER_PORT=45600
+ALLUXIO_LOGSERVER_PORT="45600"
 ```
 > Note: You MUST set BOTH variables.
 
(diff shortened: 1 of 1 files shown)


docs/en/operation/Remote-Logging.md
```diff
@@ -67,7 +67,7 @@ The following lines would need to be added to `conf/alluxio-env.sh` to enable re
 
 ```bash
 ALLUXIO_LOGSERVER_HOSTNAME=AlluxioLogServer
-ALLUXIO_LOGSERVER_PORT=45600
+ALLUXIO_LOGSERVER_PORT="45600"
```
HelloHorizon, line 70: will it cause the regression to other deployments?

apc999: why is this needed?

jiacheliu3: I thought I put an explanation on this PR this line but it seemed to be a github webpage issue that my comment didn't go through...

The reason is in K8s configmap, this value has to be string instead of number, otherwise a parse error will be thrown. I'm changing the  non-k8s instruction to use a string value too so both places are consistent.

You can find the k8s part instruction using a string value [here](https://docs.alluxio.io/os/user/stable/en/deploy/Running-Alluxio-On-Kubernetes.html#enable-remote-logging) in the `kubectl` section. I tested manually on physical machine too so no regression is expected.

jiacheliu3: @apc999 @HelloHorizon

HelloHorizon (approved):

HelloHorizon:

HelloHorizon:

PR #13540, opened 2021-06-03
Make worker start timeout configurable

jiacheliu3:
Currently the worker start timeout is 1min which is not configurable. The timeout is for worker registry startup which calls all `WorkerFactory` children classes to create corresponding workers. Currently there is only `BlockWorkerFactory` which creates a `BlockWorker`. 

The startup time of a `BlockWorker` involves initializing the tiered storage, which might take longer than 1min if the tiered storage is slow or there are too many blocks. In such cases 1 minute may be too short and cause flakiness or inability to start the worker. 

This change makes the timeout configurable.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3212,6 +3212,16 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_REGISTRY_START_TIMEOUT =
+      new Builder(Name.WORKER_REGISTRY_START_TIMEOUT)
+          .setDefaultValue("1min")
+          .setDescription("Timeout if the worker registry startup takes longer than the "
+              + "specified time. The worker registry contains all worker factory classes "
+              + "that implements alluxio.worker.workerFactory. During the startup, the "
+              + "worker registry will use the worker factories to create worker processes.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
   // The default is set to 11. One client is reserved for some light weight operations such as
   // heartbeat. The other 10 clients are used by commitBlock issued from the worker to the block
   // master.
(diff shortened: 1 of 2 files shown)


apc999:
@jiacheliu3 can you elaborate how important it is to have this configurable? One feedback from many users is that we are currently supporting too many tuning knobs to understand and manage this system.

In this case, do we want to solve a particular issue ?

calvinjia:
@apc999 In some cases 1 minute may be too short and cause flakiness or inability to start the worker. Some alternatives are to set the value fairly high (ie. 10 minutes, but we still may run into problems at larger scale) or make this a hidden configuration value (since this shouldn't need tuning except in very large scale deployments).

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3212,6 +3212,16 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_REGISTRY_START_TIMEOUT =
+      new Builder(Name.WORKER_REGISTRY_START_TIMEOUT)
+          .setDefaultValue("1min")
+          .setDescription("Timeout if the worker registry startup takes longer than the "
+              + "specified time. The worker registry contains all worker factory classes "
+              + "that implements alluxio.worker.workerFactory. During the startup, the "
+              + "worker registry will use the worker factories to create worker processes.")
```
jiacheliu3, line 3221: PTAL if this makes sense to users

jiacheliu3:
> @apc999 In some cases 1 minute may be too short and cause flakiness or inability to start the worker. Some alternatives are to set the value fairly high (ie. 10 minutes, but we still may run into problems at larger scale) or make this a hidden configuration value (since this shouldn't need tuning except in very large scale deployments).

IMO it's fine to leave this property as not hidden, as when the users face such timeout they will be able to find the property themselves.

jiacheliu3:
@HelloHorizon FYI

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3212,6 +3212,16 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_REGISTRY_START_TIMEOUT =
+      new Builder(Name.WORKER_REGISTRY_START_TIMEOUT)
+          .setDefaultValue("1min")
```
HelloHorizon, line 3217: @calvinjia Do we want to change the value to 10min instead? Or since this value is only for large-scale tuning, we can leave it as 1min?

calvinjia:
@jiacheliu3 @HelloHorizon 
I'm okay with leaving default at 1 minute and allowing the user to configure if they run into issues (1 minute will be sufficient for majority of deployments). The downside is introducing yet another public facing configuration value. @apc999 do you have a strong opinion here?
Regarding the change in calculation (callables * timeout vs just timeout) I think simplifying to a single timeout makes sense since we want to expose it as configuration and it's unreasonable to expect users to understand the internals of how many worker processes there are.

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: I prefer to leave this as 1min to keep the default behavior unchanged. Users will be able to tune it only when necessary.

HelloHorizon:
bump up the value instead of changing it to a configurable value. #13646

PR #13543, opened 2021-06-03
Run sync activities under managed blocking

ggezer:
Inode sync activities are blocking and should trigger ForkJoinPool compensation.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -15,6 +15,8 @@ import alluxio.AlluxioURI;
 import alluxio.Constants;
 import alluxio.client.WriteType;
 import alluxio.collections.Pair;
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -294,9 +297,71 @@ public InodeSyncStream(LockingScheme rootScheme, DefaultFileSystemMaster fsMaste
...
+    } else {
+      throw new IllegalStateException("Unexpected error during sync.", th);
+    }
+  }
+
+  /**
+   * @return {@code false} if sync is not needed
+   */
+  private boolean isSyncNeeded() {
+    return mRootScheme.shouldSync() || mForceSync;
```
ggezer, line 356: Will this condition cover most of the cases where we don't actually sync?

ggezer:
I found this change to be detrimental to meta operation throughput across the board.

PR #13581, opened 2021-06-07
Support bypassing by partitions

dbw9580:
### What changes were proposed in this pull request?

Following PR #13455, this PR allows bypassing by partition names.

To allow partition specification, the syntax of the config file `catalog.db.config.file` has been changed. It now looks like this:

```json
{
  "bypass": {
    "tables": [
      "table1",
      {"table": "table2", "partitions": ["table2_part1", "table2_part2"]}
    ]
  }
}
```

More fine-grained control over which partitions to bypass.

### Does this PR introduce _any_ user-facing change?

Yes. The syntax of the config file is changed.

### DIFF
--- a/docs/en/core-services/Catalog.md
+++ b/docs/en/core-services/Catalog.md
@@ -102,17 +102,28 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
     thrift://metastore_host:9083 default
 ```
 
-If you want specify config file for the UDB, please add `attachdb` option `-o catalog.db.config.file`.
-Each time you want to affect the config file, you can use `alluxio table sync`.
+If you want to specify a config file for the UDB, please append an option `-o catalog.db.config.file` to
+`attachdb` command.
+After each time you made changes to the config file, you can use `alluxio table sync` to apply the changes.
 
-If you want to write a configuration file for the UDB, please configure it like the following given example.
-It must contain an array `bypassSet`, the comma-separated table names will be bypassed.
+The configuration file is in JSON format, and can contain these configurations:
 
-```json
-{
-  "bypassSet" : ["table1", "table2"]
-}
-```
+1. Tables and partitions bypassing specification:
+
+    ```json
+    {
+      "bypass": {
+        "tables": [
+          "table1",
+          {"table": "table2", "partitions": ["table2_part1", "table2_part2"]}
+        ]
+      }
+    }
+    ```
+
+    You can specify which tables and partitions within these tables should be bypassed from Alluxio.
+    By specifying only the table name, all partitions of that table, if any, will be bypassed. Otherwise, 
+    you can specify specific partitions to bypass.
 
 > **Note:** When databases are attached, all tables are synced from the configured UDB.
 If out-of-band updates occur to the database or table and the user wants query results to reflect
--- a/docs/en/core-services/Catalog.md
+++ b/docs/en/core-services/Catalog.md
@@ -102,13 +102,20 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
     thrift://metastore_host:9083 default
 ```
 
-If you want to specify a config file for the UDB, please append an option `-o catalog.db.config.file` to
+#### UDB Configuration File
+
+To specify a configuration file for the UDB, append an option `-o catalog.db.config.file` to
 `attachdb` command.
-After each time you made changes to the config file, you can use `alluxio table sync` to apply the changes.
+Each time the configuration file is changed, you can use `alluxio table sync` to apply the changes.
 
 The configuration file is in JSON format, and can contain these configurations:
 
 1. Tables and partitions bypassing specification:
+
+    You can specify some tables and partitions to be bypassed from Alluxio, so that they will not be
+    cached in Alluxio, instead clients will be directed to access them directly from the UDB. 
+    This can be helpful when some tables and partitions are large, and accommodating them in the cache
+    is undesirable.
 
     ```json
     {
--- a/docs/en/core-services/Catalog.md
+++ b/docs/en/core-services/Catalog.md
@@ -115,7 +115,7 @@ The configuration file is in JSON format, and can contain these configurations:
     You can specify some tables and partitions to be bypassed from Alluxio, so that they will not be
     cached in Alluxio, instead clients will be directed to access them directly from the UDB. 
     This can be helpful when some tables and partitions are large, and accommodating them in the cache
-    is undesirable.
+    is undesirable. An example configuration is like the following:
 
     ```json
     {
@@ -129,8 +129,11 @@ The configuration file is in JSON format, and can contain these configurations:
     ```
 
     You can specify which tables and partitions within these tables should be bypassed from Alluxio.
-    By specifying only the table name, all partitions of that table, if any, will be bypassed. Otherwise, 
-    you can specify specific partitions to bypass.
+    By specifying only the table name, all partitions of that table, if any, will be bypassed. 
+    Otherwise, you can specify specific partitions to bypass.
+    
+    In the example above, table 1 is fully bypassed. Partition 1 and 2 of table 2 are bypassed, 
+    and any other partitions, if any, are not.
 
 > **Note:** When databases are attached, all tables are synced from the configured UDB.
 If out-of-band updates occur to the database or table and the user wants query results to reflect
--- a/docs/en/operation/User-CLI.md
+++ b/docs/en/operation/User-CLI.md
@@ -1436,7 +1436,7 @@ Here are the additional properties possible for the `-o` options:
   * `catalog.db.config.file`: the config file for the UDB, 
     you can configure which tables and partitions to bypass from Alluxio in a configuration specified 
     by this option. 
-    See [UDB Configuration File]({{ '/en/core-services/Catalog.md#udb-configuration-file' | relativize_url }}) 
+    See [UDB Configuration File]({{ '/en/core-services/Catalog.html#udb-configuration-file' | relativize_url }}) 
     for details.
   * `catalog.db.ignore.udb.tables`: comma-separated list of table names to ignore from the UDB
   * `catalog.db.sync.threads`: number of parallel threads to use to sync with the UDB. If too large,
--- a/docs/en/operation/User-CLI.md
+++ b/docs/en/operation/User-CLI.md
@@ -1433,7 +1433,8 @@ Here are the additional properties possible for the `-o` options:
     * `<UDB_TYPE>`: the UDB type
     * `<UFS_PREFIX>`: the UFS path prefix, or a regex string starts with `regex:` that the mount properties are for
     * `<MOUNT_PROPERTY>`: an Alluxio mount property
-  * `catalog.db.config.file`: the config file for the UDB, you can specify bypass table by this option.
+  * `catalog.db.config.file`: the config file for the UDB, 
+    you can configure which tables and partitions should be bypassed from Alluxio by this option.
   * `catalog.db.ignore.udb.tables`: comma-separated list of table names to ignore from the UDB
   * `catalog.db.sync.threads`: number of parallel threads to use to sync with the UDB. If too large,
   the sync may overload the UDB, and if set too low, syncing a database with many tables make take
--- a/docs/en/operation/User-CLI.md
+++ b/docs/en/operation/User-CLI.md
@@ -1434,7 +1434,10 @@ Here are the additional properties possible for the `-o` options:
     * `<UFS_PREFIX>`: the UFS path prefix, or a regex string starts with `regex:` that the mount properties are for
     * `<MOUNT_PROPERTY>`: an Alluxio mount property
   * `catalog.db.config.file`: the config file for the UDB, 
-    you can configure which tables and partitions should be bypassed from Alluxio by this option.
+    you can configure which tables and partitions to bypass from Alluxio in a configuration specified 
+    by this option. 
+    See [UDB Configuration File]({{ /en/core-services/Catalog.md#udb-configuration-file | relativize_url }}) 
+    for details.
   * `catalog.db.ignore.udb.tables`: comma-separated list of table names to ignore from the UDB
   * `catalog.db.sync.threads`: number of parallel threads to use to sync with the UDB. If too large,
   the sync may overload the UDB, and if set too low, syncing a database with many tables make take
--- a/docs/en/operation/User-CLI.md
+++ b/docs/en/operation/User-CLI.md
@@ -1436,7 +1436,7 @@ Here are the additional properties possible for the `-o` options:
   * `catalog.db.config.file`: the config file for the UDB, 
     you can configure which tables and partitions to bypass from Alluxio in a configuration specified 
     by this option. 
-    See [UDB Configuration File]({{ /en/core-services/Catalog.md#udb-configuration-file | relativize_url }}) 
+    See [UDB Configuration File]({{ '/en/core-services/Catalog.md#udb-configuration-file' | relativize_url }}) 
     for details.
   * `catalog.db.ignore.udb.tables`: comma-separated list of table names to ignore from the UDB
   * `catalog.db.sync.threads`: number of parallel threads to use to sync with the UDB. If too large,
--- a/table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
@@ -18,6 +18,11 @@ import java.util.Set;
  * Tables and partitions bypassing specification.
  */
 public final class UdbBypassSpec {
+  /**
+   * Map of table name to set of partition names.
+   * Keyed by a table's name, the value set contains names of partitions in that table.
+   * An empty set indicates all partitions of that table, if any, should be bypassed.
+   */
   private final Map<String, Set<String>> mTablePartMap;
 
   /**
@@ -32,9 +37,9 @@ public final class UdbBypassSpec {
    *
    * @param tableName the table name
    * @return true if the table is configured to be bypassed, false otherwise
-   * @see UdbBypassSpec#isFullyBypassedTable(String)
+   * @see UdbBypassSpec#hasFullTable(String)
    */
-  public boolean isBypassedTable(String tableName) {
+  public boolean hasTable(String tableName) {
     return mTablePartMap.containsKey(tableName);
   }
 
@@ -43,11 +48,11 @@ public final class UdbBypassSpec {
    *
    * @param tableName the table name
    * @return true if the table is configured to be fully bypassed, false otherwise
-   * @see UdbBypassSpec#isBypassedTable(String)
+   * @see UdbBypassSpec#hasTable(String)
    */
-  public boolean isFullyBypassedTable(String tableName) {
+  public boolean hasFullTable(String tableName) {
     // empty set indicates all partitions should be bypassed
-    return isBypassedTable(tableName) && mTablePartMap.get(tableName).size() == 0;
+    return hasTable(tableName) && mTablePartMap.get(tableName).size() == 0;
   }
 
   /**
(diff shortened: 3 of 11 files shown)


yuzhu:
@dbw9580 thanks for the improvement. Could you add some unit tests to show that spec parsing and error handling works as intended?

dbw9580:
I already did in `DbConfigTest`. Do I need to add more?

docs/en/core-services/Catalog.md
```diff
@@ -102,17 +102,28 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
+The configuration file is in JSON format, and can contain these configurations:
+
+1. Tables and partitions bypassing specification:
+
+    ```json
+    {
+      "bypass": {
+        "tables": [
+          "table1",
+          {"table": "table2", "partitions": ["table2_part1", "table2_part2"]}
```
HelloHorizon, line 118: (nit) Different from the number of tables in the database, there can be tens of thousands of partitions in single tables. Can we also come up with an easy way for user to configure the bypass partitions, like bypass major part of partitions? We can do it in a follow-up pr.

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+   */
+  public boolean isBypassedTable(String tableName) {
+    return mTablePartMap.containsKey(tableName);
+  }
+
+  /**
+   * Checks if all partitions of a table should be bypassed.
+   *
+   * @param tableName the table name
+   * @return true if the table is configured to be bypassed, false otherwise
```
HelloHorizon, line 45:
```suggestion
   * @return true if the table is configured to be fully bypassed, false otherwise
```

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+   * @param tableName the table name
+   * @param partitionName the partition name
+   * @return true if the partition should be bypassed, false otherwise
+   */
+  public boolean isBypassedPartition(String tableName, String partitionName) {
+    if (!isBypassedTable(tableName)) {
+      return false;
+    } else {
+      Set<String> parts = mTablePartMap.get(tableName);
+      if (parts.size() == 0) {
```
HelloHorizon, line 65: (nit) use `isFullyBypassedTable` since we already have the helper function.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -11,25 +11,198 @@
...
+   * partitions will be bypassed;
+   * or an object of form
+   * {"table": "tableName", "partitions": ["part1", "part2"]}
+   * can be used, and individual partitions can be specified.
    */
-  public Set<String> getBypassSet() {
-    return mBypassSet;
+  public static class BypassTableEntryDeserializer extends JsonDeserializer<BypassTableEntry> {
+    @Override
+    public BypassTableEntry deserialize(JsonParser jp, DeserializationContext cxt)
```
HelloHorizon, line 181: add a unit test to cover this function.

HelloHorizon (changes requested):
@dbw9580 thanks for the improvements! left some comments.

docs/en/core-services/Catalog.md
dbw9580: @HelloHorizon 
> come up with an easy way for user to configure the bypass partitions, like bypass major part of partitions

Yes, that's the plan.

@maobaolong Baolong, could you please shed some light on how we should handle partition specifications to make it practical for real use cases?

maobaolong: With the bypass table and partition feature, we can use it to bypass some older tables or partitions and mount a small range of tables or partitions to alluxio.

It just like a rolling forward window which stand for a range of active valuable dataset(partitions), the window can move forward by a time unit, let's say it is a week, so we just want alluxio cache the current week partition, and bypass the older partitions, if some query need the older partition, it will read data from under filesystem directly rather than read it from alluxio.

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+   * Checks by a partition's name if it should be bypassed.
+   *
+   * @param tableName the table name
+   * @param partitionName the partition name
+   * @return true if the partition should be bypassed, false otherwise
+   */
+  public boolean isBypassedPartition(String tableName, String partitionName) {
+    if (!isBypassedTable(tableName)) {
+      return false;
+    } else {
```
apc999, line 63: nit: you don't need `else` branch. In general, the following pattern is preferred for readability

```java
if (some Error case) {
  return;
}
// proceed for normal case;
```

table/server/master/src/test/java/alluxio/master/table/DbConfigTest.java
```diff
@@ -0,0 +1,88 @@
...
+  public void empty() throws Exception {
+    List<String> src = ImmutableList.of(
+        "{}",
+        "{\"bypass\": null}",
+        "{\"bypass\": {}}",
+        "{\"bypass\": {\"tables\": []}}"
+    );
+    for (String input : src) {
+      DbConfig config = mMapper.readValue(input, DbConfig.class);
+      assertEquals(config.getBypassEntry().getBypassedTables().size(), 0);
```
apc999, line 47: expected result goes first in test assert.

```suggestion
      assertEquals(0, config.getBypassEntry().getBypassedTables().size());
```

docs/en/core-services/Catalog.md
```diff
@@ -102,17 +102,28 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
-If you want to write a configuration file for the UDB, please configure it like the following given example.
-It must contain an array `bypassSet`, the comma-separated table names will be bypassed.
-
-```json
-{
-  "bypassSet" : ["table1", "table2"]
-}
-```
+If you want to specify a config file for the UDB, please append an option `-o catalog.db.config.file` to
+`attachdb` command.
```
apc999, line 106: nits in language tuning:

```suggestion
To specify a configuration file for the UDB, append an option `-o catalog.db.config.file` to
`attachdb` command.
```

docs/en/core-services/Catalog.md
```diff
@@ -102,17 +102,28 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
+After each time you made changes to the config file, you can use `alluxio table sync` to apply the changes.
```
apc999, line 107:
```suggestion
Each time the configuration file is changed, you can use `alluxio table sync` to apply the changes.
```

docs/en/core-services/Catalog.md
```diff
@@ -102,17 +102,28 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
+
+The configuration file is in JSON format, and can contain these configurations:
+
+1. Tables and partitions bypassing specification:
```
apc999, line 111: please explain here what do we mean by "Tables and partitions bypassing "

docs/en/operation/User-CLI.md
```diff
@@ -1433,7 +1433,8 @@ Here are the additional properties possible for the `-o` options:
     * `<UDB_TYPE>`: the UDB type
     * `<UFS_PREFIX>`: the UFS path prefix that the mount properties are for
     * `<MOUNT_PROPERTY>`: an Alluxio mount property
-  * `catalog.db.config.file`: the config file for the UDB, you can specify bypass table by this option.
+  * `catalog.db.config.file`: the config file for the UDB, 
+    you can configure which tables and partitions should be bypassed from Alluxio by this option.
```
apc999, line 1437:
```suggestion
    you can configure which tables and partitions to bypass from Alluxio in a configuration specified by this option.
```

apc999 (changes requested):
put some initial comments

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
dbw9580: `BypassTableEntry` is registered to use this class to deserialize. In `DbConfigTest` I added server deserailization tests to test this function.

docs/en/core-services/Catalog.md
dbw9580: Added in 9ad0139.

docs/en/core-services/Catalog.md
dbw9580: @maobaolong thanks for you input. I will come up with an implementation outline and discuss with you later. Resolve this for now.

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -211,8 +212,12 @@ public SyncStatus sync(JournalContext context) throws IOException {
 
     if (Files.exists(Paths.get(mConfigPath))) {
       ObjectMapper mapper = new ObjectMapper();
-      DbConfig dbConfig = mapper.readValue(new File(mConfigPath), DbConfig.class);
-      mBypassSet.addAll(dbConfig.getBypassSet());
+      try {
+        mDbConfig = mapper.readValue(new File(mConfigPath), DbConfig.class);
+      } catch (JsonProcessingException e) {
+        LOG.error(String.format("failed to deserialize UDB config file %s, stays unsynced",
+            mConfigPath), e);
```
apc999, line 219:
```suggestion
        LOG.error("Failed to deserialize UDB config file {}, stays unsynced", mConfigPath, e);
```

Check logging convention in Alluxio,
https://docs.alluxio.io/os/user/edge/en/contributor/Code-Conventions.html#best-practice

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+package alluxio.table.common.udb;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Tables and partitions bypassing specification.
+ */
+public final class UdbBypassSpec {
+  private final Map<String, Set<String>> mTablePartMap;
```
apc999, line 21: nit: javadoc what is stored in this map to ease future maintainance.

docs/en/core-services/Catalog.md
```diff
@@ -102,17 +102,35 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
+          "table1",
+          {"table": "table2", "partitions": ["table2_part1", "table2_part2"]}
+        ]
+      }
+    }
+    ```
+
+    You can specify which tables and partitions within these tables should be bypassed from Alluxio.
+    By specifying only the table name, all partitions of that table, if any, will be bypassed. Otherwise, 
+    you can specify specific partitions to bypass.
```
apc999, line 133: can we explainthe above json file? e.g., are we by-passing entire table of `"table1"` and two partitions in `table2`? Am I correct? better to relate to our examples.

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Checks if a table should be bypassed.
+   *
+   * @param tableName the table name
+   * @return true if the table is configured to be bypassed, false otherwise
+   * @see UdbBypassSpec#isFullyBypassedTable(String)
+   */
+  public boolean isBypassedTable(String tableName) {
```
apc999, line 37: nit: since this class is already called `UdbBypassSpec`, you can simply name this method `hasTable`

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Checks if all partitions of a table should be bypassed.
+   *
+   * @param tableName the table name
+   * @return true if the table is configured to be fully bypassed, false otherwise
+   * @see UdbBypassSpec#isBypassedTable(String)
+   */
+  public boolean isFullyBypassedTable(String tableName) {
```
apc999, line 48: maybe calling this method `hasFullTable`

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Checks by a partition's name if it should be bypassed.
+   *
+   * @param tableName the table name
+   * @param partitionName the partition name
+   * @return true if the partition should be bypassed, false otherwise
+   */
+  public boolean isBypassedPartition(String tableName, String partitionName) {
```
apc999, line 60: `hasPartition`

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+
+package alluxio.table.common.udb;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Tables and partitions bypassing specification.
+ */
+public final class UdbBypassSpec {
```
apc999, line 20: please add unit test of this class for its public methods

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @param tableName the table name
+   * @param partitionName the partition name
+   * @return true if the partition should be bypassed, false otherwise
+   */
+  public boolean isBypassedPartition(String tableName, String partitionName) {
+    if (!isBypassedTable(tableName)) {
+      return false;
+    }
+
```
apc999, line 64: nit: unnecessary empty line. We prefer src code to be compact.

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+    Set<String> parts = mTablePartMap.get(tableName);
+    if (parts.size() == 0) {
+      // empty set indicates all partitions should be bypassed
+      return true;
+    }
+
```
apc999, line 70: same here

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
```diff
@@ -0,0 +1,73 @@
...
+  public boolean isBypassedPartition(String tableName, String partitionName) {
+    if (!isBypassedTable(tableName)) {
+      return false;
+    }
+
+    Set<String> parts = mTablePartMap.get(tableName);
+    if (parts.size() == 0) {
+      // empty set indicates all partitions should be bypassed
+      return true;
+    }
```
apc999, line 69: this case can be covered by `return parts.contains(partitionName);`, right?
if that's the case, let's go with simpler code

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -11,25 +11,205 @@
...
+          throw new JsonParseException(mapper.treeAsTokens(node), "missing table name");
+        }
+        tableName = node.get("table").asText();
+        JsonNode partitionsList = node.get("partitions");
+        partitions = mapper.convertValue(partitionsList,  new TypeReference<Set<String>>() {});
+        if (partitions == null) {
+          partitions = new HashSet<>();
+        }
+      } else {
+        throw new JsonParseException(mapper.treeAsTokens(node), "invalid syntax");
```
apc999, line 210: nit: for code readability, try to do error-handling first before processing different cases.

apc999 (changes requested):
Some more comments. 
@maobaolong can you help review this PR and make sure this PR addresses your request?

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
dbw9580: I think calling `.contains(partitionName)` on an empty set returns `false` with any `partitionName`, but here we want it to return `true` because an empty set means any partition *should* be bypassed.

docs/en/core-services/Catalog.md
dbw9580: Yes, you are correct. Clarified in 6664afc4.

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: Thanks. I also changed the log level to warn.

table/server/common/src/main/java/alluxio/table/common/udb/UdbBypassSpec.java
dbw9580: Added in 6664afc4.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -11,25 +11,205 @@
...
+      } else {
+        // a {"table": "table", "partitions": ["part1", "part2"]} object
+        if (!node.hasNonNull("table")) {
+          throw new JsonParseException(mapper.treeAsTokens(node), "missing table name");
+        }
+        tableName = node.get("table").asText();
+        JsonNode partitionsList = node.get("partitions");
+        partitions = mapper.convertValue(partitionsList,  new TypeReference<Set<String>>() {});
+        if (partitions == null) {
+          partitions = new HashSet<>();
```
maobaolong, line 209:
```suggestion
          partitions = Collections.emptySet()
```

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -211,8 +212,12 @@ public SyncStatus sync(JournalContext context) throws IOException {
 
     if (Files.exists(Paths.get(mConfigPath))) {
       ObjectMapper mapper = new ObjectMapper();
-      DbConfig dbConfig = mapper.readValue(new File(mConfigPath), DbConfig.class);
-      mBypassSet.addAll(dbConfig.getBypassSet());
+      try {
+        mDbConfig = mapper.readValue(new File(mConfigPath), DbConfig.class);
+      } catch (JsonProcessingException e) {
+        LOG.warn(String.format("Failed to deserialize UDB config file %s, stays unsynced: {}",
```
maobaolong, line 218: Here swallow the exception, and keep PC continue, is that your expect logic? Maybe we need to throw the exception and exit the sync method.

BTW, if `(Files.exists(Paths.get(mConfigPath)))` return false, we should throw exception too.

For the LOG, it will help us to find the root cause if print the call stack.

A suggestion way is 

```
LOG.warn("Failed to deserialize UDB config file {}, stays unsynced.", mConfigPath, e);
```

maobaolong (changes requested):
Thanks for the improvement of SDS, left some minor comments.

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: > swallow the exception, and keep PC continue

Yes, I thought an invalid config file would warrant a warning, but should not block the sync process. But on second thought, maybe it's better that we make it fail and let the user notice the syntax error ASAP.

> if (Files.exists(Paths.get(mConfigPath))) return false, we should throw exception too

I'm not sure about that. `DB_CONFIG_FILE` in `CatalogProperty` doesn't have a default value, so when the user did not specify any config file when attaching the DB (i.e. they don't want any bypassing behaviour), they would be prevented to do sync because of the exception. I suggest we provide an empty config file as the default value for the config entry.

Also, can the user apply a config file after they have attached the DB? I.e. they did not use a config file initially, but want to add one afterwards? If so, how do we handle this situation?

dbw9580: > I suggest we provide an empty config file as the default value for the config entry.

On second thought, `DB_CONFIG_FILE` is the path of the config file, while a default empty config file exists only in code and does *not* have a path. So I would suggest we stay the way it is now (i.e. use `mDbConfig = DbConfig.empty();` in constructor, and ignore the missing config file).

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
dbw9580: Fixed in b5c33ad.

dbw9580:
@maobaolong thanks for the review. Made some changes, please take another look.

maobaolong:
@apc999 I think this PR looks good, Please have a double check.

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: Resolved in 1244eb7.

apc999 (approved):

apc999:
Thanks for the improvement @dbw9580 !

apc999:

PR #13614, opened 2021-06-09
Remove precondition string concat cost

jiacheliu3:

The `Precondition` check will generate a string per block and induce extra method invocation. The cost is significant both in execution time and heap allocation.

The change saves heap allocation for registerWorker call by 75%, compared to without this change. However the Precondition check was brought in by https://github.com/Alluxio/alluxio/pull/13461. That means this PR does not improve the performance of `registerWorker` compared to previous Alluxio version, instead it brings the performance back to normal.

NA

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
@@ -1067,7 +1067,7 @@ public class DefaultBlockMaster extends CoreMaster implements BlockMaster {
             BlockLocation location = entry.getKey();
             Preconditions.checkState(location.getWorkerId() == workerInfo.getId(),
                 "BlockLocation has a different workerId %s from the request sender's workerId %s!",
-                        location.getWorkerId(), workerInfo.getId());
+                    location.getWorkerId(), workerInfo.getId());
             mBlockStore.addLocation(blockId, location);
             mLostBlocks.remove(blockId);
           } else {
--- a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
@@ -1066,8 +1066,8 @@ public class DefaultBlockMaster extends CoreMaster implements BlockMaster {
             workerInfo.addBlock(blockId);
             BlockLocation location = entry.getKey();
             Preconditions.checkState(location.getWorkerId() == workerInfo.getId(),
-                "BlockLocation has a different workerId %s from the request sender's workerId %s!",
-                    location.getWorkerId(), workerInfo.getId());
+                "BlockLocation has a different workerId %s from the request sender's workerId %s",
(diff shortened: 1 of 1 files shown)


jiacheliu3:
@HelloHorizon FYI

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1064,12 +1064,7 @@ private void processWorkerAddedBlocks(MasterWorkerInfo workerInfo,
           Optional<BlockMeta> block = mBlockStore.getBlock(blockId);
           if (block.isPresent()) {
             workerInfo.addBlock(blockId);
-            BlockLocation location = entry.getKey();
-            Preconditions.checkState(location.getWorkerId() == workerInfo.getId(),
-                String.format("BlockLocation has a different workerId %s from "
-                    + "the request sender's workerId %s!",
-                        location.getWorkerId(), workerInfo.getId()));
-            mBlockStore.addLocation(blockId, location);
+            mBlockStore.addLocation(blockId, entry.getKey());
```
ggezer, line 1067: Did you try an `if` statement instead of `Preconditions`? Don't we still need this check?

apc999: @ggezer @jiacheliu3 Preconditions supports passing string format arguments

```

this can be replaced by

```java
Preconditions.checkState(location.getWorkerId() == workerInfo.getId(),
    "BlockLocation has a different workerId %s from the request sender's workerId %s!",
    location.getWorkerId(), workerInfo.getId());
```

jiacheliu3: Passing string format arguments is proven to save the performance the same way. Although `Preconditions.checkState` is effectively `if(!condition) throw new IllegalStateException`, I think it makes sense to use `Preconditions` instead of manually managing if-else.

Updated according to Bin's advice, thanks!

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1066,9 +1066,8 @@ private void processWorkerAddedBlocks(MasterWorkerInfo workerInfo,
             workerInfo.addBlock(blockId);
             BlockLocation location = entry.getKey();
             Preconditions.checkState(location.getWorkerId() == workerInfo.getId(),
-                String.format("BlockLocation has a different workerId %s from "
-                    + "the request sender's workerId %s!",
-                        location.getWorkerId(), workerInfo.getId()));
+                "BlockLocation has a different workerId %s from the request sender's workerId %s!",
+                    location.getWorkerId(), workerInfo.getId());
```
apc999, line 1070: is the indentation correct?

ggezer (approved):

HelloHorizon:

HelloHorizon:

PR #13616, opened 2021-06-09
[DOCFIX] Add javadoc links to Filesystem API doc

ZhuTopher:

This PR adjusts/adds javadoc links, adjusts some wording, and changed the ordering of sections for clarity in the Filesystem API doc.

Clarification on the Filesystem API doc.

Adjusted the public doc for the Filesystem API.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
@@ -27,11 +27,16 @@ import java.util.stream.Collectors;
 
 import javax.annotation.concurrent.ThreadSafe;
 /**
- * A policy that returns local host first, and if the local worker doesn't have enough availability,
- * it randomly picks a worker from the active workers list for each block write.
- * If no worker meets the demands, return local host.
- * USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES is used to reserve some space of the worker
- * to store the block, for the values mCapacityBytes minus mUsedBytes is not the available bytes.
+ * A policy that returns the local worker first, and if the local worker doesn't
+ * exist or have enough availability, will select the nearest worker from the active
+ * workers list with sufficient availability.
+ *
+ * The calculation of which worker gets selected is done for each block write.
+ *
+ * The PropertyKey `USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES`
+ * (alluxio.user.block.avoid.eviction.policy.reserved.size.bytes)
+ * is used as buffer space on each worker when calculating available space
+ * to store each block.
  */
 @ThreadSafe
 public final class LocalFirstAvoidEvictionPolicy implements BlockLocationPolicy {
@@ -78,9 +83,12 @@ public final class LocalFirstAvoidEvictionPolicy implements BlockLocationPolicy 
   }
 
   /**
-   * The information of BlockWorkerInfo is update after a file complete write. To avoid evict,
-   * user should configure "alluxio.user.block.avoid.eviction.policy.reserved.size.bytes"
-   * to reserve some space to store the block.
+   * Calculate the available bytes for a worker with the added buffer of
+   * USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES
+   * (alluxio.user.block.avoid.eviction.policy.reserved.size.bytes)
+   *
+   * Since the information of BlockWorkerInfo is updated <em>after</em> a file
+   * completes a write, mCapacityBytes minus mUsedBytes may not be the true available bytes.
    *
    * @param workerInfo BlockWorkerInfo of the worker
    * @return the available bytes of the worker
--- a/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
@@ -31,9 +31,12 @@ import javax.annotation.concurrent.ThreadSafe;
  * exist or have enough availability, will select the nearest worker from the active
  * workers list with sufficient availability.
  *
+ * The definition of 'nearest worker' is based on {@link alluxio.wire.TieredIdentity}.
+ * @see alluxio.wire.TieredIdentityUtils#nearest()
+ *
  * The calculation of which worker gets selected is done for each block write.
  *
- * The PropertyKey `USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES`
+ * The {@link alluxio.conf.PropertyKey.USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES}
  * (alluxio.user.block.avoid.eviction.policy.reserved.size.bytes)
  * is used as buffer space on each worker when calculating available space
  * to store each block.
@@ -84,7 +87,7 @@ public final class LocalFirstAvoidEvictionPolicy implements BlockLocationPolicy 
 
   /**
    * Calculate the available bytes for a worker with the added buffer of
-   * USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES
+   * {@link alluxio.conf.PropertyKey.USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES}
    * (alluxio.user.block.avoid.eviction.policy.reserved.size.bytes)
    *
    * Since the information of BlockWorkerInfo is updated <em>after</em> a file
--- a/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstPolicy.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstPolicy.java
@@ -36,6 +36,9 @@ import javax.annotation.concurrent.ThreadSafe;
  * exist or have enough availability, will select the nearest worker from the active
  * workers list with sufficient availability.
  *
+ * The definition of 'nearest worker' is based on {@link alluxio.wire.TieredIdentity}.
+ * @see alluxio.wire.TieredIdentityUtils#nearest()
+ *
  * The calculation of which worker gets selected is done for each block write.
  */
 @ThreadSafe
--- a/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstPolicy.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstPolicy.java
@@ -32,8 +32,11 @@ import java.util.stream.Collectors;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * A policy that returns local host first, and if the local worker doesn't have enough availability,
- * it randomly picks a worker from the active workers list for each block write.
+ * A policy that returns the local worker first, and if the local worker doesn't
+ * exist or have enough availability, will select the nearest worker from the active
+ * workers list with sufficient availability.
+ *
+ * The calculation of which worker gets selected is done for each block write.
  */
 @ThreadSafe
 public final class LocalFirstPolicy implements BlockLocationPolicy {
--- a/core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
@@ -24,11 +24,8 @@ import com.google.common.base.Preconditions;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * Always returns a worker with the specified hostname. Returns null if no active worker on that
- * hostname found.
- *
- * Note that since this does not have an empty constructor, this cannot
- * be used as a default policy.
+ * Always returns a worker with the hostname specified by
+ * {@link PropertyKey.WORKER_HOSTNAME} (alluxio.worker.hostname).
  */
 @ThreadSafe
 public final class SpecificHostPolicy implements BlockLocationPolicy {
@@ -52,6 +49,10 @@ public final class SpecificHostPolicy implements BlockLocationPolicy {
     mHostname = Preconditions.checkNotNull(hostname, "hostname");
   }
 
+  /**
+   * Returns null if no active worker matches the hostname
+   * provided in WORKER_HOSTNAME (alluxio.worker.hostname).
+   */
   @Override
   public WorkerNetAddress getWorker(GetWorkerOptions options) {
     // find the first worker matching the host name
--- a/core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
@@ -26,6 +26,9 @@ import javax.annotation.concurrent.ThreadSafe;
 /**
  * Always returns a worker with the specified hostname. Returns null if no active worker on that
  * hostname found.
+ *
+ * Note that since this does not have an empty constructor, this cannot
+ * be used as a default policy.
  */
 @ThreadSafe
 public final class SpecificHostPolicy implements BlockLocationPolicy {
--- a/docs/en/api/FS-API.md
+++ b/docs/en/api/FS-API.md
@@ -86,8 +86,8 @@ FileSystem fs = FileSystem.Factory.get();
 All metadata operations as well as opening a file for reading or creating a file for writing are
 executed through the `FileSystem` object. Since Alluxio files are immutable once written, the
 idiomatic way to create files is to use
-[FileSystem#createFile(AlluxioURI)](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/file/FileSystem.html#createFile-alluxio.AlluxioURI-)
-, which returns a stream object that can be used to write the file. For example:
+[FileSystem#createFile(AlluxioURI)](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/file/FileSystem.html#createFile-alluxio.AlluxioURI-),
+which returns a stream object that can be used to write the file. For example:
 
 ```java
 FileSystem fs = FileSystem.Factory.get();
@@ -220,7 +220,12 @@ Users can override the default policy class in the
   > exist or have enough availability, will select the nearest worker from the active
   > workers list with sufficient availability.
   >
+  > The definition of 'nearest worker' is based on
+  > ['Tiered Locality']({{ '/en/operation/Tiered-Locality.html' | relativize_url }}).
+  >
   > The calculation of which worker gets selected is done for each block write.
+
+  * If no worker meets availability criteria, will randomly select a worker from the list of all workers.
 
 * [LocalFirstAvoidEvictionPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.html)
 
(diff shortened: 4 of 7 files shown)


docs/en/api/FS-API.md
```diff
@@ -187,23 +212,40 @@ Users can override the default policy class in the
 [configuration file]({{ '/en/operation/Configuration.html' | relativize_url }}) at property
 `alluxio.user.block.write.location.policy.class`. The built-in policies include:
 
-* **LocalFirstPolicy (alluxio.client.block.policy.LocalFirstPolicy)**
+* [LocalFirstPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/LocalFirstPolicy.html)
+
+  **This is the default policy.**
+
+  > A policy that returns local host first, and if the local worker doesn't have enough availability,
```
Xenorith, line 219: `A policy that returns the worker local to the client first`

i noticed we don't use `local host` anywhere, usually referred to as `localhost`.  but localhost refers to the machine rather than the worker, which i think is what this is referring to

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,40 @@ Users can override the default policy class in the
...
+  > it randomly picks a worker from the active workers list for each block write.
+
+* [LocalFirstAvoidEvictionPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.html)
+
+  This is the same as `LocalFirstPolicy` with the following addition:
 
-    Returns the local worker first, and if it does not have enough capacity of a block,
-    randomly picks a worker from the active workers list. This is the default policy.
+  > If no worker meets the demands, return local host. USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES
```
Xenorith, line 226: ditto here about `local host`. i think it is more proper to use local worker

can you provide a bit more context on what `USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES ` is? right now i'm reading it as a plain string; is it a configuration property? a metric key? also add backticks around it

also i'm just noticing, did this policy exist before? or did we just not document it?

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,40 @@ Users can override the default policy class in the
...
-    Chooses the worker for the next block in a round-robin manner and skips workers that do not have
-    enough capacity.
+  > A policy that chooses the worker for the next block in a round-robin manner
+  > and skips workers that do not have enough space. The policy returns null if no worker can be found.
 
-* **SpecificHostPolicy (alluxio.client.block.policy.SpecificHostPolicy)**
+* [SpecificHostPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/SpecificHostPolicy.html)
 
-    Returns a worker with the specified host name. This policy cannot be set as default policy.
+  > Always returns a worker with the specified hostname. Returns null if no active worker on that hostname found.
```
Xenorith, line 241: did you purposefully omit the `This policy cannot be set as the default policy` line?

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,40 @@ Users can override the default policy class in the
...
-    randomly picks a worker from the active workers list. This is the default policy.
+  > If no worker meets the demands, return local host. USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES
+  > is used to reserve some space of the worker to store the block, for the values
+  > mCapacityBytes minus mUsedBytes is not the available bytes.
 
-* **MostAvailableFirstPolicy (alluxio.client.block.policy.MostAvailableFirstPolicy)**
+* [MostAvailableFirstPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/MostAvailableFirstPolicy.html)
 
-    Returns the worker with the most available bytes.
+  > A policy that returns the worker with the most available bytes. The policy returns null if no worker is qualified.
```
Xenorith, line 232: from the perspective of someone not reading the codebase, what behavior does `policy returns null` result in? does alluxio end up not writing the data to any worker?

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,40 @@ Users can override the default policy class in the
...
 
-* **SpecificHostPolicy (alluxio.client.block.policy.SpecificHostPolicy)**
+* [SpecificHostPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/SpecificHostPolicy.html)
 
-    Returns a worker with the specified host name. This policy cannot be set as default policy.
+  > Always returns a worker with the specified hostname. Returns null if no active worker on that hostname found.
+
+* [DeterministicHashPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/DeterministicHashPolicy.html)
+
+  > This policy maps blockId to several deterministic Alluxio workers.
```
Xenorith, line 245: maps the blockId

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,40 @@ Users can override the default policy class in the
...
+  > The number of workers a block can be mapped to can be passed through the constructor.
+  > The default is 1. It skips the workers that do not have enough capacity to hold the block.
+  > Note that the hash function relies on the number of workers in the cluster, so if the number of workers changes, the workers chosen by the policy for a given block will likely change. This policy is useful for limiting the amount of replication that occurs when reading blocks from the UFS with high concurrency. With 30 workers and 100 remote clients reading the same block concurrently, the replication level for the block would get close to 30 as each workers reads and caches the block for one or more clients. If the clients use DeterministicHashPolicy with 3 shards, the 100 clients will split their reads between just 3 workers, so that the replication level for the block will be only 3 when the data is first loaded.
```
Xenorith, line 248: 👍  the explanation makes a lot of sense even though i had no idea what this does.

split this up into separate lines for editability

`Note that the hash function relies on the number of workers in the cluster, so if the number of workers changes, the workers chosen by the policy for a given block will likely change.` feels like it belongs at the end as a warning, after explaining how the policy internals operate.

docs/en/api/FS-API.md
```diff
@@ -300,8 +319,8 @@ worker on each compute node.
 Alluxio has a [Python Client](https://github.com/Alluxio/alluxio-py) for interacting with Alluxio through its
 [REST API](#rest-api). The Python client exposes an API similar to the [Alluxio Java API](#java-client).
 See the [doc](http://alluxio-py.readthedocs.io) for detailed documentation about all available
-methods. See the [example](https://github.com/Alluxio/alluxio-py/blob/master/example.py) of how to perform basic filesystem
-operations in Alluxio.
+methods. See the [example](https://github.com/Alluxio/alluxio-py/blob/master/example.py) of how to perform basic
```
Xenorith, line 322: See the example `on` how to

Xenorith (changes requested):

docs/en/api/FS-API.md
ZhuTopher: I copied the descriptions verbatim from the Javadoc descriptions as I felt it'd be best to have the doc & Javadoc align; but I can fix that here for now.

docs/en/api/FS-API.md
ZhuTopher: I don't know when it was added; I noticed it wasn't documented when I looked at the Javadoc so I added it here. I'll look into what the `USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES` is and give a description underneath the quoted block.

docs/en/api/FS-API.md
ZhuTopher: Since I copied the description verbatim from the Javadoc, I guess so yeah. I didn't notice that so I'll add it back outside of the quoted block, thanks!

docs/en/api/FS-API.md
ZhuTopher: I actually don't know what happens semantically when a "policy returns null", maybe someone else can chime in?

Xenorith: i think this is code is pretty old, so maybe @calvinjia ?

docs/en/api/FS-API.md
ZhuTopher: I updated the source-code's javadoc description of `getAvailableBytes()` based on what I could understand, but if someone could fact-check me on that that'd be much appreciated!

core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
```diff
@@ -78,9 +83,12 @@ public WorkerNetAddress getWorker(GetWorkerOptions options) {
...
   /**
-   * The information of BlockWorkerInfo is update after a file complete write. To avoid evict,
-   * user should configure "alluxio.user.block.avoid.eviction.policy.reserved.size.bytes"
-   * to reserve some space to store the block.
+   * Calculate the available bytes for a worker with the added buffer of
+   * USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES
+   * (alluxio.user.block.avoid.eviction.policy.reserved.size.bytes)
+   *
+   * Since the information of BlockWorkerInfo is updated <em>after</em> a file
+   * completes a write, mCapacityBytes minus mUsedBytes may not be the true available bytes.
```
ZhuTopher, line 91: Fact-check me on this please; I just updated this according to the prior description.

ZhuTopher, line 91: Fact-check me on this please; I just updated this according to the prior description.

core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
```diff
@@ -26,6 +26,9 @@
 /**
  * Always returns a worker with the specified hostname. Returns null if no active worker on that
  * hostname found.
+ *
+ * Note that since this does not have an empty constructor, this cannot
+ * be used as a default policy.
```
ZhuTopher, line 31: Is this actually true? This is what the comments in `BlockLocationPolicy` said, but neither `LocalFirstPolicy` nor `LocalFirstAvoidEvictionPolicy` have empty constructors from what I saw.

ZhuTopher (commented):
I can't "Request Changes" on my own PR but could someone please address my comments? Thank you!

Xenorith (approved):
the changes look good to me, but i cannot validate the veracity of some of the new additions

core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
```diff
@@ -27,11 +27,16 @@
...
 import javax.annotation.concurrent.ThreadSafe;
 /**
- * A policy that returns local host first, and if the local worker doesn't have enough availability,
- * it randomly picks a worker from the active workers list for each block write.
- * If no worker meets the demands, return local host.
- * USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES is used to reserve some space of the worker
- * to store the block, for the values mCapacityBytes minus mUsedBytes is not the available bytes.
+ * A policy that returns the local worker first, and if the local worker doesn't
+ * exist or have enough availability, will select the nearest worker from the active
+ * workers list with sufficient availability.
```
jiacheliu3, line 32: Oh nice catch! Maybe you want to further explain "nearest" here. The definition of near is based on tier identity (calling `BlockLocationUtils.nearest()` -> `TiereddentityUtils.nearest)`

core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
jiacheliu3: Good call. Can you try running locally with this policy? I don't think this policy is documented enough both on the purpose and how to use it.

docs/en/api/FS-API.md
```diff
@@ -82,9 +84,10 @@ FileSystem fs = FileSystem.Factory.get();
 #### Creating a File
 
 All metadata operations as well as opening a file for reading or creating a file for writing are
-executed through the FileSystem object. Since Alluxio files are immutable once written, the
-idiomatic way to create files is to use `FileSystem#createFile(AlluxioURI)`, which returns
-a stream object that can be used to write the file. For example:
+executed through the `FileSystem` object. Since Alluxio files are immutable once written, the
+idiomatic way to create files is to use
+[FileSystem#createFile(AlluxioURI)](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/file/FileSystem.html#createFile-alluxio.AlluxioURI-)
+, which returns a stream object that can be used to write the file. For example:
```
jiacheliu3, line 90: nit: maybe the comma belongs to the previous line?

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,66 @@ Users can override the default policy class in the
...
+  > The calculation of which worker gets selected is done for each block write.
+
+* [LocalFirstAvoidEvictionPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.html)
+
+  This is the same as `LocalFirstPolicy` with the following addition:
 
-    Returns the local worker first, and if it does not have enough capacity of a block,
-    randomly picks a worker from the active workers list. This is the default policy.
+  > A policy that returns the local worker first, and if the local worker doesn't
+  > exist or have enough availability, will select the nearest worker from the active
```
jiacheliu3, line 230: You might wanna add definition of "nearest" by pointing to the tiered identity doc page.

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,66 @@ Users can override the default policy class in the
...
+  > is used as buffer space on each worker when calculating available space
+  > to store each block.
 
-* **MostAvailableFirstPolicy (alluxio.client.block.policy.MostAvailableFirstPolicy)**
+* [MostAvailableFirstPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/MostAvailableFirstPolicy.html)
 
-    Returns the worker with the most available bytes.
+  > A policy that returns the worker with the most available bytes.
+  >
+  > The policy returns null if no worker is qualified.
```
jiacheliu3, line 244: I think if this policy returns null, the client will select someone else. So from a user perspective, they don't need to know that the method can return null (nor should they worry about it). Let's just mention the behavior, which is (correct me if i'm wrong): local worker not available -> select random or select nearest?

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,66 @@ Users can override the default policy class in the
...
 
-* **RoundRobinPolicy (alluxio.client.block.policy.RoundRobinPolicy)**
+* [RoundRobinPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/RoundRobinPolicy.html)
 
-    Chooses the worker for the next block in a round-robin manner and skips workers that do not have
-    enough capacity.
+  > A policy that chooses the worker for the next block in a round-robin manner
+  > and skips workers that do not have enough space.
+  >
+  > The policy returns null if no worker can be found.
```
jiacheliu3, line 251: Same, let's describe the behavior instead of explaining internals.

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,66 @@ Users can override the default policy class in the
...
 
-* **SpecificHostPolicy (alluxio.client.block.policy.SpecificHostPolicy)**
+* [SpecificHostPolicy](https://docs.alluxio.io/os/javadoc/{{site.ALLUXIO_MAJOR_VERSION}}/alluxio/client/block/policy/SpecificHostPolicy.html)
 
-    Returns a worker with the specified host name. This policy cannot be set as default policy.
+  > Always returns a worker with the specified hostname.
+  >
+  > Returns null if no active worker on that hostname found.
```
jiacheliu3, line 257: Same comment on null

docs/en/api/FS-API.md
```diff
@@ -187,23 +212,66 @@ Users can override the default policy class in the
...
+  > Note that since this does not have an empty constructor, this cannot be used as a default policy.
```
jiacheliu3, line 258: This is also too much internals. 

If this policy cannot be used, let's not even mention it in the doc.

If this policy works, could you elaborate how to "specify some hosts"?

core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstPolicy.java
```diff
@@ -32,8 +32,11 @@
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * A policy that returns local host first, and if the local worker doesn't have enough availability,
- * it randomly picks a worker from the active workers list for each block write.
+ * A policy that returns the local worker first, and if the local worker doesn't
+ * exist or have enough availability, will select the nearest worker from the active
```
jiacheliu3, line 36: Same here, could you add some more on "nearest"?

core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
jiacheliu3: this is the fact

core/client/fs/src/main/java/alluxio/client/block/policy/LocalFirstAvoidEvictionPolicy.java
```diff
@@ -78,9 +83,12 @@ public WorkerNetAddress getWorker(GetWorkerOptions options) {
   }
 
   /**
-   * The information of BlockWorkerInfo is update after a file complete write. To avoid evict,
-   * user should configure "alluxio.user.block.avoid.eviction.policy.reserved.size.bytes"
-   * to reserve some space to store the block.
+   * Calculate the available bytes for a worker with the added buffer of
+   * USER_FILE_WRITE_AVOID_EVICTION_POLICY_RESERVED_BYTES
+   * (alluxio.user.block.avoid.eviction.policy.reserved.size.bytes)
```
jiacheliu3, line 88: Maybe change this string property name to a link to the property?

jiacheliu3 (changes requested):
@ZhuTopher Thanks for the improvements! This looks very helpful (although I left some comments).

jiacheliu3:
One high level comment is: I think it's fine to go into details (like methods and return values etc) but in the user-facing doc, we should focus on the behaviors only. The users will read the doc and choose between options. Let's just give them enough information that they can make the right choice. In other words, the doc should be different from the code comments.

Oh and pls fix checkstyle too.

core/client/fs/src/main/java/alluxio/client/block/policy/SpecificHostPolicy.java
jiacheliu3: If this policy works, can you add a few lines (explaining how the configuration should look like) both in the code comment and the doc? From the code it seems this will make the allocation only go to the WORKER_HOSTNAME?

docs/en/api/FS-API.md
ZhuTopher: I asked in the Dev channel on Slack and it seems no one is familiar enough with this part of the code to say what exactly happens when the returned worker is `null`. When I tested it appears that a worker is selected at random.

docs/en/api/FS-API.md
ZhuTopher: I tested this policy, it does indeed put all blocks onto the worker with the hostname specified by `alluxio.worker.hostname`. Interestingly, when short-circuit + domain socket were enabled they bypassed the `alluxio.user.block.write.location.policy.class=SpecificHostPolicy` setting and just wrote to the local worker instead.

I'm not sure if that's correct behaviour or not; correct for the short-circuit but incorrect for the policy.

HelloHorizon:

HelloHorizon:

Issue #13595, opened 2021-06-08
Format error when hostname is different with dns hostname


maobaolong:
**Is your feature request related to a problem? Please describe.**

```
2021-05-24 22:24:27,660 ERROR Format - Failed to format
java.lang.IllegalStateException: The cluster addresses ([my-alluxio-master1.xxx.com:19200, my-alluxio-master2.xxx.com:19200, my-alluxio-master3.xxx.com:19200]) must contain the local master address (HOST-192-168-100-111:19200)
        at com.google.common.base.Preconditions.checkState(Preconditions.java:824)
```

![image](https://user-images.githubusercontent.com/17329931/121108613-a3679680-c83c-11eb-8c37-1f360af37f3c.png)

**Describe the solution you'd like**

We can use nslookup tools to resolve the hostname from dns server and cache it for this Process.

**Describe alternatives you've considered**

Run 

**Urgency**

Normal

**Additional context**

ssz1997:
Fixed

PR #13620, opened 2021-06-10
Fix format error when raft cluster hostname different with localhost

bzheng888:
This pr is to fix issue #13595
 If hostname is not one of raft cluster hostname list, add a second round check related ip. 
In some case, we can't change cluster adresss dns name specificed by config file and the node's hostname, so maybe  local hostname is not one of raft cluster hostname list, then the format action will throw error in the validate process.  
No change

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
@@ -57,7 +57,9 @@ public class RaftJournalConfiguration {
     Preconditions.checkState(getMaxLogSize() <= Integer.MAX_VALUE,
         "{} has value {} but must not exceed {}", PropertyKey.MASTER_JOURNAL_LOG_SIZE_BYTES_MAX,
         getMaxLogSize(), Integer.MAX_VALUE);
-    Preconditions.checkState(getClusterAddresses().contains(getLocalAddress()),
+    Preconditions.checkState(getClusterAddresses().contains(getLocalAddress())
+            || NetworkAddressUtils.containsLocalIp(getClusterAddresses(),
+        ServerConfiguration.global()),
         "The cluster addresses (%s) must contain the local master address (%s)",
         getClusterAddresses(), getLocalAddress());
   }
--- a/core/common/src/main/java/alluxio/util/network/NetworkAddressUtils.java
+++ b/core/common/src/main/java/alluxio/util/network/NetworkAddressUtils.java
@@ -694,4 +694,28 @@ public final class NetworkAddressUtils {
       channel.shutdown();
     }
   }
+
+  /**
+   * @param clusterAddresses addresses of all nodes in the Raft cluster
+   * @param conf Alluxio configuration
+   * @return true if the cluster addresses contain the local IP, false otherwise
+   */
+  public static boolean containsLocalIp(List<InetSocketAddress> clusterAddresses,
+      AlluxioConfiguration conf) {
+    String localAddressIp = getLocalIpAddress((int) conf.getMs(PropertyKey
+        .NETWORK_HOST_RESOLUTION_TIMEOUT_MS));
+    for (InetSocketAddress addr : clusterAddresses) {
+      String clusterNodeIp;
+      try {
+        clusterNodeIp = InetAddress.getByName(addr.getHostName()).getHostAddress();
+        if (clusterNodeIp.equals(localAddressIp)) {
+          return true;
+        }
+      } catch (UnknownHostException e) {
+        LOG.error("Get raft cluster node ip by hostname({}) failed",
+            addr.getHostName(), e);
+      }
+    }
+    return false;
+  }
 }
--- a/core/common/src/test/java/alluxio/util/network/NetworkAddressUtilsTest.java
+++ b/core/common/src/test/java/alluxio/util/network/NetworkAddressUtilsTest.java
@@ -12,6 +12,8 @@
 package alluxio.util.network;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 import alluxio.ConfigurationRule;
 import alluxio.ConfigurationTestUtils;
@@ -29,6 +31,8 @@ import org.junit.Test;
 import java.io.Closeable;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Tests for the {@link NetworkAddressUtils} class.
@@ -40,6 +44,33 @@ public class NetworkAddressUtilsTest {
   @After
   public void after() {
     mConfiguration = ConfigurationTestUtils.defaults();
+  }
+
+  @Test
+  public void testContainsLocalIP() {
+    List<InetSocketAddress> clusterAddresses = new ArrayList<>();
+    InetSocketAddress raftNodeAddress1 = new InetSocketAddress(NetworkAddressUtils
+        .getLocalHostName(
+            (int) mConfiguration.getMs(PropertyKey.NETWORK_HOST_RESOLUTION_TIMEOUT_MS)),
+        10);
+    InetSocketAddress raftNodeAddress2 = new InetSocketAddress("host2", 20);
+    InetSocketAddress raftNodeAddress3 = new InetSocketAddress("host3", 30);
+    clusterAddresses.add(raftNodeAddress1);
+    clusterAddresses.add(raftNodeAddress2);
+    clusterAddresses.add(raftNodeAddress3);
+    assertTrue(NetworkAddressUtils.containsLocalIp(clusterAddresses, mConfiguration));
+  }
+
+  @Test
+  public void testNotContainsLocalIP() {
+    List<InetSocketAddress> clusterAddresses = new ArrayList<>();
+    InetSocketAddress raftNodeAddress1 = new InetSocketAddress("host1", 10);
+    InetSocketAddress raftNodeAddress2 = new InetSocketAddress("host2", 20);
+    InetSocketAddress raftNodeAddress3 = new InetSocketAddress("host3", 30);
+    clusterAddresses.add(raftNodeAddress1);
+    clusterAddresses.add(raftNodeAddress2);
+    clusterAddresses.add(raftNodeAddress3);
+    assertFalse(NetworkAddressUtils.containsLocalIp(clusterAddresses, mConfiguration));
   }
 
   /**
--- a/core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalConfigurationTest.java
+++ b/core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalConfigurationTest.java
@@ -15,6 +15,8 @@ import static org.junit.Assert.assertEquals;
 
 import alluxio.conf.PropertyKey;
 import alluxio.conf.ServerConfiguration;
+import alluxio.master.journal.JournalUtils;
+import alluxio.util.network.NetworkAddressUtils;
 import alluxio.util.network.NetworkAddressUtils.ServiceType;
 
 import com.google.common.collect.Sets;
@@ -22,8 +24,11 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.io.File;
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 
 /**
  * Units tests for {@link RaftJournalConfiguration}.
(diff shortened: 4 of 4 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +54,29 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
         .setPath(new File(JournalUtils.getJournalLocation().getPath()));
   }
 
+  /**
+   * @param clusterAddresses addresses of all nodes in the Raft cluster
+   * @return true if clusterAddress have an object of same ip with local Address, otherwise false
+   */
+  public boolean checkContainIp(List<InetSocketAddress> clusterAddresses) {
```
maobaolong, line 61: Make this method private as you just need to access it in this class.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +54,29 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
...
+    boolean contain = false;
```
maobaolong, line 62: contain -> exist

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +54,29 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
...
+    try {
+      for (InetSocketAddress addr : clusterAddresses) {
+        clusterItemIp = InetAddress.getByName(addr.getHostName()).getHostAddress();
+        if (clusterItemIp.equals(localAddressIp)) {
+          contain = true;
+          break;
+        }
+      }
+    } catch (Throwable t) {
+      return contain;
```
maobaolong, line 75: We need an error level log here for helping us to find the root cause why it return false.

maobaolong (changes requested):
Left minor comments inline, otherwise, look good overall.

BTW, if you have some tests locally, please feel free to share the result here.

maobaolong (approved):
Thank for this improvement, It LGTM.

maobaolong:
@apc999 Would you please take another look at this PR? Thanks.

LuQQiu (commented):
@bzheng888 where is the check result caching logics? the check looks heavy

bzheng888:
@LuQQiu I think the check logic is acceptable. the second round check works only when the first check failed, so actually the origin logic is not change. Secondly, although the second round check do not have a cache logic, it takes only one loop to match up,  as we can use the InetAddress address cache after the first lookup, PTAL.
image
image

maobaolong:
@bzheng888 Sounds good, it looks like the InetAddress.getByName supply the cache logic already.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,30 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
...
+  private boolean checkContainIp(List<InetSocketAddress> clusterAddresses) {
+    boolean exist = false;
+    String clusterItemIp;
+    String localAddressIp =  NetworkAddressUtils.getLocalIpAddress((int) ServerConfiguration
+        .global().getMs(PropertyKey.NETWORK_HOST_RESOLUTION_TIMEOUT_MS));
+    try {
+      for (InetSocketAddress addr : clusterAddresses) {
+        clusterItemIp = InetAddress.getByName(addr.getHostName()).getHostAddress();
+        if (clusterItemIp.equals(localAddressIp)) {
+          exist = true;
```
LuQQiu, line 74: remove `exist` variable and directly return true here and return false in exception and in the end of the method block?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,30 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
...
+          break;
+        }
+      }
+    } catch (UnknownHostException e) {
+      LOG.error("Get raft cluster node ip by hostname failed", e);
```
LuQQiu, line 79: is it possible that cluster addresses contain host A, host B, host C.
Host B is the local host and can be resolved, but the method error out when resolving host A?
will it cause issues in this case

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,30 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
         .setPath(new File(JournalUtils.getJournalLocation().getPath()));
   }
 
+  /**
+   * @param clusterAddresses addresses of all nodes in the Raft cluster
+   * @return true if clusterAddress have an object of same ip with local Address, otherwise false
+   */
+  private boolean checkContainIp(List<InetSocketAddress> clusterAddresses) {
```
LuQQiu, line 65: checkContainIp -> checkContainLocalIp

LuQQiu (changes requested):
Thanks for the detailed explanation of the question. The feature sounds safe (only take action when first check failed) and efficient (loop for once). Left some comments

LuQQiu:
@ggezer PTAL this PR, thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
bzheng888: @LuQQiu Thanks for the suggestion. I think this may happen and cause error, I will check again.

LuQQiu:
Ping @ggezer PTAL, thanks!

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,30 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
...
+   * @return true if the cluster addresses contain the local IP, false otherwise
+   */
+  private boolean checkContainLocalIp(List<InetSocketAddress> clusterAddresses) {
+    boolean exist = false;
+    String clusterItemIp;
+    String localAddressIp =  NetworkAddressUtils.getLocalIpAddress((int) ServerConfiguration
+        .global().getMs(PropertyKey.NETWORK_HOST_RESOLUTION_TIMEOUT_MS));
+    for (InetSocketAddress addr : clusterAddresses) {
+      try {
+        clusterItemIp = InetAddress.getByName(addr.getHostName()).getHostAddress();
```
ggezer, line 72: nit: clusterNodeIp?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
ggezer: +1

ggezer (changes requested):

bzheng888:
@LuQQiu @maobaolong @ggezer Thanks for reviewing this, PTAL.

LuQQiu (approved):
LGTM， thanks for improving the ip check!

ggezer (approved):
Thanks!

LuQQiu:
Error: 3.473 [ERROR] Errors: 
Error: 3.473 [ERROR] alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterCatchup
Error: 3.474 [ERROR]   Run 1: RaftJournalTest.gainPrimacyAfterCatchup:390->promoteFollower:404 » Timeout Tim...
Error: 3.474 [ERROR]   Run 2: RaftJournalTest.gainPrimacyAfterCatchup:390->promoteFollower:404 » Timeout Tim...
17:41:03.474 [INFO] 
Error: 3.474 [ERROR] alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterSuspend
Error: 3.474 [ERROR]   Run 1: RaftJournalTest.gainPrimacyAfterSuspend:350->promoteFollower:404 » Timeout Tim...
Error: 3.474 [ERROR]   Run 2: RaftJournalTest.gainPrimacyAfterSuspend:350->promoteFollower:404 » Timeout Tim...

LuQQiu:
RaftJournalTest failed consistently. @bzheng888 PTAL, thanks

LuQQiu:
@bzheng888 can you merge master branch into you master-bzheng branch? I doubt that RaftJournalTest is already fixed in the master branch

bzheng888:
@LuQQiu Got it, thanks for the suggestion, I'll have a try.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,28 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
...
+        .global().getMs(PropertyKey.NETWORK_HOST_RESOLUTION_TIMEOUT_MS));
+    for (InetSocketAddress addr : clusterAddresses) {
+      try {
+        clusterNodeIp = InetAddress.getByName(addr.getHostName()).getHostAddress();
+        if (clusterNodeIp.equals(localAddressIp)) {
+          return true;
+        }
+      } catch (UnknownHostException e) {
+        LOG.error(String.format("Get raft cluster node ip by hostname(%s) failed",
+            addr.getHostName()), e);
```
apc999, line 77:
```suggestion
        LOG.error("Get raft cluster node ip by hostname ({}) failed",
            addr.getHostName(), e);
```

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -63,7 +90,8 @@ public void validate() {
     Preconditions.checkState(getHeartbeatIntervalMs() < getElectionTimeoutMs() / 2,
         "Heartbeat interval (%sms) should be less than half of the election timeout (%sms)",
         getHeartbeatIntervalMs(), getElectionTimeoutMs());
-    Preconditions.checkState(getClusterAddresses().contains(getLocalAddress()),
+    Preconditions.checkState(getClusterAddresses().contains(getLocalAddress())
+            || checkContainLocalIp(getClusterAddresses()),
```
apc999, line 94: can you add a unit test on this?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,28 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
         .setPath(new File(JournalUtils.getJournalLocation().getPath()));
   }
 
+  /**
+   * @param clusterAddresses addresses of all nodes in the Raft cluster
+   * @return true if the cluster addresses contain the local IP, false otherwise
+   */
+  private boolean checkContainLocalIp(List<InetSocketAddress> clusterAddresses) {
+    String clusterNodeIp;
+    String localAddressIp =  NetworkAddressUtils.getLocalIpAddress((int) ServerConfiguration
```
apc999, line 67:
```suggestion
    String localAddressIp = NetworkAddressUtils.getLocalIpAddress((int) ServerConfiguration
```

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,28 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
         .setPath(new File(JournalUtils.getJournalLocation().getPath()));
   }
 
+  /**
+   * @param clusterAddresses addresses of all nodes in the Raft cluster
+   * @return true if the cluster addresses contain the local IP, false otherwise
+   */
+  private boolean checkContainLocalIp(List<InetSocketAddress> clusterAddresses) {
+    String clusterNodeIp;
```
apc999, line 66: nit: can we define this variable inside the for-loop?
we don't need it outside the for-loop, right?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalConfiguration.java
```diff
@@ -53,6 +58,28 @@ public static RaftJournalConfiguration defaults(ServiceType serviceType) {
         .setPath(new File(JournalUtils.getJournalLocation().getPath()));
   }
 
+  /**
+   * @param clusterAddresses addresses of all nodes in the Raft cluster
+   * @return true if the cluster addresses contain the local IP, false otherwise
+   */
+  private boolean checkContainLocalIp(List<InetSocketAddress> clusterAddresses) {
```
apc999, line 65: nits:
- How about calling this method `containsLocalIp` -- the convention of this codebase is to have `checkFoo` as a method that returns void but throw exceptions if the input is not meeting expectation.
- seems no member variable is really used in this method, in this case, better making it static

apc999 (changes requested):
minor comments left. 
can you add a test on this?

bzheng888:
@apc999  Thanks for reviewing this,  I'll add a test and  fix the problems you mentioned.

LuQQiu:
@apc999 PTAL, thanks!

apc999 (changes requested):
the code change looks good to me
Could you add a unit test on method `containsLocalIp` --- actually feel free to make it a `static` method since no member variable is used. You may also consider move this static method to `NetworkAddressUtils`

bzheng888:
@apc999 I have a problem about the unit test, how can I construct a scene like this: clusterAddresses has hostname list that different with local hostname, but the local ip is one of  clusterAddress's ip. I thought maybe I can learn the  this :
image
but it seems that I can't set hostname and ip both, is there any other way to do this? Can you  give me some advice.

apc999:
> @apc999 I have a problem about the unit test to ask, how can I construct a scene like this: clusterAddresses has hostname list that different with local hostname, but the local ip is one of clusterAddress's ip. I thought maybe I can learn the this :
> image
> but it seems that I can't set hostname and ip both, is there any other way to do this? Can you give me some advice.

Can you create a list of `InetSocketAddress`  including elements of:
1 one `InetSocketAddress` with a local IP using https://docs.oracle.com/javase/7/docs/api/java/net/InetSocketAddress.html#InetSocketAddress(java.net.InetAddress,%20int). the local IP can be obtained by `NetworkAddressUtils.getLocalIpAddress`
2 a set of InetSocketAddress with host name set manually https://docs.oracle.com/javase/7/docs/api/java/net/InetSocketAddress.html#InetSocketAddress(java.lang.String,%20int)

And you pass this list of `InetSocketAddress` to `containsLocalIp` ?

maobaolong:
@bzheng888  Do not use close&open to trigger the ci-checks restart, you can take a look at the failed test, if you think it isn't related to this PR, just left a comment to clarify and left the failed ci test for maintainer.

apc999 (approved):

apc999:

LuQQiu:

PR #13652, opened 2021-06-16
Port changes of #13241 to branch-2.4 (Manual)

calvinjia:
In certain environments, the read and write timeout may not make sense
to be equal. For example, it may be desirable to fail fast on reads and
retry while allowing for a long timeout on writes because retries cannot
be attempted.

pr-link: Alluxio/alluxio#13241
change-id: cid-8326ed56a90e3cb3c9cb6bc1a9958295d321bf9d


ggezer (approved):

calvinjia:

PR #13654, opened 2021-06-16
Group mount tables and partitions in a database

### LINKED ISSUE #13529: Improve the attachdb mount logic of Alluxio SDS
If we attach a db like test into alluxio, so alluxio sds load the table and craete one mount point for each table, so the result like the following, let's imagine that if there are 1000+ tables, there are 1000+ mount point.

/catalog/test/tables/table1/hive	hdfs://ns1/user/tdw/warehouse/test.db/table1	hdfs	13.02PB	10.96PB	NO	NO
/catalog/test/tables/table2/hive	hdfs://ns2/user/tdw/warehouse/test.db/table2	hdfs	37.64PB	35.03PB	NO	NO
/catalog/test/tables/table3/hive	hdfs://ns1/user/tdw/warehouse/test.db/table3	hdfs	13.02PB	10.96PB	NO	NO
/catalog/test/tables/table4/hive	hdfs://ns1/user/tdw/warehouse/test.db/table4	hdfs	13.02PB	10.96PB	NO	NO
/catalog/test/tables/table5/hive	hdfs://ns1/user/tdw/warehouse/test.db/table5	hdfs	13.02PB	10.96PB	NO	NO
/catalog/test/tables/table6/hive	hdfs://ns2/user/tdw/warehouse/test.db/table6	hdfs	37.64PB	35.03PB	NO	NO
/catalog/test/tables/table7/hive	hdfs://ns2/user/tdw/warehouse/test.db/table7	hdfs	37.64PB	35.03PB	NO	NO
/catalog/test/tables/table8/hive	hdfs://ns2/user/tdw/warehouse/test.db/table8	hdfs	37.64PB	35.03PB	NO	NO
/catalog/test/tables/table9/hive	hdfs://ns3/user/tdw/warehouse/test.db/table9	hdfs	42.68PB	35.51PB	NO	NO
/catalog/test/tables/table10/hive	hdfs://ns1/user/tdw/warehouse/test.db/table10	hdfs	13.02PB	10.96PB	NO	NO
...
-->

In fact, we can reduce the mount point from 1000+ to the following 3, because we can analysis all table location and find they are in 3 hdfs nameservices, so we only need to create the following three mount point.

/catalog/test/dbs/test.part1/hive	hdfs://ns1/user/tdw/warehouse/test.db	hdfs	13.02PB	10.96PB	NO	NO
/catalog/test/dbs/test.part2/hive	hdfs://ns2/user/tdw/warehouse/test.db	hdfs	13.02PB	10.96PB	NO	NO
/catalog/test/dbs/test.part3/hive	hdfs://ns3/user/tdw/warehouse/test.db	hdfs	13.02PB	10.96PB	NO	NO

Then, we add the following path mapping to test udb.

hdfs://ns1/user/tdw/warehouse/test.db/table1 -> /catalog/test/dbs/test.part1/hive/table1
hdfs://ns1/user/tdw/warehouse/test.db/table2 -> /catalog/test/dbs/test.part2/hive/table2
hdfs://ns1/user/tdw/warehouse/test.db/table3 -> /catalog/test/dbs/test.part1/hive/table3
hdfs://ns1/user/tdw/warehouse/test.db/table4 -> /catalog/test/dbs/test.part1/hive/table4
hdfs://ns1/user/tdw/warehouse/test.db/table5 -> /catalog/test/dbs/test.part1/hive/table5
hdfs://ns1/user/tdw/warehouse/test.db/table6 -> /catalog/test/dbs/test.part2/hive/table6
hdfs://ns1/user/tdw/warehouse/test.db/table7 -> /catalog/test/dbs/test.part2/hive/table7
hdfs://ns1/user/tdw/warehouse/test.db/table8 -> /catalog/test/dbs/test.part2/hive/table8
hdfs://ns1/user/tdw/warehouse/test.db/table9 -> /catalog/test/dbs/test.part3/hive/table9
hdfs://ns1/user/tdw/warehouse/test.db/table10 -> /catalog/test/dbs/test.part1/hive/table10
...

Change the SDS HiveDatabase related code.

**Urgency**

Normal

**Additional context**

This improvement can pretty reduce mount points, and improve the sync table performance significantly.

dbw9580:

If multiple partitions in a table share the same storage location prefix, we can mount the parent table instead of mounting every partitions, thus reducing number of mount points.

Summary of the changes:

1. Some variables are renamed to better reflect their role and functionality: 
    * previously, `ufsUri`, `alluxioUri` and `hiveUfsUri` are used for both table and partition uris, and reused between the two sections where table and partitions mount are handled, respectively. This can be confusing when reading the code because one needs to track what a variable actually stands for, a table uri or a partition uri.
    * previously, `hiveUfsUri` is actually a string representation of a UFS URI (for both the table's and partitions'), and `ufsUri` and `hiveUfsUri` are real `AlluxioURI` objects. It is now split and renamed to `tableUfsPath` and `partitionUfsPath`, respectively.
2. The logic of grouping partition mounts is:
    1. for each partition of a table, if its storage location shares the same prefix with its containing table (`tableUfsUri.isAncestorOf(partitionUfsUri)`), it does not need to be mounted separately. We have already mounted the table, so the partition can be accessed through the table's path.
    2. otherwise, check a config entry to see of we allow partitions that reside outside the directory of the containing table. If so, mount the partition individually. If not, just ignore the partition.
    3. add path mappings for all partitions that share the common path prefix with the table.
3. The code throwing IOException now uses `String.format()` instead of string concatenations. 

Fixes #13529.

Yes. As of commit 5326f0c146e68f9d3923e2d42d2a2d3089f33d8f, the command line config opiton `catalog.db.ignore.udb.tables` is moved into the config file `catalog.db.config.file` and is thus deprecated.

### DIFF
--- a/table/server/common/src/main/java/alluxio/table/common/CatalogPathUtils.java
+++ b/table/server/common/src/main/java/alluxio/table/common/CatalogPathUtils.java
@@ -29,11 +29,18 @@ import org.slf4j.LoggerFactory;
  *                                                  /_internal_/...
  * /<catalog base dir>/<dbName2>/tables/<tableName3>/<udbType>/...
  *                                                  /_internal_/...
+ *
+ * Catalog paths for fragments look like:
+ * /<catalog base dir>/<dbName2>/fragments/<fragment1>/...
+ *                                        /<fragment2>/...
+ * A fragment of a database is a UFS location that stores a part of the tables
+ * of the database. A UDB may be split into several fragments.
  */
 public class CatalogPathUtils {
   private static final Logger LOG = LoggerFactory.getLogger(CatalogPathUtils.class);
   private static final String TABLES_ROOT = "tables";
   private static final String INTERNAL_ROOT = "_internal_";
+  private static final String FRAGMENTS_ROOT = "fragments";
 
   private CatalogPathUtils() {} // prevent instantiation
 
@@ -58,5 +65,33 @@ public class CatalogPathUtils {
     return new AlluxioURI(PathUtils
         .concatPath(ServerConfiguration.get(PropertyKey.TABLE_CATALOG_PATH), dbName, TABLES_ROOT,
             tableName, INTERNAL_ROOT));
+  }
+
+  /**
+   * @param dbName the database name
+   * @param fragmentUri fragment ufs uri
+   * @return the AlluxioURI for the fragments path of the database
+   */
+  public static AlluxioURI getFragmentsPath(String dbName, AlluxioURI fragmentUri) {
+    return new AlluxioURI(PathUtils.concatPath(
+        ServerConfiguration.get(PropertyKey.TABLE_CATALOG_PATH), dbName, FRAGMENTS_ROOT,
+        escapeFragmentUfsPath(fragmentUri)));
+  }
+
+  /**
+   * Escape a ufs URI to be embedded in a fragment path.
+   *
+   * @param fragmentUfsUri fragment URI to escape
+   * @return escaped fragment URI
+   */
+  public static String escapeFragmentUfsPath(AlluxioURI fragmentUfsUri) {
+    String scheme = fragmentUfsUri.getScheme();
+    scheme = scheme == null || scheme.isEmpty() ? "none" : scheme;
+    String authority = fragmentUfsUri.getAuthority().toString();
+    authority = authority == null || authority.isEmpty() ? "none" : authority;
+    String escapedFragmentPath = String.format("%s_%s", scheme, authority)
+        .replaceAll(":", "_")
+        .replaceAll("/", "_");
+    return escapedFragmentPath;
   }
 }
--- a/table/server/common/src/main/java/alluxio/table/common/udb/UdbContext.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/UdbContext.java
@@ -98,4 +98,12 @@ public class UdbContext {
   public AlluxioURI getTableLocation(String tableName) {
     return CatalogPathUtils.getTablePathUdb(mDbName, tableName, mType);
   }
+
+  /**
+   * @param ufsUri ufs location of the fragment
+   * @return the AlluxioURI of the fragment in Alluxio filesystem
+   */
+  public AlluxioURI getFragmentLocation(AlluxioURI ufsUri) {
+    return CatalogPathUtils.getFragmentsPath(mDbName, ufsUri);
+  }
 }
--- a/table/server/common/src/main/java/alluxio/table/common/udb/UdbUtils.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/UdbUtils.java
@@ -36,32 +36,31 @@ public class UdbUtils {
   /**
    * Mount ufs path to alluxio path.
    *
-   * @param tableName Table name
    * @param ufsUri the uri of ufs
-   * @param tableUri the alluxio uri for table
+   * @param alluxioUri alluxio uri of the mount point
    * @param udbContext udb context
    * @param udbConfiguration Udb configurations
-   * @return table uri
+   * @return mounted Alluxio path
    * @throws IOException
    * @throws AlluxioException
    */
-  public static String mountAlluxioPath(String tableName, AlluxioURI ufsUri, AlluxioURI tableUri,
-      UdbContext udbContext, UdbConfiguration udbConfiguration)
+  public static String mountAlluxioPath(AlluxioURI ufsUri, AlluxioURI alluxioUri,
+                                        UdbContext udbContext, UdbConfiguration udbConfiguration)
       throws IOException, AlluxioException {
     if (Objects.equals(ufsUri.getScheme(), Constants.SCHEME)) {
       // already an alluxio uri, return the alluxio uri
       return ufsUri.toString();
     }
     try {
-      tableUri = udbContext.getFileSystem().reverseResolve(ufsUri);
-      LOG.debug("Trying to mount table {} location {}, but it is already mounted at location {}",
-          tableName, ufsUri, tableUri);
-      return tableUri.getPath();
+      alluxioUri = udbContext.getFileSystem().reverseResolve(ufsUri);
+      LOG.debug("Trying to mount ufs location {}, but it is already mounted at location {}",
+          ufsUri, alluxioUri);
+      return alluxioUri.getPath();
     } catch (InvalidPathException e) {
       // ufs path not mounted, continue
     }
     // make sure the parent exists
-    udbContext.getFileSystem().createDirectory(tableUri.getParent(),
+    udbContext.getFileSystem().createDirectory(alluxioUri.getParent(),
         CreateDirectoryPOptions.newBuilder().setRecursive(true).setAllowExists(true).build());
     Map<String, String> mountOptionMap = udbConfiguration.getMountOption(
         String.format("%s://%s/", ufsUri.getScheme(), ufsUri.getAuthority().toString()));
@@ -75,10 +74,10 @@ public class UdbUtils {
         option.putProperties(entry.getKey(), entry.getValue());
       }
     }
-    udbContext.getFileSystem().mount(tableUri, ufsUri, option.build());
+    udbContext.getFileSystem().mount(alluxioUri, ufsUri, option.build());
 
-    LOG.info("mounted table {} location {} to Alluxio location {} with mountOption {}",
-        tableName, ufsUri, tableUri, option.build());
-    return tableUri.getPath();
+    LOG.info("mounted ufs location {} to Alluxio location {} with mountOption {}",
+        ufsUri, alluxioUri, option.build());
+    return alluxioUri.getPath();
   }
 }
--- a/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
+++ b/table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
@@ -13,6 +13,7 @@ package alluxio.table.under.hive;
 
 import alluxio.AlluxioURI;
 import alluxio.exception.AlluxioException;
+import alluxio.exception.InvalidPathException;
 import alluxio.exception.status.NotFoundException;
 import alluxio.grpc.table.ColumnStatisticsInfo;
 import alluxio.grpc.table.Layout;
@@ -53,10 +54,15 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
 
 /**
  * Hive database implementation.
@@ -72,6 +78,8 @@ public class HiveDatabase implements UnderDatabase {
   private final String mConnectionUri;
   /** the name of the hive db. */
   private final String mHiveDbName;
+  /** path translator that records mappings between ufs paths and Alluxio paths. */
+  private final PathTranslator mPathTranslator;
 
   private static final HiveClientPoolCache CLIENT_POOL_CACHE = new HiveClientPoolCache();
   /** Hive client is not thread-safe, so use a client pool for concurrency. */
@@ -94,6 +102,7 @@ public class HiveDatabase implements UnderDatabase {
     mConnectionUri = connectionUri;
     mHiveDbName = hiveDbName;
     mClientPool = clientPool;
+    mPathTranslator = new PathTranslator();
   }
 
   /**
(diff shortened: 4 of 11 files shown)


maobaolong:
@dbw9580 This PR looks like base on your bypass partition branch? Would you please rebase to master branch?

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -148,63 +149,82 @@ public String getName() {
...
       }
       return pathTranslator;
     } catch (AlluxioException e) {
-      throw new IOException(
-          "Failed to mount table location. tableName: " + tableName
-              + " hiveUfsLocation: " + hiveUfsUri
-              + " AlluxioLocation: " + alluxioUri
-              + " error: " + e.getMessage(), e);
+      throw new IOException(String.format(
+          "Failed to mount table location. " +
```
maobaolong, line 212: table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java:212:46: error: '+' should be on a new line.

maobaolong (commented):
A style issue, please fix it.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
dbw9580: Yep. Fixed in 3f089c6.

maobaolong (changes requested):
Thanks for the useful feature you did. After further review, please make sure this PR support group mount table in the same prefix.

A suggestion approach, we don't mount path one by one, but do a analysis and group the table in same prefix.

maobaolong:
@dbw9580 Could you please show some example to explain how this PR can reduce the number of mount points?

Or you can add some necessary test cases, it will also helps to let me understand the expected behavior.

dbw9580:
@maobaolong The logic is:
1. The mounting happens when the UDB is created. Previously, the mounting happens when each table is requested.
2. For example, when the UDB contains 3 tables that are spread across 2 HDFS nodes:
    Table 1: `hdfs://ns1/db1/table1`
    Table 2: `hdfs://ns2/db1/table2`
    Table 3: `hdfs://ns1/db1/table3`
    We find each table's parent path, and group the common prefixes: `hdfs://ns1/db1` and `hdfs://ns2/db2`. We create 2 mount points for these 2 UFS paths on `/catalog/db1/fragments/ns1` and `/catalog/db1/fragments/ns2`, respectively. Now the tables can be accessed by appending the table's name to the URIs: `/catalog/db1/fragments/ns1/table1`, `/catalog/db1/fragments/ns2/table2` and `/catalog/db1/fragments/ns1/table3`.
3. For partitions, there are 2 cases: one, if the partition  resides in the parent table's directory, it can be found with the URI of the table, e.g. a partition of table 1 `hdfs://ns1/db1/table1/part1` is mapped to `/catalog/db1/fragments/ns1/table1/part1`, so no special handling is needed. Two, if otherwise the partition is stored somewhere else, then mount the partition specifically, and create mapping for it.

maobaolong:
@dbw9580 It can be more accurate to include scheme into the Alluxio mount point, so the mount point would be /catalog/db1/fragments/hdfs/ns1 and /catalog/db1/fragments/hdfs/ns12 to handle the following case

Table 1: hdfs://ns1/db1/table1
Table 2: hdfs://ns2/db1/table2
Table 3: hdfs://ns1/db1/table3
Table 4: odfs://ns1/db1/table4

dbw9580:
Exactly. I left an escape method for this purpose. It's not really implemented for now, but we can discuss how it should be implemented.
https://github.com/Alluxio/alluxio/blob/57d2688f796a70e71a33b42b4641b19351e11e83/table/server/common/src/main/java/alluxio/table/common/CatalogPathUtils.java#L87-L90

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -146,70 +155,142 @@ public String getName() {
...
+   * Mounts the database to Alluxio filesystem. This method mounts tables and partitions
+   * (if applicable) within this database to Alluxio filesystem.
+   *
+   * @param bypassSpec bypass spec
+   */
+  private void mount(UdbBypassSpec bypassSpec) throws IOException {
+    List<String> tableNames = getTableNames();
+    Map<Table, List<Partition>> tables = new HashMap<>(tableNames.size());
+    for (String tableName : tableNames) {
+      Table table;
```
maobaolong, line 168: If we only use the `table` in the try block, just define it in the try block

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -196,27 +198,27 @@ private void mount(List<Table> tables, UdbBypassSpec bypassSpec)
         continue;
       }
       AlluxioURI tableUfsUri = new AlluxioURI(tableUfsPath);
-      AlluxioURI parentUfsUri = tableUfsUri.getParent();
-      colocatedTables.put(parentUfsUri, table);
+      AlluxioURI rootUfsUri = new AlluxioURI(tableUfsUri, "/", false);
+      colocatedTables.put(rootUfsUri, table);
```
dbw9580: After discussion with @maobaolong, I extended the scope of mount to the root of UFS. This is based on the assumption that if the table is accessible by Alluxio, then any leading path before the table, including the root, is accessible, too. In other words, we only need to mount the roots of every UFS to be able to access the tables contained in various locations in the UFS's, instead of finding and mounting them one by one. I tested it with a non-clustered setup of Hive and Presto, and it worked.

Would appreciate your comments on this approach, @apc999 .

table/server/common/src/main/java/alluxio/table/common/CatalogPathUtils.java
```diff
@@ -85,8 +85,13 @@ public static AlluxioURI getFragmentsPath(String dbName, AlluxioURI fragmentUri)
...
-    // e.g. path = path.replaceAll("/", "_")
-    return fragmentUfsUri.toString();
+    String scheme = fragmentUfsUri.getScheme();
+    scheme = scheme == null || scheme.isEmpty() ? "none" : scheme;
+    String authority = fragmentUfsUri.getAuthority().toString();
+    authority = authority == null || authority.isEmpty() ? "none" : authority;
+    String escapedFragmentPath = String.format("%s_%s", scheme, authority)
+        .replaceAll(":", "_")
+        .replaceAll("/", "_");
+    return escapedFragmentPath;
```
dbw9580: This is to make sure we have some kind of identifier to distinguish these fragments. Since we mount the root of each UFS, the path leading to the tables is no longer relevent.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -340,7 +334,11 @@ public UdbTable getTable(String tableName, UdbBypassSpec bypassSpec) throws IOEx
         }
       }
 
-      mount(bypassSpec);
+      if (!mIsMounted) {
+        // TODO(bowen): bypass spec updates are ignored
+        mount(bypassSpec);
+        mIsMounted = true;
+      }
```
dbw9580, line 341: This is problematic for now. If tables are dropped from outside of alluxio between two calls of getTable, then the path mapping info in the path translator will become stale.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -146,70 +155,142 @@ public String getName() {
...
-              mUdbContext,
-              mConfiguration),
-          hiveUfsUri);
+      AlluxioURI tableUfsUri = new AlluxioURI(tableUfsPath);
+      AlluxioURI rootUfsUri = new AlluxioURI(tableUfsUri, "/", false);
+      colocatedTables.put(rootUfsUri, table);
+    }
+    for (Map.Entry<AlluxioURI, Collection<Table>> entry : colocatedTables.asMap().entrySet()) {
+      AlluxioURI rootUfsUri = entry.getKey();
+      Collection<Table> childTables = entry.getValue();
```
maobaolong, line 206: It seems you defined this variable but never used it

maobaolong (commented):
Left a couple of comments, please take a double check.

dbw9580:
The lastest design of this feature is as follows:

Suppose we have 3 tables spread across two HDFS nodes:

```
Table 1: hdfs://ns1/path1/db1/table1
Table 2: hdfs://ns1/path2/db1/table2
Table 3: hdfs://ns2/path3/db1/table3
```

We create 2 mount points on `/catalog/db1/fragments/hdfs_ns1` and `/catalog/db1/fragments/hdfs_ns2`, pointing to `hdfs://ns1` and `hdfs://ns2`, respectively. We add path mappings as follows:

```
hdfs://ns1 -> /catalog/db1/fragments/hdfs_ns1
hdfs://ns2 -> /catalog/db1/fragments/hdfs_ns2
```

The tables will now be accessible from Alluxio on these Paths:

```
Table 1: /catalog/db1/fragments/hdfs_ns1/path1/db1/table1
Table 2: /catalog/db1/fragments/hdfs_ns1/path2/db1/table2
Table 3: /catalog/db1/fragments/hdfs_ns2/path3/db1/table3
```

If there are partitions in the tables, and they are not stored along with their parent tables physically, e.g. a partition `part1` of table 1 that is stored in another directory `path4` on ns1, and a partition `part2` of table 2 that is stored on another name node ns3:

```
Partition 1: hdfs://ns1/path4/db1/table1/part1
Partition 2: hdfs://ns3/path2/db1/table2/part2
```

They will be mounted individually. In summary, as long as a partition is not stored in the same directory tree of its parent table, then it will be handled as an exception and mounted specifically.

table/server/common/src/main/java/alluxio/table/common/udb/UdbContext.java
```diff
@@ -98,4 +98,12 @@ public String getUdbDbName() {
   public AlluxioURI getTableLocation(String tableName) {
     return CatalogPathUtils.getTablePathUdb(mDbName, tableName, mType);
```
maobaolong, line 99: This method cannot return the right AlluxioURI?

table/server/common/src/main/java/alluxio/table/common/udb/UdbUtils.java
```diff
@@ -45,23 +45,50 @@ private UdbUtils() {}
    * @throws IOException
    * @throws AlluxioException
    */
+  // Maybe rename to mountTable?
```
maobaolong, line 48: Remove this comment

table/server/common/src/main/java/alluxio/table/common/udb/UdbContext.java
dbw9580: This method will be deprecated once this PR gets approved, since the directory `/catalog/db/tables` will not be used any more.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
dbw9580: Removed in 047139a5.

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
dbw9580: Fixed in 047139a5b49da41a4c8da063897dd0f39fee5098.

maobaolong:
@dbw9580 Thanks for the improvement, @apc999 Would you please take a look at this PR, it change the catalogs mount point layout

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -146,72 +152,124 @@ public String getName() {
...
-    AlluxioURI alluxioUri = mUdbContext.getTableLocation(tableName);
-    String hiveUfsUri = table.getSd().getLocation();
+  /**
+   * Mounts the database to Alluxio filesystem. This method mounts tables and partitions
+   * (if applicable) within this database to Alluxio filesystem.
+   *
+   * @param bypassSpec bypass spec
+   */
+  private void mount(UdbBypassSpec bypassSpec) throws IOException {
+    List<String> tableNames = getTableNames();
```
maobaolong, line 162: This method will get all tables from udb, it don't respect the `catalog.db.ignore.udb.tables` options

dbw9580: This logic is handled by `alluxio.master.table.Database#sync` method. The ignored tables may be mounted along with others, but they will not be visible to Alluxio client.

dbw9580: Fixed in 5326f0c146e68f9d3923e2d42d2a2d3089f33d8f

table/server/common/src/main/java/alluxio/table/common/udb/UdbInExClusionSpec.java
```diff
@@ -15,44 +15,69 @@
...
    * Keyed by a table's name, the value set contains names of partitions in that table.
    * An empty set indicates all partitions of that table, if any, should be bypassed.
    */
-  private final Map<String, Set<String>> mTablePartMap;
+  private final Map<String, Set<String>> mBypassed;
 
   /**
-   * @param tablePartMap table to partition map
+   * Set of ignored tables.
+   * Tables are ignored as a whole, no partition configuration is needed.
```
apc999, line 30: then shall we name this `mIgnoredTables`

table/server/underdb/hive/src/main/java/alluxio/table/under/hive/HiveDatabase.java
```diff
@@ -253,15 +314,19 @@ public UdbTable getTable(String tableName, UdbBypassSpec bypassSpec) throws IOEx
         }
       }
 
-      PathTranslator pathTranslator = mountAlluxioPaths(table, partitions, bypassSpec);
+      if (!mIsMounted) {
```
maobaolong, line 317: This condition check is not thread-safe, suggest to move this mount logic before invoke tasks.

dbw9580:
Added commits from @maobaolong 's [branch](https://github.com/maobaolong/alluxio/commits/sds-tested), and added a switch `mount.group_mount_points` to control which mounting behaviour to use.

There are open questions on the new `mount(tableNames, bypassSpec)` method of the `UnderDatabase` interface:

1. when it is called multiple times with different sets of table names, there can be new tables to mount which are not present during a previous call; tables that were mounted during a previous call, but are removed in a second call. We should skip the tables that are present in both calls, and mount the new tables, and maybe unmount the removed tables. If group mount points is enabled, there can be updates to mount points as well;
2. Updates can happen to the bypass specification between two calls to `mount`. That means some tables were be ignored before and are no longer ignored now.

dbw9580:
@HelloHorizon @apc999 Rebased on current master, and added unit tests for the group mount feature. Please have a look, thanks!

Issue #13657, opened 2021-06-17
Alluxio master metrics web page hung for several minutes


maobaolong:
**Alluxio Version:**
master branch 

**Describe the bug**
Open metrics page of master `http://localhost:19999/metrics` cost several minutes while the master server running over 100 days. We should record only enough point for dataPoints of TimeSeries

**To Reproduce**

Running a master over 100 days.
Open  `http://localhost:19999/metrics` or open `http://localhost:8081/api/v1/master/webui_metrics`

![image](https://user-images.githubusercontent.com/17329931/122342102-c994e100-cf76-11eb-9e24-a59ab95a19a6.png)

We only use the latest 20 points of data.

![image](https://user-images.githubusercontent.com/17329931/122342335-18db1180-cf77-11eb-91b5-200dfe5b20e3.png)

**Expected behavior**

Open page in a second, but wait for minutes.

**Urgency**

High

**Additional context**

We should limit the size of `TimeSeries#mDataPoints` 

```java
public class TimeSeries {
  private final List<DataPoint> mDataPoints;
...
}

```

LuQQiu:
@tieujason330 can you help take a look this issue, thanks

apc999:
Fixed by #13660 (thanks for @tieujason330 )
Closing this issue and feel free to re-open it if you still it happening.

PR #13660, opened 2021-06-18
Use latest 20 time series data points for Web UI metrics

tieujason330:

Only send the latest 20 time series data points to the Web UI for graphing. Since the Web UI already only uses the latest 20, this will prevent a slowdown in the metrics page for a long running Alluxio cluster. 

Addresses #13657


apc999 (approved):

LuQQiu:

core/common/src/main/java/alluxio/wire/MasterWebUIMetrics.java
```diff
@@ -660,11 +660,11 @@ public MasterWebUIMetrics setOperationMetrics(Map<String, Metric> operationMetri
   }
 
   /**
-   * @param timeSeries the time series metrics to set
+   * @param timeSeries the time series metrics to set. The latest 20 data points will be set
    * @return the updated masterWebUIMetrics object
    */
   public MasterWebUIMetrics setTimeSeriesMetrics(List<TimeSeries> timeSeries) {
```
maobaolong, line 666: If we don't need so many  point at all, we should not record so many time series into `List<TimeSeries>`, any thoughts?

webui/master/src/store/metrics/test/reducer.test.tsx
```diff
@@ -1,143 +0,0 @@
-/*
```
maobaolong, line 1: Why to remove this file?

PR #13663, opened 2021-06-18
Fix PathTranslator to compare path prefixes by path components

dbw9580:

Previously, `PathTranslator` erroneously:

1. finds a prefix of `ufsPath` by simply checking with `String.startsWith()`. This fails to handle the case where the boundary falls within a path component. This case is now covered in the `prefixBoundaryWithinPathComponent` test;
2. chooses an arbitary prefix of `ufsPath` (the first returned during map iteration) even if there are multiple eligible choices. This is fixed by choosing the deepest prefix instead.

It's a bug fix.

No.

### DIFF
--- a/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
@@ -11,8 +11,11 @@
 
 package alluxio.table.common.udb;
 
+import alluxio.AlluxioURI;
 import alluxio.conf.ServerConfiguration;
+import alluxio.exception.InvalidPathException;
 import alluxio.util.ConfigurationUtils;
+import alluxio.util.io.PathUtils;
 
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
@@ -27,7 +30,7 @@ import java.io.IOException;
 public class PathTranslator {
   private static final Logger LOG = LoggerFactory.getLogger(PathTranslator.class);
 
-  private BiMap<String, String> mPathMap;
+  private final BiMap<AlluxioURI, AlluxioURI> mPathMap;
 
   /**
    * Construct a path translator.
@@ -45,16 +48,7 @@ public class PathTranslator {
    * @return PathTranslator object
    */
   public PathTranslator addMapping(String alluxioPath, String ufsPath) {
-    while (alluxioPath.endsWith("/")) {
-      // strip trailing slashes
-      alluxioPath = alluxioPath.substring(0, alluxioPath.length() - 1);
-    }
-
-    while (ufsPath.endsWith("/")) {
-      // strip trailing slashes
-      ufsPath = ufsPath.substring(0, ufsPath.length() - 1);
-    }
-    mPathMap.put(alluxioPath, ufsPath);
+    mPathMap.put(new AlluxioURI(alluxioPath), new AlluxioURI(ufsPath));
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
```diff
@@ -76,24 +76,24 @@ public void multipleSubdir() throws Exception {
 
   @Test
   public void samePathDifferentUfs() throws Exception {
-    String alluxioPath  = "alluxio:///my/table/directory";
+    String alluxioPath  = "alluxio://master/my/table/directory";
     String ufsPath = "/the/ufs/location";
-    mTranslator.addMapping(alluxioPath, "ufs_a://" + ufsPath);
+    mTranslator.addMapping(alluxioPath, "ufs-a://" + ufsPath);
```
yuzhu, line 81: why was this changed? does '_' not work?

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
```diff
@@ -76,24 +76,24 @@ public void multipleSubdir() throws Exception {
 
   @Test
   public void samePathDifferentUfs() throws Exception {
-    String alluxioPath  = "alluxio:///my/table/directory";
+    String alluxioPath  = "alluxio://master/my/table/directory";
```
yuzhu, line 79: we should also be able to support path like 'alluxio:///dir/file' does this change not work with that kind of path?

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -45,16 +48,7 @@ public PathTranslator() {
...
-      // strip trailing slashes
-      alluxioPath = alluxioPath.substring(0, alluxioPath.length() - 1);
-    }
-
-    while (ufsPath.endsWith("/")) {
-      // strip trailing slashes
-      ufsPath = ufsPath.substring(0, ufsPath.length() - 1);
-    }
-    mPathMap.put(alluxioPath, ufsPath);
+    mPathMap.put(new AlluxioURI(alluxioPath), new AlluxioURI(ufsPath));
```
yuzhu, line 51: what is the reason to move this to the translation time? would this add potentially duplicate keys like `/path1/path2' and `/path1/path2/` ?

yuzhu (commented):
@dbw9580 Thanks for the fix! good catch, have some questions about adding path and tests.

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
dbw9580: `java.lang.IllegalArgumentException: java.net.URISyntaxException: Illegal character in scheme name at index 3: ufs_a:/the/ufs/location`

It seems "_" is forbidden in schemes.

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
dbw9580: https://github.com/Alluxio/alluxio/blob/90b3c7b010f926be3c2df7519eec067df30fd2a7/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java#L96-100

I check for the existence of both scheme and authority. Is this too strict?

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
dbw9580: > move this to the translation time

Do you mean the creation of `AlluxioURI`s? I simply use them for the convenient methods like `isAncestorOf` and that it trims trailing slashes already.

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
yuzhu: I think it is ok to not have authority, specifically i think for HA URIs. see https://docs.alluxio.io/os/user/stable/en/deploy/Running-Alluxio-On-a-HA-Cluster.html#specify-alluxio-service-in-configuration-parameters-or-java-options

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
yuzhu: oh somehow i missed that you create AlluxioURIs out of them, yeah this works

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
yuzhu: yup, it makes sense now that I see you create AlluxioURI out of the path

yuzhu (approved):
LGTM once the test passes and URI without authority is addressed

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
dbw9580: Changed in 825c0cc.

dbw9580:
@yuzhu forgot to add test for the deepest match part. PTAL, thanks.

yuzhu (approved):

yuzhu:

PR #13665, opened 2021-06-19
Add CacheContext to URIStatus to enable per-read metrics

yuanzhanhku:
Add functions to URIStatus to track bytes read from cache and external
file system in a per-read basis.
The default implementation does nothing. It is up to the subclass to
decide how to track them.
Presto will use this information to compute query level metrics.
See https://github.com/prestodb/presto/issues/16237 for details.

There are no user facing changes in this commit.

### DIFF
--- a/core/common/src/main/java/alluxio/client/file/PrestoContext.java
+++ b/core/common/src/main/java/alluxio/client/file/PrestoContext.java
@@ -1,0 +1,104 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.client.file;
+
+import alluxio.client.quota.CacheQuota;
+import alluxio.client.quota.CacheScope;
+import alluxio.wire.FileInfo;
+
+import com.google.common.base.MoreObjects;
+
+import java.util.Objects;
+
+/**
+ * Context when running in Presto.
+ */
+public class PrestoContext {
+  private CacheQuota mCacheQuota = CacheQuota.UNLIMITED;
+  private CacheScope mCacheScope = CacheScope.GLOBAL;
+
+  /**
+   * Constructs PrestoContext.
+   */
+  public PrestoContext() {
+  }
+
+  /**
+   * @return the cache quota
+   */
+  public CacheQuota getCacheQuota() {
+    return mCacheQuota;
+  }
+
+  /**
+   * @return the cache scope
+   */
+  public CacheScope getCacheScope() {
+    return mCacheScope;
+  }
+
+  /**
+   * @param cacheQuota the cache quota
+   * @return the updated {@link FileInfo}
+   */
+  public PrestoContext setCacheQuota(CacheQuota cacheQuota) {
+    mCacheQuota = cacheQuota;
+    return this;
+  }
+
+  /**
+   * @param cacheScope the cache quota
+   * @return the updated {@link FileInfo}
+   */
+  public PrestoContext setCacheScope(CacheScope cacheScope) {
+    mCacheScope = cacheScope;
+    return this;
+  }
+
+  /**
+   * Increments the counter {@code name} by {@code value}.
+   * <p>
+   * Default implementation does nothing. Subclass can implement its own tracking mechanism.
+   *
+   * @param name name of the counter
+   * @param value value of the counter
+   */
+  public void incrementCounter(String name, long value) {
+    // Default implementation does nothing
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    PrestoContext that = (PrestoContext) o;
+    return Objects.equals(mCacheQuota, that.mCacheQuota) && Objects
+        .equals(mCacheScope, that.mCacheScope);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(mCacheQuota, mCacheScope);
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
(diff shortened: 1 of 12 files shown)


yuanzhanhku:
@apc999 Bin, could you please help review this PR? Thanks.

yuanzhanhku:
Hi Bin, I updated this commit to track arbitrary metrics.

I thought about using a map to track them in the URIStatus class itself. But I think it might be better to allow each subclass to have its own tracking mechanism. For example, in Presto, we can use the RuntimeStats to track it instead of a simple map. Let me know what you think. Thanks.

core/common/src/main/java/alluxio/client/file/URIStatus.java
```diff
@@ -310,17 +313,18 @@ public String getUfsFingerprint() {
...
+    return mPrestoContext;
   }
 
   /**
-   * @return the cache scope
+   * @param prestoContext Presto context to set
    */
-  public CacheScope getCacheScope() {
-    return mInfo.getCacheScope();
+  public void setPrestoContext(PrestoContext prestoContext) {
```
ggezer, line 326: from where will this be called?

core/common/src/main/java/alluxio/client/file/PrestoContext.java
```diff
@@ -0,0 +1,104 @@
...
+import alluxio.wire.FileInfo;
+
+import com.google.common.base.MoreObjects;
+
+import java.util.Objects;
+
+/**
+ * Context when running in Presto.
+ */
+public class PrestoContext {
```
ggezer, line 25: CacheContext?

core/common/src/main/java/alluxio/client/file/PrestoContext.java
```diff
@@ -0,0 +1,104 @@
...
+import alluxio.client.quota.CacheQuota;
+import alluxio.client.quota.CacheScope;
+import alluxio.wire.FileInfo;
+
+import com.google.common.base.MoreObjects;
+
+import java.util.Objects;
+
+/**
+ * Context when running in Presto.
```
ggezer, line 23: @apc999 quotas will only be supported for Presto?

yuanzhanhku: @apc999 I also have the same question.  Is this class specific for Presto? Is it better to make this class generic and add a Presto specific subclass in the Presto repo?

beinan: I think both quota and scope are generic fields, which would be helpful to presto, spark-sql and etc..
So we could rename this class to CacheContext, can we?

apc999: @ggezer @yuanzhanhku @beinan  these two fields are only added to support Presto previously. Note that all the other URIStatus field are retrieved from Alluxio master but these two are only set in Presto code https://github.com/prestodb/presto/blob/master/presto-cache/src/main/java/com/facebook/presto/cache/alluxio/AlluxioCachingFileSystem.java#L90-L92

I also agree that these are generic field but unfortunately they are not currently something supported generally by Alluxio. So I prefer to group these Presto-specific logic into PrestoContext to have better management. We can always move them out like other fields once they are supported

core/common/src/main/java/alluxio/client/file/URIStatus.java
apc999: this will be only called in Presto land, likely https://github.com/prestodb/presto/blob/master/presto-cache/src/main/java/com/facebook/presto/cache/alluxio/AlluxioCachingFileSystem.java#L80

core/common/src/main/java/alluxio/client/file/PrestoContext.java
apc999: > @apc999 I also have the same question. Is this class specific for Presto? Is it better to make this class generic and add a Presto specific subclass in the Presto repo?

I will create a subclass of this one in Presto-side, e.g., storing `HiveFileContext` and extend `incrementCounter`

core/common/src/main/java/alluxio/client/file/PrestoContext.java
apc999: After second thoughts, I agree `CacheContext` can be better,

core/common/src/main/java/alluxio/client/file/URIStatus.java
beinan: I have the similar question.   Why do we need a setter for PrestoContext? 

The existing calling chain is starting from AlluxioCachingFileSystem::openFile
```
if (hiveFileContext.getCacheQuota().getQuota().isPresent()) {
                    info.setCacheQuota(new CacheQuota(ImmutableMap.of(scope.level(), hiveFileContext.getCacheQuota().getQuota().get().toBytes())));
                }
                else {
                    info.setCacheQuota(CacheQuota.UNLIMITED);
                }
``` 
Can we put PrestoContext as a a parameter in the constructor of this class just like FileInfo?

core/common/src/main/java/alluxio/client/file/URIStatus.java
```diff
@@ -40,14 +40,17 @@
 @ThreadSafe
 public class URIStatus {
   private final FileInfo mInfo;
+  /** Context related to Presto runtime associated with this URI. */
+  private PrestoContext mPrestoContext;
```
beinan, line 44: How do we distinguish the responsibilities of info and context?  especially the two fields in prestoContext looks generic.  I might say the quota and scope should be under mInfo rather than mPrestoContext.

beinan, line 44: any reason we do not make mPrestoContext final?

ggezer: `FileInfo` is coming from Alluxio which doesn't have such concepts.

ggezer (approved):

yuanzhanhku (commented):
LGTM. Thanks for making the change.

core/common/src/main/java/alluxio/client/file/URIStatus.java
apc999: Actually, I would like to consolidate `FileInfo` with `URIStatus` too. There are historical reasons these are two different classes.

apc999: changed to `final`. thanks

apc999 (approved):
Failing tests `Error: 1.984 [ERROR] alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterCatchup` is unrelated.

apc999:

apc999:

PR #13456, opened 2021-05-20
Improve error message for taking snapshot on single master

bf8086:
Display a more actionable message instead of complaining no follower available when running in single master mode.

PR #13670, opened 2021-06-21
Check if exist follower when request snapshot from followers

qian0817:

Check if exist follower when request snapshot from followers.

If there is only the master node, then the log file will be filled with `Failed to request snapshot info from followers No recent snapshot found from followers`  when request snapshot from followers.

no user facing changes.


ggezer (approved):

ggezer:
Will this now spin rather than printing? 
Should we put a sleep here?

qian0817:
> Will this now spin rather than printing?
> Should we put a sleep here?

Now when the checkpoint threshold is exceeded, the log will be printed when a raft log is applied.

ggezer:
I meant to ask how frequently this new `return` statement will be hit?

qian0817:
Almost all entries will be hit this statement if there is no follower.

ggezer:
@qian0817 I'm trying to understand  if there's a busy-loop here. If yes, then putting a sleep here will reduce CPU pressure even though count remains the same.

qian0817:
@ggezer It's strictly not a loop, but I find that adding sleep will reduce CPU pressure. I have already added the statement.

core/server/common/src/main/java/alluxio/master/journal/raft/SnapshotReplicationManager.java
```diff
@@ -449,6 +450,12 @@ private void requestSnapshotFromFollowers() {
                   .newBuilder()
                   .setSnapshotInfoRequest(GetSnapshotInfoRequest.getDefaultInstance())
                   .build()))));
+      // no follower in the cluster
+      if (jobs.isEmpty()) {
+        CommonUtils.sleepMs(Constants.SECOND_MS);
```
ggezer, line 455: Let's make this a minute.

qian0817:
solved by #13456 ,close it.

PR #13671, opened 2021-06-21
Add a metrics to track master RPC tput

apc999:

This PR adds a new metric key to track master-side RPCs / sec

As requested from users, it is useful to know how busy the master is serving RPCs

Yes


yuzhu (commented):
My only concern is that we are adding another metric to the common path. We already have individual rpc statistics, can this be computed by summing the statistics for each RPC type?

maobaolong:
@yuzhu Thank you for the summing the statistics suggestion, but It cannot be as same as this common rpc metrics, because we are not sure all of the individual rpc are computed to the aggregate result, from this common rpc path, it is easy to record the rpc throughput metrics.

maobaolong:
Have an offline discussion with @apc999 , I'm ok for change the metrics implementation, as long as we supply the total rpc related metrics.

ggezer:
Let's make sure there is no impact to meta operations before checking this in.

maobaolong:
@ggezer @apc999 Is there any update about meta operations impact? Hope to see this metrics.

core/server/common/src/main/java/alluxio/RpcUtils.java
```diff
@@ -106,6 +108,7 @@ private RpcUtils() {} // prevent instantiation
   public static <T> T callAndReturn(Logger logger, RpcCallableThrowsIOException<T> callable,
       String methodName, boolean failureOk, String description, Object... args)
       throws StatusException {
+    Metrics.MASTER_RPC_OPS_THROUGHPUT.mark(1);
     // avoid string format for better performance if debug is off
     String debugDesc = logger.isDebugEnabled() ? String.format(description, args) : null;
     try (Timer.Context ctx = MetricsSystem.timer(getQualifiedMetricName(methodName)).time()) {
```
maobaolong, line 114: Could we add a new reload method `timer` can accept more than one `MetricsName`? If so, it would reduce the metric related code in the main logic.

One possible usage like

```java
try (Timer.Context ctx = MetricsSystem.timer(getQualifiedMetricName(methodName), "rpcCall").time()) {
...
}
```

If we want to count the total `InProgress` call, we may want to use 

```java
MetricsSystem.counter(getQualifiedInProgressMetricName(methodName), "rpcCallInProgress").inc();
```

apc999: I updated the PR similar to what you proposed.
PTAL

apc999:
@yuzhu  summarizing the metrics of total RPCs from individual RPC types is hard: these are `timer` with statistics like p50, p75, hard to aggregate to a new and total RPC metrics

@ggezer I created a new timer context to feed two timers a time. Let me know if you feel better in this way.

I feel we need a revamp of metrics counting after this PR. I will discuss with you two offline.

maobaolong:
LGTM, That's what I want. Thank you

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -840,6 +841,40 @@ public static void resetCountersAndGauges() {
     }
   }
 
+  public static class MultiTimerContext implements AutoCloseable {
```
maobaolong, line 844: As the name of this class, would you like to support [0, n] timers?

```java
    private final Timer[] mTimers;
    public MultiTimerContext(Timer... timers) {
      for (Timer timer : timers) {
        Preconditions.checkNotNull(timer, "timer");
      }
      mTimers = timers;
      mStartTime = System.nanoTime();
    }

    public long stop() {
      final long elapsed = System.nanoTime() - mStartTime;
      for (Timer timer : mTimers) {
        timer.update(elapsed, TimeUnit.NANOSECONDS);
      }
      return elapsed;
    }
```

maobaolong (commented):
Hi @apc999 I have a minor suggest inline.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
apc999: good call

yuzhu (approved):

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -23,6 +23,7 @@
 import alluxio.util.network.NetworkAddressUtils;
 
 import com.codahale.metrics.CachedGauge;
+import com.codahale.metrics.Clock;
```
maobaolong, line 26: Remove this import

apc999:
@maobaolong @ggezer  FYI:
@dbw9580  is helping using his RPC stressbench tool to verify if this PR introduces performance overhead.
Bowen, please post it here the results from RPC stressbench.

As discussed offline with @ggezer , if this PR is not introducing performance degradation, we will merge it

dbw9580:
I've got some preliminary results.

I use the `StressMasterBench` to create 10000 test files at the master, and measure the throughput of doing the `ListDir` operation in a cluster of 1 master and 3 workers. Depending on how many simulated clients and how many threads each worker is using, the stress put on the master varies. 

The command to start the benchmark looks like:

```console
bin/alluxio runClass alluxio.stress.cli.StressMasterBench --operation ListDir --warmup 5s --duration 30s
 --clients 2 --threads 2 --fixed-count 10000 --cluster --cluster-limit 3 --target-throughput 100000
 --base 'alluxio://AlluxioSandbox-bowen-masters-1:19998/stress-master-base'
```

The result is below, where `c` is no. clients, and `t` is no. threads.

Throughput of `ListDir`:

| parameters | c=1, t=1 | c=2, t=2 | c=4, t=4 | c=1, t=10 | c=10, t=10| c=10, t=20| c=10, t=50 |
|-----------|---------|---------|---------|-----------|---------|---------|---------|
| baseline | 17.739098 | 30.405294 | 35.441532 | 35.800377 | 35.580093 | 36.06496 | 38.521786|
| with metric added | 18.597622 | 32.468178 | 37.721527 | 37.56974 | 37.519115 | 38.609425 |  39.19275 |

In every test the master with the metric added outperformed the baseline.  I'm pretty sure I didn't mistake one for another.

Really have no idea why it's like this... I thought it's due to [a bug](https://github.com/Alluxio/alluxio/pull/13958) in the `StressMasterBench`, but even after I applied the fix for the baseline and this branch, and retested, the results remained the same.

More detailed results can be found [here](https://drive.google.com/drive/folders/1wuEfDzbRTcSOM-2EDHLaUKv08H2Dc1Ye?usp=sharing).

dbw9580:
BTW, the metric is working as expected, although I think the time unit should be seconds instead of milliseconds.

```json
$ curl -G http://AlluxioSandbox-bowen-masters-1:19999/metrics/json/ | jq '.timers["Master.TotalRpcs"]'
{
  "count": 217992,
  "max": 0.25309699999999996,
  "mean": 0.08051577074260588,
  "min": 0.002881,
  "p50": 0.022174,
  "p75": 0.14609999999999998,
  "p95": 0.162221,
  "p98": 0.16825199999999998,
  "p99": 0.17374799999999999,
  "p999": 0.25309699999999996,
  "stddev": 0.06746854157123174,
  "m15_rate": 9.854821419151405,
  "m1_rate": 6.241497317990331,
  "m5_rate": 6.325676776548836,
  "mean_rate": 53.71243696103727,
  "duration_units": "milliseconds",
  "rate_units": "calls/second"
}
```

dbw9580:
More results in this spreadsheet: https://docs.google.com/spreadsheets/d/1ThgFXkYS_v7tryQkDudApeuW2hCwsC7D/edit?usp=sharing&ouid=108844305971006725786&rtpof=true&sd=true

I added this commit 29437ad3c53f8b648ee481d3a9cedb3229768a5a to simulate the old behaviour that calls a clock's `getStep` method to get the current time, instead of calling `system.nanotime()` directly. That's column `withMetricEnabledUseClock` in the spreadsheet.

The result is contrary to what I found yesterday: with this PR merged, the RPC throughput is somewhat reduced. Though the command I used is changed:

```console
bin/alluxio runClass alluxio.stress.cli.StressMasterBench \
--operation ListDir --fixed-count 10000 \
--warmup 5s --duration 30s \
--cluster --cluster-limit 3 \
--target-throughput 100 \
--clients 8 --threads 64 \
--java-opt " -Xmx8g" \
--conf alluxio.user.file.writetype.default=MUST_CACHE \
--conf alluxio.user.block.master.client.pool.size.max=512 \
--conf alluxio.user.file.master.client.pool.size.max=512 \
--base 'alluxio://AlluxioSandboxbowen3-masters-1:19998/stress-master-base' >"listdir/run$i.json"
```

Notably no. clients is 8 and no. threads is 64. Also added a bunch of confs from a wrapper script that Overmind uses.

ggezer:
> More results in this spreadsheet: https://docs.google.com/spreadsheets/d/1ThgFXkYS_v7tryQkDudApeuW2hCwsC7D/edit?usp=sharing&ouid=108844305971006725786&rtpof=true&sd=true
> 
> I added this commit [29437ad](https://github.com/Alluxio/alluxio/commit/29437ad3c53f8b648ee481d3a9cedb3229768a5a) to simulate the old behaviour that calls a clock's `getStep` method to get the current time, instead of calling `system.nanotime()` directly. That's column `withMetricEnabledUseClock` in the spreadsheet.
> 
> The result is contrary to what I found yesterday: with this PR merged, the RPC throughput is somewhat reduced. Though the command I used is changed:
> 
> ```
> bin/alluxio runClass alluxio.stress.cli.StressMasterBench \
> --operation ListDir --fixed-count 10000 \
> --warmup 5s --duration 30s \
> --cluster --cluster-limit 3 \
> --target-throughput 100 \
> --clients 8 --threads 64 \
> --java-opt " -Xmx8g" \
> --conf alluxio.user.file.writetype.default=MUST_CACHE \
> --conf alluxio.user.block.master.client.pool.size.max=512 \
> --conf alluxio.user.file.master.client.pool.size.max=512 \
> --base 'alluxio://AlluxioSandboxbowen3-masters-1:19998/stress-master-base' >"listdir/run$i.json"
> ```
> 
> Notably no. clients is 8 and no. threads is 64. Also added a bunch of confs from a wrapper script that Overmind uses.

Thanks @dbw9580. Could you please retry this after-before test with this same command and share how much throughput change you're seeing?

dbw9580:
> Thanks @dbw9580. Could you please retry this after-before test with this same command and share how much throughput change you're seeing?

@ggezer 
The results from running this command with and without this PR merged are summarized in [this spreadsheet](https://docs.google.com/spreadsheets/d/1ThgFXkYS_v7tryQkDudApeuW2hCwsC7D/edit#gid=1751699954), and the raw output JSON files generated by `StressMasterBench` are in [this folder](https://drive.google.com/drive/folders/1wuEfDzbRTcSOM-2EDHLaUKv08H2Dc1Ye).

`GetFileStatus` and `GetBlockLocation` saw -2.5% and -3.8% performance impact, respectively, compared to the baseline which is [this commit](https://github.com/Alluxio/alluxio/commit/2c792f987193cfa89d4c3798c898e942b7b85185). For `ListDir` the performance penalty is up to -13.8%. To be extra sure about this, I ran `ListDir` tests twice in a new cluster and got an identical result.

I also added [a new commit](https://github.com/dbw9580/alluxio/commit/29437ad3c53f8b648ee481d3a9cedb3229768a5a) that replaced the direct usage of `System.nanotime()` with a call to `clock.getTick()`, which is what it used to be without this PR. To my surprise, doing so seems to improve the performance a bit (compared to using `System.nanotime()` directly).

dbw9580:
Did some micro benchmarks with [this commit](https://github.com/dbw9580/alluxio/commit/9309961f7740d003ac7ba377eb5a9ec2e70027e0).

I find that with this PR merged, on my machine it adds a constant ~200ns on each invocation to `callAndReturn`, regardless of how long the RPC call actually takes. For example, calling `callAndReturn` with a no-op dummy RPC call takes ~500ns on baseline, and ~700ns with this pr merged. Increase the delay, the time goes up but the difference remains the same. In other words, the longer the rpc call takes, the more negligible impact this PR would have on performance.

dbw9580:
> Did some micro benchmarks with [this commit](https://github.com/dbw9580/alluxio/commit/9309961f7740d003ac7ba377eb5a9ec2e70027e0).
> 
> I find that with this PR merged, on my machine it adds a constant ~200ns on each invocation to `callAndReturn`, regardless of how long the RPC call actually takes. For example, calling `callAndReturn` with a no-op dummy RPC call takes ~500ns on baseline, and ~700ns with this pr merged. Increase the delay, the time goes up but the difference remains the same. In other words, the longer the rpc call takes, the more negligible impact this PR would have on performance.

I've got more reliable results with the microbenchmark tools in https://github.com/Alluxio/alluxio/pull/14003.
Details are in this summary spreadsheet: https://docs.google.com/spreadsheets/d/1ThgFXkYS_v7tryQkDudApeuW2hCwsC7D/edit#gid=1377711729

Conclusions:
1. the PR introduce a constant ~200ns additional delay.
2. the relative performance impact dwindles as the rpc takes longer time.
    when the rpc takes <1us to finish, the relative performance impact is -10.5%.
    when the rpc takes 9us to finish, the relative performance impact is -2%.

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -840,6 +840,44 @@ public static void resetCountersAndGauges() {
...
+      mStartTime = System.nanoTime();
+    }
+
+    /**
+     * Updates the timer with the difference between current and start time. Call to this method
+     * will not reset the start time. Multiple calls result in multiple updates.
+     *
+     * @return the elapsed time in nanoseconds
+     */
+    public long stop() {
```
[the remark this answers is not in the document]
apc999: inlined `stop` to keep the code simple

core/common/src/main/java/alluxio/metrics/MetricsSystem.java
```diff
@@ -840,6 +840,44 @@ public static void resetCountersAndGauges() {
...
+      final long elapsed = System.nanoTime() - mStartTime;
+      for (Timer timer : mTimers) {
+        timer.update(elapsed, TimeUnit.NANOSECONDS);
```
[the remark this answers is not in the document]
apc999: no we can not `close` them -- `Timer`s don't have `close` method

apc999:
@ggezer  PTAL

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -368,6 +368,13 @@ public MetricKey build() {
           .setDescription("Total number of unique blocks in Alluxio")
           .setMetricType(MetricType.GAUGE)
           .build();
+  public static final MetricKey MASTER_TOTAL_RPCS =
+      new Builder("Master.TotalRpcs")
+          .setDescription("Throughput of master UFS operations. This metrics indicates how busy the"
```
ggezer, line 373:
```suggestion
          .setDescription("Throughput of master RPC calls. This metrics indicates how busy the"
```

ggezer (approved):

apc999:

maobaolong:
@apc999 Wow, I wait for this great metric for a long time, very excited it merged. Thanks @apc999 and @ggezer @dbw9580 @yuzhu

PR #13678, opened 2021-06-22
Add a metrics to monitor grpc job pending queue

jayzhenghan:

Add a metrics to monitor grpc job pending queue

You can observe the pressure of the master through this metrics

### DIFF
--- a/core/common/src/main/java/alluxio/metrics/MetricKey.java
+++ b/core/common/src/main/java/alluxio/metrics/MetricKey.java
@@ -368,6 +368,11 @@ public final class MetricKey implements Comparable<MetricKey> {
           .setDescription("An estimate of the blocks heap size")
           .setMetricType(MetricType.GAUGE)
           .build();
+  public static final MetricKey MASTER_RPC_QUEUE_LENGTH =
+      new Builder("Master.RpcQueueLength")
+          .setDescription("Length of the master rpc queue")
+          .setMetricType(MetricType.GAUGE)
+          .build();
 
   // Backup Restore
   public static final MetricKey MASTER_LAST_BACKUP_ENTRIES_COUNT =
--- a/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
+++ b/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
@@ -27,6 +27,7 @@ import alluxio.master.journal.JournalMasterClientServiceHandler;
 import alluxio.master.journal.JournalSystem;
 import alluxio.master.journal.JournalUtils;
 import alluxio.master.journal.raft.RaftJournalSystem;
+import alluxio.metrics.MetricKey;
 import alluxio.metrics.MetricsSystem;
 import alluxio.metrics.sink.MetricsServlet;
 import alluxio.metrics.sink.PrometheusMetricsServlet;
@@ -50,7 +51,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
 import java.net.URI;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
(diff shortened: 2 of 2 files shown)


maobaolong:
@jayzhenghan Thanks for this useful metrics, before reviewing, would you please fix the style issues

```
/usr/src/alluxio/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java:321: error: Line is longer than 100 characters (found 137).
/usr/src/alluxio/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java:321:103: error: 'typecast' is not followed by whitespace.
```

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
```diff
@@ -317,6 +318,9 @@ protected void startServingRPCServer() {
           TimeUnit.MILLISECONDS);
 
       serverBuilder.executor(mRPCExecutor);
+      MetricsSystem.registerGaugeIfAbsent(MetricKey.MASTER_RPC_PENDING_JOBS.getName(),
```
maobaolong, line 321: As we create mRPCExecutor as a concrete class `ForkJoinPool`, and using the `getQueuedTaskCount` method which only belong to the `ForkJoinPool`, so my suggestion is define mRPCExecutor as the ForkJoinPool.

maobaolong (changes requested):
Left a minor comment inline.

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
jayzhenghan: > As we create mRPCExecutor as a concrete class `ForkJoinPool`, and using the `getQueuedTaskCount` method which only belong to the `ForkJoinPool`, so my suggestion is define mRPCExecutor as the ForkJoinPool.

I agree with you

maobaolong: Please remove the unused `import java.util.concurrent.ExecutorService;` also

maobaolong:
@jayzhenghan LGTM, @yuzhu Could you please take another look?

yuzhu (approved):

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
ggezer: JFYI; `ForkJoinPool` might go away soon.

apc999: > JFYI; `ForkJoinPool` might go away soon.

shall we leave a TODO or doc when defining `mRPCExecutor` ?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -368,6 +368,11 @@ public MetricKey build() {
           .setDescription("An estimate of the blocks heap size")
           .setMetricType(MetricType.GAUGE)
           .build();
+  public static final MetricKey MASTER_RPC_PENDING_JOBS =
+      new Builder("Master.RpcPendingJobs")
```
apc999, line 372: How about `Master.RpcQueueLength`? is it accurate?

jayzhenghan: RpcCallQueueLength？

maobaolong: `Master.RpcQueueLength` is better. RpcCallQueueLength is from hdfs, and there is call in the context of hdfs, for Alluxio, just call it `RpcQueueLength`.

jayzhenghan: ok👌

core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java
jayzhenghan: @apc999 How should todo be described

apc999 (approved):

apc999:

PR #13687, opened 2021-06-23
Support cache Not exist status

maobaolong:

If a file or dir is not exist, `getStatus` will throw FileDoesNotExistException, and metadataCache will never cache this result, client will try to call rpc server again next time.

Cache the special URIStatus `NOT_FOUND_STATUS` while met FileDoesNotExistException, when cache hit, re-throw the exception again.

NO.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
@@ -52,7 +52,6 @@ public class MetadataCachingBaseFileSystem extends BaseFileSystem {
   private final MetadataCache mMetadataCache;
   private final ExecutorService mAccessTimeUpdater;
   private final boolean mDisableUpdateFileAccessTime;
-  private final boolean mNotFoundResultCacheEnabled;
 
   /**
    * @param context the fs context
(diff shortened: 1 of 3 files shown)


core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
```diff
@@ -50,6 +51,7 @@
   private final MetadataCache mMetadataCache;
   private final ExecutorService mAccessTimeUpdater;
   private final boolean mDisableUpdateFileAccessTime;
+  private static final URIStatus NOT_FOUND_STATUS = new URIStatus(new FileInfo());
```
apc999, line 54: nit: move this `NOT_FOUND_STATUS` up with other static final members?

apc999 (changes requested):
@maobaolong  on one hand, I can see adding negative result caching in this PR can improve performance for certain workloads (e.g., multiple threads querying the same negative results); On the other hand, this may create consistency issue. For example, if some one query a non-existing path and get a negative result, then create a path, then next time querying this path will still yield negative result. Note, this pattern is very often for Spark or MR jobs when committing data.

What do you think to prevent this situation?

core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
maobaolong: Done.

maobaolong:
@apc999 Thanks for let me know the two hand while bring this PR, I add a new propertykey to let user chose whether enable the FileDoesNotExistException cache. PTAL.

apc999:
hi @maobaolong  in general we want to avoid introducing more new property keys unless it is necessary.
In your case here, I view the potential inconsistency introduced by caching negative result similar to existing potential inconsistency by caching positive result (but got deleted / renamed on server side).  So basically, the metadata caching is eventual consistent with certain time bound. So we shall be fine without adding a property.

Can you also add a test?

apc999:
Add a Unit Test, @maobaolong ?

maobaolong:
@apc999 Sure, will add a ut.

maobaolong:
@apc999 A new UT added, PTAL.

apc999 (approved):

apc999:

PR #13688, opened 2021-06-23
Add scheme and authority if authority is missing in PathTranslator

yuzhu:

The path translator needs to return a full URI path because the client might not be able to resolve relative path. 

Fixing an issue where the catalog service client can't locate the table path. 

No

### DIFF
--- a/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
@@ -93,8 +93,8 @@ public class PathTranslator {
       String difference = PathUtils.subtractPaths(ufsUri.getPath(),
           longestPrefix.getValue().getPath());
       AlluxioURI mappedUri = longestPrefix.getKey().join(difference);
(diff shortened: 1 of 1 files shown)


dbw9580:
Reproduced this bug. With commit 499d2b5686f9113c25e0ce665bf5c65c7eef3a7d, the returned path does not include a scheme:

```
2021-06-26 21:22:20,988 INFO  PathTranslator - ufsPath: file:/home/dbw/testdb/person, alluxioPath: /catalog/person/tables/person/hive
```

It should be:

```
2021-06-26 22:11:30,451 INFO  PathTranslator - ufsPath: file:/home/dbw/testdb/person, alluxioPath: alluxio://localhost:19998/catalog/person/tables/person/hive
```

Turns out it was the exact match optimization that escaped scheme and authority check: 

https://github.com/Alluxio/alluxio/blob/499d2b5686f9113c25e0ce665bf5c65c7eef3a7d/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java#L65-L68

I'll push a commit to fix it soon.

PR #13690, opened 2021-06-23
Add journal checkpoint warnings on UI

ZacBlanco:

Adding warnings to the UI overview page when the journal has not checkpointed in some time after reaching the journal entry checkpoint period. Also, add the two checkpoint metrics (last checkpoint time and entries since checkpoint) to the metrics tab.

New overview page warning:

image

Metrics page addition:

image

This is required to help alert users to potential issues with their system if the journal has not checkpointed when it is supposed to. 

- Addition of a hidden property key to change the time period which causes the warning to show up on the overview page. It is marked as hidden because no one _should_ need to configure this value, but it may be desirable. It is also useful for testing.
- Web UI additions (see screenshots). New section on the metrics page and new potential warning in the overview.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
@@ -345,10 +345,10 @@ public final class AlluxioMasterRestServiceHandler {
               + "prevent the master from serving requests while checkpointing.";
           response.setJournalCheckpointTimeWarning(String.format("Journal has not checkpointed in "
               + "a timely manner since passing the checkpoint threshold (%d/%d). Last checkpoint:"
-              + " %s. t is recommended to use the fsadmin tool to force a checkpoint. " + advice,
+              + " %s. %s",
               entriesSinceCkpt,
               ServerConfiguration.getLong(PropertyKey.MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES),
-              time));
+              time, advice));
         }
       }
 
--- a/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
@@ -340,9 +340,12 @@ public final class AlluxioMasterRestServiceHandler {
           String time = lastCkptTime > 0 ? ZonedDateTime
               .ofInstant(Instant.ofEpochMilli(lastCkptTime), ZoneOffset.UTC)
               .format(DateTimeFormatter.ISO_INSTANT) : "N/A";
+          String advice = ConfigurationUtils.isHaMode(ServerConfiguration.global()) ? ""
+              : "It is recommended to use the fsadmin tool to checkpoint the journal. This will "
+              + "prevent the master from serving requests while checkpointing.";
           response.setJournalCheckpointTimeWarning(String.format("Journal has not checkpointed in "
               + "a timely manner since passing the checkpoint threshold (%d/%d). Last checkpoint:"
-              + " %s. It is recommended to use the fsadmin tool to force a checkpoint.",
+              + " %s. t is recommended to use the fsadmin tool to force a checkpoint. " + advice,
               entriesSinceCkpt,
               ServerConfiguration.getLong(PropertyKey.MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES),
               time));
--- a/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
+++ b/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
@@ -1082,20 +1082,26 @@ public final class AlluxioMasterRestServiceHandler {
         response.setJournalDiskMetrics(Collections.emptyList());
       }
 
-      long lastCheckpointTime =
-          (Long) gauges.get(MetricKey.MASTER_JOURNAL_LAST_CHECKPOINT_TIME.getName()).getValue();
-      long entriesSinceCheckpoint =
-          (Long) gauges.get(MetricKey.MASTER_JOURNAL_ENTRIES_SINCE_CHECKPOINT.getName()).getValue();
-      response.setJournalEntriesSinceCheckpoint(entriesSinceCheckpoint);
-      String time;
-      if (lastCheckpointTime > 0) {
-        time = ZonedDateTime
-            .ofInstant(Instant.ofEpochMilli(lastCheckpointTime), ZoneOffset.UTC)
-            .format(DateTimeFormatter.ISO_INSTANT);
-      } else {
-        time = "N/A";
+      Gauge lastCheckpointTimeGauge =
+          gauges.get(MetricKey.MASTER_JOURNAL_LAST_CHECKPOINT_TIME.getName());
+      Gauge entriesSinceCheckpointGauge =
+          gauges.get(MetricKey.MASTER_JOURNAL_ENTRIES_SINCE_CHECKPOINT.getName());
+
+      if (entriesSinceCheckpointGauge != null) {
+        response.setJournalEntriesSinceCheckpoint((long) entriesSinceCheckpointGauge.getValue());
       }
-      response.setJournalLastCheckpointTime(time);
+      if (lastCheckpointTimeGauge != null) {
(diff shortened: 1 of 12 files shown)


ZacBlanco:
FYI @ns1123

webui/master/src/containers/pages/Overview/Overview.tsx
```diff
@@ -134,6 +135,21 @@ export class OverviewPresenter extends React.Component<AllProps> {
     );
   }
 
+  private renderJournalCheckpointWarning(warning: string, className: string): JSX.Element | null {
+    if (!warning || warning == "") {
+      return null;
+    }
+
+    return (
+      <tr key="0">
```
tieujason330, line 144: make the key something more unique - '0' might collide with `renderJournalDiskWarnings()` table rows

webui/master/src/containers/pages/Metrics/Metrics.tsx
```diff
@@ -318,6 +318,21 @@ export class MetricsPresenter extends React.Component<AllProps> {
...
         </div>
+        <div className="col-12">
+          <h5>Alluxio Master Journal Checkpoint Status</h5>
+          <Table hover={true}>
+            <tbody>
+              <tr>
+                <th>Last Checkpoint Time</th>
+                <th>Journal Entries Since Checkpoint</th>
+              </tr>
+              <tr key="0">
```
tieujason330, line 329: you can remove `key` here. it's usually needed for <tr> elements that get rendered when iterating through a list

tieujason330 (changes requested):

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -319,6 +323,32 @@ public Response getWebUIOverview() {
...
+      Gauge entriesSinceGauge = MetricsSystem.METRIC_REGISTRY.getGauges()
+              .get(MetricKey.MASTER_JOURNAL_ENTRIES_SINCE_CHECKPOINT.getName());
+      Gauge lastCkPtGauge = MetricsSystem.METRIC_REGISTRY.getGauges()
+          .get(MetricKey.MASTER_JOURNAL_LAST_CHECKPOINT_TIME.getName());
+
+      if (entriesSinceGauge != null && lastCkPtGauge != null) {
+        long entriesSinceCkpt = (Long) entriesSinceGauge.getValue();
+        long lastCkptTime = (Long) lastCkPtGauge.getValue();
+        long timeSinceCkpt = System.currentTimeMillis() - lastCkptTime;
+        boolean over72Hrs = timeSinceCkpt > ServerConfiguration.getMs(
```
ggezer, line 335: nit: overThreshold

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -319,6 +323,32 @@ public Response getWebUIOverview() {
...
+            PropertyKey.MASTER_WEB_JOURNAL_CHECKPOINT_WARNING_THRESHOLD_TIME);
+        boolean passedThreshold = entriesSinceCkpt > ServerConfiguration
+            .getLong(PropertyKey.MASTER_JOURNAL_CHECKPOINT_PERIOD_ENTRIES);
+        if (lastCkptTime > 0 && passedThreshold && over72Hrs) {
```
ggezer, line 339: why lastCkptTime > 0 ?
How about the very first checkpoint?

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -319,6 +323,32 @@ public Response getWebUIOverview() {
...
+          String time = ZonedDateTime
+              .ofInstant(Instant.ofEpochMilli(lastCkptTime), ZoneOffset.UTC)
+              .format(DateTimeFormatter.ISO_INSTANT);
+          response.setJournalCheckpointTimeWarning(String.format("Journal has not checkpointed in "
```
ggezer, line 343: Will this be always printed for a single-master cluster?
Do we have checkpoint capability from webUI? I believe we should at least direct users to an appropriate action.

ZacBlanco: On a single-master cluster, a user can remove this message by forcing a checkpoint with `alluxio fsadmin journal checkpoint`. We can direct users to run this command to force a checkpoint?

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
ZacBlanco: the `lastCkptTime` is an epoch timestamp. Basically 0 represents the Jan 1st 1970 date.

The metric is not updated when rebooting a master. The gauge is 0 until the first checkpoint is taken during the life of the process. Since we don't know when the last checkpoint was after a master is rebooted (it's not stored afaik?). I guess we can still issue the warning if a checkpoint hasn't been taken yet, I'll just need to adjust the error message to indicate no checkpoint taken since reboot. Maybe just with `N/A`?

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
ggezer: Right. But do we want to?

tieujason330 (approved):

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
madanadit: I would replace "It is recommended to use the fsadmin tool to force a checkpoint." with something like "No action is required with multiple masters. If using a single master, use fsadmin to force a checkpoint. User activity will be blocked."

ggezer (approved):

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
ZacBlanco: I think that message is good. Just one comment: 

> No action is required with multiple masters. ...

My only concern is that..what's the point in showing the warning if we don't want the user to do anything about it?

madanadit: Good point. Can you hide the action if you detect multiple masters from the config?

ZacBlanco: I changed the advice displayed based on the config: https://github.com/Alluxio/alluxio/pull/13690/commits/a68f75c86c5b9a71509a578c0af9ff525c2bc453#diff-78fd86b461c2be256edc9a1da53a384a8d7ac413537c95e25891b32f33b90d2eL342

madanadit: thx

core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java
```diff
@@ -319,6 +323,35 @@ public Response getWebUIOverview() {
...
+        if (passedThreshold && overThreshold) {
+          String time = lastCkptTime > 0 ? ZonedDateTime
+              .ofInstant(Instant.ofEpochMilli(lastCkptTime), ZoneOffset.UTC)
+              .format(DateTimeFormatter.ISO_INSTANT) : "N/A";
+          String advice = ConfigurationUtils.isHaMode(ServerConfiguration.global()) ? ""
+              : "It is recommended to use the fsadmin tool to checkpoint the journal. This will "
+              + "prevent the master from serving requests while checkpointing.";
+          response.setJournalCheckpointTimeWarning(String.format("Journal has not checkpointed in "
+              + "a timely manner since passing the checkpoint threshold (%d/%d). Last checkpoint:"
+              + " %s. t is recommended to use the fsadmin tool to force a checkpoint. " + advice,
```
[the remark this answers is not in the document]
ZacBlanco: updated, ty

ZacBlanco:
@ggezer could you merge this?

ggezer:

PR #13707, opened 2021-06-26
Add a name suffix to distinguish AsyncJournalWriter threads

jiacheliu3:

There can be multiple AsyncJournalWriter threads, add a thread name suffix to distinguish them.

Makes jstack debugging easier

NA

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -448,7 +448,6 @@ public class RaftJournalSystem extends AbstractJournalSystem {
 
     Preconditions.checkState(mRaftJournalWriter == null);
     mRaftJournalWriter = new RaftJournalWriter(nextSN, client);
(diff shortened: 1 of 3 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -448,6 +448,7 @@ public synchronized void gainPrimacy() {
 
     Preconditions.checkState(mRaftJournalWriter == null);
     mRaftJournalWriter = new RaftJournalWriter(nextSN, client);
+    // TODO(jiacheng): pass sth here?
```
jiacheliu3, line 451: @ggezer how many `AsyncJournalWriter` threads are there for embedded journal? Is the `LocalFirstRaftClient` a suitable distinguisher?

ggezer: There's only one async-writer. So should be 1 only.

ggezer (approved):

jiacheliu3:

PR #13708, opened 2021-06-26
Fix PathTranslator to compare path prefixes by path components

dbw9580:
Fix a bug introduced in #13663. See https://github.com/Alluxio/alluxio/pull/13688#issuecomment-869016271

The exact match optimization was not properly handled to return a URI with scheme and authority. This has been fixed and tests added.

### DIFF
--- a/table/server/common/pom.xml
+++ b/table/server/common/pom.xml
@@ -56,6 +56,8 @@
       <version>${project.version}</version>
       <scope>compile</scope>
     </dependency>
+
+    <!-- Internal test dependencies -->
     <dependency>
       <groupId>org.alluxio</groupId>
       <artifactId>alluxio-core-common</artifactId>
--- a/table/server/common/pom.xml
+++ b/table/server/common/pom.xml
@@ -56,5 +56,12 @@
       <version>${project.version}</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-core-common</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
@@ -17,7 +17,6 @@ import alluxio.exception.InvalidPathException;
 import alluxio.util.ConfigurationUtils;
 import alluxio.util.io.PathUtils;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
 import org.slf4j.Logger;
@@ -33,8 +32,12 @@ public class PathTranslator {
 
   private final BiMap<AlluxioURI, AlluxioURI> mPathMap;
 
-  @VisibleForTesting
-  static String sSchemeAuthorityPrefix = null;
+  private static final String sSchemeAuthorityPrefix;
+
+  static {
+    sSchemeAuthorityPrefix =
+        ConfigurationUtils.getSchemeAuthority(ServerConfiguration.global());
+  }
 
   /**
    * Construct a path translator.
@@ -110,10 +113,6 @@ public class PathTranslator {
 
   private static AlluxioURI checkAndAddSchemeAuthority(AlluxioURI input) {
     if (!input.hasScheme()) {
-      if (sSchemeAuthorityPrefix == null) {
-        sSchemeAuthorityPrefix =
-            ConfigurationUtils.getSchemeAuthority(ServerConfiguration.global());
-      }
       AlluxioURI baseUri = new AlluxioURI(sSchemeAuthorityPrefix);
       return new AlluxioURI(baseUri, input.getPath(), false);
     }
--- a/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
+++ b/table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
@@ -17,6 +17,7 @@ import alluxio.exception.InvalidPathException;
 import alluxio.util.ConfigurationUtils;
 import alluxio.util.io.PathUtils;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
 import org.slf4j.Logger;
@@ -64,7 +65,12 @@ public class PathTranslator {
     AlluxioURI ufsUri = new AlluxioURI(ufsPath);
     // first look for an exact match
     if (mPathMap.inverse().containsKey(ufsUri)) {
-      return mPathMap.inverse().get(ufsUri).toString() + suffix;
+      AlluxioURI match = mPathMap.inverse().get(ufsUri);
+      if (match.equals(ufsUri)) {
+        // bypassed UFS path, return as is
+        return ufsPath;
+      }
+      return checkAndAddSchemeAuthority(mPathMap.inverse().get(ufsUri)) + suffix;
     }
     // otherwise match by longest prefix
     BiMap.Entry<AlluxioURI, AlluxioURI> longestPrefix = null;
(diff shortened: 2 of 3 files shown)


table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
```diff
@@ -102,4 +103,87 @@ public void trailingSeparator() throws Exception {
     assertEquals(alluxioPath, mTranslator.toAlluxioPath(ufsPath));
     assertEquals(alluxioPath + "/", mTranslator.toAlluxioPath(ufsPath + "/"));
   }
+
+  @Test
+  public void prefixBoundaryWithinPathComponent() throws Exception {
+    mTranslator.addMapping("alluxio:///table_a", "ufs://a/table1");
+    mTranslator.addMapping("alluxio:///table_b", "ufs://a/table11");
+    assertEquals("alluxio:///table_b/part1",
```
apc999, line 111: let's assert `ufs://a/table1/part1` too here?

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -66,23 +61,60 @@ public PathTranslator addMapping(String alluxioPath, String ufsPath) {
...
+  private AlluxioURI checkAndAddSchemeAuthority(AlluxioURI input) {
+    if (!input.hasScheme()) {
+      AlluxioURI baseUri = new AlluxioURI(getSchemeAuthority() + "/");
+      return new AlluxioURI(baseUri, input.getPath(), false);
+    }
+    return input;
+  }
+
+  @VisibleForTesting
+  String getSchemeAuthority() {
```
apc999, line 117: since there is no member variable used, and this is changed over time, can we change this method to be a `static` variable instead, to save computation on every call?

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -66,23 +61,60 @@ public PathTranslator addMapping(String alluxioPath, String ufsPath) {
...
+      String difference = PathUtils.subtractPaths(ufsUri.getPath(),
+          longestPrefix.getValue().getPath());
+      AlluxioURI mappedUri = longestPrefix.getKey().join(difference);
+      return checkAndAddSchemeAuthority(mappedUri) + suffix;
+    } catch (InvalidPathException e) {
+      throw new IOException(e);
+    }
+  }
+
+  private AlluxioURI checkAndAddSchemeAuthority(AlluxioURI input) {
```
apc999, line 108: same here,  make this method `static`? (after making `getSchemeAuthority` static)

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -66,23 +61,60 @@ public PathTranslator addMapping(String alluxioPath, String ufsPath) {
...
+    if (!input.hasScheme()) {
+      AlluxioURI baseUri = new AlluxioURI(getSchemeAuthority() + "/");
+      return new AlluxioURI(baseUri, input.getPath(), false);
```
apc999, line 111: I don't think you need to append "/" to `baseUri`, `AlluxioURI(baseUri, input.getPath(), false);` will add "/" if I recall correctly.

apc999 (changes requested):

dbw9580:
@apc999 Addressed comments, PTAL.

yuzhu (commented):
@dbw9580 did this pass the automated tests?

dbw9580:
> @dbw9580 did this pass the automated tests?

Do you mean Github actions?

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -27,7 +31,10 @@
 public class PathTranslator {
   private static final Logger LOG = LoggerFactory.getLogger(PathTranslator.class);
 
-  private BiMap<String, String> mPathMap;
+  private final BiMap<AlluxioURI, AlluxioURI> mPathMap;
+
+  @VisibleForTesting
+  static String sSchemeAuthorityPrefix = null;
```
apc999, line 37: Can we try `private static final` (making it constant) and initialized this constant to `ConfigurationUtils.getSchemeAuthority(ServerConfiguration.global());`

In test (e.g., https://github.com/Alluxio/alluxio/pull/13708/files#diff-5c564844c705a2abf294409baa2f79a7bde047243560c17ee60c824405239a5bR159),
you can follow https://docs.alluxio.io/os/user/stable/en/contributor/Code-Conventions.html#managing-global-state to inject the prefix to be `alluxio://master`

dbw9580: Done. PTAL.

table/server/common/pom.xml
```diff
@@ -56,5 +56,12 @@
       <version>${project.version}</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
apc999, line 59: Follow the convention in other POM files:
```suggestion

    <!-- Internal test dependencies -->
    <dependency>
```

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -27,7 +30,14 @@
...
 
-  private BiMap<String, String> mPathMap;
+  private final BiMap<AlluxioURI, AlluxioURI> mPathMap;
+
+  private static final String SCHEME_AUTHORITY_PREFIX;
+
+  static {
+    SCHEME_AUTHORITY_PREFIX =
+        ConfigurationUtils.getSchemeAuthority(ServerConfiguration.global());
+  }
```
apc999, line 40: why not to consolidate the code?

```suggestion
  private static final String SCHEME_AUTHORITY_PREFIX =
      ConfigurationUtils.getSchemeAuthority(ServerConfiguration.global());

```

table/server/common/pom.xml
apc999: also why is this dependency needed in this PR?

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
```diff
@@ -66,23 +67,55 @@ public PathTranslator addMapping(String alluxioPath, String ufsPath) {
...
+      AlluxioURI mappedUri = longestPrefix.getKey().join(difference);
+      return checkAndAddSchemeAuthority(mappedUri) + suffix;
+    } catch (InvalidPathException e) {
+      throw new IOException(e);
+    }
+  }
+
+  private static AlluxioURI checkAndAddSchemeAuthority(AlluxioURI input) {
+    if (!input.hasScheme()) {
+      AlluxioURI baseUri = new AlluxioURI(SCHEME_AUTHORITY_PREFIX);
```
apc999, line 116: nit: you can even define this to be a static constant

table/server/common/src/test/java/alluxio/table/common/udb/PathTranslatorTest.java
```diff
@@ -27,6 +31,13 @@
   @Rule
   public ExpectedException mException = ExpectedException.none();
 
+  @Rule
+  public ConfigurationRule mConfiguration =
+      new ConfigurationRule(
+          ImmutableMap.of(PropertyKey.MASTER_HOSTNAME, "master",
+            PropertyKey.MASTER_RPC_PORT, "19998"),
```
apc999, line 38: Let's define two string constant `MASTER_HOSTNAME` and  `MASTER_RPC_PORT`
then replace the literal in the following and also in the test cases with the constant.
This is a better practise for writing tests -- more maintainable and more flexible

```suggestion
          ImmutableMap.of(PropertyKey.MASTER_HOSTNAME,  MASTER_HOSTNAME,
            PropertyKey.MASTER_RPC_PORT, MASTER_RPC_PORT),
```
Also, please define `MASTER_RPC_PORT` to be non-default value.

apc999 (changes requested):
my final batch of comments. thanks for fixing this issue

table/server/common/pom.xml
dbw9580: `alluxio.ConfigurationRule` is in module `alluxio-core-common`. I need to add the dependency to import it.

table/server/common/src/main/java/alluxio/table/common/udb/PathTranslator.java
dbw9580: My understanding is that static items are initialized when the class is loaded. Can we know for sure that `ServerConfiguration.global()` will return something meaningful when `PathTranslator` is loaded?

dbw9580:
@apc999 addressed your comments in 254ba593d7e17f4d20c2d32ff2b5040ab9be35c7, please take another look. Thanks!

apc999 (approved):

apc999:
@yuzhu  can you take another look?

apc999:
@yuzhu ？

yuzhu (approved):
Lgtm

apc999:

Issue #13732, opened 2021-07-02
Add unbuffer support for impala


kuszz:
**Is your feature request related to a problem? Please describe.**
The impala need canunbuffer ability to cache file handles

**Describe the solution you'd like**
Add canUnBuffer ability on stream which BaseFileSystem use

**Urgency**
normal

PR #13733, opened 2021-07-02
Add unbuffer support for impala

kuszz:
Address #13732

Add canUnBuffer ability on stream which BaseFileSystem use

The impala need canunbuffer ability to cache file handles

User can unbuffer the some streams


core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -32,6 +32,7 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.fs.CanUnbuffer;
```
apc999, line 35: in `core/client/fs` module, we shall not introduce a dependency to hadoop which will cause dependencies issues.
Module `core/client/hdfs` is the wrapper on top of Alluxio In/out streams and make them HDFS compatible.

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -520,4 +521,13 @@ public BlockInStreamSource getSource() {
   public long getId() {
     return mId;
   }
+
+  @Override
```
apc999, line 525: Just claim it without introducing a interface
```suggestion
```

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -520,4 +521,13 @@ public BlockInStreamSource getSource() {
...
+  public void unbuffer() {
+    try {
+      closeDataReader();
+    } catch (IOException e) {
+      LOG.error("Failed to close DataReader: {}", e.getMessage());
```
apc999, line 530:
```suggestion
      LOG.error("Failed to close DataReader", e);
```

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
```diff
@@ -33,6 +33,9 @@
 import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.io.Closer;
+import org.apache.hadoop.fs.CanUnbuffer;
+import org.apache.hadoop.fs.StreamCapabilities;
+import org.apache.hadoop.util.StringUtils;
```
apc999, line 38: same here, no hadoop dependency here.

core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
```diff
@@ -39,7 +43,7 @@
  */
 @NotThreadSafe
 public class HdfsFileInputStream extends InputStream implements Seekable, PositionedReadable,
-    ByteBufferReadable {
+    ByteBufferReadable, StreamCapabilities, CanUnbuffer {
```
apc999, line 46: do we know staring from which hadoop version , these interfaces are introduced?

core/client/fs/pom.xml
```diff
@@ -92,6 +92,12 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <version>3.2.1</version>
+      <scope>compile</scope>
+    </dependency>
```
apc999, line 100: No hadoop dependency shall be added to `core/client/fs/pom.xml`
```suggestion
```

apc999 (changes requested):
Comments and questions are left

core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
maobaolong: https://issues.apache.org/jira/browse/HDFS-7694 
This ticket show that it string from 2.6.4

core/client/fs/pom.xml
maobaolong: +1

maobaolong (changes requested):
@kuszz Thanks for this feature, but please keep no hadoop dependency is it was. So think about to move the canbuffer logic to the underfs/hdfs module or find some other approach instead of involve hadoop dependency to core module.

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
kuszz: Got it, I have modified code for these issues

core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
apc999: then one problem will be: if we build Alluxio src code with Hadoop version < 2.6.4, the build will fail.
is it possible that we add `hasCapability` and `unbuffer` here, but without claiming implementing
interfaces `StreamCapabilities`, `CanUnbuffer`? will that work for Impala? e.g., does Impala checks if the InputStream extends these interfaces explicitly?

apc999 (changes requested):
one question left about building this with Hadoop version < 2.6

core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
kuszz: @apc999 Ok, I will ask impala contributor who ask for this feature if can not check it

apc999 (changes requested):
Can you add unit tests on the new method?

apc999:
ping @kuszz

kuszz:
> ping @kuszz

Sorry, I missed the github email，I will add UT as soon as possible

core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
apc999: as discussed offline, we can follow the same convention as introducing `ByteBufferReadable`

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -224,7 +224,8 @@ public void initMetricKeys() {
             long total =
                 MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_DIRECT.getName()).getCount()
                 + MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_REMOTE.getName()).getCount()
-                + MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_DOMAIN.getName()).getCount();
+                + MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_DOMAIN.getName()).getCount()
+                + MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_LOCAL.getName()).getCount();
```
apc999, line 228: this change is unrelated and already merged in by da3210a470025172dc9d49ea246909ca61305549, right?

kuszz: Yes, sorry, I merged local master and add this change by accident, should I do something for this?

apc999 (approved):

apc999:
@ggezer  any input on this PR?

core/client/hdfs/src/main/java/alluxio/hadoop/HdfsFileInputStream.java
apc999: @ggezer  in short, this PR is not introducing any new change in backward compatibility when building Alluxio.

This is because `org.apache.hadoop.fs.ByteBufferReadable` (introduced a few years back into `HdfsFileInputStream`) already breaks the build. As a result, if someone wants to build the Alluxio with north-bound HDFS client to be 2.2 or 1.x, they will already fail. Actually the two interfaces `StreamCapabilities`, `CanUnbuffer` is introduced to Hadoop earlier than `ByteBufferReadable`, so this PR is not making this situation  worse. 

Thus I think this change shall be fine, given we don't even notice this break in the last few years --- largely meaning no one attempts to build Alluxio north bound to be HDFS 1.x or 2.2. So we shall be good.

apc999:
@ggezer I addressed your qestion https://github.com/Alluxio/alluxio/pull/13733/files#r680319496. PTAL

ggezer (approved):

apc999:

Issue #13734, opened 2021-07-02
Add master_ufs_metadata_access_count metric


kuszz:
Is your feature request related to a problem? Please describe.
We want to add a metric master_ufs_metadata_access_count and now we can not get this info

Describe the solution you'd like
Add a gauge metric in InodeSyncStream, increase it when loadMetadata

Urgency
normal

PR #13735, opened 2021-07-02
Add master_ufs_metadata_access_count metric

kuszz:
#13734

Add a counter metric in InodeSyncStream, increase it when loadMetadata. And initialize it in DefaultFileSystemMaster.

We want to add a metric master_ufs_metadata_access_count and now we can not get ufs metadata access num info

Users can get a new metric

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
@@ -153,7 +153,7 @@ public class BlockMasterClient extends AbstractMasterClient {
    * @return a flattened and deduplicated list
    */
   @VisibleForTesting
-  List<LocationBlockIdListEntry> convertBlockListMapToProto(
+  public List<LocationBlockIdListEntry> convertBlockListMapToProto(
           Map<BlockStoreLocation, List<Long>> blockListOnLocation) {
     final List<LocationBlockIdListEntry> entryList = new ArrayList<>();
 
--- a/job/server/src/main/java/alluxio/job/plan/stress/StressBenchDefinition.java
+++ b/job/server/src/main/java/alluxio/job/plan/stress/StressBenchDefinition.java
@@ -88,6 +88,7 @@ public final class StressBenchDefinition
     workerList = workerList.subList(0, clusterLimit);
 
     for (WorkerInfo worker : workerList) {
+      LOG.info("Generating job for worker {}", worker.getId());
       ArrayList<String> args = new ArrayList<>(2);
       // Add the worker hostname + worker id as the unique task id for each distributed task.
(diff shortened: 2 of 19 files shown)


maobaolong (approved):
It LGTM, thanks for the contribution. @yuzhu  Please take another look.

yuzhu (approved):
LGTM, thanks

apc999 (commented):
@maobaolong @kuszz  @yuzhu  is this (https://github.com/Alluxio/alluxio/pull/13735/files#diff-885d20a474b15524a5e7f1de04d3fbe28e10bc01fd7e5630eef5e004402b9f11R756) the only place to access UFS metadata? I don't think so.

yuzhu:
@apc999 It depends on what we are trying to measure.  

If we are measuring the number of metadata accesses to the ufs, the best place to perform this count might be inside UfsStatusCache class, because that's where we actually access ufs. we can accumulate how many accesses to the ufs we do in each UfsStatusCache and then aggregate them together. 

However, if we are counting how many files we synced, then this location or `syncInodeMetadata` might be a good place to perform the count. 

Those two measurements are different because with one listStatus call to the ufs, we can sync many files.

apc999:
Can you take a look at `UnderFileSystemWithLogging`

maobaolong:
@kuszz You can reference this PR https://github.com/Alluxio/alluxio/pull/13671

kuszz:
@yuzhu If the ufsStatus==null, should we count the UFS_METADATA_ACCESS?
      UfsStatus ufsStatus = ufs.getStatus(ufsUri.toString());
      Metrics.UFS_METADATA_ACCESS.inc();  // Should put here?
      if (ufsStatus == null) {
        mAbsentCache.addSinglePath(path);
        return null;
      }
      Metrics.UFS_METADATA_ACCESS.inc();  // Or put here?
      ...
}

yuzhu:
@kuszz for your specific question, we should also increment when ufsStatus is null. 
My higher level question is what are we trying to measure? Are we measuring how many metadata accesses we do as part of metadata sync? or in all situations? 
If the first one is the goal, then measuring here is fine. If the second one is the goal, we should instrument `UnderFileSystemWithLogging`  class as suggested by @apc999

kuszz:
> @kuszz for your specific question, we should also increment when ufsStatus is null.
> My higher level question is what are we trying to measure? Are we measuring how many metadata accesses we do as part of metadata sync? or in all situations?
> If the first one is the goal, then measuring here is fine. If the second one is the goal, we should instrument `UnderFileSystemWithLogging` class as suggested by @apc999
Thanks, we want to measure the file num sync, and I have wondered after read codes, and so sorry I have hesitated how to reply so that forget to reply, very sorry...

PR #13741, opened 2021-07-04
Support regex and exclusion in UDB config file

### LINKED ISSUE #13667: Support more sophisticated table and partition bypassing options
During review of #13581, @HelloHorizon suggested it should be made more user-friendly to configure bypassed partitions:

> Different from the number of tables in the database, there can be tens of thousands of partitions in single tables. Can we also come up with an easy way for user to configure the bypass partitions, like bypass major part of partitions?

@maobaolong offered insights from real use cases:

> With the bypass table and partition feature, we can use it to bypass some older tables or partitions and mount a small range of tables or partitions to alluxio.

> It just like a rolling forward window which stand for a range of active valuable dataset(partitions), the window can move forward by a time unit, let's say it is a week, so we just want alluxio cache the current week partition, and bypass the older partitions, if some query need the older partition, it will read data from under filesystem directly rather than read it from alluxio.

1. allow use of glob, regex, etc. to match partitions when specifying which partitions to bypass.
2. introduce some kind of filter functions that are invoked on table sync and determine which tables and partitions should be bypassed.

For example:
    
```
{
  "bypass": {
    "tables": [
      {
        "table": "table1",
        "partitions": [
          "table1_part1",
          {"filter": "regex", "regex": "table1_part(2|3|4)", "on": "name"},
          {"filter": "topN", "top_n": 100, "on": "column(id)"}
        ]
      }
    ]
  }
}
```

Open to discussion.

**Urgency**

Normal

dbw9580:

TBD.

Fixes #13667.

Yes. The config file has a new syntax.


dbw9580:
@maobaolong @HelloHorizon I added a formal syntax specification in 8af2dbc, and explained the inclusion and exclusion behaviour in a57ff01: https://github.com/Alluxio/alluxio/blob/a57ff019895e1ad118946ee87cd319250acfd782/table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java#L98-L131

Please help take a look to ensure the design and implementation makes sense to you, before I proceed to add tests. Thanks!

dbw9580:
> Please help take a look to ensure the design and implementation makes sense to you, before I proceed to add tests. Thanks!

Especially I have this unresolved question here: https://github.com/Alluxio/alluxio/blob/a57ff019895e1ad118946ee87cd319250acfd782/table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java#L129-L130

I decided to throw an exception when two explicit names clash, while wanted to provide a sane default when two patterns overlap and conflict. The reasoning is:

* Two identical names present in both the include and the exclude list at the same time are obvious to spot and fix, so it's ideal that we report this immediately to the user and abort the sync process. 

* But the fact that two patterns overlap and therefore conflict with each other may be hard to realize and fix. We should provide a default -- hopefully sane and safe -- and quietly ignore the conflicting patterns. 

I found it hard to convice myself which is right: to go with the include pattern or the exclude pattern. Would appreciate your opinions.

HelloHorizon:
@dbw9580 I am confused if the user already uses the regex to filter out the tables, why we need to do the `include` and `exclude` at the same time? Can it only be `include` or `exclude` to simplify the process? I feel it will hard to use since it's easy to have some unknown overlaps for users if we provide the `include` and `exclude` at the same time.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+    UdbMountSpec.Builder builder = new UdbMountSpec.Builder();
+    for (TableEntry entry : mBypassEntry.getList().getIncludedEntries()) {
+      if (entry.isPattern()) {
+        builder.bypass().include().addPattern(entry.getPattern());
+        continue;
+      }
+      if (entry.getPartitions().isEmpty()) {
+        builder.bypass().include().addName(entry.getName());
+        continue;
+      }
```
dbw9580, line 130: `entry` can be:
* a plain name
* a regex
* a table-partition specification

When entry is regex, it does not have a specific name (`getName` or `getTable` returns `null`), so we deal with this  case first. Otherwise, when partition specification is absent, simple add the table name.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+      builder.bypass().include().addPartition(entry.getTable(), partitionBuilder);
+    }
+
+    for (NamePatternEntry entry : mBypassEntry.getList().getExcludedEntries()) {
+      if (entry.isPattern()) {
+        builder.bypass().exclude().addPattern(entry.getPattern());
+      } else {
+        builder.bypass().exclude().addName(entry.getName());
+      }
+    }
```
dbw9580, line 156: Exclusion list on the table level does not support partition specification; to exclude a partition, include the table, and exclude that partition from its partition specfifcation, e.g. `{"include": [{"table": "table1", "partition": {"exclude": ["part1"]}}]}`.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+      if (node.isObject()) {
+        // otherwise, deserialize as an IncludeExcludeList object
+        return mapper.convertValue(
+            node,
+            mapper.getTypeFactory().constructParametricType(IncludeExcludeList.class, type)
+        );
+      }
+      throw new JsonParseException(mapper.treeAsTokens(node),
+          "invalid syntax, expecting array or object");
+    }
```
dbw9580, line 267: The object has a required field `tables`, and it accept either a list of (implicitly) included tables, or an explicit inclusion and exclusion specification object.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+  static class BypassTablesSpecDeserializer extends JsonDeserializer<BypassTablesSpec> {
+    @Override
+    public BypassTablesSpec deserialize(JsonParser jp, DeserializationContext cxt)
+        throws IOException, JsonProcessingException {
+      TablesEntryDeserializer<TableEntry> deserializer = new TablesEntryDeserializer<>();
+      IncludeExcludeList<TableEntry> list =
+          deserializer.deserializeToList(TableEntry.class, jp, cxt);
+      return new BypassTablesSpec(list);
+    }
+  }
```
dbw9580, line 290: These two deserializers differ only in their type arguments: ignored tables does not support partition specification, so the inner type is `NamePatternEntry`; bypassed tables do, so the inner type is `TableEntry`, which is a subclass of `NamePatternEntry`, and supports partition specification.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+      if (other == null) {
         return false;
-      } else if (getClass() != other.getClass()) {
+      }
+      if (getClass() != other.getClass()) {
         return false;
       }
-      BypassTableEntry entry = (BypassTableEntry) other;
-      return Objects.equals(mTableName, entry.mTableName);
+      return super.equals(other);
```
dbw9580, line 356: We do not compare partitions here, because one table cannot have two different partition specifications. When used in `Set`s, two table-partition spec objects with the same table name will be considered the same and either one is preserved.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
     }
 
     @Override
     public int hashCode() {
-      return Objects.hashCode(mTableName);
+      return super.hashCode();
```
dbw9580, line 361: Same as `equals`.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+        NamePatternEntryDeserializer deserializer = new NamePatternEntryDeserializer();
+        NamePatternEntry namePatternEntry =
+            deserializer.deserialize(mapper.treeAsTokens(node), cxt);
+        return new TableEntry(namePatternEntry);
+      } catch (JsonProcessingException e) {
+        // ignore, and try deserialize as a `TableEntry` object
+      }
+      if (node == null) {
+        return null;
+      }
```
dbw9580, line 386: `NamePatternEntryDeserializer` should have handled the `null` case itself, so we check it after the deserialization attempt.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+    /**
+     * @return pattern if the entry contains a regex pattern
+     */
+    @Nullable
+    public Pattern getPattern() {
+      return mPattern;
+    }
+
+    /**
+     * @return table name
```
dbw9580, line 474: Note to self: this is not necessarily a table name, it can be a partition name.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+      }
+      if (other == null) {
+        return false;
+      }
+      if (getClass() != other.getClass()) {
+        return false;
+      }
+      NamePatternEntry entry = (NamePatternEntry) other;
+      if (mIsPattern) {
+        return Objects.equals(mPattern.pattern(), entry.mPattern.pattern());
```
dbw9580, line 494: Compare pattern obejcts using the string representation it was created from.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+    }
+
+    /**
+     * Creates an empty list.
+     * @param <T> type of contained entry
+     * @return an empty list
+     */
+    public static <T extends NamePatternEntry> IncludeExcludeList<T> empty() {
+      return new IncludeExcludeList<>(Collections.emptySet(), Collections.emptySet());
+    }
```
dbw9580, line 571: Note to self: cache the empty object in a static field.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+
+  /**
+   * Checks if a table should be bypassed.
+   *
+   * @param tableName the table name
+   * @return true if the table is configured to be bypassed, false otherwise
+   * @see UdbMountSpec#hasFullyBypassedTable(String)
+   */
+  public boolean hasBypassedTable(String tableName) {
+    return !hasIgnoredTable(tableName) && mBypassed.hasTable(tableName);
```
dbw9580, line 58: If the same table is configured to be ignored and bypassed at the same time, ignore should take precedence, I think.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+   */
+  public boolean hasBypassedPartition(String tableName, String partitionName) {
+    return !hasIgnoredTable(tableName) && mBypassed.hasPartition(tableName, partitionName);
+  }
+
+  /**
+   * Wrapper class that takes into account inclusion and exclusion behaviour.
+   */
+  abstract static class InclusionExclusionWrapper<T extends NamePatternWrapper> {
+    protected final T mIncluded;
```
dbw9580, line 97: `T` can be a `NamePatternWrapper` or a `TablePartitionNamePatternWrapper`.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    // generic implementation
+    boolean has(String name) {
+      if (mExcluded.isEmpty()) {
+        return mIncluded.has(name);
+      } else {
+        return !mExcluded.has(name);
+      }
+    }
```
dbw9580, line 115: Since it cannot be an inclusion + exclusion list at the same time, the simplified inclusion/exclusion logic is simple: in case of an inclusion list, check if the name is present; in case of an exclusion list, check if the name is absent.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+  }
+
+  static final class TablePartitionWrapper
+      extends InclusionExclusionWrapper<TablePartitionNamePatternWrapper> {
```
dbw9580, line 119: This wrapper is subclass of `InclusionExclusionWrapper`, and also incorporates a `TablePartitionNamePatternWrapper`, therefore calling `.hasTable()`, `.hasFullTable()`, etc. on it takes the following factors into account by calling a chain of inherited methods:
* whether the table name is included or excluded;
* whether the table name is matched explicitly, or by a pattern;
* whether it has a partition specification.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+      // if a table is listed with an explicit literal name, then the partition specification
+      // or regex patterns should not be consulted:
+      // here use parent's hasExplicit since self's takes into account partition specs
+      if (super.hasExplicit(tableName)) {
+        return true;
+      }
+      if (mTablePartMap.containsKey(tableName)) {
+        // otherwise, check if it has an empty partition specification,
+        // or a non-empty partition spec.
+        // either way, it is explicitly listed, so no need to consult the regex patterns.
```
dbw9580, line 193: By these words, I mean in the following example:

```
{
  "tables": [
    {"regex": "table\\d"},
    {"table": "table1", "partitions": ["part1"]}
  ]
}
```

`table1` has non-empty partition specifications, although it is also captured by the regex, its name is explicitly listed through table-partition specification, so we consider it not a fully bypassed table.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+    @Override
+    boolean isEmpty() {
+      return super.isEmpty() && mTablePartMap.isEmpty();
+    }
+
+    boolean hasFullTable(String tableName) {
+      // if a table is listed with an explicit literal name, then the partition specification
+      // or regex patterns should not be consulted:
+      // here use parent's hasExplicit since self's takes into account partition specs
+      if (super.hasExplicit(tableName)) {
```
dbw9580, line 187: Should I write `mNames.contains(tableName)` instead?

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+
+    @Override
+    boolean has(String tableName) {
+      return super.has(tableName) || mTablePartMap.containsKey(tableName);
+    }
+
+    @Override
+    boolean hasExplicit(String tableName) {
+      return super.hasExplicit(tableName) || mTablePartMap.containsKey(tableName);
+    }
```
dbw9580, line 208: Here a table is considered explicitly listed if it is specified by a plain string, or a table-partition specification:

* `"tableName"`
* `{"table": "tableName", "partitions": [...]}`

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+     */
+    abstract T build();
+
+    private static <T> Set<T> addOrNewSet(Set<T> someSet, T element) {
+      if (someSet == null) {
+        Set<T> newSet = new HashSet<>();
+        newSet.add(element);
+        return newSet;
+      } else {
+        someSet.add(element);
```
dbw9580, line 354: Caveat: we cannot know for sure if the passed-in set is mutable or not. This might cause some unsupported operation exception.

table/server/common/src/main/java/alluxio/table/common/udb/UdbMountSpec.java
```diff
@@ -0,0 +1,589 @@
...
+
+    /**
+     * Attaches partition specification to a table.
+     * @param tableName the table name
+     * @param builder builder for partition specification
+     * @return the builder
+     */
+    public TablePartitionNamePatternWrapperBuilder addPartition(String tableName,
+                                                                SimpleWrapperBuilder builder) {
+      return addPartition(tableName, builder.build());
```
dbw9580, line 458: Here the second parameter is the builder instead of the built `SimpleWrapper`, since `SimpleWrapper` is package-private, and therefore `SimpleWrapperBuilder.build()` is made package-private. I tried to hide these inner classes that are implementation details from outside code.

dbw9580 (commented):
First batch of comments from me.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,537 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+      if (entry.isPattern()) {
+        builder.bypass().include().addPattern(entry.getPattern());
+        continue;
+      }
+      if (entry.getPartitions().isEmpty()) {
+        builder.bypass().include().addName(entry.getName());
+        continue;
+      }
+      IncludeExcludeList<NamePatternEntry> partitions = entry.getPartitions();
+      UdbMountSpec.SimpleWrapperBuilder partitionBuilder = new UdbMountSpec.SimpleWrapperBuilder();
```
dbw9580, line 132: `SimpleWrapperBuilder` and `SimpleWrapper` are not good names. Let's either give them a more descriptive name, or hide them altogether and find another way to build the partition spec (they are meant to be implementation details of the `UdbMountSpec` class).

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -28,25 +28,67 @@
...
    * @param bypassEntry bypass entry
+   * @param ignoreEntry ignore entry
    */
   @JsonCreator
-  public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
-    mBypassEntry = bypassEntry == null ? new BypassEntry(Collections.emptySet()) : bypassEntry;
+  public DbConfig(@JsonProperty("bypass") @Nullable BypassTablesSpec bypassEntry,
+                  @JsonProperty("ignore") @Nullable IgnoreTablesSpec ignoreEntry) {
+    mBypassEntry = bypassEntry == null ? new BypassTablesSpec(null) : bypassEntry;
+    mIgnoreEntry = ignoreEntry == null ? new IgnoreTablesSpec(null) : ignoreEntry;
```
dbw9580, line 91:
```suggestion
    mBypassEntry = new BypassTablesSpec(bypassEntry);
    mIgnoreEntry = new IgnoreTablesSpec(ignoreEntry);
```

dbw9580:
todo: add standalone json syntax checker to `alluxio table ` cli.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +97,535 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+     * Returns table name if the entry is not a regex entry.
+     * @return table name, null if the entry is a regex entry
+     */
+    @Nullable
+    public String getTable() {
+      return getName();
+    }
+
     @Override
     public boolean equals(Object other) {
```
dbw9580, line 346: Trying to remove this override gives me this error:

```
[INFO] --- spotbugs-maven-plugin:4.0.4:check (default) @ alluxio-table-server-master ---
[INFO] BugInstance size is 1
[INFO] Error size is 0
[INFO] Total bugs: 1
[ERROR] Medium: alluxio.master.table.DbConfig$TableEntry doesn't override DbConfig$NamePatternEntry.equals(Object) [alluxio.master.table.DbConfig$TableEntry] At DbConfig.java:[line 1] EQ_DOESNT_OVERRIDE_EQUALS
```

Maybe we should leave it as

```java
@Override
public boolean equals(Object other) {
  return super.equals(other);
}
```

so that we make our intention clear that we do not want to compare the partitions.

FYI @maobaolong .

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
dbw9580: Maybe `InclusionPatternWrapper`.

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -216,11 +212,16 @@ public SyncStatus sync(JournalContext context) throws IOException {
...
+      File configFile = new File(mConfigPath);
+      if (configFile.length() == 0) {
+        mDbConfig = DbConfig.empty();
+      } else {
+        try {
+          mDbConfig = mapper.readValue(configFile, DbConfig.class);
+        } catch (JsonProcessingException e) {
+          LOG.error("Failed to deserialize UDB config file {}, stays unsynced", mConfigPath, e);
+          throw e;
+        }
```
HelloHorizon, line 224: Make this part a method.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +148,559 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
    * @return an empty config instance
    */
   public static DbConfig empty() {
-    return new DbConfig(new BypassEntry(Collections.emptySet()));
+    return EMPTY_INSTANCE;
+  }
+
+  /**
+   * @return the {@link UdbAttachSpec} object
+   */
```
HelloHorizon, line 156: Add some descriptions for this complex method.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +148,559 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    } else if (other == null) {
+      return false;
+    } else if (getClass() != other.getClass()) {
+      return false;
+    }
```
HelloHorizon, line 222: Add some comments to explain why we want to compare with the class name.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
```diff
@@ -55,153 +148,559 @@ public DbConfig(@JsonProperty("bypass") @Nullable BypassEntry bypassEntry) {
...
+      if (this == other) {
+        return true;
+      } else if (other == null) {
+        return false;
+      } else if (getClass() != other.getClass()) {
+        return false;
+      }
+      TablesEntry that = (TablesEntry) other;
+      return Objects.equals(mList, that.mList);
+    }
```
HelloHorizon, line 288: Do we still need this? If we need, change to a helper function since we saw this code several times.

HelloHorizon (changes requested):
@dbw9580 thanks for the improvement, I left some comments

docs/en/core-services/Catalog.md
```diff
@@ -110,35 +115,224 @@ Each time the configuration file is changed, you can use `alluxio table sync` to
...
+   There are 2 ways to specify tables and partitions:
     
-    In the example above, table 1 is fully bypassed. Partition 1 and 2 of table 2 are bypassed, 
-    and any other partitions, if any, are not.
-
-> **Note:** When databases are attached, all tables are synced from the configured UDB.
-If out-of-band updates occur to the database or table and the user wants query results to reflect
-the updates, the database must be synced. See [Syncing Databases](#syncing-databases) for more
-information. 
+   1. Plain names: Use the exact name of a table or partition;
```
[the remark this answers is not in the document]
dbw9580: > There are 2 ways to specify tables and partitions:
>    
>   1. Plain names: Use the exact name of a table or partition;
>   2. Regular expressions: Use a pattern to capture a set of table or partition names.

Since we are explaining the possible ways to specify tables and partitions, I mentioned table and partition names before and after "plain names", I think it's unambiguous that "plain names" refer to tables' and partitions' names. Plus, "plain names" is in contrast to "regular expressions" afterwards.

table/server/common/src/main/java/alluxio/table/common/udb/UdbAttachSpec.java
```diff
@@ -0,0 +1,612 @@
...
+
+    boolean isEmpty() {
+      return mNames.size() == 0 && mPatterns.size() == 0;
+    }
+
+    boolean has(String name) {
+      return hasExplicit(name) || hasCoveredByPattern(name);
+    }
+
+    boolean hasExplicit(String name) {
```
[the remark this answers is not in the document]
dbw9580: A `NamePatternWrapper` contains a bunch of explicitly specified names, and a bunch of patterns. `hasExplicit` checks if a name is among those explicitly specified names, in contrast to `hasCoveredByPattern`, which checks if a name is covered by one of the patterns.

table/server/common/src/main/java/alluxio/table/common/udb/UdbAttachSpec.java
```diff
@@ -0,0 +1,612 @@
...
+    return mightBeShadowedIfTrue;
+  }
+
+  /**
+   * Wrapper class that takes into account inclusion and exclusion behaviour.
+   */
+  abstract static class InclusionExclusionWrapper<T extends NamePatternWrapper> {
+    protected final T mIncluded;
+    // excluded entries don't allow partition specifications -- they are name and regex only
+    protected final SimpleNamePatternWrapper mExcluded;
```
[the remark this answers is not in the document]
dbw9580: > can you remind me why we don't allow the partition specifications in exclusion entry?

The design is to make the `partitions` entry itself a `InclusionExclusionWrapper`, so excluded partitions are specified with itself. If we allowed partition specification on the table level, e.g.

```json
{
    "bypass": {
        "tables": {
            "exclude": [
                {"table":"table1", "partitions": ["part1"]}
            ]
        }
    }
}
```

This can be confusing because one may wonder what exactly is excluded, `table1` or `part1`?

As opposed to the above, by prohibiting partition specification in exclusion list, we can be more clear on this:

```json
{
    "bypass": {
        "tables": {
            "include": [
                {
                    "table":"table1", 
                    "partitions": {
                        "exclude": ["part1"]
                    }
                }
            ]
        }
    }
}
```

This is the correct way to say "`table1` is included in bypassing, except for its partition `part1`".

dbw9580:
@HelloHorizon addressed comments. PTAL, thanks!

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
dbw9580: Extracting the code into a function triggers spotbugs error:

```
[ERROR] Medium: Equals method for alluxio.master.table.DbConfig assumes the argument is of type DbConfig [alluxio.master.table.DbConfig] At DbConfig.java:[line 231] BC_EQUALS_METHOD_SHOULD_WORK_FOR_ALL_OBJECTS
```

table/server/common/src/main/java/alluxio/table/common/udb/UdbAttachSpec.java
```diff
@@ -0,0 +1,662 @@
...
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.annotation.Nullable;
+
+/**
+ * Tables and partitions inclusion and exclusion specification.
+ */
+public final class UdbAttachSpec {
```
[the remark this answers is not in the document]
apc999: besides, this entire class looks to me unnecessarily complicated compared to the functionality it provides. 
It introduces too many classes and look to me pretty hard to maintain.
@HelloHorizon can you work with @dbw9580  to simplify the design here?
To me , we just need to provide inclusion and exclusion rules, either by name or by regex.
Not sure if it is required to have the class hierarchy here

table/server/common/src/test/java/alluxio/table/common/udb/UdbAttachSpecTest.java
```diff
@@ -0,0 +1,264 @@
...
+    assertTrue(spec.isFullyBypassedTable(TABLE("a")));
+    assertTrue(spec.isBypassedPartition(TABLE(2), PART(1)));
+    assertTrue(spec.isBypassedPartition(TABLE(2), PART(2)));
+    assertTrue(spec.isBypassedPartition(TABLE(2), PART("a")));
+    assertTrue(spec.isBypassedPartition(TABLE("a"), PART(1)));
+    assertTrue(spec.isBypassedPartition(TABLE("a"), PART(2)));
+    assertTrue(spec.isBypassedPartition(TABLE("a"), PART("a")));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
apc999, line 129: `assertThrows` is the preferred way nowadays for expecting exception in JUNIT.
https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertThrows-java.lang.Class-org.junit.jupiter.api.function.Executable-

apc999 (changes requested):

table/server/common/src/main/java/alluxio/table/common/udb/UdbAttachSpec.java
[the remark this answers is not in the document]
HelloHorizon: `UdbAttachOptions` sounds good to me, and I will talk with @dbw9580 to simply the design

table/server/common/src/test/java/alluxio/table/common/udb/UdbAttachSpecTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+  @Before
+  public void before() {
+    mBuilder = new UdbAttachSpec.Builder();
+  }
+
+  private static final String ANY_TABLE = "any_table";
+  private static final String ANY_PART = "any_part";
+
+  private static String TABLE(int i) {
```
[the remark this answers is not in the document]
dbw9580: I intended to make the method names as succinct as possible, so `table(1)` would work the same  and look like as if I had written `"table1"` manually. Can we go with `table(int i)` instead?

table/server/common/src/main/java/alluxio/table/common/udb/UdbAttachSpec.java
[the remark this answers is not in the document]
dbw9580: Hi @apc999 @HelloHorizon I refactored both `DbConfig` and `UdbAttachOptions`, reduced number of inner classes and eliminated the use of custom deserializers. Please take another look. Thanks.

table/server/master/src/main/java/alluxio/master/table/DbConfig.java
dbw9580: This is circumvented with `    @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")` and `
@SuppressFBWarnings("EQ_UNUSUAL")
`.

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: Done in 9b05c31 .

table/server/common/src/test/java/alluxio/table/common/udb/UdbAttachSpecTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+  @Test
+  public void excludeNothingIsIncludeNothing() {
+    mBuilder.bypass()
+        .exclude().addNames(Collections.emptySet()).addPatterns(Collections.emptySet());
+    UdbAttachSpec spec = mBuilder.build();
+    assertFalse(spec.hasBypassedTable("any_table"));
+  }
+
+  /* Ignoring tests */
```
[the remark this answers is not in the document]
dbw9580: Added in c364bb1 .

dbw9580:
@apc999 @HelloHorizon we were talking about renaming `UdbAttachOptions` (aka `UdbBypassSpec`, `UdbMountSpec`) to `UdbConfig` during last review meeting. Today it occurs to me that there's already a `UdbConfiguration` class, and I already made an attempt in #13672 to make this class part of the existing `UdbConfiguration` class. 

I think `UdbConfig` is a bad name when there's already a class `UdbConfiguration`. Shall I proceed to merge `UdbAttachOptions` with the `UdbConfiguration` class, or look for a different name for it?

dbw9580:
I went with `UdbFilterSpec` and `UdbFilterSpecJsonSource` after talking to Jiacheng.

dbw9580:
@HelloHorizon I have done manual tests on this PR, and most things are working as expected, with a few minor problems, which have been fixed in the latest commits. Note that to have partition bypassing work properly, you need to rebase the base branch to include this commit fd8ff2c6a8ebb3765aa9ef2d3fcc4727fc9cb010.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+                .noneMatch(template -> template instanceof ExactNameTemplate
+                    && template.matches(tableName)),
+            "Table %s is already specified with exact name, "
+                + "cannot have partition specification at the same time",
+            tableName);
+      }
+    }
+  }
+
+  public static UdbFilterSpec empty() {
```
HelloHorizon, line 139: `/usr/src/alluxio/table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java:139:3: error: Missing a Javadoc comment.
`

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+    boolean isSpecifiedByName = matchedByTemplates(mBypassedTables, tableName);
+    if (isSpecifiedByName) {
+      if (mTableBypassMode == Mode.EXCLUDE) {
+        return false;
+      }
+      return shadowedByIgnore(tableName, true);
+    }
+    boolean hasBypassedPartition = mBypassedPartitions.containsKey(tableName);
+    if (hasBypassedPartition) {
+      // no need to check mTableBypassMode here, bc if a table has partition specifications,
```
HelloHorizon, line 160: what is bc mean here?

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+      case NONE:
+      default:
+        return false;
+    }
+  }
+
+  /**
+   * Checks if a table is shadowed by ignoring when it is also bypassed.
+   * @param tableName the table name
+   * @param mightBeShadowedIfTrue the return value of {@link #isBypassedTable(String)}, etc
```
HelloHorizon, line 230: what is the meaning of `mightBeShadowedIfTrue `?

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+   * @param tableName the table name
+   * @return true if the table is ignored, false otherwise
+   */
+  public boolean isIgnoredTable(String tableName) {
+    boolean isIgnored = matchedByTemplates(mIgnoredTables, tableName);
+    return handleInclusionExclusion(mIgnoreMode, isIgnored);
+  }
+
+  /**
+   * Checks by a partition's name if it should be bypassed.
```
HelloHorizon, line 196:
```suggestion
   * Check whether it is a bypassed partition by name.
```

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+    if (!mBypassedPartitions.containsKey(tableName)) {
+      return false;
+    }
+    Mode partitionMode = mBypassedPartitions.get(tableName).getFirst();
+    boolean isBypassed =
+        matchedByTemplates(mBypassedPartitions.get(tableName).getSecond(), partitionName);
+    return shadowedByIgnore(tableName, handleInclusionExclusion(partitionMode, isBypassed));
+  }
+
+  private static boolean handleInclusionExclusion(Mode mode, boolean selected) {
```
HelloHorizon, line 215: add some description here

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+
+  /**
+   * Checks if a table is shadowed by ignoring when it is also bypassed.
+   * @param tableName the table name
+   * @param mightBeShadowedIfTrue the return value of {@link #isBypassedTable(String)}, etc
+   * @return false if the table is shadowed by ignoring, mightBeShadowedIfTrue if it is not
+   */
+  private boolean shadowedByIgnore(String tableName, boolean mightBeShadowedIfTrue) {
+    if (mightBeShadowedIfTrue && isIgnoredTable(tableName)) {
+      // todo(bowen): this warning is never triggered bc if a table is ignored,
```
HelloHorizon, line 235: what needs to be changed here?

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -211,18 +207,7 @@ public SyncStatus sync(JournalContext context) throws IOException {
...
-      }
-      ObjectMapper mapper = new ObjectMapper();
-      try {
-        mDbConfig = mapper.readValue(new File(mConfigPath), DbConfig.class);
-      } catch (JsonProcessingException e) {
-        LOG.error("Failed to deserialize UDB config file {}, stays unsynced", mConfigPath, e);
-        throw e;
-      }
-    }
+    loadFilterSpec();
```
HelloHorizon, line 210: Add some description here

HelloHorizon (changes requested):
@dbw9580  Overall looks good to me, left some minor comments.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+          Mode.NONE,
+          Collections.emptyMap(),
+          Collections.emptySet(),
+          Mode.NONE);
+
+  /**
+   * Bypass/ignore entries.
+   */
+  public enum Entry {
+    BYPASS, IGNORE, NONE;
```
apc999, line 48:
```suggestion
    // what BYPASS means
    BYPASS, 
    // what IGNORE means
    IGNORE, 
    // what NONE means
    NONE;
```

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+  /**
+   * Inclusion/exclusion modes of tables and partitions.
+   */
+  public enum Mode {
+    INCLUDE, EXCLUDE, NONE;
```
apc999, line 55: - same here, javadoc these enum items
- Maybe called `EntryMode` ?

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+          mIgnoredTables,
+          mIgnoreMode
+      );
+      // release references to prevent reuse of builder
+      mBypassedTables = ImmutableSet.of();
+      mTableBypassMode = Mode.NONE;
+      mBypassedPartitions = ImmutableMap.of();
+      mPartitionBypassModes = ImmutableMap.of();
+      mIgnoredTables = ImmutableSet.of();
+      mIgnoreMode = Mode.NONE;
```
apc999, line 501: I view this unnecessary. We have a lot of builder pattern in Alluxio repo. Typically we don't clean the references for builder. Following the convention is fine.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+        } else {
+          set.add(partition);
+          return set;
+        }
+      });
+      return this;
+    }
+
+    /**
+     * Build a {@link UdbFilterSpec}.
```
apc999, line 473:
```suggestion
     * Builds a {@link UdbFilterSpec}.
```

We typically use Singular Tense in Java doc

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+     * The builder should not be used after this method returns.
+     * @return {@link UdbFilterSpec}
+     * @throws IllegalArgumentException when invalid or conflicting configurations
+     *                                          have been provided
+     */
+    public UdbFilterSpec build() {
+      ImmutableMap.Builder<String, Pair<Mode, Set<NameTemplate>>> partitionsBuilder
+          = ImmutableMap.builder();
```
apc999, line 481:
```suggestion
      ImmutableMap.Builder<String, Pair<Mode, Set<NameTemplate>>> partitionsBuilder =
            ImmutableMap.builder();
```

table/server/common/src/test/java/alluxio/table/common/udb/UdbFilterSpecTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+import alluxio.table.common.udb.UdbFilterSpec.Mode;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.regex.Pattern;
+
+public class UdbFilterSpecTest {
+  private UdbFilterSpec.Builder mBuilder;
```
apc999, line 26:
```suggestion
  private UdbFilterSpec.Builder mBuilder = new UdbFilterSpec.Builder();
```
and you don't need `before()`

better to keep code concise as long as readability is not reduced

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -353,6 +338,30 @@ public SyncStatus sync(JournalContext context) throws IOException {
...
+    File configFile = new File(mConfigPath);
+    if (configFile.length() == 0) {
+      mFilterSpec = UdbFilterSpec.empty();
+      return;
+    }
+    try {
+      mFilterSpec = mapper.readValue(configFile, UdbFilterSpecJsonSource.class).getUdbFilterSpec();
+    } catch (JsonProcessingException e) {
+      LOG.error("Failed to parse UDB config file {}, likely syntax error", mConfigPath, e);
+      throw new IOException(String.format("Failed to parse UDB config file %s", mConfigPath), e);
```
apc999, line 358: typically LOG and throw is anti-pattern as it may cause the same error logged multiple times. 
please be very sure this is required or let's just throw the IOE and leave the caller side to decide if this shall be logged

table/server/master/src/main/java/alluxio/master/table/Database.java
```diff
@@ -353,6 +338,30 @@ public SyncStatus sync(JournalContext context) throws IOException {
...
+    } catch (IllegalArgumentException e) {
+      LOG.error("Filter specification contains invalid configuration", e);
+      throw new IOException("Filter specification contains invalid configuration", e);
```
apc999, line 361: same here

table/server/master/src/test/java/alluxio/master/table/UdbFilterSpecJsonSourceTest.java
```diff
@@ -0,0 +1,320 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+public class UdbFilterSpecJsonSourceTest {
+  private ObjectMapper mMapper;
+
+  @Before
+  public void before() {
+    mMapper = new ObjectMapper();
```
apc999, line 52: same here , unnecessary `before`

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
```diff
@@ -21,7 +21,7 @@
 import alluxio.table.common.UdbPartition;
 import alluxio.table.common.layout.HiveLayout;
 import alluxio.table.common.udb.PathTranslator;
-import alluxio.table.common.udb.UdbBypassSpec;
+import alluxio.table.common.udb.UdbFilterSpec;
```
apc999, line 24: keep each import section alphabetically

table/server/master/src/main/java/alluxio/master/table/Database.java
apc999: do we need to load filter spec everytime on `sync`?

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+     * @return true if the names match, false otherwise
+     */
+    boolean matches(String name);
+  }
+
+  static class ExactNameTemplate implements NameTemplate {
+    private final String mExactName;
+
+    public ExactNameTemplate(String exactName) {
+      Preconditions.checkArgument(!exactName.isEmpty(), "empty name");
```
apc999, line 264: `checkNotNull` too?

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+  private static boolean matchedByTemplates(Set<NameTemplate> templates, String name) {
+    return templates.stream().anyMatch(p -> p.matches(name));
+  }
+
+  /**
+   * Abstract name for tables and partitions.
+   */
+  interface NameTemplate {
```
apc999, line 250: How about `EntryName`

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+    /**
+     * Checks whether the argument name matches the configured name pattern.
+     *
+     * @param name the concrete name to check for
+     * @return true if the names match, false otherwise
+     */
+    boolean matches(String name);
+  }
+
+  static class ExactNameTemplate implements NameTemplate {
```
apc999, line 260: `StringEntryName` or `ExactEntryName`

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+
+    @Override
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .add("exactName", mExactName)
+          .toString();
+    }
+  }
+
+  static class PatternNameTemplate implements NameTemplate {
```
apc999, line 298: `PatternEntryName`

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+    private final Pattern mPattern;
+
+    public PatternNameTemplate(Pattern pattern) {
+      mPattern = pattern;
```
apc999, line 302:
```suggestion
      mPattern = Preconditions.checkNotNull(pattern, "pattern");
```

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
```diff
@@ -0,0 +1,505 @@
...
+      switch (entry) {
+        case BYPASS:
+          mBypassedTables.add(item);
+          break;
+        case IGNORE:
+          mIgnoredTables.add(item);
+          break;
+        case NONE:
+        default:
+          // unreachable
```
apc999, line 424: if this is really unreachable, let's throw an IllegalStateException exception or at least `LOG.error`

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
apc999: maybe `EntryType` ?
also do we really need NONE?

apc999 (changes requested):
thanks @dbw9580  we are getting there.
Left a bunch of minor comments

docs/en/core-services/Catalog.md
```diff
@@ -102,43 +102,219 @@ $ ${ALLUXIO_HOME}/bin/alluxio table attachdb --db alluxio_db hive \
...
+    
+   In the above example, table 1, 2, and 3 are fully bypassed.
+   Partition 1, 2, and 3 of table 4 are bypassed, and any other partitions, if any, are not.
+   
+2. Ignoring tables:
+
+   You can specify some tables to be ignored when attaching the database to Alluxio. Compared to 
+   bypassed tables, Alluxio will not try to mount the ignored tables, therefore they are 
+   invisible to the Alluxio client. 
+
```
dbw9580, line 170: Provide use cases for ignored tables, i.e. when this feature is useful, similar to the documentation about bypassed tables above.

dbw9580: @HelloHorizon @apc999 opinions?

HelloHorizon: LGTM

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Added doc in cef77153696b4028a525829d6d17d898fa972e6c

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Means because. bad choice of acronym 😣

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: this boolean is going to be the return value of `isBypassedTable` and `isBypassedPartition`, etc. When it is `true`, before we can actually return `true` to the caller, we need to make sure the table or partition in question is not ignored (`shadowed`). When it is `false`, we don't need to. So the check is only necessary if the argument is `true`, hence the name.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Fixed in cef77153696b4028a525829d6d17d898fa972e6c. Removed unnecessary mention of "by name" in description as it is unambiguous given the parameters' description.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Fixed in cef77153696b4028a525829d6d17d898fa972e6c

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: The check for ignored tables happens before the `mount` call. The checks for bypassed tables are done in the `mount` method, and carried out against the non-ignored tables. Therefore, we would never know if a table is misconfigured to be ignored while it is actually meant to be bypassed, since it is filtered out for the check.

I'm thinking of doing the check in the `isIgnoredTable` method, instead of doing it in the bypass methods.

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: > do we need to load filter spec everytime on sync?

The original intention by @maobaolong is to make the config file part of the synchronization job.  https://github.com/Alluxio/alluxio/pull/13455#discussion_r638114342

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Added doc in cef77153696b4028a525829d6d17d898fa972e6c. Removed unnecessary `NONE` type in 2719fbf31d84e0f3f486d1892a734da5f3471e62.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Renamed in 0c85e845413da76eefa233ded6e98076946dd94f. Added doc in cef77153696b4028a525829d6d17d898fa972e6c.

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Removed in 40f91a5798b714c552f0fc2659a3549429ca3dfa

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Did a sweep in cef77153696b4028a525829d6d17d898fa972e6c

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Renamed in 4c0380b7c97d564304be200db2f40eb680da9f84

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Renamed to `ExactEntryName` in 4c0380b7c97d564304be200db2f40eb680da9f84

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: 4c0380b7c97d564304be200db2f40eb680da9f84

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Fixed in 4c0380b7c97d564304be200db2f40eb680da9f84

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Fixed in 4c0380b7c97d564304be200db2f40eb680da9f84

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Fixed in 2719fbf31d84e0f3f486d1892a734da5f3471e62

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: This is deleted in 81a1fd41a67bf67c1dd3839cb5103f74f172f1a6

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Done in 81a1fd41a67bf67c1dd3839cb5103f74f172f1a6

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: > Add some description here

Added in 2361b59807ed5d1b5cc50ef3e1e70391699c87d8

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Fixed in 391b8b369413e0ba5c9528c2c0645cc54ee76932.

table/server/common/src/test/java/alluxio/table/common/udb/UdbFilterSpecTest.java
dbw9580: Fixed in 391b8b369413e0ba5c9528c2c0645cc54ee76932

table/server/master/src/test/java/alluxio/master/table/UdbFilterSpecJsonSourceTest.java
dbw9580: Fixed in 391b8b369413e0ba5c9528c2c0645cc54ee76932

table/server/underdb/glue/src/main/java/alluxio/table/under/glue/GlueDatabase.java
dbw9580: Optimized and sorted imports in d528def944255c58bb67779fb96c016afb41c5e7

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: Fixed in 14411fbc2c1e88fe8ad8993bba095b0e13f4d213

table/server/master/src/main/java/alluxio/master/table/Database.java
dbw9580: fixed in 14411fbc2c1e88fe8ad8993bba095b0e13f4d213

table/server/common/src/main/java/alluxio/table/common/udb/UdbFilterSpec.java
dbw9580: Throws IllegalStateException in 2719fbf31d84e0f3f486d1892a734da5f3471e62

dbw9580:
@HelloHorizon @apc999 addressed comments in last batch of commits. PTAL, thank you!

HelloHorizon (approved):
LGTM, @apc999 do you mind to take another pass?

dbw9580:
@HelloHorizon 

> what will happen if there isn't any DB configuration file?

0. when the user did not specify a config file, an empty filter spec is used, and it has no effect;
1. when the user specified the `CatalogProperty.DB_CONFIG_FILE` option on the command line, but the option points to a non-existent file, the `loadFilterSpec` method will throw a `FileNotFoundException` and thus cause the sync to fail;
2. when the file is present but empty, `loadFilterSpec` will complete successfully, and an empty filter spec will be returned. This is checked by `Files.exists()` and `File.length()` methods;
3. otherwise, the parser will try to parse the file as JSON, and if the file has invalid syntax (including invalid JSON syntax and invalid syntax as our config file), an `IOException` will be thrown and abort the sync process.

HelloHorizon:
@dbw9580 sounds good! do you mind putting it into code comments?

dbw9580:
> do you mind putting it into code comments?

Done.

dbw9580:
ping @apc999 Can you please take another look?

HelloHorizon:
ping @apc999

dbw9580:
ping @apc999

PR #13761, opened 2021-07-08
Remove embedded journal heartbeat interval configuration

### LINKED ISSUE #13739: Configuration of embedded journal heartbeat interval is invalid
**Alluxio Version:**
2.5.0

**Describe the bug**
modifying alluxio.master.embedded.journal.heartbeat.interval should modify the raft heartbeat interval, but actually heartbeat interval will always be half of the election timeout.

**To Reproduce**
modifying alluxio.master.embedded.journal.heartbeat.interval value.

**Expected behavior**
raft heartbeat interval should change.

**Urgency**
not urgency.

**Additional context**
[related code](https://github.com/Alluxio/alluxio/blob/master/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java#L329)

qian0817:

Remove embedded journal heartbeat interval configuration.

embedded journal heartbeat interval configuration has no use, it's better to remove it.

remove property key `alluxio.master.embedded.journal.heartbeat.interval`
Fixes #13739


apc999:
@ggezer  can you take a look at this PR and provide some advice as we discussed offline?

ggezer:
@qian0817 Heartbeat interval is fundamental RAFT construct. We cannot get rid of it.

apc999:
@qian0817 any reason you think heartbeat interval is not used here?

qian0817:
Ratis will set heartbeat interval to half of the election timeout and it does't provide a configuration item to modify. This configuration item is currently ineffective, so I think we should remove this configuration.

ggezer:
Right, RATIS seems to have disabled this conf with min-max timeout concept?
@qian0817 Do you think we should repurpose this conf and provide MIN/MAX election timeout instead of hardcoding MAX to 2*MIN?

qian0817:
It is best to provide configuration of MIN/MAX election timeout, and this can be discussed in a new PR.

ggezer (approved):

ggezer:
> It is best to provide configuration of MIN/MAX election timeout, and this can be discussed in a new PR.

Sounds good. Looking forward to min/max PR.

ggezer:

Issue #13763, opened 2021-07-08
Add more information when registering the worker


codings-dan:
**Is your feature request related to a problem? Please describe.**
When upgrading Alluxio in our cluster, we typically do a rolling upgrade in our K8s cluster. During that, we typically upgrade the workers in batches, so some workers in the cluster will have different version/revision from others. We would like that to be reported to the master so the master can display those information on the UI and in `alluxio fsadmin` commands. 

It will also be very helpful if we can associate issues/performance with the worker version/revision, so during the rolling upgrade we can be notified if the upgrade is bad and we should roll back.

Especially when we are using `SNAPSHOT` versions, it is hard to tell which version/revision the worker is on.

**Describe the solution you'd like**
We want to get more information about workers on the master side. Add the version and revision information of the worker

**Urgency**
Normal

**Additional context**
No

EDIT: rephrased use case and requirements on 2021/10/27

LuQQiu:
@maobaolong @codings-dan As we discussed in the last AI sync, can you help add more detailed upper level information about why this change is needed? Like the background usecase

PR #13764, opened 2021-07-08
Add the version information and start time of the worker

codings-dan:

As described in #13763，add the version information and start time of the worker。


maobaolong:
@codings-dan I think you'd better to rebase your branch to alluxio master, for now, it have more than 20 commits.

maobaolong (changes requested):
Please revert this changes. webui/npm-shrinkwrap.json

core/server/master/src/main/java/alluxio/master/block/meta/StaticWorkerMeta.java
```diff
@@ -29,7 +29,10 @@
   final long mId;
   /** Start time of the worker in ms. */
   final long mStartTimeMs;
-
+  /** the start time of version. */
+  long mStartTime;
```
jiacheliu3, line 33: What is the use of this start time? Will this be very different from the `mStartTimeMs`? If so can you add some context about the scenario?

core/server/master/src/main/java/alluxio/master/block/meta/StaticWorkerMeta.java
```diff
@@ -29,7 +29,10 @@
...
+  /** the version of worker. */
+  String mVersion;
```
jiacheliu3, line 35: This class is intended to be static and therefore thread-safe. So please make this `final`.

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterSync.java
```diff
@@ -109,7 +113,7 @@ private void registerWithMaster() throws IOException {
     mMasterClient.register(mWorkerId.get(),
         storageTierAssoc.getOrderedStorageAliases(), storeMeta.getCapacityBytesOnTiers(),
         storeMeta.getUsedBytesOnTiers(), storeMeta.getBlockListByStorageLocation(),
-        storeMeta.getLostStorage(), configList);
+        storeMeta.getLostStorage(), configList, mStartTime, ProjectConstants.VERSION);
```
jiacheliu3, line 116: For the version, maybe add the revision too? `version` is like `2.6.0` and revision is the hashcode which means the commit in the git repo. You can see how that is got from the `alluxio.cli.Version` command.

jiacheliu3 (changes requested):
Thanks @codings-dan . I added a few comments.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -902,19 +902,19 @@ public long getWorkerId(WorkerNetAddress workerNetAddress) {
   public void workerRegister(long workerId, List<String> storageTiers,
       Map<String, Long> totalBytesOnTiers, Map<String, Long> usedBytesOnTiers,
       Map<BlockLocation, List<Long>> currentBlocksOnLocation,
-      Map<String, StorageList> lostStorage, RegisterWorkerPOptions options)
-      throws NotFoundException {
+      Map<String, StorageList> lostStorage, RegisterWorkerPOptions options,
+      long startTime, String version) throws NotFoundException {
```
apc999, line 906: nit: leave `RegisterWorkerPOptions options` as the last argument in this method

core/server/master/src/main/java/alluxio/master/block/meta/StaticWorkerMeta.java
```diff
@@ -29,7 +29,10 @@
   final long mId;
   /** Start time of the worker in ms. */
   final long mStartTimeMs;
-
+  /** the start time of version. */
```
apc999, line 32: what does it mean "start time of version'?

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterSync.java
maobaolong: Good suggestion, it helps.

LuQQiu:
@jiacheliu3 PTAL, thanks!

core/server/master/src/main/java/alluxio/master/block/meta/StaticWorkerMeta.java
jiacheliu3: Again this class is immutable and fields should be final

codings-dan: Because of the limitation of the fianl keyword, these data cannot be stored. We can store the data in a newly created class or in MasterWorkerInfo. What do you think?

jiacheliu3: @codings-dan Yes having a separated class will be a good way. Maybe `WorkerTimeMeta` can be a candidate name. In order to avoid yet another lock on this object (which makes it even more complicated to acquire the necessary locks), you can either guarantee thread safety within this class, or use an `AtomicReference<WorkerTimeMeta>` and create one more object each time any time field is updated.

ggezer (changes requested):
Why do you need these information coming from worker? 
Please clarify what you're trying to achieve in either PR description or the linked issue.

jiacheliu3:
> Why do you need these information coming from worker?
> Please clarify what you're trying to achieve in either PR description or the linked issue.

+1
@maobaolong Could you add more information to the linked issues, especially which field is for what use case? Thx!

codings-dan:
@jiacheliu3 I have add some information to the linked issues, PTAL,thx!

LuQQiu:
@jiacheliu3 @ggezer PTAL, thanks!

codings-dan:
@ggezer @jiacheliu3 I have added the relevant information about the added field, considering that startTime is a redundant field, delete it.PTAL

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -435,7 +437,7 @@ public StorageTierAssoc getStorageTierAssoc() {
    * @return the start time in milliseconds
    */
   public long getStartTime() {
```
ggezer, line 439: can you change the method name as well?

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -611,4 +613,38 @@ ReentrantReadWriteLock getLock(WorkerMetaLockSection lockType) {
...
+   * @return the version of worker
+   */
+  public String getVersion() {
+    return mWorkerTimeMeta.getVersion();
+  }
+
+  /**
+   * No locking required.
+   *
+   * @return the version of worker
```
ggezer, line 645: update the javadoc.

core/server/master/src/main/java/alluxio/master/block/meta/WorkerTimeMeta.java
```diff
@@ -0,0 +1,86 @@
...
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.master.block.meta;
+
+/**
+ *An object representation of the worker metadata.This class is not thread safe
+ *so accessing or updating the fields need to use AtomicReference<WorkerTimeMeta>.
+ */
+public class WorkerTimeMeta {
```
ggezer, line 18: This class is overloaded imo. it has time and version info together.

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -126,6 +126,7 @@
   /** Block Store manager. */
   private final BlockStore mLocalBlockStore;
   private WorkerNetAddress mAddress;
+  private long mStartTime;
```
ggezer, line 129: redundant.

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -237,7 +238,8 @@ public void start(WorkerNetAddress address) throws IOException {
 
     // Setup BlockMasterSync
     mBlockMasterSync = mResourceCloser
-        .register(new BlockMasterSync(this, mWorkerId, mAddress, mBlockMasterClientPool));
+        .register(new BlockMasterSync(this, mWorkerId, mAddress,
+                mBlockMasterClientPool));
```
ggezer, line 242: no need to move down.

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -252,6 +252,10 @@ message RegisterWorkerPRequest {
   map<string, StorageList> lostStorage = 7;
   /** use repeated fields to represent mapping from BlockStoreLocationProto to TierList */
   repeated LocationBlockIdListEntry currentBlocks = 8;
+  /** the version of the worker. */
+  optional string version = 9;
+  /** the revision of the worker. */
+  optional string revision = 10;
```
ggezer, line 258: why do you need revision?

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterSync.java
```diff
@@ -73,6 +74,8 @@
 
   /** Last System.currentTimeMillis() timestamp when a heartbeat successfully completed. */
   private long mLastSuccessfulHeartbeatMs;
+  /** the start time of worker. */
+  private long mStartTime;
```
ggezer, line 78: redundant.

core/server/master/src/main/java/alluxio/master/block/meta/WorkerTimeMeta.java
```diff
@@ -0,0 +1,86 @@
...
+  /** the start time of worker. */
+  private long mStartTime;
```
ggezer, line 20: This is not even used. 
Also, I don't see how this will be different than registration time.

ggezer (changes requested):

PR #13767, opened 2021-07-08
Fix process local read write client side logics and add unit tests

### LINKED ISSUE #13766: Worker Fuse should support internal read when cache miss
**Alluxio Version:**
Master Alluxio 2.6

**Describe the bug**
When a block is cached in the local Alluxio worker and fuse is launched in the worker process, fuse read can go through worker internal read mechanism without grpc.
However, when the block is not cached in any of the Alluxio workers, even when we assign the local worker to be the worker to read from UFS and do the cache job, we still go through external gRPC call to connect to the local worker.

**To Reproduce**
launch a worker with internal fuse.
Prepare a file in UFS and is not cached in Alluxio.
Read the file through fuse mount point.
View worker metrics, will find BytesReadDirect = 0

**Expected behavior**
All read communication between the worker and the embedded fuse should go through internal method call without gRPC involved. No matter the block is in worker storage or in UFS.

**Urgency**
Urgent

**Additional context**

LuQQiu:
Fixes #13766 
Support fuse read from worker through internal method call when cache miss.
Add worker is local check to correct fuse write to process local worker logics.
Add process local read write related metrics

When cache miss, the local worker can still help fetch the data from UFS and provide to the internal client without RPC involved.
Remove the unneeded RPC overhead between Fuse and Worker.
Correct the fuse write to local worker logics

No

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
@@ -204,9 +204,20 @@ public final class AlluxioBlockStore {
               .setLocations(locations))
           .setBlockWorkerInfos(blockWorkerInfo);
       dataSource = policy.getWorker(getWorkerOptions);
-      if (dataSource != null && mContext.hasProcessLocalWorker()
-          && dataSource.equals(mContext.getNodeLocalWorker())) {
-        dataSourceType = BlockInStreamSource.PROCESS_LOCAL;
+      if (dataSource != null) {
+        if (mContext.hasProcessLocalWorker()
+            && dataSource.equals(mContext.getNodeLocalWorker())) {
+          dataSourceType = BlockInStreamSource.PROCESS_LOCAL;
+          LOG.debug("Create BlockInStream to read data from UFS through process local worker {}",
+              dataSource);
+        } else {
+          LOG.debug("Create BlockInStream to read data from UFS through worker {} "
+              + "(client embedded in local worker process: {},"
+                  + "client co-located with worker in different processes: {}, "
+                  + "local worker address: {})",
+              dataSource, mContext.hasProcessLocalWorker(), mContext.hasNodeLocalWorker(),
+              mContext.hasNodeLocalWorker() ? mContext.getNodeLocalWorker() : "N/A");
+        }
       }
     }
 
--- a/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
@@ -204,7 +204,6 @@ public final class AlluxioBlockStore {
               .setLocations(locations))
           .setBlockWorkerInfos(blockWorkerInfo);
       dataSource = policy.getWorker(getWorkerOptions);
-      // TODO(lu) update the bytesReadUfs metrics
       if (dataSource != null && mContext.hasProcessLocalWorker()
           && dataSource.equals(mContext.getNodeLocalWorker())) {
         dataSourceType = BlockInStreamSource.PROCESS_LOCAL;
--- a/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
@@ -204,7 +204,13 @@ public final class AlluxioBlockStore {
               .setLocations(locations))
           .setBlockWorkerInfos(blockWorkerInfo);
       dataSource = policy.getWorker(getWorkerOptions);
+      // TODO(lu) update the bytesReadUfs metrics
+      if (dataSource != null && mContext.hasProcessLocalWorker()
+          && dataSource.equals(mContext.getNodeLocalWorker())) {
+        dataSourceType = BlockInStreamSource.PROCESS_LOCAL;
+      }
     }
+
     if (dataSource == null) {
       throw new UnavailableException(ExceptionMessage.NO_WORKER_AVAILABLE.getMessage());
     }
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -39,6 +39,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
@@ -460,10 +461,12 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
   }
 
   /**
-   * @return whether the reader is reading data directly from a local file
+   * @return the underlying data reader
    */
-  public DataReader.DataReaderType getDataReaderType() {
-    return mDataReaderFactory.getDataReaderType();
+  @VisibleForTesting
+  @Nullable
+  public DataReader getDataReader() {
+    return mDataReader;
   }
 
   /**
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -39,7 +39,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
@@ -461,12 +460,11 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
   }
 
   /**
-   * @return the underlying data reader
+   * @return the underlying data reader factory
    */
   @VisibleForTesting
-  @Nullable
-  public DataReader getDataReader() {
-    return mDataReader;
+  public DataReader.Factory getDataReaderFactory() {
+    return mDataReaderFactory;
   }
 
   /**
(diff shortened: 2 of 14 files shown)


LuQQiu:
Will have a followup PR about how to correctly record the BytesReadUfs metrics.

apc999:
can you add UT or integration tests on this?

core/client/fs/src/main/java/alluxio/client/block/stream/DataWriter.java
```diff
@@ -62,12 +62,15 @@ public static DataWriter create(FileSystemContext context, long blockId, long bl
...
-      if (context.hasProcessLocalWorker() && !ufsFallbackEnabled) {
+      if (workerIsLocal && context.hasProcessLocalWorker() && !ufsFallbackEnabled) {
+        LOG.debug("Creating worker process local output stream for block {} @ {}",
+            blockId, address);
         return BlockWorkerDataWriter.create(context, blockId, blockSize, options);
       }
 
-      if (CommonUtils.isLocalHost(address, alluxioConf)
+      if (workerIsLocal
           && shortCircuit
```
apc999, line 74: wiered format

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
```diff
@@ -204,7 +204,13 @@ public BlockInStream getInStream(BlockInfo info, InStreamOptions options,
               .setLocations(locations))
           .setBlockWorkerInfos(blockWorkerInfo);
       dataSource = policy.getWorker(getWorkerOptions);
+      // TODO(lu) update the bytesReadUfs metrics
+      if (dataSource != null && mContext.hasProcessLocalWorker()
+          && dataSource.equals(mContext.getNodeLocalWorker())) {
```
ggezer, line 209: I'm not sure if this is right. Why do we consider node-local worker as process-local? Would this cause an issue somewhere when during process-local communication?

apc999: `mContext.hasProcessLocalWorker() == true` assures the client is running together with a worker.
In this case, there is no need to go RPC to communicate with this worker

core/client/fs/src/main/java/alluxio/client/block/stream/DataWriter.java
```diff
@@ -62,13 +62,15 @@ public static DataWriter create(FileSystemContext context, long blockId, long bl
           alluxioConf.getBoolean(PropertyKey.USER_SHORT_CIRCUIT_PREFERRED);
       boolean ufsFallbackEnabled = options.getWriteType() == WriteType.ASYNC_THROUGH
           && alluxioConf.getBoolean(PropertyKey.USER_FILE_UFS_TIER_ENABLED);
+      boolean workerIsLocal = CommonUtils.isLocalHost(address, alluxioConf);
 
-      if (context.hasProcessLocalWorker() && !ufsFallbackEnabled) {
+      if (workerIsLocal && context.hasProcessLocalWorker() && !ufsFallbackEnabled) {
+        LOG.debug("Creating worker process local output stream for block {} @ {}",
+            blockId, address);
```
jiacheliu3, line 69: Thanks a lot for the extra logs!

core/client/fs/src/main/java/alluxio/client/block/stream/DataReader.java
```diff
@@ -21,6 +21,17 @@
...
 
+  /**
+   * The data reader type.
+   */
+  enum DataReaderType {
+    BLOCK_WORKER,
+    GRPC,
+    SHORT_CIRCUIT,
+    SHARED,
+    TEST
```
jiacheliu3, line 32: Comment to explain each one of them?

core/client/fs/src/main/java/alluxio/client/block/stream/DataWriter.java
```diff
@@ -62,13 +62,15 @@ public static DataWriter create(FileSystemContext context, long blockId, long bl
...
         return BlockWorkerDataWriter.create(context, blockId, blockSize, options);
       }
 
-      if (CommonUtils.isLocalHost(address, alluxioConf)
-          && shortCircuit
+      if (workerIsLocal && shortCircuit
```
jiacheliu3, line 73: Other than the logs when the if conditions are all met, I think it is also very important to know which condition is NOT met. In other words, we may expect the if condition to be true but when it is not true, there should be some hints how it is not true. I'd recommend this:
```
if (conditionA && conditionB && conditionC) {
  LOG.debug(...)
  // do sth
  // Actually if you can return here, just return explicitly to avoid confusing
}
LOG.debug("a message that contains conditionA={} conditionB={} conditionC={}")
...
```
This is not logging more than you are already doing right now, because we are logging one line when the IF block executes, we should also accept logging one message for the implicit ELSE block.

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
jiacheliu3: What I get from this conversation is that people don't understand what is `PROCESS_LOCAL`, when should it happen and what does that imply. Those information should either be commented here (when `PROCESS_LOCAL` value is given) or in the enum definition (comment after each field).

jiacheliu3 (changes requested):
@LuQQiu Thanks! Added some comment requesting for more comments. Requesting these just because last week I had to read this part of the code when debugging and it took me quite a while to understand the intention. The if condition is a combination of many conditions (which each comes from yet a few other conditions). It is hard to reverse-guess what each scenario means and hard to debug. Those are gaps I hope we can close by comments and logging (as you are doing right now).

core/client/fs/src/main/java/alluxio/client/block/AlluxioBlockStore.java
LuQQiu: The definition is in the enum comments

LuQQiu:
@apc999 @jiacheliu3 Addressed the comments, PTAL, thanks!

core/client/fs/src/main/java/alluxio/client/block/stream/DataReader.java
```diff
@@ -22,14 +22,21 @@
 public interface DataReader extends Closeable {
 
   /**
-   * The data reader type.
+   * The data reader type represents different ways to read data.
    */
   enum DataReaderType {
-    BLOCK_WORKER,
-    GRPC,
+    BLOCK_WORKER, // Read from a worker in the same process
```
jiacheliu3, line 28: nit: If this only applies to FUSE, maybe mention this in the comment? People may not understand when you have a worker in the same process.

core/client/fs/src/main/java/alluxio/client/block/stream/DataReader.java
```diff
@@ -22,14 +22,21 @@
 public interface DataReader extends Closeable {
 
   /**
-   * The data reader type.
+   * The data reader type represents different ways to read data.
    */
   enum DataReaderType {
```
jiacheliu3, line 27: Which one applies for domain socket? If domain socket and short circuit both use `SHORT_CIRCUIT` maybe we split them or rename?

jiacheliu3 (approved):
Thanks for the improvements! LGTM with some minor comments.

core/client/fs/src/main/java/alluxio/client/block/stream/DataReader.java
LuQQiu: Domain socket using GrpcDataReader. I don't feel I have enough knowledge to do domain socket related changes. The reader type here refects the data reader names/classes that are using.

LuQQiu:
@apc999 @jiacheliu3  please take a look, thanks!

core/client/fs/src/test/java/alluxio/client/block/AlluxioBlockStoreTest.java
```diff
@@ -284,28 +289,49 @@ public Object answer(InvocationOnMock invocation) {
...
+        Lists.newArrayList(new BlockWorkerInfo(remote, 100, 0),
+            new BlockWorkerInfo(local, 100, 0)));
 
-    // Location policy chooses worker1 first.
-    assertEquals(worker1, mBlockStore.getInStream(BLOCK_ID, options).getAddress());
-    // Location policy chooses worker2 second.
-    assertEquals(worker2, mBlockStore.getInStream(BLOCK_ID, options).getAddress());
+    BlockInStream stream = mBlockStore.getInStream(BLOCK_ID, options);
+    assertEquals(local, stream.getAddress());
+    assertEquals(DataReader.DataReaderType.GRPC, stream.getDataReaderType());
```
apc999, line 334: I feel it is fine to directly assert on `stream.getDataReader()`

apc999 (changes requested):

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileInStream.java
```diff
@@ -430,7 +430,8 @@ private void triggerAsyncCaching(BlockInStream stream) {
 
   private void handleRetryableException(BlockInStream stream, IOException e) {
     WorkerNetAddress workerAddress = stream.getAddress();
-    LOG.warn("Failed to read block {} of file {} from worker {}, will retry: {}",
+    LOG.warn("Failed to read block {} of file {} from worker {}. "
+        + "This worker will be skipped for future read operations, will retry: {}.",
         stream.getId(), mStatus.getPath(), workerAddress, e.getMessage());
```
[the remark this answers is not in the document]
LuQQiu: Good call. The related comparison between e.toString() & e.getMessage() https://stackoverflow.com/questions/28272473/java-logging-exceptions-use-getmessage-or-tostring-log-warnex-getmessage FYI @ssz1997

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -458,8 +462,8 @@ public void close() throws IOException {
   /**
    * @return whether the reader is reading data directly from a local file
    */
-  public boolean isShortCircuit() {
-    return mDataReaderFactory.isShortCircuit();
+  public DataReader.DataReaderType getDataReaderType() {
```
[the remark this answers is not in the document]
LuQQiu: Yeah, get class way is better! Changed to getClass, PTAL!

LuQQiu:
@apc999 PTAL, thanks!

apc999 (approved):

apc999:

Issue #13746, opened 2021-07-06
Flaky test in RaftJournalTest


apc999:
**Alluxio Version:**
2.7.0-SNAPSHOT

**Describe the bug**
```
Error: 9.329 [ERROR] Errors:
Error: 9.330 [ERROR] alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterSuspend
Error: 9.330 [ERROR]   Run 1: RaftJournalTest.gainPrimacyAfterSuspend:350->promoteFollower:404 » Timeout Tim...
Error: 9.330 [ERROR]   Run 2: RaftJournalTest.gainPrimacyAfterSuspend:350->promoteFollower:404 » Timeout Tim...
```

**To Reproduce**

**Expected behavior**
Unrelated change shall not fail these tests

**Urgency**
high. these flaky tests are blocking a lot of PRs

**Additional context**

apc999:
more info https://pipelines.actions.githubusercontent.com/110HvfICxw4TXSOnsd72jk32naeMsNgZF0xPcAxH5PwqdiBHUN/_apis/pipelines/1/runs/25506/signedlogcontent/4?urlExpires=2021-07-10T16%3A46%3A55.4529457Z&urlSigningMethod=HMACV1&urlSignature=t7zS2lro60IekQVYejJGj8LpsU5ZOwRrHYG8%2BalPjdI%3D

```
2021-07-10T06:55:59.3215303Z 06:55:59.310 [ERROR] alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterSuspend  Time elapsed: 32.288 s  <<< ERROR!
2021-07-10T06:55:59.3358087Z java.util.concurrent.TimeoutException: Timed out waiting for follower becomes leader options: WaitForOptions{interval=20, timeout=30000} last value: false
2021-07-10T06:55:59.3359957Z 	at alluxio.util.CommonUtils.waitForResult(CommonUtils.java:383)
2021-07-10T06:55:59.3361522Z 	at alluxio.util.CommonUtils.waitFor(CommonUtils.java:341)
2021-07-10T06:55:59.3363534Z 	at alluxio.master.journal.raft.RaftJournalTest.promoteFollower(RaftJournalTest.java:404)
2021-07-10T06:55:59.3366513Z 	at alluxio.master.journal.raft.RaftJournalTest.gainPrimacyAfterSuspend(RaftJournalTest.java:350)
[...]

PR #13779, opened 2021-07-12
Force election after switching leader modes in RaftJournalTest

ggezer:
Sending correct argument via reflection to force election upon switching election modes for RATIS nodes.

Without this change, election might not get triggered and test will timeout waiting for seeing the new leader.

Fix https://github.com/Alluxio/alluxio/issues/13746


apc999 (approved):
thanks for fixing this. Let's wait for tests passed

ggezer:

PR #13786, opened 2021-07-14
Update and improve conformance of S3 API

ZacBlanco:

Improve conformance of the S3 API to the specification defined by AWS.

The AWS CLI does not work for many standard operations (e.g. listing buckets, setting max number of return keys, copying between buckets)

This affects some of the values from the S3 API to be in-line with more of the S3 REST API specification. It can potentially break backwards compatibility with previous versions. 

### PR Summary

This PR is a a serious of small changes and improvements to the S3
API which improves the conformance to the specification defined
by the AWS Rest API reference at
https://docs.aws.amazon.com/AmazonS3/latest/API/Type_API_Reference.html

A short summary of the changes:

- Implement the CopyObjectResult type for bucket-bucket transfers
- Make the ListAllMyBucketsResult conform to the spec. Added
creationDate and proper timestamps to results
- ListBucketOptions now includes the delimiter and encoding type
parameters
- the listBucketResult uses the CommonPrefix type and now includes
the delimiter and encoding type. Additionally, make sure keyCount
and truncation takes into account the common prefixes
-  Update the S3RestServiceHandler to use Java 8 lambda language
features
- Update credentials to use the latest amazon authorization format
- A potentially breaking change which causes a ListAllMyBuckets result
to only return folders

### DIFF
--- a/core/server/proxy/src/main/java/alluxio/proxy/s3/ListAllMyBucketsResult.java
+++ b/core/server/proxy/src/main/java/alluxio/proxy/s3/ListAllMyBucketsResult.java
@@ -73,7 +73,7 @@ public class ListAllMyBucketsResult {
     }
 
     /**
-     * @return the name of the bucket
+     * @return the creation timestamp for the bucket
      */
     @JacksonXmlProperty(localName = "CreationDate")
     public String getCreationDate() {
--- a/core/server/proxy/src/main/java/alluxio/proxy/s3/ListAllMyBucketsResult.java
+++ b/core/server/proxy/src/main/java/alluxio/proxy/s3/ListAllMyBucketsResult.java
@@ -11,10 +11,14 @@
 
 package alluxio.proxy.s3;
 
+import alluxio.client.file.URIStatus;
+
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
 
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


core/server/proxy/src/main/java/alluxio/proxy/s3/ListAllMyBucketsResult.java
```diff
@@ -61,5 +71,13 @@ private Bucket(String name) {
     public String getName() {
       return mName;
     }
+
+    /**
+     * @return the name of the bucket
```
LuQQiu, line 76: Update doc?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
```diff
@@ -153,6 +173,22 @@ public String getPrefix() {
...
+  /**
+   * @return the delimiter
+   */
+  @JacksonXmlProperty(localName = "Delimiter")
+  public String getDelimiter() {
+    return mDelimiter;
+  }
+
+  /**
+   * @return the prefix
```
LuQQiu, line 185: Update the doc?

core/server/proxy/src/main/java/alluxio/proxy/s3/S3RestServiceHandler.java
```diff
@@ -153,32 +162,30 @@ private FileSystem getFileSystem(String authorization) {
   }
 
   /**
-   * @summary lists all buckets owned by you
+   * Lists all buckets owned by you
```
LuQQiu, line 165: add a .?

LuQQiu (commented):
I don't have much knowledge of the PROXY S3 API and nor do the other people in the team I think lol
Did the proxy S3 work well before your changes or it already broke?

core/server/proxy/src/main/java/alluxio/proxy/s3/ListAllMyBucketsResult.java
ZacBlanco: fixed

core/server/proxy/src/main/java/alluxio/proxy/s3/ListBucketResult.java
ZacBlanco: fixed

ggezer:
@ZacBlanco Can/Should backport these? Is there any dependency on AWS SDK that's not present on older branches?

ZacBlanco:
They probably don't need to be backported unless someone specifically needs these all applied. I mentioned that this potentially breaks backcompat for some responses formats that we previously returned. I personally don't think it is a great idea.

Otherwise though, there's nothing stopping us from backporting.

ZacBlanco:
> I don't have much knowledge of the PROXY S3 API and nor do the other people in the team I think lol
> Did the proxy S3 work well before your changes or it already broke?

The API "worked" but didn't conform to the responses that the latest AWS CLI was expecting, so it would error

ggezer (approved):

ggezer:

Xenorith:

PR #13795, opened 2021-07-15
Avoid caching invalid entity (null entiry) into memory

shenghui361:
Avoid caching invalid entity (null entiry) into memory

if entry is null，there is no need to add it to the cache. There is no point in caching a null entity, and it will be evicted soon。

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/metastore/caching/Cache.java
+++ b/core/server/master/src/main/java/alluxio/master/metastore/caching/Cache.java
@@ -176,21 +176,22 @@ public abstract class Cache<K, V> implements Closeable {
    * @param value the value
    */
   public void put(K key, V value) {
-    mMap.compute(key, (k, entry) -> {
-      onPut(key, value);
-      if (entry == null && cacheIsFull()) {
-        writeToBackingStore(key, value);
-        return null;
-      }
-      if (entry == null || entry.mValue == null) {
-        onCacheUpdate(key, value);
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/metastore/caching/Cache.java
```diff
@@ -176,21 +176,22 @@ public Cache(CacheConfiguration conf, String name, MetricKey evictionsKey, Metri
    * @param value the value
    */
   public void put(K key, V value) {
-    mMap.compute(key, (k, entry) -> {
-      onPut(key, value);
-      if (entry == null && cacheIsFull()) {
-        writeToBackingStore(key, value);
-        return null;
```
apc999, line 183: is this case what you are preventing?
according to Java doc https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#compute-K-java.util.function.BiFunction- 

> "If the function returns null, the mapping is removed (or remains absent if initially absent). If the function itself throws an (unchecked) exception, the exception is rethrown, and the current mapping is left unchanged."

ggezer: Right. Returning `null` from `compute` block is a valid operation.

We also rely on synchronization provided by `compute` method, which is removed by this change, to coordinate changes between backing store and cache.

PR #13798, opened 2021-07-15
Replace total active job count limit with a single batch job limit

yuzhu:

Before this change, the replication checker issues O(n) rpc calls to check the status of every job it submitted, resulting in a large delay.
Instead of limiting the number of active job outstanding submitted by the replication checker, we will limit the number of jobs submitted in one iteration.  This is a temporary solution until proper accounting and resource isolation is implemented on the job service. 

Free replication checker from doing many rpcs and not being able to make progress

No

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
+++ b/core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
@@ -205,9 +205,6 @@ public final class ReplicationChecker implements HeartbeatExecutor {
       if (Thread.interrupted()) {
         throw new InterruptedException("ReplicationChecker interrupted.");
       }
-      if (singleBatch >= mMaxSingleBatch) {
-        break;
-      }
       try (LockedInodePath inodePath = mInodeTree.lockFullInodePath(inodeId, LockPattern.READ)) {
         InodeFile file = inodePath.getInodeFile();
         for (long blockId : file.getBlockIds()) {
@@ -230,7 +227,11 @@ public final class ReplicationChecker implements HeartbeatExecutor {
           for (Map.Entry<String, String> entry
               : findMisplacedBlock(file, blockInfo).entrySet()) {
             try {
+              if (singleBatch >= mMaxSingleBatch) {
+                return;
+              }
               handler.migrate(inodePath.getUri(), blockId, entry.getKey(), entry.getValue());
+              singleBatch++;
             } catch (Exception e) {
               LOG.warn(
                   "Unexpected exception encountered when starting a migration job (uri={},"
@@ -249,15 +250,11 @@ public final class ReplicationChecker implements HeartbeatExecutor {
   private Set<Long> check(Set<Long> inodes, ReplicationHandler handler, Mode mode)
       throws InterruptedException {
     Set<Long> processedFileIds = new HashSet<>();
-    int singleBatch = 0;
     for (long inodeId : inodes) {
       Set<Triple<AlluxioURI, Long, Integer>> requests = new HashSet<>();
       // Throw if interrupted.
       if (Thread.interrupted()) {
         throw new InterruptedException("ReplicationChecker interrupted.");
-      }
-      if (singleBatch >= mMaxSingleBatch) {
-        break;
       }
       // TODO(binfan): calling lockFullInodePath locks the entire path from root to the target
       // file and may increase lock contention in this tree. Investigate if we could avoid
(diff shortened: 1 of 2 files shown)


core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -137,20 +130,6 @@ public void heartbeat() throws InterruptedException {
       return;
     }
 
-    final Iterator<Long> jobIterator = mActiveJobToInodeID.keySet().iterator();
-
-    while (jobIterator.hasNext()) {
```
yuzhu, line 142: This loop used to take forever and blocks replication checker essentially.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -219,17 +199,15 @@ public void close() {
 
   private void checkMisreplicated(Set<Long> inodes, ReplicationHandler handler)
       throws InterruptedException {
+    int singleBatch = 0;
```
apc999, line 202: I don't see this local variable got updated.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -273,18 +249,16 @@ private void checkMisreplicated(Set<Long> inodes, ReplicationHandler handler)
   private Set<Long> check(Set<Long> inodes, ReplicationHandler handler, Mode mode)
       throws InterruptedException {
     Set<Long> processedFileIds = new HashSet<>();
+    int singleBatch = 0;
```
apc999, line 252: same here, how is this local variable got updated?

apc999 (changes requested):
is this PR ready to review?

yuzhu:
@apc999 oh wow, clearly it was not ready for review. made the change in a hurry, now fixed.

ggezer:
@yuzhu @apc999 Is batch query capability in the plans for job service?

yuzhu:
> @yuzhu @apc999 Is batch query capability in the plans for job service?
 
In general, we want to improve the visibility of job status and allow filtering of jobs so that the user/client can get back more useful job status information rather than a dump of every job in the system.  The initial idea is to introduce group id for jobs so that we can find all jobs in a particular group and associate that with a command that the user issued  (like a distributedLoad) or a background task(like replication checker), but welcome feedback.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -252,9 +228,11 @@ private void checkMisreplicated(Set<Long> inodes, ReplicationHandler handler)
...
               : findMisplacedBlock(file, blockInfo).entrySet()) {
             try {
-              final long jobId =
-                  handler.migrate(inodePath.getUri(), blockId, entry.getKey(), entry.getValue());
-              mActiveJobToInodeID.put(jobId, inodeId);
+              if (mRequestBatchCount >= mMaxSingleBatch) {
+                return;
+              }
+              handler.migrate(inodePath.getUri(), blockId, entry.getKey(), entry.getValue());
+              mRequestBatchCount++;
```
apc999, line 235: is it possible to have race here?

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
```diff
@@ -220,12 +202,6 @@ public void close() {
   private void checkMisreplicated(Set<Long> inodes, ReplicationHandler handler)
       throws InterruptedException {
     for (long inodeId : inodes) {
-      if (mActiveJobToInodeID.size() >= mMaxActiveJobs) {
-        return;
-      }
-      if (mActiveJobToInodeID.containsValue(inodeId)) {
```
apc999, line 226: don't we check duplicated job any more?

apc999 (changes requested):

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: Replication checker is currently single threaded. Will need to consider race when we multi thread it.

core/server/master/src/main/java/alluxio/master/file/replication/ReplicationChecker.java
yuzhu: Unfortunately we.do need to. And given the current job master interface, that means checking each job for completion status to maintain the active job list. I will try to see if we can make a smaller change.here without changing job master interface, but that may not be possible

yuzhu:
addressing this issue in #13831 instead

PR #13799, opened 2021-07-16
Update load command using new cache request

jja725:
This is a first version of draft PR. Add new cache request method to blockWorker and use new method in loadCommand.
There's several TODO: add error message in cache request, add metrics tracking cache status, etc...

The load command currently uses the file streaming API to read data from the worker, which also caches the data onto the worker. This is very inefficient because the data has to be transferred to the client and waste a lot of bandwidth. Using the cache API should be more lightweight as it only instructs the worker to cache the block and not introduce any data transfer to the client.
probably add more option to Load command


shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +207,140 @@ public String getDescription() {
...
+    readFromUfs = readFromUfs || storedAsUfsBlock;
+    if (!readFromUfs) {
+      return Protocol.OpenUfsBlockOptions.getDefaultInstance();
+    }
+    BlockInfo info = status.getBlockInfo(blockId);
+    long blockStart = BlockId.getSequenceNumber(blockId) * status.getBlockSizeBytes();
+    Protocol.OpenUfsBlockOptions openUfsBlockOptions = Protocol.OpenUfsBlockOptions.newBuilder()
+        .setUfsPath(status.getUfsPath()).setOffsetInFile(blockStart).setBlockSize(info.getLength())
+        .setMaxUfsReadConcurrency(options.getMaxUfsReadConcurrency())
+        .setNoCache(!ReadType.fromProto(options.getReadType()).isCache())
```
yuzhu, line 333: This is the problem here.. the  readType was not set in the options, and NoCache is set to true..  which causes the cache function to not cache anything at all. I think it makes sense to setNoCache to false here..

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3093,6 +3093,24 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_NETWORK_CACHE_MANAGER_QUEUE_MAX =
+      new Builder(Name.WORKER_NETWORK_CACHE_MANAGER_QUEUE_MAX)
+          .setAlias(Name.WORKER_NETWORK_ASYNC_CACHE_MANAGER_QUEUE_MAX)
```
apc999, line 3098: Remove `PropertyKey.WORKER_NETWORK_ASYNC_CACHE_MANAGER_QUEUE_MAX`  since it is now replaced by `PropertyKey.WORKER_NETWORK_CACHE_MANAGER_QUEUE_MAX` ?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -87,41 +145,52 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
       }
     } else {
-      OpenFilePOptions options =
-          OpenFilePOptions.newBuilder().setReadType(ReadPType.CACHE).build();
       if (local) {
         if (!mFsContext.hasNodeLocalWorker()) {
-          System.out.println("When local option is specified,"
-              + " there must be a local worker available");
+          System.out.println(
+              "When local option is specified," + " there must be a local worker available");
```
apc999, line 162:
```suggestion
              "When local option is specified there must be a local worker available");
```

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
```diff
@@ -0,0 +1,89 @@
...
+  private FileSystemContext mFsContext;
+
+  /**
+   * Sets up all dependencies before a test runs.
+   */
+  @Before
+  public void before() throws IOException {
+    mRandom = new Random();
+    mBlockWorker = mock(DefaultBlockWorker.class);
+    mFsContext = mock(FileSystemContext.class);
```
apc999, line 48: in general, I am not a big fan of using mock to test important features.
Can you check `BlockWorkerDataReaderTest` and see if we can do similar setup to test Cache api?

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -257,11 +298,92 @@ public void run() {
...
 
+    public Boolean call() {
+      boolean result = false;
+      long blockId = mRequest.getBlockId();
+      long blockLength = mRequest.getLength();
+      try {
+        boolean isSourceLocal = mLocalWorkerHostname.equals(mRequest.getSourceHost());
+        // Check if the block has already been cached on this worker
+        if (mBlockWorker.hasBlockMeta(mRequest.getBlockId())) {
+          CACHE_DUPLICATE_REQUESTS.inc();
```
apc999, line 357: I don't think we need these metric counter to bump, it is not very useful.

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +207,140 @@ public String getDescription() {
...
+    List<BlockWorkerInfo> blockWorkerInfo = Collections.EMPTY_LIST;
+    TieredIdentity tieredIdentity =
+        TieredIdentityFactory.localIdentity(mFsContext.getClusterConf());
+    // Initial target workers to read the block given the block locations.
+    Set<WorkerNetAddress> workers;
+    // Note that, it is possible that the blocks have been written as UFS blocks
+    if (status.isPersisted() || status.getPersistenceState().equals("TO_BE_PERSISTED")) {
+      blockWorkerInfo = mFsContext.getCachedWorkers();
+      if (blockWorkerInfo.isEmpty()) {
+        throw new UnavailableException(ExceptionMessage.NO_WORKER_AVAILABLE.getMessage());
```
apc999, line 270: Let's avoid using `ExceptionMessage` . that's a bad practise causing duplicated messages hard to tell where it is from in source code. Just put your own words here

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +207,140 @@ public String getDescription() {
...
+      } catch (Exception e) {
+        System.out.printf("Failed to complete cache request for block %d of file %s: %s", mBlockId,
+            mStatus.getPath(), e.toString());
+      }
+      return false;
+    }
+  }
+
+  // todo(jianjian) consider refactoring with AlluxioBlockStore
+  private WorkerNetAddress getDataSouce(long blockId, URIStatus status, BlockLocationPolicy policy)
```
apc999, line 257: this is pretty long and complicated. better to be consolidated with `alluxio.client.block.AlluxioBlockStore#getInStream()`
e.g., making a util method called both by here and `AlluxioBlockStore#getInStream()`

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +207,140 @@ public String getDescription() {
...
+      // Instead, we unset the ufsPath field and fill in a flag ufsBlock to indicate the UFS file
+      // path can be derived from mount id and the block ID. Also because the entire file is only
+      // one block, we set the offset in file to be zero.
+      openUfsBlockOptions = openUfsBlockOptions.toBuilder().clearUfsPath().setBlockInUfsTier(true)
+          .setOffsetInFile(0).build();
+    }
+    return openUfsBlockOptions;
+  }
 }
+
```
apc999, line 346: remove empty line

apc999 (changes requested):
thanks for the improvement. I left a bunch of comments

core/common/src/main/java/alluxio/conf/PropertyKey.java
jja725: keep original property key since threadpool is only used for async cache

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
jja725: just through debug message for duplicate requests

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jja725: Since I reuse the function in AlluxioBlockStore for the loadCommand, do we still want to update the exception message?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jja725: refactor AlluxioBlockStore to expose a method called getDataSourceAndType instead of creating a util, because blockStore are suppose to provide block level operations.

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jja725: solved by creating the default option in the beginning instead.

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
jja725: After I read BlockWorkerDataReaderTest and discussed with David, it seems like all the operation can be finished within alluxio space(within one blockWorker), but our case need UFS or remote worker which are not obvious to me how to implement in the unit test. One thing I can think of right now is testing the scenario that Block already cached in the worker.

jja725:
Hey @apc999 , I added couple commits and comments to resolve the change requests. Could you pls take a look? In the meantime, I would work on adding error status to response(instead of just success field to indicate cache result) and add some metrics to track Cache information.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -106,7 +111,43 @@ public void submitRequest(AsyncCacheRequest request) {
...
+    CACHE_REQUESTS.inc();
+    long blockId = request.getBlockId();
+    if (mPendingSyncRequests.putIfAbsent(blockId, request) != null) {
+      // This block is already planned.
+      CACHE_DUPLICATE_REQUESTS.inc();
+      return false;
+    }
+    try {
+      Future<Boolean> task = mCacheExecutor.submit(new CacheTask(request));
+      return task.get();
```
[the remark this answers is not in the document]
ggezer: gRPC executor is shared across all activities in the worker. Eliminating a servicing executor and jumping to execution in the gRPC thread itself will adversely effect overall RPC stability of a worker. I suspect skipping a servicing layer here will constipate RPCs when worker is exercised at peak CPU capacity.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -67,12 +71,13 @@
...
-  public AsyncCacheRequestManager(ExecutorService service, BlockWorker blockWorker,
-      FileSystemContext fsContext) {
-    mAsyncCacheExecutor = service;
+  public CacheRequestManager(ExecutorService service, BlockWorker blockWorker,
+                             FileSystemContext fsContext) {
+    mCacheExecutor = service;
     mBlockWorker = blockWorker;
     mFsContext = fsContext;
-    mPendingRequests = new ConcurrentHashMap<>();
+    mPendingAsyncRequests = new ConcurrentHashMap<>();
```
ggezer, line 78: In terms of accounting, It shouldn't matter if it's async or sync. Why don't we just merge these two together?
And by merging these queues, we can also easily avoid async requests when there's already a sync request. And for the other direction, there's a race already with separate queues, so no change there.

core/transport/src/main/proto/grpc/block_worker.proto
```diff
@@ -122,6 +123,22 @@ message AsyncCacheRequest {
...
+  optional int32 source_port = 3;
+  optional alluxio.proto.dataserver.OpenUfsBlockOptions open_ufs_block_options = 4;
+  optional int64 length = 5;
+}
+
+// Response for an cache request
+message CacheResponse {
+  optional bool success = 1;
+}
+
```
ggezer, line 141: This new class is an exact copy of `AsyncCacheRequest`. We can rather rename the one above as `CacheRequest`, and provide async/sync distinction in `BlockWorker` method[s] alone.

core/common/src/main/java/alluxio/worker/block/BlockWorker.java
```diff
@@ -312,6 +313,14 @@ void requestSpace(long sessionId, long blockId, long additionalBytes)
    */
   void asyncCache(AsyncCacheRequest request);
 
+  /**
+   * Submits the async cache request to async cache manager to execute.
+   *
+   * @param request the async cache request
+   * @return cache result, succeed or not
+   */
+  boolean cache(CacheRequest request);
```
ggezer, line 322: I don't understand the `boolean` return value. It's already a synchronized operation. I think we should just throw `IOException` when it fails with something. It should be on client to handle it.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -257,11 +298,92 @@ public void run() {
...
+        mPendingAsyncRequests.remove(blockId);
       }
     }
   }
+  /**
+   * CacheTask is a callable task that can be considered equal if
+   * the blockId of the request is the same.
+   */
+  @VisibleForTesting
+  class CacheTask implements Callable<Boolean> {
```
ggezer, line 309: This seems like a duplicate of AsyncCacheTask. I have the same comment below.
BlockWorker already provides a distinction of async/sync. I don't see why we need to maintain two different names for the same storage format.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
ggezer: same for async variation.

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
```diff
@@ -70,11 +71,10 @@
...
   /**
-   * This executor passes impersonation information to the real worker thread.
-   * The proxy user is tracked by {@link AuthenticatedClientUser#sUserThreadLocal}.
-   * This executor delegates operations to the underlying executor while setting the
-   * ThreadLocal context for execution.
-   * */
+   * This executor passes impersonation information to the real worker thread. The proxy user is
+   * tracked by {@link AuthenticatedClientUser#sUserThreadLocal}. This executor delegates operations
+   * to the underlying executor while setting the ThreadLocal context for execution.
+   */
```
ggezer, line 77: nit: old formatting was neater.

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
```diff
@@ -0,0 +1,89 @@
...
+        .setMaxUfsReadConcurrency(10).setUfsPath("/a").build();
+    CacheRequest request = CacheRequest.newBuilder().setBlockId(blockId).setLength(blockLength)
+        .setOpenUfsBlockOptions(openUfsBlockOptions).setSourceHost(fakeRemoteWorker)
+        .setSourcePort(port).build();
+    mCacheRequestManager.submitRequest(request);
+    verify(mBlockWorker).createBlock(Sessions.ASYNC_CACHE_WORKER_SESSION_ID, blockId, 0, "",
+        blockLength);
+    verify(mBlockWorker).createBlockWriter(Sessions.ASYNC_CACHE_WORKER_SESSION_ID, blockId);
+    // since the remoteWorker is fake, we can not verify the following data writing process
+  }
```
ggezer, line 88: Could you add some tests for async cache as well?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -38,13 +73,13 @@
...
-      Option.builder()
-          .longOpt("local")
-          .required(false)
-          .hasArg(false)
-          .desc("load the file to local worker.")
-          .build();
+  // todo(jianjian) add option for timeout and parallelism(compatibility?)
+  private static final int DEFAULT_PARALLELISM = 4;
+
+  private static final Option LOCAL_OPTION = Option.builder().longOpt("local").required(false)
```
ggezer, line 67: nit: old formatting was neater. It helps to see when options are laid out over lines. 
BTW, I know our code-style file is not considerate to this, so I end up formatting line by line :(

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -87,41 +145,52 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
-      } catch (Exception e) {
-        throw closer.rethrow(e);
-      } finally {
-        closer.close();
+
+      queueLoadTasks(filePath, status, local, tasks);
+    }
+  }
+
+  private void queueLoadTasks(AlluxioURI filePath, URIStatus status, boolean local,
```
ggezer, line 163: I think this method is naive in that, if somebody wants to load a very large file-tree, `tasks` list here will grow quite large.

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -38,13 +73,13 @@
...
 public final class LoadCommand extends AbstractFileSystemCommand {
-  private static final Option LOCAL_OPTION =
-      Option.builder()
-          .longOpt("local")
-          .required(false)
-          .hasArg(false)
-          .desc("load the file to local worker.")
-          .build();
+  // todo(jianjian) add option for timeout and parallelism(compatibility?)
+  private static final int DEFAULT_PARALLELISM = 4;
```
ggezer, line 65: Are we making the `LoadCommand` concurrent? Isn't `DistributedLoad` is intended for such uses?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +207,140 @@ public String getDescription() {
...
+      mLocal = local;
+      mStatus = status;
+      mOptions = options;
+    }
+
+    @Override
+    public Boolean call() throws Exception {
+      BlockInfo info = mStatus.getBlockInfo(mBlockId);
+      long blockLength = info.getLength();
+      CacheRequest request = CacheRequest.newBuilder().setBlockId(mBlockId).setLength(blockLength)
```
ggezer, line 225: nit: lay out options over lines.

core/transport/src/main/proto/grpc/block_worker.proto
```diff
@@ -122,6 +123,22 @@ message AsyncCacheRequest {
...
+  optional int64 block_id = 1;
+  // TODO(calvin): source host and port should be replace with WorkerNetAddress
+  optional string source_host = 2;
+  optional int32 source_port = 3;
+  optional alluxio.proto.dataserver.OpenUfsBlockOptions open_ufs_block_options = 4;
+  optional int64 length = 5;
+}
+
+// Response for an cache request
+message CacheResponse {
```
ggezer, line 138: See my comment about returning `void` for this method.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -106,7 +111,43 @@ public void submitRequest(AsyncCacheRequest request) {
...
+    }
+  }
+  // TODO(jianjian): return error message
+  /**
+   * Handles a request to cache a block synchronously.
+   *
+   * @param request the cache request fields will be available
+   * @return request succeed or not
+   */
+  public boolean submitRequest(CacheRequest request) {
```
ggezer, line 123: This is almost identical to `submitAsyncRequest()`.
By merging these two into a single method, we could avoid lots of duplication.
I recommend merging these to all the way up to the `BlockWorker` interface.

core/transport/src/main/proto/grpc/block_worker.proto
ggezer: I'd also put sync/async as an additional flag into new unified `CacheRequest`. So it would be easier to manage sync/async distinction within `CacheRequestManager`.

ggezer (changes requested):
Thanks @jja725.
Left some comments. (mostly about duplication between async/sync logic.)

core/common/src/main/java/alluxio/worker/block/BlockWorker.java
jja725: I like the idea of just throw IOException, avoid lots of duplications caused by boolean return value.

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jja725: done, keep it simple and run the task one by one then

jja725:
Hey @ggezer, address comments, still need some updates for unit test, PTAL

jja725:
there are some tricky problems for the CacheRequestManagerTest. Basically, we have to do lots of mocks for remote reader and writer. But it's not obvious to me how to deal with remote source case in the Integration tests and it's heavyweight to cover it in overmind. Any thoughts? @apc999 @ggezer

ggezer (changes requested):
Thanks @jja725, duplication no longer seems to be an issue.
There's still some concern around servicing cache calls via separate thread. We can discuss offline if you'd like.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -0,0 +1,325 @@
...
+      if (result) {
+        CACHE_BLOCKS_SIZE.inc(blockLength);
+        CACHE_SUCCEEDED_BLOCKS.inc();
+      } else {
+        CACHE_FAILED_BLOCKS.inc();
+      }
+    }
+  }
+
+  private boolean cacheBlock(CacheRequest request) throws IOException, AlluxioException {
```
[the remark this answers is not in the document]
jja725: leave methods within the class for now because of the getRemoteBlockReader method.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -110,31 +114,137 @@ public void submitRequest(CacheRequest request) throws IOException, AlluxioExcep
...
+      }
+      return null;
+    }
+  }
+  /**
+   * AsyncCacheTask is a runnable task that can be considered equal if the blockId of the request is
+   * the same.
+   */
+  @VisibleForTesting
+  class AsyncCacheTask implements Runnable {
```
ggezer, line 192: Is this used anywhere? Can you just get rid of this?

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWorkerClientServiceHandler.java
```diff
@@ -149,7 +151,11 @@ public BlockWorkerClientServiceHandler(WorkerProcess workerProcess, FileSystemCo
   @Override
   public void cache(CacheRequest request, StreamObserver<CacheResponse> responseObserver) {
     RpcUtils.call(LOG, () -> {
-      mBlockWorker.cache(request);
+      try {
+        mBlockWorker.cache(request);
+      } catch (ExecutionException | InterruptedException e) {
+        throw AlluxioStatusException.fromThrowable(e);
```
ggezer, line 157: Let's do this inside BlockWorker implementation and avoid throwing from there in the first place.
Also, in case of ExecutionException, we actually want to throw its cause (`ExecutionException#getCause()`) instead.

ggezer (changes requested):
Few more cosmetic and exception related requests. Almost there.

core/client/fs/src/main/java/alluxio/client/block/stream/DefaultBlockWorkerClient.java
```diff
@@ -217,23 +216,11 @@ public ClearMetricsResponse clearMetrics(ClearMetricsRequest request) {
   public void cache(CacheRequest request) {
     boolean async = request.getAsync();
     if (async) {
```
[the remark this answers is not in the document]
jja725: I thought it would worsen readability if we get rid of this if-else since the exception handling is quite different. I update the code a little bit to make these two paths more clear. let me know if you have a better idea.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWorkerClientServiceHandler.java
jja725: since we throw alluxioException|IOException, so no need to catch exceptions in the service handler.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -68,35 +70,37 @@
...
     CACHE_REQUESTS.inc();
     long blockId = request.getBlockId();
-    if (mPendingCacheRequests.putIfAbsent(blockId, request) != null) {
+    if (mActiveCacheRequests.putIfAbsent(blockId, request) != null) {
       // This block is already planned.
       LOG.debug("request already planned: {}", request);
       return;
     }
     boolean async = request.getAsync();
     if (!async) {
```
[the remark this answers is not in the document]
jja725: see above

core/client/fs/src/main/java/alluxio/client/block/stream/DefaultBlockWorkerClient.java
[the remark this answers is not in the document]
ggezer: I don't see any exception handling for sync case here. Why do we handle for one case only?

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -67,32 +70,63 @@
...
     }
+    boolean async = request.getAsync();
+    if (!async) {
+      CACHE_REQUESTS_SYNC.inc();
+      submitSyncRequest(request, blockId);
+    } else {
+      CACHE_REQUESTS_ASYNC.inc();
+      submitASyncRequest(request, blockId);
+    }
+    mActiveCacheRequests.remove(blockId);
```
ggezer, line 104: AFAIU, this should be executed only when task is finished (completed or failed.)

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -67,32 +70,63 @@
...
+      LOG.debug("request already planned: {}", request);
       return;
     }
+    boolean async = request.getAsync();
+    if (!async) {
+      CACHE_REQUESTS_SYNC.inc();
+      submitSyncRequest(request, blockId);
+    } else {
+      CACHE_REQUESTS_ASYNC.inc();
+      submitASyncRequest(request, blockId);
```
ggezer, line 102: Submitting part is the same for both. Exception handling no need to be different until waiting for result, which is only relevant for sync request. So for both async and sync, we can still submit and handle pool rejections in a unified manner.

Once passed that, so the job is in the pool, then there can be follow up logic specific to sync calls, where you can handle ExecutionException and InterruptedException.

ggezer (changes requested):

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
jja725: I thought RejectedExecutionException handling would be different since async just log but sync cache tend to throw an exception. Updated to a unified manner

jja725:
https://github.com/Alluxio/alluxio/pull/13799#discussion_r681322268
@ggezer Doesn't know why I can't comment on this, but I think we just throw statusRuntimeException for sync case so I didn't handle that. For the Async case, we just log error since it's best effort

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
jja725: that's correct. Delete this removal. BTW, we want to add this to the end of the RejectedExecutionException, right?

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -137,11 +113,24 @@ private void submitASyncRequest(CacheRequest request, long blockId) {
               + " To increase, update the parameter '%s'. numRejected: {} error: {}",
           PropertyKey.Name.WORKER_NETWORK_ASYNC_CACHE_MANAGER_THREADS_MAX), mNumRejected.get(),
           e.toString());
-    } catch (Exception e) {
-      LOG.warn("Failed to submit async cache request. request: {}", request, e);
-      CACHE_FAILED_BLOCKS.inc();
       mActiveCacheRequests.remove(blockId);
```
ggezer, line 116: we should rethrow for sync requests here

ggezer (approved):
Thanks @jja725.

jja725:
Hi @apc999 @yuzhu @maobaolong, I think PR is in good shape right now, PTALK

yuzhu (approved):
LGTM. thanks for the improvement! @jja725

core/client/fs/src/main/java/alluxio/client/block/stream/DefaultBlockWorkerClient.java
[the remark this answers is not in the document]
apc999: IMO, it is better to have unified try-catch.
error handling for async cache should not be in this `DefaultBlockWorkerClient` but on the caller side (e.g., these InStreams). Important logic should not be put here as it gets easily ignored.

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
```diff
@@ -0,0 +1,169 @@
...
+    Protocol.OpenUfsBlockOptions openUfsBlockOptions = Protocol.OpenUfsBlockOptions.newBuilder()
+        .setMaxUfsReadConcurrency(10).setUfsPath("/a").build();
+    CacheRequest request = CacheRequest.newBuilder().setBlockId(blockId).setLength(blockLength)
+        .setOpenUfsBlockOptions(openUfsBlockOptions).setSourceHost(localWorkerHostname)
+        .setSourcePort(port).setAsync(true).build();
+    BlockReader mockBlockReader = mock(BlockReader.class);
+    when(mBlockWorker.createUfsBlockReader(Sessions.ASYNC_CACHE_UFS_SESSION_ID, blockId, 0, false,
+        openUfsBlockOptions)).thenReturn(mockBlockReader);
+    mCacheRequestManager.submitRequest(request);
+    Thread.sleep(100);  //sleep since it's async
```
apc999, line 122: `Thread.sleep` is not reliable to test async ops, check for `CommonUtils.waitFor` and its usage

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
```diff
@@ -0,0 +1,169 @@
...
+    long blockId = mRandom.nextLong();
+    long blockLength = mRandom.nextInt(5000);
+    Protocol.OpenUfsBlockOptions options = Protocol.OpenUfsBlockOptions.newBuilder()
+        .setMaxUfsReadConcurrency(10).setUfsPath("/a").build();
+    CacheRequest request = CacheRequest.newBuilder().setBlockId(blockId).setLength(blockLength)
+        .setOpenUfsBlockOptions(options).setSourceHost(fakeRemoteWorker)
+        .setSourcePort(port).setAsync(true).build();
+    setupMockReaderWriter(fakeRemoteWorker, port, blockId, blockLength, options);
+    mCacheRequestManager.submitRequest(request);
+    Thread.sleep(100); //sleep since it's async
```
apc999, line 142: ditto

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWorkerClientServiceHandler.java
```diff
@@ -147,12 +147,11 @@ public BlockWorkerClientServiceHandler(WorkerProcess workerProcess, FileSystemCo
   }
 
   @Override
-  public void asyncCache(AsyncCacheRequest request,
```
apc999, line 150: keep `asyncCache`, convert `AsyncCacheResponse` to the corresponding `CacheResponse` and passing it to `mBlockWorker.cache()`

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -38,6 +51,7 @@
 @ThreadSafe
 @PublicApi
 public final class LoadCommand extends AbstractFileSystemCommand {
+
```
apc999, line 54: nit: remove unnecessary empty line

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +171,33 @@ public String getDescription() {
...
+  private void cacheBlock(long blockId, WorkerNetAddress dataSource, boolean local,
+      URIStatus status, Protocol.OpenUfsBlockOptions options) {
+    BlockInfo info = status.getBlockInfo(blockId);
+    long blockLength = info.getLength();
+    CacheRequest request = CacheRequest.newBuilder()
+                                      .setBlockId(blockId)
+                                      .setLength(blockLength)
+                                      .setOpenUfsBlockOptions(options)
+                                      .setSourceHost(dataSource.getHost())
+                                      .setSourcePort(dataSource.getDataPort()).build();
```
apc999, line 184: nit: wiered formatting

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +171,33 @@ public String getDescription() {
...
+    // TODO(jianjian) support FUSE processLocalWorker?
```
apc999, line 185: why is FUSE processLocalWorker related here?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +171,33 @@ public String getDescription() {
...
+    WorkerNetAddress worker;
+    try {
+      if (local) {
+        // send request to local worker
+        worker = mFsContext.getNodeLocalWorker();
+      } else { // send request to data source
+        worker = dataSource;
+      }
```
apc999, line 193: will these throw exception? if not, moving them outside this try block

apc999 (changes requested):
thanks @jja725 for the improvement!
I left some comments. We are getting there!

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jja725: getNodeLocalWorker would throw IOexception so put it within the try clause.

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
jja725: remove comment

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3037,7 +3037,8 @@ public String toString() {
           .build();
   public static final PropertyKey WORKER_NETWORK_ASYNC_CACHE_MANAGER_THREADS_MAX =
       new Builder(Name.WORKER_NETWORK_ASYNC_CACHE_MANAGER_THREADS_MAX)
-          .setDefaultValue(8)
+          .setDefaultSupplier(() -> 2 * Runtime.getRuntime().availableProcessors(),
```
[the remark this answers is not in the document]
jja725: discussed with gokturk, since this will be working for both sync/async requests now and we feel like 8 is not enough.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockWorkerClientServiceHandler.java
jja725: a little confused, you mean cacheRequest here?

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
jja725: not sure how we can use waitFor here since it's using a lot of mock. Use timeout with a large number instead and it should offer similar functionality since it will immediately pass if it becomes true at any point

jja725:
Hey @apc999 , address comments, PTAL

apc999 (changes requested):

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +164,29 @@ public String getDescription() {
...
+    BlockInfo info = status.getBlockInfo(blockId);
+    long blockLength = info.getLength();
+    CacheRequest request = CacheRequest.newBuilder().setBlockId(blockId).setLength(blockLength)
+        .setOpenUfsBlockOptions(options).setSourceHost(dataSource.getHost())
+        .setSourcePort(dataSource.getDataPort()).build();
+    WorkerNetAddress worker;
+    try {
+      if (local) {
+        // send request to local worker
+        worker = mFsContext.getNodeLocalWorker();
```
apc999, line 179: isn't this already checked in line 140?

apc999 (changes requested):
a following batch of comments

core/server/worker/src/test/java/alluxio/worker/block/CacheRequestManagerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    mCacheRequestManager.submitRequest(request);
+    verify(mBlockWorker, timeout(30000)).createBlock(Sessions.ASYNC_CACHE_WORKER_SESSION_ID,
+        blockId, 0, "", blockLength);
+    verify(mBlockWorker, timeout(30000)).createBlockWriter(Sessions.ASYNC_CACHE_WORKER_SESSION_ID,
+        blockId);
+    verify(mBlockWorker, timeout(30000)).commitBlock(Sessions.ASYNC_CACHE_WORKER_SESSION_ID,
+        blockId, false);
+  }
+
+  private void setupMockReaderWriter(String source, int port, long blockId, long blockLength,
```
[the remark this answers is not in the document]
jja725: yeah agree. But like I mentioned in a previous comment, it's not obvious to me how to deal with remote host case in the Integration tests and it's heavyweight to cover it in overmind. So that's why I have to put a lot of mock here to mock the RemoteBlockReader. Is there a example for testing remote host so I can try to follow that?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -96,32 +110,44 @@ private void load(AlluxioURI filePath, boolean local) throws AlluxioException, I
...
-        throw closer.rethrow(e);
-      } finally {
-        closer.close();
+      runLoadTask(filePath, status, local);
+    }
+  }
+
+  private void runLoadTask(AlluxioURI filePath, URIStatus status, boolean local)
+      throws IOException {
+    AlluxioConfiguration conf = mFsContext.getPathConf(filePath);
```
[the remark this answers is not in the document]
jja725: as discussed with David offline, path conf make more sense here since it would overwrite clusterConf if pathConf is specified for the file

jja725:
@apc999  use real ufs for the unit test and finally be able to resolve a bug in the unit test. PTAL

tests/src/test/java/alluxio/client/hadoop/FileSystemRenameIntegrationTest.java
```diff
@@ -236,7 +236,7 @@ public void basicRenameTest7() throws Exception {
     // Due to Hadoop 1 support we stick with the deprecated version. If we drop support for it
     // FSDataOutputStream.hflush will be the new one.
     //#ifdef HADOOP1
-    o.sync();
+//    o.sync();
```
ggezer, line 239: Just commenting to make it seen.

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
```diff
@@ -0,0 +1,345 @@
...
+    long blockId = request.getBlockId();
+    boolean async = request.getAsync();
+    if (mActiveCacheRequests.putIfAbsent(blockId, request) != null) {
+      // This block is already planned and just just return.
+      if (async) {
+        LOG.debug("request already planned: {}", request);
+      } else {
+        try {
+          CommonUtils.waitFor("block to be loaded",
+              () -> mActiveCacheRequests.containsKey(blockId));
```
apc999, line 103: shall we put a timeout option here or it will block forever in `waitFor`?

shell/src/main/java/alluxio/cli/fs/command/LoadCommand.java
```diff
@@ -138,4 +163,22 @@ public String getDescription() {
...
+        .setSourcePort(dataSource.getDataPort()).build();
+    try {
+      try (CloseableResource<BlockWorkerClient> blockWorker =
+          mFsContext.acquireBlockWorkerClient(dataSource)) {
+        blockWorker.get().cache(request);
+      }
+    } catch (Exception e) {
+      System.out.printf("Failed to complete cache request for block %d of file %s: %s", blockId,
+          status.getPath(), e.toString());
+    }
```
apc999, line 182:
```suggestion
    try (CloseableResource<BlockWorkerClient> blockWorker =
        mFsContext.acquireBlockWorkerClient(dataSource)) {
      blockWorker.get().cache(request);
    } catch (Exception e) {
      System.out.printf("Failed to complete cache request for block %d of file %s: %s", blockId,
          status.getPath(), e.toString());
    }
```

apc999 (approved):
Minor comments left. Feel free to merge this PR after they are fixed.

Thanks for contributing this big PR! Good job @jja725

core/server/worker/src/main/java/alluxio/worker/block/CacheRequestManager.java
jja725: added wait option for 30s. catch a bug in the meantime, update to !mActiveCacheRequests.containsKey(blockId)

jja725:

yuzhu:

PR #13806, opened 2021-07-16
Create jar content always from scratch for shading

ggezer:
Shade plugin is provided fresh jar content always.

Incremental builds (without `clean` lifecycle) are broken due to shade-plugin reshading shaded content.

### DIFF
--- a/shaded/client/pom.xml
+++ b/shaded/client/pom.xml
@@ -119,6 +119,9 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <forceCreation>true</forceCreation>
+        </configuration>
         <executions>
           <execution>
             <id>empty-javadoc-jar</id>
@@ -146,6 +149,9 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-source-plugin</artifactId>
+        <configuration>
+          <forceCreation>true</forceCreation>
+        </configuration>
         <executions>
           <execution>
             <id>attach-sources</id>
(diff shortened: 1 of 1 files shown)


ggezer:
Not sure if this is the right approach. It just achieves the desired end goal of allowing incremental builds on source root.

ZacBlanco (approved):
So, while this might fix the problem, does it shade the jar every time no matter what? In my experience shading is actually the longest part of the build and it's bottlenecked by the speed of a single thread. I think this fix is fine, but it could also just make builds _always_ slow unless you skip building the jar entirely.

e.g. I use this command for all my builds inside of a local `Makefile`:

```
mvn -T 1C install -DskipTests -Dmaven.javadoc.skip -Dlicense.skip -Dcheckstyle.skip -Dfindbugs.skip -pl '!webui' -pl '!integration/fuse' -pl '!tests' -pl '!shaded/client'
```

Adding the `-pl '!shaded/client'` also allowed me to work around the issue when I don't need the client jar.

It may eventually be worth re-examining the build process from the ground up, and potentially migrating a more modern (and faster) build system like gradle or bazel. That's a task for the future though :)

yuzhu (approved):
LGTM

ggezer:
> So, while this might fix the problem, does it shade the jar every time no matter what? In my experience shading is actually the longest part of the build and it's bottlenecked by the speed of a single thread. I think this fix is fine, but it could also just make builds _always_ slow unless you skip building the jar entirely.
> 
> e.g. I use this command for all my builds inside of a local `Makefile`:
> 
> ```
> mvn -T 1C install -DskipTests -Dmaven.javadoc.skip -Dlicense.skip -Dcheckstyle.skip -Dfindbugs.skip -pl '!webui' -pl '!integration/fuse' -pl '!tests' -pl '!shaded/client'
> ```
> 
> Adding the `-pl '!shaded/client'` also allowed me to work around the issue when I don't need the client jar.
> 
> It may eventually be worth re-examining the build process from the ground up, and potentially migrating a more modern (and faster) build system like gradle or bazel. That's a task for the future though :)

Thanks. The issue is that incremental builds on shaded client fail always since shade-plugin requires clean build always.
I'll incorporate these flags into my build as well but will check this in to avoid failed builds.

ggezer:

Issue #13782, opened 2021-07-13
Enhance CSI - add pathType field in storage class.


Binyang2014:
Currently,  alluxio CSI will only mount directory if it present in alluxio namespace.
Add pathType to dynamic create directory if it is not present.

The valid pathType can be: `DirectoryOrCreate`, `Directory`. `Directory` is the default value.

If we use `DirectoryOrCreate` as the path type,  if the directory is not present in alluxio namespace, CSI server plugin will create it automatically. 

If we use `Directory` as the path type, if the directory is not present in alluxio namespace, error will return with some error message.

Updated storageClass config:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: alluxio
provisioner: alluxio
parameters:
  alluxioPath: /data
  pathType: DirectoryOrCreate
  javaOptions: "-Dalluxio.user.metadata.cache.enabled=true "
volumeBindingMode: Immediate
mountOptions:
  - kernel_cache
  - allow_other
  - entry_timeout=36000
  - attr_timeout=36000
  - max_readahead=0
```

Binyang2014:
@maobaolong @apc999 @LuQQiu Please help to take a look

maobaolong:
@Binyang2014 This is the feature what I really want to use.

LuQQiu:
This is a pretty useful feature! LGTM FYI @ssz1997

PR #13813, opened 2021-07-18
Enhance CSI - add pathType field in storage class

Binyang2014:
Currently,  alluxio CSI will only mount directory if it present in alluxio namespace.
Add pathType to dynamic create directory if it is not present.

The valid pathType can be: `DirectoryOrCreate`, `Directory`. `Directory` is the default value.

If we use `DirectoryOrCreate` as the path type,  if the directory is not present in alluxio namespace, CSI server plugin will create it automatically. 

If we use `Directory` as the path type, if the directory is not present in alluxio namespace, error will return with some error message.

Updated storageClass config:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: alluxio
provisioner: alluxio
parameters:
  alluxioPath: /data
  pathType: DirectoryOrCreate
  javaOptions: "-Dalluxio.user.metadata.cache.enabled=true "
volumeBindingMode: Immediate
mountOptions:
  - kernel_cache
  - allow_other
  - entry_timeout=36000
  - attr_timeout=36000
  - max_readahead=0
```
Fix https://github.com/Alluxio/alluxio/issues/13782

This PR also contains some bug fix and code refactor


ssz1997 (approved):
LGTM!

LuQQiu:
@Binyang2014 is this ready for review?

Binyang2014:
I need to do some tests for this change, will change to ready after test done

LuQQiu (approved):

LuQQiu:
@jiacheliu3 do you have time to review this PR, or we will merge it directly two days later?

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -547,4 +547,4 @@ csi:
     - entry_timeout=36000
     - attr_timeout=36000
     - max_readahead=0
-  javaOptions: "-Dalluxio.user.metadata.cache.enabled=true "
+  javaOptions: "-Dalluxio.user.metadata.cache.enabled=true"
```
jiacheliu3, line 550: nit: I don't have the full picture how this property is used, but if this does not break your code, I would recommend keeping the trailing whitespace. So later somewhere else if people append to this property they don't get an error.

integration/csi/go.sum
```diff
@@ -11,8 +11,6 @@ cloud.google.com/go/firestore v1.1.0/go.mod h1:ulACoGHTpvq5r8rxGJ4ddJZBZqakUQqCl
 cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2kNxGRt3I=
 cloud.google.com/go/storage v1.0.0/go.mod h1:IhtSnM/ZTZV8YYJWCY8RULGVqBDmpoyjwiyrjsg+URw=
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
-github.com/Alluxio/alluxio-csi v0.0.0-20210426175157-bbf121c94da5 h1:6MsrTti67sjwcni0uzS0xeY2Hjw4Ocv7wsxnV+aVYvA=
-github.com/Alluxio/alluxio-csi v0.0.0-20210426175157-bbf121c94da5/go.mod h1:CTFpOzx9k2rBgkdmxlNI+GXLlBz5AlJgm9slSKSnMCI=
```
jiacheliu3, line 15: What is this go.sum file? Do you know @LuQQiu ?

jiacheliu3 (approved):
LGTM from what i see. Thanks! @Binyang2014

integration/csi/go.sum
Binyang2014: It's generated by `go mod` (go package management tool). It contains the version number of go project dependencies. Refer: https://blog.golang.org/using-go-modules#TOC_3.

jiacheliu3: I'm not sure if this should be included into the repo, as it is auto-generated and can change every time. No one will be reviewing it as well. I'm fine with the change in this PR but in the future I recommend we find a way to get rid of it if possible.

Binyang2014: According to this section: https://blog.golang.org/using-go-modules#TOC_3.
```
The go command uses the go.sum file to ensure that future downloads of these modules retrieve the same bits as the first download, to ensure the modules your project depends on do not change unexpectedly, whether for malicious, accidental, or other reasons. Both go.mod and go.sum should be checked into version control.
```

Golang community recommends to check the file into version control. The file will be used when building go project. Otherwise, the build may use different dependencies in different environments

integration/kubernetes/helm-chart/alluxio/values.yaml
Binyang2014: Thx, will fix it,.

integration/csi/go.sum
jiacheliu3: Cool, thanks for the check!

LuQQiu:

PR #13817, opened 2021-07-19
Implement durable MEM-type worker cache

ZhuTopher:

Add a flag to the `alluxio-start.sh` and `alluxio-stop.sh` scripts to allow for persisting/populating the MEM-type worker caches to/from a local directory.

Allows for durable worker MEM-type cache during host machine restarts.

Changes the `USAGE` output from the `alluxio-start.sh` and `alluxio-stop.sh` scripts.

### DIFF
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -271,36 +271,44 @@ start_worker() {
       echo "Cache path ${cache} is not a directory; aborting"
       exit 2
     fi
+    if [[ -z $(ls -a "${cache}" ) ]]; then
+      echo "Cache path ${cache} is an empty directory; skipping cache population"
+    else 
+      echo "Populating worker ramcache(s) with contents from ${cache}"
 
-    echo "Populating worker ramcache(s) with contents from ${cache}"
+      get_ramdisk_array # see alluxio-common.sh
+      for dir in "${RAMDISKARRAY[@]}"; do
+        if [[ ! -e "${dir}" ]]; then
+          echo "Alluxio has a configured ramcache path ${dir}, but that path does not exist"
+          exit 2
+        fi
+        if [[ ! -d "${dir}" ]]; then
+          echo "Alluxio has a configured ramcache path ${dir}, but that path is not a directory"
+          exit 2
+        fi
 
-    get_ramdisk_array # see alluxio-common.sh
-    for dir in "${RAMDISKARRAY[@]}"; do
-      if [[ ! -e "${dir}" ]]; then
-        echo "Alluxio has a configured ramcache path ${dir}, but that path does not exist"
-        exit 2
-      fi
-      if [[ ! -d "${dir}" ]]; then
-        echo "Alluxio has a configured ramcache path ${dir}, but that path is not a directory"
-        exit 2
-      fi
+        echo "Populating worker ramcache at ${dir} with ${cache}/${dir}"
+        if [[ ! -e "${cache}/${dir}" ]]; then
+          echo "Path does not exist: ${cache}/${dir}"
+          exit 2
+        fi
+        if [[ ! -d "${cache}/${dir}" ]]; then
+          echo "Path is not a directory: ${cache}/${dir}"
+          exit 2
+        fi
 
-      echo "Populating worker ramcache at ${dir} with ${cache}/${dir}"
-      if [[ ! -e "${cache}/${dir}" ]]; then
-        echo "Path does not exist: ${cache}/${dir}"
-        exit 2
-      fi
-      if [[ ! -d "${cache}/${dir}" ]]; then
-        echo "Path is not a directory: ${cache}/${dir}"
-        exit 2
-      fi
+        # recursively delete all (including hidden) files of the ramdisk
+        find "${dir}" -mindepth 1 -delete
 
-      cp -a "${cache}/${dir}/." "${dir}/"
-      if [[ ${?} -ne 0 ]]; then
-        echo "Failed to populate ramcache at ${dir} with ${cache}/${dir}"
-        exit 2
-      fi
-    done
+        # recursively copy all contents of the src directory
+        # (including hidden files) to the destination directory
+        cp -dR "${cache}/${dir}/." "${dir}/"
+        if [[ ${?} -ne 0 ]]; then
+          echo "Failed to populate ramcache at ${dir} with ${cache}/${dir}"
+          exit 2
+        fi
+      done
+    fi
   fi
 
   echo "Starting worker @ $(hostname -f). Logging to ${ALLUXIO_LOGS_DIR}"
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -44,8 +44,8 @@ MOPT (Mount Option) is one of:
 
 -a         asynchronously start all processes. The script may exit before all
            processes have been started.
--c cache   populate the worker cache for each worker node from the specified
-           directory (relative to each worker node's host filesystem).
+-c cache   populate the worker ramcache(s) for each worker node from the
+           specified directory (relative to each worker node's host filesystem).
 -f         format Journal, UnderFS Data and Workers Folder on master.
 -h         display this help.
 -i backup  a journal backup to restore the master from. The backup should be
@@ -263,40 +263,43 @@ start_worker() {
   fi
 
   if [[ ! -z "${cache}" ]] ; then
+    if [[ ! -e "${cache}" ]] ; then
+      echo "Cache path ${cache} does not exist; aborting"
+      exit 2
+    fi
     if [[ ! -d "${cache}" ]] ; then
       echo "Cache path ${cache} is not a directory; aborting"
-      exit 1
+      exit 2
     fi
 
-    echo "Populating worker MEM-type caches with contents from ${cache}"
+    echo "Populating worker ramcache(s) with contents from ${cache}"
 
-    num_tiers=$(get_alluxio_property "alluxio.worker.tieredstore.levels")
-    for ((i=0;i<num_tiers;i++)); do
-      echo "Checking Worker tiered store level ${i}..."
+    get_ramdisk_array # see alluxio-common.sh
+    for dir in "${RAMDISKARRAY[@]}"; do
+      if [[ ! -e "${dir}" ]]; then
+        echo "Alluxio has a configured ramcache path ${dir}, but that path does not exist"
+        exit 2
+      fi
+      if [[ ! -d "${dir}" ]]; then
+        echo "Alluxio has a configured ramcache path ${dir}, but that path is not a directory"
+        exit 2
+      fi
 
-      tier_types=$(get_alluxio_property "alluxio.worker.tieredstore.level${i}.dirs.mediumtype")
-      tier_dirs=$(get_alluxio_property "alluxio.worker.tieredstore.level${i}.dirs.path")
-      # Use "Internal Field Separator (IFS)" variable to split strings on a
-      # delimeter and parse into an array
-      # - https://stackoverflow.com/a/918931
-      IFS=',' read -ra tier_types_arr <<< "${tier_types}"
-      IFS=',' read -ra tier_dirs_arr <<< "${tier_dirs}"
+      echo "Populating worker ramcache at ${dir} with ${cache}/${dir}"
+      if [[ ! -e "${cache}/${dir}" ]]; then
+        echo "Path does not exist: ${cache}/${dir}"
+        exit 2
+      fi
+      if [[ ! -d "${cache}/${dir}" ]]; then
+        echo "Path is not a directory: ${cache}/${dir}"
+        exit 2
+      fi
 
-      # iterate over the array elements using indices
-      # - https://stackoverflow.com/a/6723516
-      for j in "${!tier_types_arr[@]}"; do
-        tier_type=${tier_types_arr[$j]}
-        if [[ "${tier_type}" -eq "MEM" ]]; then
-          tier_dir=${tier_dirs_arr[$j]}
-
-          echo "Populating Worker tiered store at ${tier_dir} with ${cache}/tier${i}/${tier_dir}"
-          cp -a "${cache}/tier${i}/${tier_dir}/." "${tier_dir}/"
-          if [[ ${?} -ne 0 ]]; then
-            echo "Failed to copy directory from ${cache}/tier${i}/${tier_dir} to ${tier_dir}"
-            exit 2
-          fi
-        fi
-      done
+      cp -a "${cache}/${dir}/." "${dir}/"
+      if [[ ${?} -ne 0 ]]; then
+        echo "Failed to populate ramcache at ${dir} with ${cache}/${dir}"
+        exit 2
+      fi
     done
   fi
 
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -16,12 +16,12 @@
 
 USAGE="Usage: alluxio-start.sh [-hNwm] [-i backup] ACTION [MOPT] [-f] [-c cache]
 Where ACTION is one of:
-  all [MOPT]                \tStart all masters, proxies, and workers.
+  all [MOPT] [-c cache]     \tStart all masters, proxies, and workers.
   job_master                \tStart the job_master on this node.
   job_masters               \tStart job_masters on master nodes.
   job_worker                \tStart a job_worker on this node.
   job_workers               \tStart job_workers on worker nodes.
-  local [MOPT]              \tStart all processes locally.
+  local [MOPT] [-c cache]   \tStart all processes locally.
   master                    \tStart the local master on this node.
   secondary_master          \tStart the local secondary master on this node.
   masters                   \tStart masters on master nodes.
@@ -262,7 +262,12 @@ start_worker() {
     exit 1
   fi
 
-  if [[ ! -z "${cache}" && -d "${cache}" ]] ; then
+  if [[ ! -z "${cache}" ]] ; then
+    if [[ ! -d "${cache}" ]] ; then
+      echo "Cache path ${cache} is not a directory; aborting"
+      exit 1
+    fi
+
     echo "Populating worker MEM-type caches with contents from ${cache}"
 
     num_tiers=$(get_alluxio_property "alluxio.worker.tieredstore.levels")
--- a/bin/alluxio-start.sh
+++ b/bin/alluxio-start.sh
@@ -14,25 +14,25 @@
 
 #start up alluxio
 
-USAGE="Usage: alluxio-start.sh [-hNwm] [-i backup] ACTION [MOPT] [-f]
+USAGE="Usage: alluxio-start.sh [-hNwm] [-i backup] ACTION [MOPT] [-f] [-c cache]
 Where ACTION is one of:
-  all [MOPT]         \tStart all masters, proxies, and workers.
-  job_master         \tStart the job_master on this node.
-  job_masters        \tStart job_masters on master nodes.
-  job_worker         \tStart a job_worker on this node.
-  job_workers        \tStart job_workers on worker nodes.
-  local [MOPT]       \tStart all processes locally.
-  master             \tStart the local master on this node.
-  secondary_master   \tStart the local secondary master on this node.
-  masters            \tStart masters on master nodes.
-  proxy              \tStart the proxy on this node.
-  proxies            \tStart proxies on master and worker nodes.
-  safe               \tScript will run continuously and start the master if it's not running.
-  worker [MOPT]      \tStart a worker on this node.
-  workers [MOPT]     \tStart workers on worker nodes.
-  logserver          \tStart the logserver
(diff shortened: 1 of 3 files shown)


bin/alluxio-start.sh
```diff
@@ -415,11 +459,20 @@ main() {
...
+        ;;
+      c)
+        cache="${OPTARG}"
+        ;;
+      *)
+        echo -e "${USAGE}" >&2
+        exit 1
+        ;;
+    esac
+  done
```
ggezer, line 475: Could you explain what this change block is intended to do?

bin/alluxio-stop.sh
```diff
@@ -10,31 +10,28 @@
...
+  master              \tStop local primary master.
+  secondary_master    \tStop local secondary master.
+  masters             \tStop masters on master nodes.
+  proxy               \tStop local proxy.
+  proxies             \tStop proxies on master and worker nodes.
+  worker  [-c cache]  \tStop local worker.
+  workers [-c cache]  \tStop workers on worker nodes.
+  logserver           \tStop the logserver
+
+-c cache   save the worker MEM-type cache(s) from the worker node(s) to the
```
ggezer, line 32: ramcache?

bin/alluxio-stop.sh
```diff
@@ -78,11 +75,54 @@ stop_proxies() {
...
+      tier_dirs=$(get_alluxio_property "alluxio.worker.tieredstore.level${i}.dirs.path")
+      # Use "Internal Field Separator (IFS)" variable to split strings on a
+      # delimeter and parse into an array
+      # - https://stackoverflow.com/a/918931
+      IFS=',' read -ra tier_types_arr <<< "${tier_types}"
+      IFS=',' read -ra tier_dirs_arr <<< "${tier_dirs}"
+
+      # iterate over the array elements using indices
+      # - https://stackoverflow.com/a/6723516
+      for j in "${!tier_types_arr[@]}"; do
```
ggezer, line 100: How about using get_ramdisk_array() in alluxuo-common.sh ?

ggezer (changes requested):

bin/alluxio-start.sh
```diff
@@ -260,13 +262,55 @@ start_worker() {
     exit 1
   fi
 
+  if [[ ! -z "${cache}" ]] ; then
+    if [[ ! -d "${cache}" ]] ; then
+      echo "Cache path ${cache} is not a directory; aborting"
```
ZacBlanco, line 267: Does this error in the same way for the case of non-existent directory and if the specified path is a file? Would it be possible to separate those two cases?

bin/alluxio-start.sh
```diff
@@ -260,13 +262,55 @@ start_worker() {
...
+
+      # iterate over the array elements using indices
+      # - https://stackoverflow.com/a/6723516
+      for j in "${!tier_types_arr[@]}"; do
+        tier_type=${tier_types_arr[$j]}
+        if [[ "${tier_type}" -eq "MEM" ]]; then
+          tier_dir=${tier_dirs_arr[$j]}
+
+          echo "Populating Worker tiered store at ${tier_dir} with ${cache}/tier${i}/${tier_dir}"
+          cp -a "${cache}/tier${i}/${tier_dir}/." "${tier_dir}/"
```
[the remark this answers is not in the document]
ZacBlanco: Do we handle the case that the space used by the cache folder are greater than the worker's configured ramdisk size?

bin/alluxio-stop.sh
```diff
@@ -78,11 +75,54 @@ stop_proxies() {
...
+        tier_type=${tier_types_arr[$j]}
+        if [[ "${tier_type}" -eq "MEM" ]]; then
+          tier_dir=${tier_dirs_arr[$j]}
+
+          echo "Saving Worker tiered store at ${tier_dir} to ${cache}/tier${i}/${tier_dir}"
+          mkdir -p "${cache}/tier${i}/${tier_dir}"
+          cp -a "${tier_dir}/." "${cache}/tier${i}/${tier_dir}/"
```
ZacBlanco, line 107: Have you considered the case where the ramdisk doesn't fit into in the directory/disk that we're copying to?

bin/alluxio-start.sh
ZhuTopher: Sure I can make the error cases more granular

bin/alluxio-stop.sh
ZhuTopher: The assumption from the ticket [AE-1416](https://tachyonnexus.atlassian.net/browse/AE-1416) is that there is sufficient disk space to hold the ramdisk.

bin/alluxio-start.sh
ZhuTopher: It's intended just to parse the flags which come after `[MOPT]`. So in this case the old flag `-f` and the new flag + argument `-c [cache]`.

bin/alluxio-start.sh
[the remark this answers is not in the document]
ZhuTopher: @ggezer Do you mean like starting the `cp` as a separate process? I can do that but then we'd lose the ability to check on the result of the `cp` through `${?}`. Not sure if that's worth it when we expect the number of ramcaches to be low (<=3?).

@ZacBlanco No we don't. I can add a warning message if you think it's worth having.

bin/alluxio-start.sh
```diff
@@ -260,13 +262,55 @@ start_worker() {
...
+    fi
+
+    echo "Populating worker MEM-type caches with contents from ${cache}"
+
+    num_tiers=$(get_alluxio_property "alluxio.worker.tieredstore.levels")
+    for ((i=0;i<num_tiers;i++)); do
+      echo "Checking Worker tiered store level ${i}..."
+
+      tier_types=$(get_alluxio_property "alluxio.worker.tieredstore.level${i}.dirs.mediumtype")
+      tier_dirs=$(get_alluxio_property "alluxio.worker.tieredstore.level${i}.dirs.path")
```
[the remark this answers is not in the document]
ZhuTopher: The implementation of `get_ramdisk_array` in `alluxio-common.sh` makes the same assumption. I think it's fair to assume each path has its own medium type.
- Note that `alluxio-common.sh` should get an "array index out of bounds" error if `len(dirs.mediumtype) < len(dirs.path)` so it won't silently fail.

bin/alluxio-stop.sh
```diff
@@ -78,11 +75,48 @@ stop_proxies() {
 }
 
 stop_worker() {
+  if [[ ! -z "${cache}" ]] ; then
+    echo "Cache directory: ${cache}"
+    mkdir -p "${cache}"
```
ZhuTopher, line 80: @ggezer What do you think about overwriting the contents at `${cache}` if the directory exists already? Stale files in the cache would be pretty bad to accumulate via `cp`.


I would prefer to prompt the user for confirmation before overwriting directories. If we add the user prompt then we'd also need to add a `-y` flag to auto-approve so that this can still be automated too.

ZhuTopher:
@ggezer By the way, I should update the documentation surrounding this script. I was wondering if I should mention that users will likely have to use `sudo` when using `alluxio-start.sh workers NoMount -c /tmp/cache`.

I'm not 100% sure why, but in my testing I've found that to be the case. If the user does not use `sudo`, the command fails with an error cod (and 'Operation not permitted' is seen in the logs).

bin/alluxio-stop.sh
ZacBlanco: Thanks - let's assume that there is sufficient space in both cases (copy to/from). I'll resolve the other comment

bin/alluxio-stop.sh
ggezer: @madanadit WDYT?

ggezer:
> @ggezer By the way, I should update the documentation surrounding this script. I was wondering if I should mention that users will likely have to use `sudo` when using `alluxio-start.sh workers NoMount -c /tmp/cache`.
> 
> I'm not 100% sure why, but in my testing I've found that to be the case. If the user does not use `sudo`, the command fails with an error code and 'Operation not permitted' is seen in the logs.

Is this when the specified directory is owned by the user that's running alluxio process?

bin/alluxio-stop.sh
madanadit: on start, we should clear the contents of the directory (since we always make a complete snapshot) once copied to memory. on stop, the directory should be empty to begin with (under normal operation), and we should be okay to overwrite the directory (without a -y flag)

Both sequences begin with empty directory:

Sequence 1: start (no action, dir empty dir) > stop (write snapshot 1, dir non-empty) > start (clear snapshot, dir empty) > stop (write snapshot 2, dir non-empty)

Sequence 2: start (no action, dir empty dir) > start (no action, dir empty dir) > stop (write snapshot 1, dir non-empty) > stop (overwrite snapshot 1, dir non-empty)> start (clear snapshot, dir empty)

madanadit:
> @ggezer By the way, I should update the documentation surrounding this script. I was wondering if I should mention that users will likely have to use `sudo` when using `alluxio-start.sh workers NoMount -c /tmp/cache`.
> 
> I'm not 100% sure why, but in my testing I've found that to be the case. If the user does not use `sudo`, the command fails with an error code and 'Operation not permitted' is seen in the logs.

Would be good to understand the reason instead of documenting this as a limitation.

bin/alluxio-stop.sh
ggezer: Yeah I agree that we don't need a prompt unless otherwise requested.

@madanadit what should we do during start if the cache-dir is empty but worker dir is not? Should we just end up clearing the existing content, taking the cache-dir as authority?

madanadit: start.sh: (src = cache-dir, dst = worker-dir). for this direction an overwrite-if-src-not-empty makes sense. if src is empty, do not overwrite since we do not want to clear the Alluxio cache the first time a user starts using the -c option.

stop.sh: (src = worker-dir, dst = cache-dir).  for this direction an overwrite-always makes sense. if src is empty, for example if someone calls alluxio fs free on /) then we do want to clear the cache-dir as well.

ZhuTopher:
@madanadit @ggezer From what I can the 'Operation not permitted' is due to the fact that we are using `ramfs` in `alluxio-mount.sh` (`mount -t ramfs`). Apparently only trusted users (sudoers) are allowed to write to `ramfs`.

From the [Debian wiki page on ramfs](https://wiki.debian.org/ramfs):
> One downside of ramfs is you can keep writing data into it until you fill up all memory, and the VM can't free it because the VM thinks that files +should get written to backing store (rather than swap space), but ramfs hasn't got any backing store. Because of this, only root (or a trusted user) should be allowed write access to a ramfs mount.

Given that we already have the options `SudoMount` I think it's fair to require the users to use `sudo` where necessary? The most user-friendly way to resolve the issue would be to add a similar flag to `alluxio-start.sh` to prepend `sudo` to the `cp` command I think. Thoughts?

ggezer:
> @madanadit @ggezer From what I can the 'Operation not permitted' is due to the fact that we are using `ramfs` in `alluxio-mount.sh` (`mount -t ramfs`). Apparently only trusted users (sudoers) are allowed to write to `ramfs`.
> 
> From the [Debian wiki page on ramfs](https://wiki.debian.org/ramfs):
> 
> > One downside of ramfs is you can keep writing data into it until you fill up all memory, and the VM can't free it because the VM thinks that files +should get written to backing store (rather than swap space), but ramfs hasn't got any backing store. Because of this, only root (or a trusted user) should be allowed write access to a ramfs mount.
> 
> Given that we already have the options `SudoMount` I think it's fair to require the users to use `sudo` where necessary? The most user-friendly way to resolve the issue would be to add a similar flag to `alluxio-start.sh` to prepend `sudo` to the `cp` command I think. Thoughts?

Is write access here refers to manipulating mount point or putting data to it?
Also does that means Alluxio worker user should be sudoer?

@ZacBlanco Do you have any insight on this problem?

ZacBlanco:
Root access is required to *manipulate mount points* on any linux-based system. This is distinct from modifying permissions on the directories, adding files, etc within the mount point. During worker operation, to my knowledge, we never specifically need to modify the mount itself. Therefore, I don't think it's necessary to require the Alluxio user to have sudo access.

The sudo access is only required to set up the ramdisk mount point which is usually a one-time setup.

The debian man page referenced in the comment is recommending root-only access to a ramfs because whoever has write access to the ramfs generally can cause a DoS attack against the entire system by just writing tons of garbage data to the files in the ramdisk thus causing the system to OOM and either kill a bunch of processes or slow to a crawl. While this could be seen as a security hole, I still don't think it's necessary to make the worker's user sudoer because of that.

To summarize my opinion:

- Alluxio user doesn't need to be sudoer to write data to the mount point
- whoever creates the mount point does need to be sudoer, not necessarily the Alluxio user.

ZhuTopher:
Hmm let me try a few more tweaks to how I'm copying the files into the ramdisk; I might be modifying the folder itself which is why it's requiring the permissions then.

bin/alluxio-start.sh
```diff
@@ -271,36 +271,44 @@ start_worker() {
...
-      if [[ ! -e "${dir}" ]]; then
-        echo "Alluxio has a configured ramcache path ${dir}, but that path does not exist"
-        exit 2
-      fi
-      if [[ ! -d "${dir}" ]]; then
-        echo "Alluxio has a configured ramcache path ${dir}, but that path is not a directory"
-        exit 2
-      fi
+        # recursively delete all (including hidden) files of the ramdisk
+        find "${dir}" -mindepth 1 -delete
```
ZhuTopher, line 301: Added overwriting contents of ramdisk when `-c cache` is a non-empty directory

bin/alluxio-start.sh
```diff
@@ -271,36 +271,44 @@ start_worker() {
...
-
-      cp -a "${cache}/${dir}/." "${dir}/"
-      if [[ ${?} -ne 0 ]]; then
-        echo "Failed to populate ramcache at ${dir} with ${cache}/${dir}"
-        exit 2
-      fi
-    done
+        # recursively copy all contents of the src directory
+        # (including hidden files) to the destination directory
+        cp -dR "${cache}/${dir}/." "${dir}/"
```
ZhuTopher, line 305: Replaced `cp -a` with `cp -dR` since preserving file timestamps was causing 'Operation not permitted' errors. Specifically:
> `cp: preserving times for ‘/mnt/ramdisk/.’: Operation not permitted`

Linux manual page for cp: https://man7.org/linux/man-pages/man1/cp.1.html
- `cp -a` is equivalent to `cp -dR --preserve=all`

bin/alluxio-stop.sh
```diff
@@ -75,12 +75,21 @@ stop_proxies() {
...
-    if [[ ${?} -ne 0 ]]; then
-      echo "Failed to create directory: ${cache}"
-      exit 2
+    if [[ -d "${cache}" ]]; then
+      echo "Directory already exists at ${cache}; overwritting its contents"
+      # recursively delete all (including hidden) files of the ramdisk
+      find "${cache}" -mindepth 1 -delete
+    elif [[ -e "${cache}" ]]; then
+      echo "Non-directory file already exists at the path ${cache}; aborting"
+      exit 1
```
ZhuTopher, line 86: Added overwriting contents of `-c cache` directory if and only if that path is a directory

bin/alluxio-stop.sh
```diff
@@ -100,7 +109,10 @@ stop_worker() {
         echo "Failed to create directory: ${cache}/${dir}"
         exit 2
       fi
-      cp -a "${dir}/." "${cache}/${dir}/"
+
+      # recursively copy all contents of the src directory
+      # (including hidden files) to the destination directory
+      cp -dR "${dir}/." "${cache}/${dir}/"
```
ZhuTopher, line 115: Same situation here about replacing `cp -a` with `cp -dR`

ZhuTopher (commented):
@ggezer @ZacBlanco @madanadit PTAL and let me know what you think of these changes. Thanks!

bin/alluxio-start.sh
ggezer: AFAIK, we don't use timestamps for anything on the worker. So this is safe.

bin/alluxio-stop.sh
ZacBlanco: I also don't see `-mindepth` in the [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html#tag_20_47) either but it [does exist in BSD](https://www.freebsd.org/cgi/man.cgi?find(1))

bin/alluxio-start.sh
ZhuTopher: I agree about keeping this compliant across all platforms; should I just stick to `-R` then? I doubt the ramdisk directory would have symlinks in it right?

bin/alluxio-stop.sh
ZhuTopher: `-delete` doesn't exist in the POSIX spec either from what I can see (but it does for BSD). Any suggestions? I wanted to remove hidden files from the directory but not delete the directory itself (can't delete the ramdisk directory).

bin/alluxio-start.sh
ZacBlanco: I think it's probably safe to drop the `-d` flag. AFAIK we aren't using symlinks in the worker's ramdisk

bin/alluxio-stop.sh
ZhuTopher: Is using `shopt` cross-platform compliant?
```
shopt -s    dotglob
rm -rf  "${cache}/*"
shopt -u    dotglob
```

ZacBlanco: from my understanding, `find <dir> -mindepth 1 -delete` is just looking to delete all the files on the first level in the directory `${cache}`. You can achieve the same behavior with `rm -f ${cache}/*`

ZhuTopher: Actually let me look into using `rsync`. Since we are already using it in the `alluxio` binary I'm going to assume it's already installed on the host.

bin/alluxio-stop.sh
ZacBlanco: potentially remove `-d` from here too?

bin/alluxio-stop.sh
ZhuTopher: I think I can replicate the intended behaviour with `find <dir> -depth 1` and then `rm -rf` every directory/file listed there. The issue with `rm -rf <dir>/*` is the star-glob doesn't include hidden files.

ZacBlanco: > Is using shopt cross-platform compliant?

It's a bash built-in. Since we assume that bash is the shell in the scripts `shopt` is safe to use across platforms. Is there a reason to use `shopt` though? I've never had to use it. I just assumed globbing always works in the shell

ZacBlanco: nvm. I realize why we needed `dotglob`. It's to handle the hidden files

bin/alluxio-start.sh
```diff
@@ -298,11 +298,15 @@ start_worker() {
         fi
 
         # recursively delete all (including hidden) files of the ramdisk
-        find "${dir}" -mindepth 1 -delete
+        # - avoiding deleting the directory itself to avoid permission
+        #   changes/requirements on the parent directory
+        shopt -s dotglob
+        rm -rf "${dir}/"*
+        shopt -u dotglob
```
ZhuTopher, line 305: The `*` is outside the `"${dir}/"` double-quotes so the glob actually gets evaluated.

ZhuTopher, line 305: The `*` is outside the `"${dir}/"` double-quotes so the glob actually gets evaluated.

ZhuTopher, line 305: The `*` is outside the `"${dir}/"` double-quotes so the glob actually gets evaluated.

bin/alluxio-stop.sh
```diff
@@ -80,7 +80,11 @@ stop_worker() {
     if [[ -d "${cache}" ]]; then
       echo "Directory already exists at ${cache}; overwritting its contents"
       # recursively delete all (including hidden) files of the ramdisk
-      find "${cache}" -mindepth 1 -delete
+      # - avoiding deleting the directory itself to avoid permission
+      #   changes/requirements on the parent directory
+      shopt -s dotglob
+      rm -rf "${cache}/"*
+      shopt -u dotglob
```
ZhuTopher, line 87: See previous comment

ZhuTopher (commented):
@ZacBlanco Lmk what you think; I verified that this works as intended on a CentOS 7 EC2 instance

ZacBlanco (approved):

ZhuTopher:
@ggezer Does this look good to you? The failed Java 8 integration test looks like a flake?
```
java.lang.RuntimeException: Timed out waiting for process Alluxio job master @ bafdb604ddd2:36159 to start
```

ggezer (approved):

ggezer:

ggezer:

PR #13823, opened 2021-07-20
Add HiveCacheContext into the CacheContext

BlueStalker:

This PR add HiveCacheContext into CacheContext

HiveCacheContext contains the hive-related information like schema, table, partition, this information later will be used to identify whether the corresponding file needs to be cached and its caching TTL.

N/A

### DIFF
--- a/core/common/src/main/java/alluxio/client/file/CacheContext.java
+++ b/core/common/src/main/java/alluxio/client/file/CacheContext.java
@@ -162,9 +162,9 @@ public class CacheContext {
   public String toString() {
     return MoreObjects.toStringHelper(this)
         .add("cacheIdentifier", mCacheIdentifier)
-        .add("cacheHiveContext", mHiveCacheContext)
         .add("cacheQuota", mCacheQuota)
         .add("cacheScope", mCacheScope)
+        .add("hiveCacheContext", mHiveCacheContext)
         .toString();
   }
 }
--- a/core/common/src/main/java/alluxio/client/file/CacheContext.java
+++ b/core/common/src/main/java/alluxio/client/file/CacheContext.java
@@ -27,6 +27,7 @@ import javax.annotation.Nullable;
 public class CacheContext {
   /** Used in Prestodb to indicate the cache quota for a file. */
   private CacheQuota mCacheQuota = CacheQuota.UNLIMITED;
+
   /** Used in Prestodb to indicate the cache scope. */
   private CacheScope mCacheScope = CacheScope.GLOBAL;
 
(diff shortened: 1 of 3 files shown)


core/common/src/main/java/alluxio/client/file/CacheContext.java
```diff
@@ -28,6 +29,10 @@
   private CacheQuota mCacheQuota = CacheQuota.UNLIMITED;
   /** Used in Prestodb to indicate the cache scope. */
   private CacheScope mCacheScope = CacheScope.GLOBAL;
+
+  /** Used in Prestodb to indicate the hiveContext for a file. */
+  private HiveCacheContext mHiveCacheContext = null;
```
beinan, line 34: Might be out of scope of this PR, I'm just thinking mCacheQuota, mCacheScope and mHiveCacheContext should be final and also should be added as the parameter of the constructor.  @apc999 what do you think?

beinan, line 34: Add @Nullable?

core/common/src/main/java/alluxio/client/hive/HiveCacheContext.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Data structure that stores and returns Hive related Cache Scope
+ */
+public class HiveCacheContext
+{
+  private final String mDatabase;
+  private final String mTable;
+  private final String mPartition;
+
+  public HiveCacheContext(String database, String table, String partition)
```
beinan, line 26: The partition might be null I think in the hive context.  So put an annotation @Nullable?

beinan (commented):
Thank you for the contribution! just a couple of minor issues

core/common/src/main/java/alluxio/client/hive/HiveCacheContext.java
BlueStalker: sg

core/common/src/main/java/alluxio/client/file/CacheContext.java
BlueStalker: This makes sense to me and we can make default constructor

BlueStalker: it is in the getter?

apc999 (approved):

apc999:
the failing test "alluxio.client.rest.S3ClientRestApiTest.listAllMyBuckets" is unrelated and has been fixed in master branch

apc999:

PR #13827, opened 2021-07-20
Improve docker image to decrease size

ssz1997:
Use alpine instead of ubuntu1604-customize to reduce docker image size from 2.8G to 2.41G. Add and modify commands to make everything works in alpine.

### DIFF
--- a/integration/docker/Dockerfile
+++ b/integration/docker/Dockerfile
@@ -15,7 +15,7 @@
 # - https://stackoverflow.com/questions/30085621/why-does-chown-increase-size-of-docker-image
 # - https://github.com/moby/moby/issues/5505
 # - https://github.com/moby/moby/issues/6119
-FROM alpine:3.10.2 AS extractor
+FROM alpine:3.10.2
 # Note that downloads for *-SNAPSHOT tarballs are not available
 ARG ALLUXIO_TARBALL=http://downloads.alluxio.io/downloads/files/2.7.0-SNAPSHOT/alluxio-2.7.0-SNAPSHOT-bin.tar.gz
 # (Alert):It's not recommended to set this Argument to true, unless you know exactly what you are doing
@@ -32,10 +32,37 @@ RUN if [ ${ENABLE_DYNAMIC_USER} = "true" ] ; then \
        chmod -R 777 /opt/* ; \
     fi
 
-# instead of ubuntu:16.04, use alluxio/alluxio-ubuntu:ubuntu1604-customize
-# applies optimizations in libfuse on top of ubuntu:16.04
-# it is based off cheyang/fuse2:ubuntu1604-customize
-FROM alluxio/alluxio-ubuntu:1604-customize
+# The following setup is modified from cheyang/fuse2:ubuntu1604-customize to be applied on alpine
+WORKDIR /
+
+# fuse3 requires meson 0.42, available from stretch-backports
+# wget depends on ca-certificates
+
+RUN \
+  apk update && app upgrade && \
+  apk add --no-cache build-base ca-certificates pkgconfig wget eudev git && \
+  apk add --no-cache gcc make cmake gettext-dev libtool autoconf automake
+
+RUN \
+   git clone https://github.com/cheyang/libfuse.git && \
+   cd libfuse && \
+   git checkout fuse_2_9_5_customize_multi_threads_no_logging
+
+#ADD libfuse /libfuse
+
+WORKDIR "/libfuse"
+
+RUN bash makeconf.sh && \
+  ./configure && \
+  make -j8 && \
+  make install
+
+ENV LD_LIBRARY_PATH "/usr/local/lib:${LD_LIBRARY_PATH}"
+
+ENV MAX_IDLE_THREADS "64"
+
+WORKDIR /
+
 ARG ALLUXIO_USERNAME=alluxio
 ARG ALLUXIO_GROUP=alluxio
 ARG ALLUXIO_UID=1000
@@ -44,7 +71,6 @@ ARG ENABLE_DYNAMIC_USER=true
 
 # Specify versions for known vulnerabilities reported by trivy scan
 # RUN apk --no-cache --update add bash libc6-compat shadow tini 'libbz2>=1.0.6-r7' 'sqlite-libs>=3.28.0-r3' && \
-#    rm -rf /var/cache/apk/*
 
 # Add Tini
 # - https://github.com/krallin/tini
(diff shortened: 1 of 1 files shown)


apc999:
@ssz1997 @LuQQiu   with alpine, are we still able to run fuse related? e.g., I remember the original reason to move from Alpine to Customized ubuntu is to have libfuse working

ssz1997:
> @ssz1997 @LuQQiu with alpine, are we still able to run fuse related? e.g., I remember the original reason to move from Alpine to Customized ubuntu is to have libfuse working
> 
> also cheyang/fuse2:ubuntu1604-customize has a some optimization in libfuse.

I manually tested the fuse mount/unmount and it worked. I also migrated the optimization here.

LuQQiu:
@ssz1997 have you tested any file read write path? like write a file into fuse, see if it shows in alluxio namespace/UFS. Read some file from fuse, etc.

ssz1997:
> @ssz1997 have you tested any file read write path? like write a file into fuse, see if it shows in alluxio namespace/UFS. Read some file from fuse, etc.

I tried creating directory inside the alluxio mounting path - it appears both in alluxio namespace and mount point. Also tried copy a local file to the mounting path, and it can be read both from alluxio namespace and the mount point.

apc999:
@madanadit  do you have the context when we switch from alpine to ubuntu? I remember it is for libfuse, but it seems @ssz1997  is able to run fuse without using ubuntu

madanadit:
That's what I remember as well; that libfuse wasn't available on alpine. Could have changed since.

apc999:
thanks @ssz1997  for verifying it works! Good to know.
@LuQQiu , can you work with @ssz1997 to understand if the `MAX_IDLE_THREAD` macro still works ? That's the customization @cheyang added to the `cheyang/fuse2:ubuntu1604-customize`

apc999:
> That's what I remember as well; that libfuse wasn't available on alpine. Could have changed since.

double checked the PR, we actually checkedout and installed libfuse on top of Alpine -- that's why it works.

integration/docker/Dockerfile
```diff
@@ -32,10 +32,37 @@ RUN if [ ${ENABLE_DYNAMIC_USER} = "true" ] ; then \
...
+   git checkout fuse_2_9_5_customize_multi_threads_no_logging
+
+#ADD libfuse /libfuse
+
+WORKDIR "/libfuse"
+
+RUN bash makeconf.sh && \
+  ./configure && \
+  make -j8 && \
+  make install
```
apc999, line 58: after libfuse is created, can we remove all libfuse related src code and compilation tools to further reduce the image size?

maobaolong:
@ssz1997 Thanks for the work you did to decrease the size of docker image. I definitely agree with build a small size image for stable production env. But, before stable, we need a env with jdk tools like jps, jstack, we also need ping, vi, net-tools to helps us to debug unexpected behavior. Base on these, would you please maintain two types dockerfile, one for release, the other for debug. I suggest to use centos as the base image for debug version, it is widely used by lots of company in production env.

LuQQiu:
Centos docker image is even bigger than the ubuntu one. Should we offer the official testing docker image? If we should, a bigger Alluxio docker image based on either ubuntu or centos is acceptable and can be under alluxio-test/alluxio @apc999 what are your thoughts? @Xenorith what are your thoughts on centos vs ubuntu?

Xenorith:
i would propose naming the larger image as `alluxio-dev` instead of `alluxio-test`
i do not have a preference between ubuntu vs centos; this might be a better question for @madanadit given he's more familiar with production environments. i personally use centos more often and am less familiar with ubuntu, but that shouldn't affect this decision

apc999:
maybe still in [alluixo/alluxio](https://hub.docker.com/repository/docker/alluxio/alluxio) repo, but with different tag name . e.g, suffixed by `-dev` to indicate larger image size and more tools inside (based on different image). Let's discuss it tonight.

LuQQiu:
Redirect this PR to target feature branch `ssz-docker`.
This PR will target to simplify the docker image based on `alpine`
Next PR will target to add a profile to support building a docker image based on `centos`. @maobaolong can you share what changes may be needed to change the alluxio docker image to be on `centos`? i guess many install package commands are different

integration/docker/Dockerfile
ssz1997: We could, but the installation and uninstallation have to be on the same layer, which means we will have a giant RUN command, i.e. install all packages, install libfuse, uninstall all packages, all under the same RUN. Do we want to do that?

ssz1997:
@LuQQiu Uninstall some packages after their usage to further decrease the image size. The image can be gotten by `docker pull ssz1997/alluxio-alpine:latest`. The tarball used is the latest OS master tarball. PTAL, thanks!

LuQQiu:
@apc999 PTAL, thanks

ssz1997:
After some consideration and input from @Xenorith (Thanks Rico!), I feel like it doesn't worth to create a common Dockerfile-common for Dockerfile and Dockerfile-dev for the following reasons:
1. alpine image needs as few packages as possible, but centOS needs many packages. Also they use different package managements (alpine uses apk, and centOS uses yum), so the commands are quite different. Therefore, any installation involving packages, including installing libfuse, java, all the various tools, can't be put into the common dockerfile.
2. The commands for creating user are also different because alpine supports less commands. 
3. Even if we put the step that grants some directory permission to user Alluxio in Dockerfile-common, we have to copy them out  to Dockerfile and Dockerfile-dev. Anything in common that's new, we have to copy them out into the two Dockerfiles, which may still involve modify our Dockerfiles.
4. Other than these 3 steps, the Dockerfile doesn't do much. The ARG and ENV variables are not sharable across different Dockerfiles. So we don't really have much to put into Dockerfile-common.
@LuQQiu @apc999 Please let me know if you have any question.

LuQQiu:
@ssz1997 @Xenorith Based on our discussion, is it possible to put the common code in a script and then use that script in two dockerfiles?

ssz1997:
@LuQQiu 90% of the commands are actually different. The only common ones are grant permissions to Alluxio worker.

LuQQiu:
The libfuse related commands and grant permissions commands are common, is it possible to put them into a common script and trigger it in Dockerfile, or how hard it is

ssz1997:
@LuQQiu It's not hard. Now I think it does make sense to put these commands into a script. I'll work on it.

ssz1997:
@LuQQiu I suggest we merge this PR and the centOS one first, and then I'll create the PR for the common script.

LuQQiu (approved):
LGTM, thanks for improving the docker image size!

LuQQiu:

PR #13832, opened 2021-07-22
Use `mtime` in the Alluxio local cache to avoid stale caching

frankobe:

  1. change in user-facing APIs
  3. webui


core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheFileInStream.java
```diff
@@ -130,7 +130,8 @@ private int readInternal(byte[] b, int off, int len, ReadType readType, long pos
     }
     int totalBytesRead = 0;
     long currentPosition = pos;
-    long lengthToRead = Math.min(len, mStatus.getLength() - pos);
+    long lengthToRead = Math.min(len, mStatus.getLength() - mPosition);
```
beinan, line 133: why we use mPosition instead of pos?

core/client/fs/src/main/java/alluxio/client/file/cache/PageId.java
```diff
@@ -47,9 +50,16 @@ public long getPageIndex() {
     return mPageIndex;
   }
 
+  /**
+   * @return modification time
+   */
+  public long getmTime() {
```
beinan, line 56: getTime or getLastModified?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -184,9 +186,11 @@ private PageId getPageId(Path path) {
       if (!fileBucket.equals(getFileBucket(fileId))) {
         return null;
       }
-      String fileName = Preconditions.checkNotNull(matcher.group(3));
-      long pageIndex = Long.parseLong(fileName);
-      return new PageId(fileId, pageIndex);
+      String folderName = Preconditions.checkNotNull(matcher.group(3));
+      long pageIndex = Long.parseLong(folderName);
+      String fileName = Preconditions.checkNotNull(matcher.group(4));
+      long mTime = Long.parseLong(fileName);
```
beinan, line 192: is the filename here always a timestamp?

core/client/fs/src/test/java/alluxio/client/file/cache/store/LocalPageStoreTest.java
```diff
@@ -48,13 +48,31 @@ public void testPutGetDefault() throws Exception {
     helloWorldTest(pageStore);
   }
 
+  @Test
+  public void testmTime() throws Exception {
```
beinan, line 52: 'testModificationTime' might be a more meaningful name

core/client/hdfs/src/main/java/alluxio/hadoop/LocalCacheFileSystem.java
```diff
@@ -54,6 +54,7 @@
     {
       add(Constants.SCHEME);
       add("ws");
+      add("hdfs");
```
beinan, line 57: it's irrelevant to this PR

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3218,7 +3218,7 @@ public String toString() {
           .build();
   public static final PropertyKey WORKER_REVIEWER_PROBABILISTIC_HARDLIMIT_BYTES =
           new Builder(Name.WORKER_REVIEWER_PROBABILISTIC_HARDLIMIT_BYTES)
-          .setDefaultValue("64MB")
+          .setDefaultValue("1MB")
```
beinan, line 3221: why did you change this? I guess it's irrelevant to this pr.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4122,7 +4122,7 @@ public String toString() {
           .build();
   public static final PropertyKey USER_CLIENT_CACHE_PAGE_SIZE =
       new Builder(Name.USER_CLIENT_CACHE_PAGE_SIZE)
-          .setDefaultValue("1MB")
+          .setDefaultValue("64MB")
```
beinan, line 4125: Let's keep the original property setting.

beinan (commented):
Thank you @frankobe for this PR! 

I'm not sure if there is any other impact if we put an extra field in the pageId.   it might be good for us to implement the TTL.  @apc999 what do you think?  
If we're just talking about presto + alluxio use case, I might prefer to append the timestamp into the cache identifier on presto side.    Just like this PR did. https://github.com/prestodb/presto/pull/16453

core/client/hdfs/src/main/java/alluxio/hadoop/LocalCacheFileSystem.java
frankobe: > it's irrelevant to this PR

It is relevant in my opinion. Current implementation will cache `alluxio://` path & `ws://` path only, which doesnt apply for hive + hdfs use case

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
frankobe: > Thank you @frankobe for this PR!
> 
> I'm not sure if there is any other impact if we put an extra field in the pageId. it might be good for us to implement the TTL. @apc999 what do you think?
> If we're just talking about presto + alluxio use case, I might prefer to append the timestamp into the cache identifier on presto side. Just like this PR did. [prestodb/presto#16453](https://github.com/prestodb/presto/pull/16453)

We prefer to separate the modified time out to make it easier to implement TTL. The same reason applies for the presto pr. Including modificationTime in fileId makes it hard to reverse calculate the modifiedTime so we make the cache path like 

`root_path/page_size(ulong)/bucket(uint)/file_id(str)/page_idx(ulong)/mtime(long)` on purpose

beinan:
Hi @frankobe, sorry for the delay,  we're still discussing on the design of the persisted file-level metadata,  and I think I can send out a proposal and discuss with you soon.

PR #13834, opened 2021-07-22
Fix RPC counter metrics

LuQQiu:

Small fix for the RPC active counter metrics

Clean the code

No

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
@@ -87,10 +87,8 @@ public class BlockReadHandler implements StreamObserver<alluxio.grpc.ReadRequest
   private static final Logger SLOW_BUFFER_LOG = new SamplingLogger(LOG, Constants.MINUTE_MS);
   private static final long SLOW_BUFFER_MS =
       ServerConfiguration.getMs(PropertyKey.WORKER_REMOTE_IO_SLOW_THRESHOLD);
-  /** Metrics. */
   private static final Counter RPC_READ_COUNT =
-      MetricsSystem.counterWithTags(MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.getName(),
-            MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.isClusterAggregated());
+      MetricsSystem.counter(MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.getName());
 
   /** The executor to run {@link DataReader}. */
   private final ExecutorService mDataReaderExecutor;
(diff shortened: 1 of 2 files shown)


core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
```diff
@@ -87,10 +87,8 @@
   private static final Logger SLOW_BUFFER_LOG = new SamplingLogger(LOG, Constants.MINUTE_MS);
   private static final long SLOW_BUFFER_MS =
       ServerConfiguration.getMs(PropertyKey.WORKER_REMOTE_IO_SLOW_THRESHOLD);
-  /** Metrics. */
   private static final Counter RPC_READ_COUNT =
-      MetricsSystem.counterWithTags(MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.getName(),
-            MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.isClusterAggregated());
+      MetricsSystem.counter(MetricKey.WORKER_ACTIVE_RPC_READ_COUNT.getName());
```
apc999, line 91: isn't `WORKER_ACTIVE_RPC_READ_COUNT.isClusterAggregated() == true`, and thus this metrics will report to cluster? will your patch change the behavior?

LuQQiu: My patch didn't change the behavior, this metrics will be reported to master. Just that this metric don't need to use counterWithTags since there is no tags.

ZhuTopher: @LuQQiu From what I remember, the reason we used `counterWithTags` was to register the counter with the `SHOULD_REPORT_METRICS` in the `MetricsSystem`
- https://github.com/Alluxio/alluxio/blob/master/core/common/src/main/java/alluxio/metrics/MetricsSystem.java#L497-L503

Without this, the metric didn't appear in the master's log sink. @yuzhu has context on this as well.

ZhuTopher: @LuQQiu From what I remember, the reason we used `counterWithTags` was to register the counter with the `SHOULD_REPORT_METRICS` in the `MetricsSystem`
- https://github.com/Alluxio/alluxio/blob/master/core/common/src/main/java/alluxio/metrics/MetricsSystem.java#L497-L503

Without this, the metric didn't appear in the master's log sink. @yuzhu has context on this as well.

ZhuTopher (changes requested):
@yuzhu What do you think about these changes?

maobaolong:
Thanks for discuss this PR @LuQQiu @ZhuTopher @apc999 , is there any updates?

ZhuTopher:
Closing this PR as a won't-fix. If there are disputes we can re-open the PR, thanks!

PR #13836, opened 2021-07-22
Generate worker fuse helm yaml files

ssz1997:
This PR is a continuation of [https://github.com/Alluxio/alluxio/pull/13234](url)

To enable FUSE in worker container in k8s environment, the helm yaml files need to be modified. According to [https://github.com/Alluxio/alluxio/pull/13234](url), instead of directly providing ready-to-use templates, we want to auto-generate these templates with helm-generate.sh. User needs to include "--enable-worker-fuse" in command line to turn worker fuse on. 
The first commit is able to generate workerFuse container yaml files if .Values.worker.fuseEnabled is true.
The second commit is able to get the parameter whether to enable workerFuse from command line and pass it. 

Fix Alluxio/new-contributor-tasks#587


integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -188,13 +202,14 @@ data:
...
     {{- /* Format ALLUXIO_FUSE_JAVA_OPTS list to one line */}}
     {{ range $key := $fuseJavaOpts }}{{ printf "%v " $key }}{{ end }}
+  {{- end }}
   ALLUXIO_WORKER_TIEREDSTORE_LEVEL0_DIRS_PATH: /dev/shm
   {{- if .Values.logserver.enabled }}
   ALLUXIO_LOGSERVER_HOSTNAME: {{ $fullName }}-logserver
   {{- /* Note: The value must be a string or kubectl will fail to read it */}}
   ALLUXIO_LOGSERVER_PORT: {{ quote .Values.logserver.ports.logging }}
   {{- end }}
-
```
LuQQiu, line 200: this is the end of file line? every file should have an empty line at the end of the file

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -207,13 +232,23 @@ spec:
...
           {{- end }}
         {{- end }}
+        {{- if .Values.worker.fuseEnabled }}
+        - name: alluxio-fuse-device
+          hostPath:
+            path: /dev/fuse
+            type: CharDevice
+        - name: alluxio-fuse-mount
+          hostPath:
+            path: /mnt
```
LuQQiu, line 252: Using
```
        - name: alluxio-fuse-mount
          hostPath:
            path: {{ .Values.fuse.mountPath | dir }}
            type: DirectoryOrCreate
```
similar to fuse/daemonset.yaml?

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -134,6 +152,13 @@ spec:
             {{- if .Values.mounts -}}
 {{- include "alluxio.worker.otherVolumeMounts" . }}
             {{- end }}
+            {{- if eq .Values.worker.fuseEnabled true }}
+            - name: alluxio-fuse-device
+              mountPath: /dev/fuse
+            - name: alluxio-fuse-mount
+              mountPath: /mnt
```
LuQQiu, line 159: Using
```
            - name: alluxio-fuse-mount
              mountPath: {{ .Values.fuse.mountPath | dir }}
              mountPropagation: Bidirectional
```
similar to fuse/daemonset.yaml?

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -52,9 +52,9 @@ spec:
       hostNetwork: {{ $hostNetwork }}
       hostPID: {{ $hostPID }}
       dnsPolicy: {{ .Values.worker.dnsPolicy | default ($hostNetwork | ternary "ClusterFirstWithHostNet" "ClusterFirst") }}
-      {{ if .Values.hostAliases }}
+      {{- if .Values.hostAliases }}
```
LuQQiu, line 55: just curious, what's the differences with and without `-`

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -80,6 +80,18 @@
 {{- $workerJavaOpts = printf "-Dalluxio.worker.rpc.port=%v" .Values.worker.ports.rpc | append $workerJavaOpts }}
 {{- $workerJavaOpts = printf "-Dalluxio.worker.web.port=%v" .Values.worker.ports.web | append $workerJavaOpts }}
 
+{{- /* Worker Fuse configurations */}}
+{{- if eq .Values.worker.fuseEnabled true}}
+  {{- $workerJavaOpts = print "-Dalluxio.worker.fuse.enabled=true" | append $workerJavaOpts }}
+  {{- $workerJavaOpts = print "-Dalluxio.worker.fuse.mount.alluxio.path=/" | append $workerJavaOpts }}
+  {{- $workerJavaOpts = printf "-Dalluxio.worker.fuse.mount.point=%v" .Values.fuse.mountPath | append $workerJavaOpts }}
+  {{- $workerJavaOpts = print "-Dalluxio.worker.fuse.mount.options=allow_other" | append $workerJavaOpts }}
```
LuQQiu, line 88: Use .Values.fuse.mountOptions to replace a hard coded `allow_other` value?

LuQQiu (commented):
Thanks for supporting worker fuse in helm files! @jiacheliu3 PTAL, thanks

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
jiacheliu3: The `-`  at start of line will delete the whitespace on the previous line. `-` at the end of line will delete the following whitespace on the same line.
See
https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace

ssz1997: Without the two - here, the generated yaml template will have an empty line there

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -85,7 +85,9 @@
   {{- $workerJavaOpts = print "-Dalluxio.worker.fuse.enabled=true" | append $workerJavaOpts }}
   {{- $workerJavaOpts = print "-Dalluxio.worker.fuse.mount.alluxio.path=/" | append $workerJavaOpts }}
   {{- $workerJavaOpts = printf "-Dalluxio.worker.fuse.mount.point=%v" .Values.fuse.mountPath | append $workerJavaOpts }}
-  {{- $workerJavaOpts = print "-Dalluxio.worker.fuse.mount.options=allow_other" | append $workerJavaOpts }}
+  {{- range .Values.fuse.mountOptions}}
```
LuQQiu, line 88: I do not quite understand why we use .Values.fuse.mountOptions like that

LuQQiu: I would prefer to use one line for all fuse mount options like allow_other,kernel_cache,attr_timeout=9000.
and use
```
{{- $workerJavaOpts = print "-Dalluxio.worker.fuse.mount.options=.Values.fuse.mountOptions" | append $workerJavaOpts }}
```
here

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -307,9 +307,13 @@ fuse:
   user: 0
   group: 0
   fsGroup: 0
+  # fuse args specific to standalone fuse
   args:
     - fuse
     - --fuse-opts=allow_other
+  # fuse options specific to worker-embedded fuse
```
LuQQiu, line 314: fuse options specific to fuse embedded in the worker process?

LuQQiu (commented):
Left some minor comments, otherwise LGTM.

@ssz1997 A small tip for future PRs is to reply to the comments. This PR is fine, there are not that many comments. For a PR with tons of comments, it will be easier for the reviewer to know what changes you make. Or reply to the most important comments or the comments need discussion

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
ssz1997: I modified it to be how jvmOptions work. The options would look clearer as an array on multiple lines in values.yaml, just like jvmOptions. Then using concat instead of loop and append can make them all on one line. Do you think this should be the way to go?

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
jiacheliu3: We used to carefully get rid of all the extra empty lines. But later we found it's really hard to guarantee there's no extra empty lines in the generated YAMLs so we implicitly loosed the standard to "definitely avoid empty whitespace that breaks the YAML indentation and try your best to avoid unnecessary empty line".

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -207,13 +232,23 @@ spec:
 {{- include "alluxio.worker.secretVolumes" . }}
           {{- end }}
         {{- end }}
-        {{- if .Values.tieredstore }}
```
jiacheliu3, line 210: Good work removing extra whitespace

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -207,13 +232,23 @@ spec:
...
         {{- end }}
+        {{- if .Values.worker.fuseEnabled }}
+        - name: alluxio-fuse-device
+          hostPath:
+            path: /dev/fuse
+            type: CharDevice
+        - name: alluxio-fuse-mount
+          hostPath:
+            path: {{ .Values.fuse.mountPath | dir }}
+            type: DirectoryOrCreate
```
jiacheliu3, line 253: Note that `hostPath` that are created by `DirectoryOrCreate` are only writable by `root` user
Ref: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -100,6 +107,17 @@ spec:
               fieldRef:
                 fieldPath: status.podIP
           {{- end }}
+          {{- if eq .Values.worker.fuseEnabled true }}
+          - name: ALLUXIO_CLIENT_HOSTNAME
+            valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
```
jiacheliu3, line 114: I'm not sure if this should be the hostIP or the podIP, pls double check if this works, especially if this will enable the process-local read on the worker pod @LuQQiu

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -152,12 +166,14 @@
 {{- /*        ALLUXIO_FUSE_JAVA_OPTS         */}}
 {{- /* ===================================== */}}
 {{- $fuseJavaOpts := list }}
-{{- $fuseJavaOpts = print "-Dalluxio.user.hostname=${ALLUXIO_CLIENT_HOSTNAME}" | append $fuseJavaOpts }}
-{{- range $key, $val := .Values.fuse.properties }}
-  {{- $fuseJavaOpts = printf "-D%v=%v" $key $val | append $fuseJavaOpts }}
-{{- end }}
-{{- if .Values.fuse.jvmOptions }}
-  {{- $fuseJavaOpts = concat $fuseJavaOpts .Values.fuse.jvmOptions }}
+{{- if ne .Values.worker.fuseEnabled true }}
```
jiacheliu3, line 169: Would `{{- if eq .Values.worker.fuseEnabled false }}` be better?

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
```diff
@@ -188,13 +204,14 @@ data:
   ALLUXIO_JOB_WORKER_JAVA_OPTS: |-
     {{- /* Format ALLUXIO_JOB_WORKER_JAVA_OPTS list to one line */}}
     {{ range $key := $jobWorkerJavaOpts }}{{ printf "%v " $key }}{{ end }}
+  {{- if ne .Values.worker.fuseEnabled true }}
```
jiacheliu3, line 207: Same comment as above

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
LuQQiu: emmm, it should be similar to the ALLUXIO_WORKER_HOSTNAME, 
```
          env:
          - name: ALLUXIO_WORKER_HOSTNAME
            valueFrom:
              fieldRef:
                fieldPath: status.hostIP
          {{- if not .Values.worker.hostNetwork }}
          - name: ALLUXIO_WORKER_CONTAINER_HOSTNAME
            valueFrom:
              fieldRef:
                fieldPath: status.podIP
```
i saw we use hostIP for worker hostname. This works in my testing.
Not quite sure why we need a new property key `ALLUXIO_WORKER_CONTAINER_HOSTNAME` in container env?
should i create a similar `ALLUXIO_CLIENT_CONTAINER_HOSTNAME`?

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
ssz1997: @LuQQiu

LuQQiu: Emmm... by using fuse, the pod needs to run as root
```
      securityContext:
        runAsUser: 0
        runAsGroup: 0
        fsGroup: 0
```
so that it's fine that `DirectoryOrCreate` can only writable by root user.
And interestingly, the mount point is actually accessible by other users in my testing. 
I can double-check this part

ssz1997:
@LuQQiu PTAL. Thanks

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
LuQQiu: We can change to Directory here. in our testing, we create the mount folder first

integration/kubernetes/helm-generate.sh
```diff
@@ -22,7 +22,10 @@ function printUsage {
   echo
   echo "UFS is only for single-ufs mode. It should be one of:"
   echo -e " local             \t Use a local destination for UFS journal."
-  echo -e " hdfs              \t Use HDFS for UFS journal"
+  echo -e " hdfs              \t Use HDFS for UFS journal."
+  echo
+  echo "OPTS allows parameters. Currently the options are:"
```
LuQQiu, line 27: OPTS the options for generating desired YAML template. Currently...

integration/kubernetes/helm-generate.sh
```diff
@@ -22,7 +22,10 @@ function printUsage {
...
+  echo -e " --enable-workerFuse\tEnable FUSE in worker container."
```
LuQQiu, line 28: --worker-fuse Launch FUSE application in the worker process inside worker container.

integration/kubernetes/helm-chart/alluxio/templates/config/alluxio-conf.yaml
ssz1997: Isn't this because gitHub doesn't show the empty line at the end? I can see the empty line in ItelliJ.

integration/kubernetes/helm-generate.sh
```diff
@@ -199,10 +199,14 @@ function generateAllTemplates {
 
 workerFuse=false
 function main {
-  mode=$1
-  if [ "$3" = "--enable-workerFuse" ]; then
+  if [ "$3" = "--worker-fuse" ]; then
```
ssz1997, line 202: @LuQQiu Since we only have one option, so directly compare $3 to "--worker-fuse" is fine. But if we have more options, we should not do this but use an array storing all options and use switch statement inside the loop. Do you think we should do this?

ssz1997:
@LuQQiu PTAL. Thanks!

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
LuQQiu: @ssz1997 If you search `alluxio.worker.container.hostname` in Alluxio property key file, you will find that property key, but there is no such a thing as `alluxio.client.container.hostname`.
Next time if you make any change, make sure you understand why the changes are needed
e.g. why we use `ALLUXIO_WORKER_CONTAINER_HSOTNAME` `alluxio.worker.container.hostname` in alluxio code base. and then think if we should add a similar `alluxio.client.container.hostname` in the alluxio java code.
Adding `ALLUXIO_CLIENT_CONTAINER_HOSTNAME` here doesn't make any sense since no application consumes this information.

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -68,10 +68,17 @@ spec:
...
+      securityContext:
+        runAsUser: {{ .Values.fuse.user }}
+        runAsGroup: {{ .Values.fuse.group }}
+        fsGroup: {{ .Values.fuse.fsGroup }}
+    {{- else }}
       securityContext:
         runAsUser: {{ .Values.user }}
         runAsGroup: {{ .Values.group }}
         fsGroup: {{ .Values.fsGroup }}
+    {{- end }}
```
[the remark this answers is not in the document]
jiacheliu3: +1 to @ZhuTopher , let's err out when a discrepancy is spotted, instead of letting the container run into permission issues at runtime.

Also don't forget to add to the comments, as previously mentioned :)

ssz1997:
@jiacheliu3 @ZhuTopher I added a worker pod security context, and users should modify them to 0 if they want to embed Fuse in worker process. I feel like this makes more sense because in this way the worker pod configuration is getting the values from .Values.worker instead of .Values.fuse, which is used for standalone Fuse pod. PTAL, thanks!

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
```diff
@@ -68,16 +68,13 @@ spec:
...
       securityContext:
-        runAsUser: {{ .Values.user }}
-        runAsGroup: {{ .Values.group }}
-        fsGroup: {{ .Values.fsGroup }}
+        runAsUser: {{ .Values.worker.user }}
+        runAsGroup: {{ .Values.worker.group }}
+        fsGroup: {{ .Values.worker.fsGroup }}
+    {{- if eq .Values.worker.fuseEnabled true }}
+    {{- if or (or (ne .Values.worker.user 0) (ne .Values.worker.group 0)) (ne .Values.worker.fsGroup 0) }}
+    {{- fail ".Values.worker.{user, group, fsGroup} have to be 0 in order to launch Fuse in worker process" }}
```
ZhuTopher: Doesn't this implementation prevent the worker pods from using the deployment-wide defaults in `.Values.(user|group|fsGroup)`?

I'd prefer for `.Values.worker.(user|group|fsGroup)` to override `.Values.(user|group|fsGroup)`, and have the default in `values.yaml` be empty or some other way to indicate to use the global values. Eg:

```
worker:
  user:    # supply a value to override the top-level defaults
  group:   # supply a value to override the top-level defaults
  fsGroup: # supply a value to override the top-level defaults
```

```
securityContext:
  {{- if .Values.worker.user }}
  runAsUser: {{ .Values.worker.user }}
  {{- else }}
  runAsUser: {{ .Values.user }}
  {{- end }}
```

You'd also need to change your implementation of checking that the assigned securityContext is running with UID/GID `0` too.

integration/kubernetes/helm-chart/alluxio/values.yaml
```diff
@@ -228,6 +232,10 @@ worker:
   # The ServiceAccount provided here will have precedence over
   # the global `serviceAccount`
   serviceAccount:
+  # Set fuseEnabled to true will embed Fuse in worker process and
+  # escalated privileges and capabilities will be granted to the worker pod.
+  # Also be sure to give root access to the pod by setting the
+  # worker node security context values to 0 to turn on Fuse in worker.
```
ZhuTopher, line 238: nit: Phrasing/wording

```
Setting fuseEnabled to true will embed FUSE in the worker process.
The worker pods will run the Alluxio worker using privileged containers,
with the `SYS_ADMIN` capability added.
Be sure to give root access to the pod by setting the worker Pod's
securityContext values to `0`.
```

ZhuTopher (changes requested):

integration/kubernetes/helm-chart/alluxio/templates/worker/daemonset.yaml
ssz1997: Thanks for the clarification! I misunderstood what you meant yesterday.

ssz1997:
@jiacheliu3 @ZhuTopher PTAL. Thanks!

ZhuTopher (approved):
Thanks for the changes, Shawn! Let's wait for Jiacheng's review as well before merging @LuQQiu

jiacheliu3 (approved):
LGTM, thx for the improvements!

jiacheliu3:
The failing test is irrelevant to the change in the helm chart

jiacheliu3:

PR #13837, opened 2021-07-23
Add configuration of min and max election timeout

qian0817:

Add configuration of min and max election timeout.

The maximum election timeout is currently fixed at twice the minimum election timeout. Adding this configuration can improve flexibility.

remove conf  `alluxio.master.embedded.journal.election.timeout` .
add conf  `alluxio.master.embedded.journal.election.timeout.min` and  `alluxio.master.embedded.journal.election.timeout.max`.


apc999:
@ggezer ping

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -411,7 +417,8 @@ private RaftClient createClient() {
     RetryPolicy retryPolicy = ExponentialBackoffRetry.newBuilder()
         .setBaseSleepTime(TimeDuration.valueOf(100, TimeUnit.MILLISECONDS))
         .setMaxAttempts(10)
-        .setMaxSleepTime(TimeDuration.valueOf(mConf.getElectionTimeoutMs(), TimeUnit.MILLISECONDS))
+        .setMaxSleepTime(
+                TimeDuration.valueOf(mConf.getMinElectionTimeoutMs(), TimeUnit.MILLISECONDS))
```
ggezer, line 421:
```suggestion
                TimeDuration.valueOf(mConf.getMaxElectionTimeoutMs(), TimeUnit.MILLISECONDS))
```

ggezer: ?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -799,7 +806,7 @@ public synchronized void stopInternal() throws InterruptedException, IOException
           .setRpcPort(hp.getPort()).build();
 
       quorumMemberStateList.add(QuorumServerInfo.newBuilder().setServerAddress(memberAddress)
-          .setServerState(member.getLastRpcElapsedTimeMs() > mConf.getElectionTimeoutMs()
+          .setServerState(member.getLastRpcElapsedTimeMs() > mConf.getMinElectionTimeoutMs()
```
ggezer, line 809: min or max?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1700,18 +1700,27 @@ public String toString() {
...
+      new Builder(Name.MASTER_EMBEDDED_JOURNAL_MAX_ELECTION_TIMEOUT)
+          .setDescription(String.format(
+              "The max election timeout for the embedded journal. When a random period between "
+              + "${%s} and ${%s} elapses without a master receiving any messages, the master "
+              + "will attempt to become the primary Election timeout will be waited initially "
+              + "when the cluster is forming. So larger values for election timeout will cause "
+              + "longer start-up time. Smaller values might introduce instability to leadership.",
+              Name.MASTER_EMBEDDED_JOURNAL_MIN_ELECTION_TIMEOUT,
+              Name.MASTER_EMBEDDED_JOURNAL_MAX_ELECTION_TIMEOUT))
+          .setDefaultValue("20s")
```
ggezer, line 1720: @apc999 @ZacBlanco Is there a way to read value of `MASTER_EMBEDDED_JOURNAL_MIN_ELECTION_TIMEOUT` dynamically and set it here after multiplying by 2?

ggezer (commented):
Thanks @qian0817 .
Added some comments/questions.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ZacBlanco: AFAIK I don't think there's a "performant" way to make default values of properties depend upon one another. I think theoretically you could do the following in the lambda:

```java
() -> {
  AlluxioProperties props = ConfigurationUtils.defaults();
  return props.get(keyName) * 2;
}
```

This unfortunately causes the default properties hashmap on every call to `ConfigurationUtils.default()`. If we did this for more than a few properties it would become _very_ slow to start the JVM because of all the time spent in static initializers. We can probably make some kind of optimization in another PR which doesn't force a copy and returns a read-only version of the properties instead.

ggezer: Do you mind creating a ticket with some possible directions to tackle this issue?

ZacBlanco: done

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
qian0817: fix it.

ggezer:
@qian0817 Could you please drop your merge commit and instead replay your changes on top of a recent master branch?
We would like to keep the history clean by avoiding merge commits on PRs. @apc999 fyi.

ggezer (approved):
Thanks for this improvement @qian0817

ggezer:

PR #13853, opened 2021-07-26
[DOCFIX] Add examples for distributed load with host related options

maobaolong:

Modify documents, add some examples for distributed load.

https://docs.alluxio.io/os/user/edge/en/operation/User-CLI.html#distributedload

Add some examples.

Document page in https://docs.alluxio.io/os/user/edge/en/operation/User-CLI.html#distributedload


yuzhu:
@maobaolong thanks for the doc update. Could you add some examples of what locality means? and also how different options may interact with each other? For example, what if a user specify both hosts and excluded-hosts? which takes priority?

maobaolong:
@yuzhu Thanks for the suggestion, I submit new commits to explain the conflict cases, as the exclude related options cannot work with include related option, so I just left a NOTE to this document, not sure it is appropriate. PTAL.

apc999 (approved):

apc999:

PR #13856, opened 2021-07-27
Add counter for worker-side clients

apc999:

Introduce a new counter to track the number of active clients talking to a worker

As requested by community user, we want to expose the number of active clients on the worker side,
so user can decide if it is safe to remove a worker.

### Does this PR introduce any user-facing changes?

A new metrics `Worker.ActiveClients`


core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -415,6 +418,7 @@ public boolean hasBlockMeta(long blockId) {
 
   @Override
   public long lockBlock(long sessionId, long blockId) {
+    Metrics.WORKER_ACTIVE_CLIENTS.inc();
     return mLocalBlockStore.lockBlockNoException(sessionId, blockId);
```
yuzhu, line 423: should we still count the client as an active client if the lock acquisition failed? (e.g. returned invalid block id?)

yuzhu (commented):
@apc999  A high level question, should the inc/dec operation happen on the lock/unlock operation in the `BlockStore` class?  because there seems to be some more operations like moveBlock and localReader that  are unaccounted for?

yuzhu (approved):
LGTM

apc999:

PR #13869, opened 2021-07-29
Add FUSE read stress bench

ssz1997:
The FUSE stress bench only supports calculating reading throughput. The overall design and logic is similar to the existing clientIOBench and related files. The first commit has the data but not the final calculation or the result graph.

PTAL @LuQQiu, especially whether I read the files correctly. Thanks!


stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchOperation.java
```diff
@@ -0,0 +1,45 @@
...
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.stress.fuse;
+
+import alluxio.stress.client.ClientIOOperation;
+
+public enum FUSEBenchOperation {
```
LuQQiu, line 16: Same as other fuse classes, we use `Fuse` instead of `FUSE` in code.
Would prefer to name those classes FuseOperation, FuseParameters..... 
no need to add bench (short and clear name)

stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchOperation.java
```diff
@@ -0,0 +1,45 @@
...
+  /* first step read operation */
```
LuQQiu, line 17: Can be removed

stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchParameters.java
```diff
@@ -0,0 +1,55 @@
...
+          converter = FUSEBenchOperationConverter.class,
+          required = true)
+  public FUSEBenchOperation mOperation;
+
+  @Parameter(names = {"--threads"}, description = "the number of concurrent threads to use")
+  public int mThreads = 1;
+
+  @Parameter(names = {"--base"},
+          description = "The base directory path URI to perform operations in")
+
```
LuQQiu, line 34: extra line

stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchParameters.java
```diff
@@ -0,0 +1,55 @@
...
+  @Parameters.PathDescription(aliasFieldName = "mBaseAlias")
+  public String mBasePath = "/mnt/alluxio-fuse/stress-fuse-";
```
LuQQiu, line 36: emmm... could consider better name and description
like mFusePath or mLocalPath
and update the description to be something like
```
The local filesystem folder or Fuse mount point to perform operations in
```

stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchTaskResult.java
```diff
@@ -0,0 +1,202 @@
...
+    mTimeToFirstByte = new HashMap<>();
+  }
+
+  @Override
+  public TaskResult.Aggregator aggregator() {
+    return new Aggregator();
+  }
+
+  @Override
+  public GraphGenerator graphGenerator() {
```
LuQQiu, line 37: move this method under the Aggregater class
since the public method aggregator call the private inner class Aggregator

stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchTaskResult.java
```diff
@@ -0,0 +1,202 @@
...
+
+  private static final class Aggregator implements TaskResult.Aggregator<FUSEBenchTaskResult> {
+    @Override
+    public FUSEBenchTaskResult aggregate(Iterable<FUSEBenchTaskResult> results) throws Exception {
+      Iterator<FUSEBenchTaskResult> it = results.iterator();
+      if (it.hasNext()) {
+        FUSEBenchTaskResult taskResult = it.next();
+        if (it.hasNext()) {
+          throw new IOException(
+                  "ClientIO is a single node test, so multiple task results cannot be aggregated.");
```
LuQQiu, line 49: ClientIO -> Fuse

stress/common/src/main/java/alluxio/stress/fuse/FUSEBenchOperation.java
LuQQiu: Actually we can use FuseIO instead of FuseBench
same for all other classes

LuQQiu (commented):
The current code looks good, left some small questions
The next step is to support reading/writing multiple files per thread?

ssz1997:
@LuQQiu PTAL. Thanks!

LuQQiu:
@yuzhu Please help to see if the test implementation fulfills the stressbench test requirements/conventions. I will review more on the logics side (what test we want to achieve), thanks

stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
```diff
@@ -0,0 +1,73 @@
...
+  @Parameter(names = {"--threads"}, description = "the number of concurrent threads to use")
+  public List<Integer> mThreads = Collections.singletonList(2);
+
+  @Parameter(names = {"--fusePath"},
+      description = "The fuse directory in Alluxio to perform operations in")
+  public String mFusePath = "/stress-fuse/data-";
+
+  @Parameter(names = {"--localPath"},
+       description = "The local filesystem directory to perform operations in")
+  public String mLocalPath = "/mnt/alluxio-fuse/stress-fuse-";
```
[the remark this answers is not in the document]
LuQQiu: Oh, just realize that you pass in localPath as a file prefix. Users don't care about what file you generated, like what prefix you provided. Users are care about what the folder (either local folder or fuse mount point) they provided and you run tests target that folder.

Can you change to take a mLocalPath = /mnt/alluxio-fuse by default.
In Client IO case, we use
```
      int fileId = mThreadId;
      if (mParameters.mReadSameFile) {
        // all threads read the first file
        fileId = 0;
      }
      mFilePath = new Path(mParameters.mBasePath, "data-" + fileId);
```

LuQQiu (commented):
Overall the main logic is good. One thing that needs discussion is how we write the files and read them back through multi-worker multi-threads?

Ideally I want to define how many files one job can write and read.
The write job can be distributed among workers and threads in these workers.
For read, on the cluster wise, we write the files and then read the same files back.
But on the single node or single thread wise, I prefer it has the ability to 
1. read the same files as it writes 
2. read the files that are written by other nodes
This is also requested by @apc999 so that we can better test the local data and remote data read.

I am also thinking on my side, we can brainstorm how to achieve those or what we can achieve first if it's easier

stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
```diff
@@ -20,49 +20,45 @@
...
-
-  @Parameter(names = {"--fusePath"},
-      description = "The fuse directory in Alluxio to perform operations in")
-  public String mFusePath = "/stress-fuse/data-";
+  public List<Integer> mThreads = Collections.singletonList(1);
 
   @Parameter(names = {"--localPath"},
-       description = "The local filesystem directory to perform operations in")
-  public String mLocalPath = "/mnt/alluxio-fuse/stress-fuse-";
+      description = "The local filesystem directory to perform operations in")
```
LuQQiu, line 40: `The local filesystem directory or Fuse mount point`

stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
```diff
@@ -0,0 +1,73 @@
...
+  @Parameter(names = {"--read-random"},
+       description = "If true, threads read the file from random offsets. For streaming "
+           + "operations, seek() is called to read random offsets. If false, the file is read "
+           + "sequentially.")
+  @Parameters.BooleanDescription(trueDescription = "Random", falseDescription = "Sequential")
+  public boolean mReadRandom = false;
+
+  /**
+   * @return FuseIOOperation of this bench
+   * Converts from String to FuseIOOperation instance.
```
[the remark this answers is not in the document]
LuQQiu: Add line between method description and @return

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -224,14 +225,14 @@ public synchronized void mergeThreadResult(FuseIOTaskResult.ThreadCountResult th
...
+        new FuseIOTaskResult.ThreadCountResult();
 
     private BenchThread(BenchContext context, int threadId, int numThreads) {
       mContext = context;
       mThreadId = threadId;
       mFilesPath = new ArrayList<>();
-      for (int i = mThreadId; i < 10; i += numThreads) {
-        mFilesPath.add(mParameters.mLocalPath + i);
+      for (int i = mThreadId; i < 1000; i += numThreads) {
+        mFilesPath.add(mParameters.mLocalPath + "/fuseIOBench/data-" + i);
```
LuQQiu, line 235: I dont understand what we are doing here? 1000 threads?

LuQQiu: Oh i got it... nvm

LuQQiu:
Overall LGTM for the Fuse stress read, please add the write path and validate your changes locally.
An extended goal maybe when you done with the previous two is to see what happens when you use cluster mode with 1 master + worker (fuse mounted in this worker node) and 1 master + 2 workers (fuse mounted in the worker nodes in same localpath)

yuzhu (approved):
LGTM as far as stressbench structure is concerned.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+      }
+    }
+
+    private long applyOperation(String filePath) throws IOException {
+      if (FuseIOOperation.isRead(mParameters.mOperation) && mInStream == null) {
+        mInStream = new FileInputStream(filePath);
+      }
+      switch (mParameters.mOperation) {
+        case READ: {
```
LuQQiu, line 326: Keep read and write operation the same pattern in the beginning.
If write writes the whole file, then read reads the whole file.
If write writes up to buffer size, then read reads up to buffer size.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -0,0 +1,365 @@
...
+            "Thread missed barrier. Set the start time to a later time. start: %d current: %d",
+            mContext.getStartMs(), CommonUtils.getCurrentMs()));
+      }
+      CommonUtils.sleepMs(waitMs);
+      mStartBarrierPassed = true;
+
+      for (int i = 0; i < mFilesPath.size(); i++) {
+        String filePath = mFilesPath.get(i);
+        while (!Thread.currentThread().isInterrupted() && (!isRead
+            || CommonUtils.getCurrentMs() < mContext.getEndMs())) {
```
LuQQiu, line 295: The logic here is a little complicated ... Fix the next comment and then see how to simplify the logics here

LuQQiu: if CommonUtils.getCurrentMs() < mContext.getEndMs() for read, then we should break the big for loop

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
LuQQiu: or we can have readArray and readFully operations

ssz1997: They now read/write up to the buffer size, no whole file anymore (for now). The reading loop is by calling `applyOpeartion()` and the writing loop is inside `applyOpeartion()`, which makes it a little confusing. I have to keep a current offset for writing but not for reading and that's why they are separated. I may have to move the time check (getCurrentMS vs getEndMs) inside `applyOpeartion()` to make the logic more straightforward.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: > if CommonUtils.getCurrentMs() < mContext.getEndMs() for read, then we should break the big for loop

Thanks for pointing this out!

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -0,0 +1,365 @@
...
+          mOutStream = new FileOutputStream(filePath, true);
+          long currentOffset = 0;
+          while (true) {
+            int bytesToWrite = (int) Math.min(mFileSize - currentOffset, mBuffer.length);
+            if (bytesToWrite == 0) {
+              mOutStream.close();
+              mOutStream = null;
+              break;
+            }
+            mOutStream.write(mBuffer);
```
LuQQiu, line 343: use `mOutStream.write(mBuffer, 0, bytesToWrite)`, otherwise if we wants to write 20 bytes, the buffer is 12 bytes, we will eventually write 24 bytes instead of 20

LuQQiu: Always remember to close the input/output stream when you done using it!
you can use try (FileOutStream out = .....) {} try with resources to make sure no matter what happens when writing the file, the file out stream will always be closed

LuQQiu: No need to use mOutStream as a member variable here

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
LuQQiu: Oh, if that's the case we can do write and read logic separately to make the for & while loop logics easier

ssz1997 (commented):
Make mCurrentOffset a member variable is for reading from specific position in the future. Adding closeOutStream is for properly close outStream in input throughput test in the future

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    @Override
+    public Void call() {
+      try {
+        runInternal();
+      } catch (Exception e) {
+        LOG.error(Thread.currentThread().getName() + ": failed", e);
+        mThreadCountResult.addErrorMessage(e.getMessage());
+      }
```
LuQQiu, line 269: add a finally block here to make sure if any exception is thrown, we close the opened fileInStream and FileOutStream

stress/common/src/main/java/alluxio/stress/fuse/FuseIOOperation.java
```diff
@@ -0,0 +1,64 @@
...
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.stress.fuse;
+
+/**
+ * The operations for the FUSE stress tests.
+ */
+public enum FuseIOOperation {
+  /** The write operation, for writing data for the read operations. */
```
LuQQiu, line 18: Write operation to test the write throughput or prepare data for reading

stress/common/src/main/java/alluxio/stress/fuse/FuseIOOperation.java
```diff
@@ -0,0 +1,64 @@
...
+  WRITE("Write"),
+
+  /** Currently, only streaming reading is supported, that is, reading the whole file. */
```
LuQQiu, line 21: Sequentially read the written files.

LuQQiu (approved):
Overall LGTM, left some minor comments

stress/common/src/main/java/alluxio/stress/fuse/FuseIOOperation.java
ssz1997: Depending on how we are going to do with testing write throughput, we may be separate the data preparation part with this write, but this is for the next step

ssz1997:
@LuQQiu PTAL. Thanks for all the comments!

stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
```diff
@@ -28,7 +28,7 @@
   public static final String FIELD_READ_RANDOM = "mReadRandom";
 
   @Parameter(names = {"--operation"},
-      description = "the operation to perform. Options are [Read, Write]",
+      description = "the operation to perform. Options are [Read]",
```
LuQQiu, line 31: why remove the write operation?

ssz1997: Because currently we don't support a fully-functional write throughput test

LuQQiu: but i can still use --operation write to write the files for the following read tests?

ssz1997: Ah yes. I did this mainly because the ClientIO stress bench doesn't include write as an option in the description but it could write data for reading tests. I honestly don't know which makes more sense, include it or not..

LuQQiu: we can keep that for now and see how to address later.

LuQQiu:

PR #13891, opened 2021-08-03
[DOCFIX] Improve documentation surrounding worker tiered stores

ZhuTopher:

- Add section to 'Upgrading' doc about worker ramdisk cache persistence
- Improve descriptions to various `alluxio.worker.tieredstore` configuration options

Update docs to more accurately reflect the state of the application

- Changes the 'Upgrading' doc
- Changes the 'List of Configuration Properties' doc

### DIFF
--- a/bin/alluxio-common.sh
+++ b/bin/alluxio-common.sh
@@ -36,7 +36,8 @@ function get_alluxio_property() {
   echo "${property}"
 }
 
-# Generates an array of ramdisk paths in global variable RAMDISKARRAY 
+# Generates an array of ramdisk paths in global variable RAMDISKARRAY
+# - Only examines the top-tier worker cache level for "MEM"-type paths
 function get_ramdisk_array() {
   local tier_path=$(get_alluxio_property "alluxio.worker.tieredstore.level0.dirs.path")
   local medium_type=$(get_alluxio_property "alluxio.worker.tieredstore.level0.dirs.mediumtype")
--- a/bin/alluxio-common.sh
+++ b/bin/alluxio-common.sh
@@ -37,7 +37,7 @@ function get_alluxio_property() {
 }
 
 # Generates an array of ramdisk paths in global variable RAMDISKARRAY
-# - Only examines the top-tier worker cache level for "MEM"-type paths
+# - Only examines level0 of the tiered store for "MEM"-type paths
 function get_ramdisk_array() {
   local tier_path=$(get_alluxio_property "alluxio.worker.tieredstore.level0.dirs.path")
   local medium_type=$(get_alluxio_property "alluxio.worker.tieredstore.level0.dirs.mediumtype")
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3306,8 +3306,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
       new Builder(Template.WORKER_TIERED_STORE_LEVEL_DIRS_PATH, 0)
           .setDefaultSupplier(() -> OSUtils.isLinux() ? "/mnt/ramdisk" : "/Volumes/ramdisk",
               "/mnt/ramdisk on Linux, /Volumes/ramdisk on OSX")
-          .setDescription("A comma-separated list of paths (eg., /mnt/ramdisk1,/mnt/ramdisk2,/dev/sda1) "
-              + "of storage directories for the top storage tier. "
+          .setDescription("A comma-separated list of paths (eg., /mnt/ramdisk1,/mnt/ramdisk2,"
+              + "/mnt/ssd/alluxio/cache1) of storage directories for the top storage tier. "
               + "Note that for MacOS, the root directory should be `/Volumes/` and not `/mnt/`.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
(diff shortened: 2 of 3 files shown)


bin/alluxio-common.sh
```diff
@@ -36,7 +36,8 @@ function get_alluxio_property() {
   echo "${property}"
 }
 
-# Generates an array of ramdisk paths in global variable RAMDISKARRAY 
+# Generates an array of ramdisk paths in global variable RAMDISKARRAY
+# - Only examines the top-tier worker cache level for "MEM"-type paths
```
ZhuTopher, line 40: My current understanding is that there is no reason why we _couldn't_ specify `MEM` as the medium type for lower tiers, other than this function `get_ramdisk_array()` not checking for them.
- As a result, `alluxio-mount.sh` would not mount those ramdisk paths in lower tiers

docs/en/operation/Upgrade.md
```diff
@@ -59,3 +59,26 @@ and then start all the masters with the local backup file path.
 Alluxio 2.x makes significant changes in the RPC layer,
 so pre-2.0.0 clients do not work with post-2.0.0 servers, and vice-versa.
 Upgrade all applications to use the alluxio-2.x client.
+
+## Additional Options
+
+### Alluxio worker ramdisk cache persistence
```
ZhuTopher, line 65: This section is being added to reflect the additions added in https://github.com/Alluxio/alluxio/pull/13817

bin/alluxio-common.sh
ZacBlanco: Can your documentation be a little more specific since this is just a comment. e.g.
```suggestion
# - Only examines level0 of the tiered store for "MEM"-type paths
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3306,8 +3306,9 @@ public String toString() {
       new Builder(Template.WORKER_TIERED_STORE_LEVEL_DIRS_PATH, 0)
           .setDefaultSupplier(() -> OSUtils.isLinux() ? "/mnt/ramdisk" : "/Volumes/ramdisk",
               "/mnt/ramdisk on Linux, /Volumes/ramdisk on OSX")
-          .setDescription("The path of storage directory for the top storage tier. Note "
-              + "for MacOS the value should be `/Volumes/`.")
+          .setDescription("A comma-separated list of paths (eg., /mnt/ramdisk1,/mnt/ramdisk2,/dev/sda1) "
```
ZacBlanco, line 3309: `/dev/sda1` is a raw block device identifier and is _not_ compatible as a location to store data from the alluxio worker (AFAIK). The path needs to be a device formatted with some filesystem on it. I would suggest just removing this one from the examples or changing it to something else.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3325,7 +3326,13 @@ public String toString() {
...
       new Builder(Template.WORKER_TIERED_STORE_LEVEL_DIRS_QUOTA, 0)
           .setDefaultValue(String.format("${%s}", Name.WORKER_RAMDISK_SIZE))
-          .setDescription("The capacity of the top storage tier.")
+          .setDescription(String.format(
+              "A comma-separated list of capacities (e.g., \"500MB,500MB,5GB\") for each storage "
+                  + "directory on the top storage tier specified by %s. "
+                  + "For any \"MEM\"-type media (i.e, the ramdisks), this value should be set "
+                  + "equivalent to the value specified by %s.",
+              PropertyKey.WORKER_TIERED_STORE_LEVEL0_DIRS_PATH.mName,
+              PropertyKey.WORKER_RAMDISK_SIZE.mName))
```
ZacBlanco, line 3335: Prefer `Name.WORKER_RAMDISK_SIZE` here instead since I believe that what we use elsewhere. the `LEVEL0_DIRS_PATH` PK doesn't have a `Name.*` because it's a template property so that one is ok

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3354,7 +3361,8 @@ public String toString() {
           .build();
   public static final PropertyKey WORKER_TIERED_STORE_LEVEL1_DIRS_PATH =
       new Builder(Template.WORKER_TIERED_STORE_LEVEL_DIRS_PATH, 1)
-          .setDescription("The path of storage directory for the second storage tier.")
+          .setDescription("A comma-separated list of paths (eg., /dev/sda2,/dev/sda3,/dev/sdb1) "
+              + "of storage directories for the second storage tier. ")
```
ZacBlanco, line 3365: same comment about using `/dev/*` devices for the worker

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3400,7 +3411,8 @@ public String toString() {
           .build();
   public static final PropertyKey WORKER_TIERED_STORE_LEVEL2_DIRS_PATH =
       new Builder(Template.WORKER_TIERED_STORE_LEVEL_DIRS_PATH, 2)
-          .setDescription("The path of storage directory for the third storage tier.")
+          .setDescription("A comma-separated list of paths (eg., /dev/sda4,/dev/sdb2,/dev/sdb3) "
```
ZacBlanco, line 3414: `/dev/*` comment again

ZacBlanco (changes requested):

ZacBlanco (approved):

ggezer:

PR #13892, opened 2021-08-04
Support Web server for fuse process

maobaolong:

Support FUSE Web server, leverage this, we can get stacks, metrics restfulapi.

We want to monitor FUSE server by prometheus, and get the thread stacks of FUSE in the web browser.

Bring webserver to FUSE server.

If you want open fuse web server, set `alluxio.fuse.web.enabled=true` is necessary.

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
@@ -18,12 +18,14 @@ import alluxio.client.file.FileSystemContext;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.conf.InstancedConfiguration;
 import alluxio.conf.PropertyKey;
+import alluxio.conf.ServerConfiguration;
 import alluxio.jnifuse.FuseException;
 import alluxio.metrics.MetricsSystem;
 import alluxio.retry.RetryUtils;
 import alluxio.util.CommonUtils;
 import alluxio.util.io.FileUtils;
 
+import alluxio.util.network.NetworkAddressUtils;
 import com.google.common.base.Preconditions;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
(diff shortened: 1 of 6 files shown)


maobaolong:
@LuQQiu Please take a look at this PR, thank you.

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
```diff
@@ -119,6 +121,14 @@ public static void main(String[] args) {
     }
     CommonUtils.PROCESS_TYPE.set(CommonUtils.ProcessType.CLIENT);
     MetricsSystem.startSinks(conf.get(PropertyKey.METRICS_CONF_FILE));
+    if (conf.getBoolean(PropertyKey.FUSE_WEB_ENABLED)) {
+      FuseWebServer mWebServer = new FuseWebServer(
```
LuQQiu, line 125: local variable doesn't need to be `mName`
change to `webServer`?

LuQQiu (commented):
@tieujason330 Can you help double check is it the correct way to add a web server for Fuse process?

This PR is added to enable fuse process to expose metrics like what we did in master/worker
https://<fuse_process_host>:<fuse_web_port>/metrics/json or /metrics/prometheus/

tieujason330:
> @tieujason330 Can you help double check is it the correct way to add a web server for Fuse process?
> 
> This PR is added to enable fuse process to expose metrics like what we did in master/worker
> https://<fuse_process_host>:<fuse_web_port>/metrics/json or /metrics/prometheus/

Looks good to me, then fuse-specific configurations will go into `FuseWebServer`

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuse.java
maobaolong: Sure, thanks for remind me.

maobaolong:
@LuQQiu @tieujason330 Thanks for the previous review, PTAL.

LuQQiu:
@apc999 PTAL, this change is likely to increase the alluxio tarball size since fuse depends on common now, is that overhead acceptable?

LuQQiu (approved):
LGTM! @maobaolong in the future, we will prefer to modulize worker, so that we can easily turn on/off worker, on/off fuse when needed

LuQQiu:

Issue #13881, opened 2021-08-02
Merge conf address with server address


ggezer:
**Summary**
Path conf assumes conf can be provided from a dedicated address. This is not justified and should be deleted. Conf will always be provided from the main server.

For reference;
`AbstractClient#getAdress()`
`AbstractClient#getConfAddress()`

**Urgency**
This will help with the always present perf hit mentioned at https://github.com/Alluxio/alluxio/issues/13416

PR #13893, opened 2021-08-04
Avoid redundant query for conf address wherever possible

yyongycy:
In client, there are two addresses( MetaServer and ConfServer), use just MetaServer address.
Fixes https://github.com/Alluxio/alluxio/issues/13881 

Remove the ConfServer field, and use MetaServer instead

  1. This can help reduce the rpc call. 

  1. No


core/common/src/main/java/alluxio/AbstractJobMasterClient.java
```diff
@@ -0,0 +1,73 @@
...
+   * @param retryPolicySupplier retry policy to use
+   */
+  public AbstractJobMasterClient(MasterClientContext clientConf, InetSocketAddress address,
+                              Supplier<RetryPolicy> retryPolicySupplier) {
+    super(clientConf, address, retryPolicySupplier);
+    mConfMasterInquireClient = clientConf.getConfMasterInquireClient();
+  }
+
+  @Override
+  protected void beforeConnect()
```
ggezer, line 61: I believe we can move these implementations back to `AbstractClient` and make them use getConfAddress() instead of mAddress or mConfAddress fields.

core/common/src/main/java/alluxio/AbstractMasterClient.java
```diff
@@ -55,7 +51,15 @@ public AbstractMasterClient(MasterClientContext clientConf, InetSocketAddress ad
       Supplier<RetryPolicy> retryPolicySupplier) {
     super(clientConf, address, retryPolicySupplier);
     mMasterInquireClient = clientConf.getMasterInquireClient();
-    mConfMasterInquireClient = clientConf.getConfMasterInquireClient();
+  }
+
+  @Override
+  protected void beforeConnect()
```
ggezer, line 57: See comment above: I believe we can move these implementations back to `AbstractClient` ...

core/common/src/test/java/alluxio/AbstractClientTest.java
```diff
@@ -128,10 +142,10 @@ public void supportedVersion() throws Exception {
   public void confAddress() throws Exception {
     ClientContext context = Mockito.mock(ClientContext.class);
     Mockito.when(context.getClusterConf()).thenReturn(
-        new InstancedConfiguration(ConfigurationUtils.defaults()));
+            new InstancedConfiguration(ConfigurationUtils.defaults()));
```
ggezer, line 145: nit: fix indent.

ggezer (changes requested):

core/common/src/main/java/alluxio/AbstractJobMasterClient.java
yyongycy: If getConfAddress is called, it will talk to the server side. That would also cause a ping.
Refer to :

  @Override
  public synchronized InetSocketAddress getConfAddress() throws UnavailableException {
    return mConfMasterInquireClient.getPrimaryRpcAddress();
  }

yyongycy: Or we can add a new function, like getAddressForContext, this function just blindly returns mAddress or mConfAddress.

ggezer: But getConfAddress() does the ping only for job-master clients and not for others since you've separated the implementations between the two.

job/server/src/main/java/alluxio/worker/job/RetryHandlingJobMasterClient.java
```diff
@@ -36,7 +36,7 @@
  * workers.
  */
 @ThreadSafe
-public final class RetryHandlingJobMasterClient extends AbstractMasterClient
+public final class RetryHandlingJobMasterClient extends AbstractJobMasterClient
```
ggezer, line 39: There is another job-master client with the same name under:
`src/main/java/alluxio/client/job/RetryHandlingJobMasterClient.java`

yyongycy: Very good finding, that explains why some tests failed.

ggezer (approved):
Thanks!

ggezer:

PR #13895, opened 2021-08-04
Support list a specific status of job

maobaolong:

We want to list RUNNING,  FAILED,  COMPLETED, status job from restful api.

We can list RUNNING,  FAILED,  COMPLETED, status job through restful api. 

No


conf/metrics.properties.template
```diff
@@ -1,99 +0,0 @@
-#
```
yuzhu, line 1: why is this file removed?

maobaolong: Revert it

maobaolong:
@yuzhu Thanks for remind me, revert the deleted template file, PTAL

job/server/src/main/java/alluxio/master/job/JobMasterClientRestServiceHandler.java
```diff
@@ -127,16 +128,23 @@ public JobInfo call() throws Exception {
...
   @Path(ServiceConstants.LIST)
-  public Response list() {
+  public Response list(@QueryParam("status") final String status) {
     return RestUtils.call(new RestUtils.RestCallable<List<Long>>() {
       @Override
       public List<Long> call() throws Exception {
-        return mJobMaster.list(ListAllPOptions.getDefaultInstance());
+        if (status != null) {
+          return mJobMaster.list(ListAllPOptions.newBuilder()
+              .addStatus(Status.valueOf(status))
```
yuzhu, line 143: the api actually supports filtering jobs by multiple statuses, do we want to support that? 
users might want both completed and failed jobs in one call. It is up to you though. 

also the api supports filtering by name as well, you may find it useful to add that. we can get all the replicate jobs or evict jobs that are completed, for example.

maobaolong: This is a great idea, will support it.

maobaolong:
@yuzhu I modified the restful api to be more powerful just like you motioned, PTAL. Thank you for let me know more about `ListAllPOptions`

yuzhu (approved):

yuzhu:

PR #13906, opened 2021-08-09
Fix the job server service hangs on when set a no privileged path

ljl1988com:
Fix the job server service hangs on when  set alluxio.master.journal.folder a no privileged path.

Fix the job server service hangs on when  set alluxio.master.journal.folder a no privileged path.
Because of ratis's bug. See this pr `https://github.com/apache/ratis/pull/477`.

No

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -35,7 +35,6 @@ import alluxio.proto.journal.Journal.JournalEntry;
 import alluxio.util.CommonUtils;
 import alluxio.util.LogUtils;
 import alluxio.util.WaitForOptions;
-import alluxio.util.io.FileUtils;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
@@ -71,10 +70,12 @@ import org.apache.ratis.util.SizeInBytes;
 import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.commons.io.FileUtils;
 
 import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.nio.file.AccessDeniedException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -1026,12 +1027,18 @@ public class RaftJournalSystem extends AbstractJournalSystem {
   public void format() throws IOException {
     File journalPath = mConf.getPath();
     if (journalPath.isDirectory()) {
-      org.apache.commons.io.FileUtils.cleanDirectory(mConf.getPath());
+      if (alluxio.util.io.FileUtils.isStorageDirAccessible(journalPath.getPath())) {
+        FileUtils.cleanDirectory(journalPath);
+      } else {
+        throw new AccessDeniedException(journalPath.getPath());
+      }
     } else {
       if (journalPath.exists()) {
-        FileUtils.delete(journalPath.getAbsolutePath());
+        FileUtils.forceDelete(journalPath);
       }
-      journalPath.mkdirs();
+      if (!journalPath.mkdirs()) {
+        throw new AccessDeniedException(journalPath.getPath());
+      }
     }
   }
 
@@ -1043,7 +1050,7 @@ public class RaftJournalSystem extends AbstractJournalSystem {
   }
 
   /**
-   * @return whether it is allowed to take a local shapshot
+   * @return whether it is allowed to take a local snapshot
    */
   public boolean isSnapshotAllowed() {
     return mSnapshotAllowed.get();
--- a/job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
+++ b/job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
@@ -84,6 +84,9 @@ public class AlluxioJobMasterProcess extends MasterProcess {
     mFileSystem = FileSystem.Factory.create(mFsContext);
     mUfsManager = new JobUfsManager();
     try {
+      if (!mJournalSystem.isFormatted()) {
+        mJournalSystem.format();
+      }
       // Create master.
       mJobMaster = new JobMaster(
           new MasterContext(mJournalSystem, null, mUfsManager), mFileSystem, mFsContext,


yuzhu:
@ggezer since this is journal related, I will leave to you to review.

apc999:
ping @ggezer

job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
```diff
@@ -84,6 +84,9 @@
     mFileSystem = FileSystem.Factory.create(mFsContext);
     mUfsManager = new JobUfsManager();
     try {
+      if (!mJournalSystem.isFormatted()) {
```
ggezer, line 87: How about below?

```
if (!mJournalSystem.isFormatted()) {
        throw new RuntimeException(
            String.format("Journal %s has not been formatted!", mJournalSystem));
      }
```

This is what we do for master processes.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -996,7 +997,9 @@ public void format() throws IOException {
       if (journalPath.exists()) {
         FileUtils.delete(journalPath.getAbsolutePath());
       }
-      journalPath.mkdirs();
+      if (journalPath.mkdirs()) {
```
ggezer, line 1000: I don't understand this.
Throw if creation of folders succeed?

ggezer (changes requested):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ljl1988com: I'm so sorry.
I checkout a new branch, forget '!', and also not test it again.

job/server/src/main/java/alluxio/master/AlluxioJobMasterProcess.java
ljl1988com: Before commit, I see the master processes. There is two reasons not using it.

1.  Not support to format the job master journal in the shell `https://github.com/Alluxio/alluxio/blob/master/bin/alluxio`
2. For users, it's friendly to avoid of manual format.

ggezer: I think it's OK for the job-master processes.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -1031,7 +1032,9 @@ public void format() throws IOException {
       if (journalPath.exists()) {
         FileUtils.delete(journalPath.getAbsolutePath());
       }
-      journalPath.mkdirs();
+      if (!journalPath.mkdirs()) {
+        throw new AccessDeniedException(journalPath.getPath());
```
ggezer, line 1036: why `AccessDeniedException`?

ljl1988com: There is no privilege to operate the journal folder. Do you have good advice?

ggezer: It's kind of misleading as deletion could have failed due to any reason.
I'd rather use `org.apache.commons.io.FileUtils.forceDelete` above to make sure deletion is succeeded.

ggezer (changes requested):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ljl1988com: Thanks for your review. 
I compare `org.apache.commons.io.FileUtils.forceDelete` and `alluxio.util.io.FileUtils.delete`. I find they both invoke `java.nio.file.deleteIfExists`. There is no extra operation logic of `org.apache.commons.io.FileUtils.forceDelete` for a  journal path which is a file.

ggezer: I still don't understand throwing an exception when mkdir returns false.
`FileUtils.cleanDirectory()` will not delete the dir but just clean its children. That means `journalPath.mkdirs()` will return false according to javadoc below. So we'll be throwing for when we should not. 

> 
Returns: true if and only if the directory was created, along with all necessary parent directories; false otherwise

ljl1988com: Maybe I don't understand your discussion.
First we set alluxio.master.journal.folder. The journalPath is `${alluxio.master.journal.folder}/JobJournal`.

`if (journalPath.isDirectory()) {
      FileUtils.cleanDirectory(mConf.getPath()); // When journalPath is a  directory, it just cleans its children.
    } else {  // When journalPath is not a directory, enter this code logic
      if (journalPath.exists()) { 
        FileUtils.forceDelete(journalPath);  // When then file journalPath exists. Delete it.
      }
      if (!journalPath.mkdirs()) { // if `alluxio.master.journal.folder` is a no privileged path, return false.
        throw new AccessDeniedException(journalPath.getPath()); 
      }
    }`

ggezer: Oh. OK. I was misreading the braces.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -1026,12 +1027,18 @@ public boolean isFormatted() {
   public void format() throws IOException {
     File journalPath = mConf.getPath();
     if (journalPath.isDirectory()) {
-      org.apache.commons.io.FileUtils.cleanDirectory(mConf.getPath());
+      if (alluxio.util.io.FileUtils.isStorageDirAccessible(journalPath.getPath())) {
```
ggezer, line 1030: is this due to a new finding?

ggezer (approved):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ljl1988com: Yes, you inspire me. **`FileUtils.cleanDirectory()` will not delete the dir but just clean its children.** If journalPath is a no privileged directory and exists. but journalPath contains none file or dir. The job server service will hang on. I test it on this condition.

ggezer:

Issue #13838, opened 2021-07-23
TotalBlocks metric is negative


maobaolong:
**Alluxio Version:**
master branch 

commit hash 6f8c5d2

**Describe the bug**

After less than an hour running, the totalBlocks metric displayed a negative value.

**To Reproduce**

Not sure it can repro exactly.

**Expected behavior**

A right count of total block

**Urgency**

Normal

**Additional context**

No

LuQQiu:
@yuzhu PTAL, thanks!

maobaolong:
@yuzhu Please take a look thanks.

maobaolong:
@yuzhu Would you like to support monitor the total reported Blocks from worker? As `Master.UniqueBlocks` is the metastore UniqueBlocks which cannot show the report or replica information.

yuzhu:
@maobaolong tracking it in a new issue #14014

PR #13911, opened 2021-08-09
Fix block count metrics

yuzhu:

Provide accurate counting of blocks

Previous count of blocks were inaccurate


core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1305,16 +1301,13 @@ public void registerNewWorkerConfListener(BiConsumer<Address, List<ConfigPropert
...
-        MetricsSystem.counter(MetricKey.MASTER_TOTAL_BLOCKS.getName());
-
     /**
      * Registers metric gauges.
      *
      * @param master the block master handle
      */
     @VisibleForTesting
-    public static void registerGauges(final BlockMaster master) {
+    public static void registerGauges(final DefaultBlockMaster master) {
```
yuzhu, line 1310: This is an inner class, we might as well use the specific implementation of the block master and be able to instrument its internals.

yuzhu:
fixes #13838

apc999 (changes requested):
Can you add UTs to cover
- size() methods for BlockStore
- the metrics?

yuzhu:
@apc999 Added UTs, PTAL

core/server/master/src/main/java/alluxio/master/metastore/BlockStore.java
```diff
@@ -83,6 +83,13 @@
    */
   void close();
 
+  /**
+   * Get the number of blocks in the block store.
+   *
+   * @return size of the block store
+   */
```
apc999, line 90: nit we can simplify the javadoc here

```suggestion
  /**
   * @return size of the block store
   */
```

core/server/master/src/test/java/alluxio/master/block/BlockMasterMetricsTest.java
```diff
@@ -73,6 +75,14 @@ public void testMetricsTierCapacity() {
         getGauge(MetricKey.CLUSTER_CAPACITY_FREE.getName() + MetricInfo.TIER + HDD));
   }
 
+  @Test
+  public void testSize() {
+    BlockStore blockStore = Mockito.mock(BlockStore.class);
+    when(blockStore.size()).thenReturn(100L);
+    Whitebox.setInternalState(mBlockMaster, "mBlockStore", blockStore);
```
apc999, line 82: whitebox is hard to maintain. How about introducing a new constructor so we can inject the dependency of `blockStore`?

```java
@VisibleForTesting
DefaultBlockMaster(MetricsMaster metricsMaster, CoreMasterContext masterContext, Clock clock,
    ExecutorServiceFactory executorServiceFactory, BlockStore blockStore) {
  mBlockStore = blockStore
  ...
}
```

core/server/master/src/test/java/alluxio/master/metastore/heap/HeapBlockStoreTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+package alluxio.master.metastore.heap;
+
+import static org.junit.Assert.assertEquals;
+
+import alluxio.proto.meta.Block;
+
+import org.junit.Test;
+
+public class HeapBlockStoreTest {
```
apc999, line 20: it looks to me `HeapBlockStoreTest#blockSize` is identical to  `RocksBlockStoreTest#blockSize`

How about having a new  parametered test `BlockStoreTest`, with blockStore to be both `RocksBlockStore` and `HeapBlockStore`? check alluxio/client/file/cache/store/PageStoreTest.java as reference

apc999 (changes requested):

core/server/master/src/test/java/alluxio/master/block/BlockMasterMetricsTest.java
yuzhu: for this, since the BlockMaster was a mocked class, whitebox was the only way to do this. otherwise we will need to construct metricsMaster, masterContext etc.

yuzhu:
@apc999 addressed comments and responded with a comment

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -62,6 +63,7 @@
   private final RocksStore mRocksStore;
   private final AtomicReference<ColumnFamilyHandle> mBlockMetaColumn = new AtomicReference<>();
   private final AtomicReference<ColumnFamilyHandle> mBlockLocationsColumn = new AtomicReference<>();
+  private final LongAdder mSize = new LongAdder();
```
maobaolong, line 66: Why we use `LongAdder` here.

Should we reset mSize to `0` when `clear or close` method invoked?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -122,6 +124,7 @@ public void putBlock(long id, BlockMeta meta) {
     try {
       // Overwrites the key if it already exists.
       db().put(mBlockMetaColumn.get(), mDisableWAL, Longs.toByteArray(id), meta.toByteArray());
+      mSize.increment();
```
maobaolong, line 127: If a block is already exist in the rocksdb, can this line executed?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
```diff
@@ -131,6 +134,7 @@ public void putBlock(long id, BlockMeta meta) {
   public void removeBlock(long id) {
     try {
       db().delete(mBlockMetaColumn.get(), mDisableWAL, Longs.toByteArray(id));
+      mSize.decrement();
```
maobaolong, line 137: If the block wasn't exist in the rocksdb, can this line executed?

maobaolong (commented):
@yuzhu Thanks for fix this useful metric, it help us to measure the file blocks state in the alluxio filesystem, left some comments.

maobaolong:
@yuzhu Another high Level discussion, I think there are two `TotalBlocks.Count`, one stand for how many blocks belong to  the exist files, the other shows how many blocks and block replicas are reported by the live workers. Generally, the `reported block` count and `block belong to inodeFile` count are the same, sometime they are different.

apc999:
> @yuzhu Another high Level discussion, I think there are two `TotalBlocks.Count`, one stand for how many blocks belong to the exist files, the other shows how many blocks and block replicas are reported by the live workers. Generally, the `reported block` count and `block belong to inodeFile` count are the same, sometime they are different.

Make sense. 
How about introducing two metrics: `UniqueBlocks.Count` and `BlockReplica.Count`?

core/server/master/src/main/java/alluxio/master/metastore/rocks/RocksBlockStore.java
yuzhu: LongAdder has better performance when contention is high

yuzhu:
@maobaolong addressed your comments, will add a total block as supposed to unique blocks later

maobaolong (approved):
It LGTM now, thanks for this fix

yuzhu:

Issue #13125, opened 2021-03-25
ERROR AlluxioJniFuseFileSystem - Failed to read when setting alluxio.fuse.shared.caching.reader.enabled=true


peterroelants:
**Alluxio Version:**
2.5.0-1

**Describe the bug**

When reading a file from Alluxio FUSE which has `alluxio.fuse.shared.caching.reader.enabled=true`: I get the following errors:
```
ERROR AlluxioJniFuseFileSystem - Failed to read, path: /tmpsq57a1g8.bin size: 131072 offset: 0
java.lang.ArrayIndexOutOfBoundsException: Index -25 out of bounds for length 32
	at alluxio.client.block.stream.SharedGrpcDataReader.getLock(SharedGrpcDataReader.java:68)
	at alluxio.client.block.stream.SharedGrpcDataReader.access$000(SharedGrpcDataReader.java:43)
	at alluxio.client.block.stream.SharedGrpcDataReader$Factory.create(SharedGrpcDataReader.java:167)
	at alluxio.client.block.stream.BlockInStream.readChunk(BlockInStream.java:427)
	at alluxio.client.block.stream.BlockInStream.read(BlockInStream.java:268)
	at alluxio.client.file.AlluxioFileInStream.read(AlluxioFileInStream.java:189)
	at alluxio.fuse.AlluxioJniFuseFileSystem.readInternal(AlluxioJniFuseFileSystem.java:374)
	at alluxio.fuse.AlluxioJniFuseFileSystem.lambda$read$5(AlluxioJniFuseFileSystem.java:349)
	at alluxio.fuse.AlluxioFuseUtils.call(AlluxioFuseUtils.java:278)
	at alluxio.fuse.AlluxioJniFuseFileSystem.read(AlluxioJniFuseFileSystem.java:349)
	at alluxio.jnifuse.AbstractFuseFileSystem.readCallback(AbstractFuseFileSystem.java:180)
ERROR AlluxioJniFuseFileSystem - Failed to read, path: /tmpsq57a1g8.bin size: 4096 offset: 0
java.lang.ArrayIndexOutOfBoundsException: Index -25 out of bounds for length 32
	at alluxio.client.block.stream.SharedGrpcDataReader.getLock(SharedGrpcDataReader.java:68)
	at alluxio.client.block.stream.SharedGrpcDataReader.access$000(SharedGrpcDataReader.java:43)
[...]

PR #13914, opened 2021-08-09
Fix ArrayIndexOutOfBoundsException when using shared.caching.reader

horasal:

Fix issue https://github.com/Alluxio/alluxio/issues/13125

With shared.caching.reader is enabled, alluxio may generate `java.lang.ArrayIndexOutOfBoundsException` when accessing `BLOCK_LOCKS`.

This problem is caused by `core/client/fs/src/main/java/alluxio/client/block/stream/SharedGrpcDataReader.java`

```java
  private static ReentrantReadWriteLock getLock(long blockId) {
    return BLOCK_LOCKS[HASH_FUNC.hashLong(blockId).asInt() % BLOCK_LOCKS.length];
  }
```

`HASH_FUNC` is murmur3_32, it may generate negative number (depend on input blockId, see https://github.com/google/guava/blob/master/guava-tests/test/com/google/common/hash/Murmur3Hash32Test.java#L36 ). Thus it will finally get a negative index while accessing `BLOCK_LOCKS`. We need  a `abs` before mod or just `floorMod` instead of `%`. 

No.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/SharedGrpcDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/SharedGrpcDataReader.java
@@ -25,14 +25,13 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.lang.Math;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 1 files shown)


horasal:
oh sorry, seems I forgot import `java.lang.Math`

apc999:

Issue #13904, opened 2021-08-09
Master report Maximum lock count exceeded


maobaolong:
**Alluxio Version:**

Master branch from commit hash 7b4abe1dd481

**Describe the bug**

Exception in thread "master-rpc-pool-thread-182" java.lang.Error: Maximum lock count exceeded
        at java.util.concurrent.locks.ReentrantReadWriteLock$Sync.fullTryAcquireShared(ReentrantReadWriteLock.java:528)
        at java.util.concurrent.locks.ReentrantReadWriteLock$Sync.tryAcquireShared(ReentrantReadWriteLock.java:488)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1303)
        at java.util.concurrent.locks.ReentrantReadWriteLock$ReadLock.lockInterruptibly(ReentrantReadWriteLock.java:772)
        at alluxio.master.StateLockManager.lockShared(StateLockManager.java:162)
        at alluxio.master.AbstractMaster.createJournalContext(AbstractMaster.java:141)
        at alluxio.master.file.DefaultFileSystemMaster.createRpcContext(DefaultFileSystemMaster.java:4620)
        at alluxio.master.file.DefaultFileSystemMaster.listStatus(DefaultFileSystemMaster.java:986)
        at alluxio.master.file.FileSystemMasterClientServiceHandler.lambda$listStatus$183(FileSystemMasterClientServiceHandler.java:243)
        at alluxio.RpcUtils.callAndReturn(RpcUtils.java:113)
        at alluxio.master.file.FileSystemMasterClientServiceHandler.listStatus(FileSystemMasterClientServiceHandler.java:241)
        at alluxio.grpc.FileSystemMasterClientServiceGrpc$MethodHandlers.invoke(FileSystemMasterClientServiceGrpc.java:2328)
        at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:172)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
        at io.grpc.ForwardingServerCallListener$SimpleForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:40)
[...]

PR #13924, opened 2021-08-11
Stop iterating lost files when journal is not available

ggezer:
LostFileDetector accesses journal directly, and should back off from enumeration if the journal becomes unavailable (i.e. closed).
Fixes https://github.com/Alluxio/alluxio/issues/13904

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/LostFileDetector.java
+++ b/core/server/master/src/main/java/alluxio/master/file/LostFileDetector.java
@@ -65,7 +65,8 @@ final class LostFileDetector implements HeartbeatExecutor {
       } catch (FileDoesNotExistException e) {
         LOG.debug("Exception trying to get inode from inode tree", e);
       } catch (UnavailableException e) {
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/file/LostFileDetector.java
```diff
@@ -65,7 +65,8 @@ public void heartbeat() throws InterruptedException {
       } catch (FileDoesNotExistException e) {
         LOG.debug("Exception trying to get inode from inode tree", e);
       } catch (UnavailableException e) {
-        LOG.warn("Failed to run lost file detector: {}", e.toString());
+        LOG.warn("Journal is not available. Backing off.", e);
```
apc999, line 68: is stack trace helpful here in log warn?

ggezer: nice catch. changed back.

apc999 (approved):

apc999:

PR #13928, opened 2021-08-12
Add number of directories to FuseIOBench

ssz1997:
The number of files must be a multiple of number of directories, so each directory has same amount of files.
The number of files must be a multiple of number of threads, so each thread processes same amount of files.

@LuQQiu PTAL. Much cleaner than last time. Thanks!

### DIFF
--- a/stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
@@ -16,9 +16,6 @@ import alluxio.stress.Parameters;
 import com.beust.jcommander.IStringConverter;
 import com.beust.jcommander.Parameter;
 
-import java.util.Collections;
-import java.util.List;
-
 /**
  * This class holds all the Fuse IO related parameters. All fields are public for easier json
  * ser/de without all the getters and setters.
@@ -34,7 +31,7 @@ public final class FuseIOParameters extends Parameters {
   public FuseIOOperation mOperation;
 
   @Parameter(names = {"--threads"}, description = "the number of concurrent threads to use")
-  public List<Integer> mThreads = Collections.singletonList(1);
+  public int mThreads = 1;
 
   @Parameter(names = {"--local-path"},
       description = "The local filesystem directory or Fuse mount point to perform operations in")
@@ -48,12 +45,13 @@ public final class FuseIOParameters extends Parameters {
           description = "The buffer size for IO operations. (1k, 16k, etc.)")
   public String mBufferSize = "64k";
 
-  @Parameter(names = {"--num-files"},
-      description = "The number of files for the test")
-  public int mNumFiles = 1000;
+  @Parameter(names = {"--num-files-per-dir"},
+      description = "The number of files per directory for the test")
+  public int mNumFilesPerDir = 1000;
 
   @Parameter(names = {"--num-dirs"},
-          description = "The number of directories that the files will be evenly distributed into")
+      description = "The number of directories that the files will be evenly distributed into."
+          + "It is suggested to be a multiple of the number of threads.")
   public int mNumDirs = 1;
 
   @Parameter(names = {"--duration"},
--- a/stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
@@ -52,6 +52,10 @@ public final class FuseIOParameters extends Parameters {
       description = "The number of files for the test")
   public int mNumFiles = 1000;
 
+  @Parameter(names = {"--num-dirs"},
+          description = "The number of directories that the files will be evenly distributed into")
+  public int mNumDirs = 1;
+
   @Parameter(names = {"--duration"},
       description = "The length of time to run the benchmark. (1m, 10m, 60s, 10000ms, etc.)")
   public String mDuration = "30s";
--- a/stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
@@ -46,12 +46,12 @@ public final class FuseIOParameters extends Parameters {
   public String mBufferSize = "64k";
 
   @Parameter(names = {"--num-files-per-dir"},
-      description = "The number of files per directory for the test")
+      description = "The number of files per directory")
   public int mNumFilesPerDir = 1000;
 
   @Parameter(names = {"--num-dirs"},
       description = "The number of directories that the files will be evenly distributed into."
-          + "It is suggested to be a multiple of the number of threads.")
+          + "It must be at least the number of threads and preferably a multiple of it.")
   public int mNumDirs = 1;
 
   @Parameter(names = {"--duration"},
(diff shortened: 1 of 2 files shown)


stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -68,11 +70,16 @@ public static void main(String[] args) {
 
   @Override
   public void prepare() throws Exception {
+    if (mParameters.mNumFiles % mParameters.mNumDirs != 0) {
+      throw new IllegalArgumentException(String
+          .format("Number of files (%d) must be a multiple of the number of directories (%d)",
```
LuQQiu, line 75: I don't like the idea of forcing users to put two parameters that must be a multiplier. Every user will not be happy seeing that. 
What i recommend before is 
- we ask users to provide numDirs, numFilesPerDir, numDirs is better to be a multiplier of the thread number, but it's not mandary. 

Yes, we can simplify our developing logic by making users a little harder to use, but making them super hard to use should be avoided.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -68,11 +70,16 @@ public static void main(String[] args) {
...
+              mParameters.mNumFiles, mParameters.mNumDirs));
+    }
     for (Integer numOfThreads: mParameters.mThreads) {
-      if (numOfThreads > mParameters.mNumFiles) {
+      if (mParameters.mNumFiles % numOfThreads != 0) {
```
LuQQiu, line 79: In our case, better to not allowing thread array list. 
I took a look at Fuse logics, there are many operating system cache involved in reading, which means that when the application read a file for the first time, it goes to Fuse, but then the file may be cached in the operating system and no actual read happens at AlluxioFuse. 
We should just change mThreads to a single integer

LuQQiu (changes requested):

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: I see. Yeah this sounds more reasonable. Thanks!

ssz1997:
@LuQQiu Each directory is only written by one thread. PTAL thanks!

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -293,23 +288,25 @@ private void runInternal() throws Exception {
...
-          if (CommonUtils.getCurrentMs() > recordMs) {
-            if (ioBytes > 0) {
-              mThreadCountResult.incrementIOBytes(ioBytes);
-            } else {
-              // Done reading/writing one file
-              break;
+      for (int dirId = mThreadId; dirId < mParameters.mNumDirs; dirId += mParameters.mThreads) {
+        for (int fileId = 0; fileId < mParameters.mNumFilesPerDir; fileId++) {
+          mCurrentOffset = 0;
+          String filePath = mParameters.mLocalPath + "/" + dirId + "/" + fileId;
```
LuQQiu, line 294: Avoid string concatenation, which is low efficiency. Each + will create a new object.
use string.format here, in other cases, can use stringBuilder

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -293,23 +288,25 @@ private void runInternal() throws Exception {
...
+            }
+            long ioBytes = applyOperation(filePath);
+
+            // Start recording after the warmup
+            if (CommonUtils.getCurrentMs() > recordMs) {
+              if (ioBytes > 0) {
+                mThreadCountResult.incrementIOBytes(ioBytes);
+              } else {
+                // Done reading/writing one file
+                break;
```
LuQQiu, line 308: if (CommonUtils.getCurrentMs() < recordMs && ioBytes <= 0 ) { we should also break! we also done reading/writing one file
Move the break logic after long ioBytes = applyOperation(filePath);

LuQQiu (changes requested):

stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
```diff
@@ -48,9 +45,14 @@
...
-  @Parameter(names = {"--num-files"},
-      description = "The number of files for the test")
-  public int mNumFiles = 1000;
+  @Parameter(names = {"--num-files-per-dir"},
+      description = "The number of files per directory for the test")
+  public int mNumFilesPerDir = 1000;
+
+  @Parameter(names = {"--num-dirs"},
+      description = "The number of directories that the files will be evenly distributed into."
+          + "It is suggested to be a multiple of the number of threads.")
```
LuQQiu, line 54: The number of directories should be bigger or equal to the thread number, and it's suggested to be a multiple of the number of threads.

stress/common/src/main/java/alluxio/stress/fuse/FuseIOParameters.java
```diff
@@ -48,9 +45,14 @@
           description = "The buffer size for IO operations. (1k, 16k, etc.)")
   public String mBufferSize = "64k";
 
-  @Parameter(names = {"--num-files"},
-      description = "The number of files for the test")
-  public int mNumFiles = 1000;
+  @Parameter(names = {"--num-files-per-dir"},
+      description = "The number of files per directory for the test")
```
LuQQiu, line 49: no need to add `for the test`

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -68,12 +70,9 @@ public static void main(String[] args) {
...
   @Override
   public void prepare() throws Exception {
-    for (Integer numOfThreads: mParameters.mThreads) {
-      if (numOfThreads > mParameters.mNumFiles) {
-        throw new IllegalArgumentException(String
-            .format("Number of threads (%d) must be larger than number of files (%d)",
-                numOfThreads, mParameters.mNumFiles));
-      }
+    if (mParameters.mThreads > mParameters.mNumDirs) {
+      LOG.warn("Some threads are not being used. It is suggested to make the number of "
```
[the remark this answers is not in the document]
ssz1997: I'm going to throw an exception. Setting mThreads = mNumDirs without telling users explicitly would make them confused at the end why the mThreads was changed.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -291,22 +293,21 @@ private void runInternal() throws Exception {
...
+          String filePath = String.format("%s/%d/%d", mParameters.mLocalPath, dirId, fileId);
           while (!Thread.currentThread().isInterrupted()) {
             if (isRead && CommonUtils.getCurrentMs() > mContext.getEndMs()) {
               closeInStream();
               return;
             }
             long ioBytes = applyOperation(filePath);
 
+            // Done reading/writing one file
+            if (ioBytes < 0) {
```
LuQQiu, line 305: <= 0

LuQQiu (commented):
Overall LGTM, careful about all the corner cases, thanks!

ssz1997:
@LuQQiu Fixed the comments. PTAL thanks!

LuQQiu (approved):
LGTM, thanks for supporting writing millions of files through Fuse stress bench test

ssz1997:

LuQQiu:

PR #13936, opened 2021-08-16
Improve UfsIOBench and RpcBench command descriptions

jiacheliu3:

This change implements the missing description for `UfsIOBench` command and improves descriptions in `RpcBench` children, conforming to the same format.

Below is an example from `WorkerHeartbeatBench`. The rough structure is description -> example -> options -> parse exception.

```
$ bin/alluxio runClass alluxio.stress.cli.WorkerHeartbeatBench foofoofoo
A benchmarking tool for the WorkerHeartbeat RPC.
The test will generate a specified number of blocks in the master (without associated files). The test will also register the simulated workers with the master. Then it will keep generating heartbeats with the specified load and sending heartbeats to the master nonstop, until the specified time has elapsed.

Example:
# 2 job workers will be chosen to run the benchmark
# Each job worker runs 3 threads each simulating one worker
# Each worker will have 3000 blocks on tier 0 and 10000 blocks on tier 1
# Keep sending heartbeats for 30s
$ bin/alluxio runClass alluxio.stress.cli.WorkerHeartbeatBench --concurrency 3 \
--cluster --cluster-limit 2 --tiers "1000,1000,1000;5000,5000" --duration 30s

Usage: WorkerHeartbeatBench [options]
  Options:
    --bench-timeout
      The length of time to wait when finishing the benchmark. (10m 600s, 
      etc.) 
      Default: 20m
    --cluster
      If true, runs the benchmark via the job service cluster. Otherwise, runs 
      locally. 
      Default: false
    --cluster-limit
      If greater than 0, it will only run on that number of workers. If 0, 
      will run on all available cluster workers. If < 0, will run on the 
      workers from the end of the worker list. This flag is only used if 
      --cluster is enabled.
      Default: 0
    --concurrency
      simulate this many clients/workers on one machine
      Default: 2
    --duration
      The length of time to run the benchmark. (1m, 10m, 60s, 10000ms, etc.)
      Default: 5s
    -h, --help

    --java-opt
      The java options to add to the command line to for the task. This can be 
      repeated. The options must be quoted and prefixed with a space, to avoid 
      getting passed to the JVM. For example: --java-opt " -Xmx4g" --java-opt 
      " -Xms2g"
      Default: []
    --profile-agent
      The path to the profile agent if one is available. Providing this will 
      enable a more detailed output.
      Default: <empty string>
    --tiers
      The number of blocks in each storage dir and tier. List tiers in the 
      order of MEM, SSD and HDD separated by semicolon, list dirs inside each 
      tier separated by comma. Example: "100,200,300;1000,1500;2000"

com.beust.jcommander.ParameterException: Was passed main parameter 'foofoofoo' but no main parameter was defined in your arg class
	at com.beust.jcommander.JCommander.initMainParameterValue(JCommander.java:954)
	at com.beust.jcommander.JCommander.parseValues(JCommander.java:755)
	at com.beust.jcommander.JCommander.parse(JCommander.java:356)
	at com.beust.jcommander.JCommander.parse(JCommander.java:335)
	at alluxio.stress.cli.Benchmark.run(Benchmark.java:134)
	at alluxio.stress.cli.Benchmark.mainInternal(Benchmark.java:90)
	at alluxio.stress.cli.WorkerHeartbeatBench.main(WorkerHeartbeatBench.java:192)
```

See above

  1. Command output

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
@@ -72,10 +72,10 @@ public class GetPinnedFileIdsBench extends RpcBench<GetPinnedFileIdsParameters> 
         "Example:",
         "# 2 job workers will be chosen to run the benchmark",
         "# Each job worker runs 3 simulated clients",
-        "# Each client keeps requesting a total number of 100k pinned files for "
-            + "a total of 5 seconds",
+        "# Each client keeps requesting a total number of 10k pinned files for "
+            + "a total of 100 milliseconds",
         "$ bin/alluxio runClass alluxio.stress.cli.GetPinnedFileIdsBench --concurrency 3 \\",
-        "--cluster --cluster-limit 2 --num-files 100000 --duration 5s",
+        "--cluster --cluster-limit 2 --num-files 10000 --duration 100ms",
         ""
     ));
   }
--- a/stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
@@ -67,13 +67,15 @@ public class GetPinnedFileIdsBench extends RpcBench<GetPinnedFileIdsParameters> 
         "A benchmarking tool for the GetPinnedFileIds RPC.",
         "The test will generate a specified number of test files and pin them. "
             + "Then it will keep calling the GetPinnedFileIds RPC by the specified load until the "
-            + "specified duration has elapsed.",
+            + "specified duration has elapsed. The test files will be cleaned up in the end.",
         "",
         "Example:",
-        "4 simulated workers running on 2 job workers, requesting a total number of 100k pinned "
-            + "files for a total of 5 seconds:",
-        "$ bin/alluxio runClass alluxio.stress.cli.GetPinnedFileIdsBench --concurrency 2 "
-            + "--cluster-limit 2 --num-files 100000 --duration 5s",
+        "# 2 job workers will be chosen to run the benchmark",
+        "# Each job worker runs 3 simulated clients",
+        "# Each client keeps requesting a total number of 100k pinned files for "
+            + "a total of 5 seconds",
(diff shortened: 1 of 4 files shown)


jiacheliu3:
@LuQQiu @yuzhu  some general open questions regarding this change:
1. Should the command format be `$ bin/alluxio runClass ...` or we should provide wrappers like `runUfsIOTest` does? 
2. Following the above question, how do we add the commands to the corresponding doc page? Do they still belong to the `User Commands` page or there should be a separate page specific to the stress bench framework.

Also if the format now looks good, the `FuseIOBench` should be updated accordingly.

stress/shell/src/main/java/alluxio/stress/cli/GetPinnedFileIdsBench.java
```diff
@@ -67,13 +67,15 @@ public String getBenchDescription() {
...
-        "4 simulated workers running on 2 job workers, requesting a total number of 100k pinned "
-            + "files for a total of 5 seconds:",
-        "$ bin/alluxio runClass alluxio.stress.cli.GetPinnedFileIdsBench --concurrency 2 "
-            + "--cluster-limit 2 --num-files 100000 --duration 5s",
+        "# 2 job workers will be chosen to run the benchmark",
+        "# Each job worker runs 3 simulated clients",
+        "# Each client keeps requesting a total number of 10k pinned files for "
+            + "a total of 100 milliseconds",
+        "$ bin/alluxio runClass alluxio.stress.cli.GetPinnedFileIdsBench --concurrency 3 \\",
+        "--cluster --cluster-limit 2 --num-files 10000 --duration 100ms",
```
dbw9580, line 78: Since this does not actually work now (needs `--in-process` instead of `--cluster`), IMO we should provide a working example and change it after the blocker is fixed.

dbw9580 (approved):
Thanks for the improvement. I only have  a minor comment.

LuQQiu (approved):
LGTM, thanks for adding the descriptions

LuQQiu:

LuQQiu:
1. i would prefer a consolidate way for all the stressbench test, instead of each test has an entrypoint, things like bin/alluxio stress ufsIO will be preferred
2. i would prefer all the stressbench tests introduced in stressbench doc page
3.

PR #13944, opened 2021-08-17
Update default ZK path for jobmaster

tomscut:
IMO, "/job_election" and "/job_leader" should not be created in the root directory. So I updated default ZK path for jobmaster, and added prefix "/alluxio". 

image

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -5203,9 +5203,11 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
 
   public static final PropertyKey ZOOKEEPER_JOB_ELECTION_PATH =
-      new Builder(Name.ZOOKEEPER_JOB_ELECTION_PATH).setDefaultValue("/alluxio/job_election").build();
+      new Builder(Name.ZOOKEEPER_JOB_ELECTION_PATH)
+          .setDefaultValue("/alluxio/job_election").build();
   public static final PropertyKey ZOOKEEPER_JOB_LEADER_PATH =
-      new Builder(Name.ZOOKEEPER_JOB_LEADER_PATH).setDefaultValue("/alluxio/job_leader").build();
+      new Builder(Name.ZOOKEEPER_JOB_LEADER_PATH)
+          .setDefaultValue("/alluxio/job_leader").build();
 
   //
   // JVM Monitor related properties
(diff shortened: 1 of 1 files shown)


tomscut:
Hi @maobaolong @ggezer ,  these failed checks seem unrelated to the change. Could you please help review the code. Thanks a lot.

yuzhu (approved):
LGTM, thanks for fixing this issue!

tomscut:
> LGTM, thanks for fixing this issue!

Thanks @yuzhu for your review.

yuzhu:
@ggezer is there back compat concerns here? maybe when we do a rolling upgrade between a version before this change and a version after this change?

ggezer:
@tomscut Alluxio masters also operate under /alluxio/ path. Having job_master into same path will introduce races. Hence elections will disturb each other. 
@yuzhu your backcompat concerns are correct. We can't make rolling upgrade with this change.

tomscut:
> @tomscut Alluxio masters also operate under /alluxio/ path. Having job_master into same path will introduce races. Hence elections will disturb each other.
> @yuzhu your backcompat concerns are correct. We can't make rolling upgrade with this change.

Thanks @ggezer for your comment. However, job_master and master information are stored in different subdirectories, why should there be competition? 

I'm worried that if we put job_master in the root directory, it might conflict with the pathnames of other services. It also doesn't look very elegant.

ggezer:
@tomscut I initially thought there could be race when creating the paths the first time for both master and job_master concurrently. However, zk should take care of that.

My only concern is the breaking change to the rolling upgrade. On a second thought, there shouldn't be an issue as long as clients are also restarted. But this will definitely need to be known in advance. I don't know when do we want to merge this change. @apc999 ?

tomscut:
> @tomscut I initially thought there could be race when creating the paths the first time for both master and job_master concurrently. However, zk should take care of that.
> 
> My only concern is the breaking change to the rolling upgrade. On a second thought, there shouldn't be an issue as long as clients are also restarted. But this will definitely need to be known in advance. I don't know when do we want to merge this change. @apc999 ?

Thank you @ggezer for your reply.

tomscut:
> @tomscut I initially thought there could be race when creating the paths the first time for both master and job_master concurrently. However, zk should take care of that.
> 
> My only concern is the breaking change to the rolling upgrade. On a second thought, there shouldn't be an issue as long as clients are also restarted. But this will definitely need to be known in advance. I don't know when do we want to merge this change. @apc999 ?

Hi @apc999 , could you please help take a quick look at this. Thank you.

yuzhu:
@ggezer @apc999 are the back compat concerns addressed? do we want to delay this until later release?

apc999:
I am good with this PR. Alluxio HA with ZK is not the way the community is focusing to support so I wouldn't concern on rolling upgrade

apc999 (approved):

apc999:

tomscut:
Thanks @apc999 @yuzhu @ggezer for your review and merge.

PR #13949, opened 2021-08-17
Add job service stress bench

jja725:

The job service stress bench only supports distributed load scalability. The first commit for skeleton code

necessary for job service scalability updates

  1. additional cli


stress/common/src/main/java/alluxio/stress/jobservice/JobServiceBenchTaskResultStatistics.java
```diff
@@ -0,0 +1,122 @@
...
+import org.HdrHistogram.Histogram;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.zip.DataFormatException;
+
+/**
+ * Statistics class that is used in {@link JobServiceBenchTaskResult}.
+ */
+public class JobServiceBenchTaskResultStatistics {
```
yuzhu, line 28: this class seems very similar to `MasterBenchTaskResultStatistics`, any ways we could refactor here? what makes these two different?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * @param args command-line arguments
+   */
+  public static void main(String[] args) {
+    mainInternal(args, new StressJobServiceBench());
+  }
+
+  @Override
+  public void prepare() throws Exception {
```
yuzhu, line 75: so if i understood this correctly, this prepare is going to run on every job worker, and creating potentially duplicate files if it is a shared file system right?

yuzhu: can this be avoided?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
```diff
@@ -0,0 +1,280 @@
...
+    try (FileOutStream os = fs.createFile(fileURI, options)) {
+      byte[] arr = new byte[len];
+      for (int k = 0; k < len; k++) {
+        arr[k] = (byte) k;
+      }
+      os.write(arr);
+    }
+  }
+
+  private void deleteBasePath(FileSystem fs) throws IOException, AlluxioException {
```
yuzhu, line 113: similarly this delete is going to happen n times, once for every worker?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
```diff
@@ -0,0 +1,280 @@
...
+
+    private void applyOperation(String dirPath) throws IOException, AlluxioException {
+      switch (mParameters.mOperation) {
+        case DISTRIBUTED_LOAD:
+          // send distributed load task to job service and wait for result
+          int numReplication = 1;
+          FileSystemContext fsContext =
+              FileSystemContext.create(new InstancedConfiguration(ConfigurationUtils.defaults()));
+          DistributedLoadCommand cmd = new DistributedLoadCommand(fsContext);
+          DistributedLoadUtils.distributedLoad(cmd, new AlluxioURI(dirPath), numReplication,
```
yuzhu, line 271: do we have a way of checking for failure?

yuzhu (changes requested):
thanks for the improvement! overall looks good, except a few questions.  PTAL, thanks.

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: in line 96, every worker would try to create its own directory with mBaseParameters.mId, so I think it would be different directories(directories contains the same files)

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: yeah, you are right, this would cause delete many times even there's a check in the beginning... an idea would be let workers delete their own directory to avoid many workers delete the whole base directory. Another thought would be something mentioned by Bowen, we have a preparation stage for the whole cluster and another preparation stage for each worker(this would include larger change).

stress/common/src/main/java/alluxio/stress/jobservice/JobServiceBenchTaskResultStatistics.java
jja725: yeah it is, the only difference is the max response time buckets, we definitely can do small refactoring as we discussed before. I would personally prefer to get rid of this class and just put all the stuff into TaskResult class and init the TaskResult class with a parameter for the max response time buckets instead of using a constant.

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
yuzhu: i see. what does `mBaseParameters.mId` return when we are not using cluster mode?  is it still valid?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
yuzhu: maybe just add a TODO here? delete is less of a concern, we just have to make sure repeated deletes do not cause an error.

stress/common/src/main/java/alluxio/stress/jobservice/JobServiceBenchTaskResultStatistics.java
yuzhu: Let's add a TODO and do the refactoring in a separate PR?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: it's valid, in my local machine it's local-task-0

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: use DistributedLoadCommand.getFailedCount() for this info and record in result

yuzhu:
@LuQQiu @ssz1997 PTAL, as you have written / reviewed similar code in the recent past.

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
ssz1997: Did you see a "local-task-0" directory in cluster mode in your test? If yes, I'm more towards not creating this folder because it could be confusing for people though not affecting performance.

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
```diff
@@ -0,0 +1,280 @@
...
+
+  @Override
+  public String getBenchDescription() {
+    return "";
+  }
+
+  @Override
+  public JobServiceBenchTaskResult runLocal() throws Exception {
+    ExecutorService service =
+        ExecutorServiceFactories.fixedThreadPool("bench-thread", mParameters.mNumDirs).create();
```
ssz1997, line 129: What would happen if mNumDirs is very large?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
```diff
@@ -0,0 +1,280 @@
...
+    long timeOutMs = FormatUtils.parseTimeSize(mBaseParameters.mBenchTimeout);
+    long startMs = mBaseParameters.mStartMs;
+    if (mBaseParameters.mStartMs == BaseParameters.UNDEFINED_START_MS) {
+      startMs = CommonUtils.getCurrentMs() + 1000;
+    }
+    JobMasterClient client = JobMasterClient.Factory
```
ssz1997, line 135: I don't see this client being used? Am I missing anything?

ssz1997 (commented):
Left some comments. Thanks!

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
yuzhu: good catch @ssz1997 , maybe we do need to delete individual directories then.

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
ssz1997: I mean, each job worker will run this and creates a thread pool of mNumDirs number of threads, and each thread will issue a distributedLoad targeting one directory, right? If there are thousands of directories or more, I feel like the job worker just can't create this many of threads?

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: use numSuccess to record number of loaded files so we can know number of failure as well.

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: yeah I know what you mean, we have to be careful about the parameter for numDirs so it doesn't overload the job worker. And I feel like the test case we want to focus on is sending some concurrent requests(relative small) with lots of files per request. So I would not worry too much about this problem right now.

ssz1997 (approved):
LGTM. Thanks!

yuzhu:
LGTM, thanks.

yuzhu (approved):

yuzhu:

PR #13950, opened 2021-08-18
Make the call to transferLeadership asynchronous

jenoudet:
To avoid the `quorum elect` command from hanging indefinitely

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
@@ -37,10 +37,17 @@ public interface JournalMasterClient extends Closeable {
   void removeQuorumServer(NetAddress serverAddress) throws AlluxioStatusException;
 
   /**
-   * Changes the leading master of the quorum.
+   * Initiates changing the leading master of the quorum.
    *
-   * @param newLeaderNetAddress server address of the new leader
+   * @param newLeaderNetAddress server address of the prospective new leader
    * @throws AlluxioStatusException
    */
   void transferLeadership(NetAddress newLeaderNetAddress) throws AlluxioStatusException;
+
+  /**
+   * Resets RaftPeer priorities.
+   *
+   * @throws AlluxioStatusException
+   */
+  void resetPriorities() throws AlluxioStatusException;
 }
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -928,18 +928,19 @@ public class RaftJournalSystem extends AbstractJournalSystem {
               serverAddress, newLeaderPeerId);
       // fire and forget: need to immediately return as the master will shut down its RPC servers
       // once the TransferLeadershipRequest is initiated.
-      Thread.sleep(3_000);
+      final int sleepTimeMs = 3_000;
       new Thread(() -> {
         try {
+          Thread.sleep(sleepTimeMs);
           client.admin().transferLeadership(newLeaderPeerId, TRANSFER_LEADER_WAIT_MS);
         } catch(Throwable t){
           LOG.error("caught an error: {}", t.getMessage());
           /* checking the transfer happens in {@link QuorumElectCommand} */
         }
       }).start();
-      LOG.info("TransferLeadershipRequest sent");
+      LOG.info("Transferring leadership initiated");
     } catch (Throwable t) {
-      LOG.error(t.getMessage());
+      throw new IOException(t);
     }
   }
 
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -912,41 +912,34 @@ public class RaftJournalSystem extends AbstractJournalSystem {
     // --- end of updating priorities ---
     final int TRANSFER_LEADER_WAIT_MS = 30_000;
     try (RaftClient client = createClient()) {
-      LOG.info("Applying new peer state before transferring leadership: {}",
-              peersToString(peersWithNewPriorities));
+      String stringPeers = "[" + peersWithNewPriorities.stream().map(RaftPeer::toString)
+                      .collect(Collectors.joining(", ")) + "]";
+      LOG.info("Applying new peer state before transferring leadership: {}", stringPeers);
       // set peers to have new priorities
       RaftClientReply reply = client.admin().setConfiguration(peersWithNewPriorities);
-      processReply(reply);
+      if (!reply.isSuccess()) {
+        throw reply.getException() != null
+                ? reply.getException()
+                : new IOException(String.format("reply <%s> failed", reply));
+      }
       // transfer leadership
-      LOG.info("Transferring leadership to master with address <{}> and with RaftPeerId <{}>",
+      LOG.info("Transferring leadership to master with address <{}> and with RaftPeerId " +
+                      "<{}>",
               serverAddress, newLeaderPeerId);
-      reply = client.admin().transferLeadership(newLeaderPeerId, TRANSFER_LEADER_WAIT_MS);
-      processReply(reply);
-      // reset the peers to have the old priorities
-      LOG.info("Resetting peer state to before transfer: {}", peersToString(oldPeers));
-      reply = client.admin().setConfiguration(oldPeers);
-      processReply(reply);
-      LOG.info("Successfully reset peer state");
-    }
-  }
-
-  /**
-   * @param peers to be printed into a string
-   * @return the peers as a comma delimited list
-   */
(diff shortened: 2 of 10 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -912,41 +912,34 @@ public synchronized void transferLeadership(NetAddress newLeaderNetAddress) thro
...
+        try {
+          client.admin().transferLeadership(newLeaderPeerId, TRANSFER_LEADER_WAIT_MS);
+        } catch(Throwable t){
+          LOG.error("caught an error: {}", t.getMessage());
+          /* checking the transfer happens in {@link QuorumElectCommand} */
+        }
+      }).start();
+      LOG.info("TransferLeadershipRequest sent");
+    } catch (Throwable t) {
+      LOG.error(t.getMessage());
```
ggezer, line 942: can you wrap this in `IOException` and throw back?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -912,41 +912,34 @@ public synchronized void transferLeadership(NetAddress newLeaderNetAddress) thro
...
+      Thread.sleep(3_000);
+      new Thread(() -> {
+        try {
+          client.admin().transferLeadership(newLeaderPeerId, TRANSFER_LEADER_WAIT_MS);
+        } catch(Throwable t){
+          LOG.error("caught an error: {}", t.getMessage());
+          /* checking the transfer happens in {@link QuorumElectCommand} */
+        }
+      }).start();
+      LOG.info("TransferLeadershipRequest sent");
```
ggezer, line 940: nit: something like "Transferring leadership initiated."

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -912,41 +912,34 @@ public synchronized void transferLeadership(NetAddress newLeaderNetAddress) thro
...
-   * @throws IOException
-   */
-  private void processReply(RaftClientReply reply) throws IOException {
-    if (!reply.isSuccess()) {
-      throw reply.getException() != null
-              ? reply.getException()
-              : new IOException(String.format("reply <%s> failed", reply));
+      // fire and forget: need to immediately return as the master will shut down its RPC servers
+      // once the TransferLeadershipRequest is initiated.
+      Thread.sleep(3_000);
```
ggezer, line 931: I believe this should be inside the thread's task.
Also please define 3_000 above as a constant.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -912,41 +912,34 @@ public synchronized void transferLeadership(NetAddress newLeaderNetAddress) thro
...
+      }
       // transfer leadership
-      LOG.info("Transferring leadership to master with address <{}> and with RaftPeerId <{}>",
+      LOG.info("Transferring leadership to master with address <{}> and with RaftPeerId " +
+                      "<{}>",
               serverAddress, newLeaderPeerId);
-      reply = client.admin().transferLeadership(newLeaderPeerId, TRANSFER_LEADER_WAIT_MS);
-      processReply(reply);
-      // reset the peers to have the old priorities
-      LOG.info("Resetting peer state to before transfer: {}", peersToString(oldPeers));
```
ggezer, line 926: We should still do this, after transfer is successfully completed. It's just that it needs to be an additional call from shell to the new leader.

ggezer (changes requested):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
jenoudet: I added functionality where `transferLeadership` returns the old priorities to the shell command and then the shell command passes these priorities to a new RPC called `resetPriorities` so that the priorities may be reset the way they were prior to the transfer of leadership.

maobaolong:
@jenoudet Could you please add more context and reason why it will hang indefinitely without this PR?

jenoudet:
> @jenoudet Could you please add more context and reason why it will hang indefinitely without this PR?

Prior to this PR, we made a `SetConfigurationRequest` to reset the RaftPeer priorities after the `TransferLeadershipRequest`. This was to ensure that all the masters were equally likely to be elected in the case of an election after the `TransferLeadershipRequest`. The issue with this, is that once the transfer of leadership is initiated, the leading Alluxio master will shut down its RPC server, and therefore hangs indefinitely waiting for the response from the `SetConfigurationRequest`. 
Now, the `tranferLeadership` action returns a list of RaftPeer priorities so that they may be reset after the transfer of leadership is completed to the newly elected master.

core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
```diff
@@ -40,7 +43,18 @@
    * Changes the leading master of the quorum.
```
ggezer, line 43: Change the comment to reflect new behavior.
"Initiate changing of..."

core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
```diff
@@ -40,7 +43,18 @@
...
    *
    * @param newLeaderNetAddress server address of the new leader
+   * @return the old peer priorities before the transfer of leadership
    * @throws AlluxioStatusException
    */
-  void transferLeadership(NetAddress newLeaderNetAddress) throws AlluxioStatusException;
+  TransferLeadershipPResponse transferLeadership(NetAddress newLeaderNetAddress)
```
ggezer, line 49: Let's not mix priorities with transfer functionality at the moment. I believe `void` is fine.

core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
```diff
@@ -40,7 +43,18 @@
...
+          throws AlluxioStatusException;
+
+  /**
+   * Resets the peer priorities to the ones specified.
+   *
+   * @param peerPriorities the priorities
+   * @return
+   * @throws AlluxioStatusException
+   */
+  void resetPriorities(List<PeerPriority> peerPriorities) throws AlluxioStatusException;
```
ggezer, line 59: how about reset() with no arguments that sets all priorities to 1?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -84,8 +84,8 @@ public int run(CommandLine cl) throws IOException {
       mPrintStream.println(String.format(OUTPUT_FAIL, serverAddress));
       return 0;
     }
-
     mPrintStream.println(String.format(OUTPUT_SUCCESS, serverAddress));
+    jmClient.resetPriorities(priorities.getPrioritiesList());
```
ggezer, line 88: What if this fails?

ggezer (changes requested):

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
jenoudet: added a `try / catch`
I call `resetPriorities` independently of `transferLeadership` success because the priorities could have been changed while the transfer failed.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -80,12 +82,19 @@ public int run(CommandLine cl) throws IOException {
...
+    // setting of priorities may have succeeded while the transfer might not have.
+    try {
+      jmClient.resetPriorities();
+      mPrintStream.println(RESET_SUCCESS);
+    } catch (Exception e) {
+      mPrintStream.println(String.format(RESET_FAILED, e));
     }
-
-    mPrintStream.println(String.format(OUTPUT_SUCCESS, serverAddress));
     return 0;
```
ggezer, line 98: Shell should return -1 whenever it failed.

ggezer (approved):
Thanks!

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -69,6 +71,7 @@ public int run(CommandLine cl) throws IOException {
 
     MasterInquireClient inquireClient = MasterInquireClient.Factory
             .create(mConf, FileSystemContext.create(mConf).getClientContext().getUserState());
+    boolean success = true;
     // wait for confirmation of leadership transfer
     try {
       CommonUtils.waitFor("Waiting for leadership transfer to finalize", () -> {
```
ggezer, line 77: Does this wait still waits for too long to return?

ggezer:

PR #13965, opened 2021-08-23
Add metadata cache metric

bzheng888:
Add meta data cache size gauge metric.  
Add a new metric key called CLIENT_META_DATA_CACHE_SIZE
Register the new key to metric system in MetadataCachingBaseFileSystem.
Get client metadata cache size info, let us know how the meta cache size grows under specific workload, it will help us to adjust parameters to the load.

No
  1. change in user-facing APIs
  No
  No
  3. webui
  No

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
@@ -25,6 +25,8 @@ import alluxio.grpc.DeletePOptions;
 import alluxio.grpc.GetStatusPOptions;
 import alluxio.grpc.ListStatusPOptions;
 import alluxio.grpc.RenamePOptions;
+import alluxio.metrics.MetricKey;
+import alluxio.metrics.MetricsSystem;
 import alluxio.util.FileSystemOptions;
 import alluxio.util.ThreadUtils;
 import alluxio.wire.FileInfo;
@@ -78,6 +80,9 @@ public class MetadataCachingBaseFileSystem extends BaseFileSystem {
     // asynchronously update access time.
     mAccessTimeUpdater = new ThreadPoolExecutor(0, masterClientThreads, THREAD_KEEPALIVE_SECOND,
         TimeUnit.SECONDS, new SynchronousQueue<>());
+    MetricsSystem.registerGaugeIfAbsent(
+        MetricsSystem.getMetricName(MetricKey.CLIENT_META_DATA_CACHE_SIZE.getName()),
+        mMetadataCache::size);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


bzheng888:
image 
The picture shows metadata metric in the fuse client web.

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -1452,6 +1452,13 @@ public MetricKey build() {
           .setMetricType(MetricType.COUNTER)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey CLIENT_META_DATA_CACHE_SIZE =
+      new Builder("Client.MetadataCacheSize")
+          .setDescription("Size of cached metadata. Only valid if the filesystem is"
```
LuQQiu, line 1457: How about change to: The total number of files and directories whose metadata is cached on the client-side.

core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
```diff
@@ -78,6 +80,9 @@ public MetadataCachingBaseFileSystem(FileSystemContext context) {
     // asynchronously update access time.
     mAccessTimeUpdater = new ThreadPoolExecutor(0, masterClientThreads, THREAD_KEEPALIVE_SECOND,
         TimeUnit.SECONDS, new SynchronousQueue<>());
+    MetricsSystem.registerGaugeIfAbsent(
+        MetricsSystem.getMetricName(MetricKey.CLIENT_META_DATA_CACHE_SIZE.getName()),
+        mMetadataCache::size);
```
LuQQiu, line 85: Looks like guava cache is doing some light calculation inside the size(), will it be better to use cachedguage here? @yuzhu 
what's your thoughts on that?

LuQQiu (commented):
Thanks for adding the metadata cache metrics, left some comments

core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
maobaolong: @LuQQiu Good suggestion, but I think `10min` is too long, is there a way to cache the metric for a shorter time?

yuzhu: @maobaolong 10 mins interval is just a setting. We can overload `registerCachedGaugeIfAbsent` to pass in a time?

bzheng888:
@maobaolong @yuzhu @LuQQiu PTAL!

maobaolong (approved):
LGTM. Thanks for this improvement.

PR #13967, opened 2021-08-23
Support distributed set replication command

maobaolong:
### What changes are proposed in this pull request?

Add a new job service command line tools to support distributed set replication by cli.

### Why are the changes needed?

Support set replication by cli tools, tracked by cli tool, rather than master replicationChecker.

### Does this PR introduce any user facing changes?

A new cli tools.

TODO: a documentation work related.

```bash
bin/alluxio fs distributedSetRep /test/ --replication 1
Allow up to 3000 active jobs
Replicate /test/a to 1
Replicate /test/NOTICE to 1
Replicate /test/b to 1
Successfully set replication /test/a to 1 after 1 attempts
Successfully set replication /test/NOTICE to 1 after 1 attempts
Successfully set replication /test/b to 1 after 1 attempts
```

PR #13978, opened 2021-08-25
Remove the in alluxio check for distributedload

maobaolong:

Address suggestion in https://github.com/Alluxio/alluxio/pull/13967, remove the alluxio check for distributedload.

With this PR, we can re-load the file to a new specific replicas, something like from 1 replica to 3 replica

No


yuzhu (approved):
LGTM, thanks

PR #13986, opened 2021-08-26
Upgrade netty to the stable version to fix the existing CVE

yangjf2019:

I modified the netty version in the pom.xml file. Upgrade netty to the stable version (netty-4.1.65.Final) to fix the existing [CVE](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=Flink). 

None.

None.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
     <glusterfs-hadoop.version>2.3.13</glusterfs-hadoop.version>
     <grpc.version>1.37.0</grpc.version>
     <gson.version>2.8.6</gson.version>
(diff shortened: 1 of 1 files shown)


ggezer:
Thanks @yangjf2019. We are currently investigating this on our. Unfortunately, a direct upgrade of netty will break compatibility with older spark versions.

yangjf2019:
Okay, Thanks for your review, then I will turn off this PR.

> Thanks @yangjf2019. We are currently investigating this on our. Unfortunately, a direct upgrade of netty will break compatibility with older spark versions.

Okay, Thanks for your review, then I will turn off this PR.

PR #13998, opened 2021-08-27
Add a skip-prepare parameter for MaxThroughput

maobaolong:

We'll do a same maxThroughput test many time, so we don't want to change basedir or create&delete a lots of file before test.

Add a parameter to supply a way to skip prepare step.

Add a new parameter `--skip-prepare`

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/suite/MaxThroughput.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/suite/MaxThroughput.java
@@ -73,7 +73,9 @@ public class MaxThroughput extends Suite<MaxThroughputSummary> {
 
     List<String> baseArgs = new ArrayList<>(Arrays.asList(args));
 
-    prepareBeforeAllTests(baseArgs);
+    if (mParameters.mSkipPrepare) {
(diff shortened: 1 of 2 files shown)


stress/shell/src/main/java/alluxio/stress/cli/suite/MaxThroughput.java
```diff
@@ -73,7 +73,9 @@ public MaxThroughputSummary runSuite(String[] args) throws Exception {
 
     List<String> baseArgs = new ArrayList<>(Arrays.asList(args));
 
-    prepareBeforeAllTests(baseArgs);
+    if (mParameters.mSkipPrepare) {
```
yuzhu, line 76: i think you meant `!mParameters.mSkipPrepare` here

maobaolong: Exactly, Thanks for remind me

maobaolong:
@yuzhu Thanks for remind me, I miss a `!` here, I fix  it by the latest commit, please take a another look.

yuzhu (approved):

yuzhu:

Issue #13758, opened 2021-07-08
Alluxio worker decommission


bzheng888:
**Is your feature request related to a problem? Please describe.**
Sometimes we may encounter situations like some worker nodes need to free up for another use, or we just want to offline some worker nodes to save costs. So, it  requires us to give a way to kick these workers out of the Alluxio cluster.
**Describe the solution you'd like**
I think maybe we should  add a feature “Worker Decommission”, the new feature will complete functions described as blow:
1. Provide a client shell command to kick out workers directly.
2. Workers to be decommissioned are defined in file conf/excluded_workers. 
3. Workers complete decommission when all the replicas from them are replicated to other workers.
4. Decommissioned worker nodes are not automatically shutdown and are not chosen for writing new replicas.
**Describe alternatives you've considered**
**Urgency**
Not urgency.
**Additional context**
An early design doc is mainly about the phase 1 work, left some useful comment and background.
https://docs.google.com/document/d/14ur7rAkr4KoA9Lmaqq5z7opE7xqwuD2g7Kh7TyJ5XkI/edit?usp=sharing
A more complete design doc is as blows:
https://docs.google.com/document/d/1ErVdZdDF3_qt7yFJiZAQvYoGBHr4uR-FdEOAkOvW4mI/edit?usp=sharing

maobaolong:
@apc999 Please take a look at this design docs.

apc999:
@ggezer  and @jiacheliu3  will be the owner of this functionality and review your design doc. 
I will chime in too. 

Thanks for the improvement!

LuQQiu:
From @apc999, worker send RPC to master to unregister worker, and master remove this worker from structures (will not assign this worker to fulfill client requests)

or From @maobaolong , master send RPC to worker to shutdown them

yuzhu:
@ggezer and @jiacheliu3 any thoughts on this design doc and the associated PR?

apc999:
@maobaolong @bzheng888 @jiacheliu3 let's have a discussion on this design of worker decommission before checking in code.The design doc looks pretty thin currently.

jiacheliu3:
@dbw9580 Is this issue relevant to your work at https://github.com/Alluxio/alluxio/pull/16134 ?

elega:
@bzheng888 are you still working on this?

bzheng888:
> @bzheng888 are you still working on this?

PR #14000, opened 2021-08-27
Add worker decommission step 1

bzheng888:
This pr is related to #13758 

This change achieves below functionalities:
1. Provide a client shell command to kick out workers directly.
2. Decommissioned worker nodes are not automatically shutdown and are not chosen for new IO requests
3. Decommissioned workers are not accepted on heartbeat or register

Note there's a delay in when clients know the workers have been decommissioned. In the window, target worker will still receive requests. There's also no way for the worker to come back to the cluster. This command should only be used when this worker is going to leave the cluster (and never come back).

Offline workers more flexible, no need to wait master detection lost worker interval.
For instance,
We may encounter situations like we want to remove worker nodes, or we want to offline some worker nodes to save costs, this command is useful.
Yes


shell/src/main/java/alluxio/cli/fsadmin/refreshworkers/StartDecommissionCommand.java
```diff
@@ -0,0 +1,136 @@
...
+  private static final Option EXCLUDED_HOSTS_OPTION =
+      Option.builder()
+          .longOpt("excluded-hosts")
+          .required(false)
+          .hasArg(true)
+          .numberOfArgs(1)
+          .argName("excluded-hosts")
+          .desc("A list of excluded worker hosts separated by comma.")
+          .build();
+  private static final Option EXCLUDED_WORKER_HOST =
```
maobaolong, line 53: Maybe we can use EXCLUDED_HOSTS_OPTION instead of `EXCLUDED_WORKER_HOST ` ?

shell/src/main/java/alluxio/cli/fsadmin/refreshworkers/StartDecommissionCommand.java
```diff
@@ -0,0 +1,136 @@
...
+    FileSystemAdminShellUtils.checkMasterClientService(mAlluxioConf);
+    boolean success;
+    try {
+      success = mBlockClient.startDecommission(excludedWorkerSet);
+      if (!success) {
+        System.out.println("Worker decommission failed!");
+        return -1;
+      }
+    } catch (IOException e) {
+      return -1;
```
maobaolong, line 109: Maybe we need print the e call stack to detect the root cause of this exception.

maobaolong (changes requested):
@bzheng888 Thanks for your contribution for this, it really helps us to remove a worker immediately, left some comments, please take a look.

shell/src/main/java/alluxio/cli/fsadmin/refreshworkers/StartDecommissionCommand.java
```diff
@@ -124,8 +112,7 @@ public String getCommandName() {
 
   @Override
   public String getUsage() {
-    return "start [-h] [--excluded-hosts <hostFilePath>] "
-        + "[--excluded-worker <workerHostName>]";
+    return "start [-h] [--excluded-hosts <host1> <host2>]";
```
jiacheliu3, line 115: Would it be better to use a separator between hostnames? In the current way anything appended to the end of this command will be treated as hostnames right? That's hard to extend in the future.

bzheng888: Thanks for pointing out the problem， the usage description is wrong but actually the hostnames is separated by comma, I'll fix it.

apc999:
I left some comments on the issue. 
@jiacheliu3 @bzheng888 @maobaolong , let's have a separate design review meeting first on the worker decommission  feature which is quite core and non-trivial.

bzheng888:
@maobaolong @jiacheliu3 Please review this pr again.

dbw9580:
As per our discussion today, I'd like to propose several changes to the design implemented in this PR:

1. config file and persistent `excluded_worker` list in block master
   IMO, we should not use a config file to specify which workers to decommission. Instead, what we need is a "args file" to contain the worker hostnames when invoking the command, e.g.

```console
$ alluxio fsadmin decommissionWorkers WORKER1 WORKER2
```
is equivalent to 

```console
$ cat ./to_decommission_workers
WORKER1
WORKER2
$ alluxio fsadmin decommissionWorkers --input-file ./to_decommission_workers
```

the `--input-file` option is analogous to `--arg-file` found in `xargs`, etc.

2. decommission status
   
   We should add a status command to see which workers are in the process of being decommissioned, and which have been decommissioned. These require two additional variables tracking the workers' states. The status command can possibly go to `alluxio fsadmin report`.

bzheng888:
@dbw9580 Thanks for reviewing this,  for you suggestion, I will submit a new pr today.

core/server/master/src/main/java/alluxio/master/block/BlockMaster.java
```diff
@@ -116,6 +117,11 @@ List<WorkerInfo> getWorkerReport(GetWorkerReportOptions options)
    */
   List<WorkerLostStorageInfo> getWorkerLostStorage();
 
+  /**
+   * @return the decommission status
+   */
+  DecommissionStatus getDecommissionStatus() throws UnavailableException;
```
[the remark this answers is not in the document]
dbw9580: There is a `getWorkerReport` method on `BlockMaster` which return a list of `WorkerInfo` object. We can add a decommission status to that `WorkerInfo` object.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -116,7 +117,8 @@ public final class MasterWorkerInfo {
   private static final Logger LOG = LoggerFactory.getLogger(MasterWorkerInfo.class);
   private static final String LIVE_WORKER_STATE = "In Service";
   private static final String LOST_WORKER_STATE = "Out of Service";
-
+  private static final String DECOMMISSION_WORKER_STATE = "In Progress";
```
dbw9580, line 120:
```suggestion
  private static final String DECOMMISSION_WORKER_STATE = "Decommission In Progress";
```

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1623,6 +1670,57 @@ public void registerNewWorkerConfListener(BiConsumer<Address, List<ConfigPropert
...
+  /**
+  * Decommission workers and update excluded worker map.
+  */
+  @Override
+  public void decommissionWorkers(final Set<String> excludedWorkerSet, boolean clear) {
+    Map<String, MasterWorkerInfo> tempWorkers = new HashMap<>();
+    if (excludedWorkerSet.isEmpty()) {
+      mDecommissionedWorkers.clear();
+      mDecommissioningWorkers.clear();
+      LOG.warn("Clear decommissioned worker map.");
```
dbw9580, line 1693: Why clearing the states for decommissioning workers? If I invoke the command with an empty list, does it make the master forget about all decommission-in-progress workers?

shell/src/main/java/alluxio/cli/fsadmin/decommission/StartDecommissionCommand.java
```diff
@@ -0,0 +1,160 @@
...
+          .argName("input-file")
+          .desc("The file specific workers to be decommissioned.")
+          .build();
+  private static final Option EXCLUDED_HOSTS_CLEAR =
+      Option.builder("clear")
+          .required(false)
+          .hasArg(false)
+          .desc("If true, clear the old decommissioned map from master, and update with entries "
+              + "collected this time, so the workers are not chosen this time will release "
+              + "from the decommissioned map, and they have a chance to reactive."
```
dbw9580, line 71: I don't understand what is this for? A decommissioned worker cannot "recover" from the decommissioned state unless it is restarted. Clearing the decommissioned worker set does not automatically restart the worker and make it re-register again.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1094,6 +1133,11 @@ public void workerRegister(long workerId, List<String> storageTiers,
       throw new NotFoundException(ExceptionMessage.NO_WORKER_FOUND.getMessage(workerId));
     }
 
+    // If this is an excluded worker, do not register any more.
+    if (isExcludedWorker(worker)) {
+      return;
+    }
```
dbw9580, line 1139: Since a worker only does registration when 1) at startup, 2) explicitly told by the master to do so via worker-master heartbeat, instead of rejecting the worker's registration and resetting the decommissioned state manually by the `--clear` flag later, how about simply allowing this worker to register, because when it does, it apparently has been restarted and can be put into normal service.

dbw9580 (changes requested):
Thanks for the update! Overall I think the design looks good.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
bzheng888: Yes, this just provide a way to reactive the decommissioned workers, but the more 
 graceful way is to do this with a clear shell command.

shell/src/main/java/alluxio/cli/fsadmin/decommission/StartDecommissionCommand.java
bzheng888: Same as above, the clear option give us a chance to reactive workers that not mentioned in this time. But using a shell command to do this is better choice.

bzheng888:
Hi, @dbw9580  @Tonytan123 What else can I do with this PR? or this pr is no longer needed?

maobaolong:
@bzheng888 @dbw9580 @jiacheliu3 Look forward to merge this PR, after that we can improve more base on this.

YichuanSun:
@bzheng888 @maobaolong Sorry, recently we are busy in adding new features and other trifles. This pr 14000 is needed. We have almost implemented the base decommissionWorker cmd in [PR 16134](https://github.com/Alluxio/alluxio/pull/16134), and this PR may be a CLI for the decommissionWorker cmd. I have published the design doc of decommissionWorker Command in pr 16134, comments are welcome.

bzheng888:
@YichuanSun Thanks for current work, I will continue to complete this pr and try to adapt your design doc and codes, hope we can communicate more in future.

YichuanSun:
This function is used to remove a live worker with [#16134](https://github.com/Alluxio/alluxio/pull/16134), do not stale it.

bzheng888:
@YichuanSun @dbw9580  Update this pr base on your design and implementation, can we complete the decommission feature base on this pr as the phase1? Please review this thanks!

YichuanSun:
Thanks Bing, I will leave some comments.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -301,10 +303,10 @@ public void addLostStorage(Map<String, StorageList> lostStorage) {
    * The required locks will be determined internally based on the fields.
    *
    * @param fieldRange the client selected fields
-   * @param isLiveWorker the worker is live or not
+   * @param workerState the worker state
    * @return generated worker information
    */
-  public WorkerInfo generateWorkerInfo(Set<WorkerInfoField> fieldRange, boolean isLiveWorker) {
+  public WorkerInfo generateWorkerInfo(Set<WorkerInfoField> fieldRange, WorkerState workerState) {
```
YichuanSun, line 309: This change need to be further reviewed.

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,127 @@
...
+    int timeoutMS = FileSystemShellUtils.getIntArg(cl, TIMEOUT_OPTION, DEFAULT_TIMEOUT);
+    String workerHost = cl.getOptionValue(HOST_OPTION.getLongOpt());
+
+    DecommissionWorkerPOptions options =
+            DecommissionWorkerPOptions.newBuilder()
+                .setTimeOut(timeoutMS).setForced(cl.hasOption("f")).build();
+
+    List<BlockWorkerInfo> cachedWorkers = mFsContext.getCachedWorkers();
+
+    for (BlockWorkerInfo blockWorkerInfo :cachedWorkers)  {
```
YichuanSun, line 84: It's better to modify this loop. If extract the decommission operation from traversal cachedWorkers, such as using an obj(initialise to NULL) to save the finding result, then judge whether the obj is NULL or not, the logic may be evident and clean.

YichuanSun (commented):
The code is clear, better than what I wrote, I have a few comments for your reference.

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
YichuanSun, line 84:
```suggestion
    BlockWorkerInfo tempBlockWorkerInfo = NULL;
    for (BlockWorkerInfo blockWorkerInfo : cachedWorkers) {
      if (Objects.equals(blockWorkerInfo.getNetAddress().getHost(), workerHost)) {
        tempBlockWorkerInfo = blockWorkerInfo;
        break;
      }
    }

    if (tempBlockWorkerInfo != NULL)  {
      try {
        long start = System.currentTimeMillis();
        mFileSystem.decommissionWorker(blockWorkerInfo.getNetAddress(), options);
        long duration = System.currentTimeMillis() - start;
        System.out.printf("Decommission worker %s success, spend: %dms%n",
            workerHost, duration);
      } catch (InterruptedException ie) {
        System.out.println(ie);
      }
    }
    else {
      System.out.println("Target worker is not found in Alluxio, please input another name.");
    }

    return 0;
```

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,127 @@
...
+              workerHost, duration);
+        } catch (InterruptedException ie) {
+          System.out.println(ie);
+        }
+        return 0;
+      }
+    }
+
+    System.out.println("Target worker is not found in Alluxio, please input another name.");
+    return 0;
```
YichuanSun, line 100: how about
```suggestion
    BlockWorkerInfo tempBlockWorkerInfo = NULL;
    for (BlockWorkerInfo blockWorkerInfo : cachedWorkers) {
      if (Objects.equals(blockWorkerInfo.getNetAddress().getHost(), workerHost)) {
        tempBlockWorkerInfo = blockWorkerInfo;
        break;
      }
    }

    if (tempBlockWorkerInfo != NULL)  {
      try {
        long start = System.currentTimeMillis();
        mFileSystem.decommissionWorker(blockWorkerInfo.getNetAddress(), options);
        long duration = System.currentTimeMillis() - start;
        System.out.printf("Decommission worker %s success, spend: %dms%n",
            workerHost, duration);
      } catch (InterruptedException ie) {
        System.out.println(ie);
      }
    }
    else {
      System.out.println("Target worker is not found in Alluxio, please input another name.");
    }

    return 0;
```

YichuanSun: The old code is like the code I wrote four months ago......

bzheng888: @YichuanSun My origin implementation is far different from your design, so I just copied your codes, can we let this pr be the first step to the decommission feature?

YichuanSun: > @YichuanSun My origin implementation is far different from your design, so I just copied your codes, can we let this pr be the first step to the decommission feature?

OK, but we need more exprienced alluxio engineer to review this PR. @dbw9580 Could you please take a look when you have time? thanks!

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,127 @@
...
+          System.out.printf("Decommission worker %s success, spend: %dms%n",
+              workerHost, duration);
+        } catch (InterruptedException ie) {
+          System.out.println(ie);
+        }
+        return 0;
+      }
+    }
+
+    System.out.println("Target worker is not found in Alluxio, please input another name.");
```
YichuanSun, line 99:
```suggestion
    BlockWorkerInfo tempBlockWorkerInfo = NULL;
    for (BlockWorkerInfo blockWorkerInfo : cachedWorkers) {
      if (Objects.equals(blockWorkerInfo.getNetAddress().getHost(), workerHost)) {
        tempBlockWorkerInfo = blockWorkerInfo;
        break;
      }
    }
    if (tempBlockWorkerInfo != NULL)  {
      try {
        long start = System.currentTimeMillis();
        mFileSystem.decommissionWorker(blockWorkerInfo.getNetAddress(), options);
        long duration = System.currentTimeMillis() - start;
        System.out.printf("Decommission worker %s success, spend: %dms%n",
            workerHost, duration);
      } catch (InterruptedException ie) {
        System.out.println(ie);
      }
    }
    else {
      System.out.println("Target worker is not found in Alluxio, please input another name.");
    }
```

YichuanSun (changes requested):

bzheng888:
@jiacheliu3 Hi jiacheng, can you take a look again plz, thx!

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -331,10 +333,22 @@ public WorkerInfo generateWorkerInfo(Set<WorkerInfoField> fieldRange, boolean is
           info.setStartTimeMs(mMeta.mStartTimeMs);
           break;
         case STATE:
-          if (isLiveWorker) {
-            info.setState(LIVE_WORKER_STATE);
-          } else {
-            info.setState(LOST_WORKER_STATE);
+          switch (workerState) {
+            case LIVE:
```
jiacheliu3, line 337: can we directly use the state field in `workerState`?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -5466,4 +5467,15 @@ public void needsSync(AlluxioURI path) throws InvalidPathException {
   public Scheduler getScheduler() {
     return mScheduler;
   }
+
+  @Override
+  public void decommissionWorker(String workerName,
```
jiacheliu3, line 5472: this is no longer needed

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -942,7 +959,9 @@ public void commitBlock(long workerId, long usedBytesOnTier, String tierAlias,
       throws NotFoundException, UnavailableException {
     LOG.debug("Commit block from workerId: {}, usedBytesOnTier: {}, blockId: {}, length: {}",
         workerId, usedBytesOnTier, blockId, length);
-
+    if (isExcludedWorker(workerId)) {
```
jiacheliu3, line 962: don't skip here, we need the BlockMeta journal entry

jiacheliu3 (changes requested):
Code mostly LGTM with some comments. Thanks for the contribution!

Could you add a few unit tests? Thanks!

core/client/fs/src/main/java/alluxio/client/block/BlockMasterClient.java
```diff
@@ -104,4 +106,12 @@ List<WorkerInfo> getWorkerReport(final GetWorkerReportOptions options)
    * @return amount of used space in bytes
    */
   long getUsedBytes() throws IOException;
+
+  /**
+   * Decommission a worker.
+   * @param options method options
+   * @throws AlluxioStatusException if something goes wrong
+   */
+  default void decommissionWorker(DecommissionWorkerPOptions options) throws IOException {
```
jiacheliu3, line 115: remove `default`

core/transport/src/main/proto/proto.lock
```diff
@@ -1,10828 +0,0 @@
-{
```
jiacheliu3, line 1: don't delete this?

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,128 @@
...
+ * Decommission a specific worker, the decommissioned worker is not automatically
+ * shutdown and are not chosen for writing new replicas.
+ */
+public final class DecommissionWorkerCommand extends AbstractFileSystemCommand {
+
+  private static final int DEFAULT_TIMEOUT = 10 * Constants.MINUTE_MS;
+  private static final Option TIMEOUT_OPTION =
+      Option.builder("t")
+          .longOpt("timeout")
+          .argName("timeout in milliseconds")
```
jiacheliu3, line 41: let's use seconds as the unit

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,128 @@
...
+          .argName("host")
+          .desc("A worker host name, which is mandatory.")
+          .build();
+
+  private static final Option FORCE_OPTION =
+      Option.builder("f")
+          .required(false)
+          .hasArg(false)
+          .desc("Force to decommission a worker.")
+          .build();
```
jiacheliu3, line 63: i don't think this option is used? Let's remove this?

jiacheliu3 (changes requested):

bzheng888:
@jiacheliu3 Update, PTAL!

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,119 @@
...
+          .longOpt("timeout")
+          .argName("timeout in seconds")
+          .numberOfArgs(1)
+          .desc("Timeout in milliseconds for decommissioning a"
+              + "single worker, default: " + DEFAULT_TIMEOUT)
+          .required(false)
+          .build();
+
+  private static final Option HOST_OPTION =
+      Option.builder("host")
```
jiacheliu3, line 49:
```suggestion
      Option.builder("h")
```

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,119 @@
...
+  private static final int DEFAULT_TIMEOUT = 10 * Constants.MINUTE_MS;
+  private static final Option TIMEOUT_OPTION =
+      Option.builder("t")
+          .longOpt("timeout")
+          .argName("timeout in seconds")
+          .numberOfArgs(1)
+          .desc("Timeout in milliseconds for decommissioning a"
+              + "single worker, default: " + DEFAULT_TIMEOUT)
+          .required(false)
+          .build();
```
jiacheliu3, line 46: on 2nd thought, i think this timeout is not very necessary. The master receives the call and just removes the block locations. This is not reversible so it doesn't really matter if you timeout on the client side or not. Even if your shell is dead, the worker will still be decommissioned.

let's remove this option then

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,119 @@
...
+          System.out.printf("Decommission worker %s success, spend: %dms%n",
+              workerHost, duration);
+        } catch (IOException ie) {
+          throw new AlluxioException(ie.getMessage());
+        }
+        return 0;
+      }
+    }
+
+    System.out.println("Target worker is not found in Alluxio, please input another name.");
```
jiacheliu3, line 94: If the target worker is not found, print the list of all cached workers, including the hostname and ports. Then the admin will understand what to input.

shell/src/main/java/alluxio/cli/fs/command/DecommissionWorkerCommand.java
```diff
@@ -0,0 +1,119 @@
...
+  @Override
+  public String getUsage() {
+    return "decommissionWorker [-t <max_wait_time>] --host <worker host>";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Decommission a specific worker synchronously, the decommissioned worker are not "
+        + "automatically shutdown and are not chosen for writing new replicas. The decommission "
+        + "worker command will return within a certain timeout no matter the process is complete.";
```
jiacheliu3, line 117:
```suggestion
    return "Decommission a specific worker in the Alluxio cluster. The decommissioned worker is not shut down but will not accept new read/write operations. The ongoing operations will proceed until completion.";
```

jiacheliu3 (changes requested):
Mostly LGTM, a final batch of minor comments. Thanks!

bzheng888:
image

jiacheliu3 (approved):
LGTM, thanks for the contribution!

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -815,9 +816,25 @@ public void removeBlocks(Collection<Long> blockIds, boolean delete) throws Unava
   }
 
   @Override
-  public void decommissionWorker(long workerId)
-      throws Exception {
-    //TODO(Tony Sun): added in another pr.
+  public boolean isServing(long workerId) {
```
YichuanSun, line 819:
```suggestion
  public boolean isServing(long workerId) {
```

I have a small question here. If the function of `isServing` is judge whether target worker is decommissioned, if it is decommissioned, return `True`, so why not called the function `isNotServing` or just `isDecommissioned`? I remember if a worker is decommissioned, it will not supply any services.

bzheng888: Make sense, thanks for pointing this!

bzheng888: @jiacheliu3 Rename the `isServing` method to `isNotServing`, FYI.

jiacheliu3:

PR #14001, opened 2021-08-30
Make distributed tasks delay time configurable

codings-dan:

Make distributed tasks delay time configurable.

When the number of nodes in the cluster is large, the delay time of 10s may cause incomplete task scheduling. The delay time should be flexibly configured as required

No

### DIFF
--- a/stress/common/src/main/java/alluxio/stress/BaseParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/BaseParameters.java
@@ -34,6 +34,7 @@ public final class BaseParameters {
   public static final long UNDEFINED_START_MS = -1;
   public static final String AGENT_OUTPUT_PATH = "/tmp/stress_client.log";
   public static final String DEFAULT_TASK_ID = "local-task-0";
+  public static final String START_DELAY_MS = "--start-delay";
 
   // Public flags
   @Parameter(names = {CLUSTER_FLAG},
@@ -86,4 +87,8 @@ public final class BaseParameters {
 
   @Parameter(names = {"-h", HELP_FLAG}, help = true)
   public boolean mHelp = false;
+
+  @Parameter(names = {START_DELAY_MS},
+          description = "The start delay for the distributed tasks, in ms.")
+  public long mStartDelay = 10000;
 }
--- a/stress/common/src/main/java/alluxio/stress/BaseParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/BaseParameters.java
@@ -50,7 +50,8 @@ public final class BaseParameters {
 
   @Parameter(names = {CLUSTER_START_DELAY_FLAG},
       description = "The start delay for the jobs to wait before starting the benchmark, "
-          + "used to synchronize the jobs")
+          + "used to synchronize the jobs. For example, --start-delay 10000ms,"
+              + "--start-delay 15s, --start-delay 2m")
   public String mClusterStartDelay = "10s";
 
   @Parameter(names = {JAVA_OPT_FLAG},
--- a/stress/common/src/main/java/alluxio/stress/BaseParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/BaseParameters.java
@@ -23,6 +23,7 @@ import java.util.List;
 public final class BaseParameters {
   public static final String CLUSTER_FLAG = "--cluster";
   public static final String CLUSTER_LIMIT_FLAG = "--cluster-limit";
+  public static final String CLUSTER_START_DELAY_FLAG = "--cluster-start-delay";
   public static final String DISTRIBUTED_FLAG = "--distributed";
   public static final String ID_FLAG = "--id";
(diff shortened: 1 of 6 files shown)


stress/common/src/main/java/alluxio/stress/BaseParameters.java
```diff
@@ -86,4 +87,8 @@
 
   @Parameter(names = {"-h", HELP_FLAG}, help = true)
   public boolean mHelp = false;
+
+  @Parameter(names = {START_DELAY_MS},
```
jiacheliu3, line 91: You can make it a string value here like
https://github.com/Alluxio/alluxio/blob/58555b7de4633c6ac57e20ae73ded7c169a6f8f9/stress/common/src/main/java/alluxio/stress/rpc/RpcBenchParameters.java#L26

jiacheliu3 (changes requested):
Mostly good to me except on comment, thanks!

codings-dan:
@jiacheliu3 I have fixed it,PTAL

stress/common/src/main/java/alluxio/stress/BaseParameters.java
```diff
@@ -90,5 +90,5 @@
 
   @Parameter(names = {START_DELAY_MS},
```
jiacheliu3, line 91: Parameter name

jiacheliu3 (changes requested):
Please double check the parameter name and all the references to make sure the variables align. Also could you add an example command to your PR description so people know how to use it? Thanks.

jiacheliu3:
@dbw9580 Could you update the documentation reflecting this change after the PR's merged? Thx :)

dbw9580:
@codings-dan thanks for your contribution. I made a commit to rename and limit the new flag to cluster only execution, as I don't think you meant to apply a start delay for in-process execution.

@jiacheliu3 
> Could you update the documentation reflecting this change after the PR's merged?

Will do.

codings-dan:
@dbw9580 thanks for the improvents.  @jiacheliu3 I  have add example in the description, PTAL

stress/common/src/main/java/alluxio/stress/BaseParameters.java
```diff
@@ -23,6 +23,7 @@
 public final class BaseParameters {
   public static final String CLUSTER_FLAG = "--cluster";
   public static final String CLUSTER_LIMIT_FLAG = "--cluster-limit";
+  public static final String CLUSTER_START_DELAY_FLAG = "--cluster-start-delay";
```
jiacheliu3, line 26: `CLUSTER_START_DELAY_FLAG` -> `CLUSTER_START_DELAY`

stress/common/src/main/java/alluxio/stress/BaseParameters.java
```diff
@@ -50,7 +50,8 @@
 
   @Parameter(names = {CLUSTER_START_DELAY_FLAG},
       description = "The start delay for the jobs to wait before starting the benchmark, "
-          + "used to synchronize the jobs")
+          + "used to synchronize the jobs. For example, --start-delay 10000ms,"
```
jiacheliu3, line 53: Make sure your example aligns with your code pls

jiacheliu3 (changes requested):
Please double check the parameter naming and comments. Also please add example usage in the description of this PR. Thx

stress/common/src/main/java/alluxio/stress/BaseParameters.java
jiacheliu3: @dbw9580

dbw9580: Why stripping the flag suffix? The other options have it.

jiacheliu3: I don't have the bible with me but oftentimes when a parameter is `xxx-flag` it suggests. the parameter is a boolean field. This parameter has a value so it's a bit more logical to not have a `-flag` suffix.

maobaolong (approved):
LGTM, thanks for the contribution.

stress/common/src/main/java/alluxio/stress/BaseParameters.java
dbw9580: Yeah I agree with that, but there are already a lot of options that are named `FLAG` but take an argument, e.g. `CLUSTER_LIMIT_FLAG` and `JAVA_OPT_FLAG`.

jiacheliu3: Ok seems there are arguments from both sides now so I'm ok either way. I will recommend removing `_FLAG` suffix from non-boolean parameters, but that should be done in another PR.

jiacheliu3 (approved):

dbw9580 (approved):
LGTM. Thanks!

jiacheliu3:

PR #14006, opened 2021-08-31
Support start container id from current timestamp

maobaolong:

Without this PR, alluxio master will generate block container id from `0`, after master format and restart without worker format, the old worker will report its block to alluxio master, not sure the master will accept the block by accident.

So this PR aimed to supply configure key to enable start block container id from current timestamp, so that the blockContainer Id will never possible to be the same value in a long period.

NO

### How to test

- start a alluxio cluster with the config key `alluxio.master.block.container.id.start.from.timestamp=true`

- I can see the following log
2021-08-31 18:56:56,708 INFO  DefaultBlockMaster - The start container id is 1630407414

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2675,6 +2675,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey
+      MASTER_BLOCK_CONTAINER_ID_START_FROM_TIMESTAMP =
+      new Builder(Name.MASTER_BLOCK_CONTAINER_ID_START_FROM_TIMESTAMP)
+          .setDescription("If enabled, the first container id is current timestamp")
+          .setScope(Scope.MASTER)
+          .setDefaultValue(false)
+          .build();
 
   //
   // Secondary master related properties
@@ -5834,6 +5841,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
         "alluxio.master.journal.gc.threshold";
     public static final String MASTER_JOURNAL_TEMPORARY_FILE_GC_THRESHOLD_MS =
         "alluxio.master.journal.temporary.file.gc.threshold";
+    public static final String MASTER_BLOCK_CONTAINER_ID_START_FROM_TIMESTAMP =
+        "alluxio.master.block.container.id.start.from.timestamp";
 
     //
     // File system master related properties
--- a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
@@ -355,7 +355,13 @@ public class DefaultBlockMaster extends CoreMaster implements BlockMaster {
   public void resetState() {
     mBlockStore.clear();
     mJournaledNextContainerId = 0;
-    mBlockContainerIdGenerator.setNextContainerId(0);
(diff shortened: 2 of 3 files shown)


maobaolong:
@LuQQiu This is the PR related the issue we just discussed in the fuse sync meeting, Please take a look, Thank you.

maobaolong:
I met a NPE, because the `InodeTreePersistentState#getRoot` method use hard code `0` to get root InodeDirectory.

```java
  public InodeDirectory getRoot() {
    return mInodeStore.get(0).map(Inode::asDirectory).orElse(null);
  }
```

maobaolong:
@LuQQiu I set `0` as root directory, now, it seems everything works as expected.

I create file and directory, restart alluxio cluster, everything works as expected.

image

image

apc999:
I think the issue reported by Baolong will happen.
To solve it, alternatively, when worker reporting blocks, also reporting the cluster ID?
after master reformated, the cluster ID will change in this way, master can tell if the worker belong to a totally different cluster deployment or not. 

what do you think @ggezer @jiacheliu3 @maobaolong 


btw, @maobaolong  can you create an github issue just reporting the bug?

jiacheliu3:
> I think the issue reported by Baolong will happen.
> To solve it, alternatively, when worker reporting blocks, also reporting the cluster ID?
> after master reformated, the cluster ID will change in this way, master can tell if the worker belong to a totally different cluster deployment or not.
> 
> what do you think @ggezer @jiacheliu3 @maobaolong
> 
> btw, @maobaolong can you create an github issue just reporting the bug?

When a worker registers, the master will reject any block that is not recognized. https://github.com/Alluxio/alluxio/blob/d2e9472438057146216374b46ece361c9acd4723/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java#L1072

That said, if the master is formatted but the worker is not, you will see DEBUG logs from the master complaining about the unrecognized block https://github.com/Alluxio/alluxio/blob/d2e9472438057146216374b46ece361c9acd4723/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java#L1082 And the master will later instruct the worker to remove those blocks (in the next heartbeat). 

Above is the easy case. Let's consider a more complicated case.

The worker has block `1`. The master just formatted and starts with `1`(`0` is the root inode). Now a client asks the master to create a file and occupied this `1`. Then the worker registers with this block which does not belong to this file. I think (not 100% and not tested) this block will be acknowledged. https://github.com/Alluxio/alluxio/blob/d2e9472438057146216374b46ece361c9acd4723/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java#L871 However, in order to be sure, I recommend you can write a test (you may start with integration tests in `alluxio/tests` which gives you a `LocalAlluxioCluster` to play with).

maobaolong:
@jiacheliu3 Thanks for the sharing code blocks, it will help me to know more about block report. But I really encountered the following log

```
2021-08-26 20:21:21,503 WARN  DefaultBlockMaster - Rejecting attempt to change block length from 757049 to 925
```

The related code blocks is

image

I'm not sure what will happen if the two length are happen to be same.

jiacheliu3:
@maobaolong If would be more helpful if the `DefaultBlockMaster` log printed the blockId (and ideally the current vs new block information), but unfortunately the log message is not that helpful.

Yea I think your example can be due to this issue.

If we are to avoid this issue, I think the alternative way of checking the cluster ID (proposed by @apc999 ) is easier for people to follow and manage. The cluster ID is a UUID generated in https://github.com/Alluxio/alluxio/blob/d2e9472438057146216374b46ece361c9acd4723/core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java#L331 and will be journaled. In other words the cluster ID only changes when master starts with fresh journal.

PR #14008, opened 2021-08-31
Enhance and stabilize the `quorum elect` command

jenoudet:
The `quorum elect` command can be slow, hang, incorrectly display error messages, and has many small shortcomings. This PR tackles those problems.

No


maobaolong:
@codings-dan Would you please take a look at this PR? Thanks

codings-dan:
@jenoudet If resetPriority is not executed after the transfer leadership is completed, will subsequent elections be affected?

ggezer (changes requested):
Thanks for the improvement. 
Can you also add a test for validating the auto-reset behavior you observed?

jenoudet:
> @jenoudet If resetPriority is not executed after the transfer leadership is completed, will subsequent elections be affected?

@codings-dan In our testing we have observed that resetting priorities after the transfer of leadership has no impact on the peer priorities.  To ensure that removing this final step has no impact on subsequent elections (and the overall well being of the cluster) we've added new tests to cover these use cases. These test are contained within this PR.

jenoudet (commented):
Added a test that checks that the peer priorities are reset to 1 for all members, except the new leader to 0.

ggezer (approved):

ggezer:

PR #14010, opened 2021-08-31
Move prepare stage to local in stress bench

ssz1997:
Currently in stress bench, prepare() is called before the job is distributed to job service if cluster flag is turned on. Therefore the node that runs the command will run prepare() once independently from the actual job unnecessarily. This could mess up the stress bench test. 
This PR makes the call to prepare() to a later stage where the commands are only ran locally. 

@dbw9580 Is this what you mean? This problem is messing up my test so I need to fix it now.
@yuzhu PTAL, thanks!

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java
@@ -74,9 +74,6 @@ public abstract class Benchmark<T extends TaskResult> {
   /**
    * Prepares to run the test.
    */
-  // TODO(bowen): When the test runs in cluster mode, the prepare step will execute
-  //  both in the command side and on each job worker side. We should separate the logic
-  //  into two different calls instead of relying on the same prepare().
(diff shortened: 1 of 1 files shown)


dbw9580:
Yes this was the headache I had a while back😺. I haven't really come up with a reasonable refactoring plan that doesn't break the existing benchmarks easily because there are too many benchmarks that depends on the abstract benchmark framework.

The workaround is to test the state variables, e.g. `mBaseParameters.mDistributed`. If the preparations are only meant to run once at the caller side, put them in a `if (!mBaseParameters.mDistributed)` clause. If they are to run at each worker's side, revert the condition.

I'm afraid simply reordering the preparation step might break other benchmarks. Please be extra sure about this change, and if possible, try if the workaround meets your needs first.

ssz1997:
@dbw9580 I was also kind of worried about it. I think the workaround you mentioned should be good for my problem for now. Thanks a lot!

Close this for now.

PR #14017, opened 2021-09-02
Close gRPC input stream when finished reading

### LINKED ISSUE #14020: Thousands of gRPC connection slow down machine learning read
**Alluxio Version:**
Alluxio 2.6.1

**Describe the bug**
When using DALI data loader https://developer.nvidia.com/dali to load imagenet data. The dataset is about 1.2 millions file * 105KB on average per file = 130GB total data. DALI loads data in batches, doing some machine learning training, and then close the input streams in batches (related script can be found https://github.com/LuQQiu/DALI/blob/master/docs/examples/use_cases/pytorch/resnet50/main.py)

In normal read workload, a file is open()， read（), and then immediately close(), which the current code works well.
In machine learning world, a file is open(), read（）, wait for machine learning training job to finish, and then close(). When reading a large number of small files, many gRPC stream will be opened concurrently and consumes a lot of CPU resources. By closing the gRPC stream immediately after finishing reading can release gRPC stream earlier and improve overall read performance.

**To Reproduce**
- Set up Alluxio on Kubernetes cluster with worker fuse
- Load imagenet data into Alluxio
- Run the machine learning training job. Here is a mocked sample
```
arena --loglevel info submit pytorch \
  --name=test-job \
  --gpus=0 \
  --workers=4 \
  --cpu 6 \
  --memory 32G \
  --selector alluxio-master=false \
  --image=nvcr.io/nvidia/pytorch:21.05-py3 \
  --data-dir=/alluxio/ \
  --sync-mode=git \
  --sync-source=https://github.com/LuQQiu/DALI.git \
  "python /root/code/DALI/docs/examples/use_cases/pytorch/resnet50/main.py \
  --process 4 \
  --batch-size 256 \
  --print-freq 1000 \
  /alluxio/alluxio-mountpoint/alluxio-fuse/dali"
```
- compare the result between the current impl and close stream immediately after reading
**Expected behavior**

**Urgency**

**Additional context**

LuQQiu:
Fixes #14020
Close gRPC input stream immediately after reading all the data from worker.
In some machine learning task, they read files in batch, and close files after reading in batch. The time between they finished reading and close input stream may be relatively long. Each moment, there may be thousands of gRPC stream waiting to be closed. Close gRPC stream immediately after reading all the data can help release resources earlier and improve millions of small file Fuse read throughput. 
No.

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -326,6 +326,9 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
     byteBuffer.position(off).limit(off + toRead);
     mCurrentChunk.readBytes(byteBuffer);
     mPos += toRead;
+    if (mPos == mLength) {
+      closeDataReader();
+    }
     return toRead;
   }
 
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -330,6 +330,7 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
     mCurrentChunk.readBytes(byteBuffer);
     mPos += toRead;
     if (mPos == mLength) {
+      // a performance improvement introduced by https://github.com/Alluxio/alluxio/issues/14020
       closeDataReader();
     }
     return toRead;
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
@@ -311,6 +311,9 @@ public class BlockInStream extends InputStream implements BoundedStream, Seekabl
     if (len == 0) {
       return 0;
     }
+    if (mPos == mLength) {
+      return -1;
+    }
     readChunk();
     if (mCurrentChunk == null) {
       mEOF = true;
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
@@ -174,7 +174,12 @@ public final class GrpcDataReader implements DataReader {
       // finish reading all the data, close gRPC stream earlier to release resources
       // potentially improve machine learning millions of small file read throughput
       LOG.debug("Finished reading all data, close gRPC stream");
-      close();
+      try {
+        close();
+      } catch (Throwable t) {
+        buffer.release();
+        throw t;
+      }
       return buffer;
     }
     try {
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
@@ -170,7 +170,7 @@ public final class GrpcDataReader implements DataReader {
       return null;
     }
     mPosToRead += buffer.readableBytes();
-    if (mPosToRead == mReadRequest.getLength()) {
+    if (mPosToRead - mReadRequest.getOffset() == mReadRequest.getLength()) {
       // finish reading all the data, close gRPC stream earlier to release resources
       // potentially improve machine learning millions of small file read throughput
       LOG.debug("Finished reading all data, close gRPC stream");
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
@@ -170,18 +170,6 @@ public final class GrpcDataReader implements DataReader {
       return null;
     }
     mPosToRead += buffer.readableBytes();
-    if (mPosToRead - mReadRequest.getOffset() == mReadRequest.getLength()) {
-      // finish reading all the data, close gRPC stream earlier to release resources
-      // potentially improve machine learning millions of small file read throughput
-      LOG.debug("Finished reading all data, close gRPC stream");
-      try {
-        close();
-      } catch (Throwable t) {
-        buffer.release();
-        throw t;
-      }
-      return buffer;
-    }
     try {
       mStream.send(mReadRequest.toBuilder().setOffsetReceived(mPosToRead).build());
     } catch (Exception e) {
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
@@ -171,7 +171,8 @@ public final class GrpcDataReader implements DataReader {
     }
     mPosToRead += buffer.readableBytes();
     if (mPosToRead == mReadRequest.getLength()) {
-      // finish reading all the data, close gRPC stream earlier to release resources. Helpful in millions of small file Fuse read case
+      // finish reading all the data, close gRPC stream earlier to release resources
+      // potentially improve machine learning millions of small file read throughput
       LOG.debug("Finished reading all data, close gRPC stream");
       close();
       return buffer;
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
@@ -170,6 +170,12 @@ public final class GrpcDataReader implements DataReader {
       return null;
     }
     mPosToRead += buffer.readableBytes();
+    if (mPosToRead == mReadRequest.getLength()) {
+      // finish reading all the data, close gRPC stream earlier to release resources. Helpful in millions of small file Fuse read case
+      LOG.debug("Finished reading all data, close gRPC stream");
+      close();
+      return buffer;
+    }
     try {
       mStream.send(mReadRequest.toBuilder().setOffsetReceived(mPosToRead).build());
     } catch (Exception e) {
--- a/core/client/fs/src/test/java/alluxio/client/block/stream/BlockInStreamTest.java
+++ b/core/client/fs/src/test/java/alluxio/client/block/stream/BlockInStreamTest.java
@@ -12,6 +12,9 @@
 package alluxio.client.block.stream;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.when;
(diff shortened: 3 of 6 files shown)


LuQQiu:
Error: 1.968 [ERROR] Errors: 
Error: 1.968 [ERROR] alluxio.server.ft.MultiWorkerIntegrationTest.readRecoverFromLostWorker
Error: 1.969 [ERROR]   Run 1: MultiWorkerIntegrationTest.readRecoverFromLostWorker:125 » IllegalState Block ...
Error: 1.971 [ERROR]   Run 2: MultiWorkerIntegrationTest.readRecoverFromLostWorker:125 » IllegalState Block ...
20:57:31.971 [INFO] 
Warning: 971 [WARNING] Flakes: 
Warning: 971 [WARNING] alluxio.server.ft.ZookeeperFailureIntegrationTest.zkConnectionPolicy_Standard
Error: 1.972 [ERROR]   Run 1: ZookeeperFailureIntegrationTest.zkConnectionPolicy_Standard:125 Values should be different. Actual: 0

LuQQiu:
Error: 4.811 [ERROR] Errors: 
Error: 4.812 [ERROR] alluxio.server.ft.MultiWorkerIntegrationTest.readRecoverFromLostWorker
Error: 4.813 [ERROR]   Run 1: MultiWorkerIntegrationTest.readRecoverFromLostWorker:125 » IllegalState Block ...
Error: 4.813 [ERROR]   Run 2: MultiWorkerIntegrationTest.readRecoverFromLostWorker:125 » IllegalState Block ...

LuQQiu:
Error: 0.307 [ERROR] alluxio.master.AlluxioMasterProcessTest.startStopPrimary
Error: 0.307 [ERROR]   Run 1: AlluxioMasterProcessTest.startStopPrimary:90->startStopTest:235
Error: 0.308 [ERROR]   Run 2: AlluxioMasterProcessTest.startStopPrimary:90->startStopTest:235

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
```diff
@@ -170,6 +170,13 @@ private DataBuffer readChunkInternal() throws IOException {
...
     mPosToRead += buffer.readableBytes();
+    if (mPosToRead - mReadRequest.getOffset() == mReadRequest.getLength()) {
+      // finish reading all the data, close gRPC stream earlier to release resources
+      // potentially improve machine learning millions of small file read throughput
+      LOG.debug("Finished reading all data, close gRPC stream");
+      close();
+      return buffer;
+    }
     try {
       mStream.send(mReadRequest.toBuilder().setOffsetReceived(mPosToRead).build());
```
yuzhu, line 181: unrelated to this PR, but could we send this receipt asynchronously to return the data to the caller faster? not sure if this would even make a difference, but just a thought.

yuzhu (approved):
The PR LGTM, left a comment which might be another possible optimization.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
LuQQiu: For the small file use case, they only read one chunk and no need to send offset received. Without the fixes in this PR, client will send the offset received to worker, but worker won't stop setting data until a gap larger than 4MB. Not sure is it possible or helpful to do it asynchronously, maybe @ggezer can have a better idea.

ggezer: This call would effectively be asynchronous by setting `alluxio.user.streaming.reader.close.timeout` to `0`.

ggezer (commented):
@LuQQiu  How much improvement have you seen with this change?
My understanding is, if this fix helped with throughput, then setting `alluxio.user.streaming.reader.close.timeout` to `0` should have had the same effect. So I'm a little confused on the impact of this change.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
yuzhu: not familiar with this part of the code, but it seems to me that it is only asynchronous if there is one chunk to read right? if there are multiple chunks, is this send not in the critical path of returning the data to the caller?

ggezer: By asynchronous i meant, caller wouldn't have to wait at all since onCompleted() returns immediately. 
I agree that async execution of receipts would be another level of optimization.

ggezer (approved):

LuQQiu:
<google-sheets-html-origin><style type="text/css"></style>

Node CPU | Arena CPU | Arena Process | Alluxio Worker CPU | Arena sleep time | Use the close gRPC stream after reading first chunk tarball | Throughput
-- | -- | -- | -- | -- | -- | --
16 | 10 | 8 | 4 | 0.5 | TRUE | 7641 img/s (190MB/s/node)
16 | 10 | 8 | 4 | 0.5 | FALSE | 4711 img/s


@ggezer , i didn't do too much testing, but one of the testing i do shows that with enough resources, without the changes in this PR, the performance drops from 7.6k img/s to 4.7k img/s

LuQQiu:
@ggezer I could relaunch the cluster and test the `alluxio.user.streaming.reader.close.timeout=0`, could be it blocked on CPU resources before and not anymore

LuQQiu:
@ggezer `alluxio.user.streaming.reader.close.timeout = 0` do totally different things.
For GrpcDataReader, it has create(), readChunk(), close() functions.
In normal cases, create() -> readChunk() -> close() will be really close to each other, no much time elaspe between readChunk and close().

In DALI machine learning env, create() -> readChunk() -> 10 seconds -> close(). 

The `alluxio.user.streaming.reader.close.timeout = 0` help us do the `close()` much quicker but didn't help us to close stream in readChunk（）    -> create() -> readChunk() -> 10 seconds -> close(MUCH QUICKER).

This PR helps us `close()` in readChunk() -> 
 create() -> readChunk(DIRECTLY CLOSE AFTER READING FIRST CHUNK) -> 10 seconds -> close().

ggezer:
> @ggezer `alluxio.user.streaming.reader.close.timeout = 0` do totally different things.
> For GrpcDataReader, it has create(), readChunk(), close() functions.
> In normal cases, create() -> readChunk() -> close() will be really close to each other, no much time elaspe between readChunk and close().
> 
> In DALI machine learning env, create() -> readChunk() -> 10 seconds -> close().
> 
> The `alluxio.user.streaming.reader.close.timeout = 0` help us do the `close()` much quicker but didn't help us to close stream in readChunk（） -> create() -> readChunk() -> 10 seconds -> close(MUCH QUICKER).
> 
> This PR helps us `close()` in readChunk() ->
> create() -> readChunk(DIRECTLY CLOSE AFTER READING ALL CHUNKS) -> 10 seconds -> close().

Some questions:
-Did the last validation (with this PR) also had `alluxio.user.streaming.reader.close.timeout = 0` ?
-Why there is ` 10 seconds` wait?

LuQQiu:
@ggezer I validated without `alluxio.user.streaming.reader.close.timeout = 0`.
The 10 seconds are from machine learning logics in which the DALI data reader reads 256files, each on average 100KB in batches, and then we do a sleep(5second) to mock the machine learning training, and then DALI close those files in batches

LuQQiu:
@apc999 Do you have time to review this PR?

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
```diff
@@ -170,6 +170,13 @@ private DataBuffer readChunkInternal() throws IOException {
       return null;
     }
     mPosToRead += buffer.readableBytes();
+    if (mPosToRead - mReadRequest.getOffset() == mReadRequest.getLength()) {
+      // finish reading all the data, close gRPC stream earlier to release resources
+      // potentially improve machine learning millions of small file read throughput
+      LOG.debug("Finished reading all data, close gRPC stream");
+      close();
```
apc999, line 177: @ggezer @LuQQiu 

1 buffer here is built based upon `response.getChunk().getData().asReadOnlyByteBuffer()`.
if we close the stream before returning the buffer, is the reference to this buffer still safe?

2 what if `close` throws an exception, do we need to release any resource related to buffer?

apc999 (changes requested):
one comment left

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
LuQQiu: The 1 worry is about whether gRPC stream close() will release the Chunk in ReadResponse. I don't think it will, but better to be confirmed by @ggezer. I didn't see any code or doc indicate that scenario.

Release buffer when close throw an exception.
Considering still return the byteBuffer to client if close throws exception, but in that case, do we just rethrow that exception when the actual close() is called by the client

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
```diff
@@ -174,7 +174,12 @@ private DataBuffer readChunkInternal() throws IOException {
       // finish reading all the data, close gRPC stream earlier to release resources
       // potentially improve machine learning millions of small file read throughput
       LOG.debug("Finished reading all data, close gRPC stream");
-      close();
+      try {
+        close();
+      } catch (Throwable t) {
+        buffer.release();
```
ggezer, line 180: We had the buffer already. Why not just return it to the caller and then the caller owns releasing it after consumption.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
```diff
@@ -170,6 +170,18 @@ private DataBuffer readChunkInternal() throws IOException {
       return null;
     }
     mPosToRead += buffer.readableBytes();
+    if (mPosToRead - mReadRequest.getOffset() == mReadRequest.getLength()) {
```
ggezer, line 173: On second thought; can we do proactive closing in the BlockInStream rather than here? 
That way, you can also consume the buffer before calling close(), and effectively achieve what @yuzhu was asking.

core/client/fs/src/main/java/alluxio/client/block/stream/GrpcDataReader.java
ggezer: Currently, grpc data reader offloads releasing to the caller. By moving proactive closing logic to BlockInStream level, you wouldn't have to deal with this as well.

LuQQiu:
@apc999 @ggezer Move the logic to BlockInStream. when we read up to length then we close the stream, trick it as EOF, PTAL, thanks

ggezer (approved):
Thanks for the improvement!

apc999 (approved):
can you have UTs to cover the two corner cases added here.
Other than that, LGTM and the code looks much easier to reason.

Thanks!

core/client/fs/src/main/java/alluxio/client/block/stream/BlockInStream.java
```diff
@@ -326,6 +329,9 @@ public int read(ByteBuffer byteBuffer, int off, int len) throws IOException {
     byteBuffer.position(off).limit(off + toRead);
     mCurrentChunk.readBytes(byteBuffer);
     mPos += toRead;
+    if (mPos == mLength) {
+      closeDataReader();
```
apc999, line 333: add javadoc for tricky improvement, so in the future we will pay attention when touching this piece of code.

```suggestion
      // reference the performance issue on github
      closeDataReader();
```

apc999:
ping @LuQQiu

LuQQiu:
Thanks @apc999 ! working on the unit test

LuQQiu:
@apc999 PTAL the added unit test, thanks

apc999 (approved):

apc999:

PR #14018, opened 2021-09-02
Add three types of read and cluster mode in Fuse Stress Bench

ssz1997:

1. Implement [`LocalRead`, `RemoteRead`, `ClusterRead`] in cluster mode in Fuse IO Stress Bench.
2. Create FuseIOSummary class for aggregated results of all workers in the cluster.
3. Remove ThreadCount class under FuseIOTaskResult class to simplify the logic.

1. [design doc](https://docs.google.com/document/d/1kmrOh6blvm1fEIILnIvG-a_GbFkqwYsAkVei2GiJ6fg/edit#heading=h.tkbrhe1f6jlp)
2. Using the new FuseIOSummary class we can show the aggregated results better and more concisely. This is also adopted by master stress bench etc. where multiple nodes are used.
3. The original ThreadCount class is adopted from clientIOStressBench, where one run of the bench can test different number of threads, and each ThreadCount object is the result of one particular number of threads. FuseIOStressBench doesn't support testing multiple number of threads, so we can get rid of ThreadCount class to simplify the design.

Users could run FuseIOStressBench in a cluster with job service. `Read` operation also expands to [`LocalRead`, `RemoteRead`, `ClusterRead`]

### DIFF
--- a/stress/common/src/main/java/alluxio/stress/fuse/FuseIOTaskResult.java
+++ b/stress/common/src/main/java/alluxio/stress/fuse/FuseIOTaskResult.java
@@ -12,47 +12,81 @@
 package alluxio.stress.fuse;
 
 import alluxio.Constants;
-import alluxio.collections.Pair;
-
 import alluxio.stress.BaseParameters;
-import alluxio.stress.Parameters;
-import alluxio.stress.Summary;
 import alluxio.stress.TaskResult;
-import alluxio.stress.common.SummaryStatistics;
-import alluxio.stress.graph.Graph;
-import alluxio.stress.graph.LineGraph;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.google.common.base.Splitter;
 
-import java.io.IOException;
-import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 /**
  * The task result for the Fuse IO stress tests.
  */
-public final class FuseIOTaskResult implements TaskResult, Summary {
+public final class FuseIOTaskResult implements TaskResult {
   private long mRecordStartMs;
   private long mEndMs;
-  private Map<Integer, ThreadCountResult> mThreadCountResults;
+  private long mIOBytes;
+  private List<String> mErrors;
   private BaseParameters mBaseParameters;
   private FuseIOParameters mParameters;
-
-  private Map<Integer, Map<String, SummaryStatistics>> mTimeToFirstByte;
 
   /**
    * Creates an instance.
    */
   public FuseIOTaskResult() {
-    // Default constructor required for json deserialization
-    mThreadCountResults = new HashMap<>();
-    mTimeToFirstByte = new HashMap<>();
+    mErrors = new ArrayList<>();
+  }
+
+  @Override
+  public TaskResult.Aggregator aggregator() {
+    return new Aggregator();
+  }
+
+  private static final class Aggregator implements TaskResult.Aggregator<FuseIOTaskResult> {
+    @Override
+    public FuseIOSummary aggregate(Iterable<FuseIOTaskResult> results) throws Exception {
+      long recordStartMs = 0;
+      long endMs = 0;
+      long ioBytes = 0;
+      FuseIOParameters fuseIOParameters = null;
+      BaseParameters baseParameters = null;
+      List<String> nodes = new ArrayList<>();
+      Map<String, List<String>> errors = new HashMap<>();
+      Map<String, Float> individualThroughput = new HashMap<>();
+
+      for (FuseIOTaskResult taskResult: results) {
+        recordStartMs = taskResult.getRecordStartMs();
+        endMs = Math.max(endMs, taskResult.getEndMs());
+        ioBytes += taskResult.getIOBytes();
+        fuseIOParameters = taskResult.getParameters();
+        baseParameters = taskResult.getBaseParameters();
+
+        String workerUniqueId = taskResult.getBaseParameters().mId;
+        nodes.add(workerUniqueId);
+        individualThroughput.put(workerUniqueId, taskResult.getIOMBps());
+        errors.put(workerUniqueId, taskResult.getErrors());
+      }
+
+      float ioMBps = (float) ioBytes / (endMs - recordStartMs) * 1000.0f / Constants.MB;
+
+      return new FuseIOSummary(fuseIOParameters, baseParameters, nodes, errors, recordStartMs,
+          endMs, ioBytes, ioMBps, individualThroughput);
+    }
+  }
+
+  /**
+   * Merges this thread result into the Fuse IO task result.
+   *
+   * @param result  the result to merge
+   */
+  public void merge(FuseIOTaskResult result) {
+    mRecordStartMs = Math.min(mRecordStartMs, result.mRecordStartMs);
+    mEndMs = Math.max(mEndMs, result.mEndMs);
+    mIOBytes += result.mIOBytes;
+    mErrors.addAll(result.mErrors);
   }
 
   /**
@@ -70,17 +104,55 @@ public final class FuseIOTaskResult implements TaskResult, Summary {
   }
 
   /**
-   * @return the Fuse IO parameters
+   * @return the Fuse IO Stress Bench parameters
    */
   public FuseIOParameters getParameters() {
     return mParameters;
   }
 
   /**
-   * @param parameters the Fuse IO parameters
+   * @param parameters the Fuse IO Stress Bench parameters
    */
   public void setParameters(FuseIOParameters parameters) {
     mParameters = parameters;
+  }
+
+  /**
+   * @return the duration (in ms)
+   */
+  public long getDurationMs() {
+    return mEndMs - mRecordStartMs;
+  }
+
+  /**
+   * @param durationMs the duration (in ms)
+   */
+  @JsonIgnore
+  public void setDurationMs(long durationMs) {
+    // ignore
+  }
+
+  /**
+   * @return bytes of IO
+   */
+  public long getIOBytes() {
+    return mIOBytes;
+  }
+
+  /**
+   * Increments the bytes of IO an amount.
+   *
+   * @param ioBytes the amount to increment by
+   */
+  public void incrementIOBytes(long ioBytes) {
+    mIOBytes += ioBytes;
+  }
+
+  /**
+   * @param ioBytes bytes of IO
+   */
+  public void setIOBytes(long ioBytes) {
+    mIOBytes = ioBytes;
   }
 
   /**
@@ -98,30 +170,6 @@ public final class FuseIOTaskResult implements TaskResult, Summary {
   }
 
   /**
-   * @return client IO statistics per method
-   */
-  public Map<Integer, Map<String, SummaryStatistics>> getTimeToFirstBytePerThread() {
-    return mTimeToFirstByte;
-  }
-
-  /**
-   * @param timeToFirstByte time to first statistics
-   */
-  public void setTimeToFirstBytePerThread(Map<Integer, Map<String,
-      SummaryStatistics>> timeToFirstByte) {
-    mTimeToFirstByte = timeToFirstByte;
-  }
-
-  /**
-   * @param numThreads thread count
-   * @param statistics FuseIOTaskResultStatistics
-   */
-  public void putTimeToFirstBytePerThread(Integer numThreads,
-      Map<String, SummaryStatistics> statistics) {
-    mTimeToFirstByte.put(numThreads, statistics);
-  }
-
-  /**
    * @return the end time (in ms)
    */
   public long getEndMs() {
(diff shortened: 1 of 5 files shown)


LuQQiu:
@ssz1997 we discussed during the AI sync yesterday that we prefer to have one issue with one or more PR addressing that issue. For this PR, can you open an issue and put more detailed description/design link in that issue? and link the issue in your PR description

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -54,6 +59,10 @@
   @ParametersDelegate
   private FuseIOParameters mParameters = new FuseIOParameters();
 
+  /** Names of the directories created by workers for the test, also unique ids of the workers. */
+  private List<String> mWorkerDirNames;
+  /** 0-based id of this worker/job worker. */
+  private int mWorkerZeroBasedId;
```
LuQQiu, line 65: For Fuse stressbench, all it knows is the job worker's hostname and id.
And it knows which local folder or Fuse mount point it's write/list/read to/from.

So this job worker, this stressbench test doesn't know whether this node has a worker with Fuse, or a standalone Fuse, or nothing at all

Try to use more precise name

LuQQiu (commented):
As discussed, we try to directly tell what Fuse single/cluster mode testing do in this Fuse test and give out suggestions about how people actually doing it in Alluxio doc. 

Let's put it as a section in docs/en/operation/StressBench.md  for now, and then later we can think of where to put it and how to put those docs. @yuzhu any thought on where to put the doc of a single stressbench test?

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -75,68 +84,114 @@ public String getBenchDescription() {
...
     if (mParameters.mOperation == FuseIOOperation.WRITE) {
       LOG.warn("Cannot write repeatedly, so warmup is not possible. Setting warmup to 0s.");
       mParameters.mWarmup = "0s";
       for (int i = 0; i < mParameters.mNumDirs; i++) {
-        Files.createDirectories(Paths.get(mParameters.mLocalPath + "/" + i));
+        Files.createDirectories(Paths.get(String.format(
+            "%s/%s/dir-%d", mParameters.mLocalPath, mBaseParameters.mId, i)));
+      }
+    } else {
+      if ((mParameters.mOperation == FuseIOOperation.REMOTE_READ
```
LuQQiu, line 144: Put parameters check in the main function instead of in the prepare stage.
It doesn't belong to prepare(), it's an issue with wrong inputs

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -75,68 +84,114 @@ public String getBenchDescription() {
...
+      }
+      mWorkerDirNames = Arrays.asList(workerDirs).stream()
+              .map(file -> file.getName())
+              .collect(Collectors.toList());
+      try {
+        mWorkerZeroBasedId = mWorkerDirNames.indexOf(mBaseParameters.mId);
+      } catch (Exception e) {
+        throw new IllegalStateException(String.format(
+            "Directory %s is not found. Please use this bench to generate test files. If this is "
+                + "the case, some worker crashed or joined after data is written. The test is "
```
LuQQiu, line 177: Remove `if this is the case` and reconstruct the line

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -75,68 +84,114 @@ public String getBenchDescription() {
...
+      }
+      // Find 0-based id, and make sure dirs and workers are 1-to-1.
+      File[] workerDirs = localPath.listFiles();
+      if (workerDirs == null) {
+        throw new IOException(String.format(
+            "--local-path %s is not a valid path for this bench. Make sure using the correct path",
+                mParameters.mLocalPath
+        ));
+      }
+      int numWorkers;
```
LuQQiu, line 159: all the parameter names use jobWorker instead of worker

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -320,32 +375,38 @@ private void runInternal() throws Exception {
       mStartBarrierPassed = true;
 
       if (mParameters.mOperation == FuseIOOperation.LIST_FILE) {
+        for (String workerDirNames: mWorkerDirNames) {
+          for (int dirId = mThreadId; dirId < mParameters.mNumDirs; dirId += mParameters.mThreads) {
+            String dirPath = String.format("%s/%s/dir-%d", mParameters.mLocalPath,
```
LuQQiu, line 380: %s/%s/dir-%d make the format a constant since it's used in many places

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -320,32 +375,38 @@ private void runInternal() throws Exception {
       mStartBarrierPassed = true;
 
       if (mParameters.mOperation == FuseIOOperation.LIST_FILE) {
```
LuQQiu, line 377: same try not to use big if else statement

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -320,32 +375,38 @@ private void runInternal() throws Exception {
...
+          }
+        }
+      } else if (mParameters.mOperation == FuseIOOperation.WRITE
+          || mParameters.mOperation == FuseIOOperation.LOCAL_READ) {
         for (int dirId = mThreadId; dirId < mParameters.mNumDirs; dirId += mParameters.mThreads) {
-          String dirPath = String.format("%s/%d", mParameters.mLocalPath, dirId);
-          File dir = new File(dirPath);
-          dir.listFiles();
+          for (int fileId = 0; fileId < mParameters.mNumFilesPerDir; fileId++) {
+            String filePath = String.format("%s/%s/dir-%d/file-%d", mParameters.mLocalPath,
```
LuQQiu, line 390: for small files, i think string concatenation could largely impact the throughput performance
any ways to improve

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -320,32 +375,38 @@ private void runInternal() throws Exception {
...
-            // Start recording after the warmup
-            if (CommonUtils.getCurrentMs() > recordMs) {
-              mThreadCountResult.incrementIOBytes(ioBytes);
+      } else {
+        for (int dirRead = 0; dirRead < mWorkerDirNames.size(); dirRead++) {
+          int indexCurrentDir = (dirRead + mWorkerZeroBasedId) % mWorkerDirNames.size();
+          // Skip itself if the operation is remote read.
+          if (indexCurrentDir == mWorkerZeroBasedId
+              && mParameters.mOperation == FuseIOOperation.REMOTE_READ) {
+            indexCurrentDir++;
```
LuQQiu, line 401: if it's the last dir, indexCurrentDir++ would cause the mWorkerDirNames indexOutOfBound exception.
Double careful about some corner cases

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -320,32 +375,38 @@ private void runInternal() throws Exception {
...
-
-            // Done reading/writing one file
-            if (ioBytes <= 0) {
-              break;
-            }
-            // Start recording after the warmup
-            if (CommonUtils.getCurrentMs() > recordMs) {
-              mThreadCountResult.incrementIOBytes(ioBytes);
+      } else {
+        for (int dirRead = 0; dirRead < mWorkerDirNames.size(); dirRead++) {
```
LuQQiu, line 396: dirRead and dirId name are not informative enough
modify the name to show more clear info

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -320,32 +375,38 @@ private void runInternal() throws Exception {
...
+          int indexCurrentDir = (dirRead + mWorkerZeroBasedId) % mWorkerDirNames.size();
+          // Skip itself if the operation is remote read.
+          if (indexCurrentDir == mWorkerZeroBasedId
+              && mParameters.mOperation == FuseIOOperation.REMOTE_READ) {
+            indexCurrentDir++;
+          }
+          String nameCurrentDir = mWorkerDirNames.get(indexCurrentDir);
+          for (int dirId = mThreadId; dirId < mParameters.mNumDirs; dirId += mParameters.mThreads) {
+            for (int fileId = mWorkerZeroBasedId; fileId < mParameters.mNumFilesPerDir;
```
LuQQiu, line 405: for these two for loops, can it swap:
now we use thread id to decide dirId, and worker id to decide fileId,
can we do it like use worker id to decide fileId, and thread id to decide fileId

what's the difference between these two approaches? the reason that i suggest swapping is that it will be more engineer intuitive to go from bigger concepts (worker id, dir id) to smaller concepts together (thread id, file id)

LuQQiu (commented):
Thanks for supporting cluster read mode in Fuse stress bench! Left some comments

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -75,68 +84,114 @@ public String getBenchDescription() {
...
+        "--file-size 100m --threads 16 --warmup 5s --duration 30s --cluster",
         ""
     ));
   }
 
   @Override
   public void prepare() throws Exception {
+    if (mBaseParameters.mCluster) {
+      // For cluster mode, this function is called once before the job is submitted to the job
+      // service. Nothing should be done, otherwise the bench will break.
```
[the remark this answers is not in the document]
ssz1997: It is always called right after the parameters are parsed, before the benchmark figures out whether it is cluster mode or not (looking at --cluster flag). Therefore, if it's cluster mode, the node where we issue the runClass will run prepare() once before the job is sent to job service. Then every job worker will run prepare() themselves. 
Prepare() is called [here](https://github.com/Alluxio/alluxio/blob/8eb180b2329641682a821c923245219117a62cc4/stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java#L150) on line 150 but figuring out if it's cluster mode is [here](https://github.com/Alluxio/alluxio/blob/8eb180b2329641682a821c923245219117a62cc4/stress/shell/src/main/java/alluxio/stress/cli/Benchmark.java#L160) on line 160.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: It hasn't parsed the parameters yet at main() time. The prepare function is the earliest time after the parameters get parsed.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: To open the file we kind of have to have the path string. To construct this string I think this is the best we can do.

ssz1997: Seems like we could use either "+" or StringBuilder, which are basically the same. https://stackoverflow.com/questions/513600/should-i-use-javas-string-format-if-performance-is-important

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: Makes sense. This is definitely more intuitive.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: Using StringBuilder will look like this:
```
String filePath = new StringBuilder(mParameters.mLocalPath)
        .append("/")
        .append(mBaseParameters.mId)
        .append("/dir-")
        .append(testDirId)
        .append("/file-")
        .append(testFileId)
        .toString();
```
Using "+" will look like this:
```
String filePath = mParameters.mLocalPath + "/" + mBaseParameters.mId + "/dir-"
        + testDirId + "/file-" + testFileId;
```
I think either is fine. What do you think? @LuQQiu

LuQQiu: I don't have pretty good suggestions. We can keep the current impl, just to remind us that if future we see some performance number that is not expected, it could be the string concatenation issue especially when we are testing small files

ssz1997: It is surprising that Stirng.format() is >10 times slower than the other two.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: For this one I put the chunks of code into three separate functions and use a switch to call them, because even if we use the "if ... return; do something else" structure, it is still very large.

ssz1997:
@LuQQiu Addressed all the comments. PTAL, thanks!

stress/common/src/main/java/alluxio/stress/fuse/FuseIOTaskResult.java
```diff
@@ -136,303 +184,37 @@ public void setEndMs(long endMs) {
   }
 
   /**
-   * @return the map of thread counts to results
+   * @return the list of errors
    */
-  public Map<Integer, ThreadCountResult> getThreadCountResults() {
-    return mThreadCountResults;
+  public List<String> getErrors() {
+    return mErrors;
```
apc999, line 190: beter to return a immutable version of the errors, if you don't want to external caller to modify the returned errors
e.g., `Collections.unmodifiableList(mErrors)`

apc999 (changes requested):
had a quick pass with some minor comments.
thanks for the improvement

ssz1997:
@apc999 @LuQQiu Addressed all comments. PTAL, thanks!

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -319,42 +375,94 @@ private void runInternal() throws Exception {
...
+          numJobWorkerDirProcessed++) {
+        // find which job worker directory to read
+        int indexCurrentJobWorkerDir = (numJobWorkerDirProcessed + mJobWorkerZeroBasedId)
+            % mJobWorkerDirNames.size();
+        // skip itself if the operation is remote read
+        if (indexCurrentJobWorkerDir == mJobWorkerZeroBasedId
+            && mParameters.mOperation == FuseIOOperation.REMOTE_READ) {
+          indexCurrentJobWorkerDir = (indexCurrentJobWorkerDir + 1) % mJobWorkerDirNames.size();
+        }
+        String nameCurrentJobWorkerDir = mJobWorkerDirNames.get(indexCurrentJobWorkerDir);
```
LuQQiu, line 433: Consider 4 job workers, mJobWorkerZeroBasedId = 1 Remote_Read
numJobWorkerDirProcessed = 0
indexCurrentJobWorkerDir = (0 + 1) / 4 = 1
indexCurrentJobWorkerDir == mJobWorkerZeroBasedId
            && mParameters.mOperation == FuseIOOperation.REMOTE_READ
indexCurrentJobWorkerDir = (1 + 1) % 4;  = 2
nameCurrentJobWorkerDir = ...dir2

numJobWorkerDirProcessed = 1
int indexCurrentJobWorkerDir = (1 + 1)
            % mJobWorkerDirNames.size(); = 2
String nameCurrentJobWorkerDir = dir2

did we read the same dir for twice?

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -319,42 +375,94 @@ private void runInternal() throws Exception {
...
+
+        // find which files to read under this job worker directory
+        for (int testDirId = mJobWorkerZeroBasedId; testDirId < mParameters.mNumDirs;
+            testDirId += mJobWorkerDirNames.size()) {
+          for (int testFileId = mThreadId; testFileId < mParameters.mNumDirs;
+              testFileId += mParameters.mThreads) {
+            String filePath = String.format(TEST_DIR_STRING_FORMAT + "/file-%d",
```
LuQQiu, line 440: Use constant

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -319,42 +375,94 @@ private void runInternal() throws Exception {
...
+            processFile(filePath, FuseIOOperation.isRead(mParameters.mOperation));
           }
         }
       }
-      // Done reading all files
-      if (isRead) {
-        throw new IllegalArgumentException(String.format("Thread %d finishes reading all its files "
-            + "before the bench ends. For more accurate result, use more files, or larger files, "
-            + "or a shorter duration", mThreadId));
+      finishProcessingFiles();
```
LuQQiu, line 446: always error out?

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -319,42 +375,94 @@ private void runInternal() throws Exception {
...
+    }
+
+    private void processFile(String filePath, boolean isRead) throws IOException {
+      mCurrentOffset = 0;
+      while (!Thread.currentThread().isInterrupted()) {
+        if (isRead && CommonUtils.getCurrentMs() > mContext.getEndMs()) {
+          closeInStream();
+          return;
```
LuQQiu, line 454: we stop processing this file but didn't stop processing the following files
if current > end, we should stop processing all the files

LuQQiu (commented):
@ssz1997 Are you able to set up a running cluster to test out the cluster read/remote read? It may be helpful to add logs about which exact file is read by which job worker in which thread
From what i see in the code, i don't think the code work like what we expect

stress/common/src/main/java/alluxio/stress/fuse/FuseIOTaskResult.java
```diff
@@ -12,47 +12,82 @@
...
+  }
+
+  @Override
+  public TaskResult.Aggregator aggregator() {
+    return new Aggregator();
+  }
+
+  private static final class Aggregator implements TaskResult.Aggregator<FuseIOTaskResult> {
+    @Override
+    public FuseIOSummary aggregate(Iterable<FuseIOTaskResult> results) throws Exception {
```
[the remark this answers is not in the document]
ssz1997: The logic here is similar to WorkerBenchTaskResult merge(), but update more information. FYI @yuzhu

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -75,68 +85,116 @@ public String getBenchDescription() {
...
+    }
+    if (numJobWorkers != jobWorkerDirs.length) {
+      throw new IllegalStateException("Some worker crashed or joined after data are written. "
+          + "The test is stopped.");
+    }
+    mJobWorkerDirNames = Arrays.asList(jobWorkerDirs).stream()
+        .map(file -> file.getName())
+        .collect(Collectors.toList());
+    try {
+      mJobWorkerZeroBasedId = mJobWorkerDirNames.indexOf(mBaseParameters.mId);
```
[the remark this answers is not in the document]
ssz1997: For cluster mode, the time when it is called before figuring out it's cluster mode, it will enter https://github.com/Alluxio/alluxio/blob/5a14e865dda2248db30a692bc4056d4403db6045/stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java#L127-L129 and does nothing. The whole prepare() method is only useful when the job workers are executing.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: The directories at this level are created by job workers for distinguishing what files are written by whom. They are not `--num-dirs` given by users. And we agreed that the remote read is like cluster read, where each job worker will read from files created by every other job workers.

LuQQiu: Yeam but we read dirBelongingToJobWorker2 for twice, isn't it?

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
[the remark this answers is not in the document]
LuQQiu: oh got it, so it's like prepare(cluster_mode) -> submit to job service without cluster_mode tag -> prepare() in each job worker without cluster mode.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: Need a flag to tell the caller whether the return is because of time up or done processing file. If the caller sees time up flag is on, it should stop and also not error out.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
ssz1997: Here https://github.com/Alluxio/alluxio/blob/5a14e865dda2248db30a692bc4056d4403db6045/stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java#L436-L437 no job worker will go into the same dir.

ssz1997: Ah I see...For remote read, I'm going to pretend it done reading the files it wrote without actually doing any reading.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -319,42 +375,94 @@ private void runInternal() throws Exception {
...
+        }
+        String nameCurrentJobWorkerDir = mJobWorkerDirNames.get(indexCurrentJobWorkerDir);
+
+        // find which files to read under this job worker directory
+        for (int testDirId = mJobWorkerZeroBasedId; testDirId < mParameters.mNumDirs;
+            testDirId += mJobWorkerDirNames.size()) {
+          for (int testFileId = mThreadId; testFileId < mParameters.mNumDirs;
+              testFileId += mParameters.mThreads) {
+            String filePath = String.format(TEST_DIR_STRING_FORMAT + "/file-%d",
+                mParameters.mLocalPath, mBaseParameters.mId, testDirId, testFileId);
```
ssz1997, line 441: mId here is incorrect that it will always read from itself

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -319,51 +376,123 @@ private void runInternal() throws Exception {
...
+        if (indexCurrentJobWorkerDir == mJobWorkerZeroBasedId
+            && mParameters.mOperation == FuseIOOperation.REMOTE_READ) {
+          continue;
+        }
+        String nameCurrentJobWorkerDir = mJobWorkerDirNames.get(indexCurrentJobWorkerDir);
+
+        // find which files to read under this job worker directory
+        for (int testDirId = mJobWorkerZeroBasedId; testDirId < mParameters.mNumDirs;
+            testDirId += mJobWorkerDirNames.size()) {
+          for (int testFileId = mThreadId; testFileId < mParameters.mNumDirs;
```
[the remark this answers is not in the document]
ssz1997: I did print out the logs but didn't realize it's missing files...I really need to pay more attention to the correctness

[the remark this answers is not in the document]
LuQQiu: This kind of complicated scenario is easy to have typos and bugs, that's why tests are important. No worries

LuQQiu (approved):
LGTM, please address the merge conflicts

LuQQiu:
@apc999 PTAL if you have time, otherwise this PR will be merged by Sep 17, thanks!

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -153,30 +153,34 @@ public void prepare() throws Exception {
...
     }
-    // find 0-based id, and make sure directories and job workers are 1-to-1
     File[] jobWorkerDirs = localPath.listFiles();
     if (jobWorkerDirs == null) {
       throw new IOException(String.format(
           "--local-path %s is not a valid path for this bench. Make sure using the correct path",
               mParameters.mLocalPath
       ));
     }
+    if (!mBaseParameters.mDistributed) {
```
ssz1997, line 163: Previously single-node mode will ask for number of live job worker. This is unnecessary and will break the test if the number of job worker is not 1.

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -153,30 +153,34 @@ public void prepare() throws Exception {
...
+      throw new IllegalStateException("Some job worker crashed or joined after data are written. "
           + "The test is stopped.");
     }
     mJobWorkerDirNames = Arrays.asList(jobWorkerDirs).stream()
         .map(file -> file.getName())
         .collect(Collectors.toList());
-    try {
-      mJobWorkerZeroBasedId = mJobWorkerDirNames.indexOf(mBaseParameters.mId);
-    } catch (Exception e) {
+    mJobWorkerZeroBasedId = mJobWorkerDirNames.indexOf(mBaseParameters.mId);
```
ssz1997, line 182: indexOf will return -1 if element not found. Thought it would throw exception.

ssz1997 (commented):
@LuQQiu PTAL. Thanks.

LuQQiu (approved):
LGTM, thanks for fixing it

LuQQiu:

Issue #14030, opened 2021-09-07
Add a  command to manually reset the priority


codings-dan:
**Is your feature request related to a problem? Please describe.**
When transfer leader, in some scenarios, the priority may produce an exception. For example, I encountered this scenario during the switching process,

`bin/alluxio fsadmin journal quorum info -domain MASTER           
Journal domain  : MASTER
Quorum size     : 3
Quorum leader   : localhost:19201

STATE       | PRIORITY | SERVER ADDRESS
UNAVAILABLE | 2        | localhost:19200
AVAILABLE   | 0        | localhost:19201
AVAILABLE   | 1        | localhost:19202`

I hope to provide a manual reset priority command as a solution

**Describe the solution you'd like**
Add a  command to manually reset the priority

**Urgency**
No

PR #14031, opened 2021-09-07
Add a command to manually reset the priority

codings-dan:

According to #14030 ,add a command to manually reset the priority

I hope to provide a manual reset priority command as a solution

No

### DIFF
--- a/shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumCommand.java
+++ b/shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumCommand.java
@@ -38,6 +38,7 @@ public class QuorumCommand extends AbstractFsAdminCommand {
     SUB_COMMANDS.put("info", QuorumInfoCommand::new);
     SUB_COMMANDS.put("remove", QuorumRemoveCommand::new);
     SUB_COMMANDS.put("elect", QuorumElectCommand::new);
+    SUB_COMMANDS.put("reset",QuorumResetPriorityCommand::new);
   }
 
   private Map<String, Command> mSubCommands = new HashMap<>();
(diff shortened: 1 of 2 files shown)


codings-dan:
> A `SetPrioritiesRequest` to a Ratis cluster seems to only have an effect if immediately followed by a `TransferLeadershipRequest`, as discussed in [this PR](https://github.com/Alluxio/alluxio/pull/14008#issuecomment-914633767). This separate command therefore seems unnecessary and is why we remove the `resetPriorities` call from the `elect` command in the first place.
Please take a look at the situation encountered in the issue #14030 . I think that in some scenarios, the priority of masters may be in an abnormal state. This command can be used as a solution for this scenario, so I think it is valuable to add this command.

codings-dan:
@jenoudet Please take a look at the situation encountered in my issue #14030 . I think that in some scenarios, the priority of masters may be in an abnormal state. This command can be used as a solution for this scenario, so I think it is valuable to add this command.

maobaolong:
@jenoudet @codings-dan For compromise, I suggest to add a `raft` top level sub-command as debug purpose, to collect the commands like this resetPriority command which aim to get or set raft internal state. 

@codings-dan This command it worth to have, but we could think twice if we should put this command under `journal` command.

ggezer:
@codings-dan Did you verify if having this command helped with #14030?

codings-dan:
@ggezer That scene is not easy to reproduce, so I can try it when I reproduce it again. Now I can only confirm that this command can be used normally

ggezer:
@maobaolong @codings-dan @jenoudet
How about we get a setPriority() calls instead? It could have a -reset flag that performs this exact function. 

@jenoudet will setting priorities have an effect on next natural leader election?

jenoudet:
@ggezer I will test this scenario and tell you if it has an impact or not.

codings-dan:
@jenoudet I noticed that in #14273, you added the resetPriority logic when transferLeader, do we also need to add a manual reset priority command?

PR #14035, opened 2021-09-07
Create smaller alpine and centos development docker image

LuQQiu:
We introduce two dockerfiles: Dockerfile and Dockerfile-dev. Dockerfile is for production use based on Alpine, where we try to minimize the size of the image as much as possible. Dockerfile-dev is for development use based on CentOS, where more tools are provided. dockerfile-common.sh contains some common commands used both for Dockerfile and Dockerfile-dev for better maintainability.

On one hand, docker image with smaller size uses less time to build, is easier to run, and takes less space, which is why we have the Dockerfile. On the other hand, many users need to do further development, where more tools are needed, which is why we have the Dockerfile-dev.

- To build the Dockerfile-dev, need to include `-t Dockerfile-dev`. 
- Ubuntu, which is the base operating system for the old Dockerfile, is replaced by Alpine in Dockerfile and CentOS in Dockerfile-dev


LuQQiu:
@ssz1997 Can you help modify the PR description and review the `Fix conflict` commit?
This PR brings docker changes back to master, thanks

ssz1997:
@LuQQiu LGTM. Thanks for fixing the conflicts!

LuQQiu:

LuQQiu:
alluxio-bot, feature-merge this please

ZhuTopher:
@Xenorith @ssz1997 I _think_ I might have found the issue causing the latest Kubernetes-related Autobots failures:

See [this StackOverflow post](https://stackoverflow.com/a/56528840) for details but I think what's happened is our [entrypoint.sh](https://github.com/Alluxio/alluxio/blob/master/integration/docker/entrypoint.sh) uses `bash` which is not available by default in Alpine.

So either change `#!/usr/bin/env bash` to `#!/bin/sh` or (probably more preferably so other shell scripts we use don't break too) add the following in the Dockerfile:
```
apk add --no-cache bash
```
---
**Context**
When investigating the Autobots clusters Kubernetes logs I saw this error:
```
$ kubectl logs --previous alluxio-master-0 -c alluxio-master
standard_init_linux.go:219: exec user process caused: no such file or directory
```
which led me to the linked StackOverflow post. I tried messing with some configuration stuff in k8s and never managed to get the containers to start, so I think this is the most probable explanation.

PR #14037, opened 2021-09-08
[SMALLFIX] Fix some test case under Windows

qian0817:

fix some test case under winows.

When I run the test under windows I found that some test cases under `AlluxioURITest` can't pass. This PR fixes the problem.

No user facing change.


core/common/src/main/java/alluxio/AlluxioURI.java
```diff
@@ -492,7 +492,7 @@ public String toString() {
       String path = mUri.getPath();
       if (path.indexOf('/') == 0 && hasWindowsDrive(path, true) && // has windows drive
           mUri.getScheme() == null && // but no scheme
-          mUri.getAuthority() == null) { // or authority
+          mUri.getAuthority() instanceof NoAuthority) { // or authority
```
apc999, line 495: shall we keep null check too>?
```suggestion
          (mUri.getAuthority() == null || mUri.getAuthority() instanceof NoAuthority)) { // or authority
```

apc999 (approved):

LuQQiu:

PR #14044, opened 2021-09-08
Add property to skip listing broken symlinks on local UFS

ZacBlanco:

Add a property to skip broken symlinks when calling listStatus with the LocalUnderFileSystem

When a broken symlink appears when using a local UFS, all listings will fail because of the additional call to `#readProperties` in `LocalUndeFileSystem#listStatus`. This behavior can help alert users to broken symlinks, but also breaks functionality until the symlink is removed, which may be undesirable.

A new option that can be used with a local UFS mount.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -779,6 +779,15 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey UNDERFS_LOCAL_SKIP_BROKEN_SYMLINKS =
+      new Builder(Name.UNDERFS_LOCAL_SKIP_BROKEN_SYMLINKS)
+          .setDefaultValue(false)
+          .setDescription("When set to true, any time the local underfs lists a broken "
+              + "symlink, it will treat the entry as if it didn't exist at all."
+              + "")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
+          .build();
   public static final PropertyKey UNDERFS_WEB_HEADER_LAST_MODIFIED =
       new Builder(Name.UNDERFS_WEB_HEADER_LAST_MODIFIED)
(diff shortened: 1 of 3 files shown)


ggezer (approved):
Thanks for this improvement.

ggezer:
On second thought; could you also treat  getStatus() result under this flag?

ZacBlanco:
In the case of `getStatus`, what would you think the expected behavior should be? There's no status to actually retrieve if a symlink is broken, so if we were to "skip" it, what would the getStatus result look like?

ggezer:
> In the case of `getStatus`, what would you think the expected behavior should be? There's no status to actually retrieve if a symlink is broken, so if we were to "skip" it, what would the getStatus result look like?

`FileNotFoundException` thrown with an internal message saying the symlink is broken.

ZacBlanco:
So the `getStatus` already returns with a `FileNotFound` exception even before this change on a broken symlink, so you just want to add that to the error message? I don't think that should be controlled by this configuration flag though

ggezer:
> So the `getStatus` already returns with a `FileNotFound` exception even before this change on a broken symlink. So you just want to add the symlink portion to the error message? I don't think that should be controlled by this configuration flag though

Yeah right. Well, regardless it would still be useful to include a better message. I don't know the current exc message on broken symlink, so you can be the judge whether to improve the message or leave it as it is.

ZacBlanco:
This is what the current stacktrace looks like for a broken symlink:

```
java.io.FileNotFoundException: /tmp/junit8779467026714153574/test
	at alluxio.underfs.local.LocalUnderFileSystem.getStatus(LocalUnderFileSystem.java:267)
	at alluxio.underfs.UnderFileSystemWithLogging$27.call(UnderFileSystemWithLogging.java:618)
	at alluxio.underfs.UnderFileSystemWithLogging$27.call(UnderFileSystemWithLogging.java:615)
	at alluxio.underfs.UnderFileSystemWithLogging.call(UnderFileSystemWithLogging.java:1213)
	at alluxio.underfs.UnderFileSystemWithLogging.getStatus(UnderFileSystemWithLogging.java:615)
```

Even though it lacks the information about the path being a symlink, I still think that the user should be able to figure out that the link is broken quite easily. Plus, doing the extra check when it isn't found requires more calls which slows down the error path.

ggezer:

ggezer:

PR #14046, opened 2021-09-09
Fix a null pointer exception

bzheng888:
Fix a null pointer problem in journal system.
When construct the RaftJournalSystem, the member mStateMachine is null pointer, because the registerMetrics is called before mStateMachine initialized。
No
No

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -396,7 +396,6 @@ public class RaftJournalSystem extends AbstractJournalSystem {
         .setProperties(properties)
         .setParameters(parameters)
         .build();
-    super.registerMetrics();
   }
 
   /**
@@ -725,6 +724,7 @@ public class RaftJournalSystem extends AbstractJournalSystem {
         .collect(Collectors.toSet());
     mRaftGroup = RaftGroup.valueOf(RAFT_GROUP_ID, peers);
     initServer();
+    super.registerMetrics();
     List<InetSocketAddress> clusterAddresses = mConf.getClusterAddresses();
     LOG.info("Starting Raft journal system. Cluster addresses: {}. Local address: {}",
         clusterAddresses, mConf.getLocalAddress());
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -527,8 +527,11 @@ public class RaftJournalSystem extends AbstractJournalSystem {
 
   @Override
   public synchronized Map<String, Long> getCurrentSequenceNumbers() {
-    long currentGlobalState = mStateMachine.getLastAppliedSequenceNumber();
     Map<String, Long> sequenceMap = new HashMap<>();
+    if (mStateMachine == null) {
+      return sequenceMap;
+    }
+    long currentGlobalState = mStateMachine.getLastAppliedSequenceNumber();
     for (String master : mJournals.keySet()) {
       // Return the same global sequence for each master.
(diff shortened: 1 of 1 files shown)


apc999 (approved):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -527,8 +527,11 @@ public synchronized void losePrimacy() {
 
   @Override
   public synchronized Map<String, Long> getCurrentSequenceNumbers() {
-    long currentGlobalState = mStateMachine.getLastAppliedSequenceNumber();
     Map<String, Long> sequenceMap = new HashMap<>();
+    if (mStateMachine == null) {
```
ggezer, line 531: We shouldn't return empty sequences. Please replace this with below:

`Preconditions.checkState(mStateMachine != 1, "State machine not initialized");
`

ggezer (changes requested):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
apc999: then is this more like a bug and check why `mStateMachine` is `null`?
@bzheng888  have you seen this in your environment that `mStateMachine` is `null`?

bzheng888: > then is this more like a bug and check why `mStateMachine` is `null`?
> @bzheng888 have you seen this in your environment that `mStateMachine` is `null`?
Yeah, I debugged in my environment and it's really a must present bug, I think the registerMetrics() in the construct method  is useless, the metrics is never taken effect.

bzheng888: Yeah, I debugged in my environment and it's really a must present bug, I think the registerMetrics() in the construct method is useless, the metrics is never taken effect.

maobaolong: @bzheng888 So would you like to move the `registerMetrics` to the end of `initServer` ?

bzheng888:
@apc999 @ggezer @maobaolong Move the call of registerMetrics() after stateMachine initialized,  this fix is more appropriate, PTAL!

ggezer (approved):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -401,6 +396,7 @@ private synchronized void initServer() throws IOException {
         .setProperties(properties)
         .setParameters(parameters)
         .build();
+    super.registerMetrics();
```
ggezer, line 399: is it OK to reregister at every gainPrimacy() call?

ggezer (changes requested):

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
bzheng888: image
the register strategy in registerMetrics() is ifAbsent, so it doesn't seem to reregister every time.

apc999:
any more comments, @ggezer ?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
ggezer: OK. Still, please move this line after `initServer();` call in `RaftJournalSystem.startInternal()` method.

bzheng888:
@ggezer Done, PTAL!

ggezer (approved):
Thanks for the fix.

ggezer:

PR #14050, opened 2021-09-10
Fix test startStopPrimary failure from time to time

yyongycy:
Fix test startStopPrimary failure from time to time


core/server/master/src/test/java/alluxio/master/AlluxioMasterProcessTest.java
```diff
@@ -230,6 +230,10 @@ private void startStopTest(AlluxioMasterProcess master) throws Exception {
     waitForServing(ServiceType.MASTER_WEB);
     assertTrue(isBound(mRpcPort));
     assertTrue(isBound(mWebPort));
+    boolean testMode = ServerConfiguration.getBoolean(PropertyKey.TEST_MODE);
```
ggezer, line 233: In the before() method we reset configuration for each test. So no need to remember and reset.
Also, I don't see why you need to explicitly set it to false as it's false by default.

yyongycy: From the debug info, default value is true. And to avoid impacting rest of the test, set it to original value after finished the wait.

image

ggezer (approved):
Thanks!

ggezer:

PR #14052, opened 2021-09-10
Replace monitor from proc to port.

ljl1988com:
Replace monitor from proc to port.

Better to replace monitor from proc to port.
There are two mainly reasons:

1. Every server ssh port can be different. Not default 22. And that must be trusting with each other hosts.
2. There is a proc on a host server. Not indicating that the master server service is working.

None.


ZacBlanco (commented):
So one reason that we simply checked the process rather than the port previously, is that in multi-master deployments, masters who were not the leader, did not have open RPC ports. Thus, for any deployments with more than one master, the check would fail even if the non-primary masters _were_ alive.

So before I review further, has the behavior with respect to open RPC ports on non-primary masters changed?

ljl1988com:
> So one reason that we simply checked the process rather than the port previously, is that in multi-master deployments, masters who were not the leader, did not have open RPC ports. Thus, for any deployments with more than one master, the check would fail even if the non-primary masters _were_ alive.
> 
> So before I review further, has the behavior with respect to open RPC ports on non-primary masters changed?

Thanks for your review.
I just read the doc and code. I find there is three types of master in the doc [https://docs.alluxio.io/os/user/edge/en/overview/Architecture.html#masters](url). Then I test running three types of master. Only the secondary master doesn't open RPC ports. But a secondary master is on the same server as the leading master. Anyhow, we can replace monitor from proc to port.
@ZacBlanco PTAL. Thanks.

ZhuTopher:
Commenting so I can watch this PR. Thanks!

ggezer:
@ljl1988com How did you test this? I don't think this will work. Are you bringing this change from a feature branch where you have the secondaries open their RPC servers somehow?

ljl1988com:
@ggezer I think the leading master and standby masters and secondary masters are like the design of `hdfs namenode`. The  secondary master should not open RPC server. If secondary open rpc server, it works like standby master. I'm not bringing this change. I‘m not familiar with the working principle of secondary. I only test two conditions:

1. The leading master and standby masters starting service  are the same. so I test running ha mode.
2. To secondary master. I only test on local mode. Start leading and secondary with the same `alluxio.master.journal.folder` on local machine.

Can you give me some advise? Thanks.

ggezer:
@ljl1988com This change is described to be an improvement I fail to see any required change here. There's nothing I can provide as I believe this PR builds on expectations that are not compatible with Alluxio. To keep the discussion going, I'd recommend opening an github issue and explaining what problems are you trying to overcome.

ljl1988com:
@ggezer I open an github issue #14121 . I find I misunderstand the codes. I want to check every master server is running.

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
```diff
@@ -199,24 +189,18 @@ public void run() {
...
-            String output = ShellUtils.execCommand("bash", "-c", cmd);
-            if (output.isEmpty()) {
-              throw new IllegalStateException(
-                  String.format("Master process is not running on the host %s", host));
-            } else if (output.contains("Connection refused")) {
-              throw new IllegalStateException(
-                  String.format("Connection refused while connecting to the host %s on port %d",
-                      host, port));
+            LOG.debug("Master health check on node {}", address.getHostName());
+            try (Socket socket = new Socket()) {
```
maobaolong, line 193: Why we need to make sure all master of HA members are available ?
For now, the standby master don't start rpc service, and they are not necessary to start other masters while start a specific master

maobaolong (commented):
Thanks for this contribution. Left a comment inline.

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
ljl1988com: The standby masters don't start rpc service. But standby masters start up with the port. That indicates the service status of standby masters.

maobaolong: @ljl1988com AFAIK, only primary master start and listen rpc port, the standby masters didn't start rpc port.

even, if i start only two master of three master, the started two master will not pass the monitor for cannot connect to the third not started master?

ljl1988com: @maobaolong I make sure that the standby masters start rpc port in zk and embedded mode. But the standby masters don't service. If there is one master which is down, the check will not pass.

I will add a new checker which monitor port. Don't replace the monitor proc.

maobaolong: @ljl1988com 
> I make sure that the standby masters start rpc port in zk and embedded mode

you mean the rpc port which default 19998? Could you give some arguments or details to prove the standby masters start and listen the rpc(default 19998) port?

May be you can use netstat anp|grep 19998

Or just paste some related codes.

ljl1988com: Yes. I use netstat anp | grep 19998. I find the rpc port exists. So I make usre that the standby masters start rpc port.
 
https://github.com/Alluxio/alluxio/blob/4da0071c805923489d9da506b58cc829b82cfe71/core/server/master/src/main/java/alluxio/master/AlluxioMasterProcess.java#L188-L214

`startRejectingServers` will start the rpc port.

https://github.com/Alluxio/alluxio/blob/4da0071c805923489d9da506b58cc829b82cfe71/core/server/common/src/main/java/alluxio/master/MasterProcess.java#L181-L190

https://github.com/Alluxio/alluxio/blob/4da0071c805923489d9da506b58cc829b82cfe71/core/common/src/main/java/alluxio/network/RejectingServer.java#L44-L62

PS: I also thought standby servers didn't start rpc port before i started this issue.

LuQQiu:
@ggezer PTAL, thanks!

maobaolong:
@ljl1988com As we discussed this PR in the core community sync, we reach a consensus that we think each monitor of service should only check the healthy or port for itself, do not have to check the peers state. If you have any idea, we can have a discussion for this topic. @jiacheliu3 Any input?

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
jiacheliu3: The `RejectingServer` will close all connections on the socket, so I wonder if your `if(socket.isConnected())` check will return true?

jiacheliu3:
@ljl1988com thanks for the improvements!

There are two scenarios we are talking about, when we talk about monitoring.

```
S1. The alluxio-monitor.sh is typically used when the user manages the cluster with alluxio-start.sh. In this case, the user has password-less SSH to all nodes.
S2. In some other environments, the user does NOT have SSH to other nodes. For example, in K8s.
```

In S2, we are already using the port check to determine if a process (master/worker) is serving. We are checking the RPC port to see if the process accepts RPCs. For HA standby masters, we are checking the embedded journal port instead of the RPC port. No one uses ZK-based HA in K8s environment yet.

You can see an example in our K8s deployment YAML template:
https://github.com/Alluxio/alluxio/blob/da3697b7164ca2c948f48abc7bb14bbe305f701b/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml#L180

In S1, the alluxio-monitor.sh will check EITHER the process exists OR the port!

The process check is done by `MasterHealthCheckClient#ProcessCheckRunnable`, which is the part you updated. This SSH to the node and checks if the process exists.

The port check is actually done by `MasterHealthCheckClient#MasterServingHealthCheck`. Under the hood, it uses https://github.com/Alluxio/alluxio/blob/da3697b7164ca2c948f48abc7bb14bbe305f701b/core/common/src/main/java/alluxio/util/network/NetworkAddressUtils.java#L698 to connect to the gRPC service. So when the port is backed by `RejectingServer`, the gRPC connection will not succeed. In your code you are just connecting to the socket with `if (socket.isConnected())`, I'm not 100% sure if that will return true, but since the `RejectingServer` is closing all connections on the socket, I don't think your socket will be able to connect, although you are seeing that port as open with `netstat anp | grep 19998`. Could you double check this? 

To check, you can hack some code to make the Alluxio master start and make the 19998 backed by the `RejectingServer`, then use your monitor code to connect to that socket.

shell/src/main/java/alluxio/master/MasterHealthCheckClient.java
ljl1988com: I test the checker returns true.

ljl1988com:
@jiacheliu3 Just as you comment. I will want to refactor the ssh script and code. But I'm busy now.

PR #14021, opened 2021-09-04
Support logical name as HA address

qian0817:
### What changes are proposed in this pull request?

Support logical name as HA address.

### Why are the changes needed?

Currently, We have [two way to connect to HA cluster](https://docs.alluxio.io/os/user/stable/en/deploy/Running-Alluxio-On-a-HA-Cluster.html#ha-authority), but some frameworks may not accept either of these ways to connect to a Alluxio HA cluster, so we need to support connecting to Alluxio HA cluster via a logical name.

fix #13867 
fix #14019 

### Does this PR introduce any user facing changes?

if user set config `alluxio.master.rpc.addresses.[xxx]` , then can use `alluxio://[xxx]` to connect to alluxio embedded ha mode.
if user set config `alluxio.zookeeper.address.[xxx]` , then can use `alluxio://zk@xxx` to connect to alluxio zk ha mode.

PR #14056, opened 2021-09-11
Support multi-nameservices uri for creating HDFS fileSystem instance

hfutatzhanghb:
## Feature Introduction
Currently, when we use HDFS Command Line to access URL that is Alluxio style,  it can be like this:
`hdfs dfs -ls alluxio:///user/xxxx`. 

The command-line above does not support multiple alluxio cluster in the same time. So, the goal of this issue is to support such feature.  

After applying this patch , we can use `hdfs dfs -ls alluxio://ns1/user/xxx` , `hdfs dfs -ls alluxio://ns2/user/xxx` and so on in the same HDFS Client environment， where ns1 and ns2 are different alluxio cluster.

This patch also takes compatibility into account，if the uri does not have authority, it will fall back to original implementation.

## How to Use?

Suppose that we want to use two alluxio cluster which are named "ns1"  and "ns2" in HDFS command-line . We need to add **new** configuration entries below  to hdfs-site.xml:

```xml

<property>
    <name>alluxio.ns1.zookeeper.enabled</name>
    <value>true</value>
</property>

<property>
    <name>alluxio.ns1.zookeeper.address</name>
    <value>your ZK cluster address</value>
</property>

<property>
    <name>alluxio.ns1.zookeeper.election.path</name>
    <value>/alluxio/ns1/election</value>
</property>

<property>
    <name>alluxio.ns1.zookeeper.leader.path</name>
    <value>/alluxio/ns1/leader</value>
</property>

<property>
    <name>alluxio.ns2.zookeeper.enabled</name>
    <value>true</value>
</property>

<property>
    <name>alluxio.ns2.zookeeper.address</name>
    <value>your ZK cluster address</value>
</property>

<property>
    <name>alluxio.ns2.zookeeper.election.path</name>
    <value>/alluxio/ns2/election</value>
</property>

<property>
    <name>alluxio.ns2.zookeeper.leader.path</name>
    <value>/alluxio/ns2/leader</value>
</property>
```

The below picture displays results in product.

image


qian0817:
similar to #14021?

hfutatzhanghb:
> similar to #14021?

Hi, @qian0817 . I think it is not similar with #14021 , because this patch focuses on multiple alluxio clusters, not only one HA alluxio cluster.

qian0817:
You can specify alluxio service with URL authority like `hdfs dfs -ls alluxio://zk@host:2181/path` and `hdfs -ls alluxio://host:19998/path`. Releated doc is [here](https://docs.alluxio.io/os/user/stable/en/deploy/Running-Alluxio-On-a-HA-Cluster.html#ha-authority).

hfutatzhanghb:
> You can specify alluxio service with URL authority like `hdfs dfs -ls alluxio://zk@host:2181/path` and `hdfs -ls alluxio://host:19998/path`.
> In addition, you can use `hdfs --confg [config path] dfs -ls alluxio:///` or `hdfs dfs -D[conf key]=[conf value] -ls alluxio:/// ` to specify config.

Hi, @qian0817 . sry to reply too late.  I  have watched your code and doc,  it's very nice. In fact, our goals is similar. You have extended both master.rpc and zookeeper.enable, I think my solution is a specific supplement in zookeeper.enable sence？

ZhuTopher:
I'll come back to review if this should be added or not after #14021

hfutatzhanghb:
> I'll come back to review if this should be added or not after #14021

@ZhuTopher ,  thx a lot !

maobaolong:
@hfutatzhanghb Could you give an example config for raft based embedded journal ha?

Issue #14029, opened 2021-09-07
Elect return success while master report timeout exception


maobaolong:
**Alluxio Version:**
master branch from d2e9472438057146216374b46ece361c9acd4723

**Describe the bug**

Running elect return success while master report timeout exception

**To Reproduce**

- Start a ha mode 3 master cluster.
- Close one master.
- Elect the closed master to be leader

```
bin/alluxio fsadmin journal quorum elect -address localhost:19200
Transferred leadership to server: localhost:19200
Quorum priorities were reset to 1
```

But I found the error log in master.log, In fact the master encountered some unexpected error.

```
2021-09-07 11:46:51,426 ERROR RaftJournalSystem - caught an error: localhost_19201@group-ABB3109A44C1: Failed to transfer leadership to localhost_19200 (timed out 30000ms): current leader is localhost_19201
```

**Expected behavior**

Return exception to client, report fail to user.

**Urgency**
Normal

**Additional context**
No

maobaolong:
Another related exception is 

```
2021-09-07 14:57:18,645 ERROR RaftJournalSystem - caught an error: localhost_19201@group-ABB3109A44C1 refused to transfer leadership to peer localhost_19202 as it does not has highest priority 242: [localhost_19202|rpc:localhost:19202|admin:|client:|dataStream:|priority:1, localhost_19201|rpc:localhost:19201|admin:|client:|dataStream:|priority:1, localhost_19200|rpc:localhost:19200|admin:|client:|dataStream:|priority:1], old=null

```

maobaolong:
@codings-dan Would you like to take this ticket?

codings-dan:
> @codings-dan Would you like to take this ticket?

will do

apc999:
thanks @maobaolong  do you mean a new cli or a new option in existing cli?

maobaolong:
@apc999 No, I mean we can add a new client-master api to query the transferleadership schedule and result, if the transfer failed, the original master should reply the error reason to the client, so that the client will fail at once rather than waitfor forever like the current behavior.

apc999:
let's have a design discussion on this

ggezer:
Isn't this fixed with https://github.com/Alluxio/alluxio/pull/14027 ?

maobaolong:
@ggezer No, it's two different issue.

ggezer:
It was returning a success because `localhost:19200` and `localhost:19201` had the same hostname part. 
[...]

PR #14057, opened 2021-09-11
Return directly when an exception is thrown during transfer leader

codings-dan:

According to #14029 ,while master report timeout exception,return directly.

Now we can't perceive when throwing an exception, and there may be wrong results.

No

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
+++ b/core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
@@ -13,6 +13,7 @@ package alluxio.client.journal;
 
 import alluxio.exception.status.AlluxioStatusException;
 import alluxio.grpc.GetQuorumInfoPResponse;
+import alluxio.grpc.GetTransferLeaderMessagePResponse;
 import alluxio.grpc.NetAddress;
 
 import java.io.Closeable;
@@ -50,4 +51,11 @@ public interface JournalMasterClient extends Closeable {
    * @throws AlluxioStatusException
    */
   void resetPriorities() throws AlluxioStatusException;
+
+  /**
+   * Resets RaftPeer priorities.
+   *
+   * @return list of exception message throwing when transfer leader.
+   */
+  GetTransferLeaderMessagePResponse getTransferLeaderMessage() throws AlluxioStatusException;
 }
--- a/core/server/common/src/main/java/alluxio/master/journal/JournalMaster.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/JournalMaster.java
@@ -53,4 +53,9 @@ public interface JournalMaster {
    * @throws IOException if error occurs while performing the operation
    */
   void resetPriorities() throws IOException;
+
+  /**
+   * Gets list of exception message throwing when transfer leader.
+   */
+  void getTransferLeaderMessage();
 }
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
@@ -24,6 +24,7 @@ import alluxio.grpc.NetAddress;
 import alluxio.grpc.QuorumServerInfo;
 import alluxio.grpc.QuorumServerState;
 import alluxio.grpc.ServiceType;
+import alluxio.grpc.TransferLeaderMessage;
 import alluxio.master.Master;
 import alluxio.master.PrimarySelector;
 import alluxio.master.journal.AbstractJournalSystem;
@@ -224,6 +225,7 @@ public class RaftJournalSystem extends AbstractJournalSystem {
   private final ClientId mRawClientId = ClientId.randomId();
   private RaftGroup mRaftGroup;
   private RaftPeerId mPeerId;
+  private List<TransferLeaderMessage> mTransferMsgs;
 
   static long nextCallId() {
     return CALL_ID_COUNTER.getAndIncrement() & Long.MAX_VALUE;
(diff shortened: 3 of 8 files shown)


core/transport/src/main/proto/grpc/journal_master.proto
```diff
@@ -58,6 +58,17 @@ message ResetPrioritiesPRequest {
 }
 message ResetPrioritiesPResponse {}
 
+message TransferLeaderMessage {
+    optional string msg = 1;
```
maobaolong, line 62: We can use only one field `errorMessage`

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -69,6 +71,15 @@ public int run(CommandLine cl) throws IOException {
...
       CommonUtils.waitFor("Waiting for election to finalize", () -> {
         try {
+          GetTransferLeaderMessagePResponse transferMsg = jmClient.getTransferLeaderMessage();
+          Optional<TransferLeaderMessage>
+                  exceptionMsgOpt = transferMsg.getTransMsgList().stream()
+                  .filter(TransferLeaderMessage::getIsException).findAny();
+          String msg = exceptionMsgOpt.isPresent() ? exceptionMsgOpt.get().getMsg() : null;
+          if (msg != null) {
+            System.out.println(msg);
+            return false;
```
maobaolong, line 81: If you return false here, the `waitFor` loop will continue until 30 minutes past, so would you like to throw an exception to terminate the `waitFor` loop? So that the it can be found at once.

maobaolong (changes requested):
It look good overall, left some comment.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -1011,10 +1021,21 @@ private void processReply(RaftClientReply reply, String msgToUser) throws IOExce
               ? reply.getException()
               : new IOException(String.format("reply <%s> failed", reply));
       LOG.error("{}. Error: {}", msgToUser, ioe);
-      throw new IOException(msgToUser);
+      IOException exception = new IOException(msgToUser);
+      mTransferMsgs.add(TransferLeaderMessage.newBuilder()
```
maobaolong, line 1025: I didn't find where is responsible for remove or clear element of `mTransferMsgs `, would you like to have a double check?

maobaolong (changes requested):
Left comment, please take a look.

codings-dan:
@maobaolong I have cleared the element `mTransferMsgs`,PTAL

core/client/fs/src/main/java/alluxio/client/journal/JournalMasterClient.java
```diff
@@ -50,4 +51,11 @@
    * @throws AlluxioStatusException
    */
   void resetPriorities() throws AlluxioStatusException;
+
+  /**
+   * Resets RaftPeer priorities.
```
jenoudet, line 56:
```suggestion
   * Gets list of exception messages thrown when transferring the leader.
```

core/server/common/src/main/java/alluxio/master/journal/JournalMaster.java
```diff
@@ -53,4 +54,10 @@
    * @throws IOException if error occurs while performing the operation
    */
   void resetPriorities() throws IOException;
+
+  /**
+   * Gets list of exception message throwing when transfer leader.
```
jenoudet, line 59:
```suggestion
   * Gets list of exception messages thrown when transferring the  leader.
```

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -69,6 +71,16 @@ public int run(CommandLine cl) throws IOException {
       final int TIMEOUT_3MIN = 3 * 60 * 1000; // in milliseconds
       CommonUtils.waitFor("Waiting for election to finalize", () -> {
         try {
+          GetTransferLeaderMessagePResponse transferMsg = jmClient.getTransferLeaderMessage();
```
jenoudet, line 74: In `RaftJournalSystem` the exceptions are added to a list and then thrown. Therefore, if exceptions are encountered in `RaftJournalSystem` this call will never be executed.
Either we add them to the list without throwing exceptions in `RaftJournalSystem` and check them here, or we transform this `getTransferLeaderMessage` call into its own shell command.

core/transport/src/main/proto/grpc/journal_master.proto
```diff
@@ -82,4 +92,9 @@ service JournalMasterClientService {
      * Reset all the RaftPeer priorities.
      */
     rpc ResetPriorities(ResetPrioritiesPRequest) returns (ResetPrioritiesPResponse);
+
+    /**
+     * Gets list of exception message throwing when transfer leader.
```
jenoudet, line 97:
```suggestion
     * Gets list of exception messages thrown when transferring the leader.
```

jenoudet (changes requested):

ggezer:
@codings-dan Thanks for the improvement. 
I'd like to propose a different way for handling this case.
- JournalMaster.transferLeader() can return a randomly generated transfer-Id. 
- Add a new JournalMaster.getMessage(GUID transferId) method that returns the error message for that particular invocation. Internally we can keep map<GUID,String> of errors rather than a list.
- Shell can now get error only for its own invocation.

codings-dan:
@ggezer Ok, I will do it this way

codings-dan:
@ggezer I use a map to store exception information,PTAL.Do we need to consider deleting the data in the map at the right time？

jenoudet (commented):
Some minor comments.

jiacheliu3:
@ggezer could you take a look if the new commits align with your design in the high-level? Thanks!

codings-dan:
@jenoudet I fix it according to your comment,PTAL.

codings-dan:
@jenoudet The result of the ci test seems to have nothing to do with this code change, PTAL,thx!

Issue #10928, opened 2020-02-17
Unable to load native-hadoop library for your platform


loushang0826:
**Alluxio Version:**
2.1.1

**Describe the bug**
When alluxio master starts, there are some Warn messages as below:
`2020-01-17 16:57:37,306 WARN  NativeCodeLoader - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2020-01-17 16:57:37,312 WARN  DomainSocketFactory - The short-circuit local reads feature cannot be used because libhadoop cannot be loaded.`

as the Warn message implies, the short-circuit local reads feature cannot be used, will the impact the perfomance when reading data from HDFS

**To Reproduce**
1. configure according the doc: https://docs.alluxio.io/os/user/stable/en/ufs/HDFS.html#use-hadoop-native-library
2. start alluxio master
3. the Warn message arise

**Expected behavior**
1.  the libhadoop.so is successfully loaded into alluxio jvm

**Urgency**
 1. not urgent

**Additional context**
 1. i changed the log level to DEBUG for  `NativeCodeLoader` class and found the reason
`DEBUG NativeCodeLoader - Failed to load native-hadoop with error: java.lang.UnsatisfiedLinkError: Native Library /home/hadoop/hadoop/lib/native/libhadoop.so.1.0.0 already loaded in another classloader`
2. it seems that alluxio will have a separate classloader for a core jar file,such as one classloader for alluxio-underfs-hdfs-2.7.3-2.1.1.jar and another for alluxio-table-server-underdb-hive-2.1.1.jar, which will cause the repeated loading of /home/hadoop/hadoop/lib/native/libhadoop.so.1.0.0 file
3. i think the relevant code should be reviewed

bf8086:
@loushang0826 Thanks for reporting this issue. We will look into it and keep you updated.

jinhang:
I also care about this issue @loushang0826

jinhang:
I reproduce this issue on version 2.2.0 too.

bf8086:
@loushang0826 I have a couple questions:
- Are you using catalog service in Alluxio master? Does the error still show up if you disable catalog service?
- Does the error message show up on Alluxio workers?

loushang0826:
@bf8086 
as far as i know, i dont use catalog service in my alluxio cluster, and the alluxio workers have the same error message when started up

loushang0826:
@bf8086 any clues on the issue? 
i have some spare time to help study on it

[...]

PR #14058, opened 2021-09-13
Make sure ufs to only once load a native shared library

jhonxue:
Make sure ufs to only once load a native shared library(.so) at runtime, through making sure the same classloader for the same 
 ufs

Signed-off-by: Xue Yantao xueyantao2114@163.com

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
+++ b/core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
@@ -200,6 +200,7 @@ public class ExtensionFactoryRegistry<T extends ExtensionFactory<?, S>,
           LOG.debug("Discovered a factory implementation {} - {} in jar {}", factory.getClass(),
               factory, jarPath);
           register(factory, factories);
+          register(factory);
         }
       } catch (Throwable t) {
         LOG.warn("Failed to load jar {}: {}", jar, t.toString());
--- a/core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
+++ b/core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
@@ -278,8 +278,8 @@ public class ExtensionFactoryRegistry<T extends ExtensionFactory<?, S>,
    * @param factories list of factories
    * @return list of factories that support the given path which may be an empty list
    */
-  public List<T> select(String path, S conf, List<T> factories) {
-    Preconditions.checkArgument(path != null, "path may not be null");
+  private List<T> select(String path, S conf, List<T> factories) {
+    Preconditions.checkNotNull(path, "path may not be null");
 
     List<T> eligibleFactories = new ArrayList<>();
     for (T factory : factories) {
--- a/core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
+++ b/core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
@@ -131,19 +131,18 @@ public class ExtensionFactoryRegistry<T extends ExtensionFactory<?, S>,
     Preconditions.checkArgument(path != null, "path may not be null");
 
     List<T> factories = new ArrayList<>(mFactories);
+    List<T> eligibleFactories = select(path, conf, factories);
+    if (!eligibleFactories.isEmpty()) {
+      return eligibleFactories;
+    }
+
+    factories.clear();
     String libDir = PathUtils.concatPath(conf.get(PropertyKey.HOME), "lib");
     String extensionDir = conf.get(PropertyKey.EXTENSIONS_DIR);
     scanLibs(factories, libDir);
     scanExtensions(factories, extensionDir);
 
-    List<T> eligibleFactories = new ArrayList<>();
-    for (T factory : factories) {
-      if (factory.supportsPath(path, conf)) {
-        LOG.debug("Factory implementation {} is eligible for path {}", factory, path);
-        eligibleFactories.add(factory);
-      }
-    }
-
+    eligibleFactories = select(path, conf, factories);
     if (eligibleFactories.isEmpty()) {
       LOG.warn("No factory implementation supports the path {}", path);
     }
@@ -269,5 +268,26 @@ public class ExtensionFactoryRegistry<T extends ExtensionFactory<?, S>,
 
     LOG.debug("Unregistered factory implementation {} - {}", factory.getClass(), factory);
     factories.remove(factory);
+  }
+
+  /**
+   * Selects all the factories that support the given path from the given list of factories.
+   *
+   * @param path path
+   * @param conf configuration of the extension
+   * @param factories list of factories
+   * @return list of factories that support the given path which may be an empty list
+   */
+  public List<T> select(String path, S conf, List<T> factories) {
+    Preconditions.checkArgument(path != null, "path may not be null");
+
+    List<T> eligibleFactories = new ArrayList<>();
+    for (T factory : factories) {
+      if (factory.supportsPath(path, conf)) {
+        LOG.debug("Factory implementation {} is eligible for path {}", factory, path);
+        eligibleFactories.add(factory);
+      }
+    }
+    return eligibleFactories;
   }
 }
--- a/underfs/cephfs/pom.xml
+++ b/underfs/cephfs/pom.xml
@@ -31,7 +31,6 @@
       <groupId>io.github.opendataio</groupId>
       <artifactId>libcephfs</artifactId>
       <version>0.0.1</version>
-      <scope>provided</scope>
     </dependency>
 
     <!-- Internal dependencies -->


core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
```diff
@@ -270,4 +269,25 @@ private void unregister(T factory, List<T> factories) {
...
+
+  /**
+   * Selects all the factories that support the given path from the given list of factories.
+   *
+   * @param path path
+   * @param conf configuration of the extension
+   * @param factories list of factories
+   * @return list of factories that support the given path which may be an empty list
+   */
+  public List<T> select(String path, S conf, List<T> factories) {
```
apc999, line 281: do we need this method to be public?

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
```diff
@@ -270,4 +269,25 @@ private void unregister(T factory, List<T> factories) {
...
+    Preconditions.checkArgument(path != null, "path may not be null");
```
apc999, line 282: alluxio code convention:
```suggestion
    Preconditions.checkNotNull(path, "path");
```

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
```diff
@@ -131,19 +131,18 @@ private synchronized void init() {
     Preconditions.checkArgument(path != null, "path may not be null");
 
     List<T> factories = new ArrayList<>(mFactories);
+    List<T> eligibleFactories = select(path, conf, factories);
+    if (!eligibleFactories.isEmpty()) {
+      return eligibleFactories;
+    }
+
+    factories.clear();
```
apc999, line 139: i don't understand why we can prevent loading native shared lib multiple times by calling clear here.
can you elaborate?

apc999 (changes requested):
comments left.

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
jhonxue: > do we need this method to be public?

may be not. private is better

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
jhonxue: > i don't understand why we can prevent loading native shared lib multiple times by calling clear here.
> can you elaborate?

At this line, we did not find any supported factories from the list of factories. So we clear it to avoid to twicely check them wether support or not.

maobaolong:
https://github.com/Alluxio/alluxio/issues/10928 Not sure it is same root cause with this issue.

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
```diff
@@ -131,19 +131,18 @@ private synchronized void init() {
     Preconditions.checkArgument(path != null, "path may not be null");
 
     List<T> factories = new ArrayList<>(mFactories);
+    List<T> eligibleFactories = select(path, conf, factories);
```
maobaolong, line 134: You mean the first time `findAll ` invoked, the `mFactories ` should be empty, and this method will run like original logic, later new `findAll ` will not load the extension or lib jar at all? So please show the effect.

underfs/cephfs/pom.xml
```diff
@@ -31,7 +31,6 @@
       <groupId>io.github.opendataio</groupId>
       <artifactId>libcephfs</artifactId>
       <version>0.0.1</version>
-      <scope>provided</scope>
```
maobaolong, line 34: What's the purpose you remove the `provided ` scope, you want to include the `libcephfs ` and related share lib into the ufs shaded with dependencies jar?

maobaolong (commented):
Thanks for this improvement, lt looks like solve our problem, left some question inline, just make sure there are no downside for this PR

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
jhonxue: > You mean the first time `findAll ` invoked, the `mFactories ` should be empty, and this method will run like original logic, later new `findAll ` will not load the extension or lib jar at all? So please show the effect.

Yes, the new 'findAll' will not dynamicly load the ext or lib jar after the first load done. However, we can restart the alluxio master or worker to load the new  updated jar.

underfs/cephfs/pom.xml
jhonxue: > What's the purpose you remove the `provided ` scope, you want to include the `libcephfs ` and related share lib into the ufs shaded with dependencies jar?

Yes, this pr will solve the problem about the libcephfs-jni.so is already loaded in another classloader, so libcephfs can be included.

jhonxue:
The native shared library was loaded in another classloader:

Below is the master log in my test:

2021-09-18 15:49:21,444 WARN  UnderFileSystem$Factory - Failed to create UnderFileSystem by factory alluxio.underfs.cephfs.CephFSUnderFileSystemFactory@5e2694c1: java.lang.UnsatisfiedLinkError: Native Library /usr/lib64/libcephfs_jni.so.1.0.0 already loaded in another classloader
2021-09-18 15:49:21,444 ERROR FileSystemMasterClientServiceHandler - Exit (Error): Mount: request=alluxioPath: "/fs2"

jhonxue:
After used this pr:  the native shared lib problem is solved

We test underfs-cephfs and underfs-cephfs-hadoop
[root@s131 /data/alluxio-2.7.0-SNAPSHOT]# bin/alluxio fs  mount
cephfs://xx.xx.xx.A:6789/                         on  /fs2  (cephfs, capacity=711.02TB, used=86.41GB(0%), not read-only, not shared})
cephfs://xx.xx.xx.B:6789/                          on  /fs3  (cephfs, capacity=24.59TB, used=58.50GB(0%), not read-only, not shared})
ceph://xx.xx.xx.A:6789/                                on  /fs1  (hdfs, capacity=-1B, used=-1B, read-only, not shared)
ceph://xx.xx.xx.B:6789/                            on  /fs4  (hdfs, capacity=-1B, used=-1B, read-only, not shared})
/data/alluxio-2.7.0-SNAPSHOT/underFSStorage  on  /     (local, capacity=31.90GB, used=29.72GB(93%), not read-only, not shared, properties={})

jhonxue:
@apc999 PTAL

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
maobaolong: Please highlight add this important limitation to the related issue or the description of this PR. It would be better if you write an approach description shortly to let reviewer know your solution and make it easy to review this PR

LuQQiu:
Mainly saw in Ceph, C++ library

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
```diff
@@ -201,6 +200,7 @@ private void scan(List<File> files, List<T> factories) {
           LOG.debug("Discovered a factory implementation {} - {} in jar {}", factory.getClass(),
               factory, jarPath);
           register(factory, factories);
+          register(factory);
```
LuQQiu, line 203:
```
  /**
   * The base list of factories, which does not include any lib or extension factories. The only
   * factories in the base list will be built-in factories, and any additional factories
   * registered by tests. All other factories will be discovered and service loaded when extension
   * creation occurs.
   */
  private final List<T> mFactories = new CopyOnWriteArrayList<>();
```
we modify the basic assumption of what `mFactories` represents

will it be better to include a new list call `mDynamicLoadedFactories`?

LuQQiu: 1. If mFactories contains -> return
2. If mDynamicLoadedFactories contains -> return
3. If not, renew the mDynamicLoadedFactories?

LuQQiu: @apc999 What's your thoughts

LuQQiu (commented):
Thanks for making class loader changes, left some comments

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
jhonxue: > 1. If mFactories contains -> return
> 2. If mDynamicLoadedFactories contains -> return
> 3. If not, renew the mDynamicLoadedFactories?

Yea, i think it sounds better! By the way, may be that we can add a method "boolean supportsDynamicLoad()" in the public interface ExtensionFactory to distinguish the underFileSystemFactories, likes  the existed method "boolean supportsPath(String path, S conf)"，so we can still only use mFactories. 

What's your thoughts?   @apc999 @LuQQiu @maobaolong

jhonxue: > I prefer this way, but is it complicated to add the logic?
Based on your idea, the steps are :  
          1. If mFactories contains &&  !supportsDynamicLoad()     -> return
          2. If mDynamicLoadedFactories contains && !supportsDynamicLoad()    -> return
          3. If not, renew the mDynamicLoadedFactories?

LuQQiu: The workload SGTM, @apc999 @ggezer @yuzhu PTAL

maobaolong (approved):
As we patching this PR and it can works in our Env, It LGTM.

LuQQiu:
Sorry for the late review. @jhonxue can you help merge the latest master into the branch to solve the conflicts?

jhonxue:
> Sorry for the late review. @jhonxue can you help merge the latest master into the branch to solve the conflicts?

ok, my pleasure. Also done the conflicts

core/common/src/main/java/alluxio/extensions/ExtensionFactoryRegistry.java
```diff
@@ -159,6 +165,7 @@ public List<T> findAllWithRecorder(String path, S conf, Recorder recorder) {
       recorder.record("alluxio.underfs.version is not set by user");
     }
 
+    eligibleFactories = select(path, conf, factories);
     for (T factory : factories) {
```
LuQQiu, line 169: is select duplicate with the following for loop

jiacheliu3:
@LuQQiu @jhonxue Is this PR still valid? thanks!

PR #14063, opened 2021-09-13
Add and reorganize embedded journal tests

jenoudet:


minicluster/src/main/java/alluxio/multi/process/PortCoordination.java
```diff
@@ -33,14 +33,11 @@
       // add ports here to avoid conflicting with other processes on those ports.
   ));
 
-  public static final List<ReservedPort> EMBEDDED_JOURNAL_FAILOVER = allocate(3, 0);
-  public static final List<ReservedPort> EMBEDDED_JOURNAL_RESTART = allocate(3, 0);
-  public static final List<ReservedPort> EMBEDDED_JOURNAL_RESTART_STRESS = allocate(3, 0);
```
ggezer, line 38: what do these tests use now?

tests/src/test/java/alluxio/server/ft/journal/raft/BaseEmbeddedJournalTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.After;
+import org.junit.Rule;
+
+import java.util.concurrent.TimeoutException;
+import java.util.function.Predicate;
+
+public class BaseEmbeddedJournalTest extends BaseIntegrationTest {
+
+  @Rule
+  public ConfigurationRule mConf =
```
ggezer, line 36: fix indent style.

tests/src/test/java/alluxio/server/ft/journal/raft/FaultToleranceEBJTest.java
```diff
@@ -0,0 +1,278 @@
...
+import org.junit.Test;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class FaultToleranceEBJTest extends BaseEmbeddedJournalTest {
```
ggezer, line 45: Let's use a convention where all these tests start with the same header. For example:
`EmbeddedJournalIntegrationTest_FaultTolerance`
`EmbeddedJournalIntegrationTest_Resizing`
`EmbeddedJournalIntegrationTest_TransferLeaderShip`

tests/src/test/java/alluxio/server/ft/journal/raft/FaultToleranceEBJTest.java
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  private static class OperationThread extends Thread {
+    private final FileSystem mFs;
+    private final int mThreadNum;
+    private final AtomicReference<Throwable> mFailure;
+    private final AtomicInteger mSuccessCounter;
+
+    public OperationThread(FileSystem fs, int threadNum, AtomicReference<Throwable> failure,
+                           AtomicInteger successCounter) {
```
ggezer, line 223: fix indent style.

tests/src/test/java/alluxio/server/ft/journal/raft/FaultToleranceEBJTest.java
```diff
@@ -0,0 +1,278 @@
...
+        runInternal();
+      } catch (Exception e) {
+        e.printStackTrace();
+        mFailure.set(e);
+      }
+    }
+
+    public void runInternal() throws Exception {
+      while (!Thread.interrupted()) {
+        for (int i = 0; i < 100; i++) {
```
ggezer, line 242: let's make this a constant and bump it to 1000 now that it runs in parallel with the rest.

ggezer (changes requested):
Looks good. Left few comments.

minicluster/src/main/java/alluxio/multi/process/PortCoordination.java
jenoudet: @ZacBlanco Is one list necessary per each test, or can tests grouped under the same class reuse the same ports? Also, could the `allocate` method be made public so that tests can dynamically requests new ports at runtime, or would that cause collisions? Thanks in advance.

tests/src/test/java/alluxio/server/ft/journal/raft/FaultToleranceEBJTest.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class FaultToleranceEBJTest extends BaseEmbeddedJournalTest {
+
+  private static final int RESTART_TIMEOUT_MS = 2 * Constants.MINUTE_MS;
+  private static final int NUM_MASTERS = 3;
+  private static final int NUM_WORKERS = 0;
+
+  private void standardBefore() throws Exception {
```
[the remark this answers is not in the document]
jenoudet: see convo in `PortCoordination`

minicluster/src/main/java/alluxio/multi/process/PortCoordination.java
ZacBlanco: Multiple lists is better because we've seen cases where tests that run sequentially and attempt to re-use the same port fail to start because the OS in the background hasn't actually fully closed the socket (usually due to some TCP timeout or high system load). It just prevents some flakiness in the test if we can allocate ports for each test individually, even if their execution doesn't overlap.

Dynamically allocating at runtime could lead to more potential problems if there are conflicts. AFAIK the port coordination uses filesystem locks (`man 2 flock`) to try to grab exclusive access to the ports, but the locks are simply advisory and aren't a strong guarantee.

ggezer:
`EmbeddedJournalIntegrationTestResizing.replaceAll` seems to be failing

jenoudet:
> `EmbeddedJournalIntegrationTestResizing.replaceAll` seems to be failing

And only on Java 8, not Java 11. I'm investigating.

ggezer (approved):
Thanks!

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -522,6 +531,42 @@ public synchronized void stopMaster(int i) throws IOException {
...
+   */
+  public synchronized void removeMaster(int i) throws IOException {
+    stopMaster(i);
+    mMasterAddresses.remove(i);
+    mMasterIds.remove(i);
+    mMasters.remove(i);
+    mNumMasters--;
+    if (mMasterProperties.containsKey(i)) {
+      mMasterProperties.remove(i);
+    }
```
ZacBlanco, line 547: technically an unecessary check because `Map#remove` will just return null if the key doesn't exist.

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
```diff
@@ -0,0 +1,305 @@
...
+    extends EmbeddedJournalIntegrationTestBase {
+
+  private static final int RESTART_TIMEOUT_MS = 6 * Constants.MINUTE_MS;
+  private static final int NUM_MASTERS = 3;
+  private static final int NUM_WORKERS = 0;
+
+  @Test
+  public void failover() throws Exception {
+    mCluster = MultiProcessCluster.newBuilder(PortCoordination.EMBEDDED_JOURNAL_FAILOVER)
+        .setClusterName("EmbeddedJournalFaultTolerance_failover")
```
ZacBlanco, line 55: potential to automatically generate the name using the class name and then getting the calling method name: https://www.geeksforgeeks.org/get-name-of-current-method-being-executed-in-java/

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
```diff
@@ -0,0 +1,305 @@
...
+      for (int i = 0; i < 10; i++) {
+        EmbeddedJournalIntegrationTestFaultTolerance.OperationThread t =
+            new EmbeddedJournalIntegrationTestFaultTolerance
+                .OperationThread(fs, i, failure, successes);
+        t.start();
+        threads.add(t);
+      }
+      for (int i = 0; i < 2; i++) {
+        restartMasters();
+        System.out.printf("---------- Iteration %s ----------\n", i);
```
ZacBlanco, line 215: remove debugging printf?

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestResizing.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.stream.Collectors;
+
+public class EmbeddedJournalIntegrationTestResizing extends EmbeddedJournalIntegrationTestBase {
+
+  @Test
+  public void resizeCluster() throws Exception {
+    final int NUM_MASTERS = 5;
+    final int NUM_WORKERS = 0;
+    mCluster = MultiProcessCluster.newBuilder(PortCoordination.EMBEDDED_JOURNAL_RESIZE)
+        .setClusterName("EmbeddedJournalResizing")
```
ZacBlanco, line 45: this name isn't the same format at the other class. Intentional?

ZacBlanco (changes requested):
These are just minor comments overall. I don't think they should block merging if it is critical, but can improve maintainability in the future.

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
jenoudet: fixed

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
jenoudet: I wanted to refactor these cluster setup lines into one `@Before` method, but that would cause problems with `PortCoordination` as we've discussed before, so I made all the name explicit.

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
jenoudet: fixed, this was a direct port from the previous `EmbeddedJournalIntegrationTest` file

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestResizing.java
jenoudet: fixed

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
ZacBlanco: I understand that. I just meant you could do something like:

```java
...
      .setClusterName(generateClusterName())
...

public String generateClusterName() {
  return this.class.getName() + "_" + new Throwable().getStackTrace()[1].getMethodName();
}
```

ZacBlanco (approved):

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
```diff
@@ -522,6 +531,40 @@ public synchronized void stopMaster(int i) throws IOException {
     mMasters.get(i).close();
   }
 
+  /**
+   * Removes master i from the cluster.
+   * @param i index of the master to be removed
+   * @throws IOException
+   */
+  public synchronized void removeMaster(int i) throws IOException {
```
ggezer, line 539: nit: stopAndRemoveMaster()

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
jenoudet: There's actually a simpler way of doing this in JUnit:
```
@Rule 
public TestName testName = new TestName();
...
 testName.getMethodName();
```
This gets you the name of the test that is currently running, even in `@Before` methods.

minicluster/src/main/java/alluxio/multi/process/MultiProcessCluster.java
jenoudet: fixed

tests/src/test/java/alluxio/server/ft/journal/raft/EmbeddedJournalIntegrationTestFaultTolerance.java
ZacBlanco: :O never knew this. Nice!

ggezer:

PR #14069, opened 2021-09-14
Move transferleader logic from master to client-side

### LINKED ISSUE #14068: Move transferleader logic from master to client-side

Now, alluxio master like a remote proxy for alluxio elect shell from client side, which send transferleader rpc to ratis server asynchronously, so the elect shell will not receive any message although the transferleader rpc call timeout or encountered an exception

If we move the tranfer leadership code logic from master to the client-side, we can maintain the progress, status, result easily.

image

let admin shell trigger transfer leader.

**Urgency**

Normal

**Additional context**

No

maobaolong:

Fix #14068

Make it easy to handle the whole transferleader transaction.

```
bin/alluxio fsadmin journal quorum elect -address localhost:19200
Applying new peer state before transferring leadership: [localhost_19202|rpc:localhost:19202|priority:1, localhost_19201|rpc:localhost:19201|priority:1, localhost_19200|rpc:localhost:19200|priority:2]
Transferring leadership to master with address <localhost:19200> and with RaftPeerId <localhost_19200>
Transferring leadership initiated
Successfully elected localhost:19200 as the new leader
```

### DIFF
--- a/shell/pom.xml
+++ b/shell/pom.xml
@@ -67,6 +67,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-compress</artifactId>
     </dependency>
+    <dependency>
+      <artifactId>ratis-client</artifactId>
+      <groupId>org.apache.ratis</groupId>
+    </dependency>
 
     <!-- Internal dependencies -->
     <dependency>
@@ -97,6 +101,12 @@
       <version>${project.version}</version>
       <type>test-jar</type>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-core-server-common</artifactId>
+      <version>2.7.0-SNAPSHOT</version>
+      <scope>compile</scope>
     </dependency>
   </dependencies>
 </project>
--- a/shell/pom.xml
+++ b/shell/pom.xml
@@ -93,6 +93,11 @@
       <artifactId>alluxio-job-client</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.alluxio</groupId>
+      <artifactId>alluxio-core-server-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
 
       <!-- Test dependencies -->
     <dependency>
@@ -101,12 +106,6 @@
       <version>${project.version}</version>
       <type>test-jar</type>
       <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.alluxio</groupId>
-      <artifactId>alluxio-core-server-common</artifactId>
-      <version>2.7.0-SNAPSHOT</version>
-      <scope>compile</scope>
     </dependency>
   </dependencies>
 </project>
--- a/shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
+++ b/shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
@@ -137,7 +137,7 @@ public class QuorumElectCommand extends AbstractFsAdminCommand {
 
     RaftPeerId newLeaderPeerId =
         RaftJournalUtils.getPeerId(address.getHost(), address.getRpcPort());
-    /* update priorities to enable transfer */
+    // update priorities to enable transfer
     List<RaftPeer> peersWithNewPriorities = new ArrayList<>();
     for (RaftPeer peer : oldPeers) {
       peersWithNewPriorities.add(
@@ -151,33 +151,30 @@ public class QuorumElectCommand extends AbstractFsAdminCommand {
           .collect(Collectors.joining(", ")) + "]";
       mPrintStream.printf(
           "Applying new peer state before transferring leadership: %s%n", stringPeers);
-      RaftClientReply reply = client.admin().setConfiguration(peersWithNewPriorities);
-      processReply(reply, "failed to set master priorities before initiating election");
-      /* transfer leadership */
+      RaftClientReply setConfigurationReply =
+          client.admin().setConfiguration(peersWithNewPriorities);
+      processReply(setConfigurationReply,
+          "failed to set master priorities before initiating election");
+      // transfer leadership
       mPrintStream.printf(
           "Transferring leadership to master with address <%s> and with RaftPeerId <%s>%n",
           serverAddress, newLeaderPeerId);
       // fire and forget: need to immediately return as the master will shut down its RPC servers
       // once the TransferLeadershipRequest is initiated.
-      final int SLEEP_TIME_MS = 3_000;
-      final int TRANSFER_LEADER_WAIT_MS = 60_000;
       try {
-        Thread.sleep(SLEEP_TIME_MS);
-        RaftClientReply reply1 = client.admin().transferLeadership(newLeaderPeerId,
-            TRANSFER_LEADER_WAIT_MS);
-        processReply(reply1, "election failed");
+        Thread.sleep(3_000);
+        RaftClientReply transferLeadershipReply =
+            client.admin().transferLeadership(newLeaderPeerId, 60_000);
+        processReply(transferLeadershipReply, "election failed");
       } catch (Throwable t) {
         mPrintStream.printf("caught an error when executing transfer: %s%n", t.getMessage());
         return -1;
       }
       mPrintStream.println("Transferring leadership initiated");
-    } catch (Throwable t) {
-      throw new IOException(t);
     }
 
     try {
       // wait for confirmation of leadership transfer
-      final int TIMEOUT_3MIN = 3 * 60 * 1000; // in milliseconds
       CommonUtils.waitFor("Waiting for election to finalize", () -> {
         try {
           GetQuorumInfoPResponse quorumInfo = jmClient.getQuorumInfo();
@@ -191,7 +188,7 @@ public class QuorumElectCommand extends AbstractFsAdminCommand {
         } catch (AlluxioStatusException e) {
           return false;
         }
-      }, WaitForOptions.defaults().setTimeoutMs(TIMEOUT_3MIN));
+      }, WaitForOptions.defaults().setTimeoutMs(3 * 60 * 1000));
 
       mPrintStream.println(String.format(TRANSFER_SUCCESS, serverAddress));
       return 0;
@@ -242,7 +239,7 @@ public class QuorumElectCommand extends AbstractFsAdminCommand {
         .setBaseSleepTime(TimeDuration.valueOf(100, TimeUnit.MILLISECONDS))
         .setMaxAttempts(10)
         .setMaxSleepTime(
-            TimeDuration.valueOf(100000, TimeUnit.MILLISECONDS))
+            TimeDuration.valueOf(100_000, TimeUnit.MILLISECONDS))
         .build();
     return RaftClient.newBuilder()
         .setRaftGroup(mRaftGroup)
--- a/shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
+++ b/shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
@@ -15,7 +15,6 @@ import static alluxio.master.journal.raft.RaftJournalSystem.RAFT_GROUP_ID;
 
 import alluxio.cli.fsadmin.command.AbstractFsAdminCommand;
 import alluxio.cli.fsadmin.command.Context;
-import alluxio.client.journal.JournalMasterClient;
 import alluxio.conf.AlluxioConfiguration;
 import alluxio.exception.ExceptionMessage;
 import alluxio.exception.status.AlluxioStatusException;
@@ -99,7 +98,6 @@ public class QuorumElectCommand extends AbstractFsAdminCommand {
 
   @Override
   public int run(CommandLine cl) throws IOException {
-    JournalMasterClient jmClient = mMasterJournalMasterClient;
     String strAddr = cl.getOptionValue(ADDRESS_OPTION_NAME);
     NetAddress address = QuorumCommand.stringToAddress(strAddr);
     InetSocketAddress serverAddress = InetSocketAddress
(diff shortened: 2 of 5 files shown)


LuQQiu:
FYI @jiacheliu3

shell/pom.xml
```diff
@@ -67,6 +67,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-compress</artifactId>
     </dependency>
+    <dependency>
```
jiacheliu3, line 70: @LuQQiu as mentioned today by Baolong, this only adds dependency to shell module so no dependency added to the client

shell/pom.xml
```diff
@@ -98,5 +102,11 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
jiacheliu3, line 105: However I'm not sure if  this is desirable

jiacheliu3: I believe generally we want to separate the client and server so one does not depend on the other. If some logic is shared, I think what we do is to extract that into somewhere like `alluxio-core-common` so both can use it.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+      RaftClientReply reply = client.admin().setConfiguration(peersWithNewPriorities);
+      processReply(reply, "failed to set master priorities before initiating election");
+      /* transfer leadership */
+      mPrintStream.printf(
+          "Transferring leadership to master with address <%s> and with RaftPeerId <%s>%n",
+          serverAddress, newLeaderPeerId);
+      // fire and forget: need to immediately return as the master will shut down its RPC servers
+      // once the TransferLeadershipRequest is initiated.
+      final int SLEEP_TIME_MS = 3_000;
+      final int TRANSFER_LEADER_WAIT_MS = 60_000;
```
jiacheliu3, line 163: static variables?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+      }
+      mPrintStream.println("Transferring leadership initiated");
+    } catch (Throwable t) {
+      throw new IOException(t);
+    }
+
     try {
-      jmClient.transferLeadership(address);
       // wait for confirmation of leadership transfer
       final int TIMEOUT_3MIN = 3 * 60 * 1000; // in milliseconds
```
jiacheliu3, line 180: static variable?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+    final JournalMasterClient jmClient;
+    if (cl.hasOption(DOMAIN_OPTION_NAME)) {
+      String domainVal = cl.getOptionValue(DOMAIN_OPTION_NAME);
+      try {
+        JournalDomain domain = JournalDomain.valueOf(domainVal);
+        if (domain == JournalDomain.JOB_MASTER) {
+          jmClient = mJobMasterJournalMasterClient;
+          serviceType = NetworkAddressUtils.ServiceType.JOB_MASTER_RAFT;
+        } else {
+          jmClient = mMasterJournalMasterClient;
```
jiacheliu3, line 104: initialize `jmClient` above to avoid `else`?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+      /* transfer leadership */
+      mPrintStream.printf(
+          "Transferring leadership to master with address <%s> and with RaftPeerId <%s>%n",
+          serverAddress, newLeaderPeerId);
+      // fire and forget: need to immediately return as the master will shut down its RPC servers
+      // once the TransferLeadershipRequest is initiated.
+      final int SLEEP_TIME_MS = 3_000;
+      final int TRANSFER_LEADER_WAIT_MS = 60_000;
+      try {
+        Thread.sleep(SLEEP_TIME_MS);
```
jiacheliu3, line 165: `CommonUtils.sleep()`

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -118,7 +227,44 @@ public String getDescription() {
...
+  private RaftClient createClient() {
+    RaftProperties properties = new RaftProperties();
+    Parameters parameters = new Parameters();
+    RaftClientConfigKeys.Rpc.setRequestTimeout(properties,
+        TimeDuration.valueOf(15, TimeUnit.SECONDS));
+    ExponentialBackoffRetry retryPolicy = ExponentialBackoffRetry.newBuilder()
+        .setBaseSleepTime(TimeDuration.valueOf(100, TimeUnit.MILLISECONDS))
+        .setMaxAttempts(10)
+        .setMaxSleepTime(
+            TimeDuration.valueOf(100000, TimeUnit.MILLISECONDS))
```
jiacheliu3, line 245: nit: `100_000`

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+      mPrintStream.printf(
+          "Transferring leadership to master with address <%s> and with RaftPeerId <%s>%n",
+          serverAddress, newLeaderPeerId);
+      // fire and forget: need to immediately return as the master will shut down its RPC servers
+      // once the TransferLeadershipRequest is initiated.
+      final int SLEEP_TIME_MS = 3_000;
+      final int TRANSFER_LEADER_WAIT_MS = 60_000;
+      try {
+        Thread.sleep(SLEEP_TIME_MS);
+        RaftClientReply reply1 = client.admin().transferLeadership(newLeaderPeerId,
```
jiacheliu3, line 166: nit: better name?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
jiacheliu3: and why sleep here?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+      );
+    }
+    try (RaftClient client = createClient()) {
+      String stringPeers = "[" + peersWithNewPriorities.stream().map(RaftPeer::toString)
+          .collect(Collectors.joining(", ")) + "]";
+      mPrintStream.printf(
+          "Applying new peer state before transferring leadership: %s%n", stringPeers);
+      RaftClientReply reply = client.admin().setConfiguration(peersWithNewPriorities);
+      processReply(reply, "failed to set master priorities before initiating election");
+      /* transfer leadership */
```
jiacheliu3, line 156: nit: `//`

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+    List<RaftPeer> oldPeers = new ArrayList<>(mRaftGroup.getPeers());
+    // if you cannot find the address in the quorum, throw exception.
+    if (oldPeers.stream().map(RaftPeer::getAddress).noneMatch(addr -> addr.equals(strAddr))) {
+      throw new IOException(String.format("<%s> is not part of the quorum <%s>.",
+          strAddr, oldPeers.stream().map(RaftPeer::getAddress).collect(Collectors.toList())));
+    }
+
+    RaftPeerId newLeaderPeerId =
+        RaftJournalUtils.getPeerId(address.getHost(), address.getRpcPort());
+    /* update priorities to enable transfer */
```
jiacheliu3, line 140: nit: `//`

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+        .collect(Collectors.toSet());
+    mRaftGroup = RaftGroup.valueOf(RAFT_GROUP_ID, peers);
+
+    String strAddr = cl.getOptionValue(ADDRESS_OPTION_NAME);
+    NetAddress address = QuorumCommand.stringToAddress(strAddr);
+    InetSocketAddress serverAddress = InetSocketAddress
+        .createUnresolved(address.getHost(), address.getRpcPort());
+    List<RaftPeer> oldPeers = new ArrayList<>(mRaftGroup.getPeers());
+    // if you cannot find the address in the quorum, throw exception.
+    if (oldPeers.stream().map(RaftPeer::getAddress).noneMatch(addr -> addr.equals(strAddr))) {
```
jiacheliu3, line 133: trim?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
```diff
@@ -60,11 +90,92 @@ public static String description() {
...
+        RaftClientReply reply1 = client.admin().transferLeadership(newLeaderPeerId,
+            TRANSFER_LEADER_WAIT_MS);
+        processReply(reply1, "election failed");
+      } catch (Throwable t) {
+        mPrintStream.printf("caught an error when executing transfer: %s%n", t.getMessage());
+        return -1;
+      }
+      mPrintStream.println("Transferring leadership initiated");
+    } catch (Throwable t) {
+      throw new IOException(t);
```
jiacheliu3, line 175: Why catch the `Throwable` in the above block but throw here?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
jiacheliu3, line 245: Are these default values good enough or there's a chance a user may tune them?

jiacheliu3 (changes requested):
Mostly look good to me. Most of my comments are related to format. Thx for the improvement!

shell/pom.xml
maobaolong: If we extract the raft related logic to `alluxio-core-common` which means ratis dependencies will involved into alluxio client.

shell/pom.xml
maobaolong: Yeah.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: No, just use hardcode here as no one need to know this detailed concept now.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: No, remove it.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: I have to use final to limit the `jmClient` since it is used in the following lambda expression L180

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: @jenoudet Could you please give some explaination?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: Yeah

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: Sure

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: Got it

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: Done

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: Would you like to keep hardcode here until someone want to change it?

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: You mean the `strAddr`? It is from option, so it cannot be surrounded by whitespace I think.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
maobaolong: No need for this place, thanks for point it out, have remove this useless catch block

maobaolong:
@jiacheliu3 Thanks for your review and suggestion, the comments has been addressed and replied, PTAL.

shell/src/main/java/alluxio/cli/fsadmin/journal/QuorumElectCommand.java
jenoudet: The leading master shuts down its RPC server when it loses its leadership. Therefore, we use `Thread.sleep` to let the transferLeadership RPC return to the caller before the leader shuts down its RPC server. 
Without the `sleep` call, there could be a situation where the leader shuts down its RPC server before responding to the transferLeadership RPC. This would cause the caller to retry the RPC erroneously.

maobaolong:
@ggezer @jiacheliu3 @jenoudet Thanks for your review, as we discussed in the `slack` and `wechat`, I decided to close this PR.

Issue #13959, opened 2021-08-19
Avoid or warn lock pool resource shortage


apc999:
**Is your feature request related to a problem? Please describe.**

When `BlockWorkerClientPool` is used up (default 1024 threads), applications e.g., writing through Alluxio will stall.
Currently it is hard to tell the resource usage

**Describe the solution you'd like**
Ideally avoid resource starvation , or at least showing good message to indicate what's wrong

**Describe alternatives you've considered**

There might be multiple solutions, as proposed by different maintainers:
1 increase this pool size to much larger
2 remove the pool size bound as the purpose is to resuse initialized connection, not for resource binding
3 acquire resource with timeout

**Urgency**
Relatively high

**Additional context**

apc999:
fixed by #14075, so that we introduce a metrics to track the usage of all dynamic resource pool

PR #14075, opened 2021-09-15
Add lock pool metrics

yuzhu:
Add lock pool metrics

Give users a way to monitor lock pool usage. 

Additional metrics

### DIFF
--- a/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
+++ b/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
@@ -282,8 +282,7 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
 
   private void registerGauges() {
     MetricsSystem.registerGaugeIfAbsent(
-        MetricsSystem.getMetricName("ResourcePoolSize_" + getClass().getName() + "_"
-            + Integer.toHexString(System.identityHashCode(this))),
+        MetricsSystem.getResourcePoolMetricName(this),
         () -> size());
   }
 
--- a/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
+++ b/core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
@@ -276,6 +276,8 @@ public abstract class DynamicResourcePool<T> implements Pool<T> {
         }
       }
     }, options.getInitialDelayMs(), options.getGcIntervalMs(), TimeUnit.MILLISECONDS);
+
+    registerGauges();
   }
 
   private void registerGauges() {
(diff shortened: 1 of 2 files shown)


yuzhu:
Alleviate https://github.com/Alluxio/alluxio/issues/13959

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
```diff
@@ -275,6 +276,15 @@ public DynamicResourcePool(Options options) {
...
       }
     }, options.getInitialDelayMs(), options.getGcIntervalMs(), TimeUnit.MILLISECONDS);
+
+    registerGauges();
+  }
+
+  private void registerGauges() {
+    MetricsSystem.registerGaugeIfAbsent(
+        MetricsSystem.getMetricName("ResourcePoolSize_" + getClass().getName() + "_"
+            + Integer.toHexString(System.identityHashCode(this))),
```
yuzhu, line 286: This produces a class type + unique class id string as the name for the gauge.

jiacheliu3: I wonder how the hex string will help. For example, if you have `ResourcePoolSize_BlockMasterClientPool_abcdef` and `ResourcePoolSize_BlockMasterClientPool_ghijkl`, how do you further associate them back to the resource user? 

nit: `ResourcePoolSize` -> `ResourcePool`?

jiacheliu3 (changes requested):
Thanks for the improvement! A metric on resource pool is definitely necessary, i'm all in for this idea.

Added some comments on the naming, PTAL

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
yuzhu: it probably won't, but provides unique naming at the very least.

core/common/src/main/java/alluxio/resource/DynamicResourcePool.java
```diff
@@ -275,6 +276,15 @@ public DynamicResourcePool(Options options) {
         }
       }
     }, options.getInitialDelayMs(), options.getGcIntervalMs(), TimeUnit.MILLISECONDS);
+
+    registerGauges();
+  }
+
+  private void registerGauges() {
+    MetricsSystem.registerGaugeIfAbsent(
+        MetricsSystem.getMetricName("ResourcePoolSize_" + getClass().getName() + "_"
```
apc999, line 285: shall we make this into a util, so other resource pool can follow the same convention to create metrics?

yuzhu:
@jiacheliu3 @apc999 PTAL

apc999 (approved):

apc999:

PR #14101, opened 2021-09-18
Add application/octet-stream to consumed content types in S3 Handler

ZacBlanco:
The AWS Java client uses this content type in its requests. Also, the
body content type is only consumed in a single method anyways, and is
treated the same regardless of its type.

Add `application/octet-stream` to the default mime types which are allowed

Some S3 clients (e.g. Java AWS SDK) use the `application/octet-stream` by default for the request content type. We only ever consume a byte stream in the handlers anyways, so adding `application/octet-stream` to the types should be harmless and increase support.

No


ggezer (approved):

ggezer:

Xenorith:

PR #14104, opened 2021-09-20
[DOCFIX] Improve Async cache docs

yuzhu:
Improve user documentation on async caching

### DIFF
--- a/docs/en/operation/Performance-Tuning.md
+++ b/docs/en/operation/Performance-Tuning.md
@@ -249,6 +249,10 @@ This is most commonly effective in cases where the files being cached are relati
 However, increase this number sparingly, as it will consume more CPU resources on the worker node
 as the number is increased.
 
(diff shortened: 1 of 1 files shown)


jja725:
FYI: after merging the load command updates, those metrics @yuzhu  mentioned in the updates are replaced by the following metrics. 
Worker.CacheRequests: "Total number of cache requests received by this worker"

Worker.CacheRequestsAsync: "Total number of async cache requests received by this worker"

Worker.CacheRequestsSync: "Total number of sync cache requests received by this worker")

Worker.CacheSucceededBlocks: "Total number of cache succeeded blocks in this worker")

yuzhu:
@jja725 is there a way for the user to monitor how many AsyncCache requests were submitted vs succeeded like before? seems like we lost that if we just have `Worker.CacheSucceededBlocks` ?

jja725:
> @jja725 is there a way for the user to monitor how many AsyncCache requests were submitted vs succeeded like before? seems like we lost that if we just have `Worker.CacheSucceededBlocks` ?

yeah during the review, we think it's not that valuable to track the sync/async success number and just total number and a single ratio metric to see the separation between the two should be good. If you think individual metrics provide useful information we can add that back easily I think.

apc999:

Issue #14110, opened 2021-09-21
Introduce deadlines for sync RPC stubs


ggezer:
**Alluxio Version:**
2.6+

**Describe the bug**
In certain conditions, an RPC call might hang for a long time (observed 15+ mins in testing) without throwing any error. This is more probable when the target server is partitioned from the network after receiving the call. RPC calls stuck in this state are still contained in encompassing retry duration. Retry duration is much smaller, so such RPCs will not be retried when they finally give up after a long time.

**To Reproduce**
Pick a random synchronous RPC. (for example, getStatus()). Suspend host's networking after the RPC is received by the target server.

**Expected behavior**
RPC clients should not blocked on particular call for more than X minutes. X should be smaller than total RPC retry duration. 

**Urgency**
ASAP.

**Additional context**
`BlockWorkerClient` employs deadlines already. `AbstractClient` stubs should do the same.

PR #14116, opened 2021-09-21
Introduce deadlines for sync RPC stubs

ggezer:
Fixes https://github.com/Alluxio/alluxio/issues/14110

### DIFF
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.channels.UnresolvedAddressException;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import javax.annotation.concurrent.ThreadSafe;
@@ -85,6 +86,9 @@ public abstract class AbstractClient implements Client {
 
   private final long mRpcThreshold;
 
+  /** Deadline duration for RPC stubs. */
+  protected final long mRpcCallDeadlineDurationMs;
+
   /**
    * Creates a new client base.
    *
@@ -112,6 +116,8 @@ public abstract class AbstractClient implements Client {
     mRetryPolicySupplier = retryPolicySupplier;
     mServiceVersion = Constants.UNKNOWN_SERVICE_VERSION;
     mRpcThreshold = mContext.getClusterConf().getMs(PropertyKey.USER_LOGGING_THRESHOLD);
+    mRpcCallDeadlineDurationMs =
+        mContext.getClusterConf().getMs(PropertyKey.USER_RPC_CALL_DEADLINE_DURATION);
   }
 
   /**
@@ -230,7 +236,8 @@ public abstract class AbstractClient implements Client {
             .setClientType(getServiceName())
             .build();
         // Create stub for version service on host
-        mVersionService = ServiceVersionClientServiceGrpc.newBlockingStub(mChannel);
+        mVersionService = ServiceVersionClientServiceGrpc.newBlockingStub(mChannel)
+            .withDeadlineAfter(mRpcCallDeadlineDurationMs, TimeUnit.MILLISECONDS);
         mConnected = true;
         afterConnect();
         checkVersion(getServiceVersion());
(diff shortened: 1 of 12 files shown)


ZacBlanco (changes requested):

ZacBlanco (approved):

jenoudet (approved):

ggezer:

ggezer:

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4671,6 +4671,13 @@ public String toString() {
               + "an exponential backoff. This property determines the maximum duration to retry for"
               + " before giving up. Note that, this value is set to 5s for fs and fsadmin CLIs.")
           .build();
+  public static final PropertyKey USER_RPC_CALL_DEADLINE_DURATION =
+      new Builder(Name.USER_RPC_CALL_DEADLINE_DURATION)
+          .setDefaultValue("30sec")
```
ZhuTopher, line 4676: @ggezer I have two questions:

1. I'm pretty sure this isn't a valid default value? It is a string with the suffix `"sec"` but [gRPC stub's withDeadlineAfter](https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/AbstractStub.html#withDeadlineAfter-long-java.util.concurrent.TimeUnit-) expects a `long`. Also we use `TimeUnit.MILLISECONDS` so shouldn't this be `30000`?

2. Should we let users configure the deadline duration `TimeUnit` as a property too? Otherwise we should mention in the description below that it is always `TimeUnit.MILLISECONDS`.

ZhuTopher (commented):
I know this was merged already, but I think we should address my comment; if the branch still exists we can repurpose it and open a follow-up PR?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4671,6 +4671,13 @@ public String toString() {
...
+          .setDescription("Alluxio client RPCs could potentially get stuck "
+              + "when underlying keep-alive settings are disabled. "
+              + "This property determines the maximum duration to wait per each invocation.")
+          .build();
```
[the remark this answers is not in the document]
ZhuTopher: Looks like we don't have a dedicated doc regarding RPCs in Alluxio
- https://docs.alluxio.io/os/user/stable/en/search.html?q=rpc

So we can probably add that if/when that gets made 🤞

core/common/src/main/java/alluxio/AbstractClient.java
```diff
@@ -112,6 +116,8 @@ public AbstractClient(ClientContext context, InetSocketAddress address,
     mRetryPolicySupplier = retryPolicySupplier;
     mServiceVersion = Constants.UNKNOWN_SERVICE_VERSION;
     mRpcThreshold = mContext.getClusterConf().getMs(PropertyKey.USER_LOGGING_THRESHOLD);
+    mRpcCallDeadlineDurationMs =
```
ggezer, line 119: @ZhuTopher For 1;
Here's where that value `("30sec")` is converted to a `long` in milliseconds.

core/common/src/main/java/alluxio/AbstractClient.java
```diff
@@ -230,7 +236,8 @@ public synchronized void connect() throws AlluxioStatusException {
             .setClientType(getServiceName())
             .build();
         // Create stub for version service on host
-        mVersionService = ServiceVersionClientServiceGrpc.newBlockingStub(mChannel);
+        mVersionService = ServiceVersionClientServiceGrpc.newBlockingStub(mChannel)
+            .withDeadlineAfter(mRpcCallDeadlineDurationMs, TimeUnit.MILLISECONDS);
```
ggezer, line 240: @ZhuTopher For 2;
Here's how that `long` value in milliseconds is supplied to `.withDeadlineAfter(long, TimeUnit)` method.
We use milliseconds within the code as convention.

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: I provided places where your questions are answered. 
Having said that, I'll anyway provide a follow up PR as I discovered not all sync stubs are covered in this PR

PR #14119, opened 2021-09-22
Introduce deadlines for sync RPC stubs

### LINKED ISSUE #14110: Introduce deadlines for sync RPC stubs

ggezer:
Manual CP from #14116

Fixes #14110

pr-link: #14116
change-id: cid-1be9a2ddc00881e1c03731841ce3e4a1f56af3aa

### DIFF
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.channels.UnresolvedAddressException;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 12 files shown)


ggezer:

PR #14126, opened 2021-09-22
Display block count for each worker in the master webui

maobaolong:

Display block count of each worker in alluxio master ui.

Leverage this, we can clear found the skew of block storage.

image

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
+++ b/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
@@ -278,6 +278,9 @@ public final class MasterWorkerInfo {
         case ADDRESS:
           info.setAddress(mMeta.mWorkerAddress);
           break;
+        case BLOCK_COUNT:
+          info.setBlockCount(getBlockCount());
+          break;
         case WORKER_CAPACITY_BYTES:
           info.setCapacityBytes(mUsage.mCapacityBytes);
           break;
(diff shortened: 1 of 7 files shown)


yuzhu:
@tieujason330 could  you take a look as it relate to UI

tieujason330 (approved):
UI changes LGTM

yuzhu (approved):

yuzhu:

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -278,6 +278,9 @@ public WorkerInfo generateWorkerInfo(Set<WorkerInfoField> fieldRange, boolean is
         case ADDRESS:
           info.setAddress(mMeta.mWorkerAddress);
           break;
+        case BLOCK_COUNT:
```
jiacheliu3, line 281: @maobaolong @yuzhu A shared lock on `MasterWorkerInfo.mBlockListLock` is missing here. You can find how this method should be properly locked in the method comment.

jiacheliu3: sure

maobaolong: @jiacheliu3 Thanks for your remind, would you lease explain more why lost `MasterWorkerInfo.mBlockListLock` here, and what will happen if missing `mBlockListLock `

jiacheliu3: @maobaolong You can see the comments in `MasterWorkerInfo` where the locking mechanism of this class is further explained. If you miss the read lock when reading the block list you may see inconsistent result.

maobaolong: @jiacheliu3 Since there I just read the block count here, I know `mBlocks` is guarded by `mBlockListLock`, and `mBlocks.size()` is more heavy if guarded by `mBlockListLock`, and we get the size just want to expose it for metrics  purpose, so, can we expose a inconsistent block count by this not thread-safe method?

PR #14135, opened 2021-09-23
Revert "Introduce deadlines for sync RPC stubs"

ggezer:
This reverts commit 70112cb226119c7bb73f8ce0c25252d099cdd840.

### DIFF
--- a/core/common/src/main/java/alluxio/AbstractClient.java
+++ b/core/common/src/main/java/alluxio/AbstractClient.java
@@ -44,7 +44,6 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.channels.UnresolvedAddressException;
-import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import javax.annotation.concurrent.ThreadSafe;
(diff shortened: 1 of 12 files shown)


ggezer:

ggezer:
@ZacBlanco @ZhuTopher FYI. I reverted this as I believe keep-alive configuration could help better. Testing that approach in parallel.

ggezer:
@ZacBlanco @ZhuTopher I wanted to follow up with why I reverted this change. Just FYI...

The reason that I introduced deadlines in the first place was that RPC connections to leading master were getting hung when the leader is switched. So those RPCs were not returning back to callers even though there's no RPC server on the target anymore. This was happening randomly, and was happening more frequently in containerized environments.

Attaching deadlines to stubs means that the stub will wait for configured time before it receives a response, or timeout otherwise. Even though this was solution to hung connections on old masters, this was not desired considering some calls take long time, and some calls will take longer than expected when target master is under heavy load. Overall, deadline means a deadline and it's so preventive for adaptability under load. So I reverted it.

As I mentioned, I wanted to try keep-alive settings and I did. See https://github.com/grpc/grpc/blob/master/doc/keepalive.md
By enabling keep-alive for RPC clients, I was able to get the same outcome with this PR without putting a hard bound on timing requirements of RPCs. Just putting below property was sufficient to avoid hung connections.
`alluxio.user.network.rpc.keepalive.time=30sec` (It's set to INT.MAX by default which means no keep-alive pings are employed.)

PR #14138, opened 2021-09-23
Make jvm pause monitor display gc info correctly

qian0817:

Make jvm pause monitor to display gc info correctly.

I found that the jvm pause monitor could not display gc info correctly and always print `No GCs detected` even after gc. This PR fixes this issue.

No user facing changes.

### DIFF
--- a/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
+++ b/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
@@ -180,6 +180,8 @@ public class JvmPauseMonitor {
     List<GarbageCollectorMXBean> gcBeanList = ManagementFactory.getGarbageCollectorMXBeans();
     Map<String, GarbageCollectorMXBean> gcBeanMap = new HashMap<>();
     for (GarbageCollectorMXBean gcBean : gcBeanList) {
+      // #getGarbageCollectorMXBeans will return the same gc bean
+      // so we need to clone a new gc bean.
       gcBeanMap.put(gcBean.getName(), new GarbageCollectorMXBeanView(gcBean));
     }
     return gcBeanMap;
--- a/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
+++ b/core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
@@ -23,7 +23,6 @@ import com.google.common.collect.Sets;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import javax.annotation.concurrent.NotThreadSafe;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
```diff
@@ -143,22 +145,23 @@ private String formatLogString(long extraSleepTime,
...
       oldBean = gcMXBeanMapBeforeSleep.get(name);
       newBean = gcMXBeanMapAfterSleep.get(name);
       if (oldBean == null) {
-        beanDiffs.add("new GCBean created name= '" + newBean.getName() + " count="
+        beanDiffs.add("new GCBean created name= '" + newBean.getName() + "' count="
             + newBean.getCollectionCount() + " time=" + newBean.getCollectionTime() + "ms");
       } else if (newBean == null) {
-        beanDiffs.add("old GCBean canceled name= '" + oldBean.getName() + " count="
+        beanDiffs.add("old GCBean canceled name= '" + oldBean.getName() + "' count="
             + oldBean.getCollectionCount() + " time=" + oldBean.getCollectionTime() + "ms");
```
ZacBlanco, line 158: should we just use string.format for these too?

core/common/src/main/java/alluxio/util/JvmPauseMonitor.java
```diff
@@ -228,4 +233,59 @@ public void run() {
       }
     }
   }
+
+  /**
+   * An unmodifiable view of a garbage collector MX Bean.
+   */
+  public static class GarbageCollectorMXBeanView implements GarbageCollectorMXBean {
```
ZacBlanco, line 240: I don't fully understand why this class is needed

Also, wouldn't it be more memory efficient (and easier to code) to just hold a reference to the bean, and call the methods themselves instead of storing all these variables?

qian0817: If you run this code
```java
public static void main(String[] args) {
  List<GarbageCollectorMXBean> oldBean = ManagementFactory.getGarbageCollectorMXBeans();
  System.out.println(oldBean);
  System.out.println("old bean info:");
  oldBean.forEach(it -> System.out.println(formatGcBean(it)));
  System.gc();
  System.out.println("===== system gc =====");
  List<GarbageCollectorMXBean> newBean = ManagementFactory.getGarbageCollectorMXBeans();
  System.out.println(newBean);
  System.out.println("old bean info:");
  oldBean.forEach(it -> System.out.println(formatGcBean(it)));
  System.out.println("new bean info:");
  newBean.forEach(it -> System.out.println(formatGcBean(it)));
}

private static String formatGcBean(GarbageCollectorMXBean it) {
  return "name:" + it.getName() + " count:" + it.getCollectionCount();
}
```
and the result is 
```
[sun.management.GarbageCollectorImpl@378fd1ac, sun.management.GarbageCollectorImpl@49097b5d]
old bean info:
name:PS Scavenge count:0
name:PS MarkSweep count:0
===== system gc =====
[sun.management.GarbageCollectorImpl@378fd1ac, sun.management.GarbageCollectorImpl@49097b5d]
old bean info:
name:PS Scavenge count:1
name:PS MarkSweep count:1
new bean info:
name:PS Scavenge count:1
name:PS MarkSweep count:1
```
so you can see `ManagementFactory.getGarbageCollectorMXBeans()` will return the same gc bean instance, so I need this class to store previous data.

qian0817:
Please take a look. @ZacBlanco @ggezer

ZacBlanco (approved):

LuQQiu:

LuQQiu:
@Xenorith cherry-pick?

PR #14140, opened 2021-09-23
Add worker threadPool and task metrics

xichen01:

Add metrics to identify how many jobs started in master.

Add a way to monitor worker load info. Include worker read/write thread pool allowed max size,  current size, actively count, and completed task count

no

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
+++ b/core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
@@ -14,6 +14,8 @@ package alluxio.worker.grpc;
 import alluxio.conf.ServerConfiguration;
 import alluxio.Constants;
 import alluxio.conf.PropertyKey;
+import alluxio.metrics.MetricKey;
+import alluxio.metrics.MetricsSystem;
 import alluxio.security.User;
 import alluxio.security.authentication.AuthenticatedClientUser;
 import alluxio.util.ThreadFactoryUtils;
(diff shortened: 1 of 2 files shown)


core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
```diff
@@ -45,11 +47,29 @@
               PropertyKey.WORKER_NETWORK_ASYNC_CACHE_MANAGER_QUEUE_MAX)),
           ThreadFactoryUtils.build("CacheManagerExecutor-%d", true)));
 
+  private static final ThreadPoolExecutor BLOCK_READER_THREAD_POOL_EXECUTOR =
+      new ThreadPoolExecutor(THREADS_MIN, ServerConfiguration.getInt(
+          PropertyKey.WORKER_NETWORK_BLOCK_READER_THREADS_MAX), THREAD_STOP_MS,
+          TimeUnit.MILLISECONDS, new SynchronousQueue<>(),
+          ThreadFactoryUtils.build("BlockDataReaderExecutor-%d", true));
+
+  static {
```
maobaolong, line 56: Would you please put the following code to L87 to keep only a static block in this file.

maobaolong (changes requested):
Thanks for provide these useful metric, left a minor comment inline, please take a look

core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java
xichen01: I have moved it to an init block

yuzhu:
@ggezer do you have any perf concern about these newly added metrics?

PR #14144, opened 2021-09-23
Retry the eviction on delete page file failures

beinan:

Retry the eviction when the deletion of the victim page file failed (it has been deleted by others)

  As FB team shared with us, we might see a spike of page put failures if we delete all the page files on the disk.  I think this one might also occur when there is a race on delete the same page file.   So add a retry might help the put operation get through in case the victim file has been deleted by some one else.

No


rongrong (approved):

beinan:

PR #14156, opened 2021-09-27
Fix leader master priority always shows 0

bzheng888:
Fix the problem that leader priority always shows 0 when use 'bin/alluxio fsadmin journal quorum info' command get raft information.
Shell command shows wrong information about the leader priority.
No
No


maobaolong (approved):
Good catch, LGTM, thanks for your contribution!

ggezer (approved):
Thanks for catching this @bzheng888 !

jenoudet:
Thanks for the fix. Could you merge master into your branch? The test that was failing should be removed from master and so your fix should pass all tests.

ggezer:
@bzheng888 @jenoudet We don't prefer merge commits in PRs. Please remove merge commit and rebase on master.

Issue #14113, opened 2021-09-21
Improve partition tolerance for Alluxio FS clients


ggezer:
**Summary**
Alluxio FS is not partition tolerant. When an RPC modifies FS state but fails at client due to sudden partitioning, that client will retry its request blindly and might end up failing still when network partition disappears.

A typical scenario demonstrates the problem:

1-Client calls createFile()
2-Master executes the request
3-Before returning success
   -- transient network partition happens
   -- Alluxio master crashes
   -- ...
4-Client receives failure and initiates a retry
5-Retried request lands on a master that has already created the file
6-Client receives `FileAlreadyExists` exception for the file created by itself

Similar situation can arise for completeFile/delete/rename.

**Special case of embedded journal:**
In addition to crash/partition scenarios, embedded journal introduces another case where such failures could happen. In embedded journal, modifications to journal are advertised to enough master nodes before returning success to client. However, this advertisement can be partial due to many reasons. In that case, client will receive an error but master nodes that have been introduced to its change will still hold on to that new knowledge. This is common especially during leadership changes, which makes it likely for client that received failure to retry its action on another master node that potentially has already executed that change against its state. Hence, the same issue.

**Solution:**
Possible catch-all solution to greatly improve partition tolerance is to link individual RPCs to corresponding journal entries they end up creating. By caching this link information, we could provide a safety net over such RPCs. This solution would update above scenario as follows:

1-Client calls createFile() **(with opId:12345)**
2-Master executes the request
    **- standby masters replay the journal and caches opId:12345 as executed.**
3-Before returning success
   -- transient network partition happens
   -- Alluxio master crashes
   -- ...
4-Client receives failure and initiates a retry **(again with opId:12345)**
5-Retried request lands on a master that has already created the file
     **- Master detects that opId:12345 was already journaled.
     - Returns success for the RPC**
6-Client receives success

**Urgency**
This is a systemic problem that can't be fully addressed but could be greatly improved with proposed solution.
Special case of embedded journal adds to the urgency of this problem.

PR #14162, opened 2021-09-27
Implement FS operation tracking for improving partition tolerance

ggezer:
Fix https://github.com/Alluxio/alluxio/issues/14113

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
@@ -24,12 +24,14 @@ import alluxio.exception.ExceptionMessage;
 import alluxio.exception.PreconditionMessage;
 import alluxio.exception.status.UnavailableException;
 import alluxio.grpc.CompleteFilePOptions;
+import alluxio.grpc.FileSystemMasterCommonPOptions;
 import alluxio.metrics.MetricKey;
 import alluxio.metrics.MetricsSystem;
 import alluxio.resource.CloseableResource;
 import alluxio.util.CommonUtils;
 import alluxio.util.FileSystemOptions;
 import alluxio.wire.BlockInfo;
+import alluxio.wire.OperationId;
 import alluxio.wire.WorkerNetAddress;
 
 import com.codahale.metrics.Counter;
@@ -41,6 +43,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.UUID;
 
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.annotation.concurrent.ThreadSafe;
@@ -142,6 +145,8 @@ public class AlluxioFileOutStream extends FileOutStream {
       }
 
       CompleteFilePOptions.Builder optionsBuilder = CompleteFilePOptions.newBuilder();
+      optionsBuilder.setCommonOptions(FileSystemMasterCommonPOptions.newBuilder()
+          .setOperationId(new OperationId(UUID.randomUUID()).toFsProto()).buildPartial());
       if (mUnderStorageType.isSyncPersist()) {
         if (mCanceled) {
           mUnderStorageOutputStream.cancel();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2728,7 +2728,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_SIZE =
       new Builder(Name.MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_SIZE)
-          .setDefaultValue(100000)
+          .setDefaultValue(100_000)
           .setDescription("Size of fs operation retry cache.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2721,7 +2721,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED =
       new Builder(Name.MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED)
           .setDefaultValue(true)
-          .setDescription("If enabled, each fs operation will be tracked on all masters, "
+          .setDescription("If enabled, each filesystem operation will be tracked on all masters, "
               + "in order to avoid re-execution of client retries.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
@@ -4214,7 +4214,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey USER_FILE_INCLUDE_OPERATION_ID =
       new Builder(Name.USER_FILE_INCLUDE_OPERATION_ID)
           .setDefaultValue(true)
-          .setDescription("Whether to send a unique operation id with designated FS calls.")
+          .setDescription(
+              "Whether to send a unique operation id with designated filesystem operations.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2718,6 +2718,21 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED =
+      new Builder(Name.MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("If enabled, each fs operation will be tracked on all masters, "
+              + "in order to avoid re-execution of client retries.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_SIZE =
+      new Builder(Name.MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_SIZE)
+          .setDefaultValue(100000)
+          .setDescription("Size of fs operation retry cache.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.MASTER)
+          .build();
 
   //
   // Worker related properties
@@ -4193,6 +4208,13 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               + "bottom going up (-1 identifies the last tier, -2 identifies the second to "
               + "last tier, and so on). If the absolute value of the provided value is "
               + "greater than the number of tiers, it identifies the first tier.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.CLIENT)
+          .build();
+  public static final PropertyKey USER_FILE_INCLUDE_OPERATION_ID =
+      new Builder(Name.USER_FILE_INCLUDE_OPERATION_ID)
+          .setDefaultValue(true)
+          .setDescription("Whether to send a unique operation id with designated FS calls.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
@@ -5871,6 +5893,10 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     //
     public static final String MASTER_FILE_SYSTEM_LISTSTATUS_RESULTS_PER_MESSAGE =
         "alluxio.master.filesystem.liststatus.result.message.length";
+    public static final String MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED =
+        "alluxio.master.filesystem.operation.retry.cache.enabled";
+    public static final String MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_SIZE =
+        "alluxio.master.filesystem.operation.retry.cache.size";
 
     //
     // Secondary master related properties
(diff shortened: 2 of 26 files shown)


core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
```diff
@@ -142,6 +145,8 @@ public void close() throws IOException {
       }
 
       CompleteFilePOptions.Builder optionsBuilder = CompleteFilePOptions.newBuilder();
+      optionsBuilder.setCommonOptions(FileSystemMasterCommonPOptions.newBuilder()
+          .setOperationId(new OperationId(UUID.randomUUID()).toFsProto()).buildPartial());
```
ZacBlanco, line 149: how fast is `UUID.randomUUID()`? Should we worry about this being called very frequently since it's in a somewhat hot code path? does it use a normal `Random` instance or `ThreadLocalRandom` under the hood?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2718,6 +2718,21 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED =
+      new Builder(Name.MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED)
+          .setDefaultValue(true)
+          .setDescription("If enabled, each fs operation will be tracked on all masters, "
```
ZacBlanco, line 2724: prefer the full expression for the sake of our documentation: `... each filesystem operation ...`

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4196,6 +4211,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_FILE_INCLUDE_OPERATION_ID =
+      new Builder(Name.USER_FILE_INCLUDE_OPERATION_ID)
+          .setDefaultValue(true)
+          .setDescription("Whether to send a unique operation id with designated FS calls.")
```
ZacBlanco, line 4217: same comment about `fs` vs filesystem. Anyways, if you're going to keep it short, at least keep capital/lowercases consistent.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
```diff
@@ -83,7 +83,7 @@ public void write(JournalEntry entry) throws IOException, JournalClosedException
     if (mClosed) {
       throw new JournalClosedException("Cannot write to journal. Journal writer has been closed");
     }
-    Preconditions.checkState(entry.getAllFields().size() <= 1,
+    Preconditions.checkState(entry.getAllFields().size() <= 2,
```
ZacBlanco, line 86: Can you help me understand the change to this precondition check?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -399,8 +399,8 @@ public synchronized void applyJournalEntryCommand(TransactionContext commit) {
    */
   private void applyEntry(JournalEntry entry) {
     Preconditions.checkState(
-        entry.getAllFields().size() <= 1
-            || (entry.getAllFields().size() == 2 && entry.hasSequenceNumber()),
+        entry.getAllFields().size() <= 2
+            || (entry.getAllFields().size() == 3 && entry.hasSequenceNumber()),
```
ZacBlanco, line 403: can you help me understand why we needed to update this precondition check?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -119,6 +133,31 @@ public InodeTreePersistentState(InodeStore inodeStore, InodeLockManager lockMana
     mBucketCounter = new BucketCounter(
         ServerConfiguration.getList(MASTER_METRICS_FILE_SIZE_DISTRIBUTION_BUCKETS, ",")
             .stream().map(FormatUtils::parseSpaceSize).collect(Collectors.toList()));
+    mRetryCacheEnabled =
+        ServerConfiguration.getBoolean(MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_ENABLED);
+    mCache = CacheBuilder.newBuilder()
+        .maximumSize(ServerConfiguration.getInt(MASTER_FILE_SYSTEM_OPERATION_RETRY_CACHE_SIZE))
+        .build();
```
ZacBlanco, line 140: just a consideration to reduce the configuration footprint, could you make the cache enabled if the size > 0 and disabled if <= 0? This way you only need one config key

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -217,7 +261,12 @@ public long applyAndJournal(Supplier<JournalContext> context, NewBlockEntry entr
   public void applyAndJournal(Supplier<JournalContext> context, RenameEntry entry) {
     try {
       applyRename(entry);
-      context.get().append(JournalEntry.newBuilder().setRename(entry).build());
+      JournalEntry.Builder builder = JournalEntry.newBuilder().setRename(entry);
```
ZacBlanco, line 264: This is in a different order than the `applyDelete` operation. Does the ordering of these matter?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -78,6 +85,13 @@
   private final InodeStore mInodeStore;
   private final InodeLockManager mInodeLockManager;
 
+  private final boolean mRetryCacheEnabled;
+  /**
+   * This cache is used by this persistent state and InodeTree
+   * for keeping track of recently applied operations on inodes.
+   */
+  private final Cache<OperationId, Boolean> mCache;
```
ZacBlanco, line 93: nit: `mOpIdCache`?

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
jiacheliu3: I remember seeing complaints that UUID is slow in Java, [this](http://braveo.blogspot.com/2013/05/uuidrandomuuid-is-slow.html) is not alone. 
I think essentially we are using UUID to guard the operations affected in a master flip, so global uniqueness is not a must. We just want uniqueness in the active operations right? That can be an argument favoring `ThreadLocalRandom`.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
jiacheliu3: There is an `OpId` added to the journal entry so there's one additional field now. 

But it's good to elaborate in the comments.

core/transport/src/main/proto/proto/journal/journal.proto
```diff
@@ -16,9 +16,20 @@ import "proto/journal/table.proto";
...
+message JournalOpPId {
+    optional int64 mostSignificantBits = 1;
+    optional int64 leastSignificantBits = 2;
+}
+
+// next available id: 53
 message JournalEntry {
+  // shared fields.
   optional int64 sequence_number = 1;
+  optional JournalOpPId operationId = 52;
```
jiacheliu3, line 31: Is it intended that the order here is messed up? LOL

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
jiacheliu3: If we just need uniqueness across active operations, we may also reduce the size of OpId from 2 int64 to just one. Well the performance difference will not be too much though.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -727,6 +798,8 @@ public void resetState() {
     mInodeStore.clear();
     mReplicationLimitedFileIds.clear();
     mPinnedInodeFileIds.clear();
+    mCache.invalidateAll();
+    mCache.cleanUp();
```
ggezer, line 802: remove this.

jiacheliu3:
I have some general high level questions. 

1. What is the testing plan? I feel like we need some unit test coverage.
2. How much of backward compatibility does that break? Does old clients not work with the new master? How about the old journal?

core/client/fs/src/main/java/alluxio/client/file/AlluxioFileOutStream.java
ggezer: We want uniqueness across all clients of Alluxio. I don't see how can we achieve this other than using `UUID`.

ggezer: As for speed, I haven't seen any degradation on overall master throughput.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
ggezer: What is a waste?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
ggezer: Same with above. Due to additional operationId field.

core/transport/src/main/proto/proto/journal/journal.proto
ggezer: No idea!

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
ggezer: I'm pretty sure somebody made a quick hack on some issue years ago.
This part is where I'm making sure followers will see this entry with an operationId. So from their perspective, it didn't happen if it's not on the journals. 

There has to be back-off on the `DefaultFileSystemMaster` for when things go wrong for this particular master, for which we'll get here again.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalWriter.java
ggezer: Consider the scenario where retries handled by multiple Alluxio masters. We need to mark each journal entry with a unique operationId. Or followers would have no idea which operation is applied or not.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
ggezer: I'm favoring to keep this as is. In the future, we might need to change the eviction model to a timeout.

ggezer:
> I have some general high level questions.
> 
> 1. What is the testing plan? I feel like we need some unit test coverage.
> 2. How much of backward compatibility does that break? Does old clients not work with the new master? How about the old journal?

I will work on 1.
For backward compatibility, we're just adding an additional field to journal entries, so it should be covered.
 - new masters can work with old journal
 - old clients can work with new masters.

jenoudet (approved):

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
```diff
@@ -410,64 +408,52 @@ private void promoteFollower() throws Exception {
 
   @Test
   public void gainPrimacyDuringCatchup() throws Exception {
-    // TODO(feng): remove this test when remote journal write is deprecated
-    after();
-    try (Closeable r = new ConfigurationRule(
-        PropertyKey.MASTER_EMBEDDED_JOURNAL_WRITE_REMOTE_ENABLED, "true",
-        ServerConfiguration.global()).toResource()) {
-      before();
```
ZacBlanco, line 418: Do we not need this config value any more?

ZacBlanco (commented):
two minor questions

ZacBlanco (approved):

core/server/master/src/main/java/alluxio/master/file/contexts/CreateDirectoryContext.java
```diff
@@ -108,6 +109,14 @@ public CreateDirectoryContext setDefaultAcl(List<AclEntry> defaultAcl) {
     return mDefaultAcl;
   }
 
+  @Override
+  public OperationId getOperationId() {
+    if (getOptions().hasCommonOptions() && getOptions().getCommonOptions().hasOperationId()) {
+      return OperationId.fromFsProto(getOptions().getCommonOptions().getOperationId());
+    }
+    return super.getOperationId();
```
jiacheliu3, line 117: Will this return a `null` from `super.getOperationId()` and possibly introduce an NPE for someone else?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
```diff
@@ -687,6 +753,11 @@ private long getIdFromPath(Path path) {
 
   @Override
   public boolean processJournalEntry(JournalEntry entry) {
+    // Account for entry in retry-cache.
+    if (entry.hasOperationId()) {
```
jiacheliu3, line 757: Will it be better to cache after the entry is applied?

tests/src/test/java/alluxio/server/ft/FileSystemMasterFaultToleranceIntegrationTest.java
```diff
@@ -0,0 +1,353 @@
...
+    // It should just return successfully.
+    FileInfo fileInfo = leadingFsMaster.createFile(testPath1, context);
+    Assert.assertEquals(fileInfo.getFileId(), leadingFsMaster.getFileId(testPath1));
+
+    // Create the file again with a different context.
+    // It should fail with `FileAlreadyExistException`.
+    try {
+      leadingFsMaster.createFile(testPath1, context2);
+      Assert.fail("Retry with a different operation-id should have failed.");
+    } catch (FileAlreadyExistsException e) {
```
jiacheliu3, line 108: Maybe use `assertThrows()` like https://github.com/Alluxio/alluxio/blob/0d565aee2ffeb44a0f1ee9d579677a8a3134eec1/core/client/fs/src/test/java/alluxio/client/file/FileOutStreamTest.java#L457

jiacheliu3 (approved):
LGTM, thanks for the fix! Left a few minor comments

core/server/common/src/test/java/alluxio/master/journal/raft/RaftJournalTest.java
ggezer: I don't think so. The old logic that was protected with this property was not right anyway.
Created this to reconsider this property and its client: https://github.com/Alluxio/alluxio/issues/14198

core/server/master/src/main/java/alluxio/master/file/meta/InodeTreePersistentState.java
ggezer: Right.

core/server/master/src/main/java/alluxio/master/file/contexts/CreateDirectoryContext.java
ggezer: Right. All code path is marked with @Nullable annotation. I have now added where it was missing as well.

ggezer:

PR #14164, opened 2021-09-28
Update JobService Stress bench

jja725:

Add NoOp operation to Jobservice Stress bench 

Better test scalability of job master

  1. change in user-facing APIs


stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
```diff
@@ -238,40 +259,106 @@ private void applyOperation(String dirPath) throws IOException, AlluxioException
...
+            info = mJobMasterClient.getJobStatus(jobId);
+            if (statuses.contains(info.getStatus())) {
+              singleton.set(info);
+            }
+            return statuses.contains(info.getStatus());
+          } catch (IOException e) {
+            throw Throwables.propagate(e);
+          }
+        }, WaitForOptions.defaults().setTimeoutMs(30 * Constants.SECOND_MS));
+    JobInfo jobInfo = singleton.get();
```
yuzhu, line 359: is the reason for using an AtomicReference here because the waitfor signature does not allow return values? what if the reference is never set (e.g. an exception was thrown)?

yuzhu (changes requested):
@jja725 Overall looks good, thanks for the improvement! 

could you add some more description on how this noop test can be launched? as well describe any user-facing changes

stress/shell/src/main/java/alluxio/stress/cli/StressJobServiceBench.java
jja725: yes, I'm using the same function as JobTestUtils(think of refactoring them in the following PR). I think if any exception happened during the waiting time, It would propagate the exception and would try to catch that exception in line 215.

yuzhu (approved):

yuzhu:

PR #14174, opened 2021-09-28
Implement Resource Leak Detection

ZacBlanco:

Implement resource leak detection by extending Netty's detection framework in Alluxio

This requires no API changes

Introduces a new property key `alluxio.leak.detector.level`. Its use is documented in the description of the PK.

### Additional Information

I confirmed this basic PoC works with the following code. It needs to open a resource, then not close. After which, a GC must be triggered and a message will be logged.

```
  @Test
  public void testRLD() throws Exception {
    for (int i = 0; i < 1000000; i++) {
      createNotCloseIter();
    }
    System.gc();
    for (int i = 0; i < 50; i++) {
      byte[] mem = new byte[1024 * 1024 * 1024];
      mem[ThreadLocalRandom.current().nextInt(1024 * 1024)] += 1;
    }
    System.gc();
    Thread.sleep(1000L);
  }
```

Sample log message (truncated)

```
2021-09-28 09:21:34,826 [main] ERROR resource.AlluxioResourceLeakDetector (AlluxioResourceLeakDetector.java:reportTracedLeak) - LEAK: CloseableResource.close() was not called before resource is garbage-collected. See https://docs.alluxio.io/blah/blah/blah for more information about this message.
Recent access records: 
Created at:
	alluxio.resource.CloseableResource.<init>(CloseableResource.java:51)
	alluxio.resource.CloseableIterator.<init>(CloseableIterator.java:35)
	alluxio.resource.CloseableIterator$1.<init>(CloseableIterator.java:46)
	alluxio.resource.CloseableIterator.noopCloseable(CloseableIterator.java:46)
	alluxio.resource.CloseableIteratorTest.createNotCloseIter(CloseableIteratorTest.java:58)
	alluxio.resource.CloseableIteratorTest.testRLD(CloseableIteratorTest.java:45)
	java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	java.base/java.lang.reflect.Method.invoke(Method.java:566)
        .....

```


core/common/src/main/java/alluxio/resource/AlluxioResourceLeakDetector.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param resourceType the resource class
+   * @param samplingInterval on average, how often a resource should be tracked
+   */
+  public AlluxioResourceLeakDetector(Class<?> resourceType, int samplingInterval) {
+    super(resourceType, samplingInterval);
+  }
+
+  @Override
+  protected void reportTracedLeak(String resourceType, String records) {
```
jiacheliu3, line 39: For our better understanding, if you figure out how this traced/untraced leak differ, I think it makes sense to add to the comment here although it is `@Override`.

jiacheliu3:
Is this still a draft?

jiacheliu3 (commented):
I think this overall looks great. Surprised at the fact that we are not touching too much code at all! Does it make sense to add a few unit tests?

ZacBlanco:
Yes, still a draft, I just tagged you guys in case you wanted an early look. I'll comment when it's ready for review.

@jiacheliu3 the hard part is that testing is not very straightforward. I'll have to think about it a bit more.

The reason is because the leak detector only logs and reports the leaks when the JVM GC is triggered (because that's when it collects the unreachable references). In order to test the leak detector we'd need a reliable way of triggering GC, which I've found is actually quite difficult (and time consuming, from a CI perspective). I'm open to other ideas if you have any w.r.t to testing this

core/common/src/main/java/alluxio/resource/AlluxioResourceLeakDetector.java
ZacBlanco: Good point. I'll add some comments

jiacheliu3:
@ZacBlanco yea testing something triggered only by GC is difficult, especially some JVMs don't even let you trigger it manually. I'll help thinking.

ZacBlanco:
ok, I might have a bit of polish left, but I think this is the meat of it. Feel free to review

PR #14180, opened 2021-09-29
Optimize the worker page of the master web ui

### LINKED ISSUE #14179: Optimize the worker page of the master web ui
**Alluxio Version:**
2.6.1

**Describe the bug**
We deploys alluxio services on kubernetes. When we re-install alluxio worker daemonset. The master web ui displays the same node name of `Live Workers` and `Lost Workers`.

**To Reproduce**
We deploys alluxio services on kubernetes. When I re-install alluxio worker daemonsets. The master web ui displays the same node name of `Live Workers` and `Lost Workers`.

**Expected behavior**
Want to display `Live Workers` and not display `Lost Workers`.

**Urgency**
Normal.

**Additional context**
I find #13861. When I test alluxio worker daemonsets using master branch. The web display

I want to display only `192.168.118.128 (127.0.0.3)` that is the newest live node. It replaces the old one.

ljl1988com:
Fixes #14179 .

Optimize the worker page of the master web ui.

We deploys alluxio services on kubernetes. When we re-install alluxio worker daemonset. The master web ui displays the same node name of Live Workers and Lost Workers.

None


ljl1988com:
I will refresh the container host of the old worker.  If not refresh, the ui display the old container host. It will be a bug.

core/common/src/main/java/alluxio/wire/WorkerNetAddress.java
```diff
@@ -181,7 +181,6 @@ public boolean equals(Object o) {
     }
     WorkerNetAddress that = (WorkerNetAddress) o;
     return mHost.equals(that.mHost)
-        && mContainerHost.equals(that.mContainerHost)
```
maobaolong, line 184: Why you remove `mContainerHost`

core/common/src/main/java/alluxio/wire/WorkerNetAddress.java
```diff
@@ -191,7 +190,7 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(mHost, mContainerHost, mDataPort, mRpcPort, mWebPort,
```
maobaolong, line 194: Why remove `mContainerHost ` here

maobaolong (commented):
Left a couple of comment.

yuzhu:
@tieujason330 PTAL, thanks

ljl1988com:
@tieujason330 I add unit test. PTAL.

yuzhu:
@ljl1988com could you please also address @maobaolong 's comment?

ljl1988com:
@maobaolong Thanks for your review. When in k8s `daemonset` ensures to run only a copy of a pod on a node. So the worker lost and new worker re-register on the same node. I think only one of them will be scheduled, and the other will be destroyed.

maobaolong:
@ljl1988com Oh, that is the different between us, we deploy alluxio worker to k8s as a deployment, because the k8s is a public resource pool with 2k+ node, we cannot use daemonset, we can only scale the replicas to less than 1k.

So, if the worker workload is deployment, the container host is necessary, right?

ljl1988com:
@maobaolong I know your scene. In k8s the workers are often lost. The worker page of the master web ui displays many lost workers which never re-register in place. I will think how to resolve the problem better.

PR #14190, opened 2021-09-30
Add missing lock in DefaultBlockMaster.extractWorkerInfo

jiacheliu3:

Add missing lock in https://github.com/Alluxio/alluxio/pull/14126#discussion_r718334076 and refactored some locking logic for `MasterWorkerInfo`.

Add the missing read lock to ensure the data consistency.

NA

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
@@ -15,9 +15,12 @@ import alluxio.grpc.GetWorkerReportPOptions;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableSet;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.Set;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -195,6 +198,8 @@ public final class GetWorkerReportOptions implements Serializable {
     STATE,
     WORKER_USED_BYTES,
     WORKER_USED_BYTES_ON_TIERS;
+
+    public static final Set<WorkerInfoField> ALL = EnumSet.allOf(WorkerInfoField.class);
 
     /**
      * @return the proto representation of this worker info fields
--- a/core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
@@ -15,11 +15,9 @@ import alluxio.grpc.GetWorkerReportPOptions;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
-import com.google.common.collect.ImmutableSet;
 
 import java.io.Serializable;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -261,19 +267,16 @@ public void addLostStorage(Map<String, StorageList> lostStorage) {
...
+   * The required locks will be determined internally based on the fields.
    *
    * @param fieldRange the client selected fields
    * @param isLiveWorker the worker is live or not
    * @return generated worker information
    */
   public WorkerInfo generateWorkerInfo(Set<WorkerInfoField> fieldRange, boolean isLiveWorker) {
     WorkerInfo info = new WorkerInfo();
-    Set<WorkerInfoField> checkedFieldRange = fieldRange != null ? fieldRange :
-        new HashSet<>(Arrays.asList(WorkerInfoField.values()));
```
jiacheliu3, line 275: We used to allow `null` and handle here. `null` is no longer allowed, in order to avoid confusion.

core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -623,4 +626,23 @@ ReentrantReadWriteLock getLock(WorkerMetaLockSection lockType) {
...
   }
+
+  /**
+   * Locks the corresponding read locks for the specified worker information fields.
+   * This is a wrapper of {@link MasterWorkerInfo#lockWorkerMeta(EnumSet, boolean)}
+   *
+   * @param fieldRange a set of {@link WorkerInfoField}
+   * @return a {@link LockResource} of the {@link WorkerMetaLock}
+   */
+  public LockResource lockWorkerMetaForInfo(
```
jiacheliu3, line 637: This is just a wrapper to make the logic easier for user.

In terms of why we don't lock internally in `MasterWorkerInfo`, the methods in this class are usually called externally in combination so it's not easy for ALL methods here to embed locking internally. For consistency (so that we don't have some methods that lock internally and the other externally), we require ALL methods here to lock externally if needed.

Feel free to drop a comment if you have better ideas here.

core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
```diff
@@ -15,9 +15,12 @@
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableSet;
```
maobaolong, line 18:
```suggestion
```

core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
```diff
@@ -15,9 +15,12 @@
...
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.Collections;
```
maobaolong, line 22:
```suggestion
```

maobaolong:
@jiacheliu3 Left a couple style fix suggestion. And left a question in https://github.com/Alluxio/alluxio/pull/14126#discussion_r718334076

core/client/fs/src/main/java/alluxio/client/block/options/GetWorkerReportOptions.java
jiacheliu3: nice catch!

elega (approved):

jiacheliu3:

PR #14191, opened 2021-09-30
Support free all blocks from specified worker

xichen01:
Add fsadmin command "freeWorkerTier"
```
$ .bin/alluxio fsadmin freeWorkerTier --help
free(evict) all blocks from all worker 
usage: freeWorkerTier [--force] [--workers <worker_hostname>]
Options:
 -f, --force    If given, will remove all blocks that can be found in worker, whether persistent or not
 -w, --workers  If given, free blocks from specified workers. Pass in the worker hostnames separated by comma
```
This command is similar alluxio fs free, but It can free all blocks of a specified worker node rather than a directory. It doesn‘t depend on the metadata of the master node. So you can use it when all master nodes are destroyed.
The worker cache in the tier will be invalid when all master nodes are destroyed, because all block metadata was lost. In this case, you should use this command to free all workers block to reset cluster.
This command will free the MUST_CACHE block, which is another difference with "alluxio fs free". Same with "alluxio fs free" Any pinned/un persisted files cannot be freed unless -f option is specified.

no

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
@@ -71,6 +71,7 @@ import java.net.InetSocketAddress;
 import java.nio.channels.FileChannel;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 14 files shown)


core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -537,6 +538,26 @@ public void removeBlock(long sessionId, long blockId)
     mLocalBlockStore.removeBlock(sessionId, blockId);
   }
 
+  @Override
+  public void removeAllBlock(long sessionId)
+      throws InvalidWorkerStateException, IOException {
+    BlockStoreMeta mBlockStoreMeta = getStoreMetaFull();
+    for (List<Long> mBlockIds : mBlockStoreMeta.getBlockList().values()) {
```
maobaolong, line 545: As you want to remove all block, so you can remove all block files in the cache storage tier location and clear all the blocks in `mLocalBlockStore`, any thoughts?

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorker.java
```diff
@@ -537,6 +538,26 @@ public void removeBlock(long sessionId, long blockId)
...
+        } catch (BlockDoesNotExistException e) {
+          LOG.info("No such block id: " + mBlockId);
+        }
+      }
+    }
+  }
+
+  @Override
+  public void tryRemoveAllBlock(long sessionId) throws IOException {
+    mLocalBlockStore.tryFreeAllSpace(sessionId);
```
maobaolong, line 558: why we have to use two api method here, can we use only one method with a removeAllBlockOption to specify how to free.

maobaolong (commented):
Thanks for submit this useful feature, left some comments inline.

PR #14193, opened 2021-09-30
Add Jobservice bench maxThroughput for noop

jja725:

Add Jobservice bench maxThroughput for noop

  1. change in user-facing APIs
  3. webui


stress/common/src/main/java/alluxio/stress/jobservice/JobServiceMaxThroughputSummary.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * The summary for the max throughput test.
+ */
+public final class JobServiceMaxThroughputSummary implements Summary {
```
yuzhu, line 32: could you describe at a high level how this one is different from MaxThroughputSummary class and the rationale for having two separate classes?  would it be possible to have a templated class?

stress/shell/src/main/java/alluxio/stress/cli/suite/JobServiceMaxThroughput.java
```diff
@@ -0,0 +1,165 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A max throughput suite.
+ */
+public class JobServiceMaxThroughput extends Suite<JobServiceMaxThroughputSummary> {
```
yuzhu, line 39: Similarly, any opportunity for refactor here?

yuzhu (commented):
high level comment, i saw quite a bit of duplicated code. Maybe we can discuss offline the reasons why they could not be refactored?

stress/common/src/main/java/alluxio/stress/jobservice/JobServiceMaxThroughputSummary.java
jja725: I thought we can put refactoring as another PR but I can do it within this PR as well.

PR #14194, opened 2021-09-30
Avoid printing too many blocks on logging

apc999:

Instead of showing 

```
2021-09-28 05:03:00,999 WARN  FileSystemMasterClient - GetPinList() returned [12884901887, 10737418239, 15032385535, ... 4000 others ]] in 2312 ms (>=1000 ms)
```

printing in logs

```
2021-09-28 05:03:00,999 WARN  FileSystemMasterClient - GetPinList() returned in 2312 ms (>=1000 ms)
```

- Worker logs can be spammed by the warning log printing all blocks
- Can be slow down

No

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
+++ b/core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
@@ -28,7 +28,6 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -38,7 +37,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-import java.util.stream.Collectors;
 
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.NotThreadSafe;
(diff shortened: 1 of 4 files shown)


core/server/master/src/main/java/alluxio/master/block/meta/MasterWorkerInfo.java
```diff
@@ -501,21 +499,16 @@ public boolean hasLostStorage() {
...
         .add("id", mMeta.mId)
         .add("workerAddress", mMeta.mWorkerAddress)
         .add("capacityBytes", mUsage.mCapacityBytes)
         .add("usedBytes", mUsage.mUsedBytes)
         .add("lastUpdatedTimeMs", mLastUpdatedTimeMs.get())
         .add("blockCount", mBlocks.size())
-        .add(blockFieldName, blocks)
+        // We truncate the list of block IDs to print, unless it is for DEBUG logs
+        .add("blocks", LOG.isDebugEnabled() ? mBlocks :
+            CommonUtils.summarizeCollection(mBlocks, BLOCK_SIZE_LIMIT))
```
yuzhu, line 511: Default  `BLOCK_SIZE_LIMIT` is 100, is that still too spammy?  IMHO, i really don't see the difference between 10 entries vs 100 entries..  if we are hoping to see a particular block id in the 100 entries, we will likely be out of luck, considering we tend to lose blocks in batches (e.g lost worker)

apc999: I agree. I think it serves very little meaning to show 10 entries.
I will simply remove the blocks here but just showing the number of blocks

yuzhu:
checkstyle?

jiacheliu3 (approved):

apc999:

PR #14202, opened 2021-10-05
Remove one constraint for listfile in FuseIOStressBench

ssz1997:
In FuseIOStressBench, num-dirs is not needed for running ListFile operation, but the logic compares number of threads and number of directories, and throw error if the number of threads is larger than number of directories. This constraint should not apply to ListFile operation. This PR fixes this problem.


ssz1997:
@LuQQiu PTAL. Thanks!

LuQQiu (approved):
Thanks for fixing!

LuQQiu:

PR #14205, opened 2021-10-07
Fix copyFromLocal command

ryantotti:

Fix copyFromLocal command

Can not copy local dir which has only a subdir to alluxio via "copyFromLocal" command.
e.g. Copy a local ”dir0“ like:

>     dir0
>       |
>        — dir1
>             |
>             — file0  

alluxio will throw "Source /dir0/dir1 is not a file."

No


apc999:
thanks @ryantotti for fixing this issue!

PR #14207, opened 2021-10-07
Add maxThroughput for job service

jja725:
Add MaxThroughput for job service, introduce several abstract classes in stressBench so, later on, we can refactor other stressBench. Would introduce another PR refactoring MaxThroughput(master) after understanding the usage of that component.

The reason we are doing refactoring is that there's lots of duplicate code between job service maxThroughput and maxThroughput(master). And there's also some duplication between the stress benchs.
Notice the key change we introduce is adding a class `GeneralParameter`. We are introducing this class because we need some abstract Parameter class with the operation member variable in MaxThroughput Summary. The operation parameter needs a specific static converter to convert the string to an Enum type. and we can not do resource injection or template(since we have to put converter.class as a value to the Parameter notation for `mOperation`). So we decide to go with an alternative way to add an method to the Parameter class to get the corresponding Operation.

add maxThroughput command
`runClass alluxio.stress.cli.suite.JobServiceMaxThroughput --operation NoOp --target-throughput 2000`

### DIFF
--- a/stress/common/src/main/java/alluxio/stress/common/GeneralParameters.java
+++ b/stress/common/src/main/java/alluxio/stress/common/GeneralParameters.java
@@ -20,8 +20,7 @@ public abstract class GeneralParameters extends Parameters {
 
   /**
    * Notice the function name can't be getOperation since Jackson would transfer this function in to
-   * json value, break Parameter serialization and cause serialization error
-   * 
+   * json value, break Parameter serialization and cause serialization error.
    * @return the Operation Enum
    */
   public abstract Enum<?> operation();


stress/common/src/main/java/alluxio/stress/common/AbstractMaxThroughputSummary.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * abstract class for result of MaxThroughput stressBench.
+ * @param <T> the stress bench parameter
+ * @param <S> the Bench Summary
+ */
+public abstract class AbstractMaxThroughputSummary<T extends GeneralParameters, S extends Summary>
```
jiacheliu3, line 34: @shiqi1122 Will you need to update your PRs due to changes here?

PR #14212, opened 2021-10-08
Add a backoff before worker registers

jiacheliu3:

This adds an option for the worker to backoff before registering with the master. When there are a lot of workers, this can avoid all workers trying to register with the master at once.

All workers attempt to register at the same time, the master will experience a flush of load.

If turned on, the user will experience a potential slow down when all workers are started before they are all registered.


jiacheliu3:
> I don't really understand the impact of this change. I believe this heuristic will be rendered useless with register leases. Don't you think?

Great question. Yes the impact should be really small if we use the register leases. But if people don't have register leases and don't use streaming, I think this will still have its value. Maybe this can be a very safe patch to backport too?

jiacheliu3:
Closed because we chose to implement https://github.com/Alluxio/alluxio/pull/14220

PR #14131, opened 2021-09-22
Add metric of audit log

codings-dan:
### What changes are proposed in this pull request?

Fix #14130 

### Does this PR introduce any user facing changes?

No

PR #14214, opened 2021-10-08
Fix null pointer exception for AuditLog

codings-dan:

Fix a null pointer exception problem when registering AuditLog metric in #14131

When registering `MASTER_AUDIT_LOG_ENTRIES_SIZE` metric, if manually transferLeader , `mAsyncAuditLogWriter` will be set to null , and a null pointer exception will occur at this time.

No


codings-dan:
@apc999 I think this CI test error has nothing to do with this code change, PTAL,thx!

PR #14216, opened 2021-10-09
Add recursive to create a file in case of missing parent directory

yyongycy:
Add recursive to create a file in case of missing parent directory

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/client/StressClientIOBench.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/client/StressClientIOBench.java
@@ -576,7 +576,7 @@ public class StressClientIOBench extends Benchmark<ClientIOTaskResult> {
         case WRITE: {
           if (mOutStream == null) {
             mOutStream = mFs.createFile(new AlluxioURI(mFilePath.toString()),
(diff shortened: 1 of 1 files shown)


jiacheliu3:
You should fetch the latest `UPSTREAM/master` before creating a branch,  then create a branch based on that
```
$ git fetch --all
$ git checkout upstream/master
# Do your changes
$ git checkout -b change-branch
# create PR based on change-branch
```

yyongycy:
> You should fetch the latest `UPSTREAM/master` before creating a branch, then create a branch based on that
> 
> ```
> $ git fetch --all
> $ git checkout upstream/master
> # Do your changes
> $ git checkout -b change-branch
> # create PR based on change-branch
> ```

Sound good. Looks like we won't keep a dev branch for a long time.

jiacheliu3:
> > You should fetch the latest `UPSTREAM/master` before creating a branch, then create a branch based on that
> > ```
> > $ git fetch --all
> > $ git checkout upstream/master
> > # Do your changes
> > $ git checkout -b change-branch
> > # create PR based on change-branch
> > ```
> 
> Sound good. Looks like we won't keep a dev branch for a long time.

Yea we don't have a `dev` vs `master` branch distinction. Rather we only use the `master` branch for simplicity. And by each PR we intend to make them as small and atomic as possible.

PR #14223, opened 2021-10-11
Convert worker registration to a stream

jiacheliu3:

This change adds the following items:
1. Add an option for the worker register logic to be a streaming call instead of unary
2. Add a stressbench implementation

As a side effect, the locking mechanism in `MasterWorkerInfo` is changed from `ReentrantReadWriteLock` to `StampedLock`. See the discussion in this PR for why.

When the worker has a huge amount of blocks to report to the master, the worker register call will be a huge unary RPC call. This incurs a lot of heap allocation on the master side and it's mostly not GC-able before the call finishes. Changing this to streaming makes it more GC-friendly. 

Being able to do some GC during the stream can allow higher concurrency too, so more workers can successfully register at the same time.

Streaming also enables more flow control logic that was impossible when using a unary call. This is not included in this PR but may be added in the future.

The stream will hold the write locks on the `MasterWorkInfo` during the stream. As a side effect, the Web UI and other readers of `MasterWorkerInfo` will block when they are inquiring the status of all workers. When the worker finishes registering, or when the stream times out, the locks will be released and the readers will be unblocked. This is one field to improve on in future changes.


xichen01:
@jiacheliu3 I'm adding some features to the worker register function in this pr:  https://github.com/Alluxio/alluxio/pull/14258 
 I noticed that you had to do some refactoring on this function on the current pr, did I need to sync those changes in my pr now to prevent a mass conflict, or waiting this pr was merged？

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3352,6 +3363,55 @@ public String toString() {
...
+          .setDefaultValue("true")
+          .setDescription("When the worker registers with the master, whether the request should be"
+              + " broken into a stream of smaller batches. This is useful when the worker's storage"
+              + " is large and we expect a large number of blocks. ")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_REGISTER_STREAM_BATCH_SIZE =
+      new Builder(Name.WORKER_REGISTER_STREAM_BATCH_SIZE)
+          .setDefaultValue("500000")
```
ggezer, line 3377:
```suggestion
          .setDefaultValue("1000000")
```

ggezer (approved):
Thanks for the huge effort Jiacheng!

ggezer:

PR #14224, opened 2021-10-11
[DOCFIX] Add one known issue for JVM running in containers

jiacheliu3:
In some early Java 8 versions the jvm in a container do not see resource defined in cgroup. Mention this known issue and solution in the troubleshooting section in Alluxio-on-K8s doc.


jiacheliu3:
> #14243

@yuzhu  I think we should deal with #14243 in a separate PR, as that can involve both code and doc changes.

Issue #14011, opened 2021-09-01
Invalidate client-side metadata cache to be consistent


bzheng888:
**Is your feature request related to a problem? Please describe.**
When we enable client metadata cache(like fuse), the default cache expired time is 2 hour,  we  can't update the cache when the metadata is actually changed until the old metadata cache expired. It affects the workload sometimes.
Based on the above reasons，It's useful to a new command to clear the client metadata cache 
**Describe the solution you'd like**
A new command maybe to clear metadata cache on our own initial.

**Describe alternatives you've considered**
Maybe we can supply a not just clear operation but also other useful operation to handle the client side cache data.

**Urgency**
No
**Additional context**
NO

apc999:
one way possibly to handle this is to make worker services modular and then make it possible to turn on FUSE service and Proxy service, then we can run generic commands to the Proxy server, e.g. `fs ls -f` to force update cache

bzheng888:
> one way possibly to handle this is to make worker services modular and then make it possible to turn on FUSE service and Proxy service, then we can run generic commands to the Proxy server, e.g. `fs ls -f` to force update cache

Thanks for your reply, Let me think about this problem carefully and give a better way to complete this.

LuQQiu:
Fixed by @bzheng888 via https://github.com/Alluxio/alluxio/commit/b892c642ecbce3dd465eb46f634cc419c2c07430

PR #14230, opened 2021-10-11
Add clear metadata cache logic

bzheng888:
A way to complete  issue #14011
Clear client metadata cache by "ls command"

User can clear metadata cache by executing "ls -l /fuse_mount_point/path/.alluxiocli.meatadatacache.drop"


bzheng888:
@LuQQiu Any suggestion or better idea?

LuQQiu:
@bzheng888 it's a great idea to use special folder and ls to fulfill the function!
discussed with @apc999 that we think special folder + ls is awesome.
For the command, we are consider using `ls ./alluxio.metadatacache.drop` or `ls -al ./alluxio.metadatacache.size`.
This is because 
* `:` can be a reserved character in linux/UNIX according to https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/
* this can be convenient for adding other special commands!
@bzheng888 @maobaolong what's your opinions?

bzheng888:
> @bzheng888 it's a great idea to use special folder and ls to fulfill the function! discussed with @apc999 that we think special folder + ls is awesome. For the command, we are consider using `ls ./alluxio.metadatacache.drop` or `ls -al ./alluxio.metadatacache.size`. This is because
> 
> * `:` can be a reserved character in linux/UNIX according to https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/
> * this can be convenient for adding other special commands!
>   @bzheng888 @maobaolong what's your opinions?

Agree! I'm wondering are there any other questions for this pr, like the performance concern and code style?

LuQQiu:
@maobaolong 
```
    mPathResolverCache = CacheBuilder.newBuilder()
        .maximumSize(conf.getInt(PropertyKey.FUSE_CACHED_PATHS_MAX))
        .build(new CacheLoader<String, AlluxioURI>() {
          @Override
          public AlluxioURI load(String fusePath) {
            // fusePath is guaranteed to always be an absolute path (i.e., starts
            // with a fwd slash) - relative to the FUSE mount point
            final String relPath = fusePath.substring(1);
            final Path tpath = mAlluxioRootPath.resolve(relPath);
            return new AlluxioURI(tpath.toString());
          }
        });
```
so we can directly use the path given without resolve it to alluxio uri `  private int getattrInternal(String path, FileStat stat) {`
if it's starting with `/.alluxio` then we go to special path?

bzheng888:
@LuQQiu @maobaolong Solve the problem that alluxio-fuse doesn't mount the root alluxio, change the reserved name to ".alluxio.reserved". PTAL!

bzheng888:
@LuQQiu @maobaolong PTAL!

LuQQiu:
@bzheng888 We can discuss this PR on tonight's AI sync, basically we should try not to directly touch the MetadataCacheFileSystem getStatus or other core operation path for some special operations. But metadata cache filesystem should expose the public methods for APIs to call. It's API's responsibility of thinking how to call the methods. 
For other applications like Presto, we can discuss on the presto side.

On Fuse side, the cleanest way as we discussed so far is to use ioctl to call Fuse, Fuse directly call getMetadataSize/ clearMetadataCache public methods of MetadataCacheFileSystem. In this way, even if special command enabled, we won't largely impact the getStatus performance.

bzheng888:
> pecial operations. But metadata cache filesystem should expose the public methods for APIs to call. It's API's responsibility of thinking how to call the methods. For other applications like Presto, we can discuss on the presto side.
> 
> On Fuse side, the cleanest way as we discussed so far is to use ioctl to call Fuse, Fuse directly call getMetadataSize/ clearMetadataCache public methods of MetadataCacheFileSystem. In this way, even if special command enabled, we won't largely impact the getStatus performance.

@LuQQiu @yyongycy I tried use 'ioctl' clear metadata cache, there are several problems
1.  Java has no ioctl API, so we must write a 'C' program call ioctl POSIX API,  where we put this binary file and let user use?
2. The posix ioctl API require a device or file 'fd', it means we must create a true file in the fuse mount point, and open the file to get fd, then call ioctl, I'm not sure is this a good way.
3. Linux and Mac OS's fuse ioctl is not compatible, i tried to compile in different platform, the clear metadata cache function works on linux, but not work in Mac os and cause an error.
FYI.

bzheng888:
@LuQQiu this pr is updated, please review this at your convenience.

core/client/fs/src/main/java/alluxio/client/file/BaseFileSystem.java
```diff
@@ -98,6 +98,17 @@
...
+  /**
+   * Define the special commands prefix starting with alluxio reserved dir. User can use "ls -l"
+   * command to complete some special operations. For fuse client examples:
+   * "ls -l /mnt_point/.alluxiocli.metadatacache.drop", drop all the metadata cache.
+   * "ls -l /mnt_point/path/.alluxiocli.metadatacache.drop" can clear client metadata cache of
+   * specific path.
+   * "ls -la /mnt_point/.alluxiocli.metadatacache.size" will get metadata cache size in the file
+   * size field.
+   */
+  protected static final String CLEAR_METADATACACHE_RESERVED = Constants.ALLUXIO_RESERVED_DIR
```
LuQQiu, line 110: This is only related to Fuse, move this to fuse package

core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
```diff
@@ -228,4 +263,64 @@ public synchronized void close() throws IOException {
       super.close();
     }
   }
+
+  /**
+   * @param path the path need to drop metadata cache
+   * @return a mock RUIStatus object
+   */
+  private URIStatus dropMetadataCache(String path, String arg) {
```
LuQQiu, line 271: Can we keep MetadataCachingBaseFileSystem clean and move all the logics to Fuse?
here we only need

dropMetadataCache(AlluxioURI uri)
dropAllMetadataCache()
getMetadataCacheSize()

all the other uri processing should be done in the FUSE side to keep this class as clean as possible

core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
```diff
@@ -228,4 +263,64 @@ public synchronized void close() throws IOException {
...
+    return new URIStatus(fi);
+  }
+
+  /**
+   * @param uri include the client metadata cache operation info, the info pattern as blows
+   *             /path/.alluxiocli.metadatacache.cmd.[args]
+   *            the args field is for future using.
+   * @return a mock RUIStatus object
+   */
+  public URIStatus clientMetadataCacheOPHandler(AlluxioURI uri) {
```
LuQQiu, line 297: URI processing should be done in FUSE instead of here

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4212,6 +4212,13 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.CLIENT)
           .build();
+  public static final PropertyKey USER_CLIENT_SPECIAL_COMMAND_ENABLED =
```
LuQQiu, line 4215: Change to FUSE SPECIAL COMMAND ENABLED since it's not generalized.
HDFS Client side should have their own special command

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -229,7 +230,14 @@ public int getattr(String path, FileStat stat) {
   private int getattrInternal(String path, FileStat stat) {
     final AlluxioURI uri = mPathResolverCache.getUnchecked(path);
     try {
-      URIStatus status = mFileSystem.getStatus(uri);
+      URIStatus status = null;
+      // Handle special metadata cache operation
+      if (mConf.getBoolean(PropertyKey.USER_CLIENT_SPECIAL_COMMAND_ENABLED)
+          && mConf.getBoolean(PropertyKey.USER_METADATA_CACHE_ENABLED)) {
```
LuQQiu, line 236: If Fuse special command enabled & path starts with `".alluxiocli"`
-> enter a new method to process the path and get subcommand
-> if subcommand is metadata cache -> then go to new method to process metadata cache related commands

LuQQiu (commented):
Left some comments, thanks!

bzheng888:
@LuQQiu @maobaolong  Update this pr, PTAL!

core/client/fs/src/main/java/alluxio/client/file/MetadataCachingBaseFileSystem.java
```diff
@@ -227,4 +227,39 @@ public synchronized void close() throws IOException {
       super.close();
     }
   }
+
+  /**
+   * @param uri the uri need to drop metadata cache
+   * @return a mock RUIStatus object
+   */
+  public URIStatus dropMetadataCache(AlluxioURI uri) {
```
maobaolong, line 235: I try to guess what @LuQQiu and @apc999 want is to make the `MetadataCachingBaseFileSystem.java` be more clean, so I suggest to make each of the new added method just return the `void` or long type cache size

maobaolong (commented):
Left a comment inline @bzheng888

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -228,7 +229,13 @@ public int getattr(String path, FileStat stat) {
   private int getattrInternal(String path, FileStat stat) {
     final AlluxioURI uri = mPathResolverCache.getUnchecked(path);
     try {
-      URIStatus status = mFileSystem.getStatus(uri);
+      URIStatus status = null;
+      // Handle special metadata cache operation
+      if (mConf.getBoolean(PropertyKey.FUSE_SPECIAL_COMMAND_ENABLED)
+          && mFuseCLientTools.isReservedPath(uri)) {
```
maobaolong, line 235: If the AlluxioFuse mount point mapping to Alluxio no-root path, the `uri` should always starts with the mounted Alluxio path, so you should remove prefix of uri when call `isReservedPath` method

bzheng888: @maobaolong The reserved dir is not only be identified by fuse root /.alluxiocli, but for any path under the the fuse mount point works, for example:
If mount point  mapping to Alluxio no-root path /dir, the uri above will be /dir/.alluxiocli, it is considered to be a reserved path too, so the follow procedure will be executed normally.

maobaolong: @bzheng888 OK

maobaolong (approved):
It LGTM, @LuQQiu Would you like to take a double check?

LuQQiu:
Reorganize the code, @bzheng888 @maobaolong @yyongycy PTAL, thanks!

maobaolong:
@LuQQiu The code after refactor is pretty good, I like the FuseShell concept here, it make this feature can be more stronger in the future.

bzheng888:
@LuQQiu LGTM, thanks for the improvement！

integration/fuse/src/test/java/alluxio/fuse/cli/MockFuseFileSystemMasterClient.java
```diff
@@ -0,0 +1,214 @@
...
+    return false;
+  }
+
+  @Override
+  public boolean isClosed() {
+    return false;
+  }
+
+  @Override
+  public void close() throws IOException {
```
yyongycy, line 212: Just curious, is there any default one proper? If not necessary, can it be removed?

yyongycy:
LGTM thanks for the improvement.

LuQQiu:

PR #14244, opened 2021-10-12
Update alluxio-maven to use go1.12 + update cacerts

Xenorith:
### What changes are proposed in this pull request and why are they needed?

The alluxio-maven image is used to verify PR builds. Currently it is installing go1.11 but the generate tarball go code uses strings.ReplaceAll which is only available starting go1.12.

In addition, some commands that depend on curl/wget are failing due to client certificates being out of date. This is fixed by upgrading the ca-certificates package.

They better not...


Xenorith:

PR #14245, opened 2021-10-13
Allow higher keep-alive frequency for worker data server

ggezer:
RPC channels that are opened to worker are configured to send keep-alive pings every 30sec. This change increases worker server's tolerance to that default.


ggezer:

Issue #14243, opened 2021-10-12
WORKER_NETWORK_ASYNC_CACHE_MANAGER_THREADS_MAX is invalid in some k8s containers


ssz1997:
**Alluxio Version:**
2.6.2

**Describe the bug**
The property key [`WORKER_NETWORK_ASYNC_CACHE_MANAGER_THREADS_MAX`](https://github.com/Alluxio/alluxio/blob/71165a5da66df06425265247f32d631c34692489/core/common/src/main/java/alluxio/conf/PropertyKey.java#L3141-L3149), used at [`GrpcExecutors.java`](https://github.com/Alluxio/alluxio/blob/71165a5da66df06425265247f32d631c34692489/core/common/src/main/java/alluxio/conf/PropertyKey.java#L3141-L3149), is set by `2 * Runtime.getRuntime().availableProcessors()`. In some environments this function always returns `1`, for example, in Fluid.
However, we also require the minimum thread to be at least 4 [here](https://github.com/Alluxio/alluxio/blame/master/core/server/worker/src/main/java/alluxio/worker/grpc/GrpcExecutors.java#L38), so this check would fail. While launching worker, it returns:

2021-10-12 17:50:25,490 ERROR AlluxioWorker - Fatal error: Failed to create worker process
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.ExceptionInInitializerError
	at alluxio.worker.AlluxioWorkerProcess.<init>(AlluxioWorkerProcess.java:162)
	at alluxio.worker.WorkerProcess$Factory.create(WorkerProcess.java:46)
	at alluxio.worker.WorkerProcess$Factory.create(WorkerProcess.java:38)
	at alluxio.worker.AlluxioWorker.main(AlluxioWorker.java:72)
Caused by: java.util.concurrent.ExecutionException: java.lang.ExceptionInInitializerError
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at alluxio.util.CommonUtils.invokeAll(CommonUtils.java:593)
	at alluxio.util.CommonUtils.invokeAll(CommonUtils.java:555)
	at alluxio.worker.AlluxioWorkerProcess.<init>(AlluxioWorkerProcess.java:113)
	... 3 more
Caused by: java.lang.ExceptionInInitializerError
[...]

PR #14248, opened 2021-10-13
Add fall back logic for counting cpu in container is 1

ssz1997:
Check if Runtime.getProcessorCount() returns 1. If so, use default value 8 for property key WORKER_NETWORK_ASYNC_CACHE_MANAGER_THREADS_MAX.

This is a quick fix for #14243.


ggezer:
> > @ssz1997 Could you please address all similar issues in this PR? Also, I think the notation could be like this everywhere: `.setDefaultSupplier(() -> Math.max(8, Runtime.getRuntime().availableProcessors())`
> 
> I was thinking about the same thing, but wasn't sure in the case where the number of cores is 2. Will do this style.

I think there's no end to handling exceptions as cpu count might be 1 as well.

ssz1997:
@ggezer Update all usage of runtime.availableProcessors(), and added some dummy numbers. Let me know what would be the correct, or more suitable, numbers, and I'll update the descriptions accordingly.

PR #14249, opened 2021-10-13
Add probe improvements to Helm chart

ZhuTopher:
This PR introduces the following:

1. Addition of `startupProbe`s
When using the Helm chart with HA Alluxio master pods, sometimes the leader (re-)election can take longer than the `livenessProbe`'s maximum duration causing the pod to be terminated. To combat needing to specify extremely long-running `livenessProbe`s, Kubernetes docs recommend the use of [startupProbes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes). This feature was moved to beta (and so is enabled by default) [in Kubernetes 1.18+](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-graduated-or-deprecated-features). 

2. Use of the Alluxio master Web port for non-embedded HA Alluxio
When using HA Alluxio, not all master pods will be serving the master RPC port (default 19998). Thus it is a poor choice for a `livenessProbe`. We already circumvent this in embedded journal deployments with the use of the embedded journal port (default: 19200), however we were still using the RPC port in non-embedded HA deployments. This PR changes the `livenessProbe` and `readinessProbe` ports to use the master's "web" port in those cases.

3. Addition of `successThreshold` for `readinessProbe`s
This isn't necessarily required for anything, but is nice to be able to have the flexibility to configure. It is only added for `readinessProbe`s since [it is required to be 1](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) for `livenessProbe`s and `startupProbe`s.

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
+++ b/integration/kubernetes/helm-chart/alluxio/templates/master/statefulset.yaml
@@ -129,8 +129,7 @@ spec:
           envFrom:
           - configMapRef:
               name: {{ $fullName }}-config
-{{- $probePort := $isSingleMaster | ternary "rpc" "web" }}
-{{- $probePort = $isHaEmbedded | ternary "embedded" $probePort }}
+{{- $probePort := $isHaEmbedded | ternary "embedded" "rpc" }}
           readinessProbe:
             tcpSocket:
               port: {{ $probePort }}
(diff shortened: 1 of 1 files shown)


ZhuTopher:
@jiacheliu3 I removed the `web` probe definition, lmk if there's anything else (like k8s doc links) you think I should add

PR #14253, opened 2021-10-14
Add a pattern to verify locks are held in MasterWorkerInfo

jiacheliu3:

Add a syntax to double check certain locks are held, within methods that require external locking.
The idea is from HDFS

https://github.com/apache/hadoop/blob/cb2b7970ee190edca3b1a33427e629e32acf0ad7/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/blockmanagement/ProvidedStorageMap.java#L147

It's easier to enforce locking mechanism from methods that need external locking. Without checking, we can only advise people to follow the locking pattern in the comment.


jiacheliu3:
@ggezer PTAL if this pattern makes sense to you. If it does, there are some tests like `MasterWorkerInfoTest` that i need to conform to this patter.

jiacheliu3:
Closed as the lock pattern in MasterWorkerInfo has changed. This PR no longer applies.

Issue #14209, opened 2021-10-08
Formatted master incorrectly accepts blocks created before the format due to blockID(containerID) clash


jiacheliu3:
**Alluxio Version:**
What version of Alluxio are you using?

**Describe the bug**
alluxio master will generate block container id from 0, after master format and restart without worker format, the old worker will report its block to alluxio master, not sure the master will accept the block by accident.

**To Reproduce**
Introduced in
https://github.com/Alluxio/alluxio/pull/14006#issuecomment-913334686

The new master has generated some blocks with container IDs starting from `0`. Then the worker from the previous cluster registers with old blocks (created with the old master, with container ID starting from `0`). If you are unlucky, a block on this old worker may have old blockID `0` and the new blockID `0` has been allocated to a totally irrelevant file. The new master, in this case, will mistakenly think this blockID is recognized and accepts this copy from the old worker. 

Then if the block lengths do not match (the new `0` vs old `0`), the master will throw the error referred in the link.

What's worse, if the block lengths DO match, the master will think this copy belongs to this totally irrelevant file. Then the Alluxio data is messed up without you noticing!

**Expected behavior**

**Urgency**
HIGH

**Additional context**

jiacheliu3:
For the sake of completeness, I'm listing all possible combinations of scenarios, to make sure all the corner cases are scrutinized.

1. Formatted&restarted master + formatted&restarted worker
2. Formatted&restarted master + unformatted&restarted worker
3. Formatted&restarted master + unformatted&unrestarted worker
4. Unformatted&restarted master + formatted&restarted worker
5. Unformatted&restarted master + unformatted&restarted worker
6. Unformatted&restarted master + unformatted&unrestarted worker
7. Unformatted&unrestarted master + formatted&restarted worker
8. Unformatted&unrestarted master + unformatted&restarted worker
9. Unformatted&unrestarted master + unformatted&unrestarted worker

Also don't forget, we have the K8s environment where all paths are ephemeral unless you specify a persistent volume and mount it. On the worker we only have the following volumes. Adding more volumes adds operation overheads.
1. Volumes for tiered storage, as you need
2. Domain socket volume, to be shared with the alluxio client pod (within an application)
3. Log volume if you specifies

If we add a persistent state like `${ALLUXIO_HOME}`, we must not forget the possibility this is overlooked by the admin. We need to define what happens if the file is not copied during a worker update, and how this can be integrated into the existing scripts like `alluxio-start.sh`.

[...]

PR #14258, opened 2021-10-14
Add worker persisted cluster ID

xichen01:

Add persisted clusterID

fixed the issue: https://github.com/Alluxio/alluxio/issues/14209
[DESIGN DOC](https://docs.google.com/document/d/1tCs3fXyI0IzccQSU9HnArlzyEZ-q9bkuPDRub9Nne3c/edit?usp=sharing)

Two main functions in this PR
1. Add a file DB implementation  for worker use to  persist the clusterId of worker

2. Add a preRegistration phase before the worker register to master. The worker will not do anything about register(include getWorkerId with master) until this phase is completed when the worker starts
    In this stage, the worker will report the persisted clusterID to the master. The master will compare it with current 
     clusterId and return command to the worker.
    - if the worker is new, it will report the "INVALID_CLUSTER_ID" to master and the master will return a `Persist` command and the clusterId string. Then the worker will persist the clusterId.
    - if the worker is beginning another cluster , it will report the different clusterId with the current cluster to master and the master will return a `Reset` command and the clusterId string. Then the worker will reset itself(reset == clear all blocks in tier and resetState) and persist the new clusterId (this is todo).
    - the most normal case, the worker will report the same clusterId with the current cluster to master. So master  will return a `Nothing` command, and the worker needn't do anything

usage:
```bash
[root@linux alluxio_home]# bin/alluxio-start.sh local
//...
-----------------------------------------
Starting to monitor all local services.
-----------------------------------------
//...
[root@linux alluxio_home]# cat clusterid  // the file use to persist clusterId
#generated by worker process, don't edit it
#Tue Oct 19 16:25:03 CST 2021
clusterId=d177aeaa-a7cf-4938-8859-3f3e54ffde16

[root@linux alluxio_home]# rm clusterid  // simulate a new worker, worker save cluster in the file "clusterid"
[root@linux alluxio_home]# bin/alluxio-start.sh worker
//...
[root@linux alluxio_home]# cat clusterid 
#generated by worker process, don't edit it
#Tue Oct 19 16:27:31 CST 2021
clusterId=d177aeaa-a7cf-4938-8859-3f3e54ffde16

[root@linux alluxio_home]# vim clusterid // simulate a another cluster worker, worker will reset and save cluster in the file "clusterid"
[root@linux alluxio_home]# cat clusterid 
#generated by worker process, don't edit it
#Tue Oct 19 16:27:31 CST 2021
clusterId=d177aeaa-a7cf-4938-8859-3f3e54ffde17
[root@linux alluxio_home]# bin/alluxio-start.sh worker
//...
[root@linux alluxio_home]# cat clusterid 
#generated by worker process, don't edit it
#Tue Oct 19 16:28:16 CST 2021
clusterId=d177aeaa-a7cf-4938-8859-3f3e54ffde16
[root@linux alluxio_home]# 
```

There are two more features about this "clusterid feature" that will be implemented but may not be included in this PR
- clear all blocks in tier  when worker receive `Reset` command 
- report the clusterId in workerMasterSync heartbaet

no


jiacheliu3:
As you are proposing an RPC before worker registers, this can probably be merged with https://github.com/Alluxio/alluxio/pull/14220?

core/transport/src/main/proto/grpc/meta_master.proto
```diff
@@ -264,6 +272,11 @@ service MetaMasterMasterService {
    */
   rpc GetMasterId(GetMasterIdPRequest) returns (GetMasterIdPResponse);
 
+  /**
+ * Returns a master id for the given master address.
+ */
+  rpc GetClusterIdInternal(GetClusterIdInternalPRequest) returns (GetClusterIdInternalPResponse);
```
maobaolong, line 278: If the server-side is running in old version, and the worker use the version of yours which contains this PR, what will happen? Please list the consistent issue if we upgrade a part of master or worker cluster.

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -233,6 +233,32 @@ message GetWorkerIdPResponse {
   optional int64 workerId = 1;
 }
 
+message PreRegisterCommand {
+  optional PreRegisterCommandType preRegisterCommandType = 1;
+  optional string data = 2;
+}
+
+enum PreRegisterCommandType {
+  Unknown = 0;
```
maobaolong, line 242: We usually use upper case name as the name of enum member.

maobaolong (changes requested):
Before further review, please describe how you handle and keep backward consistent

core/transport/src/main/proto/grpc/meta_master.proto
xichen01: Does Alluxio have requirements for the health status of the cluster during the upgrade process? 
I may need to add compatibility design to the upgrade process

xichen01:
@jiacheliu3 the [DESIGN DOC](https://docs.google.com/document/d/1tCs3fXyI0IzccQSU9HnArlzyEZ-q9bkuPDRub9Nne3c/edit#) has been updated, please take a look.

core/transport/src/main/proto/grpc/block_master.proto
maobaolong: @xichen01 How about this comment?

xichen01:
@jiacheliu3 the WORKER_CLUSTERID_PATH default is ${ALLUXIO_HOME}/persist, this is the path used to save the clusterId file. By default, the worker must have permission to this directory, so that the clusterId can be persisted. If there is no permission, the worker will throw an RuntimeException and exit. This directory (/opt/alluxio/persist/) does not have permission in the community test environment. So many unit tests failed.
Can we add permissions to this directory? Otherwise, I need to set WORKER_CLUSTERID_PATH of all related tests set to the temporary directory.

jiacheliu3 (changes requested):
@xichen01 A high level comment, could you remove the PreRegister and merge the logic into GetWorkerId RPC?

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -168,10 +168,11 @@ message StorageList {
 message BlockHeartbeatPOptions {
   repeated grpc.Metric metrics = 1;
   map<string, int64> capacityBytesOnTiers = 2;
+  optional bool hasClusterId = 3;
```
jiacheliu3, line 171: I dont think you need this field. You can check BlockHeartbeatPRequest.hasClusterId() directly because `clusterId` is an optional field. This method is automatically generated.

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -223,14 +226,22 @@ message CommitBlockInUfsPRequest {
 message CommitBlockInUfsPOptions {}
 message CommitBlockInUfsPResponse {}
 
-message GetWorkerIdPOptions {}
+message GetWorkerIdPOptions {
+  optional bool extendedRegisterInfo = 1; // Identify whether there is preRegister information
```
jiacheliu3, line 230: There's no need for this flag. You can just check `GetWorkerIdRequest.hasClusterId()`

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -223,14 +226,22 @@ message CommitBlockInUfsPRequest {
...
   /** the worker network address */
   optional grpc.WorkerNetAddress workerNetAddress = 1;
+  /** for preRegister information **/
   optional GetWorkerIdPOptions options = 2;
+  optional string clusterId = 3;
+  optional bool hasBlockInTier = 4;
 }
 message GetWorkerIdPResponse {
   optional int64 workerId = 1;
+  optional grpc.WorkerNetAddress workerNetAddress = 2;
```
jiacheliu3, line 242: do you need to return the worker address?

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -223,14 +226,22 @@ message CommitBlockInUfsPRequest {
...
+  optional WorkerPreRegisterInfo workerPreRegisterInfo = 3;
```
jiacheliu3, line 243: You don't need this `WorkerPreRegisterInfo` message. You can just add one `PreRegisterCommandType` in the response.

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -223,14 +226,22 @@ message CommitBlockInUfsPRequest {
...
+  optional GetWorkerIdPOptions options = 4;
```
jiacheliu3, line 244: Do you need the option in the response? The client already has the option when sending the request.

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -246,6 +257,20 @@ message GetRegisterLeasePResponse {
   optional GetRegisterLeasePOptions options = 4;
 }
 
+message WorkerPreRegisterInfo {
+  optional PreRegisterCommandType preRegisterCommandType = 1;
+  optional string clusterId = 2;
+  optional int64 workerId = 3;
+}
```
jiacheliu3, line 264: this message type can be removed

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -223,14 +226,22 @@ message CommitBlockInUfsPRequest {
...
+message GetWorkerIdPOptions {
+  optional bool extendedRegisterInfo = 1; // Identify whether there is preRegister information
+}
 message GetWorkerIdPRequest {
   /** the worker network address */
   optional grpc.WorkerNetAddress workerNetAddress = 1;
+  /** for preRegister information **/
   optional GetWorkerIdPOptions options = 2;
+  optional string clusterId = 3;
+  optional bool hasBlockInTier = 4;
```
jiacheliu3, line 238: instead of a boolean, pass the current block count in the message so later the master can add some extra logic on that

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -246,6 +257,20 @@ message GetRegisterLeasePResponse {
...
+  optional string clusterId = 2;
+  optional int64 workerId = 3;
+}
+
+enum PreRegisterCommandType {
+  UNKNOWN = 0;
+  ACK_REGISTER = 1;
+  REGISTER_PERSIST_CLUSTERID = 2;
+  REGISTER_CLEAN_BLOCKS = 3;
+  REJECT_REGISTER = 4;
```
jiacheliu3, line 271:
```suggestion
  REJECT = 4;
```

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -246,6 +257,20 @@ message GetRegisterLeasePResponse {
...
 }
 
+message WorkerPreRegisterInfo {
+  optional PreRegisterCommandType preRegisterCommandType = 1;
+  optional string clusterId = 2;
+  optional int64 workerId = 3;
+}
+
+enum PreRegisterCommandType {
+  UNKNOWN = 0;
```
jiacheliu3, line 267: what is UNKNOWN for?

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -246,6 +257,20 @@ message GetRegisterLeasePResponse {
...
+  ACK_REGISTER = 1;
+  REGISTER_PERSIST_CLUSTERID = 2;
+  REGISTER_CLEAN_BLOCKS = 3;
```
jiacheliu3, line 270:
```suggestion
  REGISTER_FORMAT_BLOCKS = 3;
```
"format" is better?

core/transport/src/main/proto/grpc/common.proto
```diff
@@ -76,6 +76,7 @@ enum CommandType {
   Free = 3;     // Ask the worker to free files.
   Delete = 4;   // Ask the worker to delete files.
   Persist = 5;  // Ask the worker to persist a file for lineage
+  PreRegisterAndRegister = 6;  // Ask the worker to PreRegister
```
jiacheliu3, line 79: If the "PreRegister" is embedded in the register logic, you don't need this extra command.

core/transport/src/main/proto/grpc/meta_master.proto
```diff
@@ -224,6 +224,14 @@ message GetMasterIdPResponse {
   optional int64 masterId = 1;
 }
 
+message GetClusterIdInternalPOptions {}
+message GetClusterIdInternalPRequest {
+  optional GetClusterIdInternalPOptions options = 1;
+}
+message GetClusterIdInternalPResponse {
+  optional string clusterId = 1;
+}
```
jiacheliu3, line 233: when does the client need to call this? I thought the clusterID is returned in GetWorkerIdPResponse?

jiacheliu3 (changes requested):
Thanks for the changes! I added a batch of comments on the RPC design first. PTAL.

core/transport/src/main/proto/grpc/block_master.proto
xichen01: `UNKNOWN` is a default value. `UNKNOWN` should not appear unless an error occurs
I refer to the heartbeat CommandType
core/transport/src/main/proto/grpc/common.proto
```
enum CommandType {
  Unknown = 0;
  Nothing = 1;
  Register = 2; // Ask the worker to re-register.
 //...
}
```

core/transport/src/main/proto/grpc/common.proto
xichen01: When the ClusterId in the heartbeat is different from the current ClusterId: `PreRegisterAndRegister`

core/transport/src/main/proto/grpc/block_master.proto
xichen01: This is different from `formatWorker` but removes Block and cleans up some related data(metrics, BlockMasterSync.report, ...)

core/transport/src/main/proto/grpc/block_master.proto
xichen01: done

core/transport/src/main/proto/grpc/block_master.proto
xichen01: done

core/transport/src/main/proto/grpc/block_master.proto
xichen01: done

core/transport/src/main/proto/grpc/block_master.proto
xichen01: has Removed

core/transport/src/main/proto/grpc/block_master.proto
xichen01: done

core/transport/src/main/proto/grpc/block_master.proto
xichen01: done

xichen01:
// TODO Integration Tests

core/transport/src/main/proto/grpc/common.proto
jiacheliu3: Since the "pre-register" logic happens in GetWorkerId, I think you should call this command GetWorkerId. I don't want to define the "pre-register" concept here because we may want to use this word for other things.

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
```diff
@@ -125,11 +131,24 @@ public void commitBlockInUfs(CommitBlockInUfsPRequest request,
...
-    }, "getWorkerId", "request=%s", responseObserver, request);
+    if (request.hasClusterId()) {
+      // For compatibility, reuse the RPC of getWorkerID
+      // Determine whether it is preRegistration by the extendedRegisterInfo flag
+      String clusterId = request.getClusterId();
+      int blocksNum = request.getBlocksNum();
+      RpcUtils.call(LOG,
+          (RpcUtils.RpcCallableThrowsIOException<GetWorkerIdPResponse>) () -> {
+            // By the Flag, Let the Worker know that preRegister information is included
+            return mBlockMaster.workerPreRegister(
```
jiacheliu3, line 142: is it possible to merge the logic into `mBlockMaster.getWorkerId()` call instead of having a separate method?

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
```diff
@@ -341,6 +341,7 @@ public void start(Boolean isPrimary) throws IOException {
       } else {
         LOG.info("Detected existing cluster ID {}", mState.getClusterID());
       }
+      mBlockMaster.setClusterId(mState.getClusterID());
```
jiacheliu3, line 344: Instead of MetaMaster setting the clusterID on the BlockMaster, can you make BlockMaster get clusterID from the MetaMaster every time?

core/server/master/src/main/java/alluxio/master/meta/RetryHandlingMetaMasterMasterClient.java
```diff
@@ -82,6 +83,17 @@ public long getId(final Address address) throws IOException {
         .getMasterId(), LOG, "GetId", "address=%s", address);
   }
 
+  /**
+   * Returns current cluster id.
+   *
+   * @return a master id
+   */
+  public String getClusterIdInternal() throws IOException {
```
jiacheliu3, line 91: who will make this call?

core/server/worker/src/main/java/alluxio/worker/block/BlockMasterClient.java
```diff
@@ -269,6 +274,44 @@ public void acquireRegisterLeaseWithBackoff(
...
+    GetWorkerIdPRequest request = GetWorkerIdPRequest.newBuilder()
+        .setClusterId(clusterId).setWorkerNetAddress(GrpcUtils.toProto(address))
+        .setBlocksNum(blocksNum).build();
+
+    return retryRPC(() -> {
+      // For compatibility, reuse the RPC of getWorkerID.
+      GetWorkerIdPResponse response = mClient.getWorkerId(request);
+      // For compatibility,  If preRegister information is included,
+      // the Master will set hasExtendedRegisterInfo
+      if (response.hasClusterId()) {
```
jiacheliu3, line 301: i think you are doing this compatibility check anyway, so you can just merge this perRegisterWithMaster() with the existing getWorkerId()?

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
xichen01: Yes, but this may require changing the return type of `mBlockMaster.getWorkerId()`, Let it return `GetWorkerIdPResponse`

core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java
xichen01: The `MetaMaster` is constructed after the `BlockMaster` is constructed, and the clusterid is generated by the `MetaMaster`.
When `BlockMaster` get  clusterId, it will wait for MetaMaster to `mBlockMaster.setClusterId`

core/server/master/src/main/java/alluxio/master/meta/RetryHandlingMetaMasterMasterClient.java
xichen01: This function is the previous design, I will delete it.

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
```diff
@@ -242,11 +255,86 @@ public void registerCleansUpOrphanedBlocks() throws Exception {
...
 
     // Check that the worker heartbeat tells the worker to remove the blocks.
-    alluxio.grpc.Command heartBeat = mBlockMaster.workerHeartbeat(workerId, null,
+    alluxio.grpc.Command heartBeat = mBlockMaster.workerHeartbeat(workerId, mCurrentClusterId, null,
         memUsage, NO_BLOCKS, NO_BLOCKS_ON_LOCATION, NO_LOST_STORAGE, mMetrics);
     assertEquals(orphanedBlocks, heartBeat.getDataList());
   }
 
+  @Test
+  public void TestNewWorkerRegisterWithMaster() throws IOException {
```
jiacheliu3, line 264: Normally we don't need "test" in the test name. Also we use camel case here. So `newWorkerRegisterWithMaster()` will be a better name.

core/server/worker/src/main/java/alluxio/worker/AlluxioWorkerProcess.java
```diff
@@ -246,7 +247,9 @@ public void start() throws Exception {
     }
 
     // Start serving RPC, this will block
-    LOG.info("Alluxio worker started. id={}, bindHost={}, connectHost={}, rpcPort={}, webPort={}",
+    LOG.info("Alluxio worker started. clusterId={}, workerId={}, bindHost={}, "
+            + "connectHost={}, rpcPort={}, webPort={}",
+        mRegistry.get(BlockWorker.class).getClusterId().get(),
```
jiacheliu3, line 252: Can you `mRegistry.get(BlockWorker.class)` once and use the reference?

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
```diff
@@ -0,0 +1,109 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.UUID;
+
+/**
+ * Persistent info for worker.
+ */
+@NotThreadSafe
+public interface BlockWorkerDB {
```
jiacheliu3, line 34: `WorkerMetaStore`?

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -384,6 +392,78 @@ public void updatePinList() {
...
+
+    // the new cluster ID will be persisted
+    mBlockWorker.handlePreRegisterInfo(response);
+    assertEquals(newClusterId,
+        mBlockWorker.getOrDefaultClusterIdFromDB(IdUtils.EMPTY_CLUSTER_ID).get());
+    assertEquals(mWorkerId1, (long) mBlockWorker.getWorkerId().get());
+  }
+
+  @Test
+  public void handlePreRegisterCommandREGISTER_CLEAN_BLOCKS()
```
jiacheliu3, line 433: I think you can just use camel case here instead of using underscores. So `handleRegisterCleanBlocks()`

jiacheliu3 (commented):
Added some more comments. PTAL and let me know if there's sth you'd like to discuss. Thx :)

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
xichen01: There are already classes with similar names, Will it cause confusion? such as: 
```java
alluxio.master.block.meta.WorkerMetaLockSection
alluxio.master.block.meta.MasterWorkerInfo#lockWorkerMeta
```

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
xichen01: done

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
xichen01: done

core/server/master/src/main/java/alluxio/master/block/BlockMasterWorkerServiceHandler.java
xichen01: @jiacheliu3  do you mean that:

alluxio.master.block.DefaultBlockMaster#getWorkerId
old
``` java
public long getWorkerId(WorkerNetAddress workerNetAddress)
```
new
```java
  public GetWorkerIdPResponse getWorkerId(String workerClusterId,
      WorkerNetAddress workerNetAddress, int blocksNum) throws IOException {
   //....
    PreRegisterCommandType commandType = checkWorker(curClusterId, workerClusterId, blocksNum);
    long workerId = getWorkerIdInternal(workerNetAddress); // the old getWorkerId rename to getWorkerIdInternal
    GetWorkerIdPResponse response = GetWorkerIdPResponse.newBuilder()
        .setPreRegisterCommandType(commandType)
        .setClusterId(curClusterId)
        .setWorkerId(workerId)
        .build();
    //...
    return response;
  }
}
```
the new getworkerid will contain logic for preRegister and old getWorkerId

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
jiacheliu3: I think this is fine. `WorkerMetaStore` suggests it's a storage of worker metadata. `WorkerMeta` suggests it's worker metadata.

@beinan We are storing the clusterID on the worker locally so the worker remembers clusterID on restart. Do you think `WorkerMetaStore` is a fair name? Do you anticipate the worker to store more similar metadata locally? Do you have preference on where the worker should store those metadata (like we are putting them under `${ALLUXIO_HOME}/metadata` now i believe, instead of in the tiered storage)?

beinan: is this class keeping the meta data of the block level meta or worker level meta?   
if it's worker level, I agree with you, I might use alluxio.worker.meta.WorkerMetaStore or alluxio.worker.WorkerMetaStore
If it's block level, it might be more complicated, because @yyongycy and I are both working some block level meta data, I think we could consolidate these metadata somehow

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorkerDB.java
```diff
@@ -0,0 +1,145 @@
...
+import java.nio.file.Paths;
+import java.util.Properties;
+
+/**
+ * persistence worker info to a file.
+ */
+@NotThreadSafe
+public class DefaultBlockWorkerDB implements BlockWorkerDB {
+  private static final String CLUSTER_ID_KEY = "clusterId";
+  private static final Logger LOG = LoggerFactory.getLogger(MetricsConfig.class);
```
beinan, line 43:
```suggestion
  private static final Logger LOG = LoggerFactory.getLogger(DefaultBlockWorkerDB .class);
```

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorkerDB.java
```diff
@@ -0,0 +1,145 @@
...
+  /*Use Properties to persistence info */
+  final Properties mProperties;
```
beinan, line 45: It's ok to keep the metadata as a property file just for now,  I might prefer a protobuf or json file if we will have more data entries in the future.

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorkerDB.java
```diff
@@ -0,0 +1,145 @@
...
+  /*Properties file path */
+  final String mFile;
+  boolean mIsInit;
+
+  DefaultBlockWorkerDB(String path) throws IOException {
+    mProperties = new Properties();
+    mFile = path;
+    mIsInit = false;
```
beinan, line 53: I don't think mIsInit is necessary, we could safely remove it

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
beinan: Ok, I see, it's a worker level metadata, so I think WorkerMetaStore is a decent name.  But I think it's a little bit over kill to make a metadata store just for a cluster_id, we could just load the cluster_id from a local file on its name e.g.   /worker/cluster_id/000765  (the file is empty, we could get everything we need from the filename).

xichen01:
@jiacheliu3 I have merged preRegister into getWokreID and changed the signature of getWokreID

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
xichen01: @beinan I don't think there is a big difference between using the file name and the file content, because the IO of this file only happens when the worker starts and registers, which is an infrequent operation.
Let the user know the function of the file through the file name and some comments

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorkerDB.java
xichen01: This is old code, I have removed it

core/server/worker/src/main/java/alluxio/worker/block/DefaultBlockWorkerDB.java
xichen01: done

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
beinan: @xichen01 I'm not talking about the performance, I'm just thinking if it's overkilled by adding a couple of classes just for reading a cluster id from the disk.

beinan:
Hi @xichen01 & @jiacheliu3,
@LuQQiu and I have some discussion offline, we think this pr might help for cleaning the staled blocks to prevent the storage leak also.     So both of us propose to consider the worker's format op also, because the worker's format might be interrupted or failed to delete some of the files(see https://github.com/Alluxio/alluxio/issues/15171).  We just have a rough idea,  we could put the last format timestamp in the worker's storage path, something like
{master_last_format_timestamp}/{worker_last_format_timestamp}/block_files

master_last_format_timestamp is identical to your cluster_id, then it would be much easier to clean the staled or dirty block files

xichen01:
@beinan If the leader of the raft-HA cluster changes, will the new master have the same {master_last_format_timestamp} as the old master?

beinan:
> @beinan If the leader of the raft-HA cluster changes, will the new master have the same {master_last_format_timestamp} as the old master?

Great question!  but I'm afraid I don't know. @jiacheliu3 do you have any ideas?

LuQQiu:
> > @beinan If the leader of the raft-HA cluster changes, will the new master have the same {master_last_format_timestamp} as the old master?
> 
> Great question! but I'm afraid I don't know. @jiacheliu3 do you have any ideas?

Can we use the first leader master up time, and master_last_format_time can be one of the earliest journal entry?

core/server/worker/src/main/java/alluxio/worker/block/BlockWorkerDB.java
jiacheliu3: Yea if we have a complicated design here, that'd be an overkill because it's a design without real use cases.

For the moment I think it's fine. The `BlockWorkerDB`(to be renamed to sth else) is not too much code and the usage is self-explanatory. It's wrapping a few file operations. I will actually recommend we remove this interface design here. You don't need an interface here. It's fine to just have a concrete class `BlockWorkerMetaStore` which handles reading and writing cluster ID from a file.

xichen01: Storing the metadata(current is ClusterId) in a file makes it easier for workers to store more metadata, The current interface is relatively simple, I think there is an interface that may be more extensible.

jiacheliu3:
> Hi @xichen01 & @jiacheliu3, @LuQQiu and I have some discussion offline, we think this PR might help for cleaning the staled blocks to prevent the storage leak also. So both of us propose to consider the worker's format op also, because the worker's format might be interrupted or failed to delete some of the files(see #15171). We just have a rough idea, we could put the timestamp of last format in the worker's storage path, something like {master_last_format_timestamp}/{worker_last_format_timestamp}/block_files
> 
> master_last_format_timestamp is identical to your cluster_id, then it would be much easier to clean the staled or dirty block files

@beinan I see your point. I think handling the format time (at master and worker and probably consider HA master too) will add too many states and corner cases for this single PR to handle. If you have a github issue or WIP PR feel free to link to this PR, we'd like to stay tuned on that idea as well.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1005,6 +1012,31 @@ public long getWorkerId(WorkerNetAddress workerNetAddress) {
     return workerId;
   }
 
+  @Override
+  public GetWorkerIdPResponse getWorkerId(WorkerNetAddress workerNetAddress,
+      String workerClusterId, int blocksNum) throws IOException {
+    // todo The return type may be the Alluxio class(such as WorkerMetaInfo)
+    //  instead of the GRPC class ?
```
jiacheliu3, line 1019: Do you intend to use the `WorkerMetaInfo` for something?

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1005,6 +1012,31 @@ public long getWorkerId(WorkerNetAddress workerNetAddress) {
...
+    GetWorkerIdPResponse response = GetWorkerIdPResponse.newBuilder()
+        .setRegisterCommandType(commandType)
+        .setClusterId(curClusterId)
+        .setWorkerId(workerId)
+        .build();
+
+    LOG.info("worker {} ,Register result: RegisterCommand {}, workerId {},"
+            + " current clusterId {}",
+        workerNetAddress.getHost(), response.getRegisterCommandType(),
+        response.getWorkerId(), response.getClusterId());
```
jiacheliu3, line 1036:
```suggestion
    LOG.info("Allocated ID {} to worker {} with command {}",
        response.getWorkerId(), 
        workerNetAddress.getHost(), 
        response.getRegisterCommandType());
```
And you can add another log in your `checkWorker()` like when worker and master cluster ID don't match:
```
LOG.warn("Worker {} has a different clusterID {} from master {}. The worker may be outdated and the master command is {}, workerHostname, workerClusterId, masterClusterId, commandType);
```

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1005,6 +1012,31 @@ public long getWorkerId(WorkerNetAddress workerNetAddress) {
     return workerId;
   }
 
+  @Override
+  public GetWorkerIdPResponse getWorkerId(WorkerNetAddress workerNetAddress,
+      String workerClusterId, int blocksNum) throws IOException {
+    // todo The return type may be the Alluxio class(such as WorkerMetaInfo)
+    //  instead of the GRPC class ?
+    LOG.info("worker {} Register, workerClusterId {}, blocksNum {}",
+        workerNetAddress.getHost(), workerClusterId, blocksNum);
```
jiacheliu3, line 1021: This log line can be removed

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1020,6 +1052,87 @@ public void releaseRegisterLease(long workerId) {
...
+  public String getClusterId() {
+    try {
+      // ClusterId is generated by MetaMaster,
+      // MetaMaster is constructed later than BlockMaster,
+      // so wait for MetaMaster to generate and set ClusterId.
+      CommonUtils.waitFor(this + " getClusterId",
+          () -> !mClusterId.get().equals(IdUtils.INVALID_CLUSTER_ID),
+          WaitForOptions.defaults().setTimeoutMs(5000));
+    } catch (Exception e) {
+      LOG.error("getClusterId {}: ", IdUtils.INVALID_CLUSTER_ID, e);
```
jiacheliu3, line 1065: When we cannot get a cluster ID, I think the master logic cannot continue and we should exit with a RuntimeException

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1020,6 +1052,87 @@ public void releaseRegisterLease(long workerId) {
     mRegisterLeaseManager.releaseLease(workerId);
   }
 
+  @Override
+  public String getClusterId() {
+    try {
+      // ClusterId is generated by MetaMaster,
+      // MetaMaster is constructed later than BlockMaster,
+      // so wait for MetaMaster to generate and set ClusterId.
+      CommonUtils.waitFor(this + " getClusterId",
```
jiacheliu3, line 1061: `this + "getClusterId"` will concat strings every time this is called. So either put a constant string here or use String.format().

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -1020,6 +1052,87 @@ public void releaseRegisterLease(long workerId) {
...
+          //   3. The clusterId file is deleted.
+          //   4. The worker starts for the first time after the upgrade (from
+          //      version “without persist clusterId” upgrade to version “persist clusterId” )
+          //
+          //   In the case 1 ~ 3 the worker will get the default clusterId IdUtils.EMPTY_CLUSTER_ID
+          //   We cannot distinguish whether the worker belongs to the current cluster.
+          //   Although the worker may contain dirty data. But it is an undefined behavior, the
+          //   clusterId file should be be persisted correctly.
+          //   All of these cases are due to clusterId not being persisted correctly the branch
+          //   should not be entered except the case 4.
```
jiacheliu3, line 1108: This is a great question. I personally prefer deleting these blocks because losing cache is better than having incorrect blocks. WDYT @beinan ?

core/server/master/src/test/java/alluxio/master/block/BlockMasterTest.java
```diff
@@ -106,10 +116,13 @@ public void before() throws Exception {
     mClock = new ManualClock();
     mExecutorService =
         Executors.newFixedThreadPool(2, ThreadFactoryUtils.build("TestBlockMaster-%d", true));
-    mBlockMaster = new DefaultBlockMaster(mMetricsMaster, masterContext, mClock,
-        ExecutorServiceFactories.constantExecutorServiceFactory(mExecutorService));
+    mBlockMaster = spy(new DefaultBlockMaster(mMetricsMaster, masterContext, mClock,
```
jiacheliu3, line 119: You can write an easy extension of `DefaultBlockMaster` to override `getClusterId()` logic. We prefer that to using mocks because it's much easier to maintain. For example if the method signature changes, mocks will fail (change behavior) silently but children classes will complain.

core/server/master/src/test/java/alluxio/master/block/ConcurrentBlockMasterTest.java
```diff
@@ -118,8 +122,10 @@ public void before() throws Exception {
         ThreadFactoryUtils.build("TestBlockMaster-%d", true));
     // No one is listening to this latch, so do any preparation for the test
     CountDownLatch voidLatch = new CountDownLatch(1);
-    mBlockMaster = new SignalBlockMaster(mMetricsMaster, mMasterContext, mClock,
-        ExecutorServiceFactories.constantExecutorServiceFactory(mExecutorService), voidLatch);
+    mBlockMaster = spy(new SignalBlockMaster(mMetricsMaster, mMasterContext, mClock,
```
jiacheliu3, line 125: same comment on mock vs extension

core/server/worker/src/main/java/alluxio/worker/AlluxioWorkerProcess.java
```diff
@@ -311,6 +314,13 @@ public boolean waitForReady(int timeoutMs) {
           () -> isServing() && mRegistry.get(BlockWorker.class).getWorkerId() != null
               && mWebServer != null && mWebServer.getServer().isRunning(),
           WaitForOptions.defaults().setTimeoutMs(timeoutMs));
+      if (mRegistry.get(BlockWorker.class).getClusterId().get()
+          .equals(IdUtils.EMPTY_CLUSTER_ID)) {
+        LOG.warn("BlockWorker clsuter ID: {} "
+                + "This may be caused by the worker process not be able to persist the clusterid"
+                + " to the file {}", IdUtils.EMPTY_CLUSTER_ID,
+            ServerConfiguration.get(PropertyKey.WORKER_METASTORE_PATH));
```
jiacheliu3, line 322:
```suggestion
        LOG.warn("Worker has an empty cluster ID! "
                + "Please check if the worker has permission to persist cluster ID in {}/clusterID", 
            ServerConfiguration.get(PropertyKey.WORKER_METASTORE_PATH));
```

core/server/worker/src/main/java/alluxio/worker/block/BlockHeartbeatReporter.java
```diff
@@ -67,13 +67,23 @@ public BlockHeartbeatReport generateReport() {
...
+    synchronized (mLock) {
+      ClearReportInternal();
+    }
+  }
+
+  private void ClearReportInternal() {
+    mAddedBlocks.clear();
+    mRemovedBlocks.clear();
+    mLostStorage.clear();
+  }
```
jiacheliu3, line 85: good catch :)
But if it's not necessary in this PR, let's fix in a separate PR

core/server/worker/src/main/java/alluxio/worker/block/DefaultWorkerMetaStore.java
```diff
@@ -0,0 +1,128 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Properties;
+
+/**
+ * persistence worker info to a file.
+ */
+@NotThreadSafe
+public class DefaultWorkerMetaStore implements WorkerMetaStore {
```
jiacheliu3, line 40: We don't need an interface and default implementation. You can make `WorkerMetaStore` the one class that has the logic. If the Factory is not necessary, we can remove that too.

I admit that in our existing code, we have a lot of interfaces + 1 default implementation. That's not too necessary and we don't have to follow that pattern. But I appreciate the try.

core/server/worker/src/test/java/alluxio/worker/block/DefaultBlockWorkerTest.java
```diff
@@ -105,8 +112,9 @@ public void before() throws IOException {
     mSessions = mock(Sessions.class);
     mUfsManager = mock(UfsManager.class);
 
-    mBlockWorker = new DefaultBlockWorker(mBlockMasterClientPool, mFileSystemMasterClient,
-        mSessions, mBlockStore, mUfsManager);
+    setPersistenceConf();
+    mBlockWorker = spy(new DefaultBlockWorker(mBlockMasterClientPool, mFileSystemMasterClient,
```
jiacheliu3, line 116: same comment on mock vs extension

core/server/worker/src/test/java/alluxio/worker/block/NoopBlockWorker.java
```diff
@@ -48,6 +48,11 @@
     return null;
   }
 
+  @Override
+  public AtomicReference<String> getClusterId() {
+    return null;
```
jiacheliu3, line 53: It doesn't matter very much here in the test, but generally speaking returning null is not the rule of thumb. Instead return something like `new AtomicReference("");`

core/transport/src/main/proto/grpc/block_master.proto
```diff
@@ -171,7 +171,7 @@ message BlockHeartbeatPOptions {
 }
 
 message LocationBlockIdListEntry {
-  optional BlockStoreLocationProto key= 1;
+  optional BlockStoreLocationProto key = 1;
```
jiacheliu3, line 174: lol

jiacheliu3 (changes requested):
Just had a quick pass. Thanks for the improvements and thorough tests. I appreciate it!

Left some comments, and I will add some more in later passes in the following few days.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
beinan: My concern is case 4, some of the users might be not happy to lose all the blocks they cached when they perform the upgrade.

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
xichen01: Maybe use `WorkerMetaInfo` to store some metadata of the worker, so that It does not need to return a GRPC Type. This reduces the coupling of the code

core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
xichen01: yes, In the current design, if it is case4, the user will not lose any data, and the upgrade can be completed normally

core/server/worker/src/main/java/alluxio/worker/block/DefaultWorkerMetaStore.java
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * persist a key-value to property file.
+   *
+   * @param key   the key to be persist into this file
+   * @param value the value corresponding to key
+   * @throws IOException I/O error if create or write file failed
+   */
+  @Override
+  public void set(String key, String value) throws IOException {
```
jiacheliu3, line 105: I think we don't need a general `get()` and `set()` KV store. Let's start simple and just define `setClusterId()` and `getClusterId()`. If in the future we see more need, more methods will be added to this class, and then we can choose to abstract an interface.

core/server/master/src/test/java/alluxio/master/block/ConcurrentBlockMasterTest.java
```diff
@@ -417,7 +423,7 @@ public void concurrentCommitWithSameWorkerHeartbeatSameBlock() throws Exception
           // The same block is removed on worker in this heartbeat
           // This should succeed as commit locks the block exclusively and finishes first
           // When the block heartbeat processes the same block, it has been committed
-          Command cmd = mBlockMaster.workerHeartbeat(worker1,
+          Command cmd = mBlockMaster.workerHeartbeat(worker1, CLUSTER_ID,
```
jiacheliu3, line 426: Should we consider concurrent tests? Like when the worker registers with master, there are concurrent read/write requests? The concurrency happens in two sides:
1. There are concurrent `removeBlock(blockId, worker)` requests on the BlockMaster with worker register
2. There are concurrent `writeBlock()` requests on the BlockWorker with worker register.

I think there can be too many combinations of events so we can never enumerate them all. But we can start with reasoning what operations can be concurrent and if there could be race conditions. If we identify a race condition, try to translate that into a unit test.

jiacheliu3 (changes requested):
Added some more comments. Thanks!

jiacheliu3:
@xichen01 I see some outstanding comments and some conflicts in this PR, is this still actively worked on?

PR #14263, opened 2021-10-14
Fix flaky test:BlockMasterJournalIntegrationTest

ggezer:
This test fails sometimes because validation happens before cluster is out of safe mode.


ggezer:

PR #14265, opened 2021-10-15
Fix resource leak due to Files.walk

lujiefsi:
Files.walk will open stream, we should close it.

 

see jdk: 
the {try} -with-resources construct should be used to ensure that the
stream's

{[@link|https://github.com/link] Stream#close close}
method is invoked after the stream
operations are completed.


yuzhu:
@lujiefsi thanks for your contribution. Could you sign the contributor agreement? and also the title of the PR needs to follow the convention. Thanks!

PR #14267, opened 2021-10-18
[DOCFIX] Fix navtab names for CephFS.md

Xenorith:
Fixes the navtabs in CephFS pages and adds static checks for navtabs to:
- prevent the use of invalid characters in the future
- prevent duplicate names


Xenorith:
@sg-c thanks for catching this

PR #14270, opened 2021-10-18
Make the time unit consistent

codings-dan:
Use uniform time unit

sec,second -> s
min,minute -> m
hour,hr -> h
day -> d


dbw9580:
Can you please elaborate on why this change is necessary? `s` and `sec`, etc. are all supported suffixes for time. See `alluxio.util.FormatUtils#parseTimeSize`

jiacheliu3:
Thanks for the attempt. Honestly I prefer "sec", "min", "hr" to just one letter because that's less ambiguous. Do you have some other references why "s" and "h" should be the preferred way? We can further discuss.

I agree we should be consistent on the time unit, just a bit not sure which option is better.

codings-dan:
@jiacheliu3 Using s and h to represent time may be a more general solution, we can refer to https://github.com/airlift/units/blob/master/src/main/java/io/airlift/units/Duration.java#L189

maobaolong:
@codings-dan Thanks for your reply, can you list what famous softwares are using this dependency?

codings-dan:
@maobaolong As far as I know, Presto uses this library.

maobaolong:
@codings-dan I have no objection for this PR, LGTM, thanks.

jiacheliu3:
> @jiacheliu3 Using s and h to represent time may be a more general solution, we can refer to https://github.com/airlift/units/blob/master/src/main/java/io/airlift/units/Duration.java#L189

I don't really think `airlift/units` is a common dependency people confirm to. Are there any other arguements?

PR #14271, opened 2021-10-18
Replace guava's Files.createTempDir()

lujiefsi:
see https://www.cvedetails.com/cve/CVE-2020-8908/

A temp directory creation vulnerability exists in all versions of Guava, allowing an attacker with access to the machine to potentially access data in a temporary directory created by the Guava API com.google.common.io.Files.createTempDir(). By default, on unix-like systems, the created directory is world-readable (readable by an attacker with access to the system). The method in question has been marked @deprecated in versions 30.0 and later and should not be used. For Android developers, we recommend choosing a temporary directory API provided by Android, such as context.getCacheDir(). For other Java developers, we recommend migrating to the Java 7 API java.nio.file.Files.createTempDirectory() which explicitly configures permissions of 700, or configuring the Java runtime's java.io.tmpdir system property to point to a location whose permissions are appropriately configured.

### DIFF
--- a/core/common/src/main/java/alluxio/util/io/FileUtils.java
+++ b/core/common/src/main/java/alluxio/util/io/FileUtils.java
@@ -344,6 +344,18 @@ public final class FileUtils {
   }
 
   /**
+   * Creates a temporary directory like guava's Files.createTempDir().
+   *
+   * @return the created temporary directory
+   */
+  public static File createTempDir() throws IOException {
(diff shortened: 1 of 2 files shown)


lujiefsi:
> @lujiefsi any reason why we don't use ` java.nio.file.Files.createTempDirectory()`

hi
this patch uses **java.nio.file.Files.createTempDirectory()**

PR #14277, opened 2021-10-19
Fix worker cannot report worker cluster aggregated metrics issue

maobaolong:
### What changes are proposed in this pull request?

Worker cannot report worker related cluster aggregated metrics

### Why are the changes needed?

With this fix, worker can report both client and worker related cluster aggregated metrics to master side.

### Does this PR introduce any user facing changes?

No

PR #14280, opened 2021-10-19
Add worker and client metrics size to log info

bzheng888:
Add helpful information, this is inspired by pr #14277
 Developers can get metrics debug information directly in log.
No


maobaolong:
@LuQQiu Would you please take a look at this PR, with this PR, we can find the root cause of `worker never report metric to master` issue easily.

PR #14292, opened 2021-10-22
Set the range in the GetObjectRequest when reading S3

beinan:

Set the range in the GetObjectRequest when reading S3

This change would reduce the number of bytes alluxio read from S3

No

### DIFF
--- a/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
+++ b/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
@@ -29,8 +29,6 @@ import javax.annotation.concurrent.NotThreadSafe;
  */
 @NotThreadSafe
 public class S3AInputStream extends InputStream {
-  private static final Logger LOG = LoggerFactory.getLogger(S3AInputStream.class);
-
   /** Client for operations with s3. */
   protected AmazonS3 mClient;
   /** Name of the bucket the object resides in. */
@@ -144,12 +142,16 @@ public class S3AInputStream extends InputStream {
       return;
     }
     if (mIn != null && mPos > mInPos) { // stream is already open but the caller is seeking forward
-      long skip = mPos - mInPos;
-      if (skip < mIn.available()) {
+      if (mPos - mInPos < mIn.available()) {
         try {
-          //TODO(beinan): we might need check the max skip length
-          if (skip == mIn.skip(skip)) {
-            mInPos += skip;
+          while (mPos - mInPos > 0) {
+            long actualSkipped = mIn.skip(mPos - mInPos);
+            if (actualSkipped <= 0) {
+              break;
+            }
+            mInPos += actualSkipped;
+          }
+          if (mPos == mInPos) {
             return; //able to reuse the existing stream
           }
         } catch (IOException e) {


underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -125,29 +133,54 @@ public long skip(long n) throws IOException {
...
    */
-  private void openStream() throws IOException {
-    if (mIn != null) { // stream is already open
+  private void seekStream(int length) throws IOException {
+    if (mIn != null && mPos == mInPos) { // stream is already open and at the expected position
       return;
     }
+    if (mIn != null && mPos > mInPos) { // stream is already open but the caller is seeking forward
+      long skip = mPos - mInPos;
+      if (skip < mIn.available()) {
```
ZacBlanco, line 150: IIRC, `available` is not supported on all input streams, and [may just return 0](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#available--). Do you know if this method is implemented properly for the underlying stream?

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -90,12 +92,11 @@ public void close() {
 
   @Override
   public int read() throws IOException {
-    if (mIn == null) {
-      openStream();
-    }
+    seekStream(1);
```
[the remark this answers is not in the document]
beinan: we need seek because the pos might changed by the skip() call

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -125,29 +125,52 @@ public long skip(long n) throws IOException {
...
+          if (skip == mIn.skip(skip)) {
+            mInPos += skip;
+            return; //able to reuse the existing stream
+          }
+        } catch (IOException e) {
+          //ignore to create a new stream
+        }
+      }
+    }
+    closeStream();
```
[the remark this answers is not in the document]
beinan: good call, not really, mostly the stream would be reused if there are still some bytes available (line 155)

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
beinan: Good question!  I believed so, I will double check and get back to you later.

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -125,29 +125,52 @@ public long skip(long n) throws IOException {
...
+  private void seekStream(int length) throws IOException {
+    if (mIn != null && mPos == mInPos) { // stream is already open and at the expected position
       return;
     }
+    if (mIn != null && mPos > mInPos) { // stream is already open but the caller is seeking forward
+      long skip = mPos - mInPos;
+      if (skip < mIn.available()) {
+        try {
+          //TODO(beinan): we might need check the max skip length
+          if (skip == mIn.skip(skip)) {
```
ZacBlanco, line 153: in the case skip returns a value which is not equal to `skip` (e.g. < skip), but not negative, the stream may still have more to available to skip. Should we try calling this in a loop? 

ref: https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#skip-long-

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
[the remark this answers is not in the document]
ZacBlanco: I see -- above when we check that we can just skip forward there is an early return. I missed that before.

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
jja725: the same question, we need to do something like the following.
```
do {
      skip(pos - mPos);
    } while (pos > mPos && in.available() > 0);
```

jja725 (commented):
we also have https://github.com/Alluxio/alluxio/pull/16631 to support seek in s3. Better merge these two.

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
beinan: @jia725 loop added

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -123,30 +129,59 @@ public long skip(long n) throws IOException {
...
 
   /**
-   * Opens a new stream at mPos if the wrapped stream mIn is null.
+   * Seek if we could reuse the existing stream or create a new one.
+   * @param length the content length to be read
    */
-  private void openStream() throws IOException {
-    if (mIn != null) { // stream is already open
+  private void seekStream(int length) throws IOException {
+    if (mIn != null && mPos == mInPos) { // stream is already open and at the expected position
```
LuQQiu, line 141: when mPos == mInPos == mInPosEnd, when reading length > 0, then 
mIn.read() may out of range?

LuQQiu, line 141: when mPos == mInPos == mInPosEnd, when reading length > 1, then
mIn.read() may out of range?

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -123,30 +129,59 @@ public long skip(long n) throws IOException {
...
       return;
     }
+    if (mIn != null && mPos > mInPos) { // stream is already open but the caller is seeking forward
+      if (mPos - mInPos < mIn.available()) {
```
LuQQiu, line 145: we already recorded `mInPosEnd` so no need of checking mIn.available() which may not be correct?

I expect if read is out of range, then getObject will already fail?
as long as mInPos< mPos < mInPosEnd, it should exist?

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AInputStream.java
```diff
@@ -123,30 +129,59 @@ public long skip(long n) throws IOException {
...
+        try {
+          while (mPos - mInPos > 0) {
+            long actualSkipped = mIn.skip(mPos - mInPos);
```
LuQQiu, line 148: same here, if we setRange(0, 10), and new Pos is 20, are we able to skip to 20 given that we only set range up to 10?

LuQQiu (commented):
Left some comments, need a unit test to verify all the tricky cases

LuQQiu:
The API doc mentions that
```
Be extremely careful when using this method; the returned Amazon S3 object contains a direct stream of data from the HTTP connection. The underlying HTTP connection cannot be reused until the user finishes reading the data and closes the stream. Also note that if not all data is read from the stream then the SDK will abort the underlying connection, this may have a negative impact on performance. Therefore:
Use the data from the input stream in Amazon S3 object as soon as possible
Read all data from the stream (use GetObjectRequest.setRange(long, long) to request only the bytes you need)
Close the input stream in Amazon S3 object as soon as possible
If these rules are not followed, the client can run out of resources by allocating too many open, but unused, HTTP connections.

public S3Object getObject(String bucketName, String key) throws SdkClientException,
            AmazonServiceException;

```

beinan:
> @beinan probably can refactor the code in the following way
> 
> ```
>   @Override
>   public int read() throws IOException {
>     byte[] buf = new byte[1];
>     if (read(buf) == -1) {
>       return -1;
>     }
>     return buf[0];
>   }
> 
>   @Override
>   public int read(byte[] b) throws IOException {
>     return read(b, 0, b.length);
>   }
> 
>   @Override
>   public int read(byte[] b, int offset, int length) throws IOException {
>     GetObjectRequest getReq = new GetObjectRequest(mBucketName, mKey);
>     getReq.setRange(mPos, mPos + length);
>     S3ObjectInputStream in = getClient().getObject(getReq).getObjectContent();
>     int ret = mIn.read(b, offset, length);
>     mIn.close();
>     if (ret == -1) {
>       return ret;
>     }
>     mPos += ret;
>     return ret;
>   }
> ```
> 
> and using BufferedInputStream to provide buffer ability for too small read operations(e.g. those from FUSE, or keep doing read() for read 1 bytes), and we can find the best buffer size for application by testing. Do you want me to take over this PR?

Sorry for my delay, Yes ,really appreciate it if you could take over this pr!

PR #14302, opened 2021-10-25
Enable collecting metrics from user

LuQQiu:

Enable collect metrics from user

so that user knows that they enabled short-circuit read from the related metrics.

Web UI will show the correct user metrics

### DIFF
--- a/shell/src/test/java/alluxio/cli/LogLevelTest.java
+++ b/shell/src/test/java/alluxio/cli/LogLevelTest.java
@@ -25,6 +25,8 @@ import alluxio.conf.InstancedConfiguration;
 import alluxio.conf.PropertyKey;
 import alluxio.job.wire.JobWorkerHealth;
 import alluxio.master.MasterInquireClient;
+import alluxio.uri.MultiMasterAuthority;
+import alluxio.uri.ZookeeperAuthority;
 import alluxio.wire.WorkerNetAddress;
 
 import org.apache.commons.cli.CommandLine;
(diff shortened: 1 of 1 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -4411,7 +4411,7 @@ public String toString() {
           .build();
   public static final PropertyKey USER_METRICS_COLLECTION_ENABLED =
       new Builder(Name.USER_METRICS_COLLECTION_ENABLED)
-          .setDefaultValue(false)
+          .setDefaultValue(true)
```
maobaolong, line 4414: This change will make the default behavior changed, the user report metric will harm for performance, do you really need to do this change while users do not aware this change?

maobaolong (commented):
Left a question inline.

core/common/src/main/java/alluxio/conf/PropertyKey.java
LuQQiu: @apc999 @yuzhu

yuzhu: so far we have not seen much perf degradation (maybe 5-10%)? do you have specific instances where this makes perf much worse? @maobaolong

yuzhu: one reason we want to make this default true, is users get confused why there is no local read metric when they just use it out of the box. we want to make the default experience better.

PR #14324, opened 2021-10-28
Support use default block size to replace the Hdfs ufs

maobaolong:

Now the cephfs, ozone and cosn are all implemented based on hdfs ufs, but we don‘t want alluxio use the ufs block size, for example, our block size of ceph is 4MB, which is too small for alluxio, we should supply a way to customize hdfs-like ufs block size for alluxio. 

Add a propertyKey to enable the user default block size.

set `alluxio.user.block.size.enabled` to true to enable using the user block size for cephfs, ozone, hdfs, cosn.

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3931,7 +3931,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey USER_BLOCK_SIZE_OVERRIDE_UFS_ENABLED =
       new Builder(Name.USER_BLOCK_SIZE_ENABLED)
           .setDefaultValue(false)
-          .setDescription("Use Default block size for Alluxio files.")
+          .setDescription("When reading a file from UFS, whether Alluxio should ignore the UFS block size and rely on the Alluxio configuration.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.ALL)
           .build();


underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
```diff
@@ -395,8 +395,12 @@ public UfsFileStatus getFileStatus(String path) throws IOException {
     FileStatus fs = hdfs.getFileStatus(tPath);
     String contentHash =
         UnderFileSystemUtils.approximateContentHash(fs.getLen(), fs.getModificationTime());
+    long blockSize = fs.getBlockSize();
+    if (mUfsConf.getBoolean(PropertyKey.USER_BLOCK_SIZE_ENABLED)) {
```
ggezer, line 399: With this flag enabled, `getFileStatus()` will start returning wrong block size. So I don't think this the right place to patch up. 

What's the exact reason you want to override UFS block-size with Alluxio block-size?

ggezer (changes requested):

underfs/hdfs/src/main/java/alluxio/underfs/hdfs/HdfsUnderFileSystem.java
maobaolong: image

I just reference the ObjectUnderFileSystem#getFileStatus method.

The exact reason for customize the block size is that in our alluxio cluster, cephfs is our underfilesystem, the blockSize from cephfs is 4MB, so a 1 GB big file will cost 1GB/4MB blocks, i want to customize the alluxio block size, for example 64MB, it would be better to reduce the block count.

maobaolong:
@ggezer Thanks for the previous review, I move the customize block size logic to high level. PTAL.

jiacheliu3:
fyi @HelloHorizon @beinan

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -880,6 +880,9 @@ static void loadFileMetadataInternal(RpcContext rpcContext, LockedInodePath inod
       }
       ufsLength = ((UfsFileStatus) context.getUfsStatus()).getContentLength();
       long blockSize = ((UfsFileStatus) context.getUfsStatus()).getBlockSize();
+      if (ServerConfiguration.getBoolean(PropertyKey.USER_BLOCK_SIZE_ENABLED)) {
+        blockSize = ServerConfiguration.getBytes(PropertyKey.USER_BLOCK_SIZE_BYTES_DEFAULT);
```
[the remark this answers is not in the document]
beinan: +1

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -880,6 +880,9 @@ static void loadFileMetadataInternal(RpcContext rpcContext, LockedInodePath inod
       }
       ufsLength = ((UfsFileStatus) context.getUfsStatus()).getContentLength();
       long blockSize = ((UfsFileStatus) context.getUfsStatus()).getBlockSize();
+      if (ServerConfiguration.getBoolean(PropertyKey.USER_BLOCK_SIZE_ENABLED)) {
```
jiacheliu3, line 883: I think this only reads the server-side default, so it doesn't take your path config or configuration from your client command. This defeats the purpose of having this property.

@beinan @yuzhu Do you know how to make this respect both the client-side config and path config? I'm afraid making this respect user-side config needs this property value to be in `message FileSystemCommonPOptions`. Getting the path config is easier than modifying RPC, but I forgot how :(

jiacheliu3 (changes requested):
Thanks for the request! This setting makes sense to me. But it needs to respect the user-side config and path config in order to be really useful.

jiacheliu3:
After a few discussions on this feature, we decided to still enable users to configure this per path, because at first it's hard to set the block size correctly without need to override in the future.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
jiacheliu3: In order to pick up the correct path-specific setting, we will need to resolve the config for the path. The path configs are returned from master https://github.com/Alluxio/alluxio/blob/ef6b7efffb2e18246b7fae4b73551db73c8b7943/core/server/master/src/main/java/alluxio/master/meta/DefaultMetaMaster.java#L430
to the client
https://github.com/Alluxio/alluxio/blob/ef6b7efffb2e18246b7fae4b73551db73c8b7943/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java#L433
Then the client will resolve the path config. This is how you set `sync.interval` using `alluxio pathConf add /path` then how that takes effect.

The annoying thing is, in the sync process the master is actually the client that needs to do the path resolution, and currently there's no API you can directly call. @maobaolong I will take over this part and add one API for you to use. So in your PR here it should be just an easy call.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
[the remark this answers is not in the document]
elega: +1 if you have different block sizes for different paths, you still need to update them with path config anyways so no need to create a new property for this.

maobaolong:
@jiacheliu3 We have a use case about this PR recently, lets discuss whether we need this PR in os repo.

elega:
ack. let's discuss this on our next community meeting

maobaolong:
please do not mark this stale

Issue #14257, opened 2021-10-14
Support automatic abortion  s3 multipartUpload after a timeout.


jffree:
If a `MultipartUpload` request fails and the `MultipartUpload` request is not aborted,  directory  `xxx_s3_multipart_tmp` will be preserved, and future `MultipartUpload` requests will always error.

**Describe the solution you'd like**

1.  Create a callable instance when a `MultipartUpload` request received which used to abort this `MultipartUpload`. Callable instance contains:
    1.  upload file name (`xxx`)
    2. uploadId
    3. last_update_time (`xxx_s3_multipart_tmp` modify time or `xxx` modify time)

2. Create a `ScheduledExecutorService ` instance used to schedule  auto abort `MultipartUpload` callable instance.
    1. last_update_time + timeout = delay

3. When execute auto abort func:
    1. check `xxx_s3_multipart_tmp` if exist
    2. check `xxx_s3_multipart_tmp` dir id if equals `uploadId`
    3. check if `xxx_s3_multipart_tmp`  modify time changed
    4. check `xxx` if exist and if `xxx` modify time changed

**May caused problem**

* Upload object part or finish `MultipartUpload` request may be distributed to different proxies, how to coordinate different proxies? Leader to do this?

**Urgency**

Normal

ZacBlanco:
Regarding

> Upload object part or finish MultipartUpload request may be distributed to different proxies, how to coordinate different proxies? Leader to do this?

I don't see a clear way to get around this if someone is using multiple proxies. Alluxio just does not allow multiple clients to upload+complete the same file. I think we will just have to document a restriction that the same proxy must be used during a multipart upload to properly complete the request. If a load balancer in front of the proxies is desired it should deterministically redirect the request to the correct proxy

HelloHorizon:
Hey @jffree , is this answer you question?

HelloHorizon:
@ZacBlanco are there any code changes needed for this issue? feel free to open this issue if that is the case.

jffree:
> yes, we do need to make code changes -- in fact there is a TODO in the code for it left by the original author of the S3 handler

@ZacBlanco I can do the preliminary implementation first. Then you can help to review the code.

ZacBlanco:
How does S3 handle this? I think that there should still be some timeout where the upload becomes invalid.

[...]

PR #14328, opened 2021-10-28
Support auto abort multiupload after timeout

jffree:

 #14257

Add these PropertyKey:
1. `alluxio.proxy.s3.multipart.upload.timeout` 
2. `alluxio.proxy.s3.multipart.upload.cleaner.retry.count`
3. `alluxio.proxy.s3.multipart.upload.cleaner.retry.delay`
4. `alluxio.proxy.s3.multipart.upload.cleaner.pool.size`

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3710,14 +3710,14 @@ public final class PropertyKey implements Comparable<PropertyKey> {
   public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_COUNT =
       new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_COUNT)
           .setDefaultValue(3)
-          .setDescription("The retry count when abort multipart upload failed.")
+          .setDescription("The retry count when aborting a multipart upload fails.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.SERVER)
           .build();
   public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_DELAY =
       new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_DELAY)
           .setDefaultValue("10sec")
-          .setDescription("The retry delay time when abort multipart upload failed.")
+          .setDescription("The retry delay time when aborting a multipart upload fails.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.SERVER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3700,6 +3700,34 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_TIMEOUT =
+      new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_TIMEOUT)
+          .setDefaultValue("10min")
+          .setDescription("The timeout for aborting proxy s3 multipart upload automatically.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
+          .build();
+  public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_COUNT =
+      new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_COUNT)
+          .setDefaultValue(3)
+          .setDescription("The retry count when abort multipart upload failed.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
+          .build();
+  public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_DELAY =
+      new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_DELAY)
+          .setDefaultValue("10sec")
+          .setDescription("The retry delay time when abort multipart upload failed.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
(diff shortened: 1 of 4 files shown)


HelloHorizon:
@ZhuTopher please also take a look, thanks!

ZhuTopher (approved):
LGTM, but @ZacBlanco should review this too.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3700,6 +3700,34 @@ public String toString() {
...
+      new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_TIMEOUT)
+          .setDefaultValue("10min")
+          .setDescription("The timeout for aborting proxy s3 multipart upload automatically.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
+          .build();
+  public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_COUNT =
+      new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_COUNT)
+          .setDefaultValue(3)
+          .setDescription("The retry count when abort multipart upload failed.")
```
ZacBlanco, line 3713:
```suggestion
          .setDescription("The retry count when aborting a multipart upload fails.")
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3700,6 +3700,34 @@ public String toString() {
...
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
+          .build();
+  public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_DELAY =
+      new Builder(Name.PROXY_S3_MULTIPART_UPLOAD_CLEANER_RETRY_DELAY)
+          .setDefaultValue("10sec")
+          .setDescription("The retry delay time when abort multipart upload failed.")
```
ZacBlanco, line 3720:
```suggestion
          .setDescription("The retry delay time when aborting a multipart upload fails.")
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3700,6 +3700,34 @@ public String toString() {
...
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.SERVER)
+          .build();
+  public static final PropertyKey PROXY_S3_MULTIPART_UPLOAD_CLEANER_POOL_SIZE =
```
ZacBlanco, line 3724: Personally, I'm not sure if we really need this configuration property. I think it should suffice to just set some static value because I can't imagine a scenario where someone would actually want to configure this -- however, I think it's fine to leave it.

core/server/proxy/src/main/java/alluxio/proxy/s3/MultipartUploadCleaner.java
```diff
@@ -0,0 +1,288 @@
...
+        if (delay > 0) {
+          apply(this, delay);
+        } else {
+          removeTaskRecord(this);
+        }
+      } catch (IOException | AlluxioException e) {
+        mRetryCount++;
+        LOG.error("Failed abort multipart upload {} in bucket {} with uploadId {} with error {}"
+            + " after retry {} count.", mObject, mBucket, mUploadId, e, mRetryCount);
+        e.printStackTrace();
```
ZacBlanco, line 252: we should not print stack traces directly to the stdout. Instead I suggest logging with the following format:

```java
 LOG.error("Failed abort multipart upload {} in bucket {} with uploadId {} after {} retries with error {}"
            + " after retry {} count.", mObject, mBucket, mUploadId, retryCount, e);
```

core/server/proxy/src/main/java/alluxio/proxy/s3/MultipartUploadCleaner.java
```diff
@@ -0,0 +1,288 @@
...
+    return delay;
+  }
+
+  /**
+   * Get multipart uploadId.
+   *
+   * @param bucket the bucket name
+   * @param object the object name
+   */
+  private Long getMultipartUploadId(String bucket, String object)
```
ZacBlanco, line 202: Annotate with `@Nullable`?

core/server/proxy/src/main/java/alluxio/proxy/s3/MultipartUploadCleaner.java
```diff
@@ -0,0 +1,288 @@
...
+   * @param bucket bucket name
+   * @param object object name
+   * @return true if add a abort task
+   */
+  public boolean apply(String bucket, String object)
+      throws IOException, AlluxioException {
+    // Use schedule pool do everyThing.
+    long delay = tryAbortMultipartUpload(bucket, object, null);
+    if (delay > 0) {
+      long uploadId = getMultipartUploadId(bucket, object);
```
ZacBlanco, line 79: This can result in NPE if `getMultipartUploadId` returns null. We should either prevent that method from returning null or handle that here

core/server/proxy/src/main/java/alluxio/proxy/s3/MultipartUploadCleaner.java
```diff
@@ -0,0 +1,288 @@
...
+   * @param bucket bucket name
+   * @param object object name
+   * @param uploadId multipart upload tmp directory fileId
+   * @return true if add a abort task
+   */
+  public boolean apply(String bucket, String object, Long uploadId)
+      throws IOException, AlluxioException {
+    // Use schedule pool do everyThing.
+    if (uploadId == null) {
+      uploadId = getMultipartUploadId(bucket, object);
```
ZacBlanco, line 97: This can result in NPE if getMultipartUploadId returns null. We should either prevent that method from returning null or handle that here

ZacBlanco (changes requested):
Apologies to delaying this review. Thank you for implementing this feature

core/common/src/main/java/alluxio/conf/PropertyKey.java
jffree: We only need to write a few lines of code to expose this configuration, and it is up to the user to decide whether it should be used or not.

core/server/proxy/src/main/java/alluxio/proxy/s3/MultipartUploadCleaner.java
jffree: Do you means this?
```
LOG.error("Failed abort multipart upload {} in bucket {} with uploadId {} after {} retries with error {}.", 
            mObject, mBucket, mUploadId, mRetryCount, e);
```

ZacBlanco (approved):
Thanks for the improvement!

ggezer:

Xenorith:

PR #14351, opened 2021-10-29
Implement configurable ExecutorService creation for Alluxio processes

### LINKED ISSUE #14326: Configurable ExecutorService provider for Alluxio processes
Main executors that are used by master/worker processes are not configurable.
This leaves us with custom code changes to mitigate issues.

By leveraging propertry-key templates, we could easily provide common infra to configure executor-services for Alluxio nodes.
There are not really many options when it comes to providing executor-services to gRPC servers:
- FJP
- Fixed Thread Pool
- Cached Thread Pool
For the last two, we could also provide support for several internal queue implementations. (LinkedBlocking, SynchronousBlocking etc...)

**Urgency**
It would be very nice to have before the next release.

ggezer:
Fix #14326

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2737,41 +2737,26 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setScope(Scope.MASTER)
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
-  // In Java8 in container environment Runtime.availableProcessors() always returns 1,
-  // which is not the actual number of cpus, so we set a safe default value 8.
-  public static final PropertyKey MASTER_RPC_EXECUTOR_PARALLELISM =
-      new Builder(Name.MASTER_RPC_EXECUTOR_PARALLELISM)
-          .setDefaultSupplier(() -> Math.max(8, 2 * Runtime.getRuntime().availableProcessors()),
-              "2 * {CPU core count}")
-          .setDescription("The parallelism level of master RPC executor service .")
-          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
-          .setScope(Scope.MASTER)
-          .build();
-  public static final PropertyKey MASTER_RPC_EXECUTOR_MIN_RUNNABLE =
-      new Builder(Name.MASTER_RPC_EXECUTOR_MIN_RUNNABLE)
-          .setDefaultValue(1)
-          .setDescription("the minimum allowed number of core threads not blocked. "
-              + "To ensure progress, when too few unblocked threads exist and unexecuted tasks may "
-              + "exist, new threads are constructed up to the value of "
-              + Name.MASTER_RPC_EXECUTOR_MAX_POOL_SIZE
-              + ". A value of 1 ensures liveness. A larger value might improve "
-              + "throughput but might also increase overhead.")
+  public static final PropertyKey MASTER_RPC_EXECUTOR_TYPE =
+      new Builder(Name.MASTER_RPC_EXECUTOR_TYPE)
+          .setDefaultValue("ThreadPoolExecutor")
+          .setDescription("Type of ExecutorService for Alluxio master gRPC server. "
+              + "Supported values are ThreadPoolExecutor and ForkJoinPool.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
   public static final PropertyKey MASTER_RPC_EXECUTOR_CORE_POOL_SIZE =
       new Builder(Name.MASTER_RPC_EXECUTOR_CORE_POOL_SIZE)
           .setDefaultValue(500)
-          .setDescription("the number of threads to keep in thread pool of master RPC executor "
-              + "service. By default it is same as the parallelism level, but may be "
-              + "set to a larger value to reduce dynamic overhead if tasks regularly block. ")
+          .setDescription(
+              "The number of threads to keep in thread pool of master RPC ExecutorService.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
   public static final PropertyKey MASTER_RPC_EXECUTOR_MAX_POOL_SIZE =
       new Builder(Name.MASTER_RPC_EXECUTOR_MAX_POOL_SIZE)
           .setDefaultValue(500)
-          .setDescription("the maximum number of threads allowed for master RPC executor service."
+          .setDescription("The maximum number of threads allowed for master RPC ExecutorService."
               + " When the maximum is reached, attempts to replace blocked threads fail.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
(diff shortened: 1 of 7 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2737,53 +2737,90 @@ public String toString() {
           .setScope(Scope.MASTER)
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .build();
-  // In Java8 in container environment Runtime.availableProcessors() always returns 1,
-  // which is not the actual number of cpus, so we set a safe default value 8.
-  public static final PropertyKey MASTER_RPC_EXECUTOR_PARALLELISM =
```
yuzhu, line 2742: can we keep these around and set the default FJP or TPE parallelism to this for better back compat?

core/server/master/src/test/java/alluxio/master/AlluxioMasterProcessTest.java
```diff
@@ -171,32 +171,35 @@ public void restoreFromBackupLocal() throws Exception {
 
   @Test
   public void startZeroParallelism() {
-    ServerConfiguration.set(PropertyKey.MASTER_RPC_EXECUTOR_PARALLELISM, "0");
+    ServerConfiguration.set(PropertyKey.MASTER_RPC_EXECUTOR_TYPE, "FJP");
```
yuzhu, line 174: can we also add some sanity test for TPE?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2737,53 +2737,90 @@ public String toString() {
...
-          .setDefaultValue(1)
-          .setDescription("the minimum allowed number of core threads not blocked. "
-              + "To ensure progress, when too few unblocked threads exist and unexecuted tasks may "
-              + "exist, new threads are constructed up to the value of "
-              + Name.MASTER_RPC_EXECUTOR_MAX_POOL_SIZE
-              + ". A value of 1 ensures liveness. A larger value might improve "
-              + "throughput but might also increase overhead.")
+  public static final PropertyKey MASTER_RPC_EXECUTOR_TYPE =
+      new Builder(Name.MASTER_RPC_EXECUTOR_TYPE)
+          .setDefaultValue("TPE")
```
yuzhu, line 2742: can we use full class name instead of TPE and FJP?

yuzhu (changes requested):
thanks @ggezer , some early comments

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2737,53 +2737,90 @@ public String toString() {
...
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_RPC_EXECUTOR_TPE_QUEUE_TYPE =
+      new Builder(Name.MASTER_RPC_EXECUTOR_TPE_QUEUE_TYPE)
+          .setDefaultValue("LINKED_BLOCKING_QUEUE")
+          .setDescription(String.format(
+              "This property is effective when %s is set to TPE. "
+                  + "It specifies the internal task queue that's used by RPC ExecutorService. "
+                  + "Supported values are: LINKED_BLOCKING_QUEUE, LINKED_BLOCKING_QUEUE_WITH_CAP, "
```
jiacheliu3, line 2778: If you have an `enum TpeType` you may do `TpeType.values()` here.

core/server/common/src/main/java/alluxio/master/AlluxioExecutorService.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * @return the current RPC queue size
+   */
+  public long getRpcQueueLength() {
+    if (mExecutor instanceof ThreadPoolExecutor) {
+      return ((ThreadPoolExecutor) mExecutor).getQueue().size();
+    } else if (mExecutor instanceof ForkJoinPool) {
+      return ((ForkJoinPool) mExecutor).getQueuedSubmissionCount();
+    } else {
+      throw new IllegalArgumentException("Executor is not supported.");
```
jiacheliu3, line 50: print the `Executor.getClassName()` here so we know what other types people may want?

job/server/src/main/java/alluxio/master/job/plan/PlanTracker.java
```diff
@@ -105,8 +105,8 @@ public PlanTracker(long capacity, long retentionMs,
     Preconditions.checkArgument(retentionMs >= 0);
     mRetentionMs = retentionMs;
     mMaxJobPurgeCount = maxJobPurgeCount <= 0 ? Long.MAX_VALUE : maxJobPurgeCount;
-    mCoordinators = new ConcurrentHashMap<>(0,
-        0.95f, ServerConfiguration.getInt(PropertyKey.MASTER_RPC_EXECUTOR_PARALLELISM));
+    mCoordinators = new ConcurrentHashMap<>(0, 0.95f,
```
jiacheliu3, line 108: Move the comment/description from the property key here so we keep why the `Math(8, 2*nCPU)` here?

core/server/common/src/main/java/alluxio/master/ExecutorServiceBuilder.java
```diff
@@ -0,0 +1,225 @@
...
+   * @param executorHost Where the executor is needed
+   * @return instance of {@link ExecutorService}
+   */
+  public static AlluxioExecutorService buildExecutorService(RpcExecutorHost executorHost) {
+    // Get executor type for given host.
+    RpcExecutorType executorType = ServerConfiguration.getEnum(
+        PropertyKey.Template.RPC_EXECUTOR_TYPE.format(executorHost.toString()),
+        RpcExecutorType.class);
+    // Build thread name format.
+    String threadNameFormat = executorHost + "-rpc-executor-thread-%d";
```
jiacheliu3, line 134: Does it make sense to reflect the threadpool type in the thread names too?

core/server/common/src/main/java/alluxio/master/ExecutorServiceBuilder.java
```diff
@@ -0,0 +1,225 @@
...
+          queue = new LinkedBlockingQueue<>(maxPoolSize);
+          break;
+        case ARRAY_BLOCKING_QUEUE:
+          queue = new ArrayBlockingQueue<>(maxPoolSize);
+          break;
+        case SYNCHRONOUS_BLOCKING_QUEUE:
+          queue = new SynchronousQueue<>();
+          break;
+        default:
+          throw new IllegalArgumentException();
```
jiacheliu3, line 210: Add a message here on what the `queueType` is?

core/server/common/src/test/java/alluxio/master/ExecutorServiceBuilderTest.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.master;
+
+import alluxio.conf.PropertyKey;
+import alluxio.conf.ServerConfiguration;
+import org.junit.Before;
+import org.junit.Rule;
```
jiacheliu3, line 17: oops checkstyle will complain

jiacheliu3 (changes requested):
Thanks for the improvement! Overall this looks great, but I have some high level questions.

I was imagining we do this in a way similar to how we specify `alluxio.worker.allocator.class`. There's an interface for an `ExecutorFactory` and we have `FJPFactory` and `FixedThreadPoolFactor` and the user can write their own factories. This will probably make the configuration easier because the user will go to that class and see how the code(property keys) apply. 

However the current approach also looks good. Partly that's because the TPE constructor only has 7 parameters. Not toooo hard to wrap in property keys.

Just raising an alternative for discussion. Either looks good to me. What do people think?

codings-dan:
@ggezer hi Gorkturk，I would like to know why we chose to add the option of Thread Pool. Compared with FrokJoin Pool, what are its advantages? Are there any performance test results?  thx!

ggezer:
> @ggezer hi Gorkturk，I would like to know why we chose to add the option of Thread Pool. Compared with FrokJoin Pool, what are its advantages? Are there any performance test results? thx!

Hey. With ThreadPoolExecutor we can avoid pitfalls of FJP. FJP is throughout optimized and assumes short execution times. When a task is expected to run for a long time, it should notify FJP via `ForkJoinPool.managedBlock()`. When this is not done, FJP has trouble scaling up its internal thread count. So you could start seeing issues like; worker heartbeats not going through, RPCs not being served etc...

job/server/src/main/java/alluxio/master/job/plan/PlanTracker.java
ggezer: @yuzhu Why did we relied on this property in the first place?

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: Can't do unless I move interfaces down to core/common.

core/server/master/src/test/java/alluxio/master/AlluxioMasterProcessTest.java
ggezer: Al those that are asserted are tested already. (keepalive etc is effective for both pools)

core/common/src/main/java/alluxio/conf/PropertyKey.java
ggezer: I set aliases for old FJP related settings. However they won't take effect due to switching to TPE by default.
Rest are untouched already (core/max pool etc...)

Beyond that, any custom configuration that touches parallelism is most probably wrong.

ggezer:
@yuzhu @jiacheliu3 Now that we have an infra, what do you guys think about introducing keys for job-master and worker processes?

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: right....

job/server/src/main/java/alluxio/master/job/plan/PlanTracker.java
yuzhu: Should really make a new property, but everyone hates that..

yuzhu (approved):

yuzhu:
Lgtm if build and test passes

jiacheliu3 (approved):
LGTM, thanks!

Xenorith:

Xenorith:

PR #14382, opened 2021-11-02
Fix incorrect available bytes when change space in StorageDir

qian0817:

Fix incorrect available bytes when change space in StorageDir.

Otherwise the value of available bytes will greater than real available bytes And also make it thread-safe.

No user facing changes.

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/meta/DefaultStorageDir.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/meta/DefaultStorageDir.java
@@ -382,18 +382,23 @@ public final class DefaultStorageDir implements StorageDir {
   }
 
   private void reclaimSpace(long size, boolean committed) {
-    Preconditions.checkState(mCapacityBytes >= mAvailableBytes.get() + size,
-        "Available bytes should always be less than total capacity bytes");
-    mAvailableBytes.addAndGet(size);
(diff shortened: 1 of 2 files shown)


core/server/worker/src/main/java/alluxio/worker/block/meta/DefaultStorageDir.java
```diff
@@ -382,18 +382,23 @@ public long getReservedBytes() {
...
   }
 
   private void reserveSpace(long size, boolean committed) {
-    Preconditions.checkState(size <= mAvailableBytes.get() + mReservedBytes.get(),
-        "Available bytes should always be non-negative");
-    mAvailableBytes.getAndSet(Math.max(0, mAvailableBytes.get() - size));
+    mAvailableBytes.getAndUpdate(oldAvailableBytes -> {
+      Preconditions.checkState(size <= oldAvailableBytes + getReservedBytes(),
+              "Available bytes should always be non-negative");
+      return oldAvailableBytes - size;
```
qian0817, line 400: If there is no available bytes, then require space less than reserved bytes, It will sucess and don't change  the value of available bytes. If afterwards reclaim space the value of available bytes will greater than real available bytes.
In this case, It is reasonable for available bytes to have a negative number.

ggezer (approved):
Thanks @qian0817 !

ggezer:

jja725:

Xenorith:

Issue #14379, opened 2021-11-02
alluxio-fuse get fuse mount  process id incorrectly


kevincai:
**Alluxio Version:**
v2.6.2

**Describe the bug**
when doing csi umount, alluxio-fuse was trying to get process pid by `ps aux`, script function `fuse_stat` returned process user instead its pid, and incorrectly kill it. which caused crash during exit

**To Reproduce**
* create a pvc and bind to a pod
* after pod started the volume is mounted to pod
* delete the pod, csi nodeplugin was trying to umount the volume, and error out
```
I1102 09:17:35.251923       1 utils.go:97] GRPC call: /csi.v1.Node/NodeUnpublishVolume
I1102 09:17:35.251942       1 utils.go:98] GRPC request: {"target_path":"/var/lib/kubelet/pods/50810713-80f9-4a9e-8ff1-75e179ce6334/volumes/kubernetes.io~csi/alluxio-csi-inline-vol/mount","volume_id":"csi-1d1386662f230eed9ef2db50e770672bb711bb7db67b3e95b7b4520273663840"}
I1102 09:17:35.366812       1 main.go:84] Signal received: child exited
I1102 09:17:35.368138       1 nodeserver.go:113] /opt/alluxio/integration/fuse/bin/alluxio-fuse: line 91: kill: root: arguments must be process or job IDs
Successfully unmount fuse at /var/lib/kubelet/pods/50810713-80f9-4a9e-8ff1-75e179ce6334/volumes/kubernetes.io~csi/alluxio-csi-inline-vol/mount (PID: root).

W1102 09:17:35.391828       1 mount_helper_common.go:129] Warning: "/var/lib/kubelet/pods/50810713-80f9-4a9e-8ff1-75e179ce6334/volumes/kubernetes.io~csi/alluxio-csi-inline-vol/mount" is not a mountpoint, deleting

```

**Expected behavior**
Clean umount, no crash

**Urgency**
Describe the impact and urgency of the bug.

**Are you planning to fix it**
Please indicate if you are already working on a PR. 

**Additional context**

```
ps aux command output

PID   USER     TIME  COMMAND
    1 root      0:00 /usr/local/bin/alluxio-csi --v=5 --nodeid=minikube-m02 --endpoint=unix:///csi/csi.sock
[...]

PR #14422, opened 2021-11-04
Fix alluxio-fuse inaccurate parse ps output

kevincai:
* explicitly set required field by '-o' option to eliminate ambiguous
  output column order

explicit set '-o' option to eliminate different version of ps command ambiguous output format

ps command in alpine image is from busybox, which pid is the first column by default.

No

Fixes https://github.com/Alluxio/alluxio/issues/14379


LuQQiu (approved):
LGTM, thanks for fixing the pid info

LuQQiu:
@ssz1997 FYI

LuQQiu:

PR #14425, opened 2021-11-04
[DOCFIX] Improve file system api doc

dbw9580:
Improve file system api doc.


docs/en/api/FS-API.md
```diff
@@ -1,36 +1,38 @@
 ---
 layout: global
-title: Filesystem API
+title: File System API
```
dbw9580, line 3: Our [documentation convention](https://docs.alluxio.io/os/user/edge/en/contributor/Documentation-Conventions.html#terminology-table) says that `filesystem` is wrong.

docs/en/api/FS-API.md
```diff
@@ -353,120 +359,13 @@ Alluxio has a [Python Client](https://github.com/Alluxio/alluxio-py) for interac
...
-
-The Python client interacts with Alluxio through the REST API provided by the Alluxio proxy.
-
-The proxy is a standalone server that can be started using
-`${ALLUXIO_HOME}/bin/alluxio-start.sh proxy` and stopped using `${ALLUXIO_HOME}/bin/alluxio-stop.sh
-proxy`. By default, the REST API is available on port 39999.
-
-There are performance implications of using the HTTP proxy. In particular, using the proxy requires
-an extra hop. For optimal performance, it is recommended to run the proxy server and an Alluxio
-worker on each compute node.
```
dbw9580, line 368: Duplicate content.

docs/en/api/FS-API.md
```diff
@@ -353,120 +359,13 @@ Alluxio has a [Python Client](https://github.com/Alluxio/alluxio-py) for interac
...
-
-    info("asserting that %s is deleted" % py_test_root_dir)
-    assert not client.exists(py_test_root_dir)
-    info("done")
-
-
-if __name__ == '__main__':
-    main()
-```
-
```
dbw9580, line 469: This is the `example.py` file mentioned above.

docs/en/api/FS-API.md
```diff
@@ -475,18 +374,6 @@ See the [godoc](http://godoc.org/github.com/Alluxio/alluxio-go) for detailed doc
...
-The Go client talks to Alluxio through the REST API provided by the Alluxio proxy.
-
-The proxy is a standalone server that can be started using
-`${ALLUXIO_HOME}/bin/alluxio-start.sh proxy` and stopped using `${ALLUXIO_HOME}/bin/alluxio-stop.sh
-proxy`. By default, the REST API is available on port 39999.
-
-There are performance implications of using the HTTP proxy. In particular, using the proxy requires
-an extra hop. For optimal performance, it is recommended to run the proxy server and an Alluxio
-worker on each compute node.
-
```
dbw9580, line 489: Duplicate content.

docs/en/api/FS-API.md
Xenorith: i think it is better to keep this here and point the user to the example.py file for ease of utilizing or modifying. it is easier for someone to read from the doc than to take the extra step to open the example file. what do you think?

docs/en/api/FS-API.md
Xenorith: okay with removing duplicate content but it should still be contained somewhere in this doc for the python and go client sections to point to. i agree it is not good to have two copies but it is also not good to have zero copies

dbw9580: Yes there is one copy left. See here: https://github.com/Alluxio/alluxio/blob/90fc2207eaee1813c4f66ccacc3bdc6308046b8d/docs/en/api/FS-API.md#L345-L351

These are the duplicates:
https://github.com/Alluxio/alluxio/blob/90fc2207eaee1813c4f66ccacc3bdc6308046b8d/docs/en/api/FS-API.md#L365-L371
https://github.com/Alluxio/alluxio/blob/90fc2207eaee1813c4f66ccacc3bdc6308046b8d/docs/en/api/FS-API.md#L485-L491

I will add a pointer to that section under the python and go sections, so that impatient people have a chance to read about the performance implications.

docs/en/api/FS-API.md
dbw9580: Makes sense. I think it's helpful to add a line that points out that this example is the example found in the python package repository.

docs/en/api/FS-API.md
Xenorith: great!

Xenorith (approved):

Xenorith:

Xenorith:

PR #14481, opened 2021-11-09
Fix S3 region when not value set in instance

LuQQiu:

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
+++ b/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
@@ -86,6 +86,9 @@ import javax.annotation.Nullable;
 public class S3AUnderFileSystem extends ObjectUnderFileSystem {
   private static final Logger LOG = LoggerFactory.getLogger(S3AUnderFileSystem.class);
 
+  /** The special S3 region which can be used to talk to any bucket. */
+  private static final String S3_DEFAULT_REGION = "us-east-1";
+
   /** Static hash for a directory's empty contents. */
   private static final String DIR_HASH;
 
@@ -236,6 +239,7 @@ public class S3AUnderFileSystem extends ObjectUnderFileSystem {
       // access bucket without region information
       // at the cost of an extra HEAD request
       clientBuilder.withForceGlobalBucketAccessEnabled(true);
+      clientBuilder.setRegion(S3_DEFAULT_REGION);
     }
 
     AmazonS3 amazonS3Client = clientBuilder.build();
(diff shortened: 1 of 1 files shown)


yuzhu:
i think if we initialize the s3 client with `AmazonEC2ClientBuilder.standard()` then we won't have this issue. https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html

LuQQiu:
@yuzhu I tested on my side that this fix is sufficient, the standard may also works, can you help test on your side as well?

LuQQiu:
I didn't see standard() modify the s3 region information, looks like it's similar to what we do now
```
  public static AmazonS3ClientBuilder standard() {
    return (AmazonS3ClientBuilder)(new AmazonS3ClientBuilder()).withCredentials(new S3CredentialsProviderChain());
  }
```
```
    AWSCredentialsProvider credentials = createAwsCredentialsProvider(conf);
  public static AWSCredentialsProvider createAwsCredentialsProvider(
      UnderFileSystemConfiguration conf) {
    // Set the aws credential system properties based on Alluxio properties, if they are set;
    // otherwise, use the default credential provider.
    if (conf.isSet(PropertyKey.S3A_ACCESS_KEY)
        && conf.isSet(PropertyKey.S3A_SECRET_KEY)) {
      return new AWSStaticCredentialsProvider(new BasicAWSCredentials(
          conf.get(PropertyKey.S3A_ACCESS_KEY), conf.get(PropertyKey.S3A_SECRET_KEY)));
    }
    // Checks, in order, env variables, system properties, profile file, and instance profile.
    return new DefaultAWSCredentialsProviderChain();
  }
    AmazonS3ClientBuilder clientBuilder = AmazonS3Client.builder()
        .withCredentials(credentials)
```

LuQQiu:
Tested with s3 access key and secret key in ~/.aws/credentials
s3 bucket in us-east-1, us-west-1, Singapore
bin/alluxio runTests
all passed

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -86,6 +86,9 @@
 public class S3AUnderFileSystem extends ObjectUnderFileSystem {
   private static final Logger LOG = LoggerFactory.getLogger(S3AUnderFileSystem.class);
 
+  /** The special S3 region which can be used to talk to any bucket. */
+  private static final String S3_DEFAULT_REGION = "us-east-1";
```
yuzhu, line 90: can we use their enum for this? aws sdk has defined enums

yuzhu (approved):
Thanks @LuQQiu LGTM

Xenorith:

Xenorith:

PR #14490, opened 2021-11-10
Add stacks page in worker web UI

qian0817:

Add stacks page in worker web UI. Also refactored the code about stacks webui.

It's useful to also see thread dump in worker web ui.

A new stack page.

configuration `aluxio.web.threaddump.log.enabled` determines whether print thread info to the log when access stacks ui.

image


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -536,11 +536,54 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey WEB_CORS_ALLOW_CREDENTIAL =
```
maobaolong, line 539: Please give some description for these batch of config. As this PR is a little big, give more information will be easy to review, thanks

qian0817: The purpose is to make cors origin, methods and headers configurable. FYI https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#the_http_response_headers

maobaolong: Could you create a new ticket? So that we can review this PR focus on worker stacks page.

qian0817: @maobaolong Ok, I submit a new PR #14562 .

core/server/common/src/main/java/alluxio/web/WebServer.java
```diff
@@ -45,6 +45,7 @@
 public abstract class WebServer {
   private static final Logger LOG = LoggerFactory.getLogger(WebServer.class);
   private static final String DISABLED_METHODS = "TRACE,OPTIONS";
+  private static final String THREAD_DUMP_PATH = "/api/v1/common/thread_dump";
```
maobaolong, line 48: Why you changed the `stack` path which is user interface level, suggest to keep backward compatible, in another world, it should keep consistent with `routePaths.tsx`

maobaolong (commented):
It Looks good to me almost, left a single comment.

core/server/common/src/main/java/alluxio/web/WebServer.java
qian0817: This URL is reused by the both frontend and backend, and you can find the unexcepted behavior by refreshing page in the stacks page. Also I think api interface should always start with `/api` to ensure consistency. So i change the path.

webui/worker/src/constants/data/routePaths.tsx
```diff
@@ -15,5 +15,6 @@ export const routePaths = {
   blockInfo: '/blockInfo',
   logs: '/logs',
   metrics: '/metrics',
+  stacks: '/stacks',
```
maobaolong, line 18: Do you mean to `"/api/v1/common/thread_dump`?

qian0817: These two path values should be different.

LuQQiu:
can stacks tab be moved after `configuration` tab since it's more advanced and more for developer/maintainer purpose?

LuQQiu (approved):
LGTM, except the previous comment

qian0817:
> can stacks tab be moved after `configuration` tab since it's more advanced and more for developer/maintainer purpose?

done.

core/server/common/src/main/java/alluxio/web/WebServer.java
LuQQiu: @tieujason330  is this thread dump path follows convention?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -551,6 +551,14 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.SERVER)
           .build();
+  public static final PropertyKey WEB_THREAD_DUMP_TO_LOG =
```
jiacheliu3, line 554: How many lines of log will this print?

core/server/common/src/main/java/alluxio/web/WebServer.java
jiacheliu3: Please make this configurable by a property, as we are hearing different opinions about the path.

jiacheliu3 (changes requested):
@qian0817 Thanks for the improvements! I'm a bit confused as the PR title does not match what it does. Is this updated?

core/common/src/main/java/alluxio/conf/PropertyKey.java
qian0817: All threads will be print here without limit.

core/server/common/src/main/java/alluxio/web/WebServer.java
qian0817: The frontend needs to use this api and if it is configurable then it means that the frontend needs to additionally fetch the configuration and thus increase the complexity. I don't think it's necessary to make this configurable.

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: I've seen many deployments have 1000 or even 2000 threads. This means you are dumping at least 1000*50 lines at once. I don't think I want 50K lines of thread dumps while I'm checking the logs. What is the use case here? Do you know @maobaolong ?

maobaolong: @jiacheliu3 How about set this default as false? It would not be a good idea if put 50k line into logger as default.

maobaolong (commented):
@qian0817 Thanks for the changes, I suggest to set WEB_THREAD_DUMP_TO_LOG default to false, how about your idea?

webui/common/src/components/Stacks/Stacks.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+      it('Matches snapshot with Table listing', () => {
+        expect(shallowWrapper).toMatchSnapshot();
+      });
+    });
+
+    describe('Renders FileView', () => {
+      it('Matches snapshot with File', () => {
+        expect(shallowWrapper).toMatchSnapshot();
+      });
+    });
```
tieujason330, line 73: both of these tests do the same thing. remove one of them and rename the other to be Stacks-related

core/server/common/src/main/java/alluxio/web/WebServer.java
tieujason330: i think this is fine since it is consistent with `/api/v1/master` and `/api/v1/worker`

tieujason330 (changes requested):
minor change for the stacks test file

webui/common/src/components/Stacks/Stacks.test.tsx
qian0817: done

core/common/src/main/java/alluxio/conf/PropertyKey.java
qian0817: I don't know why require printing thread information to the log and why the user need to get it from the log when they can already get it from webui.

webui/common/src/components/Stacks/Stacks.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { createBrowserHistory, History, LocationState } from 'history';
+import React from 'react';
+import sinon from 'sinon';
+import { AllStacksProps, Stacks } from './Stacks';
+import { initialState } from '../../../../master/src/store';
+import { createAlertErrors } from '../../utilities';
+
+configure({ adapter: new Adapter() });
+
+describe('Stacks-related', () => {
```
tieujason330, line 23: sorry my previous comment was unclear, i didn't mean to say it should be renamed to "Stacks-related". You can just call it "Stacks test".

webui/common/src/components/Stacks/Stacks.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    it('Renders without crashing', () => {
+      expect(shallowWrapper.length).toEqual(1);
+    });
+
+    it('Contains a div with class col-12', () => {
+      expect(shallowWrapper.find('.col-12').length).toEqual(1);
+    });
+
+    describe('Renders Log', () => {
```
tieujason330, line 63: this `describe` can be removed

webui/common/src/components/Stacks/Stacks.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      it('Matches snapshot with Table listing', () => {
```
tieujason330, line 64: rename to 'Matches snapshot'. I don't think this has to do with tables

tieujason330 (changes requested):

webui/common/src/components/Stacks/Stacks.test.tsx
qian0817: done

webui/common/src/components/Stacks/Stacks.test.tsx
qian0817: done

webui/common/src/components/Stacks/Stacks.test.tsx
qian0817: done

tieujason330 (approved):
lgtm, thanks!

core/server/common/src/main/java/alluxio/web/WebServer.java
jiacheliu3: If we change the `v1` to `v2` in the webui code (probably the jetty part?) this will be an incompatibility. @tieujason330 Do you think there's a way to at least leave a marker so we will notice before upgrade?

core/common/src/main/java/alluxio/conf/PropertyKey.java
jiacheliu3: @LuQQiu Do you think a separate logger and log destination will be preferable? Or at least there should be a rate limiter on this request. If I blindly refresh on this page, i can DDoS  the server in less than 10 requests.

jiacheliu3: @tieujason330 how much overhead would there be to add a button in the webUI to explicitly ask the process to dump it?

core/server/common/src/main/java/alluxio/web/WebServer.java
qian0817: @jiacheliu3 Is it ok if I modify the code in this way?  I think this will help us to notice.
```suggestion
  private static final String THREAD_DUMP_PATH = REST_API_PREFIX + "/common/thread_dump";
```

tieujason330: can use constant `REST_API_PREFIX` in `Constants.java` for consistency. there isn't an automated way to track version changes between server and ui code. we'll manually have to keep track

core/common/src/main/java/alluxio/conf/PropertyKey.java
tieujason330: adding a button would be straightfoward:
- add button on stacks page
- create redux action + saga to send POST request to server
- server needs new POST handler that'll tell process to dump

qian0817: I think a cache could be added to prevent performance problems caused by multiple refreshes. @jiacheliu3

qian0817: Can I make this a standalone assignment, which kind of defeats the purpose of this PR in the first place?

LuQQiu:
@qian0817 This PR will be merged by end of week after 2.7.2 release

qian0817:
ping @LuQQiu .

LuQQiu:

PR #14520, opened 2021-11-15
fix concurrency problem when generating PropertyKey with Template object

hfutatzhanghb:

make PropertyKey#fromString method synchronized to prevent get wrong ProperKey object in concurrency scene.

Concurrency problems occurs in the following code segment of PropertyKey#fromString method.
```java
for (Template template : Template.values()) {
      key = template.getPropertyKey(input);
      if (key != null) {
        return key;
      }
    }
```
Inside template.getPropertyKey method, use static variable PropertyCreators.DEFAULT_PROPERTY_CREATOR to create 
PropertyKey object, this static variable is shared among multi threads and consequently leads to  concurrency problems

To explain this problems clearly，I add some logs in the for loop to print template object, input string and generated PropertyKey object. Then, I start 50 threads to read data on Alluxio, Here is some logs in my test program. From the logs below, we can see fromString method wrongly generates PropertyKey(alluxio.ns1.zookeeper.lead
er.path) when the input key is "alluxio.ns2.zookeeper.election.path".

```
[INFO ][2021/11/14 21:15:22,869 ][pool-1-thread-21][alluxio.conf.PropertyKey.fromString(PropertyKey.java:6869)] ZHB###propertyKey : alluxio.ns1.zookeeper.lead
er.path, key: alluxio.ns2.zookeeper.election.path, template: Template{format=alluxio.%s.zookeeper.election.path, pattern=alluxio\.(\w+)\.zookeeper\.election\.
path}
```

no user facing changes introduced .


hfutatzhanghb:
@ZhuTopher Please take a look. Thanks a lot.

hfutatzhanghb:
@ZhuTopher  Please take a look, thanks a lot~

ZhuTopher:
Thanks for the detailed description, that helps a lot! I'm concerned about whether or not it is necessary to use `synchronized` on the entire method or if we can instead protect the underlying objects/methods.

I'll make a note of this PR and someone will look into it when we have the opportunity.

hfutatzhanghb:
> Thanks for the detailed description, that helps a lot! I'm concerned about whether or not it is necessary to use `synchronized` on the entire method or if we can instead protect the underlying objects/methods.
> 
> I'll make a note of this PR and someone will look into it when we have the opportunity.

Thanks。

hfutatzhanghb:
image

hfutatzhanghb:
@apc999 @jiacheliu3 @ZhuTopher    Hi~, i have updated some descriptions about this problem. but In in Chinese.

qian0817:
@hfutatzhanghb 
looks like it only need to prevent multi threads accessing PropertyCreators.DEFAULT_PROPERTY_CREATOR. I have re-implemented this change, do you think it works?
```diff
     private BiFunction<String, PropertyKey, PropertyKey> mPropertyCreator =
-        PropertyCreators.DEFAULT_PROPERTY_CREATOR;
+        (name, baseProperty) -> new Builder(name).buildUnregistered();
```

jiacheliu3:
> @hfutatzhanghb looks like it only need to prevent multi threads accessing PropertyCreators.DEFAULT_PROPERTY_CREATOR. I have re-implemented this change, do you think it works?
> 
> ```diff
>      private BiFunction<String, PropertyKey, PropertyKey> mPropertyCreator =
> -        PropertyCreators.DEFAULT_PROPERTY_CREATOR;
> +        (name, baseProperty) -> new Builder(name).buildUnregistered();
> ```

hi @qian0817 and @hfutatzhanghb I think the solution proposed by @qian0817 is correct and should be the way to go.
The problem is basically https://github.com/Alluxio/alluxio/blob/a9d1ab0c06386d4b015483d4bb3df173f1aea6ca/core/common/src/main/java/alluxio/conf/PropertyKey.java#L7373 where the static variable uses this one `Builder` object for all requests without synchronization. `NESTED_UFS_PROPERTY_CREATOR ` and `NESTED_JOURNAL_PROPERTY_CREATOR ` do not have the same problem because the `Builder` is created within the lambda, so it's a different `Builder` every time the `BiFunction` is called. We should do the same with `DEFAULT_PROPERTY_CREATOR`.

You can change the `DEFAULT_PROPERTY_CREATOR` as below:
```
      private static final BiFunction<String, PropertyKey, PropertyKey> DEFAULT_PROPERTY_CREATOR =
          fromBuilder();

      private static BiFunction<String, PropertyKey, PropertyKey> fromBuilder() {
        return (name, baseProperty) -> 
        {
          Builder builder = new Builder();
          builder.setName(name).buildUnregistered();
        }
      }
```

fyi @apc999

hfutatzhanghb:
> 

i will try to fix it in that way and make some tests.

PR #14526, opened 2021-11-16
Display unescape the ufs URI

ljl1988com:

Display unescape the ufs URI.

Display unescape the ufs URI

original:

now:


yuzhu (changes requested):
we should unescape on the browser side? i think http requires some characters to be escaped? @tieujason330 what do you think>?

tieujason330:
@yuzhu since this is part of the response's body, i don't think it needs to be escaped. i think that'll get handled via json serialization.

ljl1988com:
@tieujason330 @yuzhu I find some codes as follows.

https://github.com/Alluxio/alluxio/blob/c642abc475878811c9e0dcbd28c842edd2639d5c/core/server/master/src/main/java/alluxio/master/meta/AlluxioMasterRestServiceHandler.java#L1062-L1070

If that'll get handled via json serialization. I think we still set serialization method of those properties. It's too complex.

maobaolong (approved):
Nice catch, LGTM, Thanks for this improvement.

yuzhu (approved):

yuzhu:

jiacheliu3:

PR #14530, opened 2021-11-16
Update DistributedLoad using cache request

jja725:

Update DistributedLoad using cache request instead of read with passive cache.

using read to cache block is not efficient and error proune. Use cache request directly.

no

### DIFF
--- a/core/common/src/main/java/alluxio/exception/ExceptionMessage.java
+++ b/core/common/src/main/java/alluxio/exception/ExceptionMessage.java
@@ -279,8 +279,8 @@ public enum ExceptionMessage {
   TRANSFORM_TABLE_URI_LACKS_AUTHORITY("URI {0} lacks authority"),
 
   // job service
-  NO_LOCAL_BLOCK_WORKER_REPLICATE_TASK(
-      "Cannot find a local block worker to replicate blockId {0,number,#}"),
+  NO_LOCAL_BLOCK_WORKER_LOAD_TASK(
+      "Cannot find a local block worker to load blockId {0,number,#}"),
 
   // job manager
   JOB_DEFINITION_DOES_NOT_EXIST("The job definition for config {0} does not exist"),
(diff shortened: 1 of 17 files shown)


maobaolong:
@jja725 As `CacheRequest` is a new come feature, I hope it would be stable, but maybe there are some bugs for several months in the future, so I suggest to keep an option to load data by original read back mode, once the cacherequest way fail, we can also using distributedload command. If we keep the option for previous read-back data mode, we can have more courage to upgrade to the new version or back porting this patch.

yuzhu (commented):
@jja725 overall looks good. @maobaolong 's suggest is a valid one, but I am also interested in the performance aspect of this. Could you do some benchmarking using two versions and see how much it improves load performance?

jja725:
> @jja725 overall looks good. @maobaolong 's suggest is a valid one, but I am also interested in the performance aspect of this. Could you do some benchmarking using two versions and see how much it improves load performance?

sure I would do the benchmarking and add a option for distributedLoad

jja725:
@yuzhu @maobaolong add one option to support the old passive cache implementation

maobaolong (commented):
It look perfect now. Do you have a change to test this in a container env? Since I test the Load command cannot work in the container env after `load` command updated to use direct cache way, luckily I have fixed that issue.

yuzhu:
@jja725 could you update the documentation for distributedLoad to reflect the strategy choices?

jja725:
> @jja725 could you update the documentation for distributedLoad to reflect the strategy choices?

@yuzhu added doc, and I set the default to not use direct cache. So probably we can merge this in and try to test it in container env later after docker hub have the newest version(hard to test feature branch). After confirmation with test docker env, we can switch the default to use direct cache.

jja725:
@yuzhu @maobaolong tested in local docker env, it works fine. So I would set using direct cache as default then.

maobaolong (approved):
@jja725 Thanks for this contribution.

yuzhu (approved):

yuzhu:

PR #14535, opened 2021-11-17
Fix the wrong compution of cache percentage

ljl1988com:

Fix the wrong compution of cache percentage.

Fix the wrong compution of cache percentage.

None.


yuzhu (commented):
actually the original way of calculation is correct. 
bytesReadRemote includes both cache hit and cache miss 
Total number of bytes read from Alluxio storage "
              + "or underlying UFS if data does not exist in Alluxio storage "
              + "reported by all workers. This does not include "
              + "short-circuit local reads and domain socket reads")

ljl1988com:
@yuzhu I find when I load the data from ufs manually, `bytesReadRemote` is zero, but `bytesReadUfs` is not zero. `cacheHitRemotePercentage` may be negative.

yuzhu:
@ljl1988com at the very least, that is an inconsistency with the documentation.. maybe a bug?

ljl1988com:
@yuzhu It's a bug. I'll read the code logic of  metric `bytesReadRemote`, and find root cause. I'll fix it.

maobaolong (changes requested):
I guess the logic here is redundant with metric `CLUSTER_CACHE_HIT_RATE`, it would be nice if you correct and complete the logic in `MetricsStore#initMetricKeys` to support the metric `CLUSTER_BYTES_READ_TOTAL`,  `CLUSTER_LOCAL_CACHE_HIT_RATE`,`CLUSTER_REMOTE_CACHE_HIT_RATE`

ljl1988com:
I devide `bytesReadRemote` into `bytesReadRemote`  and `bytesReadRemoteHit`.
 `bytesReadRemote`  is cache miss, the number of bytes read from underlying UFS if data does not exist in Alluxio storage reported by all workers. This does not include short-circuit local reads and domain socket reads.
`bytesReadRemoteHit` is cache hit, the number of bytes read from Alluxio storage reported by all workers. This does not include short-circuit local reads and domain socket reads.

@yuzhu @maobaolong PTAL.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
```diff
@@ -427,6 +427,10 @@ private void runInternal() {
                 } else {
                   mResponse.onNext(response);
                 }
+                if (mContext.getBlockReader() == null) {
+                  mContext.setCounter(MetricsSystem.counter(
```
maobaolong, line 431: Where you update the value of this metric?

core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -1010,8 +1017,16 @@ public MetricKey build() {
           .build();
   public static final MetricKey WORKER_BYTES_READ_REMOTE =
       new Builder("Worker.BytesReadRemote")
+          .setDescription("Total number of bytes read from "
```
maobaolong, line 1020: It seems the meaning of this metrics has been changed after this PR? So would you like to give a new name of this metrics?

maobaolong (changes requested):
@ljl1988com Thanks for the update. It looks better, left some more comments, please take a look.

core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java
ljl1988com: Thanks for your review. 
Only when read from alluxio cache data, `mContext.getBlockReader()` will is null. 
Updating the value using common code as follows:
https://github.com/Alluxio/alluxio/blob/221f0ed5851417ffc6cae4758d112d490e25f930/core/server/worker/src/main/java/alluxio/worker/grpc/BlockReadHandler.java#L550-L568.

core/server/master/src/main/java/alluxio/master/metrics/MetricsStore.java
```diff
@@ -178,11 +178,14 @@ public void initMetricKeys() {
...
       // worker metrics
       mClusterCounters.putIfAbsent(new ClusterCounterKey(InstanceType.WORKER,
-              MetricKey.WORKER_BYTES_READ_DIRECT.getMetricName()),
+          MetricKey.WORKER_BYTES_READ_DIRECT.getMetricName()),
           MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_DIRECT.getName()));
       mClusterCounters.putIfAbsent(new ClusterCounterKey(InstanceType.WORKER,
           MetricKey.WORKER_BYTES_READ_REMOTE.getMetricName()),
           MetricsSystem.counter(MetricKey.CLUSTER_BYTES_READ_REMOTE.getName()));
+      mClusterCounters.putIfAbsent(new ClusterCounterKey(InstanceType.WORKER,
+                      MetricKey.WORKER_BYTES_READ_REMOTE_HIT.getMetricName()),
```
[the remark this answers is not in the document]
ljl1988com: Yes. I fix it.

ljl1988com:
@maobaolong PTAL, Thanks.

maobaolong (approved):
LGTM @ljl1988com Thanks for your contribution!

ljl1988com:
@yuzhu Can you take a another look?

yuzhu (approved):

jja725:
@ljl1988com Hi sorry for the late reply, but do you have some time to resolve the conflict so I can help merge this PR, thanks!

PR #14543, opened 2021-11-18
Make short-cirtcuit can fall back to grpc when path notfound

jhonxue:
LocalFileDataReader should check the path from the local worker,
            if the path can not be found, so falling back to the network transfer.
            The case may happen on the K8S, such as, client and worker in two diffrent containers

core/client:LocalFileDataReader Factory check the path from the local worker

                 LocalFileDataReader should check the path from the local worker,
                if the path can not be found, so can falling back to the network transfer.
                The case may happen on the K8S, such as, client and worker in two diffrent containers，
                client can not see the  worker data path.

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
@@ -151,7 +151,7 @@ public final class LocalFileDataReader implements DataReader {
         OpenLocalBlockResponse response = mStream.receive(mDataTimeoutMs);
         Preconditions.checkState(response.hasPath());
         mPath = response.getPath();
-        if (Files.notExists(Paths.get(mPath))) {
+        if (!Files.exists(Paths.get(mPath))) {
           throw new NotFoundException(String.format(
                   "LocalFileDataReader can not find the path:%s", mPath));
         }
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
@@ -16,6 +16,7 @@ import alluxio.client.ReadType;
 import alluxio.client.file.FileSystemContext;
 import alluxio.client.file.options.InStreamOptions;
 import alluxio.conf.PropertyKey;
+import alluxio.exception.status.NotFoundException;
 import alluxio.grpc.OpenLocalBlockRequest;
 import alluxio.grpc.OpenLocalBlockResponse;
 import alluxio.metrics.MetricKey;
@@ -31,6 +32,7 @@ import com.google.common.base.Preconditions;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
 import java.nio.file.Paths;
 
 import javax.annotation.concurrent.NotThreadSafe;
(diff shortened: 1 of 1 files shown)


jhonxue:
@apc999 @maobaolong @jiacheliu3 @LuQQiu PTAL

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
```diff
@@ -149,6 +151,10 @@ public Factory(FileSystemContext context, WorkerNetAddress address, long blockId
         OpenLocalBlockResponse response = mStream.receive(mDataTimeoutMs);
         Preconditions.checkState(response.hasPath());
         mPath = response.getPath();
+        if (Files.notExists(Paths.get(mPath))) {
```
maobaolong, line 154: Would you like to use `!Files.exists(Paths.get(mPath))`
```suggestion
        if (!Files.exists(Paths.get(mPath))) {
```

As the following method `notExists` return `false` when encountered IOException

```
    public static boolean notExists(Path path, LinkOption... options) {
        try {
            if (followLinks(options)) {
                provider(path).checkAccess(path);
            } else {
                // attempt to read attributes without following links
                readAttributes(path, BasicFileAttributes.class,
                               LinkOption.NOFOLLOW_LINKS);
            }
            // file exists
            return false;
        } catch (NoSuchFileException x) {
            // file confirmed not to exist
            return true;
        } catch (IOException x) {
            return false;
        }
    }
```

maobaolong (changes requested):
Nice catch, left a comment.

maobaolong (approved):
Thanks, LGTM.

maobaolong:
@jiacheliu3 Would you like to take another look?

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
```diff
@@ -149,6 +151,10 @@ public Factory(FileSystemContext context, WorkerNetAddress address, long blockId
         OpenLocalBlockResponse response = mStream.receive(mDataTimeoutMs);
         Preconditions.checkState(response.hasPath());
         mPath = response.getPath();
+        if (!Files.exists(Paths.get(mPath))) {
```
LuQQiu, line 154: Can you move the mPath exist check outside this try catch block.
This try block is try to catch anything throw when creating the GrpcBlockingStream or when sending messages, so that mBlockWorker can be close.

mBlockWorker =
try {
        mStream =
        mPath =
      } catch (Exception e) {
      mBlockWorker.close();
      throw e;
}

if (!Files.exists(Paths.get(mPath))) {
          throw new NotFoundException(String.format(
                  "LocalFileDataReader can not find the path:%s", mPath));
   BlockWorker.close();
  mStream.close();
)

LuQQiu (commented):
Good catch! Thanks for finding the issue and fixing it!
Left one comment about how to arrange the code

core/client/fs/src/main/java/alluxio/client/block/stream/LocalFileDataReader.java
jhonxue: Yes, thanks!

jhonxue: @LuQQiu I have modified the code. PTAL

jhonxue:
@apc999 @maobaolong @jiacheliu3 @LuQQiu PTAL

LuQQiu (approved):
LGTM, thanks for fixing it!

LuQQiu:

PR #14545, opened 2021-11-18
Refactor s3client creation

yyongycy:
Refactor the AmazonS3 client creation so that the code can be reused.

### DIFF
--- a/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
+++ b/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
@@ -208,8 +208,41 @@ public class S3AUnderFileSystem extends ObjectUnderFileSystem {
       clientConf.setSignerOverride(conf.get(PropertyKey.UNDERFS_S3_SIGNER_ALGORITHM));
     }
 
+    AwsClientBuilder.EndpointConfiguration endpointConfiguration
+        = createEndpointConfiguration(conf, clientConf);
+
+    AmazonS3 amazonS3Client
+        = createAmazonS3(credentials, clientConf, endpointConfiguration, conf);
+
+    ExecutorService service = ExecutorServiceFactories
+        .fixedThreadPool("alluxio-s3-transfer-manager-worker",
+            numTransferThreads).create();
+
+    TransferManager transferManager = TransferManagerBuilder.standard()
+        .withS3Client(createAmazonS3(credentials, clientConf, endpointConfiguration, conf))
+        .withExecutorFactory(() -> service)
+        .withMultipartCopyThreshold(MULTIPART_COPY_THRESHOLD)
+        .build();
+
+    return new S3AUnderFileSystem(uri, amazonS3Client, bucketName,
+        service, transferManager, conf, streamingUploadEnabled);
+  }
+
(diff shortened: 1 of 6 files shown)


yyongycy:
@LuQQiu  this is to handle : https://github.com/TachyonNexus/enterprise/pull/3289/files#r745264192

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -208,17 +207,49 @@ public static S3AUnderFileSystem createInstance(AlluxioURI uri,
       clientConf.setSignerOverride(conf.get(PropertyKey.UNDERFS_S3_SIGNER_ALGORITHM));
     }
 
-    AmazonS3ClientBuilder clientBuilder = AmazonS3Client.builder()
-        .withCredentials(credentials)
+    AwsClientBuilder.EndpointConfiguration endpointConfiguration
+        = createEndpointConfiguration(conf, clientConf);
```
LuQQiu, line 211: Do we expect creating different endpoint configuration for each Amazon S3 client?
why not put create endpoint configuration into createAmazonS3?

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
```diff
@@ -208,17 +207,49 @@ public static S3AUnderFileSystem createInstance(AlluxioURI uri,
...
+   * @param endpointConfiguration the endpoint config
+   * @param conf the Ufs config
+   * @return the AmazonS3 client
+   */
+  public static AmazonS3 createAmazonS3(AWSCredentialsProvider credentialsProvider,
+      ClientConfiguration clientConf,
+      AwsClientBuilder.EndpointConfiguration endpointConfiguration,
+      UnderFileSystemConfiguration conf) {
+    AmazonS3ClientBuilder clientBuilder = AmazonS3ClientBuilder
+        .standard()
```
LuQQiu, line 244: why add `.standard()`?
This line is not tested by me or @yuzhu so we are not quite sure whether it will break something or not.
S3 client with the S3 region/endpoint is kind of tricky.

LuQQiu (commented):
Thanks for refactoring, left two questions

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
yyongycy: It is my mistaken. It should be 100% same with before.

underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AUnderFileSystem.java
yyongycy: In the other scenarios, the client would be created multiple times, and not necessary to create endpoint multiple times. The purpose is trying to share some common objects.

LuQQiu (approved):
LGTM, thanks!

LuQQiu:
Will merge it after the test passed

LuQQiu:

Xenorith:

PR #14553, opened 2021-11-19
Feature/ha journal native fix

lilyzhoupeijie:
### What changes are proposed in this pull request?

set max concurrency for master notifyTermIndexUpdated,user can set alluxio.master.journal.log.concurrency.max=*** ,default is 256 to avoid error  oom:fail to create native thread

PR #14559, opened 2021-11-19
Change the thread pool for `CompletableFuture.runAsync`

jenoudet:
Redo of PR #14553 by @lilyzhoupeijie. 

Change the thread pool used by `CompletableFuture.runAsync` from the common pool to a separate `ForkJoinPool`. This is because of the concerns raised in [this article](https://dzone.com/articles/be-aware-of-forkjoinpoolcommonpool) about the common pool.

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -67,6 +67,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -106,6 +108,8 @@ public class JournalStateMachine extends BaseStateMachine {
   private volatile boolean mSnapshotting = false;
   private volatile boolean mIsLeader = false;
 
+  private final ExecutorService mJournalPool;
+
   /**
    * This callback is used for interrupting someone who suspends the journal applier to work on
    * the states. It helps prevent dirty read/write of the states when the journal is reloading.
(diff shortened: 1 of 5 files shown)


core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -138,8 +142,13 @@
    * @param journals      master journals; these journals are still owned by the caller, not by the
    *                      journal state machine
    * @param journalSystem the raft journal system
+   * @param maxConcurrencyPoolSize the maximum concurrency for notifyTermIndexUpdated
    */
-  public JournalStateMachine(Map<String, RaftJournal> journals, RaftJournalSystem journalSystem) {
+  public JournalStateMachine(Map<String, RaftJournal> journals, RaftJournalSystem journalSystem,
+                             Integer maxConcurrencyPoolSize) {
+    mJournalPool = Executors.newFixedThreadPool(maxConcurrencyPoolSize);
```
yuzhu, line 149: It is generally preferable to use the ThreadPoolExecutor constructor which gives a lot more control over resources consumed. see https://www.baeldung.com/java-executors-cached-fixed-threadpool In this case, let's use a bounded task queue with some reasonable minimum thread count.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -302,7 +311,7 @@ public void notifyNotLeader(Collection<TransactionContext> pendingEntries) {
   @Override
   public void notifyTermIndexUpdated(long term, long index) {
     super.notifyTermIndexUpdated(term, index);
-    CompletableFuture.runAsync(mJournalSystem::updateGroup);
+    CompletableFuture.runAsync(mJournalSystem::updateGroup, mJournalPool);
```
yuzhu, line 314: Just curious is the state of the excution not important? do we care if the updateGroup call failed?

yuzhu (commented):
@jenoudet overall looks good, couple of questions

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
jenoudet: I opted for a for 
```
ArrayBlockingQueue<Runnable> boundedQ = new ArrayBlockingQueue<>(maxConcurrencyPoolSize * 2);
mJournalPool = new ThreadPoolExecutor(maxConcurrencyPoolSize / 2, maxConcurrencyPoolSize, 0L,
        TimeUnit.MILLISECONDS, boundedQ);
```
The thread pool begins with half the max pool size, and can queue up to twice the max pool size.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
jenoudet: `mJournalSystem` falls back to the group information it held previously in case `updateGroup` fails. I think this is the best we can do if it fails, as we don't have a way of inferring group information.

yuzhu (approved):
LGTM

yuzhu:

Xenorith:

PR #14561, opened 2021-11-22
Support store client cache to multidisk

maobaolong:

In some of our scenarios, we have to use multi-disk to be our cache storage.
- There are no enough space if we just store cache data into one disk
- There are no enough bandwidth if we just store cache data into one disk.

Store the cache data into multi-disk by hash & mod

User can config `alluxio.user.client.cache.dir=rootdir1,rootdir2,rootdir3`.

Meanwhile, this change support keep consistent to the original usage `alluxio.user.client.cache.dir=rootdir1`

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
@@ -29,6 +29,7 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
@@ -43,7 +44,7 @@ import javax.annotation.concurrent.NotThreadSafe;
 public class LocalPageStore implements PageStore {
   private static final Logger LOG = LoggerFactory.getLogger(LocalPageStore.class);
   private static final String ERROR_NO_SPACE_LEFT = "No space left on device";
-  private final String mRoot;
+  private final List<Path> mRoots;
   private final long mPageSize;
   private final long mCapacity;
   private final int mFileBuckets;
@@ -55,15 +56,16 @@ public class LocalPageStore implements PageStore {
    * @param options options for the local page store
    */
   public LocalPageStore(LocalPageStoreOptions options) {
-    mRoot = options.getRootDir();
+    mRoots = options.getRootDirs();
     mPageSize = options.getPageSize();
     mCapacity = (long) (options.getCacheSize() / (1 + options.getOverheadRatio()));
     mFileBuckets = options.getFileBuckets();
-    // normalize the path to deal with trailing slash
-    Path rootDir = Paths.get(mRoot);
-    // pattern encoding root_path/page_size(ulong)/bucket(uint)/file_id(str)/page_idx(ulong)/
+    // pattern encoding root_path/page_size(ulong)/bucket(uint)/file_id(str)/page_idx(ulong)
     mPagePattern = Pattern.compile(
-        String.format("%s/%d/(\\d+)/([^/]+)/(\\d+)", Pattern.quote(rootDir.toString()), mPageSize));
+        String.format("%s/%d/(\\d+)/([^/]+)/(\\d+)", "("
+                + mRoots.stream().map(path -> path.toString()).reduce((a, b) -> a + "|" + b).get()
+                + ")",
+            mPageSize));
   }
 
   @Override
@@ -84,7 +86,7 @@ public class LocalPageStore implements PageStore {
       Files.deleteIfExists(p);
       if (e.getMessage().contains(ERROR_NO_SPACE_LEFT)) {
         throw new ResourceExhaustedException(
-            String.format("%s is full, configured with %d bytes", mRoot, mCapacity), e);
+            String.format("%s is full, configured with %d bytes", getRoot(pageId), mCapacity), e);
       }
       throw new IOException("Failed to write file " + p + " for page " + pageId);
     }
@@ -159,8 +161,16 @@ public class LocalPageStore implements PageStore {
   @VisibleForTesting
   public Path getFilePath(PageId pageId) {
     // TODO(feng): encode fileId with URLEncoder to escape invalid characters for file name
-    return Paths.get(mRoot, Long.toString(mPageSize), getFileBucket(pageId.getFileId()),
-        pageId.getFileId(), Long.toString(pageId.getPageIndex()));
+    return Paths.get(getRoot(pageId).toString(), Long.toString(mPageSize),
+        getFileBucket(pageId.getFileId()), pageId.getFileId(),
+        Long.toString(pageId.getPageIndex()));
+  }
+
+  private Path getRoot(PageId pageId) {
+    // TODO(maobaolong): Refactor to support choose volume policy
+    int index = pageId.hashCode() % mRoots.size();
+    index = index < 0 ? index + mRoots.size() : index;
+    return mRoots.get(index);
   }
 
   private String getFileBucket(String fileId) {
@@ -178,12 +188,12 @@ public class LocalPageStore implements PageStore {
       return null;
     }
     try {
-      String fileBucket = Preconditions.checkNotNull(matcher.group(1));
-      String fileId = Preconditions.checkNotNull(matcher.group(2));
+      String fileBucket = Preconditions.checkNotNull(matcher.group(2));
+      String fileId = Preconditions.checkNotNull(matcher.group(3));
       if (!fileBucket.equals(getFileBucket(fileId))) {
         return null;
       }
-      String fileName = Preconditions.checkNotNull(matcher.group(3));
+      String fileName = Preconditions.checkNotNull(matcher.group(4));
       long pageIndex = Long.parseLong(fileName);
       return new PageId(fileId, pageIndex);
     } catch (NumberFormatException e) {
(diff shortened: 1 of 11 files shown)


beinan (commented):
Thank you @maobaolong for this contribution!  Add some unit test to cover this feature, could you?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -159,10 +157,16 @@ public void delete(PageId pageId) throws IOException, PageNotFoundException {
...
   public Path getFilePath(PageId pageId) {
     // TODO(feng): encode fileId with URLEncoder to escape invalid characters for file name
-    return Paths.get(mRoot, Long.toString(mPageSize), getFileBucket(pageId.getFileId()),
+    return Paths.get(getRoot(pageId), Long.toString(mPageSize), getFileBucket(pageId.getFileId()),
         pageId.getFileId(), Long.toString(pageId.getPageIndex()));
   }
 
+  private String getRoot(PageId pageId) {
+    int index = pageId.hashCode() % mRoots.length;
+    index = index < 0 ? index + mRoots.length : index;
```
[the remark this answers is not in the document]
maobaolong: We can refactor and support round many choose disk policy, like robin, storageUsageBased or IOBusyBased.

Would you mind to let me implements the abstract refactor in next PR as follow-up task?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -55,15 +55,13 @@
...
+    mRoots = options.getRootDir().split(",");
     mPageSize = options.getPageSize();
     mCapacity = (long) (options.getCacheSize() / (1 + options.getOverheadRatio()));
     mFileBuckets = options.getFileBuckets();
-    // normalize the path to deal with trailing slash
-    Path rootDir = Paths.get(mRoot);
     // pattern encoding root_path/page_size(ulong)/bucket(uint)/file_id(str)/page_idx(ulong)/
     mPagePattern = Pattern.compile(
-        String.format("%s/%d/(\\d+)/([^/]+)/(\\d+)", Pattern.quote(rootDir.toString()), mPageSize));
+        String.format("%s/%d/(\\d+)/([^/]+)/(\\d+)", ".*", mPageSize));
```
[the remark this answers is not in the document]
maobaolong: Change the pattern to `(rootdir1|rootdir2)/1024/(\\d+)/([^/]+)/(\\d+)`

maobaolong:
@beinan Thanks for the review, changed the regex pattern updated the related UTs, please take another look

maobaolong:
The failed test is not caused by this PR, @beinan please take a look at this PR, thank you.

maobaolong:
Hi @apc999 would you please take a look at this PR, thank you.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -159,8 +161,16 @@ public void delete(PageId pageId) throws IOException, PageNotFoundException {
...
-    return Paths.get(mRoot, Long.toString(mPageSize), getFileBucket(pageId.getFileId()),
-        pageId.getFileId(), Long.toString(pageId.getPageIndex()));
+    return Paths.get(getRoot(pageId).toString(), Long.toString(mPageSize),
+        getFileBucket(pageId.getFileId()), pageId.getFileId(),
+        Long.toString(pageId.getPageIndex()));
+  }
+
+  private Path getRoot(PageId pageId) {
+    // TODO(maobaolong): Refactor to support choose volume policy
+    int index = pageId.hashCode() % mRoots.size();
```
beinan, line 171: I just have a question,  what if we add a new root path?  looks like we're still loading all the pages from the existing root paths, but the index here would be changed,  am I right?
say we already have page1 and page2 in root1, then we add root2, in this case, page2's new index would be on root2, but physically it's still on root1. Would this be an issue?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
```diff
@@ -219,8 +229,11 @@ public void close() {
 
   @Override
   public Stream<PageInfo> getPages() throws IOException {
-    Path rootDir = Paths.get(mRoot);
-    return Files.walk(rootDir).filter(Files::isRegularFile).map(this::getPageInfo);
+    Stream<Path> stream = Stream.empty();
+    for (Path root : mRoots) {
+      stream = Stream.concat(stream, Files.walk(root));
+    }
+    return stream.filter(Files::isRegularFile).map(this::getPageInfo);
```
beinan, line 236: would it be more efficient if we could do filter and then do the concat?  I haven't look into the implementation or java's stream, so I'm just guessing

core/client/fs/src/main/java/alluxio/client/file/cache/store/PageStoreOptions.java
```diff
@@ -107,19 +109,19 @@ public static PageStoreOptions create(AlluxioConfiguration conf) {
   protected double mOverheadRatio;
 
   /**
-   * @param rootDir the root directory where pages are stored
+   * @param rootDirs the root directory where pages are stored
    * @return the updated options
    */
-  public PageStoreOptions setRootDir(String rootDir) {
-    mRootDir = rootDir;
+  public PageStoreOptions setRootDir(List<Path> rootDirs) {
```
beinan, line 115: setRootDirs?

core/client/fs/src/main/java/alluxio/client/file/cache/store/PageStoreOptions.java
```diff
@@ -107,19 +109,19 @@ public static PageStoreOptions create(AlluxioConfiguration conf) {
...
+    mRootDirs = rootDirs;
     return this;
   }
 
   /**
    * @return the root directory where pages are stored
    */
-  public String getRootDir() {
-    return mRootDir;
+  public List<Path> getRootDir() {
```
beinan, line 123: getRootDirs?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
maobaolong: Yeah, it would be an issue if we add a new root for the existing roots for now.

So i decided to do some improvement base on this PR, i can add more choose page policies, something like consistent hash policy, or supply a start rebalance pages during add or remove a root directory.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
maobaolong: will take a look

apc999:
ping @maobaolong , any update on this PR?

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
maobaolong:
```java
    /**
     * Creates a lazily concatenated stream whose elements are all the
     * elements of the first stream followed by all the elements of the
     * second stream.  The resulting stream is ordered if both
     * of the input streams are ordered, and parallel if either of the input
     * streams is parallel.  When the resulting stream is closed, the close
     * handlers for both input streams are invoked.
     *
     * @implNote
     * Use caution when constructing streams from repeated concatenation.
     * Accessing an element of a deeply concatenated stream can result in deep
     * call chains, or even {@code StackOverflowException}.
     *
     * @param <T> The type of stream elements
     * @param a the first stream
     * @param b the second stream
     * @return the concatenation of the two input streams
     */
    public static <T> Stream<T> concat(Stream<? extends T> a, Stream<? extends T> b) {
        Objects.requireNonNull(a);
        Objects.requireNonNull(b);

        @SuppressWarnings("unchecked")
        Spliterator<T> split = new Streams.ConcatSpliterator.OfRef<>(
                (Spliterator<T>) a.spliterator(), (Spliterator<T>) b.spliterator());
        Stream<T> stream = StreamSupport.stream(split, a.isParallel() || b.isParallel());
        return stream.onClose(Streams.composedClose(a, b));
    }
```

It seems no efficient down for `concat` operation.

maobaolong:
@apc999 Thanks for remind me! @beinan I've addressed the comments, PTAL.

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
beinan: I see, looks like the concat is lazy, which should be just a O(1) operation.   Thank you Baolong for the clarification!

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
beinan: I'm good to have another PR for this, thanks!

beinan (approved):
Thank you Baolong!  I think this pr is good starting point for multidisk support, I might recommend this one as an "experimental feature" to our users because there would be some storage leaking issue when adding or removing the disks.  I will suggest the user to clean up the cache folder if they want to add or remove a new disk

beinan:
@apc999 do you want to take a look?  thanks!

core/client/fs/src/main/java/alluxio/client/file/cache/store/LocalPageStore.java
zzcclp: @maobaolong , any updates on this PR, or will raise another PR to fix this issue?

zzcclp:
@maobaolong please rebase this pr, #14772 has been merged, there are some conflicts that need to be fixed.

beinan (approved):
thank you @maobaolong ! lgtm,  rerun the test

maobaolong:
@beinan The test has been passed, thank for trigger re-run test for me!

beinan:

PR #14565, opened 2021-11-22
Configure block and inode metastore separately

bzheng888:
Separate configuration for block and inode is make sense, because in some cases, blockstore use ROCKS is enough, but inodestore must use HEAP, we want to distinguish the difference between these two types.

Sometimes we want to config inode metastore type and block metastore type seperately.

  1. change in user-facing APIs
  Some configuration changes

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2298,40 +2298,37 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
-  /**
-   * @deprecated use {@link #MASTER_INODE_METASTORE} and {@link #MASTER_BLOCK_METASTORE} instead
-   */
-  @Deprecated
   public static final PropertyKey MASTER_METASTORE =
       enumBuilder(Name.MASTER_METASTORE, MetastoreType.class)
           .setDefaultValue(MetastoreType.ROCKS)
-          .setDescription("The type of metastore to use, either HEAP or ROCKS. If this property is "
-              + "set, the " + Name.MASTER_INODE_METASTORE + " " + Name.MASTER_BLOCK_METASTORE + " "
-              + "property will be ignored. The heap metastore keeps all metadata on-heap, "
+          .setDescription("The type of metastore to use, either HEAP or ROCKS. "
+              + "The heap metastore keeps all metadata on-heap, "
               + "while the rocks metastore stores some metadata on heap and some metadata on disk. "
               + "The rocks metastore has the advantage of being able to support a large namespace "
-              + "(1 billion plus files) without needing a massive heap size.")
+              + "(1 billion plus files) without needing a massive heap size."
+              + "The metadata storage includes inode and block metadata. "
+              + "Users can override the type of metastore using " + Name.MASTER_INODE_METASTORE
+              + " and " + Name.MASTER_BLOCK_METASTORE
+              + ". For example if " + Name.MASTER_METASTORE + "=ROCKS but "
+              + Name.MASTER_INODE_METASTORE + "=HEAP, then inodes are stored with HEAP and blocks"
+              + " are stored with ROCKS."
+          )
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
   public static final PropertyKey MASTER_INODE_METASTORE =
       enumBuilder(Name.MASTER_INODE_METASTORE, MetastoreType.class)
-          .setDefaultValue(String.format("${%s}", PropertyKey.Name.MASTER_METASTORE))
-          .setDescription("The type of inode metastore to use, either HEAP or ROCKS. The heap "
-              + "metastore keeps all metadata on-heap, while the rocks metastore stores some "
-              + "metadata on heap and some metadata on disk. The rocks metastore has the advantage "
-              + "of being able to support a large namespace (1 billion plus files) without needing"
-              + "a massive heap size.")
+          .setDefaultValue(MASTER_METASTORE.getDefaultValue())
+          .setDescription("The type of inode metastore to use, either HEAP or ROCKS. "
+              + "By default this uses " + PropertyKey.Name.MASTER_METASTORE.toString() + ".")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
   public static final PropertyKey MASTER_BLOCK_METASTORE =
       enumBuilder(Name.MASTER_BLOCK_METASTORE, MetastoreType.class)
-          .setDefaultValue(String.format("${%s}", PropertyKey.Name.MASTER_METASTORE))
-          .setDescription("The type of metastore for block, either HEAP or ROCKS. Separate "
-              + "configuration for block and inodes is make sense, because in some cases, "
-              + "block use ROCKS is enough, but we prefer to use HEAP for inode, we want to "
-              + "distinguish the difference between these two types. ")
+          .setDefaultValue(MASTER_METASTORE.getDefaultValue())
+          .setDescription("The type of block metastore to use, either HEAP or ROCKS. "
+              + "By default this uses " + PropertyKey.Name.MASTER_METASTORE.toString() + ".")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2298,14 +2298,40 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  /**
+   * @deprecated use {@link #MASTER_INODE_METASTORE} and {@link #MASTER_BLOCK_METASTORE} instead
+   */
+  @Deprecated
   public static final PropertyKey MASTER_METASTORE =
       enumBuilder(Name.MASTER_METASTORE, MetastoreType.class)
           .setDefaultValue(MetastoreType.ROCKS)
-          .setDescription("The type of metastore to use, either HEAP or ROCKS. The heap metastore "
-              + "keeps all metadata on-heap, while the rocks metastore stores some metadata on "
-              + "heap and some metadata on disk. The rocks metastore has the advantage of being "
-              + "able to support a large namespace (1 billion plus files) without needing a "
-              + "massive heap size.")
+          .setDescription("The type of metastore to use, either HEAP or ROCKS. If this property is "
+              + "set, the " + Name.MASTER_INODE_METASTORE + " " + Name.MASTER_BLOCK_METASTORE + " "
+              + "property will be ignored. The heap metastore keeps all metadata on-heap, "
+              + "while the rocks metastore stores some metadata on heap and some metadata on disk. "
+              + "The rocks metastore has the advantage of being able to support a large namespace "
+              + "(1 billion plus files) without needing a massive heap size.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_INODE_METASTORE =
+      enumBuilder(Name.MASTER_INODE_METASTORE, MetastoreType.class)
+          .setDefaultValue(String.format("${%s}", PropertyKey.Name.MASTER_METASTORE))
+          .setDescription("The type of inode metastore to use, either HEAP or ROCKS. The heap "
+              + "metastore keeps all metadata on-heap, while the rocks metastore stores some "
+              + "metadata on heap and some metadata on disk. The rocks metastore has the advantage "
+              + "of being able to support a large namespace (1 billion plus files) without needing"
+              + "a massive heap size.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_BLOCK_METASTORE =
+      enumBuilder(Name.MASTER_BLOCK_METASTORE, MetastoreType.class)
+          .setDefaultValue(String.format("${%s}", PropertyKey.Name.MASTER_METASTORE))
+          .setDescription("The type of metastore for block, either HEAP or ROCKS. Separate "
+              + "configuration for block and inodes is make sense, because in some cases, "
+              + "block use ROCKS is enough, but we prefer to use HEAP for inode, we want to "
+              + "distinguish the difference between these two types. ")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
@@ -7535,6 +7561,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     public static final String MASTER_METADATA_SYNC_UFS_PREFETCH_TIMEOUT =
         "alluxio.master.metadata.sync.ufs.prefetch.timeout";
     public static final String MASTER_METASTORE = "alluxio.master.metastore";
+    public static final String MASTER_INODE_METASTORE = "alluxio.master.inode.metastore";
+    public static final String MASTER_BLOCK_METASTORE = "alluxio.master.block.metastore";
     public static final String MASTER_METASTORE_DIR = "alluxio.master.metastore.dir";
     public static final String MASTER_METASTORE_ROCKS_PARALLEL_BACKUP =
         "alluxio.master.metastore.rocks.parallel.backup";
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -7558,8 +7558,8 @@ public final class PropertyKey implements Comparable<PropertyKey> {
     public static final String MASTER_METADATA_SYNC_UFS_PREFETCH_TIMEOUT =
         "alluxio.master.metadata.sync.ufs.prefetch.timeout";
     public static final String MASTER_METASTORE = "alluxio.master.metastore";
-    public static final String MASTER_INODE_METASTORE = "alluxio.master.inode.metastore";
-    public static final String MASTER_BLOCK_METASTORE = "alluxio.master.block.metastore";
+    public static final String MASTER_INODE_METASTORE = "alluxio.master.metastore.inode";
+    public static final String MASTER_BLOCK_METASTORE = "alluxio.master.metastore.block";
     public static final String MASTER_METASTORE_DIR = "alluxio.master.metastore.dir";
     public static final String MASTER_METASTORE_ROCKS_PARALLEL_BACKUP =
         "alluxio.master.metastore.rocks.parallel.backup";
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -2307,17 +2307,17 @@ public final class PropertyKey implements Comparable<PropertyKey> {
               + "The rocks metastore has the advantage of being able to support a large namespace "
               + "(1 billion plus files) without needing a massive heap size."
               + "The metadata storage includes inode and block metadata. "
-              + "Users can override the type of metastore using " + Name.MASTER_INODE_METASTORE
-              + " and " + Name.MASTER_BLOCK_METASTORE
+              + "Users can override the type of metastore using " + Name.MASTER_METASTORE_INODE
+              + " and " + Name.MASTER_METASTORE_BLOCK
               + ". For example if " + Name.MASTER_METASTORE + "=ROCKS but "
-              + Name.MASTER_INODE_METASTORE + "=HEAP, then inodes are stored with HEAP and blocks"
+              + Name.MASTER_METASTORE_INODE + "=HEAP, then inodes are stored with HEAP and blocks"
               + " are stored with ROCKS."
           )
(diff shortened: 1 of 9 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1941,6 +1941,16 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.ENFORCE)
           .setScope(Scope.MASTER)
           .build();
+  public static final PropertyKey MASTER_METASTORE_BLOCK_USE =
```
maobaolong, line 1944:
```suggestion
  public static final PropertyKey MASTER_BLOCK_METASTORE =
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1941,6 +1941,16 @@ public String toString() {
...
+      new Builder(Name.MASTER_METASTORE_BLOCK_USE)
+          .setDefaultValue("ROCKS")
+          .setDescription("The type of metastore for block, either HEAP or ROCKS. Separate "
```
maobaolong, line 1947: Would you like to update the `MASTER_METASTORE` besides?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -6263,6 +6273,7 @@ private static String javadocLink(String fullyQualifiedClassname) {
     public static final String MASTER_METADATA_SYNC_UFS_PREFETCH_POOL_SIZE =
         "alluxio.master.metadata.sync.ufs.prefetch.pool.size";
     public static final String MASTER_METASTORE = "alluxio.master.metastore";
+    public static final String MASTER_METASTORE_BLOCK_USE = "alluxio.master.metastore.block";
```
maobaolong, line 6276:
```suggestion
    public static final String MASTER_BLOCK_METASTORE = "alluxio.master.block.metastore";
```

core/server/master/src/main/java/alluxio/master/MasterUtils.java
```diff
@@ -64,7 +64,7 @@ public static void createMasters(MasterRegistry registry, MasterContext context)
    */
   public static BlockStore.Factory getBlockStoreFactory(String baseDir) {
     MetastoreType type =
-        ServerConfiguration.getEnum(PropertyKey.MASTER_METASTORE, MetastoreType.class);
+        ServerConfiguration.getEnum(PropertyKey.MASTER_METASTORE_BLOCK_USE, MetastoreType.class);
```
maobaolong, line 67:
```suggestion
        ServerConfiguration.getEnum(PropertyKey.MASTER_BLOCK_METASTORE, MetastoreType.class);
```

maobaolong (changes requested):
@bzheng888 Thanks for your contribution, left some comments and suggestion commit inline.

BTW, you can rename `MASTER_METASTORE` to `MASTER_INODE_METASTORE` meanwhile, add an alias name to keep backward compatible and deprecate the old name.

bzheng888:
@maobaolong Done, PTAL!

maobaolong (changes requested):
It would be better if you KEEP `MASTER_METASTORE ` propertykey and if it `setByUser`, the original logic should honor the configured value and ignored the `MASTER_BLOCK_METASTORE ` and  `MASTER_INODE_METASTORE ` config

bzheng888:
> It would be better if you KEEP `MASTER_METASTORE ` propertykey and if it `setByUser`, the original logic should honor the configured value and ignored the `MASTER_BLOCK_METASTORE ` and `MASTER_INODE_METASTORE ` config

Thanks for pointing this, update this pr, PTAL!

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1933,11 +1933,33 @@ public String toString() {
   public static final PropertyKey MASTER_METASTORE =
       new Builder(Name.MASTER_METASTORE)
           .setDefaultValue("ROCKS")
-          .setDescription("The type of metastore to use, either HEAP or ROCKS. The heap metastore "
-              + "keeps all metadata on-heap, while the rocks metastore stores some metadata on "
-              + "heap and some metadata on disk. The rocks metastore has the advantage of being "
-              + "able to support a large namespace (1 billion plus files) without needing a "
-              + "massive heap size.")
+          .setDescription("The type of metastore to use, either HEAP or ROCKS. If this property "
+              + "is set, the MASTER_INODE_METASTORE MASTER_BLOCK_METASTORE property will be "
```
maobaolong, line 1937:
```suggestion
              + "is set, the " + Name.MASTER_INODE_METASTORE + " " + Name.MASTER_BLOCK_METASTORE + " property will be "
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1933,11 +1933,33 @@ public String toString() {
   public static final PropertyKey MASTER_METASTORE =
```
maobaolong, line 1933: Please add an annotation here to mark it deprecated.

tests/src/test/java/alluxio/server/ft/journal/MultiProcessCheckpointTest.java
```diff
@@ -38,6 +38,7 @@ public void test() throws Exception {
         .addProperty(PropertyKey.ZOOKEEPER_SESSION_TIMEOUT, "2sec")
         .addProperty(PropertyKey.ZOOKEEPER_CONNECTION_TIMEOUT, "1sec")
         .addProperty(PropertyKey.MASTER_METASTORE, "ROCKS")
+        .addProperty(PropertyKey.MASTER_METASTORE, "ROCKS")
```
maobaolong, line 41: Look here

maobaolong (changes requested):
@bzheng888 Thanks for this updated, it looks better, left some minor comment.

maobaolong (approved):
LGTM Thanks @bzheng888

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -1930,14 +1930,40 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  /**
+   * @deprecated use {@link #MASTER_INODE_METASTORE #MASTER_BLOCK_METASTORE} instead
```
maobaolong, line 1934:
```suggestion
   * @deprecated use {@link #MASTER_INODE_METASTORE} and {@link #MASTER_BLOCK_METASTORE} instead
```

bzheng888: @maobaolong Done, thanks for this suggestion!

bzheng888:
Hi, @jiacheliu3, can you take some time to review this pr?

jiacheliu3:
@bzheng888 @maobaolong Can you please clarify on the use case? I'm not following. This is a very big change.

FYI: @apc999 @yuzhu @ggezer

bzheng888:
@jiacheliu3 Thank you for your reply, the main idea for this pr is to config master inode metastore and block metastore separately. Talking about cases, if our case are big files, and the amount of inode metadata is not that much, we can can config inode metastore as heap, but the block metadata will be very large, it's better to use rocks metastore to support large namespace.

bzheng888:
@maobaolong @maobaolong Update this pr, PTAL!

maobaolong:
@bzheng888 Will take a quick look soon

maobaolong (approved):
It LGTM, Thanks @bzheng888 for your contribution and patient improvement.

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
```diff
@@ -35,7 +35,11 @@ public static boolean isEmbeddedJournal() {
    * @return true, if running with rocks
    */
   public static boolean isRocks() {
-    return Configuration.get(PropertyKey.MASTER_METASTORE) == MetastoreType.ROCKS;
+    if (Configuration.isSetByUser(PropertyKey.MASTER_METASTORE)) {
+      return Configuration.get(PropertyKey.MASTER_METASTORE) == MetastoreType.ROCKS;
+    }
+    return Configuration.get(PropertyKey.MASTER_INODE_METASTORE) == MetastoreType.ROCKS
+        &&  Configuration.get(PropertyKey.MASTER_BLOCK_METASTORE) == MetastoreType.ROCKS;
```
jiacheliu3, line 42: This part is actually quite interesting. Does "Inode-on-rocks + block-on-heap" count as rocks or heap? @elega @dbw9580 WDYT?

The good news is this check is only used in `UpdateCheck` which is a minor feature in Alluxio.

tests/src/test/java/alluxio/client/cli/JournalToolTest.java
```diff
@@ -143,7 +143,8 @@ public void dumpHeapCheckpointFromUfsJournal() throws Throwable {
 
   @Test
   @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_JOURNAL_TYPE,
-      "EMBEDDED", PropertyKey.Name.MASTER_METASTORE, "HEAP"})
+      "EMBEDDED", PropertyKey.Name.MASTER_INODE_METASTORE, "HEAP",
+      PropertyKey.Name.MASTER_BLOCK_METASTORE, "HEAP"})
```
jiacheliu3, line 147: Please add tests for BLOCK_METASTORE=ROCKS and INODE_METASTORE=HEAP and vice versa. Let's make sure all combinations work.

tests/src/test/java/alluxio/client/cli/JournalToolTest.java
```diff
@@ -186,7 +187,8 @@ public void dumpHeapCheckpointFromEmbeddedJournal() throws Throwable {
   }
 
   @Test
-  @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_METASTORE, "ROCKS"})
+  @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_INODE_METASTORE,
+      "ROCKS", PropertyKey.Name.MASTER_BLOCK_METASTORE, "ROCKS"})
```
jiacheliu3, line 191: Same as above

tests/src/test/java/alluxio/client/fs/FileSystemMasterRestartIntegrationTest.java
```diff
@@ -335,7 +335,8 @@ public void ufsModeReplay() throws Exception {
    * Tests journal is updated with access time asynchronously before master is stopped.
    */
   @Test
-  @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_METASTORE, "HEAP"})
+  @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_INODE_METASTORE, "HEAP",
+      PropertyKey.Name.MASTER_BLOCK_METASTORE, "HEAP"})
```
jiacheliu3, line 339: Same as above

tests/src/test/java/alluxio/client/fs/FileSystemMasterRestartIntegrationTest.java
```diff
@@ -369,7 +370,8 @@ public void updateAccessTimeAsyncFlush() throws Exception {
    * Tests journal is not updated with access time asynchronously after delete.
    */
   @Test
-  @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_METASTORE, "HEAP"})
+  @LocalAlluxioClusterResource.Config(confParams = {PropertyKey.Name.MASTER_INODE_METASTORE, "HEAP",
+      PropertyKey.Name.MASTER_BLOCK_METASTORE, "HEAP"})
```
jiacheliu3, line 374: Same as above

jiacheliu3 (changes requested):
Mostly LGTM with some comments. Thanks for the improvement!

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2250,14 +2250,40 @@ public String toString() {
...
+          .setScope(Scope.MASTER)
+          .build();
+  public static final PropertyKey MASTER_INODE_METASTORE =
+      enumBuilder(Name.MASTER_INODE_METASTORE, MetastoreType.class)
+          .setDefaultValue(String.format("${%s}", PropertyKey.Name.MASTER_METASTORE))
+          .setDescription("The type of inode metastore to use, either HEAP or ROCKS. The heap "
+              + "metastore keeps all metadata on-heap, while the rocks metastore stores some "
+              + "metadata on heap and some metadata on disk. The rocks metastore has the advantage "
+              + "of being able to support a large namespace (1 billion plus files) without needing"
+              + "a massive heap size.")
```
jiacheliu3, line 2276:
```suggestion
          .setDescription("The type of inode metastore to use, either HEAP or ROCKS. 
              + "By default this uses " + PropertyKey.Name.MASTER_METASTORE.toString() + ".");
```

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2250,14 +2250,40 @@ public String toString() {
...
-              + "keeps all metadata on-heap, while the rocks metastore stores some metadata on "
-              + "heap and some metadata on disk. The rocks metastore has the advantage of being "
-              + "able to support a large namespace (1 billion plus files) without needing a "
-              + "massive heap size.")
+          .setDescription("The type of metastore to use, either HEAP or ROCKS. If this property is "
+              + "set, the " + Name.MASTER_INODE_METASTORE + " " + Name.MASTER_BLOCK_METASTORE + " "
+              + "property will be ignored. The heap metastore keeps all metadata on-heap, "
+              + "while the rocks metastore stores some metadata on heap and some metadata on disk. "
+              + "The rocks metastore has the advantage of being able to support a large namespace "
+              + "(1 billion plus files) without needing a massive heap size.")
```
jiacheliu3, line 2265:
```suggestion
          .setDescription("The type of metastore to use, either HEAP or ROCKS. "
              + "The heap metastore keeps all metadata on-heap, "
              + "while the rocks metastore stores some metadata on heap and some metadata on disk. "
              + "The rocks metastore has the advantage of being able to support a large namespace "
              + "(1 billion plus files) without needing a massive heap size."
              + "The metadata storage includes inode and block metadata. "
              + "Users can override the type of metastore using " + Name.MASTER_INODE_METASTORE
              + " and " + Name.MASTER_BLOCK_METASTORE
              + ". For example if " + Name.MASTER_METASTORE + "=ROCKS but "
              + Name.MASTER_INODE_METASTORE + "=HEAP, then inodes are stored with HEAP and blocks are stored with ROCKS."
              )
```

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
elega: https://github.com/Alluxio/alluxio/blob/master/core/server/master/src/main/java/alluxio/master/meta/UpdateCheck.java#L124-L136
Looking at how this method is used, I feel like the semantics of "isRocks" is closer to "isRocksUsed". The result of the method will be attached in the rpc header as a user agent for update check. So I think it's an OR condition instead of AND.

core/server/master/src/main/java/alluxio/master/MasterUtils.java
```diff
@@ -62,8 +62,9 @@ public static void createMasters(MasterRegistry registry, MasterContext context)
    * @return a block store factory of the configured type
    */
   public static BlockMetaStore.Factory getBlockStoreFactory(String baseDir) {
-    MetastoreType type =
-        Configuration.getEnum(PropertyKey.MASTER_METASTORE, MetastoreType.class);
+    MetastoreType type = Configuration.isSetByUser(PropertyKey.MASTER_METASTORE)
```
elega, line 65: this and the one below should be converted to util functions IMO

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -2250,14 +2250,40 @@ public String toString() {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.MASTER)
           .build();
+  /**
+   * @deprecated use {@link #MASTER_INODE_METASTORE} and {@link #MASTER_BLOCK_METASTORE} instead
+   */
+  @Deprecated
```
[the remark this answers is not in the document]
elega: I actually have a random idea. Can we add an alias to the property: alluxio.master.inode.metastore and introduce a new one for block: alluxio.master.block.metastore. So that the user can
1. just specify alluxio.master.metastore for both
2. specify alluxio.master.inode.metastore AND alluxio.master.block.metastore

@jiacheliu3 wdty. @maobaolong please ignore this comment for now. Not sure this is a better way or not. :)

tests/src/test/java/alluxio/client/cli/JournalToolTest.java
bzheng888: Ok, thanks for reviewing this!

core/common/src/main/java/alluxio/conf/PropertyKey.java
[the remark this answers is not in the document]
bzheng888: +1

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
bzheng888: How about separate the method to isInodeRocks() and isBlockRocks()?

bzheng888:
@jiacheliu3 @elega Update, PTAL!

tests/src/test/java/alluxio/client/cli/JournalToolTest.java
bzheng888: This test seems just test the Journal tool and only inode metastore is considered, this may be enough and no need to add new test? FYI.

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -7535,6 +7558,8 @@ public static final class Name {
     public static final String MASTER_METADATA_SYNC_UFS_PREFETCH_TIMEOUT =
         "alluxio.master.metadata.sync.ufs.prefetch.timeout";
     public static final String MASTER_METASTORE = "alluxio.master.metastore";
+    public static final String MASTER_INODE_METASTORE = "alluxio.master.inode.metastore";
+    public static final String MASTER_BLOCK_METASTORE = "alluxio.master.block.metastore";
```
jiacheliu3, line 7562: @elega Do you prefer `master.inode.metastore` or `master.metastore.inode`?

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
```diff
@@ -31,11 +31,26 @@ public static boolean isEmbeddedJournal() {
   }
 
   /**
-   * Utility to check the master metastore is rocks.
+   * Utility to check the master inode metastore is rocks.
    *
    * @return true, if running with rocks
    */
-  public static boolean isRocks() {
+  public static boolean isInodeRocks() {
```
jiacheliu3, line 38:
```suggestion
  public static boolean isInodeStoreRocks() {
```

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
```diff
@@ -31,11 +31,26 @@ public static boolean isEmbeddedJournal() {
...
+    }
+    return Configuration.get(PropertyKey.MASTER_METASTORE) == MetastoreType.ROCKS;
+  }
+
+  /**
+   * Utility to check the master block metastore is rocks.
+   *
+   * @return true, if running with rocks
+   */
+  public static boolean isBlockRocks() {
```
jiacheliu3, line 50:
```suggestion
  public static boolean isBlockStoreRocks() {
```

jiacheliu3 (changes requested):
Overall LGTM with some minor comments. Thanks a lot for the improvement!

core/common/src/main/java/alluxio/conf/PropertyKey.java
elega: i prefer metastore.inode

jiacheliu3: @bzheng888 can you pls change the property keys accordingly? Thanks!

core/server/common/src/main/java/alluxio/util/FeatureUtils.java
bzheng888: Done

bzheng888:
image
Don't know why failed this test, seems not relate to this pr?

core/common/src/main/java/alluxio/conf/PropertyKey.java
bzheng888: Done

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -7535,6 +7558,8 @@ public static final class Name {
     public static final String MASTER_METADATA_SYNC_UFS_PREFETCH_TIMEOUT =
         "alluxio.master.metadata.sync.ufs.prefetch.timeout";
     public static final String MASTER_METASTORE = "alluxio.master.metastore";
+    public static final String MASTER_INODE_METASTORE = "alluxio.master.metastore.inode";
+    public static final String MASTER_BLOCK_METASTORE = "alluxio.master.metastore.block";
```
jiacheliu3, line 7562: pls keep the property names aligned
```suggestion
    public static final String MASTER_METASTORE_INODE = "alluxio.master.metastore.inode";
    public static final String MASTER_METASTORE_BLOCK = "alluxio.master.metastore.block";
```

jiacheliu3 (changes requested):
LGTM with a minor comment

core/common/src/main/java/alluxio/conf/PropertyKey.java
bzheng888: +1

bzheng888:
@jiacheliu3 @elega PTAL

jiacheliu3 (approved):
LGTM, thanks!

elega (approved):
lgtm

bzheng888:
Resolve the conflicts

jiacheliu3:

PR #14566, opened 2021-11-22
Update register stream timestamp more often

jiacheliu3:

During a worker registration stream, we identify stale stream and `DefaultBlockMaster$WorkerRegisterStreamGCExecutor` recycles them based on the last-message-received timestamp. This change updates the timestamp more often.

There will be fewer false positives for stale streams. The false positives are expensive.

NA

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/RegisterStreamObserver.java
+++ b/core/server/master/src/main/java/alluxio/master/block/RegisterStreamObserver.java
@@ -154,6 +154,9 @@ public class RegisterStreamObserver implements StreamObserver<RegisterWorkerPReq
                 + "Probably %s was exceeded!",
             PropertyKey.MASTER_WORKER_REGISTER_STREAM_RESPONSE_TIMEOUT.toString());
 
+        // Update the TS before and after processing the request, so that when a message
+        // takes long to process, the stream does not time out.
+        mContext.updateTs();
         mBlockMaster.workerRegisterFinish(mContext);
         mContext.updateTs();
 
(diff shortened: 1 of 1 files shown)


core/server/master/src/main/java/alluxio/master/block/RegisterStreamObserver.java
```diff
@@ -94,6 +94,9 @@ public RegisterWorkerPResponse call() throws Exception {
                 + "Probably %s was exceeded!",
             PropertyKey.MASTER_WORKER_REGISTER_STREAM_RESPONSE_TIMEOUT.toString());
 
+        // Update the TS before and after processing the request, so that when a message
+        // takes long to process, the stream does not time out.
+        mContext.updateTs();
         mBlockMaster.workerRegisterStream(mContext, chunk, isHead);
         mContext.updateTs();
```
yuzhu, line 101: not familiar with this part of the code, but is the updateTs() call still necessary after the register?

ZhuTopher: From the PR description,
> This change updates the timestamp more often.

So I believe the answer is yes; we want to update the timestamp before & afterwards in case the register takes an exceedingly long time.

ZhuTopher (approved):
LGTM

core/server/master/src/main/java/alluxio/master/block/RegisterStreamObserver.java
jiacheliu3: Yes. The `workerRegisterStream` can take much time in processing. If we only update TS at the end of a message processing, the interval equals `Tmessage_process + Tmessage_in_flight` where `Tmessage_in_flight` equals network delay and worker-side preparation time. If we update TS before and after, the interval becomes more like `max(Tmessage_process, Tmessage_in_flight)`.

jiacheliu3:
David's comment reminded me that the completion logic is also long, and the TS update will be beneficial too. PTAL @yuzhu @ZhuTopher 
The TS update has no downsides as it is only used to check whether a stream is active. Ideally the more frequent we update it, the more it reflects whether the stream is really active.

yuzhu (approved):

jiacheliu3:

PR #14572, opened 2021-11-23
Add timing information to Raft failover process

jenoudet:
Adding timing information to the Raft failover process in the form of a log message and a metric

### DIFF
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -690,6 +690,20 @@ public class JournalStateMachine extends BaseStateMachine {
   }
 
   /**
+   * This function is triggered when the state machine becomes CANDIDATE.
+   */
+  @Override
+  public void notifyExtendedNoLeader(RaftProtos.RoleInfoProto roleInfoProto) {
+    long lastLeaderElapsedTimeMs = roleInfoProto.getCandidateInfo().getLastLeaderElapsedTimeMs();
+    if (roleInfoProto.getRole() == RaftProtos.RaftPeerRole.CANDIDATE) {
+      LOG.info("Transitioned to CANDIDATE, last leader was {}ms ago", lastLeaderElapsedTimeMs);
+      mJournalSystem.setLastLeaderTime(System.currentTimeMillis() - lastLeaderElapsedTimeMs);
+    } else {
+      LOG.warn("Triggered notifyExtendedNoLeader when role is {}", roleInfoProto.getRole());
+    }
+  }
+
+  /**
    * @return the snapshot replication manager
    */
   public SnapshotReplicationManager getSnapshotReplicationManager() {
--- a/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
+++ b/core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
@@ -682,6 +682,9 @@ public class JournalStateMachine extends BaseStateMachine {
   public void notifyLeaderChanged(RaftGroupMemberId groupMemberId, RaftPeerId raftPeerId) {
     if (mRaftGroupId == groupMemberId.getGroupId()) {
       mIsLeader = groupMemberId.getPeerId() == raftPeerId;
+      if (!mIsLeader) {
+        mJournalSystem.setLastLeaderTime(System.currentTimeMillis());
+      }
       mJournalSystem.notifyLeadershipStateChanged(mIsLeader);
     } else {
       LOG.warn("Received notification for unrecognized group {}, current group is {}",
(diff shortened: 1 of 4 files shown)


maobaolong:
@jenoudet Great metric!

maobaolong:
@jenoudet Is there any update?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -682,13 +682,35 @@ public boolean isSnapshotting() {
...
+   * in RaftJournalSystem.
+   */
+  @Override
+  public void notifyExtendedNoLeader(RaftProtos.RoleInfoProto roleInfoProto) {
+    long lastLeaderElapsedTimeMs = roleInfoProto.getCandidateInfo().getLastLeaderElapsedTimeMs();
+    if (roleInfoProto.getRole() == RaftProtos.RaftPeerRole.CANDIDATE) {
+      LOG.info("Transitioned to CANDIDATE, last leader was {}ms ago", lastLeaderElapsedTimeMs);
+      mJournalSystem.setLastLeaderTime(System.currentTimeMillis() - lastLeaderElapsedTimeMs);
+    } else {
+      LOG.warn("Triggered notifyExtendedNoLeader when role is {}", roleInfoProto.getRole());
```
ggezer, line 710: why not set last-leader-time, when the state-machine is say transitioned to PASSIVE or FOLLOWER?

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
```diff
@@ -682,13 +682,35 @@ public boolean isSnapshotting() {
   public void notifyLeaderChanged(RaftGroupMemberId groupMemberId, RaftPeerId raftPeerId) {
     if (mRaftGroupId == groupMemberId.getGroupId()) {
       mIsLeader = groupMemberId.getPeerId() == raftPeerId;
+      if (!mIsLeader) {
+        // If this state machine wins an election right after starting or right after losing
+        // another election, it will not trigger notifyExtendedNoLeader. This is therefore the
+        // last point of contact with another leader.
+        mJournalSystem.setLastLeaderTime(System.currentTimeMillis());
```
ggezer, line 689: what if the state-machine was not the leader to being with? are we then setting last-leader-time of a particular state-machine ahead of when it was a leader the last time?

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
```diff
@@ -482,7 +509,10 @@ public synchronized void gainPrimacy() {
     mAsyncJournalWriter
         .set(new AsyncJournalWriter(mRaftJournalWriter, () -> getJournalSinks(null)));
     mTransferLeaderAllowed.set(true);
-    LOG.info("Gained primacy.");
+
+    mLastFailoverDurationMs.set(System.currentTimeMillis() - getLastLeaderTime());
```
ggezer, line 513: I'm not sure what this measures...
AFAIU, last leader time is supposed to mark the last time this raft instance was a leader. And it might stay as non-leader for a long time...
So this subtraction yields something like; "the time this node was not a leader."
In that case I'm not sure why do we need such info.

ggezer (changes requested):
Have some comments regarding correctness of the calculation.
As a general feedback, I believe having this metric populated for both journals is a saner approach, other than getting into internals of RAFT.

`FaultTolerantAlluxioProcess` should provide all hooks that we can integrate in order to measure the time it takes to transition between roles.

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
jenoudet: See response to [this comment](https://github.com/Alluxio/alluxio/pull/14572/files#r768952525).

core/server/common/src/main/java/alluxio/master/journal/raft/JournalStateMachine.java
jenoudet: I see now how "lastLeaderTime" is confusing, it's not meant to be when this master was leader last, but rather the last time this master interacted with a leading master.
This measurement is meant to describe the time elapsed between when the last leading master lost leadership and when a new leader is up caught up and serving requests. This means measuring the time between when it becomes CANDIDATE and when it becomes LEADER.
This method (`notifyExtendedNoLeader`) is not triggered if the master launches a new election right after losing a previous one. Therefore, I use `notifyLeaderChanged` to mark the last time we saw a leading master in this corner case.

core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java
jenoudet: See response to [this comment](https://github.com/Alluxio/alluxio/pull/14572/files#r768952525).

PR #14574, opened 2021-11-24
[DOCFIX] Fix missing placeholder in metric names

dbw9580:

Fix missing placeholders in some metric names in docs, e.g. https://docs.alluxio.io/os/user/edge/en/reference/Metrics-List.html#master-metrics , by changing the placeholders to use curly braces rather than triangle parentheses. 

These dynamically generated metrics have placeholders in the name, but they are not rendered properly:

图片

That is because the markdown parser interprets the placeholders warped inside `< >` as HTML tags:

图片


Xenorith (approved):

Xenorith:

PR #14575, opened 2021-11-24
Refactor ExecutorServiceBuilder and add to Worker gRPC server

ZhuTopher:

This PR refactors `alluxio.master.ExecutorServiceBuilder` to the top level package `alluxio.ExecutorServiceBuilder`, and adds it as the executor for the Worker's `GrpcServer`.

Without specifying an executor for `GrpcServer`, it will be assigned a [default executor](https://grpc.github.io/grpc-java/javadoc/io/grpc/ServerBuilder.html#executor-java.util.concurrent.Executor-). In our case this would be the `DEFAULT_WORKER_EVENT_LOOP_GROUP_POOL` from [NettyServerBuilder](https://github.com/grpc/grpc-java/blob/ee581bfdfa03e7b0f96fd99e287f5a326648bc42/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java#L85-L86).

This is an unbounded pool which results in unbounded memory consumption by gRPC worker threads. In highly concurrent workloads this can lead to memory exhaustion.

Screen Shot 2021-11-24 at 3 38 54 PM

  1.  `./bin/alluxio runClass alluxio.stress.cli.worker.StressWorkerBench --clients 1 --threads 128`
  2. `./bin/alluxio runClass alluxio.stress.cli.worker.StressWorkerBench --clients 10 --threads 256`
  3. `./bin/alluxio runClass alluxio.stress.cli.worker.StressWorkerBench --clients 1 --threads 512`

Using our `ExecutorServiceBuilder` which is already used for the Master `GrpcServer` will allow us to bound and tune this executor for the Worker.

Screen Shot 2021-11-24 at 2 57 06 PM

  1. `./bin/alluxio runClass alluxio.stress.cli.worker.StressWorkerBench --clients 1 --threads 128`
  2. `./bin/alluxio runClass alluxio.stress.cli.worker.StressWorkerBench --clients 10 --threads 128`
  3. `./bin/alluxio runClass alluxio.stress.cli.worker.StressWorkerBench --clients 1 --threads 256`

Configured as follows
```
alluxio.worker.rpc.executor.core.pool.size=32
alluxio.worker.rpc.executor.max.pool.size=128
```
---

Screenshots were captured from a `jconsole` of the Alluxio worker processes with and without the change.
- Local alluxio cluster deployed according to [this doc](https://docs.alluxio.io/os/user/edge/en/deploy/Running-Alluxio-Locally.html)
- Tested using the [stressbench docs](https://docs.alluxio.io/os/user/edge/en/operation/StressBench.html#single-node-testing-3) for `StressWorkerBench`

New `PropertyKey`s, and one new `MetricKey`
- `alluxio.worker.rpc.executor.<...>` property keys are one-to-one with `alluxio.master.rpc.executor.<...>`
- `Worker.RpcQueueLength` is semantically identical to `Master.RpcQueueLength`, just for the Worker `GrpcServer`

### DIFF
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3852,7 +3852,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey WORKER_RPC_EXECUTOR_CORE_POOL_SIZE =
       new Builder(Name.WORKER_RPC_EXECUTOR_CORE_POOL_SIZE)
-          .setDefaultValue(500)
+          .setDefaultValue(100)
           .setDescription(
               "The number of threads to keep in thread pool of worker RPC ExecutorService.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
@@ -3860,7 +3860,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey WORKER_RPC_EXECUTOR_MAX_POOL_SIZE =
       new Builder(Name.WORKER_RPC_EXECUTOR_MAX_POOL_SIZE)
-          .setDefaultValue(500)
+          .setDefaultValue(1000)
           .setDescription("The maximum number of threads allowed for worker RPC ExecutorService."
               + " When the maximum is reached, attempts to replace blocked threads fail.")
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
@@ -3876,7 +3876,7 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .build();
   public static final PropertyKey WORKER_RPC_EXECUTOR_TPE_QUEUE_TYPE =
       new Builder(Name.WORKER_RPC_EXECUTOR_TPE_QUEUE_TYPE)
-          .setDefaultValue("LINKED_BLOCKING_QUEUE")
+          .setDefaultValue("LINKED_BLOCKING_QUEUE_WITH_CAP")
           .setDescription(String.format(
               "This property is effective when %s is set to TPE. "
                   + "It specifies the internal task queue that's used by RPC ExecutorService. "
--- a/core/common/src/main/java/alluxio/conf/PropertyKey.java
+++ b/core/common/src/main/java/alluxio/conf/PropertyKey.java
@@ -3842,6 +3842,99 @@ public final class PropertyKey implements Comparable<PropertyKey> {
           .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
           .setScope(Scope.WORKER)
           .build();
+  public static final PropertyKey WORKER_RPC_EXECUTOR_TYPE =
+      new Builder(Name.WORKER_RPC_EXECUTOR_TYPE)
+          .setDefaultValue("TPE")
+          .setDescription("Type of ExecutorService for Alluxio worker gRPC server. "
+              + "Supported values are TPE (for ThreadPoolExecutor) and FJP (for ForkJoinPool).")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_RPC_EXECUTOR_CORE_POOL_SIZE =
+      new Builder(Name.WORKER_RPC_EXECUTOR_CORE_POOL_SIZE)
+          .setDefaultValue(500)
+          .setDescription(
+              "The number of threads to keep in thread pool of worker RPC ExecutorService.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_RPC_EXECUTOR_MAX_POOL_SIZE =
+      new Builder(Name.WORKER_RPC_EXECUTOR_MAX_POOL_SIZE)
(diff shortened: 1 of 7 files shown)


core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3842,6 +3842,99 @@ public String toString() {
...
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_RPC_EXECUTOR_MAX_POOL_SIZE =
+      new Builder(Name.WORKER_RPC_EXECUTOR_MAX_POOL_SIZE)
+          .setDefaultValue(500)
+          .setDescription("The maximum number of threads allowed for worker RPC ExecutorService."
+              + " When the maximum is reached, attempts to replace blocked threads fail.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
```
ZhuTopher, line 3868: These values were copied over from `MASTER_RPC_EXECUTOR_<...>`, but I don't have any educated suggestion for a value to set for the Worker.

ZhuTopher: Maybe `WORKER_RPC_EXECUTOR_CORE_POOL_SIZE = 128` and `WORKER_RPC_EXECUTOR_MAX_POOL_SIZE = 1024`?

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3842,6 +3842,99 @@ public String toString() {
...
+      new Builder(Name.WORKER_RPC_EXECUTOR_KEEPALIVE)
+          .setDefaultValue("60sec")
+          .setDescription("The keep alive time of a thread in worker RPC ExecutorService"
+              + "last used before this thread is terminated (and replaced if necessary).")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_RPC_EXECUTOR_TPE_QUEUE_TYPE =
+      new Builder(Name.WORKER_RPC_EXECUTOR_TPE_QUEUE_TYPE)
+          .setDefaultValue("LINKED_BLOCKING_QUEUE")
```
yuzhu, line 3879: could we make the default queue bounded?

yuzhu (changes requested):

core/server/common/src/main/java/alluxio/ExecutorServiceBuilder.java
```diff
@@ -9,11 +9,12 @@
  * See the NOTICE file distributed with this work for information regarding copyright ownership.
  */
 
-package alluxio.master;
+package alluxio;
```
apc999, line 12: not a good practice to put classes in top level package
how about `alluxio.util`  ?

lilyzhoupeijie:
I have tried some ways like this serval days age,and I just test again a few minutes ago with the following code:
图片
the error still exist
图片

图片
I change the GrpcSerializationUtils.getByteBufFromReadableBuffer with AtomicLong,the executor just can not limit the times which function getByteBufFromReadableBuffer executes.
And I tried the client with semphore,it quite works well for now,emmm...so can we discuss about it tomorrow？

yuzhu:
@lilyzhoupeijie what you have there still has an unbounded queue inside it, have you tried it with an bounded queue?

lilyzhoupeijie:
@yuzhu no matter   unbounded  or bounded queue is used ,it only affect whether the queue itself  leads to OOM
if I set newFixedThreadPool to 1,there  will be  only one thread which can work.
this problem may be solved by this link ? https://github.com/salesforce/reactive-grpc
图片

core/common/src/main/java/alluxio/conf/PropertyKey.java
```diff
@@ -3842,6 +3842,99 @@ public String toString() {
...
+          .setScope(Scope.WORKER)
+          .build();
+  public static final PropertyKey WORKER_RPC_EXECUTOR_MAX_POOL_SIZE =
+      new Builder(Name.WORKER_RPC_EXECUTOR_MAX_POOL_SIZE)
+          .setDefaultValue(1000)
+          .setDescription("The maximum number of threads allowed for worker RPC ExecutorService."
+              + " When the maximum is reached, attempts to replace blocked threads fail.")
+          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
+          .setScope(Scope.WORKER)
+          .build();
```
ZhuTopher, line 3868: Reasoning for the default values chosen:
- `WORKER_RPC_EXECUTOR_MAX_POOL_SIZE=1000`
I assumed the default [Java thread stack size](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html#BABHDABI) `-Xss` to be 1024 KB, in which case 1000 threads would consume ~1GB of memory. I used that as the limit since our [software requirements doc](https://docs.alluxio.io/os/user/stable/en/deploy/Requirements.html) suggests at least 1GB of memory for workers.

- `WORKER_RPC_EXECUTOR_CORE_POOL_SIZE=100`
I chose this to be one order of magnitude lower than the maximum pool size.

yuzhu (approved):

yuzhu:

core/server/common/src/main/java/alluxio/ExecutorServiceBuilder.java
apc999: @ZhuTopher  seems this is not addressed in the merged version?

apc999:

Issue #12704, opened 2021-01-07
Support building and running Alluxio on ARM64 platform


liusheng:
**Is your feature request related to a problem? Please describe.**
Currently, more and more softwares have supported running on ARM64 platform, it would be good if Alluxio also support running and testing on ARM platform.

**Describe the solution you'd like**
Try  to make Alluxio can be sucessfully built and run all the test cases passed on ARM64 server, and then setup ARM CI job for Alluxio project to make sure future development won't break ARM64 support.

**Describe alternatives you've considered**
N/A

**Urgency**
Medium

**Additional context**
N/A

calvinjia:
@liusheng Thanks for making this request. Have you already identified platform dependent issues when using ARM64?

liusheng:
@calvinjia 
Thanks for pay attention to this issue, I haven't identified platform dependent issues of Alluxio, but I have tried to build Alluxio on ARM64 server, there are some errors of dependent issues. such as:
```
> node-sass@4.13.0 postinstall /opt/alluxio/webui/node_modules/node-sass
> node scripts/build.js

Building: /opt/alluxio/webui/node/node /opt/alluxio/webui/node_modules/node-gyp/bin/node-gyp.js rebuild --verbose --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=

lerna ERR! npm install --production  --no-save --no-package-lock --no-shrinkwrap stderr:
Cannot download "https://github.com/sass/node-sass/releases/download/v4.13.0/linux-arm64-64_binding.node": 

HTTP error 404 Not Found
...

[INFO] Alluxio UI 2.5.0-SNAPSHOT .......................... FAILURE [01:03 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:06 min (Wall Clock)
[INFO] Finished at: 2021-01-08T02:52:34Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.salesforce.servicelibs:proto-backwards-compatibility:1.0.5:backwards-compatibility-check (default) on project alluxio-core-transport: OS not supported. Unable to find a protolock binary for the classifier linux-aarch_64 -> [Help 1]
[...]

PR #14581, opened 2021-11-25
Issue #12704 - Exclude ratis-thirdparty-misc for alluxio-hub-server

martin-g:

Excludes `ratis-thirdparty-misc` as a transitive dependency in `alluxio-hub-server` module.

ratis-thirdparty-misc brings old versions of netty-tcnative-boringssl-static that breaks the **ProcessMonitorTest's

No user facing changes!


martin-g:
I don't see how my PR will increase the code coverage with 6%. It seems Codecov there is some issue with the way Codecov works.

martin-g:
Any feedback ?

PR #14585, opened 2021-11-26
Add local cache used pages size metric

bzheng888:
Add local cache pages used metric 

Sometimes we not only need space used information, but also the pages size information. The pages size give us meta information for future analyze.

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
@@ -722,6 +722,9 @@ public class LocalCacheManager implements CacheManager {
       MetricsSystem.registerGaugeIfAbsent(
           MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_SPACE_USED.getName()),
           metaStore::bytes);
+      MetricsSystem.registerGaugeIfAbsent(
+          MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_PAGE_SIZE.getName()),
+          metaStore::pages);
     }
   }
 }
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
@@ -723,7 +723,7 @@ public class LocalCacheManager implements CacheManager {
           MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_SPACE_USED.getName()),
           metaStore::bytes);
       MetricsSystem.registerGaugeIfAbsent(
-          MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_PAGE_SIZE.getName()),
+          MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_PAGE_COUNT.getName()),
           metaStore::pages);
     }
   }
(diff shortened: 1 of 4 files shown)


core/common/src/main/java/alluxio/metrics/MetricKey.java
```diff
@@ -1417,6 +1417,12 @@ public MetricKey build() {
           .setMetricType(MetricType.GAUGE)
           .setIsClusterAggregated(false)
           .build();
+  public static final MetricKey CLIENT_CACHE_PAGE_SIZE =
+      new Builder("Client.CachePageSize")
+          .setDescription("Amount of pages used by the client cache.")
```
maobaolong, line 1422: As the description you gave, do you think the metric name `CLIENT_CACHE_PAGE_COUNT` can make sense more?

maobaolong (changes requested):
Thanks for this metrics export, Left one comment inline.

maobaolong (approved):
LGTM

maobaolong:
@beinan Would you please take another look?

maobaolong:
@beinan ping

core/client/fs/src/main/java/alluxio/client/file/cache/LocalCacheManager.java
```diff
@@ -722,6 +722,9 @@ private static void registerGauges(long cacheSize, MetaStore metaStore) {
       MetricsSystem.registerGaugeIfAbsent(
           MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_SPACE_USED.getName()),
           metaStore::bytes);
+      MetricsSystem.registerGaugeIfAbsent(
+          MetricsSystem.getMetricName(MetricKey.CLIENT_CACHE_PAGE_COUNT.getName()),
```
beinan, line 726: I think there has been a metrics key called client_cache_pages.    https://github.com/Alluxio/alluxio/blob/a10823a9523e200f1665228cd4eb3ea7659a0d15/core/common/src/main/java/alluxio/metrics/MetricKey.java#L1488

beinan: If you just wanna count the total number of the pages in our local cache, why not just use this metrics?

https://github.com/Alluxio/alluxio/blob/f2825833fbf5d473fe7265aae749e8b3d52df143/core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java#L146

https://github.com/Alluxio/alluxio/blob/f2825833fbf5d473fe7265aae749e8b3d52df143/core/client/fs/src/main/java/alluxio/client/file/cache/DefaultMetaStore.java#L71

maobaolong:
Hi @beinan , we have some bad experience for the counter which counted by increase and decrease, it maybe wrong when we encounter exception, how about use `mPageMap.size()` instead of the `mPages` and `Metrics.PAGES`

beinan:
> Hi @beinan , we have some bad experience for the counter which counted by increase and decrease, it maybe wrong when we encounter exception, how about use `mPageMap.size()` instead of the `mPages` and `Metrics.PAGES`

Good call, create https://github.com/Alluxio/alluxio/pull/15121 to use the actual size of the map instead of the mPages

beinan:
hello @bzheng888 & @maobaolong , since the metrics of num of pages has been there already, can I shutdown this PR?

bzheng888:
@beinan LGTM! Thanks!

Issue #14518, opened 2021-11-15
Alluxio preferentially move  block to a lower tier when freeing space


safesky:
Currently alluxio will directly remove blocks when freeing space. When the available space in MEM is used up，the evictable block is removed directly. The next time the data is read, it is reload from the UFS. In some cases, it can impose additional performance overhead.
We want to modify this implementation so that when the space needs to be freed, the block that needs to be evicted will moved to the lower tier first. If it fails, we delete the block.

yuzhu:
@Jackson-Wang-7 the concern here is we only free blocks if a write operation needs the space, and it is a blocking operation. Freeing is much faster than moving the data to a lower tier, we do not want this to be in the critical path of the write operation.  Please also benchmark the effect on the write performance if you want to try this approach.

Jackson-Wang-7:
> @Jackson-Wang-7 the concern here is we only free blocks if a write operation needs the space, and it is a blocking operation. Freeing is much faster than moving the data to a lower tier, we do not want this to be in the critical path of the write operation. Please also benchmark the effect on the write performance if you want to try this approach.

OK, we will benchmark the effect on the write performancethe. the current content of this modification is as follows:
[14518-diff.txt](https://github.com/Alluxio/alluxio/files/7559547/14518-diff.txt)

Then we will change it to freeing directly in a write operation and moving data in a promote operation. And the promote operation is expected to be asynchronous.What do you think of that?

yuzhu:
could you submit it as a draft PR?  it is much easier to read.  also @apc999 any thoughts on this?

yuzhu:
@Jackson-Wang-7 could you join our slack channel so we could.chat about this.issue and your fix?

Jackson-Wang-7:
> @Jackson-Wang-7 could you join our slack channel so we could.chat about this.issue and your fix?

Actually, I'm already in the slack channel of #bigo-alluxio. We can discuss it there.

PR #14603, opened 2021-12-01
Alluxio preferentially move block to a lower tier when freeing space

Jackson-Wang-7:
pr-link: Alluxio/alluxio#14518

We want to modify this implementation so that when the space needs to be freed, the block that needs to be evicted will moved to the lower tier first. If it fails, we remove the block.

Currently alluxio will directly remove blocks when freeing space. When the available space in MEM is used up，the evictable block is removed directly. The next time the data is read, it is reload from the UFS. In some cases, it will impose additional performance overhead.

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
+++ b/core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
@@ -11,6 +11,8 @@
 
 package alluxio.worker.block;
 
+import alluxio.StorageTierAssoc;
+import alluxio.WorkerStorageTierAssoc;
 import alluxio.conf.PropertyKey;
 import alluxio.conf.ServerConfiguration;
 import alluxio.exception.BlockAlreadyExistsException;
@@ -114,6 +116,9 @@ public class TieredBlockStore implements BlockStore {
 
   /** Management task coordinator. */
   private ManagementTaskCoordinator mTaskCoordinator;
+
+  /** An object storing the mapping of tier aliases to ordinals. */
+  private final StorageTierAssoc mStorageTierAssoc = new WorkerStorageTierAssoc();
 
   /**
    * Creates a new instance of {@link TieredBlockStore}.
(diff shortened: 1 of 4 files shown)


apc999:
Suggest to have a PropertyKey to turn on and off this feature.

core/server/worker/src/main/java/alluxio/worker/block/allocator/MaxFreeAllocator.java
```diff
@@ -45,7 +45,7 @@ public MaxFreeAllocator(BlockMetadataView view) {
   }
 
   @Override
-  public StorageDirView allocateBlockWithView(long sessionId, long blockSize,
+  public synchronized StorageDirView allocateBlockWithView(long sessionId, long blockSize,
```
beinan, line 48: Could you help clarify why we need add `synchronized ` here?

core/server/worker/src/main/java/alluxio/worker/block/allocator/RoundRobinAllocator.java
```diff
@@ -57,7 +57,7 @@ public RoundRobinAllocator(BlockMetadataView view) {
   }
 
   @Override
-  public StorageDirView allocateBlockWithView(long sessionId, long blockSize,
+  public synchronized StorageDirView allocateBlockWithView(long sessionId, long blockSize,
```
beinan, line 60: Could you help clarify why we need add synchronized here?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -322,25 +327,26 @@ public void requestSpace(long sessionId, long blockId, long additionalBytes)
...
-          AllocateOptions.forRequestSpace(additionalBytes, tempBlockMeta.getBlockLocation()));
-      if (!allocationDir.toBlockStoreLocation().equals(tempBlockMeta.getBlockLocation())) {
-        // If reached here, allocateSpace() failed to enforce 'forceLocation' flag.
-        throw new IllegalStateException(
-            String.format("Allocation error: location enforcement failed for location: %s",
-                allocationDir.toBlockStoreLocation()));
-      }
+    StorageDirView allocationDir = allocateSpace(sessionId,
+        AllocateOptions.forRequestSpace(additionalBytes, tempBlockMeta.getBlockLocation()));
+    if (!allocationDir.toBlockStoreLocation().equals(tempBlockMeta.getBlockLocation())) {
```
beinan, line 337: what if the tempBlockMeta got updated by other threads?  looks like you're shifting these lines out from the resource lock.  any particular reason for you doing this?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -711,16 +717,18 @@ private StorageDirView allocateSpace(long sessionId, AllocateOptions options)
...
+    try (LockResource r = new LockResource(mMetadataReadLock)) {
       // NOTE: a temp block is supposed to be visible for its own writer,
       // unnecessary to acquire block lock here since no sharing.
       if (newBlock) {
         checkTempBlockIdAvailable(blockId);
       }
+    }
 
       // Allocate space.
-      StorageDirView dirView = allocateSpace(sessionId, options);
```
beinan, line 722: why do you shift allocateSpace out of the protection of mMetadataWriteLock?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -809,6 +818,30 @@ public synchronized void freeSpace(long sessionId, long minContiguousBytes,
       if (evictorView.isBlockEvictable(blockToDelete)) {
         try {
           BlockMeta blockMeta = mMetaManager.getBlockMeta(blockToDelete);
+          int tirLevel = mStorageTierAssoc.getOrdinal(blockMeta.getBlockLocation().tierAlias());
```
beinan, line 821: s/tirLevel/tierLevel?

core/server/worker/src/main/java/alluxio/worker/block/TieredBlockStore.java
```diff
@@ -809,6 +818,30 @@ public synchronized void freeSpace(long sessionId, long minContiguousBytes,
...
+          LOG.debug("block [{}] to delete or move, cur tier level [{}]", blockMeta.getBlockId(),
+              tirLevel);
+          if (mStorageTierAssoc.size() > tirLevel + 1) {
+            LOG.debug("block [{}] move to tier level [{}]", blockMeta.getBlockId(), tirLevel + 1);
+            BlockStoreLocation loc = BlockStoreLocation
+                .anyDirInTier(mStorageTierAssoc.getAlias(tirLevel + 1));
+            MoveBlockResult result = moveBlockInternal(sessionId, blockMeta.getBlockId(),
```
beinan, line 828: what will happen if there is no enough space in the next tier?

beinan (changes requested):
I think this feature would be great for multi-tier use cases.  I just put a few comments and clarification questions.  Also as @apc999 suggested, could you add a alluxio property to turn on/off this feature?   Many thanks for the contribution!

PR #14609, opened 2021-12-01
Update Evict semantic

jja725:

change the semantic of evict job: evict 1 copy -> evict down to 1 copy

make evict job idempotent so we can retry

no


jja725:
due to the limit of AlluxioBlockStore, we still want to keep the intention of choosing different replicas from different hosts. So currently it's hard to add an integration test for Evict with 2 copies, may want to resolve this later.

Issue #14276, opened 2021-10-18
Worker Fuse doesn't have resource constraint


LuQQiu:
**Is your feature request related to a problem? Please describe.**
AlluxioFuse or all other clients connect to AlluxioWorker using gRPC.
Alluxio has threadpools to constraint how many threads are processing/fulfilling those gRPC requests together.
However, Fuse inside AlluxioWorker talks to AlluxioWorker directly and there is no internal resource constraints.

When WorkerFuse in under heavy pressure, there maybe a large amount of requests going from Fuse to Worker and can easily eat out all the worker resources.

For example, in cases similar to https://github.com/Alluxio/alluxio/issues/14028
when data is not loaded in Alluxio, there is more than 5000 Fuse opening files. Each Fuse opening files hold an s3 thread from the s3 thread pool. 
When `-Dalluxio.underfs.s3.threads.max=` is smaller than 5000, all the s3 threads are used by worker fuse. 
If i launch `distributedLoad` command, the command will error out saying that `no connection from pool`. 
Have to set `-Dalluxio.underfs.s3.threads.max=` to a value much bigger than 5000 to be able to run distributedLoad to load data from s3.

**Describe the solution you'd like**
Add a constraint layer between WorkerFuse and Worker to prevent Fuse in Worker eat all the resources

**Describe alternatives you've considered**

**Urgency**
Explain why the feature is important

**Additional context**

PR #14622, opened 2021-12-03
Add worker fuse thread pool

ssz1997:
In workerFuse, reading goes through internal method calls instead of gRPC. This brings some problems, for example, issue #14276 .

This PR adds a configurable fixed-size thread pool for reading to put a resource constraint.


ssz1997:
No longer in need, thus closing this PR

PR #14627, opened 2021-12-06
Make the checkConsistency command check more cases

qian0817:

Check consistency should check case that exist in UFS but not exist in alluxio

Now if a file exists in UFS but not in alluxio, then the `checkConsistency` command will not return any consistency information. This PR fixes this issue.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1295,7 +1295,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
           throw e;
         }
         checkConsistencyRecursive(parent, inconsistentUris, false,
-            syncStatus != OK);
+            syncStatus == OK);
 
         auditContext.setSucceeded(true);
       }
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1304,7 +1304,7 @@ public final class DefaultFileSystemMaster extends CoreMaster
   }
 
   private void checkConsistencyRecursive(LockedInodePath inodePath,
-      List<AlluxioURI> inconsistentUris, boolean assertInconsistent, boolean syncMetadata)
+      List<AlluxioURI> inconsistentUris, boolean assertInconsistent, boolean metadataSynced)
           throws IOException, FileDoesNotExistException {
     Inode inode = inodePath.getInode();
     try {
@@ -1321,31 +1321,32 @@ public final class DefaultFileSystemMaster extends CoreMaster
         for (Inode child : children) {
           try (LockedInodePath childPath = inodePath.lockChild(child, LockPattern.READ)) {
             checkConsistencyRecursive(childPath, inconsistentUris, assertInconsistent,
-                syncMetadata);
+                metadataSynced);
           }
         }
         // If a file exists in ufs but not in alluxio,
         // it should be treated as an inconsistent file.
         // if it is a directory we could ignore the subpaths.
         // if the metadata has already been synced, then we could skip it.
-        if (!syncMetadata) {
-          MountTable.Resolution resolution = mMountTable.resolve(inodePath.getUri());
-          UfsStatus[] statuses;
-          try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
-            UnderFileSystem ufs = ufsResource.get();
-            String ufsPath = resolution.getUri().getPath();
-            statuses = ufs.listStatus(ufsPath);
-          }
-          if (statuses != null) {
-            HashSet<String> alluxioFileNames = Streams.stream(children)
-                .map(Inode::getName)
-                .collect(Collectors.toCollection(HashSet::new));
-            Arrays.stream(statuses).forEach(status -> {
-              if (!alluxioFileNames.contains(status.getName())) {
-                inconsistentUris.add(inodePath.getUri().join(status.getName()));
-              }
-            });
-          }
+        if (metadataSynced) {
+          return;
+        }
+        MountTable.Resolution resolution = mMountTable.resolve(inodePath.getUri());
+        UfsStatus[] statuses;
+        try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
+          UnderFileSystem ufs = ufsResource.get();
+          String ufsPath = resolution.getUri().getPath();
+          statuses = ufs.listStatus(ufsPath);
+        }
+        if (statuses != null) {
+          HashSet<String> alluxioFileNames = Streams.stream(children)
+              .map(Inode::getName)
+              .collect(Collectors.toCollection(HashSet::new));
+          Arrays.stream(statuses).forEach(status -> {
+            if (!alluxioFileNames.contains(status.getName())) {
+              inconsistentUris.add(inodePath.getUri().join(status.getName()));
+            }
+          });
         }
       }
     } catch (InvalidPathException e) {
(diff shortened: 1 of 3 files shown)


core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1313,11 +1314,31 @@ private void checkConsistencyRecursive(LockedInodePath inodePath,
...
           }
         }
+        // if a file exist in mount table but not exist in alluxio
+        // it should as regard as inconsistent.
+        // if it's a directory, we could ignore child path.
+        MountTable.Resolution resolution = mMountTable.resolve(inodePath.getUri());
+        try (CloseableResource<UnderFileSystem> ufsResource = resolution.acquireUfsResource()) {
+          UnderFileSystem ufs = ufsResource.get();
+          String ufsPath = resolution.getUri().getPath();
+          UfsStatus[] statuses = ufs.listStatus(ufsPath);
```
LuQQiu, line 1330: Can close ufs resources after listing status

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1313,11 +1314,31 @@ private void checkConsistencyRecursive(LockedInodePath inodePath,
...
+          if (statuses != null) {
+            Arrays.stream(statuses).forEach(status -> {
```
LuQQiu, line 1332: A little worried about the O(n^2) time complexity to compare the inconsistency status.

LuQQiu: Have you tested the functionality?  i am asking because i am not quite sure if 
```
  public List<AlluxioURI> checkConsistency(AlluxioURI path, CheckConsistencyContext context)
      throws AccessControlException, FileDoesNotExistException, InvalidPathException, IOException {
    List<AlluxioURI> inconsistentUris = new ArrayList<>();
    try (RpcContext rpcContext = createRpcContext(context);
        FileSystemMasterAuditContext auditContext =
            createAuditContext("checkConsistency", path, null, null)) {

      syncMetadata(rpcContext,
          path,
          context.getOptions().getCommonOptions(),
          DescendantType.ALL,
          auditContext,
          LockedInodePath::getInodeOrNull,
          (inodePath,  permChecker) -> permChecker.checkPermission(Mode.Bits.READ, inodePath),
          false);
```
 syncMetadata already avoid this case？

LuQQiu (approved):
Thanks for adding more check cases. Left some comments

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
qian0817: fix

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
qian0817: change to O(n+m) time complexity.
Also I know syncMetadata will solves this problem, but syncMetadata relies on parameters configured on both the client and the server, and if it is set to never sync metadata as the default setting, then the inconsistency will always exist and we won't realize that the metadata is inconsistent. So I think the checkConsistency command should handle this case.

LuQQiu: Are we able to avoid this check case if syncMetada already synced and add this check case if not

qian0817: Add logic to determine if metadata is synchronized and skip the check.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1302,7 +1304,7 @@ public void checkAccess(AlluxioURI path, CheckAccessContext context)
   }
 
   private void checkConsistencyRecursive(LockedInodePath inodePath,
-      List<AlluxioURI> inconsistentUris, boolean assertInconsistent)
+      List<AlluxioURI> inconsistentUris, boolean assertInconsistent, boolean syncMetadata)
```
LuQQiu, line 1307: syncMetadata -> metadataSynced

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1318,28 +1320,32 @@ private void checkConsistencyRecursive(LockedInodePath inodePath,
...
-          Arrays.stream(statuses).forEach(status -> {
-            if (!alluxioFileNames.contains(status.getName())) {
-              inconsistentUris.add(inodePath.getUri().join(status.getName()));
-            }
-          });
+        // If a file exists in ufs but not in alluxio,
+        // it should be treated as an inconsistent file.
+        // if it is a directory we could ignore the subpaths.
+        // if the metadata has already been synced, then we could skip it.
+        if (!syncMetadata) {
```
LuQQiu, line 1331: avoid using big if {} block.
use something like `if (metadataSynced) {return}` instead

jenoudet (approved):

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
jenoudet: I was about to suggest the same thing before you did.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
qian0817: done

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
qian0817: done

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1313,11 +1317,37 @@ private void checkConsistencyRecursive(LockedInodePath inodePath,
...
+          String ufsPath = resolution.getUri().getPath();
+          statuses = ufs.listStatus(ufsPath);
+        }
+        if (statuses != null) {
+          HashSet<String> alluxioFileNames = Streams.stream(children)
+              .map(Inode::getName)
+              .collect(Collectors.toCollection(HashSet::new));
+          Arrays.stream(statuses).forEach(status -> {
+            if (!alluxioFileNames.contains(status.getName())) {
+              inconsistentUris.add(inodePath.getUri().join(status.getName()));
```
maobaolong, line 1347: Once the files only in ufs added into `inconsistentUris`, the `checkConsistency command` will try to delete the `inconsistentUris` in alluxio space only, it seems tricky we try to delete files in alluxio only while these files are not in alluxio at all.

codings-dan (commented):
Thanks for working on this. I have some questions: if a path exists in ufs but not in alluxio, can it be achieved with the ls -f command?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
qian0817: The new add inconsistent file is in alluxio but not in ufs, after testing I think the current logic has no problem.

qian0817:
> Thanks for working on this. I have some questions: if a path exists in ufs but not in alluxio, can it be achieved with the ls -f command?

I think the problem can be fixed with `ls -f`, but this PR is aimed to how to find the inconsistency.

qian0817:
@maobaolong @codings-dan Do you still have questions?

LuQQiu:
@qian0817 we are entering 2.7.2 release process, this PR will be merge later (expected by end of this week)

LuQQiu:

PR #14629, opened 2021-12-06
Fix docGen and update configuration/metrics

jja725:
Fix docGen and update configuration/metrics

  1. change in user-facing APIs
  3. webui


docs/_data/table/common-configuration.csv
```diff
@@ -210,12 +209,6 @@ alluxio.zookeeper.leader.connection.error.policy,"SESSION"
 alluxio.zookeeper.leader.inquiry.retry,"10"
 alluxio.zookeeper.leader.path,"/alluxio/leader"
 alluxio.zookeeper.session.timeout,"60s"
-aws.accessKeyId,""
-aws.secretKey,""
```
ZhuTopher, line 209: The following is still in `common-configuration.yaml`
```
aws.accessKeyId:
  'The access key of S3 bucket.'
aws.secretKey:
  'The secret key of S3 bucket.'
```

Should they be removed?

ZhuTopher (changes requested):
Overall LGTM, not sure about if removing `aws.accessKeyId` and `aws.secretKey` is correct or not though.

docs/_data/table/common-configuration.csv
Xenorith: check against one of the previous release branches? ex. release-2.6.2

jja725: yeah I think we keep those two

ZhuTopher (approved):
Thanks, LGTM

Xenorith:

Issue #14620, opened 2021-12-03
workerFuse stressbench write Input/output error


ssz1997:
**Alluxio Version:**
Current master branch

**Describe the bug**
Fuse stress bench with workerFuse writing data errors out. 
Not tuning MaxDirectMemorySize, all threads fail. Logs from `worker.out`:
```
io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 16777216 byte(s) of direct memory (used: 4278190359, max: 4294967296)
        at io.netty.util.internal.PlatformDependent.incrementMemoryCounter(PlatformDependent.java:754)
        at io.netty.util.internal.PlatformDependent.allocateDirectNoCleaner(PlatformDependent.java:709)
        at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:645)
        at io.netty.buffer.PoolArena$DirectArena.newChunk(PoolArena.java:621)
        at io.netty.buffer.PoolArena.allocateNormal(PoolArena.java:204)
        at io.netty.buffer.PoolArena.tcacheAllocateNormal(PoolArena.java:188)
        at io.netty.buffer.PoolArena.allocate(PoolArena.java:138)
        at io.netty.buffer.PoolArena.allocate(PoolArena.java:128)
        at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:378)
        at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:187)
        at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:178)
        at io.netty.buffer.AbstractByteBufAllocator.buffer(AbstractByteBufAllocator.java:115)
        at alluxio.client.block.stream.BlockOutStream.allocateBuffer(BlockOutStream.java:294)
        at alluxio.client.block.stream.BlockOutStream.updateCurrentChunk(BlockOutStream.java:252)
        at alluxio.client.block.stream.BlockOutStream.write(BlockOutStream.java:141)
        at alluxio.client.file.AlluxioFileOutStream.writeInternal(AlluxioFileOutStream.java:271)
        at alluxio.client.file.AlluxioFileOutStream.write(AlluxioFileOutStream.java:229)
        at alluxio.fuse.AlluxioJniFuseFileSystem.writeInternal(AlluxioJniFuseFileSystem.java:438)
[...]

PR #14635, opened 2021-12-07
Release Bytebuf in workerFuse to fix leak

ssz1997:
The ByteBuf is never released in workerFuse writing process. This brings memory leak (out of direct memory)

Partly solve https://github.com/Alluxio/alluxio/issues/14620

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataWriter.java
+++ b/core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataWriter.java
@@ -102,6 +102,7 @@ public final class BlockWorkerDataWriter implements DataWriter {
       }
     }
     long append = mBlockWriter.append(buf);
+    buf.release();
     MetricsSystem.counter(MetricKey.WORKER_BYTES_WRITTEN_DIRECT.getName()).inc(append);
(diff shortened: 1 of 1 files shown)


ssz1997:
@LuQQiu Please take a look. Thanks!

core/client/fs/src/main/java/alluxio/client/block/stream/BlockWorkerDataWriter.java
```diff
@@ -102,6 +102,7 @@ public void writeChunk(final ByteBuf buf) throws IOException {
       }
     }
     long append = mBlockWriter.append(buf);
+    buf.release();
```
LuQQiu, line 105: Wrap the whole method internal content into a try{}finally{buf.release} block to make sure whatever happens inside the writeChunk operation, the buf can always be released.
Otherwise we may still memory leak when BlockWriter append buf failed or other exceptions thrown

LuQQiu (commented):
Thanks for finding the memory leak! Awesome!
Left a comment

LuQQiu (approved):
LGTM, thanks!

LuQQiu:
@ssz1997 can you fix the checkstyle and the TestDataWriter as well

ssz1997:
Wrong base repo. Close this PR.

PR #14649, opened 2021-12-09
Remove hadoop AbstractFileSystem default port

### LINKED ISSUE #14648: Can not submit yarn task via logical domain when enable historyserver
**Alluxio Version:**
2.7.0

**Describe the bug**
Can not submit yarn task via logical domain when enable historyserver.

**To Reproduce**
Use logical address and enable historyserver. then submit a yarn task.

**Expected behavior**
Can submit yarn task successfully.

**Urgency**
High

**Additional context**
exception stack 
```
org.apache.hadoop.fs.InvalidPathException: Invalid path name Wrong FS: alluxio://ebj@logical/tmp/hadoop-yarn/staging/root/.staging/job_1639060727248_0001, expected: alluxio://ebj@logical/
        at org.apache.hadoop.fs.AbstractFileSystem.checkPath(AbstractFileSystem.java:400)
        at org.apache.hadoop.fs.AbstractFileSystem.resolvePath(AbstractFileSystem.java:487)
        at org.apache.hadoop.mapred.YARNRunner.setupLocalResources(YARNRunner.java:394)
        at org.apache.hadoop.mapred.YARNRunner.createApplicationSubmissionContext(YARNRunner.java:573)
        at org.apache.hadoop.mapred.YARNRunner.submitJob(YARNRunner.java:325)
        at org.apache.hadoop.mapreduce.JobSubmitter.submitJobInternal(JobSubmitter.java:251)
        at org.apache.hadoop.mapreduce.Job$11.run(Job.java:1570)
        at org.apache.hadoop.mapreduce.Job$11.run(Job.java:1567)
        at java.security.AccessController.doPrivileged(Native Method)
        at javax.security.auth.Subject.doAs(Subject.java:422)
        at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1730)
        at org.apache.hadoop.mapreduce.Job.submit(Job.java:1567)
        at org.apache.hadoop.mapred.JobClient$1.run(JobClient.java:576)
        at org.apache.hadoop.mapred.JobClient$1.run(JobClient.java:571)
        at java.security.AccessController.doPrivileged(Native Method)
        at javax.security.auth.Subject.doAs(Subject.java:422)
        at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1730)
        at org.apache.hadoop.mapred.JobClient.submitJobInternal(JobClient.java:571)
        at org.apache.hadoop.mapred.JobClient.submitJob(JobClient.java:562)
        at org.apache.hadoop.mapred.JobClient.runJob(JobClient.java:873)
        at org.apache.hadoop.fs.TestDFSIO.runIOTest(TestDFSIO.java:476)
        at org.apache.hadoop.fs.TestDFSIO.writeTest(TestDFSIO.java:455)
        at org.apache.hadoop.fs.TestDFSIO.run(TestDFSIO.java:872)
        at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
        at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:90)
        at org.apache.hadoop.fs.TestDFSIO.main(TestDFSIO.java:743)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.hadoop.util.ProgramDriver$ProgramDescription.invoke(ProgramDriver.java:71)
        at org.apache.hadoop.util.ProgramDriver.run(ProgramDriver.java:144)
        at org.apache.hadoop.test.MapredTestDriver.run(MapredTestDriver.java:139)
        at org.apache.hadoop.test.MapredTestDriver.main(MapredTestDriver.java:147)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.hadoop.util.RunJar.run(RunJar.java:323)
        at org.apache.hadoop.util.RunJar.main(RunJar.java:236)
```

qian0817:

Remove hadoop AbstractFileSystem default port to fix errors when use logical address.

Currently throw an exception when using logical address.
fixes #14648 

method `alluxio.hadoop.AbstactFileSystem.getUriDefaultPort` will return 0.

### DIFF
--- a/core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioFileSystem.java
+++ b/core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioFileSystem.java
@@ -12,7 +12,6 @@
 package alluxio.hadoop;
 
 import alluxio.Constants;
-import alluxio.conf.PropertyKey;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.DelegateToFileSystem;
@@ -38,7 +37,7 @@ import java.net.URISyntaxException;
 public class AlluxioFileSystem extends DelegateToFileSystem {
   /**
    * This constructor has the signature needed by
-   * {@link AbstractFileSystem#createFileSystem(URI, Configuration)}
+   * {@link org.apache.hadoop.fs.AbstractFileSystem#createFileSystem(URI, Configuration)}
    * in Hadoop 2.x.
    *
    * @param uri the uri for this Alluxio filesystem
@@ -48,10 +47,5 @@ public class AlluxioFileSystem extends DelegateToFileSystem {
   AlluxioFileSystem(final URI uri, final Configuration conf)
       throws IOException, URISyntaxException {
     super(uri, new FileSystem(), conf, Constants.SCHEME, false);
-  }
-
-  @Override
-  public int getUriDefaultPort() {
-    return (Integer) PropertyKey.MASTER_RPC_PORT.getDefaultValue();
   }
 }


core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioFileSystem.java
```diff
@@ -54,4 +57,30 @@
...
+        throw new InvalidPathException("relative paths not allowed:" + path);
+      } else {
+        throw new InvalidPathException("path without scheme with non-null authority:" + path);
+      }
+    }
+
+    if (!StringUtils.equals(this.getUri().getHost(), uri.getHost())) {
+      throw new InvalidPathException(
+          "path not have same host: " + path + ", expected: " + this.getUri());
+    }
```
dbw9580, line 80: I don't understand - compared with the default implementation https://github.com/apache/hadoop/blob/d7c5400fbcd31ccb8d2301e244fcd7969fd0b77e/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/AbstractFileSystem.java#L390-L396

your change seems to just ignore the comparison of schemes.

If this change allows your tests to pass, then it must be that the scheme strings are not equal in #14648. However, from the exception message they look the same.

Can you please try to do a live debugging to see what are the schemes exactly at runtime?

qian0817: My implementation here is trying to modify the judgment for ports instead of schemes.As for the different logs you see, it is because of the different version of `AbstractFileSystem`.

qian0817:
In the previous implementation, class `alluxio.hadoop.FileSystem` did not implement method `#getDefaultPort` but class `alluxio.hadoop.AlluxioFileSystem` implement method `#getUriDefaultPort`, so two paths were determined to belong two different file systems because of the inconsistent port. I removed the `#getUriDefaultPort` implementation in `alluxio.hadoop.AlluxioFileSystem` to make them consistent.

qian0817:
I have reworked the implementation, please review the code when free. @dbw9580

core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioFileSystem.java
```diff
@@ -49,9 +48,4 @@
       throws IOException, URISyntaxException {
     super(uri, new FileSystem(), conf, Constants.SCHEME, false);
   }
-
-  @Override
-  public int getUriDefaultPort() {
-    return Integer.parseInt(PropertyKey.MASTER_RPC_PORT.getDefaultValue());
-  }
```
ZhuTopher, line 56: [In the original PR](https://github.com/Alluxio/alluxio/pull/14021#discussion_r714211214), we defined the URI as `alluxio://ebj@<logical name>` to avoid dropping this. I'm not sure if we should be dropping this now as opposed to some other solution.

@ggezer @yuzhu Can I get an opinion from one of you? Thanks!

ZhuTopher (changes requested):

core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioFileSystem.java
yuzhu: @qian0817 i am not sure I understand the rationale behind this change.  I have concerns over back compat over this removal of default port

qian0817: I Search for the relevant code in the hadoop project, it looks like the `#getUriDefaultPort` method is only called [here](https://github.com/apache/hadoop/blob/e103c83765898f756f88c27b2243c8dd3098a989/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/AbstractFileSystem.java#L403). I've read the code here and don't think there are compatibility issues after this modification.

ZhuTopher: ## alluxio.hadoop.FileSystem doesn't implement getDefaultPort()

As you say, `alluxio.hadoop.FileSystem` does not implement `getDefaultPort()` and so relies on the implementation from `org.apache.hadoop.fs.Filesystem`. [That implementation](https://github.com/apache/hadoop/blob/843f66f4dc012dff402bfebc183d46673cd47419/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L413-L419) always returns `0`:
```
  /**
   * Get the default port for this FileSystem.
   * @return the default port or 0 if there isn't one
   */
  protected int getDefaultPort() {
    return 0;
  }
```

Removing our implementation of `getUriDefaultPort()` would fallback into the default implementation of `org.apache.hadoop.fs.DelegateToFileSystem` since `alluxio.hadoop.AlluxioFileSystem` extends it (see [the AlluxioFileSystem definition](https://github.com/Alluxio/alluxio/blob/946a34f077382778300adf8815cd1407bcbb53a1/core/client/hdfs/src/main/java/alluxio/hadoop/AlluxioFileSystem.java#L38)).

`DelegateToFileSystem`'s implementation for `getUriDefaultPort()` is [here](https://github.com/c9n/hadoop/blob/a315107901aebc8d0c2b9f530d191673d27922a2/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/DelegateToFileSystem.java#L141-L144), and it also always returns `0`:
```
  @Override
  public int getUriDefaultPort() {
    return 0;
  }
```

So removing our implementation here does resolve the mismatch, but it means that the port will always be returned as `0`, which I believe is undesirable.

## alluxio.hadoop.AlluxioFileSystem is only used for Hadoop configuration

In fact, I found zero references to its constructor in our codebase (outside of unit tests). In our docs, it is only referenced as part of a `hbase-site.xml` or `core-site.xml` property:
```
<property>
  <name>fs.AbstractFileSystem.alluxio.impl</name>
  <value>alluxio.hadoop.AlluxioFileSystem</value>
</property>
```
- https://docs.alluxio.io/os/user/stable/cn/compute/Hadoop-MapReduce.html?q=AlluxioFileSystem
  - @yuzhu FYI we are missing this in our EN docs; looks like EN needs a docfix?
- https://docs.alluxio.io/os/user/stable/en/compute/HBase.html?q=AlluxioFilesystem
- https://docs.alluxio.io/os/user/stable/en/api/FS-API.html?q=AlluxioFileSystem

Anyway, because of this fact I agree there probably won't be any compatibility issues within our own codebase. However I have no idea how returning `0` instead of `Integer.parseInt(PropertyKey.MASTER_RPC_PORT.getDefaultValue())` will affect end-users using Alluxio with Hadoop.

I'm inclined to believe it'd be better to implement `getDefaultPort()` in `alluxio.hadoop.[Abstract]FileSystem` in a similar manner as is done in `alluxio.hadoop.AlluxioFileSystem` but I'm not familiar enough to say.

Again, @ggezer @yuzhu for feedback. Thanks!

qian0817: I have tried to implement the `#getDefaultPort` method of `alluxio.hadoop.FileSystem`, but I found that the method  [#geturi](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/AbstractFileSystem.java#L339) in `org.apache.hadoop.fs.AbstractFileSystem` use this method, which will change URI from `alluxio://ebj@logical` to `alluxio://logical:19998`.

qian0817: @ZhuTopher 
I tried different ways but others can not bypass this limitation because the code in hadoop. Also I looked at the all `getUriDefaultPort` method call inside hadoop and I don't think it will cause compatibility issues. Do you have any other ideas?

qian0817:
I would like to know if you still have any questions or concerns about this? @yuzhu @ZhuTopher

ZhuTopher:
@qian0817 Thanks for the ping, I haven't had a chance to look into this deeper so I don't know whether I have concerns or not. Perhaps if nobody is able to look at this right now we can just try a couple end-to-end tests on our end to have some confidence in not breaking prior functionality.

qian0817:
Okay, please let me know if you find any issues.

qian0817:
@ZhuTopher any update?

ZhuTopher:
@qian0817 Thanks for the inquiry, unfortunately I still have not had the chance to get back to this. Maybe @yuzhu can re-assign this based on priority.

jenoudet:
I'd like to reopen the discussion about this change as a new use case for this change has emerged. 
The [Hadoop Synthetic Load Generator](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/SLGUserGuide.html) that we have begun using for testing works by issuing a mapreduce job to the Alluxio cluster. If we cannot support mapreduce jobs through Hadoop, then we will not be able to use the Hadoop Synthetic Load Generator for HA tests.
Preliminary testing has indicated that removing this method does enable mapreduce jobs to be used in Alluxio.

qian0817:
I fixed the code conflict. Also no new issues or incompatibilities were found in our production environment. If you have other questions feel free to ask. @jenoudet

PR #14657, opened 2021-12-10
Add Fuse and client alarm metrics

LuQQiu:

Add FUSE and client metrics

Understand the Fuse or client load

Add two centralized metrics

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -313,7 +313,7 @@ public class FileSystemContext implements Closeable {
       // developers should first mark their resources as closed prior to any exceptions being
       // thrown.
       mClosed.set(true);
-      MetricsSystem.removeGaugeIfExist(TOTAL_THREAD_METRICS_NAME);
+      MetricsSystem.removeMetrics(TOTAL_THREAD_METRICS_NAME);
       LOG.debug("Closing fs master client pool with current size: {} for id: {}",
           mFileSystemMasterClientPool.size(), mId);
       mFileSystemMasterClientPool.close();
--- a/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
@@ -32,6 +32,7 @@ import alluxio.exception.status.UnavailableException;
 import alluxio.grpc.GrpcServerAddress;
 import alluxio.master.MasterClientContext;
 import alluxio.master.MasterInquireClient;
+import alluxio.metrics.MetricKey;
 import alluxio.metrics.MetricsSystem;
 import alluxio.refresh.RefreshPolicy;
 import alluxio.refresh.TimeoutRefresh;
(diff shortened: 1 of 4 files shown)


integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
```diff
@@ -277,8 +278,10 @@ public static int call(Logger logger, FuseCallable callable, String methodName,
     long startMs = System.currentTimeMillis();
     int ret = callable.call();
     long durationMs = System.currentTimeMillis() - startMs;
-    MetricsSystem.timer(methodName).update(durationMs, TimeUnit.MILLISECONDS);
     logger.debug("Exit ({}): {}({}) in {} ms", ret, methodName, debugDesc, durationMs);
+    MetricsSystem.timer(methodName).update(durationMs, TimeUnit.MILLISECONDS);
```
maobaolong, line 282: Would you please reference `MetricsSystem.MultiTimerContext` invented by #13671

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -297,6 +313,7 @@ private synchronized void closeContext() throws IOException {
       // developers should first mark their resources as closed prior to any exceptions being
       // thrown.
       mClosed.set(true);
+      MetricsSystem.removeMetrics(TOTAL_THREAD_METRICS_NAME);
```
maobaolong, line 316: Why we need to remove the metric while close context? If it should be removed, how about other metrics?

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
```diff
@@ -91,6 +94,8 @@
 @ThreadSafe
 public class FileSystemContext implements Closeable {
   private static final Logger LOG = LoggerFactory.getLogger(FileSystemContext.class);
+  private static final String TOTAL_THREAD_METRICS_NAME
+      = MetricsSystem.getMetricName(MetricKey.CLIENT_META_DATA_CACHE_SIZE.getName());
```
maobaolong, line 98: Why assign the `CLIENT_META_DATA_CACHE_SIZE` to `TOTAL_THREAD_METRICS_NAME`, it make me confused?

maobaolong (changes requested):
Thanks for bring more metrics for fuse component. Left some comments.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
LuQQiu: this is because in closeContext, all the clientPool will be reset to null.
And in the metric gauge, adding null check is not safe, adding locking will impact the performance significantly.
So directly remove the metrics is a much easier choice.

core/client/fs/src/main/java/alluxio/client/file/FileSystemContext.java
LuQQiu: good catch, fixed, PTAL, thanks

LuQQiu:
@maobaolong @ssz1997 @apc999 Addressed the comments, PTAL, thanks

ssz1997 (approved):
LGTM.

integration/fuse/src/main/java/alluxio/fuse/AlluxioFuseUtils.java
LuQQiu: Added here because we already recorded the time duration for logging, no need to use more timerContext for repeated time recording

maobaolong (approved):
LGTM, thanks!

LuQQiu:

PR #14599, opened 2021-12-01
Enable mount specific dir in Alluxio in helm chart

ssz1997:
Previously using helm chart we can only mount the whole Alluxio storage. Now by specifying `FUSE_ALLUXIO_PATH` in the helm chart, we are able to mount one specific directory in Alluxio, both for standalone Fuse and workerFuse.

PR #14661, opened 2021-12-10
Merge mount Alluxio path in k8s & resolve conflict

ssz1997:
Previously using helm chart we can only mount the whole Alluxio storage.
Now by specifying `FUSE_ALLUXIO_PATH` in the helm chart, we are able to
mount one specific directory in Alluxio, both for standalone Fuse and
workerFuse.

pr-link: #14599
change-id: cid-89a283a3246d855f2d00f9fa516f4855f082911a

### DIFF
--- a/integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
+++ b/integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
@@ -223,3 +223,7 @@
 0.6.32
 
 - Add toggles to disable master/worker resource deployment
+
+0.6.33
+
+- Enable mounting a specific directory in Alluxio through Fuse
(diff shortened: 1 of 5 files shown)


integration/kubernetes/helm-chart/alluxio/CHANGELOG.md
```diff
@@ -223,3 +223,12 @@
 0.6.32
 
 - Add toggles to disable master/worker resource deployment
+
+0.6.33
```
ssz1997, line 227: What should we do with this 0.6.33 changes that don't seem exist in branch-2.7? Do we really care?

Xenorith: err i dont even know what this is about. are these bundled as part of our helm charts? if so, better to be accurate

Xenorith:

PR #14669, opened 2021-12-13
Support register.lease.enabled can be updated

maobaolong:

Now the `alluxio.master.worker.register.lease.enabled` is read once the `BlockMasterWorkerServiceHandler` object init instance, but cannot be updated after that.

Read the propertykey from `ServerConfiguration` each time for worker register, never mind the performance down, since the register request for each worker have only one time, and get a value from `ServerConfiguration` have no io operation.

NO


jiacheliu3 (approved):

jiacheliu3:

PR #14675, opened 2021-12-13
Make recursive options consistent in the filesystem shell

### LINKED ISSUE #8498: Make recursive options consistent in the filesystem shell

We should try to make the usage and options in our filesystems shell mirror the most commonly used commands. For this issue I'm specifically requesting that options dealing with recursive operations are consistent with the most common bash/Posix implementations.

Examples:

- [cp](http://man7.org/linux/man-pages/man1/cp.1.html) can use `-r`, `-R`, or `--recursive`
- [rm](http://man7.org/linux/man-pages/man1/rm.1.html) is the same as cp
- [scp](https://linux.die.net/man/1/scp) only supports `-r`
- [ls](http://man7.org/linux/man-pages/man1/ls.1.html) `-R` or `--recursive` (`-r` is taken by `--reverse` sorting flag)
- [chmod](https://linux.die.net/man/1/chmod) `-R` or `--recursive`. There is no conflict with another `-r` option. It just doesn't take the lowercase flag.

Most of our commands only use the `-R` flag to mark an operation as recursive. It would be nice if 
we supported all variants that our standard bash commands support.

None as of now

**Additional context**

codings-dan:

Fix #8498 

In this patch, I tried to use a method similar to Option.alias(), but
org.apache.commons.cli.Option does not provide a similar function, so I chose to add an ANOTHER_RECURSIVE_OPTION. If there is a better implementation, please suggest it and I will change the code.

make the option consistent with fs shell

No

### DIFF
--- a/shell/src/main/java/alluxio/cli/fs/command/ChmodCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/ChmodCommand.java
@@ -36,7 +36,7 @@ import javax.annotation.concurrent.ThreadSafe;
 public final class ChmodCommand extends AbstractFileSystemCommand {
 
   private static final Option RECURSIVE_OPTION =
-      Option.builder("R")
+      Option.builder("R").longOpt("recursive")
           .required(false)
           .hasArg(false)
           .desc("change mode recursively")
@@ -56,7 +56,7 @@ public final class ChmodCommand extends AbstractFileSystemCommand {
   @Override
   protected void runPlainPath(AlluxioURI plainPath, CommandLine cl)
       throws AlluxioException, IOException {
-    chmod(plainPath, mModeString, cl.hasOption("R"));
+    chmod(plainPath, mModeString, cl.hasOption(RECURSIVE_OPTION.getOpt()));
   }
 
   @Override
@@ -103,12 +103,12 @@ public final class ChmodCommand extends AbstractFileSystemCommand {
 
   @Override
   public String getUsage() {
-    return "chmod [-R] <mode> <path>";
+    return "chmod [-R(--recursive)] <mode> <path>";
   }
 
   @Override
   public String getDescription() {
     return "Changes the permission of a file or directory specified by args."
-        + " Specify -R to change the permission recursively.";
+        + " Specify -R(--recursive) to change the permission recursively.";
   }
 }
--- a/shell/src/main/java/alluxio/cli/fs/command/ChmodCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/ChmodCommand.java
@@ -103,12 +103,12 @@ public final class ChmodCommand extends AbstractFileSystemCommand {
 
   @Override
   public String getUsage() {
-    return "chmod [-R(--recursive)] <mode> <path>";
+    return "chmod [-R/--recursive] <mode> <path>";
   }
 
   @Override
   public String getDescription() {
     return "Changes the permission of a file or directory specified by args."
-        + " Specify -R(--recursive) to change the permission recursively.";
+        + " Specify -R/--recursive to change the permission recursively.";
   }
 }
--- a/shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
+++ b/shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
@@ -76,11 +76,17 @@ public final class CpCommand extends AbstractFileSystemCommand {
   private static final int COPY_TO_LOCAL_BUFFER_SIZE_DEFAULT = 64 * Constants.MB;
 
   private static final Option RECURSIVE_OPTION =
-      Option.builder("R")
+      Option.builder("R").longOpt("recursive")
           .required(false)
           .hasArg(false)
           .desc("copy files in subdirectories recursively")
           .build();
+  private static final Option ANOTHER_RECURSIVE_OPTION =
+      Option.builder("r")
+              .required(false)
+              .hasArg(false)
+              .desc("copy files in subdirectories recursively")
+              .build();
   public static final Option THREAD_OPTION =
       Option.builder()
           .longOpt("thread")
@@ -340,6 +346,7 @@ public final class CpCommand extends AbstractFileSystemCommand {
   @Override
   public Options getOptions() {
     return new Options().addOption(RECURSIVE_OPTION)
+        .addOption(ANOTHER_RECURSIVE_OPTION)
         .addOption(THREAD_OPTION)
         .addOption(PRESERVE_OPTION);
   }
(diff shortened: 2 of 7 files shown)


maobaolong (changes requested):
Thanks for this contribution, left a couple of comments inline.

codings-dan:
@maobaolong Thanks for reviewing the code, PTAL again, thx!

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -300,9 +300,10 @@ private void ls(AlluxioURI path, boolean recursive, boolean forceLoadMetadata, b
   @Override
   protected void runPlainPath(AlluxioURI path, CommandLine cl)
       throws AlluxioException, IOException {
-    ls(path, cl.hasOption("R"), cl.hasOption("f"), cl.hasOption("d"), cl.hasOption("h"),
-        cl.hasOption("p"), cl.getOptionValue("sort", null), cl.hasOption("r"),
-        cl.getOptionValue("timestamp", "lastModificationTime"));
+    ls(path, cl.hasOption(RECURSIVE_OPTION.getOpt()), cl.hasOption("f"),
+            cl.hasOption("d"), cl.hasOption("h"), cl.hasOption("p"),
```
maobaolong, line 304: Wrong indent

shell/src/main/java/alluxio/cli/fs/command/RmCommand.java
```diff
@@ -114,14 +123,15 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
...
+    return "rm [-R --recursive -r] [-U] [--alluxioOnly] <path>";
   }
 
   @Override
   public String getDescription() {
-    return "Removes the specified file. Specify -R to remove file or directory recursively."
-        + " Specify -U to remove directories without checking UFS contents are in sync."
-        + " Specify -alluxioOnly to remove data and metadata from alluxio space only.";
+    return "Removes the specified file. Specify -R(-r or --recursive) to remove file"
+            + " or directory recursively. Specify -U to remove directories without checking "
```
maobaolong, line 132: Wrong indent

shell/src/main/java/alluxio/cli/fs/command/RmCommand.java
```diff
@@ -82,13 +89,15 @@ public Options getOptions() {
   protected void runPlainPath(AlluxioURI path, CommandLine cl)
       throws AlluxioException, IOException {
     // TODO(calvin): Remove explicit state checking.
-    boolean recursive = cl.hasOption("R");
+    boolean recursive = cl.hasOption(RECURSIVE_OPTION.getOpt())
+            || cl.hasOption(RECURSIVE_ALIAS_OPTION.getOpt());
```
maobaolong, line 93: Wrong indent, expected 4, but 8

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
```diff
@@ -488,7 +497,8 @@ private void copy(AlluxioURI srcPath, AlluxioURI dstPath, boolean recursive)
     } else {
       if (!recursive) {
         throw new IOException(
-            srcPath.getPath() + " is a directory, to copy it please use \"cp -R <src> <dst>\"");
+            srcPath.getPath() + " is a directory,"
+                    + " to copy it please use \"cp -R(-r,--recursive) <src> <dst>\"");
```
maobaolong, line 501: Wrong indent

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
```diff
@@ -412,10 +419,12 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
         throw new FileDoesNotExistException(
             ExceptionMessage.PATH_DOES_NOT_EXIST.getMessage(srcPath.getPath()));
       }
+      boolean recursive = cl.hasOption(RECURSIVE_OPTION.getOpt())
+              || cl.hasOption(RECURSIVE_ALIAS_OPTION.getOpt());
```
maobaolong, line 423: Wrong indent

shell/src/main/java/alluxio/cli/fs/command/ChmodCommand.java
```diff
@@ -103,12 +103,12 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
 
   @Override
   public String getUsage() {
-    return "chmod [-R] <mode> <path>";
+    return "chmod [-R(--recursive)] <mode> <path>";
```
LuQQiu, line 106: can you check other classes in the CLI to see what's the better way to say -R or --recursive?
I am asking because you use two formats 
[-R --recursive]
[-R(--recursive)]

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
```diff
@@ -793,16 +803,16 @@ private void copyFileToLocal(AlluxioURI srcPath, AlluxioURI dstPath)
   @Override
   public String getUsage() {
     return "cp "
-        + "[-R] "
+        + "[-R -r --recursive] "
```
LuQQiu, line 806: can we support same set of options, -R and --recursive should be enough
or we can research to see whether -r is a also a commonly used one

LuQQiu (commented):
Left some comments, thanks!

shell/src/main/java/alluxio/cli/fs/command/CpCommand.java
codings-dan: If we are consistent with the linux manual, according to https://man7.org/linux/man-pages/man1/cp.1.html, we need to support -r. But if -r is not supported, we can delete RECURSIVE_ALIAS_OPTION, and the code will be more elegant.

shell/src/main/java/alluxio/cli/fs/command/ChmodCommand.java
codings-dan: I changed to -R/--recursive, which I think is more user-friendly

codings-dan:
@LuQQiu @maobaolong PTAL again, thx!

shell/src/main/java/alluxio/cli/fs/command/LsCommand.java
```diff
@@ -317,7 +318,7 @@ public int run(CommandLine cl) throws AlluxioException, IOException {
 
   @Override
   public String getUsage() {
-    return "ls [-d|-f|-p|-R|-h|--sort=option|--timestamp=option|-r] <path> ...";
+    return "ls [-d|-f|-p|-R/--recursive|-h|--sort=option|--timestamp=option|-r] <path> ...";
```
LuQQiu, line 321: can we all use | instead of / as the separator?

codings-dan: On both sides of “｜”, there are options for different functions. For example, -d and -f have different meanings, but -R and --recursive are essentially the same options, so I think it might be better to use "/".

LuQQiu: it's more a coding style issue. In this case -r is also there using |, isn't them used for the same recursive meaning?

codings-dan: Here, -r means reverse, not recursive.

codings-dan: Maybe we can use other ways to distinguish, such as adding (), what do you think?

codings-dan:
@LuQQiu The CI test error seems to have nothing to do with this code change, PTAL again, thx!

LuQQiu (approved):
LGTM, thanks!

LuQQiu:
This PR will be merged by end of week after 2.7.2 release

LuQQiu:

PR #14676, opened 2021-12-13
Preload metadata in the same directory

tomscut:
If the current file or directory is not in the cache, we can load the metadata of all files and dirs in the parent directory in advance to avoid multiple metadata loading.


maobaolong (commented):
This will cause performance degradation.

tomscut:
> This will cause performance degradation.

Thanks for your comments. This feature is suitable for scenarios where the same directory is appended, so that metadata does not need to be loaded when accessing other new files in the same directory. Maybe it's better to add a option. what do you think?

LuQQiu:
i am worried that during a large amount of concurrency requests, many getstatus will be triggered under the same directory which causing multiple listStatus happening together. When the parent folder contains a large number of files, this change may significantly impact the performance.
Can you provide more information about your workload and how much percentage this change improves the performance?

Issue #13867, opened 2021-07-29
Cannot use mapreduce under HA mode


qian0817:
**Alluxio Version:**

2.6.0

**Describe the bug**

I read the [documentation](https://docs.alluxio.io/os/user/stable/en/compute/Hadoop-MapReduce.html), but it not work when set `fs.defaultFS`  under HA mode. I have tried many ways of writing but none of them work.

**To Reproduce**

I create a [minimal reproduction repository](https://github.com/qian0817/alluxio-ha-mapreduce), just run:
```shell
➜  alluxio-ha-mapreduce git:(master) ✗ docker-compose up -d master1 master2 master3 worker
Creating alluxio-ha-mapreduce_master3_1 ... done
Creating alluxio-ha-mapreduce_master2_1 ... done
Creating alluxio-ha-mapreduce_worker_1  ... done
Creating alluxio-ha-mapreduce_master1_1 ... done
➜  alluxio-ha-mapreduce git:(master) ✗ docker exec -it alluxio-ha-mapreduce_worker_1  alluxio fs chmod -R 777 /
Changed permission of / to 777
➜  alluxio-ha-mapreduce git:(master) ✗ docker-compose up -d nodemanager resourcemanager
Creating resourcemanager ... done
Creating nodemanager     ... done
➜  alluxio-ha-mapreduce git:(master) ✗ docker exec resourcemanager hadoop jar /opt/hadoop-3.2.1/share/hadoop/mapreduce/hadoop-mapreduce-client-jobclient-3.2.1-tests.jar TestDFSIO -write -nrFiles 10 -fileSize 4KB -resFile /tmp/DFSIO-write.out

2021-08-03 02:41:48,675 INFO fs.TestDFSIO: TestDFSIO.1.8
2021-08-03 02:41:48,676 INFO fs.TestDFSIO: nrFiles = 10
2021-08-03 02:41:48,676 INFO fs.TestDFSIO: nrBytes (MB) = 0.00390625
2021-08-03 02:41:48,676 INFO fs.TestDFSIO: bufferSize = 1000000
2021-08-03 02:41:48,676 INFO fs.TestDFSIO: baseDir = /benchmarks/TestDFSIO
[...]

PR #14680, opened 2021-12-14
[DOCFIX] Add fs.AbstractFileSystem.alluxio.impl to Hadoop docs

ZhuTopher:
These changes add an XML property which we suggested in the [Alluxio Hadoop CN docs](https://docs.alluxio.io/os/user/stable/cn/compute/Hadoop-MapReduce.html?q=AlluxioFileSystem#%E9%85%8D%E7%BD%AEhadoop) but which are not present in the [Alluxio Hadoop EN docs](https://docs.alluxio.io/os/user/stable/en/compute/Hadoop-MapReduce.html#configuring-hadoop-core-site-properties).

Similarly, I noticed in the CN docs there is no mention of these properties being skippable for Hadoop >= 2.x. Not sure which of the two docs are correct on that.

### DIFF
--- a/docs/en/compute/Hadoop-MapReduce.md
+++ b/docs/en/compute/Hadoop-MapReduce.md
@@ -32,13 +32,18 @@ For example, if you are using Hadoop 2.7 download this
 
 > **This step is only required for Hadoop 1.x and can be skipped by users of Hadoop 2.x or later**.
 
-Add the following property to the `core-site.xml` of your Hadoop installation:
+Add the following properties to the `core-site.xml` of your Hadoop installation:
 
 ```xml
 <property>
   <name>fs.alluxio.impl</name>
   <value>alluxio.hadoop.FileSystem</value>
-  <description>The Alluxio FileSystem</description>
+  <description>The Alluxio FileSystem (Hadoop 1.x and 2.x)</description>
+</property>
+<property>
+  <name>fs.AbstractFileSystem.alluxio.impl</name>
+  <value>alluxio.hadoop.AlluxioFileSystem</value>
+  <description>The Alluxio AbstractFileSystem (Hadoop 2.x)</description>
 </property>
 ```
 


Xenorith:
was this change motivated by an issue? without any additional context, i would generally trust the EN docs by default even if the CN docs are more verbose in comparison

ZhuTopher:
For the notice about these properties being unnecessary for Hadoop >= 2x, it looks like that has been in the EN doc [since the beginning](https://github.com/Alluxio/alluxio/commit/92029c53ccc35a05b66420a78ba18a72f70771e4#diff-f837e5606c7ad2915fc64ac46621d7b7431ba7b6cc0170788136e6fb96e0fccaR32-R33). So I'll trust EN for that.

As for the additional property for `fs.AbstractFileSystem.alluxio.impl` it came up as a discrepancy I saw while looking through a tangentially related PR (#14649).

I'm tentatively adding it to the EN docs because I wonder if CN has been seeing issues (like in PR #14649) that EN hasn't because they've been including this. I wanted to use this DOCFIX PR primarily to discuss that discrepency.

jiacheliu3:
I don't think https://github.com/Alluxio/alluxio/pull/14649 looks relevant to this issue.

I read the EN vs CN docs. I think they are saying similar things in different ways.

When we don't need to add the below configuration
```
<property>
  <name>fs.AbstractFileSystem.alluxio.impl</name>
  <value>alluxio.hadoop.AlluxioFileSystem</value>
  <description>The Alluxio AbstractFileSystem (Hadoop 2.x)</description>
</property>
```
that is because the [ServiceLoader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) is used when we publish `alluxio.hadoop.AlluxioFileSystem` as `fs.AbstractFileSystem.alluxio.impl`, you will see this defined in the `META-INF/services/` folder somewhere near `AlluxioFileSystem.java`. 

Some time ago I heard someone say Hadoop 1.x did not have ServiceLoader so we had to manually add those lines in `core-site.xml`. That's probably why the EN doc mentioned only Hadoop 1.x needed those lines, and the interface is specific to 1.x only.
```
<property>
  <name>fs.alluxio.impl</name>
  <value>alluxio.hadoop.FileSystem</value>
  <description>The Alluxio FileSystem (Hadoop 1.x and 2.x)</description>
</property>
```

That might explain the discrepancy but I'm actually not so sure. If what I thought I know is true, then we can just keep the section as the EN doc goes. Only Hadoop 1.x will need  the manual setting and only the Hadoop 1.x API needs that. @yuzhu Could you help me confirm?

qian0817:
I don't see the ServiceLoader implemented [here](https://github.com/Alluxio/alluxio/tree/master/core/client/hdfs/src/main/resources/META-INF/services), and I mentioned earlier that [removing this configuration would cause Yarn to fail to start](https://github.com/Alluxio/alluxio/issues/13867#issuecomment-916579311). Maybe we need to add the file to implement  ServiceLoader.

apc999:
The en doc is more up-to-date.
shall we close this PR, @ZhuTopher ?

ZhuTopher:
@apc999 Sure, sounds good.

PR #14689, opened 2021-12-15
Add dev shell script to capture metrics HTTP json sink

ZhuTopher:
This is primarily a convenience script for devs to capture & observe Alluxio metrics. There are no affected components of the codebase since this is a standalone shell script.

### DIFF
--- a/dev/scripts/watch-metrics.sh
+++ b/dev/scripts/watch-metrics.sh
@@ -10,8 +10,14 @@
 # See the NOTICE file distributed with this work for information regarding copyright ownership.
 #
 
-# This script is used as a convenience method for monitoring & capturing
-# the HTTP JSON sink for Alluxio metrics
+# This script is analogous to `alluxio collectInfo collectMetrics` in that
+# it periodically collects the output from the Alluxio HTTP JSON metrics sink.
+# However you may also use it as a way to monitor differences in metrics
+# over time due to the implementation using `watch`.
+# - You can filter which specific metric(s) to collect via `jq` filters
+#
+# This process will output a new file containing the metrics at each
+# polling interval into the specified metrics directory.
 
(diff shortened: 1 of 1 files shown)


dev/scripts/watch-metrics.sh
```diff
@@ -0,0 +1,167 @@
...
+port=""
+metrics_file_prefix=""
+metrics_dir="/tmp/alluxio_metrics"
+
+# Misc. arguments
+time_step=2  # 2s is the default interval for `watch`
+watch=""
+jq_filter=""
+
+USAGE="Usage:
```
yuzhu, line 27: could you add a couple of sentences describing what this tool will do at a high level? like why should i use it instead of just plain curl?

jiacheliu3:
Would this command do a similar thing? 
https://docs.alluxio.io/os/user/stable/en/operation/Troubleshooting.html#collect-alluxio-metrics
Maybe they should be merged so we have one better tool?

ZhuTopher:
In response to
> Would this command do a similar thing?
> https://docs.alluxio.io/os/user/stable/en/operation/Troubleshooting.html#collect-alluxio-metrics
> Maybe they should be merged so we have one better tool?

They do similar things in that they both periodically poll the Alluxio metrics HTTP JSON sink. The only main difference is the usage of `watch` allows for visual feedback on changes to metrics each polling period. `jq` allows us to filter the metrics JSON, but there's no reason that filtering couldn't be added to `alluxio collectInfo collectMetrics` too.

Upon discussion with @yuzhu we're merging this in standalone from `collectMetrics` as an alternative.

yuzhu:

jiacheliu3:
@dbw9580 will this be useful in 2.7?

PR #14698, opened 2021-12-17
Remove permission check from Sync in InodeSyncStream

yyongycy:
Remove permission check from Sync in InodeSyncStream. Sync is the server behavior, should not be checked permission against the client user.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -314,16 +314,7 @@ public class InodeSyncStream {
       if (mAuditContext != null && mAuditContextSrcInodeFunc != null) {
         mAuditContext.setSrcInode(mAuditContextSrcInodeFunc.apply(path));
       }
-      try {
-        if (mPermissionCheckOperation != null) {
-          mPermissionCheckOperation.accept(path, mFsMaster.getPermissionChecker());
-        }
-      } catch (AccessControlException e) {
-        if (mAuditContext != null) {
-          mAuditContext.setAllowed(false);
-        }
-        throw e;
-      }
+
       syncInodeMetadata(path);
       syncPathCount++;
       if (mDescendantType == DescendantType.ONE) {


yyongycy:
@yuzhu @madanadit @jiacheliu3 I created this PR, let's discuss here.

yyongycy:
@ggezer  any thoughts?

jiacheliu3:
There are like 20 consumers of `syncMetadata` and `loadMetadataIfNotExist` in total. Not a terrible number but not trivial either. If we want to make sure, we do have to do a full clean pass and explicitly state that the callers of sync should handle permission check.

Some other things that will be good-to-have:
1. Unit tests that clearly describe the behavior. For example, if user does `getFileInfo` on one path that he/she does not permission, but the path is not sync-ed into Alluxio yet, sync the path into Alluxio with server permission, and throw an error in the later permission check.
2. Update comments on sync() to state what it does and does not do.

I'm fully supportive of removing perm check in sync itself. IMO the sync should be done by the server regardless of who's accessing. Also removing the perm check in the sync step will reduce the total number of repetitive perm checks, because we typically do perm check in the sync and do another perm check in the sync caller.

yyongycy:
@ZacBlanco  I see the code was added by you, could you please help clarify? If the permission check is necessary.

yyongycy:
> 1. getFileInfo



> There are like 20 consumers of `syncMetadata` and `loadMetadataIfNotExist` in total. Not a terrible number but not trivial either. If we want to make sure, we do have to do a full clean pass and explicitly state that the callers of sync should handle permission check.
> 
> Some other things that will be good-to-have:
> 
> 1. Unit tests that clearly describe the behavior. For example, if user does `getFileInfo` on one path that he/she does not permission, but the path is not sync-ed into Alluxio yet, sync the path into Alluxio with server permission, and throw an error in the later permission check.
> 2. Update comments on sync() to state what it does and does not do.
> 
> I'm fully supportive of removing perm check in sync itself. IMO the sync should be done by the server regardless of who's accessing. Also removing the perm check in the sync step will reduce the total number of repetitive perm checks, because we typically do perm check in the sync and do another perm check in the sync caller.

I checked syncMetadata/loadxxx before, and there is always one permission checker following. It looks like the sync has a redundant check there.

yuzhu:
thinking about this issue some more, there are potential values to do a permission check prior to the sync, but not the way it is done today. It can save valuable system resource and avoid the sync if we know the operation itself will fail due to permission check later. For example, `ls /a/b/c` when I don't have execute on `/a`. Sometimes the permission check result of the later operation is unknown because we have not synced yet, then we need to be conservative and allow the sync to happen. (We can have false negatives but not false positives) Unfortunately, in create file, we check the write permission of a parent dir, which is not necessary even for the create operation later.  

Short version, I think this PR's approach is fine but there might be better solutions if we examine each operation individually and improve the current permission check for sync.

yyongycy:
> thinking about this issue some more, there are potential values to do a permission check prior to the sync, but not the way it is done today. It can save valuable system resource and avoid the sync if we know the operation itself will fail due to permission check later. For example, `ls /a/b/c` when I don't have execute on `/a`. Sometimes the permission check result of the later operation is unknown because we have not synced yet, then we need to be conservative and allow the sync to happen. (We can have false negatives but not false positives) Unfortunately, in create file, we check the write permission of a parent dir, which is not necessary even for the create operation later.
> 
> Short version, I think this PR's approach is fine but there might be better solutions if we examine each operation individually and improve the current permission check for sync.

Do we find a lot of sync attack before? 
Another think is the metadata should be synced first, then check against the update information. Otherwise, it will rely on the privileged user to get the latest information. 
And there is no rate limit about the metadata sync, as long as there is a request, the metadata would be synced, is it true?

yuzhu (approved):
LGTM,  I understand that @ggezer and @yyongycy had a discussion? please review as well. @ggezer

yuzhu:

PR #14700, opened 2021-12-17
Increase renameFile and deleteObjects speed by using ObsClient api

adios666:
Override renameFile and deleteObjects in class ObjectUnderFileSystem. Using renameFile and deleteObjects instead.

Background: Alluxio 2.4.1 + Huawei OBS and running spark jobs
1. Default procedure of renameFile in ObjectUnderFileSystem is copy&delete, which is really slow when UFS is Huawei OBS. 
2. Batch delete file is using deleteObject method. It is very slow when a large directory is being deleted when UFS is Huewei OBS.
Based on two problem above, I think we should implement renameFile and deleteObjects of ObsClient in OBSUnderFileSystem.

noop


HelloHorizon:
@beinan @rongrong can you take a look?

HelloHorizon:
@fuzhengjia can you take a look?

PR #14537, opened 2021-11-17
Upgrade Checkstyle version

jenoudet:
### What changes are proposed in this pull request?
This PR upgrades the [maven checkstyle plugin](http://maven.apache.org/plugins/maven-checkstyle-plugin/) version to the latest version. This allows us to use [checkstyle semantics](https://checkstyle.sourceforge.io/) v9.0.1 to interface with the [IntelliJ checkstyle plugin](https://plugins.jetbrains.com/plugin/1065-checkstyle-idea)

The import order remains the same (static imports -> `alluxio.` imports -> all other imports -> `java.` and `javax.` imports). Each of these groups is separated by a newline. Imports are sorted in alphabetical order within each group. The previous version of checkstyle did not enforce any of these rules consistently. The new version of checkstyle does. The effect is that some imports have changed location within their groups (as they were not ordered alphabetically before), and some newlines within groups have been erased (the biggest culprit was an often found empty line between `java.` and `javax.` imports).

This change enables integration with the IntelliJ checkstyle plugin (which dropped support for the version of checkstyle we used before this change). This plugin allows the checkstyle checks to happen within the IDE, and can report checkstyle errors before committing to the repo, and without running `mvn checkstyle:checkstyle` on the command line.

PR #14703, opened 2021-12-18
Include missing Javadoc comments on public methods and classes

jenoudet:
The issue is that the Checkstyle Javadoc rules previously (before #14537) checked both the existence and the format of Javadocs. When they were updated, the Javadoc rules only regarded the format of Javadocs, no longer their existence. This rectifies this oversight.

### DIFF
--- a/build/checkstyle/suppressions.xml
+++ b/build/checkstyle/suppressions.xml
@@ -10,4 +10,7 @@
     <suppress files="[\\/]src[\\/]main[\\/].*|[\\/]generated-sources[\\/].*" id="TestScope" />
     <!--  Suppress specific file  -->
     <suppress files="hub/server/src/main/java/alluxio/hub/manager/process/ManagerProcessContext.java" checks="Indentation" />
+    <!-- Excludes test files from having Javadocs for classes and methods (a lot do not have those) -->
+    <suppress files="[\\/]*[\\/]test[\\/].*" checks="MissingJavadocMethod" />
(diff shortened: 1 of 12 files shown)


Xenorith (approved):

build/checkstyle/suppressions.xml
```diff
@@ -10,4 +10,7 @@
     <suppress files="[\\/]src[\\/]main[\\/].*|[\\/]generated-sources[\\/].*" id="TestScope" />
     <!--  Suppress specific file  -->
     <suppress files="hub/server/src/main/java/alluxio/hub/manager/process/ManagerProcessContext.java" checks="Indentation" />
+    <!-- Excludes test files from having Javadocs for classes and methods (a lot do not have those) -->
+    <suppress files="[\\/]*[\\/]test[\\/].*" checks="MissingJavadocMethod" />
```
Xenorith, line 14: maybe this needs to be `Test` instead of `test`?
also would it be more precise to check for files matching `*Test.java`? dunno if the checkstyle file matcher takes into account the file extension or not

jenoudet: this is meant to escape all files who have a `test` parent directory, which would also include test utils and helpers. I think this is consistent with the old rule because when escaping only `*Test.java` file some violations occur.

Xenorith: got it

Xenorith (commented):
checkstyle failing in alluxio-tests module :(

Xenorith (approved):

tieujason330:

PR #14714, opened 2021-12-19
Set Fuse getattr size to output stream bytes written

### LINKED ISSUE #14713: Show correct bytes written in Fuse getattr
**Alluxio Version:**
Master

**Describe the bug**
Alluxio master only updates file status length after file completed.
But in Alluxio Fuse, sometimes applications need to known the bytes written by calling getattr.
At least in the same Alluxio Fuse which file is written to, we should support getting the bytes written info in getattr

**To Reproduce**

**Expected behavior**
Same Fuse mount point, getattr() can show the correct bytes written even if file is not completed

**Urgency**

**Are you planning to fix it**
Please indicate if you are already working on a PR. 

**Additional context**

LuQQiu:

Fixes #14713

Show correct bytes written in Fuse.getattr() when file is writing to that fuse mount point

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -239,16 +239,26 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
       } else {
         status = mFileSystem.getStatus(uri);
       }
-      if (!status.isCompleted()) {
-        // Always block waiting for file to be completed except when the file is writing
-        // We do not want to block the writing process
-        if (!mCreateFileEntries.contains(PATH_INDEX, path)
-            && !AlluxioFuseUtils.waitForFileCompleted(mFileSystem, uri)) {
-          LOG.error("File {} is not completed", path);
-        }
-        status = mFileSystem.getStatus(uri);
-      }
       long size = status.getLength();
+      if (!status.isCompleted()) {
+        if (mCreateFileEntries.contains(PATH_INDEX, path)) {
+          // Alluxio master will not update file length until file is completed
+          // get file length from the current output stream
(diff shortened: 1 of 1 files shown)


LuQQiu:
@ssz1997 @yyongycy PTAL, thanks!

integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
```diff
@@ -239,16 +239,26 @@ private int getattrInternal(String path, FileStat stat) {
       } else {
         status = mFileSystem.getStatus(uri);
       }
+      long size = status.getLength();
```
yyongycy, line 242: The purpose to get the size here is to make sure the size is gotten at the beginning? Size could be changed during the gap between line 244 and 259?

yyongycy: If it could be changed, why not return the value after it is done?

yyongycy: Does this change mean return the size even if the file is not finished yet?

LuQQiu: Alluxio will not update the file length until the file is completed. So when the file is being written, Alluxio Master only have the file metadata showing the file length is 0.
so at line 242, the file length is zero.

This is fine in many cases but also may cause some applications to error out.
e.g. Some applications calls create() - write() - getattr() - write().
If the getattr() between the two writes show written size 0, the application may error out

LuQQiu: we block waiting for file to be completed when the file is not written by this FUSE.
we cannot block waiting for file is completed when the file is being written by the same FUSE, then the second write() will never be called.

LuQQiu:
@maobaolong Please also take a look, thanks

yyongycy (approved):
Then LGTM,

LuQQiu:

PR #14719, opened 2021-12-20
Fix can't remove data in alluxio when under fs is readonly

bzheng888:
Fix can't remove data in alluxio when under fs is readonly

I think this is a problem. When underfs is readonly, it should not restrict us to remove data from alluxio.

  1. change in user-facing APIs
  3. webui

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
@@ -1848,7 +1848,12 @@ public class DefaultFileSystemMaster extends CoreMaster
       try (LockedInodePath inodePath = mInodeTree
               .lockInodePath(lockingScheme)) {
         mPermissionChecker.checkParentPermission(Mode.Bits.WRITE, inodePath);
-        mMountTable.checkUnderWritableMountPoint(path);
+        // If the mount point is read only, we allow removing the in-Alluxio metadata and data
+        // in order to load it from the UFS again.
+        // This can happen if Alluxio is out-of-sync with the UFS.
+        if (!context.getOptions().getAlluxioOnly()) {
+          mMountTable.checkUnderWritableMountPoint(path);
+        }
         if (!inodePath.fullPathExists()) {
           throw new FileDoesNotExistException(ExceptionMessage.PATH_DOES_NOT_EXIST
               .getMessage(path));


maobaolong (approved):
Good catch LGTM

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1792,7 +1792,9 @@ public void delete(AlluxioURI path, DeleteContext context)
             throw e;
           }
         }
-        mMountTable.checkUnderWritableMountPoint(path);
+        if (!context.getOptions().getAlluxioOnly()) {
+          mMountTable.checkUnderWritableMountPoint(path);
```
jiacheliu3, line 1796: @yuzhu Do you think we should allow rm alluxioOnly files when a mount point is read-only?

bzheng888: @jiacheliu3 We have a scenario where the underfs is read-only and the underfs's data changes. When I execute ./alluxio fs checkConsistency -r /, there are some inconsistent files that should be deleted cannot be deleted.

maobaolong: Another case, ./alluxio fs rm --alluxioOnly /test , the file cannot  be removed too.

jiacheliu3: But what if the UFS file does not change and you just delete the file in Alluxio? The change will not propagate the UFS and will not be rejected. Later the file will appear again due to sync. This will be confusing.

maobaolong: @jiacheliu3 Yeah, the purpose of deleting the file is that the file was never be consistent with the ufs, so we want to delete it from Alluxio, and then, the new consistent inode will be created.

maobaolong:
Hi @jiacheliu3 Do you have any thoughts?

HelloHorizon:
@jiacheliu3 can you take another look?

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
jiacheliu3: Thanks for the explanation, makes sense to me

jiacheliu3: I'm actually a little concerned about files that are not in UFS. For example you `--alluxioOnly`  remove an ASYNC_THROUGH file before it's persisted, or remove a MUST_CACHE file. These can all lead to data loss.

What about we only allow `--alluxioOnly` remove a file only if it's persisted? @yuzhu @maobaolong

jiacheliu3 (changes requested):
Thanks for the contribution! Left some comments, more to start a quick discussion.

Also could you add a unit test? Thx!

bzheng888:
Hi， @jiacheliu3， thanks for your reply, I'm not very clear what your concerns. If we use ASYNC_THROUGH or MUST_CACHE, and no matter ufs is readonly, it has a risk to lead data loss any way. For example, if the ufs is not readonly, and we use ASYNC_THROUGH write, then we can remove data to lead a data loss, and why when the ufs is readonly, we can't remove the in alluxio data?

HelloHorizon:
@jiacheliu3 can you take another look?

jiacheliu3:
Currently blocked by https://github.com/Alluxio/alluxio/issues/14992 Need a better definition of "readonly" before we can determine the correct behavior in this case. This issue needs to be fixed. We are only not sure how.

core/server/master/src/main/java/alluxio/master/file/DefaultFileSystemMaster.java
```diff
@@ -1792,7 +1792,9 @@ public void delete(AlluxioURI path, DeleteContext context)
             throw e;
           }
         }
-        mMountTable.checkUnderWritableMountPoint(path);
+        if (!context.getOptions().getAlluxioOnly()) {
```
jiacheliu3, line 1795: Pls leave a comment here
```
// If the mount point is read only, we allow removing the in-Alluxio metadata and data
// in order to load it from the UFS again.
// This can happen if Alluxio is out-of-sync with the UFS.
```

jiacheliu3 (approved):
LGTM with a minor comment, thanks for the fix!

HelloHorizon:
@bzheng888 can you fix the conflicting files?

bzheng888:
@jiacheliu3 PTAL, thanks!

jiacheliu3 (approved):
LGTM, will merge when tests all pass

jiacheliu3:

PR #14721, opened 2021-12-20
Make metadata cache global

tomscut:
We can globalize metadata cache so that compute nodes like `Presto Worker` can share the metadata cache, improving access efficiency and saving memory.


maobaolong (commented):
Thanks for this PR, it will help to reduce the memory of client side, but if we use one client to access multiply alluxio cluster, the cache metadata will be mess, so, is there a way to use a high level singleton map its key stand for a alluxio cluster, value is the metadata cache for that alluxio cluster?

PR #14731, opened 2021-12-22
Fix the bug which load metadata failed when the path is special

ljl1988com:
Fix the bug which load metadata failed when the path is special.

For example, there are some files whose names are like `tmp\flinkxdata\.data\2c64b72a51458ea04db1b9e25c1cd834_0_0`. Alluxio will transfer the file names into a path like `tmp/flinkxdata/.data/2c64b72a51458ea04db1b9e25c1cd834_0_0`. When loading metadata from ufs will failed every times, leaded to unable to refresh some new metadata which is at the back of special files.
Before:
image

After:

None

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
+++ b/core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
@@ -113,10 +113,10 @@ import javax.annotation.Nullable;
  * while performing a sync for a large tree. Additionally, by using a prefetch mechanism we can
  * concurrently process other inodes while waiting for UFS RPCs to complete.
  *
- * With regards to locking, this class expects to be able to take a write lock on any inode, and
+ * With regard to locking, this class expects to be able to take a write lock on any inode, and
  * then subsequently downgrades or unlocks after the sync is finished. Even though we use
  * {@link java.util.concurrent.locks.ReentrantReadWriteLock}, because we concurrently process
- * inodes on separate threads, we cannot utilize the reetrnant behavior. The implications of
+ * inodes on separate threads, we cannot utilize the reentrant behavior. The implications of
  * that mean the caller of this class must not hold a write while calling {@link #sync()}.
  *
  * A user of this class is expected to create a new instance for each path that they would like
@@ -211,7 +211,7 @@ public class InodeSyncStream {
    * {@link LockPattern#READ}.
    *
    * It is an error to initiate sync without a WRITE_EDGE lock when loadOnly is {@code false}.
-   * If loadOnly is set to {@code true}, then the the root path may have a read lock.
+   * If loadOnly is set to {@code true}, then the root path may have a read lock.
    *
    * @param rootPath The root path to begin syncing
    * @param fsMaster the {@link FileSystemMaster} calling this method
@@ -291,7 +291,7 @@ public class InodeSyncStream {
   }
 
   /**
-   * Sync the metadata according the the root path the stream was created with.
+   * Sync the metadata according the root path the stream was created with.
    *
    * @return SyncStatus object
    */
@@ -734,6 +734,7 @@ public class InodeSyncStream {
       FileDoesNotExistException, InvalidFileSizeException, InvalidPathException, IOException {
     AlluxioURI path = inodePath.getUri();
     MountTable.Resolution resolution = mMountTable.resolve(path);
+    int failedSync = 0;
     try {
       if (context.getUfsStatus() == null) {
         // uri does not exist in ufs
@@ -787,8 +788,15 @@ public class InodeSyncStream {
               loadMetadata(descendant, loadMetadataContext);
             } catch (FileNotFoundException e) {
               LOG.debug("Failed to loadMetadata because file is not in ufs:"
-                      + " inodePath={}, options={}.",
+                  + " inodePath={}, options={}.",
                   childURI, loadMetadataContext, e);
+            } catch (BlockInfoException | FileAlreadyCompletedException
+                | FileDoesNotExistException | InvalidFileSizeException
+                | IOException e) {
+              LOG.debug("Failed to loadMetadata because the ufs file or directory"
+                  + " is {}, options={}.",
+                  childStatus, loadMetadataContext, e);
+              failedSync++;
             }
           }
           mInodeTree.setDirectChildrenLoaded(mRpcContext, inodePath.getInode().asDirectory());
@@ -798,6 +806,10 @@ public class InodeSyncStream {
       LOG.debug("Failed to loadMetadata: inodePath={}, context={}.", inodePath.getUri(), context,
           e);
       throw new IOException(e);
+    }
+    if (failedSync > 0) {
+      throw new IOException(String.format("Failed to load metadata of %s files or directories "
+          + "under %s", failedSync, path));
     }
   }
 
@@ -840,7 +852,7 @@ public class InodeSyncStream {
       }
     }
     for (Pair<Integer, MutableInodeFile> pair : fileEntryMap.values()) {
-      // Replace the old entry place holder with the new entry,
+      // Replace the old entry placeholder with the new entry,
       // to create the file in the same place in the journal
       newEntries.set(pair.getFirst(),
           pair.getSecond().toJournalEntry());


jiacheliu3:
Are there better ways than just throwing an exception for such files? If you "correct" the path somehow, will the file be able to load?

ljl1988com:
@jiacheliu3 I think throwing an exception is a good way. It fix the bug which a path or file loads metadata failed, not only this situation, then still can load other metadata. But a file or path with a special name still can't load. Because `URI.create()` will transform a file or path name which is like 'tmp\flinkxdata\.data\2c64b72a51458ea04db1b9e25c1cd834_0_0' to a path. But the path is not in ufs. 
https://github.com/Alluxio/alluxio/blob/221f0ed5851417ffc6cae4758d112d490e25f930/core/common/src/main/java/alluxio/uri/URI.java#L53-L108

Because I don't know the scope of this method's influence, I dare not fix it.

ljl1988com:
@jiacheliu3 PTAL, Thanks.

HelloHorizon:
@jiacheliu3 @yuzhu can you take another look?

yuzhu:
@ljl1988com could you add in the description of the PR what file names will trigger such behavior? any filename with `\` in it?

ljl1988com:
@yuzhu Yes. the file names contain `\`. Because `AlluxioURI` transfer the file name into a path name.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -789,8 +790,19 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
               LOG.debug("Failed to loadMetadata because file is not in ufs:"
                       + " inodePath={}, options={}.",
                   childURI, loadMetadataContext, e);
+            } catch (BlockInfoException | FileAlreadyCompletedException
+                | FileDoesNotExistException | InvalidFileSizeException
+                | IOException e) {
+              LOG.debug("Failed to loadMetadata because the ufs file or directory"
+                      + " is {}, options={}.",
+                      childStatus, loadMetadataContext, e);
```
jiacheliu3, line 798: Line 797 and 798 should have 4 more spaces from line 796 instead of 8. The checkstyle currently is not able to tell the incorrect indentation, unfortunately.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -789,8 +790,19 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
...
+              failedSync++;
             }
           }
+          if (failedSync > 0) {
+            throw new IOException("Failed to load metadata of " + failedSync
+                    + " files or directories.");
+          }
```
jiacheliu3, line 805: If you throw IOE here, you are catching it on line 809 and re-throwing, which is unnecessary. I think you can throw outside the try-catch.

Also, more than just logging the failed count, can you also log the total and succeeded count?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -789,8 +790,19 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
               LOG.debug("Failed to loadMetadata because file is not in ufs:"
                       + " inodePath={}, options={}.",
                   childURI, loadMetadataContext, e);
+            } catch (BlockInfoException | FileAlreadyCompletedException
+                | FileDoesNotExistException | InvalidFileSizeException
+                | IOException e) {
```
jiacheliu3, line 795: Since you intend to catch the exceptions here, should you change the method signature?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -789,8 +790,19 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
...
+              LOG.debug("Failed to loadMetadata because the ufs file or directory"
+                      + " is {}, options={}.",
```
jiacheliu3, line 797: @yuzhu @ljl1988com Maybe not in this PR, but I think we should re-think on what exactly are we logging in DEBUG level. Logging too many lines may generate too much extra information. However we do want to be able to see the error trace sometime. 

Ideally if a file/dir succeeds to load, I only want to see 1 line max at DEBUG level, and 0 lines at INFO. If something fails to load, currently we expect at least 1 warning (and short) at INFO/WARN level, and at DEBUG level we want the stacktrace (maybe the status and options too, but they can also take many lines).

Sometime before I turn `InodeSyncStream` to DEBUG, just to collect how long each operation takes in `LOG.debug("synced {} paths ({} success, {} failed) in {} ms on {}",...)`. Maybe if you want to print much more details in InodeSyncStream, they should be in TRACE level?

jiacheliu3: Also, there are try-catch blocks in callers of `loadMetadata()`, we'd like to make sure that if an exception is thrown, it's caught once and logged once. Instead of caught-print-rethrown-caught-print-rethrown.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -762,6 +762,7 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
             LOG.debug("fetching children for {} returned null", inodePath.getUri());
             return;
           }
+          int failedSync = 0;
```
jiacheliu3, line 765: The failed count is also counted in `sync()`, how does counting work if we are also counting here?

jiacheliu3 (changes requested):
Thanks for the contribution! 

I get your point that some exceptions can be caught in `loadMetadata()` but if you want to do that, you will then need to think re-arrange the try-catch logic and counters a bit. Please let me know if you have confusions and how I can help offline :)

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ljl1988com: No, because invoking `loadFileMetadataInternal` and `loadDirectoryMetadata` throw these exceptions.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
ljl1988com: The failed count in `sync()` is the counter of loading request path failed. `failedSync` is the counter of loading children path of request path failed.

jiacheliu3: Thx for the explanation.

It can make the log more confusing if another count is logged without explanation. At least the exception or log should contain which path it is from.

And it can be worse. If I understand correctly, if you have a structure
```
/a
/a/b1
/a/b2
/a/b3
/a/b/c1
/a/b/c2
```
The `loadMetadata()` will be called on `/a` and recursively on `/a/b/`. If some files fail to load, you see one line for `/a/b/` and `/a/` like 
```
2 files failed to load in /a/b
1 files failed to load in /a/
```
It'll be very hard to make sense out of these log messages, because every dir counts for its own.

(No need to implement in this PR) @yuzhu Do you think it makes sense to make `loadMetadata()` return a status object and only log in the top level?

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -799,6 +807,10 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
           e);
       throw new IOException(e);
     }
+    if (failedSync > 0) {
+      throw new IOException("Failed to load metadata of " + failedSync
+              + " files or directories.");
```
jiacheliu3, line 812: indentation

jiacheliu3, line 812:
```suggestion
      throw new IOException(String.format("Failed to load metadata of %s files or directories under %s", failedSync, path));
```

jiacheliu3 (changes requested):
Added some comments

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
yuzhu: i agree it makes sense to log at the top level, but returning a status object might not be the right way because we also want to propagate exception messages up. anyways we can address this in a future PR.

yuzhu:
could we create an issue so that we can track these special URI handlings for the future? we do ultimately need to address that.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
jiacheliu3: The status object can include a list of exceptions (or error objects that only contains the exception class and message, to save some space). An alternative is to use `Exception.addSuppressed(Exception cause)` to avoid losing the causer exceptions.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
jiacheliu3: For this PR, I think it's fine to print the count, although this count may get confusing in the log message. We will address the confusion in future PRs.

core/server/master/src/main/java/alluxio/master/file/InodeSyncStream.java
```diff
@@ -799,6 +807,10 @@ private void loadMetadata(LockedInodePath inodePath, LoadMetadataContext context
           e);
       throw new IOException(e);
     }
+    if (failedSync > 0) {
+      throw new IOException(String.format("Failed to load metadata of %s files or directories "
+              + "under %s", failedSync, path));
```
jiacheliu3, line 812: indentation

jiacheliu3 (changes requested):
Only one minor indentation to go :)

jiacheliu3 (approved):
LGTM, thanks for the fix!

HelloHorizon:

PR #14728, opened 2021-12-22
[SMALLFIX] Create ArrayList with size for worker status report

jiacheliu3:

PR #14733, opened 2021-12-22
Update log4j configuration for performance

jiacheliu3:

Change the log format to improve performance. This impacts job master/worker log and audit log.

`%F` is the file that log event is from (java file)
`%M` is the method
Example log looks like this
```
2021-12-14 17:26:32,763 INFO master.AbstractMaster (AbstractMaster.java:stop) - JobMaster: Stopping primary master.
2021-12-14 17:26:32,764 INFO master.AbstractMaster (AbstractMaster.java:stop) - JobMaster: Stopped primary master.
```

As log4j document specifies, using `%F` and `%M` will hurt performance. We have `%c{1}` which outputs the class name anyways so these info are not necessary.
https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternFile in sections for `%F` and `%M`

The log will look different. I don't think anyone is parsing those anyways so the impact should be small enough. We are just changing job master/worker log format to be the same as master/worker.

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
+++ b/core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
@@ -620,8 +620,7 @@ public class DefaultBlockMaster extends CoreMaster implements BlockMaster {
         throw new InvalidArgumentException("Unrecognized worker range: " + workerRange);
     }
 
-    List<WorkerInfo> workerInfoList = new ArrayList<>(
-        selectedLiveWorkers.size() + selectedLostWorkers.size());
+    List<WorkerInfo> workerInfoList = new ArrayList<>();
     for (MasterWorkerInfo worker : selectedLiveWorkers) {
       // extractWorkerInfo handles the locking internally
       workerInfoList.add(extractWorkerInfo(worker, options.getFieldRange(), true));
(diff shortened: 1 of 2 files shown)


core/server/master/src/main/java/alluxio/master/block/DefaultBlockMaster.java
```diff
@@ -620,7 +620,8 @@ public long getUsedBytes() {
         throw new InvalidArgumentException("Unrecognized worker range: " + workerRange);
     }
 
-    List<WorkerInfo> workerInfoList = new ArrayList<>();
+    List<WorkerInfo> workerInfoList = new ArrayList<>(
+        selectedLiveWorkers.size() + selectedLostWorkers.size());
```
jenoudet, line 624: Isn't this a duplicate of #14728 ?

jiacheliu3: Oops my branch messed up and I didn't notice. Will clean this up.

jiacheliu3: Removed that commit from my branch.

jenoudet (approved):

maobaolong (approved):
Thanks for this performance improvement, look forward to merge this, will use it in our cluster.

yuzhu (approved):

yuzhu:

PR #14735, opened 2021-12-22
Create directory recursively in Fuse

ssz1997:
Recursively create directories in JniFuse. 

If some code, for example Fuse stressbench, wants to create directories recursively in the mount point, Alluxio will recursively create those directories as well and won't error out.

Directly using command `mkdir` in the mount point won't create the directories recursively.


LuQQiu (approved):
LGTM, thanks for fixing this issue!

LuQQiu:
After discussion, this is not a fuse bug, close this issue

ssz1997:
@LuQQiu  FYI

PR #14742, opened 2021-12-23
Put all Fuse Stressbench operations in a new dir

ssz1997:
We create a new dir `fuseIOStressBench` inside `localPath` and all files will be written and read inside this new directory. We can then avoid the case that the test files interfere with other files in Alluxio space.

### DIFF
--- a/stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
+++ b/stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
@@ -138,6 +138,8 @@ public class FuseIOBench extends Benchmark<FuseIOTaskResult> {
               + "be at least the number of threads, preferably a multiple of it."
       ));
     }
+    // All test operations happen in mLocalpath/fuseIOStressBench
+    mParameters.mLocalPath = mParameters.mLocalPath + "/fuseIOStressBench";
     File localPath = new File(mParameters.mLocalPath);
 
     if (mParameters.mOperation == FuseIOOperation.WRITE) {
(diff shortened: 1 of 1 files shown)


ssz1997:
@LuQQiu Please take a look. Thanks!

LuQQiu (approved):
LGTM, thanks!
Also a small suggestion, may be there are some path tools to help connect paths and dealing with "/"
otherwise if mLocalPath ends with "/", it may error out

stress/shell/src/main/java/alluxio/stress/cli/fuse/FuseIOBench.java
```diff
@@ -138,6 +138,8 @@ public void prepare() throws Exception {
               + "be at least the number of threads, preferably a multiple of it."
       ));
     }
+    // All test operations happen in mLocalpath/fuseIOStressBench
+    mParameters.mLocalPath = Paths.get(mParameters.mLocalPath, "fuseIOStressBench").toString();
```
LuQQiu, line 142: do we want to name the subfolder following the operation type to avoid conflicts?
but it's update to you
also, did we add cluster-id to the test path?

ssz1997: Operation type you mean "Write", "LocalRead, "RemoteRead" and "ClusterRead"? 

No we don't add cluster id to the test path. We add the job-worker id though

LuQQiu: we can keep the path for now and no need to worry about it

LuQQiu:

PR #14743, opened 2021-12-23
Add string interning to user and group names

jiacheliu3:

Use `intern()` on frequent strings like user/group names. This should change the string memory allocation from the heap into the internal string pool by an extent so we allocate fewer objects (and subsequently friendly to GC).

In a heap analysis I see many duplicate Strings. For example when I run a cluster on AWS with `ec2-user` as both user and group, I see many duplicate "ec2-user" strings from a heap analysis:

image
My heap size is 15G and 1.7G are allocated  to user/group name strings. I have around 25 million files. Note that the 50K String objects are the live objects. There should have been many more and GC-ed from the heap.

image
image
If I'm reading correctly, each string retains 56 bytes in heap.

Using `intern()` should avoid many new String allocations. https://blog.codecentric.de/en/2012/03/save-memory-by-using-string-intern-in-java/

The string intern table brings in a trade off in CPU time because there's now an explicit seek overhead. Using `intern()` is not a no-brainer.

https://blog.codecentric.de/en/2012/03/save-memory-by-using-string-intern-in-java/ This suggests that the default string intern table size has changed to 60013 in Java 8+ so we should be able to put our user/group names in the intern table without too much seek time overhead, **without extra JVM options**.

However there are contradicting complaints on performance like https://stackoverflow.com/a/10628759/4933827 This is from 2017 so maybe later version java has improved significantly.

This means we need to benchmark the performance with a little care than just reasoning if interning makes sense or not.

### DIFF
--- a/core/common/src/main/java/alluxio/grpc/GrpcUtils.java
+++ b/core/common/src/main/java/alluxio/grpc/GrpcUtils.java
@@ -106,8 +106,8 @@ public final class GrpcUtils {
       acl = new AccessControlList();
     }
 
-    acl.setOwningUser(pAcl.getOwner());
-    acl.setOwningGroup(pAcl.getOwningGroup());
+    acl.setOwningUser(pAcl.getOwner().intern());
+    acl.setOwningGroup(pAcl.getOwningGroup().intern());
     acl.setMode((short) pAcl.getMode());
 
     if (pAcl.getEntriesCount() > 0) {
--- a/core/common/src/main/java/alluxio/security/authorization/AccessControlList.java
+++ b/core/common/src/main/java/alluxio/security/authorization/AccessControlList.java
@@ -423,8 +423,8 @@ public class AccessControlList implements Serializable {
       // when the stringEntries size is 0, it can only be a DefaultAccessControlList
       acl = new DefaultAccessControlList();
     }
-    acl.setOwningUser(owner);
-    acl.setOwningGroup(owningGroup);
+    acl.setOwningUser(owner.intern());
+    acl.setOwningGroup(owningGroup.intern());
 
     for (String stringEntry : stringEntries) {
       AclEntry aclEntry = AclEntry.fromCliString(stringEntry);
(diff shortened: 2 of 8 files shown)


core/common/src/main/java/alluxio/grpc/GrpcUtils.java
```diff
@@ -106,8 +106,8 @@ public static AccessControlList fromProto(PAcl pAcl) {
       acl = new AccessControlList();
     }
 
-    acl.setOwningUser(pAcl.getOwner());
-    acl.setOwningGroup(pAcl.getOwningGroup());
+    acl.setOwningUser(pAcl.getOwner().intern());
```
jiacheliu3, line 109: I'm trying to guard all callers of 
```
ACL.setOwningUser()
ACL.setOwningGroup()
MutableInodeFile.setOwningUser()
MutableInodeFile.setOwningGroup()
MutableInodeDirectory.setOwningUser()
MutableInodeDirectory.setOwningGroup()
```
However I'm not 100% percent sure those are the best places. Maybe intern() when we create the Subject in UserState or getting from that?

jiacheliu3:
#Before
Screen Shot 2021-12-28 at 4 55 38 PM

#After
Screen Shot 2021-12-28 at 4 55 14 PM

String interning reduced String instances by around 30% and around 20% in the retained heap size. The difference can be spotted in both the `String` row and the `char[]` row. Note that there's the same number of files (`MutableInodeFile` and `MutableInodeDirectory`), and both clusters are just started from the journal (no operations).

ggezer (approved):
Thanks for the improvement.

ggezer:

LuQQiu:
@jiacheliu3 Awesome improvements!
Better to double check if any performance tradeoff.
@apc999 FYI, a good knowledge sharing topic and a lesson i am interested to learn

LuQQiu:
do we have a test can know how much memory we can improve from this PR?
or whether a benchmark exist?

jiacheliu3:
@dbw9580 Do you need this in 2.7?

PR #14744, opened 2021-12-23
Fix base dir in UFS IO bench

dbw9580:
Fix for #13326

`dataDirPath` introduced in #13326 does not actually work, because `getFilePathStr` still creates file paths from the base path specified by the `--path` argument . Additionally, when the base path is a directory but does not end with a slash, `getFilePathStr` concatenates paths as strings, without properly handling path segments.


dbw9580:
@jiacheliu3 please take a look

jiacheliu3 (approved):
LGTM, thanks for the fix!

jiacheliu3:

PR #14747, opened 2021-12-25
Fix Fuse write

LuQQiu:

Fix incorrect result in linux vim ：wq
modify file.

vim modify file will have many same write requests
it will write(offset=0,size=4096)
and then write(offset=0,size=4096) again.
The changes are validated.

correct the vim modify file results.

### DIFF
--- a/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
+++ b/integration/fuse/src/main/java/alluxio/fuse/AlluxioJniFuseFileSystem.java
@@ -473,10 +473,17 @@ public final class AlluxioJniFuseFileSystem extends AbstractFuseFileSystem
       }
     }
     FileOutStream os = ce.getOut();
-    if (offset != os.getBytesWritten()) {
+    long bytesWritten = os.getBytesWritten();
+    if (offset != bytesWritten && offset + sz > bytesWritten) {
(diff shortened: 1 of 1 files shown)


LuQQiu:
@maobaolong @yyongycy PTAL, thanks!

LuQQiu:
The error is introduced by my PR https://github.com/Alluxio/alluxio/commit/915299521dc436f4fce823c830e7b4da3785031d

maobaolong (commented):
Thanks for your bug report and PR for it.

But I did a test in my test cluster, I cannot reproduce this issue, please add some reproduce steps and I would like to have a test it in my test cluster, thanks.

LuQQiu:
@maobaolong i launch Alluxio in linux instance, create a file by `vi a.file` and `:wq` save it.
After that i use vi again to modify this file `vi a.file` and then modify the content by typing `abc` inside and `:wq`.
After https://github.com/Alluxio/alluxio/commit/915299521dc436f4fce823c830e7b4da3785031d and before this PR, those workloads will fail.
After this pr, the file is successfully modified to be a file with `abc` as content.

maobaolong:
@LuQQiu Thanks for the explaination

maobaolong (approved):
@LuQQiu Thanks for fix this.

LuQQiu:

PR #14754, opened 2021-12-28
Avoid creating new empty sets in MutableInode

jiacheliu3:

The `MutableInode` uses a Set to store mediums that the file/dir is pinned in. By default this is an empty HashSet. This PR changes empty sets to `Collection.emptySet()` which is a singleton, in order to avoid creation of empty HashSets.

The main intention is to reduce memory waste on empty collections. 

Before: 
54.5MB total
342B in a `MutableInodeFile.mMediumTypes` where we have a pin medium specified
Screen Shot 2022-03-29 at 6 31 10 PM

After
46.6MB total
82B in a `MutableInodeFile.mMediumTypes` where we have a pin medium specified
image

When there's no value, the save is one empty `HashSet` per MutableInode, < 100B.
When there's value, the save is 342B -> 82B per MutableInode.

NA

### DIFF
--- a/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
@@ -781,6 +781,8 @@ public class InodeTree implements DelegatingJournaled {
           if (context.getXAttr() != null) {
             updateInodeEntry.putAllXAttr(CommonUtils.convertToByteString(context.getXAttr()));
           }
+          updateInodeEntry.setPinned(currentInodeDirectory.isPinned());
+          updateInodeEntry.addAllMediumType(currentInodeDirectory.getMediumTypes());
           mState.applyAndJournal(rpcContext, updateInodeEntry.build());
         }
       }
@@ -808,9 +810,7 @@ public class InodeTree implements DelegatingJournaled {
           currentInodeDirectory.getId(), pathComponents[k], missingDirContext);
       if (currentInodeDirectory.isPinned() && !newDir.isPinned()) {
         newDir.setPinned(true);
-        if (!currentInodeDirectory.getMediumTypes().isEmpty()) {
-          newDir.setMediumTypes(ImmutableSet.copyOf(currentInodeDirectory.getMediumTypes()));
-        }
+        newDir.setMediumTypes(currentInodeDirectory.getMediumTypes());
       }
       inheritOwnerAndGroupIfEmpty(newDir, currentInodeDirectory);
 
@@ -916,9 +916,7 @@ public class InodeTree implements DelegatingJournaled {
     }
     if (currentInodeDirectory.isPinned() && !newInode.isPinned()) {
       newInode.setPinned(true);
-      if (!currentInodeDirectory.getMediumTypes().isEmpty()) {
-        newInode.setMediumTypes(ImmutableSet.copyOf(currentInodeDirectory.getMediumTypes()));
-      }
+      newInode.setMediumTypes(currentInodeDirectory.getMediumTypes());
     }
 
     mState.applyAndJournal(rpcContext, newInode,
--- a/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
@@ -809,6 +809,8 @@ public class InodeTree implements DelegatingJournaled {
         newDir.setPinned(true);
         if (!currentInodeDirectory.getMediumTypes().isEmpty()) {
           newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+        } else {
+          newDir.setMediumTypes(Collections.emptySet());
         }
       }
       inheritOwnerAndGroupIfEmpty(newDir, currentInodeDirectory);
@@ -917,6 +919,8 @@ public class InodeTree implements DelegatingJournaled {
       newInode.setPinned(true);
       if (!currentInodeDirectory.getMediumTypes().isEmpty()) {
         newInode.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+      } else {
+        newInode.setMediumTypes(Collections.emptySet());
       }
     }
 
--- a/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
+++ b/core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
@@ -807,7 +807,9 @@ public class InodeTree implements DelegatingJournaled {
           currentInodeDirectory.getId(), pathComponents[k], missingDirContext);
       if (currentInodeDirectory.isPinned() && !newDir.isPinned()) {
         newDir.setPinned(true);
-        newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+        if (!currentInodeDirectory.getMediumTypes().isEmpty()) {
+          newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+        }
       }
       inheritOwnerAndGroupIfEmpty(newDir, currentInodeDirectory);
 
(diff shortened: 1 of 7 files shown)


core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -807,7 +807,11 @@ public InodeDirectory getRoot() {
           currentInodeDirectory.getId(), pathComponents[k], missingDirContext);
       if (currentInodeDirectory.isPinned() && !newDir.isPinned()) {
         newDir.setPinned(true);
-        newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+        if (!currentInodeDirectory.getMediumTypes().isEmpty()) {
+          newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+        } else {
+          newDir.setMediumTypes(Collections.emptySet());
+        }
```
jiacheliu3, line 814: This is a lot of boilerplate code at the caller site. Instead of doing this, we can modify the `setMediumTypes` like below:

Instead of just setting the field to whatever is given, we do the copy internally.
```
setMediumTypes(Set<String> set) {
  if (set.isEmpty()) {
    mMediumTypes = Collections.emptySet();
  } else {
    mMediumTypes = new HashSet(set);
  }
}
```

So the caller will do
```
// Before
newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
// After
newDir.setMediumTypes(currentInodeDirectory.getMediumTypes());
```

Above should be a better way. I'm just slightly concerned that it becomes different from a pure "setter", and therefore should have a better name to reflect the underlying copy, like `copyMediumTypes()`.

ggezer: How about we declare and reorganize mediumTypes as immutable?
So that we can handle this logic in two parts;
1- `getMediumTypes()` returns a reference to existing immutable reference (Collections.emptySet() for empty list)
2- `setMediumTypes()` or any add/remove/update method always create a new map.

Otherwise, I'm still in favor of handling this in `getMediumTypes()`, (Or `copyMediumTypes()`?), and make the setter as plain as possible.

jiacheliu3: I'm in favor of this but i'm not sure I fully understood:
```
So that we can handle this logic in two parts;
1- getMediumTypes() returns a reference to existing immutable reference (Collections.emptySet() for empty list)
2- setMediumTypes() or any add/remove/update method always create a new map.
```

```
getMediumTypes() {
  // Just return the ref because it's immutable
  // The current interface doesn't say anything about immutability and because of the setter, we can't make this "final" :(
  // As of now, all callers are read-only
  return mMediumTypes;
}
setMediumTypes(Set<> types) {
  // Handle copy internally to ensure immutability
  if (types.isEmpty()) {
    mMediumTypes = Collections.emptySet();
  } else {
    mMediumTypes = new HashSet<>(types);
  }
}
// There are no add/remove/update methods in the interface so no need to worry
// The setter is the only update entrace
// Update paths include setAttribute and reading from journal etc
```

jiacheliu3: One step further, the mutability of this `Set<String>` does not mean that much and we are not using it at all. If we know all possible medium types already these `Set`s can well be singletons and `EnumSet`s.

Currently we don't have a good definition of what the medium types are. The user can define anything not limited by "MEM,SSD,HDD". There's nothing we explicitly do to check the medium types match between master and all workers. Taking more steps back, setting this HashSet is currently only used by feature "pinning to specified medium". The overhead is really non-trivial for some feature that is rarely used.

jiacheliu3: Moving to a separate thread for discussion https://github.com/Alluxio/alluxio/issues/14820

yuzhu: I think EnumSet is a good choice here. We just need to fix the medium type candidate to be one of MEM, SSD, and HDD. I have never seen anything else used anyways.   If we want flexiblility, BitSet could be used. I agree that HashSet is too expensive as a per-inode construct right now.

yuzhu:
@jiacheliu3 PTAL and wrap up this useful optimization?

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
ggezer: Jiacheng, right. That's what I meant.

For enum based representation of medium types, I think you might wanna see this section as well. I also think that's much better.
https://docs.google.com/document/d/1uAPy3aspGwzJmNZ_oUiFxLIdjGJD5-s5k0aRNUC4iqA/edit#heading=h.296qlbvti652

jiacheliu3: I tried to convert MEDIUM to an enum and everything else will start being EnumSet. This is the implementation: https://github.com/Alluxio/alluxio/pull/15223

Let's compare this PR with #15223 and choose one. @yuzhu @ggezer

core/server/master/src/main/java/alluxio/master/file/meta/MutableInodeDirectory.java
```diff
@@ -209,7 +210,11 @@ public static MutableInodeDirectory fromJournalEntry(InodeDirectoryEntry entry)
     } else {
       ret.mDefaultAcl = new DefaultAccessControlList();
     }
-    ret.setMediumTypes(new HashSet<>(entry.getMediumTypeList()));
+    if (!entry.getMediumTypeList().isEmpty()) {
+      ret.setMediumTypes(new HashSet<>(entry.getMediumTypeList()));
+    } else {
+      ret.setMediumTypes(Collections.emptySet());
```
ggezer, line 216: Don't we already set the internal mediumtype list to `Collections.emptySet()` when it's empty?
Even if we not, we could be returning `Collections.emptySet()` from the `getMediumTypeList` call to avoid these if-else statements everywhere else.

ggezer (changes requested):

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
jiacheliu3: On 2nd thought, moving to `ImmutableSet` in both member variable and the method signature answers our question on "Who does the defensive copy?" @ggezer 

Now the answer is, the caller does the copy and everything is now immutable explicitly and the signature makes the usage fool-proof enough. Using `ImmutableSet` gives us the type safety, in contrast with `unmodifiableSet` which is only `Set`. 

This pattern can be applied to many other internal collection-like member fields, if any.

core/server/master/src/main/java/alluxio/master/file/meta/InodeTree.java
```diff
@@ -807,7 +810,7 @@ public InodeDirectory getRoot() {
           currentInodeDirectory.getId(), pathComponents[k], missingDirContext);
       if (currentInodeDirectory.isPinned() && !newDir.isPinned()) {
         newDir.setPinned(true);
-        newDir.setMediumTypes(new HashSet<>(currentInodeDirectory.getMediumTypes()));
+        newDir.setMediumTypes(currentInodeDirectory.getMediumTypes());
```
jiacheliu3, line 813: No defensive copy is needed anymore, since everything is immutable. Actually, if the child inode just uses a ref to the parent dir's medium Set, that'd be even better cuz there's one less object off the table. But at the moment it doesn't seem to be the case.

core/server/master/src/main/java/alluxio/master/file/meta/MutableInode.java
```diff
@@ -610,12 +614,28 @@ public void updateFromEntry(UpdateInodeEntry entry) {
     if (entry.hasPinned()) {
       // pinning status has changed, therefore we change the medium list with it.
       if (entry.getPinned()) {
-        List<String> mediaList = ServerConfiguration.getList(
-            PropertyKey.MASTER_TIERED_STORE_GLOBAL_MEDIUMTYPE);
```
jiacheliu3, line 614: Moved to static constant

ggezer (approved):

yuzhu (approved):

jiacheliu3:

PR #14757, opened 2021-12-28
Fix otel/opentelemetry-collector-dev:latest is not backward compatible

ddadaal:

Fix otel/opentelemetry-collector-dev:latest is not backward compatible using specific image tags [as instructed in the original issue](https://github.com/Alluxio/new-contributor-tasks/issues/613).

Fixes Alluxio/new-contributor-tasks#613.

### DIFF
--- a/integration/metrics/docker-compose-master.yaml
+++ b/integration/metrics/docker-compose-master.yaml
@@ -21,7 +21,7 @@ services:
 
   # Collector
   otel-collector:
-    image: otel/opentelemetry-collector-dev:latest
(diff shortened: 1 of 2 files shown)


yuzhu (approved):

yuzhu:
@ddadaal is there a version tag we can use instead of hash value?

ddadaal:
@yuzhu https://hub.docker.com/r/otel/opentelemetry-collector-dev/tags?page=1&name=b24db545eb56e2ee7dcf6a8f3740ce49bb80608b I don't think there is a tag to this hash.

PR #14763, opened 2021-12-29
[DOCFIX] Fix tensorflow training script download

ssz1997:
The original download will download the whole html instead of the script. 

Add `?raw=true` to make it download the python script.


LuQQiu (approved):

LuQQiu:

PR #14770, opened 2021-12-31
[SMALLFIX] Modify the hdfs ufsConstants output path

qian0817:

Modify the hdfs ufsConstants output path

Make it consistent with https://github.com/Alluxio/alluxio/blob/master/core/common/pom.xml#L206

No user facing changes.


maobaolong (approved):
Reference `core/common/pom.xml`, I agree with this change, it will help ide to find the correct source path while debug within an ide like `intellij idea`.

yuzhu (approved):

HelloHorizon:

Issue #14771, opened 2021-12-31
Support to cache data in memory when using Alluxio Local Cache


zzcclp:
**Is your feature request related to a problem? Please describe.**

Currently, Alluxio Local Cache supports two types to cache data: local disk and Rocks, in some cases, there are some small data needed to be cached, the memory size is enough for them and there is no SSD on the machines.

**Describe the solution you'd like**
It's better to support caching data in the memory when using Alluxio Local Cache, adding the third PageStore type: MEM.

**Describe alternatives you've considered**
None.

**Urgency**
Explain why the feature is important

**Additional context**

zzcclp:
I'll raise a pr to implement this feature.

zzcclp:
I have raised PR #14772 to implement this feature.

apc999:
thanks for the contribution! We will start to review this PR!

PR #14772, opened 2021-12-31
Support to cache data in memory when using Alluxio Local Cache

zzcclp:

Add the third PageStore type: MEM, to support caching data in memory.

Please see the description in #14771 .

  1. Add a new PageStore type: MEM

### DIFF
--- a/core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
+++ b/core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
@@ -1,0 +1,123 @@
+/*
+ * The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+ * (the "License"). You may not use this work except in compliance with the License, which is
+ * available at www.apache.org/licenses/LICENSE-2.0
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied, as more fully set forth in the License.
+ *
+ * See the NOTICE file distributed with this work for information regarding copyright ownership.
+ */
+
+package alluxio.client.file.cache.store;
+
+import alluxio.client.file.cache.PageId;
+import alluxio.client.file.cache.PageInfo;
+import alluxio.client.file.cache.PageStore;
+import alluxio.exception.PageNotFoundException;
+import alluxio.exception.status.ResourceExhaustedException;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.concurrent.NotThreadSafe;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Stream;
+
+/**
+ * The {@link MemoryPageStore} is an implementation of {@link PageStore} which
+ * stores all pages in memory.
+ */
+@NotThreadSafe
+public class MemoryPageStore implements PageStore {
+  private static final Logger LOG = LoggerFactory.getLogger(MemoryPageStore.class);
+  private final long mPageSize;
+  private final long mCapacity;
+
+  private ConcurrentHashMap<PageId, byte[]> mPageStoreMap = null;
+
+  /**
+   * Creates a new instance of {@link MemoryPageStore}.
+   *
+   * @param options options for the buffer store
+   */
+  public MemoryPageStore(MemoryPageStoreOptions options) {
+    mPageSize = options.getPageSize();
+    mCapacity = (long) (options.getCacheSize() / (1 + options.getOverheadRatio()));
+    mPageStoreMap = new ConcurrentHashMap<>();
+  }
+
+  @Override
+  public void put(PageId pageId, byte[] page) throws ResourceExhaustedException, IOException {
+    PageId pageKey = getKeyFromPageId(pageId);
+    try {
+      byte[] mPage = new byte[page.length];
+      System.arraycopy(page, 0, mPage, 0, page.length);
+      mPageStoreMap.put(pageKey, mPage);
+    } catch (Exception e) {
+      throw new IOException("Failed to put cached data in memory for page " + pageId);
+    }
+  }
+
+  @Override
+  public int get(PageId pageId, int pageOffset, int bytesToRead, byte[] buffer, int bufferOffset)
+      throws IOException, PageNotFoundException {
+    Preconditions.checkArgument(buffer != null, "buffer is null");
+    Preconditions.checkArgument(pageOffset >= 0, "page offset should be non-negative");
+    Preconditions.checkArgument(buffer.length >= bufferOffset,
+        "page offset %s should be " + "less or equal than buffer length %s", bufferOffset,
+        buffer.length);
+    PageId pageKey = getKeyFromPageId(pageId);
+    if (!mPageStoreMap.containsKey(pageKey)) {
+      throw new PageNotFoundException(pageId.getFileId() + "_" + pageId.getPageIndex());
+    }
+    byte[] mPage = mPageStoreMap.get(pageKey);
+    Preconditions.checkArgument(pageOffset <= mPage.length, "page offset %s exceeded page size %s",
+        pageOffset, mPage.length);
+    int bytesLeft = (int) Math.min(mPage.length - pageOffset, buffer.length - bufferOffset);
+    bytesLeft = Math.min(bytesLeft, bytesToRead);
+    System.arraycopy(mPage, pageOffset, buffer, bufferOffset, bytesLeft);
+    return bytesLeft;
+  }
+
+  @Override
+  public void delete(PageId pageId) throws IOException, PageNotFoundException {
+    PageId pageKey = getKeyFromPageId(pageId);
+    if (!mPageStoreMap.containsKey(pageKey)) {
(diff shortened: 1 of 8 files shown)


zzcclp:
alluxio-bot, retest this please

zzcclp:
The test case 'alluxio.util.JvmPauseMonitorTest.testMockedInfoPause' failed, I met this  several times.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Creates a new instance of {@link MemoryPageStore}.
+   *
+   * @param options options for the buffer store
+   */
+  public MemoryPageStore(MemoryPageStoreOptions options) {
+    mPageSize = options.getPageSize();
+    mCapacity = (long) (options.getCacheSize() / (1 + options.getOverheadRatio()));
+    mFileBuckets = options.getFileBuckets();
```
beinan, line 52: do we still need mFileBuckets for mem page store?  I think it's just helpful for file system storage.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
```diff
@@ -0,0 +1,132 @@
...
+    mPageStoreMap = new ConcurrentHashMap<>();
+  }
+
+  @Override
+  public void put(PageId pageId, byte[] page) throws ResourceExhaustedException, IOException {
+    String pageKey = getKeyFromPageId(pageId);
```
beinan, line 58: is there any concern to use pageId as the key directly? I didn't see any benefit to use a string key here

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      byte[] mPage = new byte[page.length];
+      System.arraycopy(page, 0, mPage, 0, page.length);
```
beinan, line 61: is the arraycopy here necessary?  I think when we read a page from external file system, we already made a copy there.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
```diff
@@ -0,0 +1,132 @@
...
+    String pageKey = getKeyFromPageId(pageId);
+    if (!mPageStoreMap.containsKey(pageKey)) {
+      throw new PageNotFoundException(pageKey);
+    }
+    byte[] mPage = mPageStoreMap.get(pageKey);
+    Preconditions.checkArgument(pageOffset <= mPage.length, "page offset %s exceeded page size %s",
+        pageOffset, mPage.length);
+    int bytesLeft = (int) Math.min(mPage.length - pageOffset, buffer.length - bufferOffset);
+    bytesLeft = Math.min(bytesLeft, bytesToRead);
+    System.arraycopy(mPage, pageOffset, buffer, bufferOffset, bytesLeft);
```
beinan, line 85: arraycopy might make it safer, but is it necessary?

beinan (commented):
Great contribution! Our community is waiting for this one for a while. Thank you Zhichao!  it looks good to me except a couple of clarification questions, thanks!

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
zzcclp: I will change the policy for the  key

zzcclp: Already removed mFileBuckets.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
zzcclp: Use the 'MD5(file_id(str)_page_idx(ulong)).asLong' as key, the type is long.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
zzcclp: I think it's necessary, if use the byte array data directly, when the original byte array data is modified, the value is changed too.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
zzcclp: It's necessary, the reason is the same as the putting method, shown comments above.

zzcclp:
@beinan @apc999 I have modified the code according to the comments, there are some test cases failing, but they are not related to this PR, please help to review, thanks.

zzcclp:
PR is ready, @beinan @apc999 please help to review again, thanks.

zzcclp:
ping @beinan @apc999

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
beinan: but you always make a copy in the put method, I don't think the byte array would be got modified unless there is some other client to modify it (I don't think presto will modify any bytes array it reads from)

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
beinan: hmmm, do you happen to know when and how the original btye array could be modified?  Thanks!

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * @param pageId page Id
+   * @return the key to this page
+   */
+  @VisibleForTesting
+  public long getKeyFromPageId(PageId pageId) {
+    // TODO(feng): encode fileId with URLEncoder to escape invalid characters for file name
+    // Key is : MD5(file_id(str)_page_idx(ulong))
```
beinan, line 107: You're able to use the pageId direclty as the key, the extra md5 hashing is unnecessary which might cause conflict in some extreme cases.

See the hashcode and equals method in the PageId class

` @Override
  public int hashCode() {
    return Objects.hashCode(mFileId, mPageIndex);
  }

  @Override
  public boolean equals(Object obj) {
    if (this == obj) {
      return true;
    }
    if (!(obj instanceof PageId)) {
      return false;
    }
    PageId that = (PageId) obj;
    return mFileId.equals(that.mFileId) && mPageIndex == that.mPageIndex;
  }`

zzcclp: Got it. I misunderstood your means of comment before.
Done.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
zzcclp: For example:
```
        InputStream is = new FileInputStream("XXXX/alluxio/NOTICE");
        ByteArrayOutputStream buffer = new ByteArrayOutputStream();

        int nRead;
        long cnt = 0;
        byte[] data = new byte[4];   // read buffer

        while ((nRead = is.read(data, 0, data.length)) != -1) {
            // put data into memory cache
            mPageDataMap.put(cnt, data);
            cnt++;
            buffer.write(data, 0, nRead);
        }

        buffer.flush();
```
Use the byte array as a buffer, read data from a file and put the data into memory cache, at the end, all the cached data is the last one read from the file.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
zzcclp: > I don't think the byte array would be got modified unless there is some other client to modify it (I don't think presto will modify any bytes array it reads from)

Yeah, like Spark, Presto, they will not modify any data, but we can not guarantee all of the systems will not modify any data after reading from, shown as below:
```
        System.out.println(Arrays.toString(mPageDataMap.get(1l)));
        System.out.println(Arrays.toString(mPageDataMap.get(2l)));

        // read data from memory cache data
        byte[] data1 = mPageDataMap.get(1l);
        // modify the data directly
        data1[0] = 2;
        byte[] data2 = mPageDataMap.get(2l);
        // modify the data directly
        data2[1] = 3;
        
        System.out.println(Arrays.toString(mPageDataMap.get(1l)));
        System.out.println(Arrays.toString(mPageDataMap.get(2l)));
```

the results are:

image

zzcclp:
ping @beinan @apc999

zzcclp:
PR is ready, ping @beinan @apc999, please help to review, thanks.

core/client/fs/src/main/java/alluxio/client/file/cache/store/MemoryPageStore.java
beinan: gotcha, let's keep the array copy for now,  we could improve it if we saw any GC or memory issues.  Also thinking to make the bytes immutable, but it's really difficult to make an immutable array in java.

beinan (approved):
lgtm,  great contribution!  Thanks a lot!
@apc999 could you help take a look also?

beinan:

